PR #3, opened 2023-03-02
monad-p2p boilerplate

omegablitz:
Boilerplate for monad-p2p

Will figure out the `libp2p::PeerId` -> `monad_*::PeerId` conversion in a later PR.

### DIFF
--- a/monad-p2p/src/behavior/codec.rs
+++ b/monad-p2p/src/behavior/codec.rs
@@ -1,0 +1,89 @@
+use std::{io::ErrorKind, marker::PhantomData, sync::Arc};
+
+use async_trait::async_trait;
+use monad_executor::{Message, Serializable};
+
+use futures::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
+use libp2p::request_response::Codec;
+
+#[derive(Clone)]
+pub(crate) struct ReliableMessageCodec<M>
+where
+    M: Message + Serializable,
+{
+    _marker: PhantomData<M>,
+}
+
+impl<M> Default for ReliableMessageCodec<M>
+where
+    M: Message + Serializable,
+{
+    fn default() -> Self {
+        Self {
+            _marker: PhantomData,
+        }
+    }
+}
+
+#[async_trait]
+impl<M> Codec for ReliableMessageCodec<M>
+where
+    M: Message + Serializable,
+{
+    type Protocol = String;
+    // TODO can we avoid Arc?
+    type Request = Arc<M>;
+    type Response = ();
+
+    async fn read_request<T>(
+        &mut self,
+        protocol: &Self::Protocol,
+        io: &mut T,
+    ) -> std::io::Result<Self::Request>
+    where
+        T: AsyncRead + Unpin + Send,
+    {
+        let mut bytes = Vec::new();
+        io.read_to_end(&mut bytes).await?;
+        let message =
+            M::deserialize(&bytes).map_err(|err| std::io::Error::new(ErrorKind::Other, err))?;
+        Ok(Arc::new(message))
+    }
+
+    async fn read_response<T>(
+        &mut self,
+        protocol: &Self::Protocol,
+        io: &mut T,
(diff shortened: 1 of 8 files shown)


monad-p2p/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+    recipient: PeerId,
+    message_id: MessageId,
+}
+
+pub struct SentHandle {
+    sent: Sent,
+    notifier: Sender<Sent>,
+}
+
+impl Drop for SentHandle {
```
ariqchowdhury: Where does the Drop trait come from or what is it?

omegablitz: Pretty much equivalent to a destructor in C++ - it's built into the language

monad-p2p/src/behavior/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+const IDENTIFY_PROTO_NAME: &str = "/monad/identify/0.0.1";
+const REQUEST_RESPONSE_PROTO_NAME: &str = "/monad/req-res/0.0.1";
+const REQUEST_RESPONSE_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(1);
+const REQUEST_RESPONSE_KEEPALIVE: std::time::Duration = std::time::Duration::from_secs(1);
+
+impl<M> Behavior<M>
+where
+    M: Message + Serializable,
+{
+    pub(crate) fn new(identity: &libp2p::identity::Keypair) -> Self {
```
ariqchowdhury, line 23: nit: maybe just pass in libp2p::identity::PublicKey if we don't have to sign anything here

monad-p2p/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+                            request_id,
+                            response: (),
+                        } => {
+                            if let Some((message, on_ack)) =
+                                service.outbound_messages.remove(&request_id)
+                            {
+                                service
+                                    .outbound_messages_lookup
+                                    .remove(&(peer, message.id()))
+                                    .unwrap();
```
ariqchowdhury: nit: expect with an error message?

monad-p2p/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+
+use futures::{Stream, StreamExt};
+use monad_executor::{Message, Serializable};
+
+use libp2p::{request_response::RequestId, swarm::SwarmBuilder, PeerId, Transport};
+
+mod behavior;
+use behavior::Behavior;
+
+pub struct Service<M>
```
ariqchowdhury, line 11: should we add debug counters in a future PR?

monad-p2p/src/behavior/codec.rs
```diff
@@ -0,0 +1,89 @@
+use std::{io::ErrorKind, marker::PhantomData, sync::Arc};
+
+use async_trait::async_trait;
+use monad_executor::{Message, Serializable};
```
ariqchowdhury, line 4: nit, you can ignore this comment, I got confused by Message since its a generic name -- reading through the code, didn't realize for a while it was our executor trait

monad-p2p/src/lib.rs
omegablitz: Yeah definitely... we just need to decide where the best place to put them is - I think this is related to the whole logging/observability stuff

monad-p2p/src/behavior/mod.rs
omegablitz: It's used in the `authenticate(libp2p::noise::NoiseAuthenticated::xx(identity).unwrap())` for authentication - we will need to revisit this once I figure out how to unify the monad/libp2p KeyPairs

monad-p2p/src/behavior/codec.rs
omegablitz: Hmm, maybe I should use the full type... eg `monad_executor::Message`/`monad_executor::Serializable`

omegablitz:
Merging because will simplify some stuff in #64

PR #4, opened 2023-03-03
Weighted round robin leader election

michael-yxchen:
A weighted round robin leader election for validators. It assumes that the validator list is stable is in an epoch. In every view, it deterministically produces the leader, with the average probability of selection proportional to its stake. Validator stakes can be updated and will take effect on the next view.

Some WIP docs https://www.notion.so/monad-labs/LeaderElection-6f6d4bed2c0c4d058b8770fcc1a1eef4?pvs=4

### DIFF
--- a/monad-validator/src/leader_election.rs
+++ b/monad-validator/src/leader_election.rs
@@ -1,0 +1,10 @@
+use super::validator::Address;
+
+// VotingPower is i64
+pub trait LeaderElection {
+    fn new() -> Self;
+    fn start_new_epoch(&mut self, voting_powers: Vec<(Address, i64)>);
+    fn increment_view(&mut self, view: i64);
+    fn get_leader(&self) -> &Address;
+    fn update_voting_power(&mut self, addr: &Address, new_voting_power: i64) -> bool;
+}
--- a/monad-validator/src/validator.rs
+++ b/monad-validator/src/validator.rs
@@ -1,0 +1,21 @@
+use std::fmt::Debug;
+
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug)]
+pub struct Address(pub i64); // placeholder type for Address
+#[derive(Clone, Copy, Debug)]
+pub struct PubKey(pub i64); // placeholder type
+#[derive(Clone, Copy)]
+pub struct Stake(pub i64);
+
+#[derive(Clone, Copy)]
+pub struct Validator {
+    pub address: Address,
+    pub pubkey: PubKey,
+    pub stake: Stake,
+}
+
+impl Debug for Validator {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "validator {:?}", self.address.0)
+    }
+}
--- a/monad-validator/src/weighted_round_robin.rs
+++ b/monad-validator/src/weighted_round_robin.rs
@@ -1,0 +1,304 @@
+use super::{leader_election::LeaderElection, validator::Address};
+use log::warn;
+use std::cmp::Ordering;
+use std::fmt::Debug;
+
+#[derive(Eq, Clone, Copy, Debug)]
+struct Voter {
+    address: Address,
+    voting_power: i64,
+    priority: i64,
+}
+
+// ordering Voters first on priority, then on address
+// higher priority -> smaller in Ord
+impl Ord for Voter {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        (self.priority, self.address)
+            .cmp(&(other.priority, other.address))
+            .reverse()
+    }
+}
+
+impl PartialOrd for Voter {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl PartialEq for Voter {
+    fn eq(&self, other: &Self) -> bool {
+        self.address == other.address
+    }
+}
+
+impl Voter {
+    pub fn verified(&self) -> bool {
+        self.voting_power > 0
+    }
+}
+pub struct WeightedRoundRobin {
+    voters: Vec<Voter>,
+    leader: usize, // voter idx
+    total_voting_power: i64,
+}
+
+impl LeaderElection for WeightedRoundRobin {
+    fn new() -> Self {
+        Self {
+            voters: Vec::new(),
+            leader: 0,
+            total_voting_power: 0,
+        }
+    }
+
+    fn start_new_epoch(&mut self, voting_powers: Vec<(Address, i64)>) {
+        self.voters.reserve(voting_powers.len());
+        self.total_voting_power = 0;
+        for (addr, vp) in voting_powers.into_iter() {
+            let voter = Voter {
+                address: addr,
+                voting_power: vp,
+                priority: vp,
+            };
+            if !voter.verified() {
+                warn!("ignoring voter {:?} with zero voting power", voter.address);
+                continue;
+            }
+            self.voters.push(voter);
+            self.total_voting_power += vp;
+        }
+
+        if !self.voters.is_empty() {
+            self.voters.sort();
+            self.increment_one_view();
+        }
+    }
+
+    fn increment_view(&mut self, view: i64) {
+        self.panic_if_empty();
+        for _ in 0..view {
+            self.increment_one_view();
+        }
+    }
+
+    fn get_leader(&self) -> &Address {
+        self.panic_if_empty();
+        &self.voters[self.leader].address
+    }
+
+    fn update_voting_power(&mut self, addr: &Address, new_voting_power: i64) -> bool {
+        self.panic_if_empty();
+        let v = match self.voters.iter_mut().filter(|v| addr == &v.address).next() {
+            Some(v) => v,
+            None => return false,
+        };
+        self.total_voting_power -= v.voting_power;
+        self.total_voting_power += new_voting_power;
+        v.voting_power = new_voting_power;
+        v.priority = -self.total_voting_power;
+        true
+    }
+}
+
+impl WeightedRoundRobin {
+    fn get_highest_priority_validator(&self) -> usize {
+        self.panic_if_empty();
+        self.voters
+            .iter()
+            .enumerate()
+            .min_by(|(_, vx), (_, vy)| vx.cmp(vy))
+            .map(|(idx, _)| idx)
+            .unwrap()
+    }
+
+    fn increment_one_view(&mut self) {
+        for v in self.voters.iter_mut() {
+            v.priority += v.voting_power;
+        }
+        println!("voters: {:?}", self.voters);
+        self.leader = self.get_highest_priority_validator();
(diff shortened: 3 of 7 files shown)


omegablitz:
I'll go through the changes in more detail tomorrow to leave comments, but one quick thing: the validator stuff can go into a new top-level crate called monad-validator or something, I think it makes sense for our monad-crypto crate to just have our base crypto-related things that everything will depend on.

michael-yxchen:
Ah I see. Will put that into a separate crate. I was misinterpreting that as the crate everything can fit in

monad-validator/src/leader_election.rs
```diff
@@ -0,0 +1,8 @@
+use super::validator::Validator;
+pub trait LeaderElection {
+    fn new() -> Self;
+    fn start_new_epoch(&mut self, validators: Vec<Validator>);
+    fn increment_view(&mut self, view: i64);
+    fn get_proposer(&self) -> &Validator;
```
ariqchowdhury: nit: get_leader

monad-validator/src/leader_election.rs
```diff
@@ -0,0 +1,8 @@
...
+    fn update_stake(&mut self, update: Validator) -> bool;
```
ariqchowdhury: does it make sense for LeaderElection to be controlling the updating of stake? Could this be separated out as its sort of a separate concern

monad-validator/src/weighted_round_robin.rs
```diff
@@ -0,0 +1,274 @@
+use super::{
+    leader_election::LeaderElection,
+    validator::{Stake, Validator},
+};
+use log::warn;
+
+pub struct WeightedRoundRobin {
```
ariqchowdhury, line 40: I feel like Validator should own stake and updating stake and have a policy for leader election.
something like:
struct validator_set {
  v: Vec<Validator>,
  election: LeaderElection,
  staking: Stake
  etc...
}

monad-validator/src/leader_election.rs
```diff
@@ -0,0 +1,9 @@
+pub type VotingPower = i64;
+use super::validator::Address;
```
omegablitz, line 1: nit: imports at top

monad-validator/src/leader_election.rs
```diff
@@ -0,0 +1,9 @@
+pub type VotingPower = i64;
```
omegablitz: nit: NewType instead of type alias (eg `pub struct VotingPower(i64)`)

monad-validator/src/leader_election.rs
```diff
@@ -0,0 +1,9 @@
...
+use super::validator::Address;
+pub trait LeaderElection {
+    fn new() -> Self;
+    fn start_new_epoch(&mut self, addresses: Vec<Address>, voting_powers: Vec<VotingPower>);
```
omegablitz: nit: maybe `voting_powers: Vec<(Address, VotingPower)>` to enforce at compile-time that these are the same length instead of panic'ing at runtime if they're not

monad-validator/src/validator.rs
```diff
@@ -0,0 +1,18 @@
+use std::fmt::Debug;
+
+pub type Address = i64; // placeholder type for Address
+pub type PubKey = i64; // placeholder type
+pub type Stake = i64;
```
omegablitz: nit: newtype

monad-validator/src/validator.rs
```diff
@@ -0,0 +1,18 @@
...
+
+#[derive(Clone, Copy)]
+pub struct Validator {
+    pub address: Address,
+    pub pubkey: PubKey,
+    pub stake: Stake,
+}
+
+impl Debug for Validator {
```
omegablitz, line 17: Is there a reason not to derive Debug instead?

monad-validator/src/weighted_round_robin.rs
```diff
@@ -0,0 +1,326 @@
...
+use std::fmt::Debug;
+type VoterPriority = i64;
+#[derive(Eq, Clone, Copy)]
+struct Voter {
+    address: Address,
+    voting_power: VotingPower,
+    priority: VoterPriority,
+}
+
+impl Debug for Voter {
```
omegablitz: Is there a reason not to derive Debug?

monad-validator/src/weighted_round_robin.rs
```diff
@@ -0,0 +1,326 @@
+use super::{
+    leader_election::{LeaderElection, VotingPower},
+    validator::Address,
+};
+use log::warn;
+use std::cmp::Ordering;
+use std::fmt::Debug;
+type VoterPriority = i64;
```
omegablitz: nit: newtype

monad-validator/src/weighted_round_robin.rs
```diff
@@ -0,0 +1,326 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "validator {:?} pri: {}", self.address, self.priority)
+    }
+}
+
+// ordering Voters first on priority, then on address
+// higher priority -> smaller in Ord
+impl Ord for Voter {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        match self.priority.cmp(&other.priority).reverse() {
```
omegablitz: Can you do something like `(other.priority, self.address).cmp((self.priority, other.address))` instead?

If the ordering by address doesn't matter you could also do something like `(self.priority, self.address).cmp((other.priority, other.address)).reverse()`

monad-validator/src/weighted_round_robin.rs
```diff
@@ -0,0 +1,326 @@
...
+    }
+}
+
+impl PartialOrd for Voter {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl PartialEq for Voter {
```
omegablitz, line 29: This implementation of `PartialEq` is a little strange to me - I understand the purpose but semantically it doesn't make that much sense to me for 2 Voters with different voting_power to be ==.

For instance, this would have implications if we were adding `Voter`s into a hashset or whatnot.

monad-validator/src/weighted_round_robin.rs
```diff
@@ -0,0 +1,326 @@
...
+
+pub struct WeightedRoundRobin {
+    voters: Vec<Voter>,
+    leader: usize, // voter idx
+    total_voting_power: VotingPower,
+}
+
+impl LeaderElection for WeightedRoundRobin {
+    fn new() -> Self {
+        WeightedRoundRobin {
```
omegablitz: nit: you could do `Self { ... }` here

monad-validator/src/weighted_round_robin.rs
```diff
@@ -0,0 +1,326 @@
...
+            voters: Vec::new(),
+            leader: 0,
+            total_voting_power: 0,
+        }
+    }
+
+    fn start_new_epoch(&mut self, addresses: Vec<Address>, voting_powers: Vec<VotingPower>) {
+        if addresses.len() != voting_powers.len() {
+            panic!("addresses.len() != voting_powers.len()");
```
omegablitz: nit: this panic could be avoided if the input was a `Vec<(_, _)>`

monad-validator/src/weighted_round_robin.rs
```diff
@@ -0,0 +1,326 @@
...
+        }
+        self.voters.reserve(addresses.len());
+        self.total_voting_power = 0;
+        for it in addresses.iter().zip(voting_powers.iter()) {
```
omegablitz: nit: `for (addr, vp) in ...`
nit: `addresses.into_iter().....` - because this is the last usage of `addresses`, we can use an owned iter instead of a ref iter.

monad-validator/src/weighted_round_robin.rs
```diff
@@ -0,0 +1,326 @@
...
+            };
+            if !voter.verified() {
+                warn!("ignoring voter {} with zero voting power", voter.address);
+                continue;
+            }
+            self.voters.push(voter);
+            self.total_voting_power += vp;
+        }
+
+        if self.voters.len() > 0 {
```
omegablitz: nit: entirely your own preference, but noting here just so you're aware, I more commonly see `if !self.voters.is_empty()`

monad-validator/src/weighted_round_robin.rs
```diff
@@ -0,0 +1,326 @@
...
+            self.increment_one_view();
+        }
+    }
+
+    fn get_leader(&self) -> &Address {
+        self.panic_if_empty();
+        &self.voters[self.leader].address
+    }
+
+    fn update_voting_power(&mut self, addr: Address, new_voting_power: VotingPower) -> bool {
```
omegablitz: nit: can this be `addr: &Address`?

monad-validator/src/weighted_round_robin.rs
```diff
@@ -0,0 +1,326 @@
...
+        self.panic_if_empty();
+        for v in self.voters.iter_mut() {
```
omegablitz: nit: I think it's more idiomatic to do `self.voters.iter_mut().filter(...).next()`

monad-validator/src/weighted_round_robin.rs
```diff
@@ -0,0 +1,326 @@
...
+        false
+    }
+}
+
+impl WeightedRoundRobin {
+    fn get_highest_priority_validator(&self) -> usize {
+        self.panic_if_empty();
+        let mut hidx: usize = 0;
+        let mut h = &self.voters[hidx];
+        for (i, v) in self.voters.iter().enumerate() {
```
omegablitz: nit: I think it's more idiomatic to do `self.voters.iter().min().unwrap()`

monad-validator/src/weighted_round_robin.rs
```diff
@@ -0,0 +1,326 @@
...
+    fn increment_one_view(&mut self) {
+        for v in self.voters.iter_mut() {
+            v.priority += v.voting_power;
+        }
+        self.leader = self.get_highest_priority_validator();
+        self.voters[self.leader].priority -= self.total_voting_power;
+    }
+
+    fn panic_if_empty(&self) {
+        if self.voters.len() == 0 {
```
omegablitz: nit: `self.voters.is_empty()`

monad-validator/src/validator.rs
michael-yxchen: the intention was to create shorter debug messages: address is derived from pubkey, so printing one of the two would be enough

monad-validator/src/weighted_round_robin.rs
michael-yxchen: Should've derived Debug here

monad-validator/src/weighted_round_robin.rs
michael-yxchen: That would mean there could be `Voter` with the same address but different `voting_power` in a hashset. Would it make sense to use this implementation because we should only have unique (by address) voters?

PR #5, opened 2023-03-07
monad-blockstore

michael-yxchen:
Interface to RocksDB for blockstore. Any suggestion is welcomed

Some future improvements on the draft
- We might need more abstractions around the database as we add more ColumnFamily to the store.
- We would need to dive deep into the db options. The defaults are used for now.
- Error handling

WIP doc: https://www.notion.so/monad-labs/Blockstore-7b52e2bbff98440fa235271ba517f16c?pvs=4

### DIFF
--- a/monad-blockstore/src/store_db.rs
+++ b/monad-blockstore/src/store_db.rs
@@ -1,0 +1,97 @@
+use crate::{
+    store::Result,
+    types::{BlockHeight, BlockMeta, BlockPart, DoubleQC, H256, TC},
+};
+use ordcode;
+use rocksdb::{ColumnFamily, WriteBatch as RocksWriteBatch, DB};
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+// TODO: create an abstraction around database to fit different underlying interface
+
+// ColumnFamily names
+const BLOCK_PART_CF: &str = "block_part";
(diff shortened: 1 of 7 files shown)


monad-blockstore/src/store_db.rs
```diff
@@ -0,0 +1,95 @@
...
+// TODO: create an abstraction around database to fit different underlying interface
+
+// ColumnFamily names
+const BLOCK_PART_CF: &str = "block_part";
+const BLOCK_META_CF: &str = "block_meta";
+const QC_CF: &str = "qc";
+const TC_CF: &str = "tc";
+const BLOCK_HEIGHT_CF: &str = "block_height";
+
+pub fn cfs() -> Vec<&'static str> {
```
michael-yxchen: switching to static array

michael-yxchen:
yes, they should be changed to use monad-consensus types. We would need some (de)serialization for the types. I think `bincode` and `ordcode` should work fine before we have our own database storage format

michael-yxchen:
It might make more sense to swap the types later. 

We might want to go over
- if we want to a block into blockparts and store them as parts in the db
- TimeoutCertificate formats

to figure out the ColumnFamilies we need for the database.

michael-yxchen:
Replaced by persistence

PR #6, opened 2023-03-15
Initial definitions for consensus types

ariqchowdhury:
Some of the types have an Option<Signature> -- that will be replaced in a future PR with types T that are considered unsigned and a `trait Signable` where a signer function can create a signed version of the the type.

### DIFF
--- a/monad-consensus/src/types/block.rs
+++ b/monad-consensus/src/types/block.rs
@@ -45,7 +45,7 @@ impl<T: VotingQuorum> Block<T> {
         hasher.update(self.author);
         hasher.update(self.round);
         hasher.update(&self.payload.0);
-        hasher.update(self.qc.info.vote.id);
+        hasher.update(self.qc.info.vote.id.0);
         hasher.update(self.qc.signatures.get_hash());
 
         self.id = hasher.finalize().into();
--- a/monad-consensus/src/types/voting.rs
+++ b/monad-consensus/src/types/voting.rs
@@ -21,9 +21,9 @@ use sha2::Digest;
 
 #[derive(Copy, Clone, Debug, Default)]
 pub struct VoteInfo {
-    pub id: NodeId,
+    pub id: BlockId,
     pub round: Round,
-    pub parent_id: NodeId,
+    pub parent_id: BlockId,
     pub parent_round: Round,
 }
 
@@ -31,9 +31,9 @@ impl VoteInfo {
     // TODO will use something from monad-crypto later...
     pub fn get_hash(&self) -> Hash {
         let mut hasher = sha2::Sha256::new();
-        hasher.update(self.id);
+        hasher.update(self.id.0);
         hasher.update(self.round);
-        hasher.update(self.parent_id);
+        hasher.update(self.parent_id.0);
         hasher.update(self.parent_round);
 
         hasher.finalize().into()
--- a/monad-consensus/src/lib.rs
+++ b/monad-consensus/src/lib.rs
@@ -23,3 +23,7 @@ impl AsRef<[u8]> for NodeId {
         self.0.as_bytes()
     }
 }
+
+#[repr(transparent)]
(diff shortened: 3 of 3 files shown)


monad-consensus/src/types/block.rs
```diff
@@ -0,0 +1,53 @@
...
+        };
+        b.hash();
+        b
+    }
+
+    pub fn get_id(&self) -> Hash {
+        self.id
+    }
+
+    fn hash(&mut self) {
```
michael-yxchen, line 43: It might be helpful return `Hash` from this function and make it public (or have function to to check if the block is wellformed) so the receiver can verify the hash against its `id`.

monad-consensus/src/types/quorum_certificate.rs
```diff
@@ -0,0 +1,98 @@
...
+            signatures: Default::default(),
+            author: Default::default(),
+            author_signature: None,
+        }
+    }
+}
+
+impl<T: VotingQuorum> PartialEq for QuorumCertificate<T> {
+    fn eq(&self, other: &Self) -> bool {
+        self.vote_info.round == other.vote_info.round
```
michael-yxchen: Similar to the comment Aashish left me [here](https://github.com/monad-crypto/monad-bft/pull/4#discussion_r1127078499). Still thinking what would make the most sense there and here..

monad-consensus/src/types/timeout.rs
```diff
@@ -0,0 +1,75 @@
...
+}
+
+#[derive(Clone, Debug)]
+pub struct TimeoutCertificate {
+    pub round: Round,
+    pub high_qc_rounds: Vec<HighQcRound>,
+}
+
+impl TimeoutCertificate {
+    pub fn max_round(&self) -> Round {
```
michael-yxchen, line 31: Another way to write it would be
```
high_qc_rounds
            .iter()
            .map(|x| x.qc_round)
            .max_by(|x, y| x.cmp(&y))
            .unwrap_or(Round(0));
```
Not sure which way is more idiomatic

monad-consensus/src/types/block.rs
ariqchowdhury: yeah, I'll add functions for checking well-formedness of everything that needs it in another pr.

monad-consensus/src/types/quorum_certificate.rs
ariqchowdhury: Yeah, I'm using the rank (Jolteon paper terminology) for comparing QCs. 
Equality looks confusing, agreed, but QC requires 2f+1 votes so you can't have 2 different valid QCs at the same round. 
We can think about this more later...if you received conflicting proposals with 2 QCs which are equal rank but have different content, then there are more than f byzantine nodes...for that round, I guess an honest node should try and initiate a view change since it can't vote in that round?

monad-consensus/src/types/timeout.rs
ariqchowdhury: I'll leave it unless you or Aashish feel strongly about changing it

monad-consensus/src/types/quorum_certificate.rs
ariqchowdhury: I pushed a commit to wrap the qcinfo in a Rank type and then do comparisons on Rank so its more clear.

monad-consensus/src/types/voting.rs
```diff
@@ -0,0 +1,41 @@
...
+
+use sha2::Digest;
+
+#[derive(Copy, Clone, Debug, Default)]
+pub struct VoteInfo {
+    pub id: NodeId,
+    pub round: Round,
+    pub parent_id: NodeId,
+    pub parent_round: Round,
+}
```
michael-yxchen, line 28: I think `id` and `parent_id` are referring to blocks, so they would have type `Hash`

PR #8, opened 2023-03-17
Ariq/consensus verifcation

ariqchowdhury:
Take a look at the Signable trait and the implementation for VoteMessage. 
The test in signing.rs also shows an example of a signer that would use the Signable trait to create SignedVoteMessage.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -99,8 +99,10 @@ dependencies = [
 name = "monad-consensus"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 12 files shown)


monad-consensus/src/types.rs
omegablitz, line 1: Rust nit: I think it would be more idiomatic to put this stuff in `monad-consensus/src/types/mod.rs`, I think it's functionally equivalent as this but the grouping just makes more sense.

PR #9, opened 2023-03-20
monad-executor

omegablitz:
Adds support for our monad-executor abstraction (inspired by elm architecture)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 3
 
 [[package]]
+name = "autocfg"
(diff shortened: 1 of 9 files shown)


omegablitz:
Ready for review, still have some TODOs:

- [x] only include tokio features that we need
- [x] feature-gate tokio
- [x] clean up lib.rs to actually make sense - right now it's mainly just an example
- [x] more unit tests


notes: https://www.notion.so/monad-labs/monad-executor-13f78fbcaa2c405ead2a133d6687c9b8

PR #10, opened 2023-03-20
Blocktree structure

michael-yxchen:
A tree structure to store pending blocks. A block contains a QC for its parent, i.e. its parent block is `block.qc.vote_info.id`. `prune` moves the root to the given block and deallocates all non-descendant blocks. 

Modifies `consensus-types` to expose some type/fields

### DIFF
--- a/monad-blocktree/src/blocktree.rs
+++ b/monad-blocktree/src/blocktree.rs
@@ -1,0 +1,506 @@
+use monad_consensus::types::block::Block;
+use monad_consensus::types::voting::VotingQuorum;
+use monad_consensus::BlockId;
+use monad_consensus::Round;
+use ptree::print_tree;
+use std::collections::HashMap;
+use std::fmt;
+use std::result::Result as StdResult;
+
+use ptree::builder::TreeBuilder;
+
+type Result<T> = StdResult<T, BlockTreeError>;
+
+#[derive(Debug, PartialEq)]
+#[non_exhaustive]
+pub enum BlockTreeError {
+    ParentNotPresent(String),
+    BlockNotExist(String),
+}
+
+impl fmt::Display for BlockTreeError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::ParentNotPresent(s) => write!(f, "Parent not present: {}", s),
+            Self::BlockNotExist(s) => write!(f, "Block not exist: {}", s),
+        }
+    }
+}
+
+impl std::error::Error for BlockTreeError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        None
+    }
+}
+
+pub struct BlockTreeBlock<T>
+where
+    T: VotingQuorum,
+{
+    block: Block<T>,
+    children: Vec<BlockId>,
+}
+
+impl<T: VotingQuorum> BlockTreeBlock<T> {
+    const CHILD_INDENT: &str = "    ";
+    fn new(b: Block<T>) -> Self {
+        BlockTreeBlock {
+            block: b,
+            children: Vec::new(),
+        }
+    }
+
+    fn tree_fmt(
+        &self,
+        tree: &BlockTree<T>,
+        f: &mut fmt::Formatter<'_>,
+        indent: &String,
+    ) -> std::fmt::Result {
+        write!(
+            f,
+            "{}({:?}){:?}\n",
+            indent,
+            self.block.round.0,
+            self.block.get_id()
+        )?;
+        let mut child_indent: String = indent.clone();
+        child_indent.push_str(Self::CHILD_INDENT);
+        for bid in self.children.iter() {
+            let block = tree.tree.get(bid).unwrap();
+            block.tree_fmt(tree, f, &child_indent)?;
+        }
+        Ok(())
+    }
+
+    fn build_ptree<'a>(
+        &self,
+        tree: &BlockTree<T>,
+        tree_print: &'a mut TreeBuilder,
+    ) -> &'a mut TreeBuilder {
+        let mut result =
+            tree_print.begin_child(format!("({}){:?}", self.block.round.0, self.block.get_id()));
+        for bid in self.children.iter() {
+            let block = tree.tree.get(bid).unwrap();
+            result = block.build_ptree(tree, result);
+        }
+        result.end_child()
+    }
+}
+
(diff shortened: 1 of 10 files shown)


monad-blocktree/src/blocktree.rs
```diff
@@ -0,0 +1,238 @@
...
+        fn current_voting_power(&self) -> i64 {
+            0
+        }
+
+        fn get_hash(&self) -> Hash {
+            Default::default()
+        }
+
+        fn add_signature(&mut self, _s: ConsensusSignature) {}
+    }
```
ariqchowdhury: I think this was also used in a test in monad-consensus...maybe we can extract this into a test helpers?

monad-blocktree/src/blocktree.rs
```diff
@@ -0,0 +1,238 @@
...
+        let genesis = Rc::new(RefCell::new(BlockTreeBlock::<T>::new(genesis_block)));
+        let mut lookup = HashMap::new();
+        lookup.insert(genesis.borrow().get_id(), Rc::downgrade(&genesis));
+        BlockTree {
+            root: genesis,
+            lookup: lookup,
+        }
+    }
+
+    pub fn prune(&mut self, new_root: &BlockId) -> Result<()> {
```
ariqchowdhury: as discussed, return vector of blocks for the branch that is being committed for this prune

monad-blocktree/src/blocktree.rs
```diff
@@ -0,0 +1,238 @@
...
+use std::fmt;
+use std::rc::Rc;
+use std::rc::Weak;
+use std::result::Result as StdResult;
+
+type Result<T> = StdResult<T, BlockTreeError>;
+
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum BlockTreeError {
```
ariqchowdhury, line 16: can you implement a debug fmt to draw the graph?

michael-yxchen: Pushed a function to draw the graph. 

It's definitely not the cleanest way but just a quick one: block `children` information is purely for debugging purposes in this BlockTree implementation, and the printing won't be captured by `cargo test` under default args (no `--nocapture`).

Planning on implementing the `Debug` trait for BlockTree with all the tree rendering logic so it can be properly captured by test

michael-yxchen:
Yeah, I'll work on tests for the later two cases.

The BlockTree can never be empty. It's initialized with the genesis block and at least `new_root` is in the tree after pruning. (actually exactly three blocks because of the commit rule)

michael-yxchen:
The assertion removes the need for BlockExist check: we can never have a valid block from a higher round that's the same as a block we have in the tree. So removing that error and test case.

monad-blocktree/src/blocktree.rs
```diff
@@ -0,0 +1,487 @@
...
+            f,
+            "{}({}){:?}\n",
+            indent,
+            self.block.round.0,
+            self.block.get_id()
+        )?;
+        let mut child_indent: String = indent.clone();
+        child_indent.push_str(Self::CHILD_INDENT);
+        for bid in self.children.iter() {
+            if let Some(b) = tree.tree.get(bid) {
```
omegablitz: Can this happen without a broken invariant? If not, should unwrap

monad-blocktree/src/blocktree.rs
```diff
@@ -0,0 +1,487 @@
...
+
+    fn build_ptree<'a>(
+        &self,
+        tree: &BlockTree<T>,
+        tree_print: &'a mut TreeBuilder,
+    ) -> &'a mut TreeBuilder {
+        let mut result =
+            tree_print.begin_child(format!("({}){:?}", self.block.round.0, self.block.get_id()));
+        for bid in self.children.iter() {
+            if let Some(child) = tree.tree.get(bid) {
```
omegablitz: Can this happen without a broken invariant? If not, should unwrap

monad-blocktree/src/blocktree.rs
```diff
@@ -0,0 +1,487 @@
...
+    T: VotingQuorum,
+{
+    root: BlockId,
+    block_ids: VecDeque<BlockId>,
+    tree: HashMap<BlockId, BlockTreeBlock<T>>,
+}
+
+impl<T: VotingQuorum> std::fmt::Debug for BlockTree<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if let Some(root) = self.tree.get(&self.root) {
```
omegablitz: Can this happen without a broken invariant? If not, should unwrap

monad-blocktree/src/blocktree.rs
```diff
@@ -0,0 +1,487 @@
...
+        while cur != self.root {
+            let block = self.tree.remove(&cur).unwrap();
+            cur = block.block.qc.info.vote.id;
+            commit.push(block.block);
+        }
+        self.root = *new_root;
+
+        // garbage collect old blocks
+        // should only keep `n`, `n+1` and `n+2` blocks
+        while let Some(bid) = self.block_ids.front() {
```
omegablitz: Can this be None without a broken invariant?

monad-blocktree/src/blocktree.rs
```diff
@@ -0,0 +1,487 @@
...
+        if self.tree.contains_key(&parent_bid) == false {
+            return Err(BlockTreeError::ParentNotPresent(format!(
+                "{:?}",
+                parent_bid
+            )));
+        }
+
+        self.tree
+            .entry(*parent_bid)
+            .and_modify(|e| e.children.push(new_bid));
```
omegablitz: I don't think it makes sense to use and_modify here, because an invariant is broken if the entry is nonexistent.

BTW can collapse this and the above `contains_key` check into: https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.get_mut

monad-blocktree/src/blocktree.rs
```diff
@@ -0,0 +1,487 @@
...
+        self.tree.insert(new_bid, BlockTreeBlock::new(b));
+        self.block_ids.push_back(new_bid);
+
+        Ok(())
+    }
+
+    pub fn debug_print(&self) -> std::io::Result<()> {
+        let mut tree = TreeBuilder::new("BlockTree".to_owned());
+        if let Some(root) = self.tree.get(&self.root) {
```
omegablitz: Can this be None without a broken invariant?

PR #13, opened 2023-03-26
monad-state boilerplate

omegablitz:
Boilerplate for `MessageState` - the thing that'll handle the conversion to commands from sends/broadcasts/acks.

Important stuff (and tests) is in `monad-state/src/message.rs`

### DIFF
--- a/monad-state/src/message.rs
+++ b/monad-state/src/message.rs
@@ -1,0 +1,220 @@
+use std::collections::{HashSet, VecDeque};
+
+use monad_executor::{Message, PeerId};
+
+pub struct MessageState<M: Message> {
+    messages: VecDeque<HashSet<(PeerId, M::Id)>>,
+    round: u64, // round # of messages.back()
+    // min round == round - messages.len() + 1
+    peers: Vec<PeerId>,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct MessageActionPublish<M>
+where
+    M: Message,
+{
+    pub to: PeerId,
+    pub message: M,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct MessageActionUnpublish<M>
+where
+    M: Message,
+{
+    pub to: PeerId,
+    pub id: M::Id,
+}
+
+impl<M> MessageState<M>
+where
+    M: Message,
+{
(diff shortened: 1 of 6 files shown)


monad-state/src/message.rs
```diff
@@ -0,0 +1,218 @@
...
+
+    pub fn set_round(&mut self, round: u64, peers: Vec<PeerId>) -> Vec<MessageActionUnpublish<M>> {
+        assert!(round > self.round);
+        let mut commands = Vec::new();
+
+        // TODO this can be short-circuited for (round - self.round) > self.messages.len()
+        while round > self.round {
+            self.messages.rotate_left(1);
+            let back = self.messages.back_mut().unwrap();
+            for (peer, message_id) in back.drain() {
```
ariqchowdhury, line 70: I might not be understanding this; rotate_left(1) is going to move the first element to the end, and then you get it via back_mut and then back.drain() -- so is this the same as calling pop_front()?

monad-state/src/message.rs
```diff
@@ -0,0 +1,218 @@
+use std::collections::{HashSet, VecDeque};
+
+use monad_executor::{Message, PeerId};
+
+pub struct MessageState<M: Message> {
+    messages: VecDeque<HashSet<(PeerId, M::Id)>>,
```
ariqchowdhury, line 6: My understanding is that we're using the VecDeque as a ring buffer for messages. It's not contiguous, probably not worth thinking about the performance now, maybe just add a task to remember to benchmark this with real message workloads

omegablitz: Just added a `make_contiguous` call on init to make it contiguous

monad-state/src/message.rs
```diff
@@ -0,0 +1,220 @@
...
+        self.round
+    }
+
+    pub fn set_round(&mut self, round: u64, peers: Vec<PeerId>) -> Vec<MessageActionUnpublish<M>> {
+        assert!(round > self.round);
+        let mut commands = Vec::new();
+
+        // TODO this can be short-circuited for (round - self.round) > self.messages.len()
+        while round > self.round {
+            self.messages.rotate_left(1);
```
ariqchowdhury, line 68: I guess this is expensive now that we made it contiguous...
can come back to it later though

PR #14, opened 2023-04-03
Use RecoverableSignature for Signature type

ariqchowdhury:


omegablitz:
Does the Cargo.lock need to be regenerated? (It's possible that it remains the same but I would have thought that it would change)

PR #18, opened 2023-04-04
Add protocol verification

ariqchowdhury:
Verification involves checking the well-formedness of messages as well as verifying the appropriate signatures and authors of messages. Author verification requires checking membership in a validator set.

Verified type is added to wrap Signed types that have passed through validation functions.

---
mock-signing can be moved into monad-testutils if that PR gets accepted
The verified type is a little verbose, `Verified<Signed<QuorumCertificate<T>>>` but I think its okay. Open to suggestions if folks don't like it. 

Also, refactored the Pubkey in monad-validators to use the monad-crypto type (probably should have put that in its own pr)

### DIFF
--- a/monad-consensus/src/validation/message.rs
+++ b/monad-consensus/src/validation/message.rs
@@ -34,19 +34,24 @@ fn well_formed(
     tc: &Option<TimeoutCertificate>,
 ) -> Result<(), Error> {
     let prev_round = round - Round(1);
-    if qc_round == prev_round {
-        if tc.is_none() {
-            Ok(())
-        } else {
-            Err(Error::NotWellFormed)
-        }
-    } else {
-        tc.as_ref().map_or(Err(Error::NotWellFormed), |t| {
-            if t.round == prev_round {
+    let valid_qc_round = qc_round == prev_round;
+
+    match tc {
+        Some(tc) => {
+            // if there is a TC, the qc round must be invalid and the tc round must be valid
+            if !valid_qc_round && tc.round == prev_round {
                 Ok(())
             } else {
                 Err(Error::NotWellFormed)
             }
-        })
+        }
+        None => {
+            // If no TC, the qc round must be valid
+            if valid_qc_round {
+                Ok(())
+            } else {
+                Err(Error::NotWellFormed)
+            }
+        }
     }
 }
--- a/monad-consensus/src/validation/protocol.rs
+++ b/monad-consensus/src/validation/protocol.rs
@@ -143,10 +143,10 @@ where
             .collect::<Vec<(Hash, &ConsensusSignature)>>()
     };
 
-    for i in msg_sig {
-        get_pubkey(&i.0, i.1)?
+    for (hash, sig) in msg_sig {
+        get_pubkey(&hash, sig)?
             .valid_pubkey(validators)?
-            .verify(&i.0, &i.1 .0)
+            .verify(&hash, &sig.0)
             .map_err(|_| Error::InvalidSignature)?;
     }
     Ok(())


monad-consensus/src/validation/signing.rs
```diff
@@ -7,33 +7,33 @@ pub trait Signable {
...
     pub obj: M,
     pub author: NodeId,
     pub author_signature: ConsensusSignature,
 }
 
+pub trait ConsensusSigned {}
+impl<M> ConsensusSigned for Signed<M> where M: Signable {}
+
+pub struct Verified<S: ConsensusSigned> {
+    pub obj: S,
```
omegablitz: This could be error-prone in the sense that anyone can construct a `Verified<S>` given an `S` without actually verifying it

monad-consensus/src/validation/message.rs
```diff
@@ -3,29 +3,46 @@ use crate::*;
...
         }
     } else {
-        tc.is_none()
+        tc.as_ref().map_or(Err(Error::NotWellFormed), |t| {
+            if t.round == prev_round {
+                Ok(())
+            } else {
+                Err(Error::NotWellFormed)
+            }
+        })
```
omegablitz: nit: the boolean logic in this function looks like it could be simplified a lot

monad-consensus/src/validation/protocol.rs
```diff
@@ -0,0 +1,254 @@
...
+            .collect::<Vec<(Hash, &ConsensusSignature)>>()
+    } else {
+        qc.signatures
+            .get_signatures()
+            .into_iter()
+            .map(|s| (qc.signature_hash, s))
+            .collect::<Vec<(Hash, &ConsensusSignature)>>()
+    };
+
+    for i in msg_sig {
```
omegablitz: nit: unpack `i` so code is easier to read

monad-consensus/src/validation/protocol.rs
```diff
@@ -0,0 +1,254 @@
...
+        get_pubkey(&i.0, i.1)?
+            .valid_pubkey(validators)?
+            .verify(&i.0, &i.1 .0)
+            .map_err(|_| Error::InvalidSignature)?;
+    }
+    Ok(())
```
omegablitz, line 152: super nit: everything from the for loop to the end could be simplified to `msg_sig.into_iter().map(...).collect()` (iter of `Result<T, V>` can be collected into `Result<T, V>` directly

omegablitz: actually this isnt even worth noting, resolving this

monad-consensus/src/validation/signing.rs
ariqchowdhury: Yeah, good catch. 
I like your suggestion, added the Verified/Unverified types, only thing is that a type alias doesn't check the generic trait bounds (ie Signable) so I made it a newtype

monad-consensus/src/validation/signing.rs
```diff
@@ -7,76 +7,15 @@ pub trait Signable {
...
 
-pub trait Hashable<'a> {
-    type DataIter: Iterator<Item = &'a [u8]>;
-
-    fn msg_parts(&self) -> Self::DataIter;
-}
-
-pub struct Signed<M> {
+#[derive(Clone, Debug)]
+pub struct Signed<M: Signable, const VERIFIED: bool> {
```
omegablitz, line 11: It seems error-prone that a `Signed<M, true>` can be constructed by anyone. ~~The easiest hotfix for this would be to change the fields of this to be `pub(mod)` so that `Signed<M, true>` could be manually constructed from within this module.~~

~~A better approach may be to make these fields private and only allow the construction of a `Signed<M, false>` via a constructor - and then have functions within `Signed` to generate a `Signed<M, true>` given a `Signed<M, false>`.~~

monad-consensus/src/validation/protocol.rs
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// Extract the PubKey from the Signature if possible
+fn get_pubkey(msg: &[u8], sig: &ConsensusSignature) -> Result<PubKey, Error> {
+    sig.0
+        .recover_pubkey(msg)
+        .map_err(|_| Error::InvalidSignature)
+}
+
+// A verified proposal is one which is well-formed and has valid
```
omegablitz, line 45: These `verify_*` functions feel like they should belong to a trait

omegablitz:
I didn't mean to actually comment those above comments, but since I did, will just post the example traits I was playing with:

```rust
// BEGIN TYPE PLACEHOLDER
pub struct NodeId;
#[derive(PartialEq)]
pub struct ConsensusSignature;
pub struct ProposalMessage;
pub struct VoteMessage;
pub type ValidationResult<T> = Result<T, ()>;
// END TYPE PLACEHOLDER


pub struct Signed<M, const VERIFIED: bool> {
    // note that fields are private to this module
    obj: M,
    author: NodeId,
    author_signature: ConsensusSignature,
}
pub type Unverified<M> = Signed<M, false>;
pub type Verified<M> = Signed<M, true>;

pub trait Signable: Sized {
    // TODO Add whatever other params necessary for signing/verifying
    fn sign(&self, author: &NodeId) -> ConsensusSignature;
    fn verify(signed: &Unverified<Self>) -> bool;
}

impl<M: Signable> Unverified<M> {
    pub fn new(obj: M, author: NodeId, author_signature: ConsensusSignature) -> Self {
        Self {
            obj,
            author,
            author_signature,
        }
    }

    pub fn validate(self) -> ValidationResult<Verified<M>> {
        if Signable::verify(&self) {
            Ok(Verified {
                obj: self.obj,
                author: self.author,
                author_signature: self.author_signature,
            })
        } else {
            Err(())
        }
    }
}

impl<M: Signable> Verified<M> {
    // can directly construct a Verified if you let it generate its own signature
    pub fn new(obj: M, author: NodeId) -> Self {
        let author_signature = obj.sign(&author);
        Self {
            obj,
            author,
            author_signature,
        }
    }
}


// begin message-specific stuff
impl Signable for ProposalMessage {
    fn sign(&self, author: &NodeId) -> ConsensusSignature {
        todo!()
    }
    fn verify(signed: &Unverified<Self>) -> bool {
        todo!()
    }
}
impl Signable for VoteMessage {
    fn sign(&self, author: &NodeId) -> ConsensusSignature {
        todo!()
    }
    fn verify(signed: &Unverified<Self>) -> bool {
        todo!()
    }
}
// end message-specific stuff
```

monad-consensus/src/validation/protocol.rs
omegablitz: Lol I meant to elaborate here, ignore this

ariqchowdhury:
I'll merge this to unblock progress and rethink the signing abstraction in a separate PR. I think it can probably be redone in steps so I'll propose some incremental changes and go from there.

PR #22, opened 2023-04-05
protobuf for Unverified<VoteMessage>

michael-yxchen:
- Map every consensus-type struct to a protobuf struct
- (de)serialize implements glue logic to bridge the two
- PR contains serde for `Unverified<VoteMessage>`, other message types coming in separate PRs

### DIFF
--- a/monad-proto/src/proto/timeout.proto
+++ b/monad-proto/src/proto/timeout.proto
@@ -10,22 +10,3 @@ import "signing.proto";
 message ProtoHighQcRound {
   monad_proto.basic.ProtoRound qc_round = 1;
 }
-
-message ProtoUnverifiedHighQcRound {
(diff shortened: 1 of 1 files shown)


monad-proto/src/proto/timeout.proto
```diff
@@ -0,0 +1,31 @@
...
+
+message ProtoHighQcRound {
+  monad_proto.basic.ProtoRound qc_round = 1;
+}
+
+message ProtoUnverifiedHighQcRound {
+  ProtoHighQcRound high_qc_round = 1;
+  monad_proto.basic.ProtoNodeId author = 2;
+  monad_proto.signing.ProtoSecpSignature author_signature = 3;
+}
```
michael-yxchen: Corrected in #49

PR #24, opened 2023-04-06
Pacemaker

omegablitz:
Initial pacemaker implementation... need to add tests but would still appreciate a skim!


monad-consensus/src/pacemaker.rs
```diff
@@ -0,0 +1,171 @@
...
+pub struct PacemakerTimerExpire {
+    round: Round,
+}
+
+impl<T> Pacemaker<T>
+where
+    T: VotingQuorum,
+{
+    pub fn new(
+        delta: Duration,
```
omegablitz, line 58: Unfortunately can't use Duration w/ const generics

omegablitz:
Good point, fixed

PR #27, opened 2023-04-07
Implement process_qc and process_vote.

ariqchowdhury:
Process vote and pending vote state are implemented in a separate VoteState in monad-consensus.

### DIFF
--- a/monad-state/src/lib.rs
+++ b/monad-state/src/lib.rs
@@ -1,21 +1,26 @@
-use message::MessageState;
 use monad_blocktree::blocktree::BlockTree;
-use monad_consensus::types::message::{TimeoutMessage, VoteMessage};
-use monad_consensus::types::quorum_certificate::QuorumCertificate;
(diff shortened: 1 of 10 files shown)


monad-state/src/lib.rs
```diff
@@ -169,43 +181,96 @@ impl State for MonadState {
...
         todo!();
     }
+
+    // If the qc has a commit_state_hash, commit the parent block and prune the
+    // block tree
+    // Update our highest seen qc (high_qc) if the incoming qc is of higher rank
+    fn process_qc(&mut self, qc: &QuorumCertificate<T>) {
+        match qc.info.ledger_commit.commit_state_hash {
+            Some(_) => {
+                let blocks_to_commit = self.pending_block_tree.prune(&qc.info.vote.parent_id);
```
omegablitz, line 247: nit: more idiomatic to use `.unwrap()` or `.expect("<message>")` for invariants

but we can change this later

PR #29, opened 2023-04-08
Add monad-types crate

ariqchowdhury:
Create a monad-types crate for basic types like Round, NodeId etc. Refactor everything to use the same monad-types


monad-consensus/src/pacemaker.rs
```diff
@@ -141,7 +140,7 @@ where
         let timeouts = self
             .pending_timeouts
             .keys()
-            .map(|node_id| node_id.0)
+            .map(|node_id| node_id.clone()) // TODO: maybe has_honest_vote should take &Vec<&NodeId>
```
omegablitz, line 143: not worth changing but just fyi, there's a `.cloned()` fn u can call on iters

PR #31, opened 2023-04-09
`ConsensusCommand` abstraction

omegablitz:
This PR creates a `ConsensusCommand` which is a superset of the normal commands.

All consensus commands can be mapped to normal commands using what's in MonadState - an example of this is the `Broadcast` command which is mapped using `message_state`.

This file is getting kinda big... will split it up into smaller ones at some point

### DIFF
--- a/monad-state/src/lib.rs
+++ b/monad-state/src/lib.rs
@@ -1,5 +1,8 @@
+use std::time::Duration;
+
 use monad_blocktree::blocktree::BlockTree;
 use monad_consensus::{
+    pacemaker::PacemakerTimerExpire,
     signatures::aggregate_signature::AggregateSignatures,
     types::{
         block::Block,
@@ -15,12 +18,12 @@ use monad_consensus::{
     },
     vote_state::VoteState,
 };
-use monad_executor::{Command, Message, PeerId, RouterCommand, State};
+use monad_executor::{Command, Message, PeerId, RouterCommand, State, TimerCommand};
 use monad_types::Round;
 use monad_validator::leader_election::LeaderElection;
 use monad_validator::{validator_set::ValidatorSet, weighted_round_robin::WeightedRoundRobin};
 
-use message::MessageState;
+use message::{MessageActionPublish, MessageState};
 
 mod message;
 
@@ -41,15 +44,10 @@ pub enum MonadEvent {
         id: <MonadMessage as Message>::Id,
         round: u64,
     },
-    UnverifiedMessage(MonadMessage),
+    ConsensusEvent(ConsensusEvent),
 }
 
-#[derive(Clone)]
-pub enum MonadMessage {
-    Proposal(Unverified<ProposalMessage<SignatureType>>),
-    Vote(Unverified<VoteMessage>),
-    Timeout(Unverified<TimeoutMessage<SignatureType>>),
-}
+type MonadMessage = ConsensusMessage;
 
 impl Message for MonadMessage {
     type Event = MonadEvent;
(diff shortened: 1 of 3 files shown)


monad-state/src/lib.rs
```diff
@@ -97,48 +97,137 @@ impl State for MonadState {
...
-        }
+#[derive(Clone)]
+pub enum ConsensusMessage {
+    Proposal(Unverified<ProposalMessage<SignatureType>>),
+    Vote(Unverified<VoteMessage>),
+    Timeout(Unverified<TimeoutMessage<SignatureType>>),
+}
+
+impl ConsensusMessage {
+    fn round(&self) -> u64 {
```
ariqchowdhury, line 196: can this be monad_types::Round -- there's other pre-existing uses of u64 in this file, so we can also clean it up all at once in a follow up PR

ariqchowdhury: for clarification, is this meant to extract the round from the message in ConsensusMessage? Is it okay that its extracted before verification of the message?

monad-state/src/lib.rs
```diff
@@ -59,6 +57,8 @@ impl Message for MonadMessage {
 
     fn deserialize(from: PeerId, message: &[u8]) -> Result<Self, Self::ReadError> {
         // MUST assert that output is valid and came from the `from` PeerId
+        // `from` must somehow be guaranteed to be staked at this point so that subsequent
```
ariqchowdhury, line 60: the various verify_* functions check that the sender is part of the validator set and can slash for bad messages...is that sufficient?

monad-state/src/lib.rs
```diff
@@ -97,48 +97,137 @@ impl State for MonadState {
...
-                    Ok(p) => self
-                        .consensus_state
-                        .handle_vote_message::<Sha256Hash, LeaderElectionType>(
-                            &p,
-                            &self.validator_set,
-                        ),
-                    Err(_) => todo!(),
-                }
-            }
+#[derive(Clone)]
```
ariqchowdhury: nit: +Debug

monad-state/src/lib.rs
```diff
@@ -97,48 +97,137 @@ impl State for MonadState {
...
-
-                match timeout {
-                    Ok(p) => self
-                        .consensus_state
-                        .handle_timeout_message(&p, &self.validator_set),
-                    Err(_) => todo!(),
-                }
-            }
-        }
+#[derive(Clone)]
```
ariqchowdhury: nit: Debug

monad-state/src/lib.rs
```diff
@@ -97,48 +97,137 @@ impl State for MonadState {
...
+    Timeout(Unverified<TimeoutMessage<SignatureType>>),
+}
+
+impl ConsensusMessage {
+    fn round(&self) -> u64 {
+        todo!()
+    }
+}
+
+pub enum ConsensusCommand {
```
ariqchowdhury, line 202: nit: derive Debug

monad-state/src/lib.rs
omegablitz: Yes, we only actually use it for extracting round from outbound ConsensusMessage.

monad-state/src/lib.rs
omegablitz: Yeah, I guess the main thing I was trying to note was that by the time a validator message reaches the `verify_*` functions, that validator must be staked so that they can be slashed on invalid messages. Right now we don't have a mechanism for asserting that the validators are staked before they reach the `verify_*` functions.

PR #39, opened 2023-04-09
Implement more monad-state functions

ariqchowdhury:
impl process_new_round_event.
add mempool to state.
remove round from ConsensusState and use pacemaker instead


monad-state/src/lib.rs
```diff
@@ -406,19 +418,38 @@ where
...
+            &txns,
+            &self.high_qc,
+        );
+
+        let p = ProposalMessage {
+            block: b,
+            last_round_tc,
+        };
+
+        return Vec::from([ConsensusCommand::Broadcast {
```
omegablitz, line 450: not important, but just FYI there's a `vec![]` macro

PR #49, opened 2023-04-11
protobuf for Unverified<TimeoutMessage>

michael-yxchen:
Based on #22 
- for `Unverified<TimeoutMessage>`

### DIFF
--- a/monad-proto/tests/protobuf.rs
+++ b/monad-proto/tests/protobuf.rs
@@ -1,19 +1,29 @@
 #[cfg(test)]
 mod test {
     use monad_consensus::{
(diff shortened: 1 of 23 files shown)


monad-proto/tests/protobuf.rs
```diff
@@ -59,4 +69,98 @@ mod test {
...
+        };
+
+        let qcinfo = QcInfo {
+            vote: vi,
+            ledger_commit: lci,
+        };
+
+        let qcinfo_hash = Sha256Hash::hash_object(&qcinfo.ledger_commit);
+
+        let privkey1 =
```
michael-yxchen, line 93: Cleaned up in #51

PR #58, opened 2023-04-13
MonadMessage: outbound msg are Verified, inbound are Unverified

michael-yxchen:
- Added `Verified<M>::new(msg, keypair)` which signs message and produces Verified<M>
- Changed the Hashable trait a little
- Split up Outbound and Inbound types in monad-state, `ConsensusMessage::sign` returns the outbound message. `update` should take the inbound message only
- The inbound handling is not done as serde is not integrated in. planning on writing a mock serde: converting `Verified -> Unverified` in place of actual (de)serialization for tests. This will come in a later PR
- `single_node` test passes as usual

### DIFF
--- a/monad-state/src/lib.rs
+++ b/monad-state/src/lib.rs
@@ -74,7 +74,10 @@ pub enum MonadEvent {
 }
 
 #[derive(Debug, Clone)]
-pub struct MonadMessage(SignedConsensusMessage<SignatureType, SignatureCollectionType>);
+pub enum MonadMessage {
+    Verified(VerifiedConsensusMessage<SignatureType, SignatureCollectionType>),
+    Unverified(SignedConsensusMessage<SignatureType, SignatureCollectionType>),
+}
 
 impl Message for MonadMessage {
     type Event = MonadEvent;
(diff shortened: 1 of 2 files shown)


monad-state/src/lib.rs
```diff
@@ -328,29 +350,33 @@ impl<T: SignatureCollection> ConsensusMessage<SignatureType, T> {
...
-                let signature = SignatureType::sign(&hash, &keypair);
-                SignedConsensusMessage::Timeout(Unverified::new(msg, signature))
+                SignedConsensusMessage::OutTimeout(Verified::new::<H>(msg, keypair))
             }
         }
     }
 }
 
 #[derive(Debug, Clone)]
 pub enum SignedConsensusMessage<S, T> {
```
omegablitz: Would it be possible to create a `VerifiedConsensusMessage` and put all the `Verified<_>` types in there? I think the Unverified and Verified wrapped consensus messages are fundamentally different, and so should have separate types.

This will require some changes to `monad-executor/src/state.rs` but I think that's fine...

Basically will just need to add a `type UnverifiedMessage` type and a conversion function from `Message` (outbound) to `UnverifiedMessage` (inbound)

omegablitz: We could also merge this as is and punt the refactoring for down the line if it's too annoying to change now

michael-yxchen: I think I want to refactor it now, before we build more stuff on top it..

michael-yxchen: This should also make the "mock serde" easier

michael-yxchen:
- Created a separate `VerifiedConsensusMessage` type
- Took a slightly different approach, instead of creating a new `UnverifiedMessage` type in `monad-executor/src/state.rs`, changed `MonadMessage` to a enum of Verified and Unverified consensus message. `Message` trait contains both `serialize` and `deserialize`, which I think is better than having unidirectional messages (outbound implements ser only, and inbound implements deser only). But I'm open to any suggestions
    - one downside in implementation is that in `MonadMessage::event` only one branch will get hit: the `Verified` branch in tests, and the `Unverified` branch in production (as deserialized from wire)
- Conversion from `Verified*` to `Signed/Unverified`

omegablitz:
If MonadMessage contains an enum of Verified/Unverified, what will the implementation for serialize look like for Unverified? I can't think of a use-case where we'd want to have serialize implemented for Unverified (so we have a strong compile-time guarantee that all messages we send out are verified), so would that mean we'd put a panic() in the serialization path for Unverified in MonadMessage?

My bias is always towards making errors unrepresentable at compile-time, but honestly it's not that big of a deal - we can revisit later if we feel motivated enough to change it.

michael-yxchen:
you're right, I overlooked that we don't want serialize implemented for Unverified. (or deserialize for Verified)

I'm with you on making errors unrepresentable at compile-time. I'll make the change in this or a separate PR

michael-yxchen:
Merging now for cleaner dependency. Will fix in a later PR

PR #63, opened 2023-04-18
Serializable message_id

omegablitz:
Necessary for #3


omegablitz:
JK I'm gonna fix the TODO comment now... it's grossing me out too much

PR #64, opened 2023-04-19
Serializable trait refactor

michael-yxchen:
- split Serializable trait into Serializable and Deserializable: we want to serialize VerifiedMessage only and deserialize should produce Signed/Unverified message
- not sure how this might affect #3

### DIFF
--- a/monad-executor/src/executor/mock.rs
+++ b/monad-executor/src/executor/mock.rs
@@ -12,10 +12,11 @@ use crate::{state::PeerId, Command, Executor, Message, RouterCommand, TimerComma
 
 use futures::Stream;
 
-pub struct MockExecutor<E, M>
+pub struct MockExecutor<E, M, OM>
 where
     E: Unpin,
     M: Message<Event = E>,
+    OM: Into<M>,
 {
     tick: Duration,
 
@@ -26,7 +27,7 @@ where
     inbound_ack: BinaryHeap<SequencedPeerEvent<M::Id>>,
 
     // caller pop_front outbounds from this (via MockExecutor::receive_*)
-    outbound_messages: VecDeque<(PeerId, M)>,
+    outbound_messages: VecDeque<(PeerId, OM)>,
     outbound_ack: VecDeque<(PeerId, M::Id)>,
 
     sent_messages: HashMap<PeerId, HashMap<M::Id, E>>,
@@ -78,10 +79,11 @@ enum ExecutorEventType {
     Timer,
 }
 
-impl<E, M> MockExecutor<E, M>
+impl<E, M, OM> MockExecutor<E, M, OM>
 where
     E: Unpin,
     M: Message<Event = E>,
+    OM: Into<M>,
 {
     pub fn tick(&self) -> Duration {
         self.tick
@@ -105,7 +107,7 @@ where
             tx_tick,
         });
     }
-    pub fn receive_message(&mut self) -> Option<(PeerId, M)> {
+    pub fn receive_message(&mut self) -> Option<(PeerId, OM)> {
         self.outbound_messages.pop_front()
     }
     pub fn receive_ack(&mut self) -> Option<(PeerId, M::Id)> {
@@ -154,10 +156,11 @@ where
     }
 }
 
-impl<E, M> Default for MockExecutor<E, M>
+impl<E, M, OM> Default for MockExecutor<E, M, OM>
 where
     E: Unpin,
     M: Message<Event = E>,
+    OM: Into<M>,
 {
     fn default() -> Self {
         Self {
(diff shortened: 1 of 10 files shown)


monad-executor/src/state.rs
omegablitz, line 1: Doesn't this also require separate `InboundMessage`/`OutboundMessage` associated types in `State` instead of a single `type Message`? Because otherwise the outbound/inbound types are still the same.

eg something like:
```rust
type OutboundMessage: Into<Self::InboundMessage>;
type InboundMessage: Message<Event = Self::Event>;
```

monad-executor/src/state.rs
```diff
@@ -60,11 +60,17 @@ pub trait State: Sized {
     fn update(&mut self, event: Self::Event) -> Vec<Command<Self::Event, Self::Message>>;
 }
 
-pub trait Serializable: Sized + Send + 'static {
+pub trait Serializable {
```
omegablitz, line 77: It feels more intuitive to me for a Serializable trait to only have a serialize fn instead of also conversion functions

monad-state/src/lib.rs
```diff
@@ -92,33 +94,35 @@ impl std::fmt::Display for MonadReadError {
...
         todo!("proto deserialize")
     }
-
-    fn serialize(&self) -> Vec<u8> {
-        todo!("proto serialize")
-    }
 }
 
-impl Message for MonadMessage {
+impl Message for UnverifiedMonadMessage {
```
omegablitz: Related to above, I think Message should only be implemented for one of these types

monad-executor/src/state.rs
michael-yxchen: Had a first attempt at splitting inbound/outbound message this morning and got confused. Trying again this afternoon

monad-executor/src/state.rs
michael-yxchen: I agree that it's not intuitive. I put it in Serializable because we need this conversion function only if `msg.serialize().deserialize().type != msg.type`. It makes more sense with the Inbound/Outbound types

michael-yxchen:
Another attempt at splitting Inbound and Outbound messages
- Created associated traits for Inbound and Outbound Messages
    - created traits instead of associated types in `State` because `RouterCommand` takes the `OutboundMessage` and `event`, which is a type from the associated inbound type
- `InboundMessage` is like the previous `Message` trait
- I don't like that I can't create a type alias from `InboundMessage` to `OutboundMessage::Id`, but I'm getting associated type default errors trying to do that
- p2p and viz builds are breaking, will fix after confirming that the refactor makes sense

michael-yxchen:
Another try...
- It's cleaner this way without the `OutboundMessage` trait
- Having to do `OutboundMessage.clone().id()` is a little annoying, because we still need the message after getting it's id. Maybe we can lift `Message::Id` to its own trait? Not sure if this is a good idea
- Technically, `State::OutboundMessage` and `State::Message` (previously inbound) can still be the same type...

monad-executor/src/executor/mock.rs
```diff
@@ -12,10 +12,11 @@ use crate::{state::PeerId, Command, Executor, Message, RouterCommand, TimerComma
 
 use futures::Stream;
 
-pub struct MockExecutor<E, M>
+pub struct MockExecutor<E, M, OM>
```
omegablitz, line 15: This is starting to get a bit verbose (`<E, M, OM>`) - we can consider parametrizing this with only `S`.

I was flip flopping on this before when it was just `<E, M>` because on one hand it's more verbose, but on the other it's more explicit, but now that it's getting pretty long it may be beneficial to simplify it.

Don't need to change this in this PR because it's more a matter of taste than anything - we can change this in a later PR if we think it's a worthwhile tradeoff

monad-wal/Cargo.toml
omegablitz, line 1: nit: can probably delete the monad-wal stuff from this PR so the commit history is cleaner

monad-state/src/message.rs
```diff
@@ -1,22 +1,42 @@
...
+use std::{
+    collections::{HashSet, VecDeque},
+    marker::PhantomData,
+};
 
 use monad_executor::{Message, PeerId};
 use monad_types::Round;
 
-pub struct MessageState<M: Message> {
+pub struct MessageState<M: Message, OM: Into<M> + Clone> {
```
omegablitz, line 9: It feels wrong to me for this to be parametrized by 2 things... but I get why you did it.

Only other option I can think of is parametrizing this by S instead, but that might get weird (`MonadState` will contain `MessageState<MonadState>`)

That might be fine though ...? Not sure

monad-wal/Cargo.toml
michael-yxchen: oops, didn't mean to include them

monad-executor/src/executor/mock.rs
michael-yxchen: Noted, will followup in a later PR

michael-yxchen:
Thanks for the feedback! Merging now (after cleaning the wal) and opening a followup PR

PR #69, opened 2023-04-20
Add test for verifying TC

ariqchowdhury:


monad-consensus/src/validation/signing.rs
```diff
@@ -257,3 +277,55 @@ impl ValidatorPubKey for PubKey {
...
+
+        vset.insert(
+            NodeId(keypair.pubkey()),
+            Validator {
+                pubkey: keypair.pubkey(),
+                stake: 1,
+            },
+        );
+
+        let high_qc_rounds = vec![
```
michael-yxchen, line 313: The paper doesn't explicitly cover this, but is `high_qc_round` always smaller than `round`?

ariqchowdhury: it should be because a TC is only valid if its from the round-1 and the high_qc's can't be higher than the round of the TC.

PR #70, opened 2023-04-20
Fix PeerId/PubKey conversions, add more tests

omegablitz:
It turns out that `libp2p::PeerId` is actually protobuf-encoded, meaning that we weren't actually mapping our Pubkeys to the canonical PeerId.

This PR fixes that, and also adds another more exhaustive test that catches this.

### DIFF
--- a/monad-blocktree/src/blocktree.rs
+++ b/monad-blocktree/src/blocktree.rs
@@ -197,10 +197,10 @@ mod test {
     type QC = QuorumCertificate<MockSignatures>;
 
     fn node_id() -> NodeId {
-        let privkey =
+        let mut privkey =
             hex::decode("6fe42879ece8a11c0df224953ded12cd3c19d0353aaf80057bddfd4d4fc90530")
                 .unwrap();
-        let keypair = KeyPair::from_slice(&privkey).unwrap();
+        let keypair = KeyPair::from_bytes(&mut privkey).unwrap();
         NodeId(keypair.pubkey())
     }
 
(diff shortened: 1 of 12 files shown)


monad-crypto/src/secp256k1.rs
```diff
@@ -27,10 +29,33 @@ fn msg_hash(msg: &[u8]) -> secp256k1::Message {
 }
 
 impl KeyPair {
-    pub fn from_slice(keypair: &[u8]) -> Result<Self, Error> {
-        secp256k1::KeyPair::from_seckey_slice(secp256k1::SECP256K1, keypair)
+    pub fn from_bytes(mut secret: impl AsMut<[u8]>) -> Result<Self, Error> {
```
omegablitz, line 32: Changed this to `mut secret` to force the input buffer to be zero'd out.

Inspired by how the libp2p folks do it

monad-blocktree/src/blocktree.rs
```diff
@@ -197,10 +197,10 @@ mod test {
     type QC = QuorumCertificate<MockSignatures>;
 
     fn node_id() -> NodeId {
-        let privkey =
+        let mut privkey =
```
ariqchowdhury, line 200: nit: if you don't care about the value too much and want something more compact,
`let mut privkey: [u8; 32] = [127; 32];`

omegablitz: I'll change this in a subsequent PR

PR #73, opened 2023-04-20
Test cleanup

omegablitz:
Most of these crates don't need the `hex` dep


omegablitz:
Oh right, there was only 1 monad-crypto test that depends on that specific key

PR #75, opened 2023-04-21
Make monad-p2p a RouterCommand executor

omegablitz:
Very simple change - this makes monad-p2p work out of the box with `monad_executor::executor::parent::ParentExecutor`.

This means that after our ser/de stuff is done, we pretty straightforwardly test our consensus on a libp2p-backed test executor.

One unknown: rust-libp2p uses [futures_timer](https://docs.rs/futures-timer/latest/futures_timer/) to drive its timeouts, whereas we use a tokio timer in monad-executor - I don't love the asymmetry there

### DIFF
--- a/monad-executor/src/executor/mock.rs
+++ b/monad-executor/src/executor/mock.rs
@@ -190,7 +190,7 @@ impl<S> Executor for MockExecutor<S>
 where
     S: State,
 {
-    type Command = Command<S>;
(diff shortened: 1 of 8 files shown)


omegablitz:
Rebased, needed to change `Command` back to be parametrized by `M` and `OM` so that a full `State` implementation isn't necessary to test isolated executors.

PR #78, opened 2023-04-24
Lift out Verified<T> around ConsensusMessage

omegablitz:
Had to lift out `Verified<T>` in order to support `impl AsRef<MonadMessage> for OutboundMonadMessage { ... }`.

I think this makes more sense conceptually anyways... verification should wrap the entire enum instead of being inside the enum - this also simplifies a good deal of code.

### DIFF
--- a/monad-consensus/src/pacemaker.rs
+++ b/monad-consensus/src/pacemaker.rs
@@ -11,7 +11,7 @@ use crate::{
         signature::SignatureCollection,
         timeout::{HighQcRound, TimeoutCertificate},
     },
-    validation::{safety::Safety, signing::Verified},
+    validation::safety::Safety,
 };
 
 pub struct Pacemaker<S, T> {
@@ -21,7 +21,7 @@ pub struct Pacemaker<S, T> {
     last_round_tc: Option<TimeoutCertificate<S>>,
 
     // only need to store for current round
-    pending_timeouts: HashMap<NodeId, Verified<S, TimeoutMessage<S, T>>>,
+    pending_timeouts: HashMap<NodeId, (TimeoutMessage<S, T>, S)>,
 
     // used to not duplicate broadcast/tc
     phase: PhaseHonest,
@@ -55,13 +55,12 @@ where
         delta: Duration,
         current_round: Round,
         last_round_tc: Option<TimeoutCertificate<S>>,
-        pending_timeouts: HashMap<NodeId, Verified<S, TimeoutMessage<S, T>>>,
     ) -> Self {
         Self {
             delta,
             current_round,
             last_round_tc,
-            pending_timeouts,
+            pending_timeouts: HashMap::new(),
 
             phase: PhaseHonest::Zero,
         }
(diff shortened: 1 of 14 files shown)


monad-consensus/src/pacemaker.rs
```diff
@@ -123,7 +122,9 @@ where
         validators: &ValidatorSet<L>,
         safety: &mut Safety,
         high_qc: &QuorumCertificate<T>,
-        tmo: Verified<S, TimeoutMessage<S, T>>,
+        author: NodeId,
+        signature: S,
+        tmo: TimeoutMessage<S, T>,
```
ariqchowdhury, line 127: what's the reason we want to split the input to functions that required a verified message?
Or is this a result of the changes, as verified is over a ConsensusMessage? So there's no longer a `Verified<TimeoutMessage>`?
Is there something that enforces that the (author, signature, tmo) are from the same message?

monad-consensus/src/validation/signing.rs
```diff
@@ -77,6 +88,58 @@ impl<S, M> From<Verified<S, M>> for Unverified<S, M> {
...
+        Ok(match self.obj {
+            ConsensusMessage::Proposal(m) => {
+                let verified =
+                    Unverified::new(m, self.author_signature).verify::<H>(validators, sender)?;
+                Verified {
+                    author: verified.author,
+                    message: Unverified::new(
+                        ConsensusMessage::Proposal(verified.message.obj),
+                        verified.message.author_signature,
+                    ),
```
ariqchowdhury, line 114: conceptually, this seems fine, but it reads kind of funny...
`let verified = Unverified::new(self...).verify()`
and then the Verified's message field is an Unverified whose fields are populated with
`verified.message.obj`, `verified.message.author_signature`
I don't have any suggestions, so ignore this comment, but maybe we can think about this in the background.

monad-consensus/src/pacemaker.rs
omegablitz: Yeah, it's because Verified is over a ConsensusMessage now.

I thought about this before... the way we can enforce that they're from the same message is by creating another enum (similar to what he had previously)
```rust
enum VerifiedConsensusMessage {
    Proposal(Verified<Proposal>),
    Vote(Verified<Vote>),
    Timeout(Verified<Timeout>),
}
```
and then having a function
```rust
impl Verified<ConsensusMessage> {
    fn destructure(self) -> VerifiedConsensusMessage { ... }
}
```
for doing the conversion so we can match on it after.

Positives:
- coupling is enforced at the caller level

Negatives:
- there's nothing actually enforcing the coupling in the callee
- adding the VerifiedConsensusMessage type adds extra noise

I'm open to either one, don't really feel strongly either way - let me know what you think.

monad-consensus/src/validation/signing.rs
omegablitz: Yeah lol it's kinda strange... maybe `Signed` is a better name than `Unverified`? So instead of `Unverified` and `Verified`, we have `Signed` and `Verified`?

The only good thing about the name `Unverified` is that we won't accidentally use it as the parameter to any sensitive function - whereas it's more likely we mess up and use `Signed` in places we shouldn't.

omegablitz: Also I kinda alluded to this, but I think this code will also look less strange if the signature verification is on `Unverified<ConsensusMessage>` instead of `Unverified<Proposal | Vote | Timeout>`... because then we won't need to do the weird unwrap -> verify -> wrap stuff.

monad-consensus/src/pacemaker.rs
ariqchowdhury: okay, I don't think this needs to block this PR, we can address in future.

monad-consensus/src/validation/signing.rs
ariqchowdhury: Yeah, `Signed` reads better here. We can think about it more and discuss if we want to change it later, doesn't have to block this PR.

PR #79, opened 2023-04-25
Introduce libp2p simulation tests (in `monad-driver`)

omegablitz:
There's some real weird stuff going on related to the rust-libp2p dialing behavior... but that's unrelated to the monad-state simulation test. Will do a follow-up investigation on that.

### DIFF
--- a/monad-driver/src/lib.rs
+++ b/monad-driver/src/lib.rs
@@ -1,0 +1,107 @@
+#[cfg(test)]
+mod tests {
+    use std::time::Duration;
+
+    use futures::StreamExt;
+    use monad_consensus::{
+        signatures::aggregate_signature::AggregateSignatures,
+        types::quorum_certificate::genesis_vote_info, validation::hashing::Sha256Hash,
+    };
+    use monad_crypto::secp256k1::{KeyPair, SecpSignature};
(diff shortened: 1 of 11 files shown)


monad-driver/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+
+                        delta: Duration::from_millis(2),
+                        genesis_block: genesis_block.clone(),
+                        genesis_vote_info: genesis_vote_info(genesis_block.get_id()),
+                        genesis_signatures: genesis_sigs.clone(),
+                    },
+                )
+            })
+            .collect::<Vec<_>>();
+
```
ariqchowdhury, line 83: nit: I think one of my PRs that just merged added a helper function in monad-testutil that wraps this

omegablitz: Revisited this, this function isn't actually exactly the same - because I'm using `KeyPair::libp2p_from_bytes` in the libp2p in this test, whereas the tests in `monad-state` use `create_keys` which uses`KeyPair::from_bytes` under the hood. I'm too lazy to refactor the helper function to be generic over both `KeyPair::libp2p_from_bytes` and `KeyPair::from_bytes`... can revisit later

PR #80, opened 2023-04-25
Add more consensus state tests

ariqchowdhury:


monad-crypto/src/secp256k1.rs
```diff
@@ -7,6 +7,7 @@ use zeroize::Zeroize;
 
 #[derive(Copy, Clone, Debug, PartialOrd, Ord, PartialEq, Eq, Hash)]
 pub struct PubKey(secp256k1::PublicKey);
+#[derive(Copy, Clone)]
```
omegablitz: I don't think we want Copy/Clone on KeyPair, do we?

ariqchowdhury: Ah, good point, will remove that

ariqchowdhury: ✅

PR #81, opened 2023-04-25
Exporting .proto types from `monad-proto`

michael-yxchen:
- exporting .proto types from `monad-proto` so `monad-event-proto` can reuse the protobuf types from `monad-proto`
- preparing to move the conversion functions to the crate defining the rust types. 
    - Right now they are all under `monad-proto` because the proto types are defined there. 
    - Since Unverified obj field is private, we need to move the conversion function to `monad-consensus`. Otherwise it would create circular dependencies between `monad-proto` and `monad-consensus` as they reference types defined in one another.


michael-yxchen:
Thanks, noted

PR #83, opened 2023-04-26
Defining types for Hash and HighQcRoundSigTuple

michael-yxchen:
Based on #81 
- Defining new types for them so the conversion function can be defined in the crate defining the rust types
- This is an intermediate change to keep PR smaller. It breaks the build so must be merged together with #84

### DIFF
--- a/monad-consensus/src/types/timeout.rs
+++ b/monad-consensus/src/types/timeout.rs
@@ -23,17 +23,29 @@ impl Hashable for HighQcRound {
     }
 }
 
+impl AsRef<[u8]> for HighQcRound {
+    fn as_ref(&self) -> &[u8] {
+        self.qc_round.as_ref()
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct HighQcRoundSigTuple<S> {
+    pub high_qc_round: HighQcRound,
+    pub author_signature: S,
+}
+
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct TimeoutCertificate<S> {
     pub round: Round,
-    pub high_qc_rounds: Vec<(HighQcRound, S)>,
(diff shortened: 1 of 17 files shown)


monad-types/src/lib.rs
```diff
@@ -1,11 +1,27 @@
...
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl AsRef<[u8]> for Hash {
+    fn as_ref(&self) -> &[u8] {
+        self.0.as_bytes()
+    }
+}
```
michael-yxchen, line 24: `Hash` type

monad-consensus/src/types/timeout.rs
```diff
@@ -23,17 +23,29 @@ impl Hashable for HighQcRound {
...
+    fn as_ref(&self) -> &[u8] {
+        self.qc_round.as_ref()
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct HighQcRoundSigTuple<S> {
+    pub high_qc_round: HighQcRound,
+    pub author_signature: S,
+}
```
michael-yxchen, line 36: `HighQcRoundSigTuple` type

monad-consensus/src/types/timeout.rs
```diff
@@ -23,17 +23,29 @@ impl Hashable for HighQcRound {
     }
 }
 
+impl AsRef<[u8]> for HighQcRound {
+    fn as_ref(&self) -> &[u8] {
+        self.qc_round.as_ref()
+    }
+}
```
michael-yxchen, line 30: so Hasher.update() can take in `HighQcRound`

omegablitz:
Should we just squash this into #84 if it's needed for the tests to pass? It'd be nice if every commit in master passes the tests in case we want to git bisect or something in the future.

michael-yxchen:
Sure, I'll highlight the new types squashing it into #84.

PR #84, opened 2023-04-26
`monad-proto` refactor

michael-yxchen:
- new types for `Hash` and `HighQcRoundSigTuple`
- Splitting `monad-proto` into two parts: the .proto definitions remain in `monad-proto`, the conversion functions are moved to the crate defining the corresponding rust type


monad-types/src/lib.rs
```diff
@@ -1,11 +1,30 @@
...
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl AsRef<[u8]> for Hash {
+    fn as_ref(&self) -> &[u8] {
+        self.0.as_bytes()
+    }
+}
```
michael-yxchen, line 27: new `Hash` type

monad-consensus/src/types/timeout.rs
```diff
@@ -23,17 +23,29 @@ impl Hashable for HighQcRound {
...
+        self.qc_round.as_ref()
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct HighQcRoundSigTuple<S> {
+    pub high_qc_round: HighQcRound,
+    pub author_signature: S,
+}
+
```
michael-yxchen, line 31: `HighQcRoundSigTuple` type

monad-consensus/src/types/timeout.rs
```diff
@@ -23,17 +23,29 @@ impl Hashable for HighQcRound {
     }
 }
 
+impl AsRef<[u8]> for HighQcRound {
+    fn as_ref(&self) -> &[u8] {
+        self.qc_round.as_ref()
+    }
+}
```
michael-yxchen: added so `Hasher.update` can call on HighQcRound

omegablitz: Can we do `high_qc_round.hash(&hasher)` instead? Because
```rust
impl Hashable for HighQcRound {
        state.update(self.qc_round.as_bytes());
    }
}
```
is already implemented. 

Only mentioning because I think this implementation of `AsRef<[u8]>` doesn't actually make that much sense conceptually.

monad-consensus/src/validation/signing.rs
```diff
@@ -255,20 +255,21 @@ where
     H: Hasher,
 {
     for t in tc.high_qc_rounds.iter() {
-        if t.0.qc_round >= tc.round {
+        if t.high_qc_round.qc_round >= tc.round {
             return Err(Error::InvalidTcRound);
         }
 
         // TODO fix this hashing..
```
michael-yxchen, line 262: Changed to `high_qc_round.hash(&mut hasher)`. The hashing is a little funny here as they are using different interfaces. Maybe we can move the Hashable trait, etc to `monad-crypto`, and implement `Hashable` for `Round`.  So this would look like
```
      let mut h = H::new();
      tc.round.hash(&mut h);
      t.high_qc_round.hash(&mut h);
      let msg = h.hash();
```
after refactoring

PR #89, opened 2023-05-02
Enable feature `proto` in github workflow

michael-yxchen:


omegablitz:
I think that if we use the `proto` feature in any of our crates, even if as a `dev-dependency`, `cargo check` and `cargo test` will test the feature out of the box

michael-yxchen:
yea, you're right. Confirmed by inspecting the github check outputs. Closing..

PR #90, opened 2023-05-02
Persistence and replay layer

michael-yxchen:
- A generic write ahead logger that can persist anything Serializable and Deserializable: an 8-byte (should probably use a smaller one) header is added to record length of the message payload
- Replay test: assert identical state can be reconstructed from the log file 
- `WAL.push` has three implementations, benchmark coming in a later PR

### DIFF
--- a/monad-wal/Cargo.toml
+++ b/monad-wal/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "monad-wal"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+monad-executor = { version = "0.1.0", path = "../monad-executor" }
+
+[dev-dependencies]
+monad-consensus = { path = "../monad-consensus"}
+monad-crypto = { path = "../monad-crypto" }
+monad-executor = { path = "../monad-executor" }
+monad-state = { path = "../monad-state", features = ["proto"] }
+monad-testutil = { path = "../monad-testutil" }
+monad-types = { path = "../monad-types" }
+tempfile = "3.5.0"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2754,6 +2754,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "monad-wal"
+version = "0.1.0"
+dependencies = [
+ "monad-consensus",
+ "monad-crypto",
+ "monad-executor",
+ "monad-state",
+ "monad-testutil",
+ "monad-types",
+ "tempfile",
+]
+
+[[package]]
 name = "multiaddr"
 version = "0.17.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ members = [
     "monad-types",
     "monad-validator",
     "monad-viz",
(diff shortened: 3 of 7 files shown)


michael-yxchen:
good point, we probably want a WAL trait for that. Will add one in the integration PR

monad-wal/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+[dev-dependencies]
+monad-consensus = { path = "../monad-consensus"}
+monad-crypto = { path = "../monad-crypto" }
+monad-executor = { path = "../monad-executor" }
+monad-state = { path = "../monad-state", features = ["proto"] }
+monad-testutil = { path = "../monad-testutil" }
+monad-types = { path = "../monad-types" }
+tempfile = "3.5.0"
+
+[features]
```
ariqchowdhury: @omegablitz would there be any concerns for us to depend on nightly?

monad-wal/src/file.rs
```diff
@@ -0,0 +1,45 @@
...
+        self.file.write_all(data)
+    }
+
+    #[cfg(feature = "nightly")]
+    pub fn write_all_vectored(&mut self, bufs: &mut [IoSlice<'_>]) -> io::Result<()> {
+        self.file.write_all_vectored(bufs)
+    }
+
+    pub fn flush(&mut self) -> io::Result<()> {
+        self.file.flush()
```
ariqchowdhury: does this actually do anything?
https://doc.rust-lang.org/src/std/sys/unix/fs.rs.html#1160

If the intent is to do something like fsync, I think this might be what we want https://doc.rust-lang.org/std/fs/struct.File.html#method.sync_all

but we should consider whether we want to block on fsync on every write

monad-wal/src/file.rs
```diff
@@ -0,0 +1,45 @@
...
+            .open(file_path)?;
+        Ok(Self { file })
+    }
+
+    pub fn read_exact(&mut self, buf: &mut [u8]) -> io::Result<()> {
+        self.file.read_exact(buf)
+    }
+
+    // `write_all_vectored` can write header and data in one call
+    // but it's currently a nig¡htly feature
```
ariqchowdhury: nit: typo

monad-wal/src/file.rs
michael-yxchen: yea I was trying to do a fsync... 
doing more research on this to see if the fsync are needed: 
1. understanding the tradeoff between log latency and persistency: how much data loss can we tolerate? so we can do fsync less frequently
2. journaling capabilities of file systems

monad-wal/Cargo.toml
omegablitz: Not concerns per se, but we should probably stick to stable

monad-wal/src/file.rs
michael-yxchen: Ran benchmark with `sync_all` and `sync_data` instead of `flush` respectively
The latency isn't that worse on peach3, updated pages [here](https://www.notion.so/monad-labs/Append-only-file-benchmark-a837d777cd2d4837a0531ed392343199?pvs=4) and [here](https://www.notion.so/monad-labs/Ser-append-benchmark-8faf33a1525d436299bb539c2eaf9810?pvs=4)
But it's much worse on my mac tho... it takes 2 seconds to persist 400 vote messages. Probably because APFS is not journaled but xfs on peach3 is.

PR #91, opened 2023-05-02
Benchmark: append raw data blobs through WAL

michael-yxchen:
- Benchmark three ways of pushing the message header and payload
    - write(header), write(payload), flush
    - copy header + payload to one buffer, write(header + payload), flush
    - (nightly feature) write_vectored(header + payload), flush
- Benchmark result [page](https://www.notion.so/monad-labs/Append-only-file-benchmark-a837d777cd2d4837a0531ed392343199?pvs=4)

### DIFF
--- a/monad-wal/Cargo.toml
+++ b/monad-wal/Cargo.toml
@@ -5,14 +5,27 @@ edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+# this is necessary for criterion bench options
+# https://bheisler.github.io/criterion.rs/book/faq.html#cargo-bench-gives-unrecognized-option-errors-for-valid-command-line-options
+[lib]
+bench = false
+
 [dependencies]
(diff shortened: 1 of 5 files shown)


monad-wal/Cargo.toml
```diff
@@ -9,13 +9,20 @@ edition = "2021"
...
 monad-consensus = { path = "../monad-consensus"}
 monad-crypto = { path = "../monad-crypto" }
 monad-executor = { path = "../monad-executor" }
 monad-state = { path = "../monad-state", features = ["proto"] }
 monad-testutil = { path = "../monad-testutil" }
 monad-types = { path = "../monad-types" }
-tempfile = "3.5.0"
+
+
+[[bench]]
```
ariqchowdhury, line 28: need to add this as a pre-req for using criterion benchmarks with the github benchmark-actions (even though its not fully integrated yet)
https://github.com/monad-crypto/monad-bft/blob/master/monad-state/Cargo.toml#L8-L11

PR #92, opened 2023-05-02
Benchmark: append events to WAL

michael-yxchen:
- A more realistic benchmark to append events to the WAL
- Results [page](https://www.notion.so/monad-labs/Ser-append-benchmark-8faf33a1525d436299bb539c2eaf9810?pvs=4)

### DIFF
--- a/monad-testutil/src/block.rs
+++ b/monad-testutil/src/block.rs
@@ -1,0 +1,50 @@
+use monad_consensus::signatures::aggregate_signature::AggregateSignatures;
+use monad_consensus::types::block::Block;
+use monad_consensus::types::block::TransactionList;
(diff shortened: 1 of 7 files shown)


monad-testutil/src/block.rs
```diff
@@ -0,0 +1,53 @@
...
+    let txns = txns;
+    let round = Round(block_round);
+
+    let vi = VoteInfo {
+        id: BlockId(Hash([42_u8; 32].into())),
+        round: Round(qc_round),
+        parent_id: BlockId(Hash([43_u8; 32].into())),
+        parent_round: Round(0),
+    };
+    let lci = LedgerCommitInfo {
```
ariqchowdhury: nit: LedgerCommitInfo has a constructor

PR #93, opened 2023-05-02
Proposal Sync

ariqchowdhury:
https://www.notion.so/monad-labs/proposal-sync-3f040d392ef24a2e9213845550d3e63e?pvs=4

### DIFF
--- a/monad-state/src/lib.rs
+++ b/monad-state/src/lib.rs
@@ -1,3 +1,4 @@
+use std::collections::BTreeMap;
 use std::fmt::Debug;
 use std::time::Duration;
 
@@ -28,7 +29,7 @@ use monad_crypto::{
     Signature,
 };
 use monad_executor::{Command, Message, PeerId, RouterCommand, State, TimerCommand};
-use monad_types::{NodeId, Round};
+use monad_types::{BlockId, NodeId, Round};
 use monad_validator::{
     leader_election::LeaderElection, validator::Validator, validator_set::ValidatorSet,
     weighted_round_robin::WeightedRoundRobin,
@@ -69,6 +70,14 @@ where
 
     pub fn ledger(&self) -> &Vec<Block<SCT>> {
         self.consensus_state.ledger.get_blocks()
+    }
+
+    pub fn pacemaker(&self) -> &Pacemaker<ST, SCT> {
+        &self.consensus_state.pacemaker
+    }
+
+    pub fn blocktree(&self) -> &BlockTree<SCT> {
+        &self.consensus_state.pending_block_tree
     }
 }
 
@@ -380,6 +389,11 @@ where
                         ConsensusCommand::Unschedule => {
                             cmds.push(Command::TimerCommand(TimerCommand::Unschedule))
                         }
+                        ConsensusCommand::RequestSync { blockid } => {
+                            // TODO: respond with the Proposal matching the blockID.
+                            //       right now, all 'missed' proposals are actually in-flight
+                            //       so we don't need to implement this yet for things to work
+                        }
                     }
                 }
                 cmds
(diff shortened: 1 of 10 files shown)


ariqchowdhury:
This PR depends on PR #98 and #99

monad-state/src/lib.rs
```diff
@@ -396,12 +402,16 @@ pub enum ConsensusCommand<S, T: SignatureCollection> {
...
+    RequestSync {
+        blockid: BlockId,
+    },
     // TODO add command for updating validator_set/round
     // - to handle this command, we need to call message_state.set_round()
 }
 
 struct ConsensusState<S, T, L, M> {
     pending_block_tree: BlockTree<T>,
+    pending_proposals: BTreeMap<Round, ProposalMessage<S, T>>,
```
michael-yxchen, line 453: Since we accept multiple proposals at the same round now, should we update the `pending_proposals` to buffer multiple proposals on the same level?

monad-state/src/lib.rs
```diff
@@ -492,6 +507,78 @@ where
...
+    // is possible to apply it, send the proposal to self
+    #[must_use]
+    fn process_pending_proposal(&mut self) -> Vec<ConsensusCommand<S, T>> {
+        let mut cmds = Vec::new();
+
+        if !self.pending_proposals.is_empty() {
+            // the only valid thing that could be applied is the lowest pending round
+            if let Some(entry) = self.pending_proposals.first_entry() {
+                let e = entry.get();
+                if e.block.round == self.pacemaker.next_round() {
```
michael-yxchen, line 607: consider a situation:\
we are at round(2)\
we haven't receive the round(3) block\
round(4) is timed out\
round(5) proposal is received first, with a TC for round(4), and buffered

---

when round(3) proposal finally arrives, we apply the round(3) proposal\
then we call `process_pending_proposals`\
`pacemaker.round == 3` and `block.round == 5`, so we might never apply this proposal from round(5).

so I'm thinking `e.block.round == self.pacemaker.next_round() || e.last_round_tc.map_or(false, |tc| tc.round == self.pacemaker.next_round())`.

I'm actually not sure if this case is covered by the code already, maybe I should write up this test case first.

monad-state/src/lib.rs
ariqchowdhury: yeah, good point

monad-state/src/lib.rs
ariqchowdhury: If we had a testutil that generated a valid sequence of proposals and timed-out-rounds with TC, then we could just create a sequence over N rounds and then just try applying all the permutation orders of messages. 
I can work on that and add that test.

ariqchowdhury: I'm pretty sure this test case is covered now by the `run_one_delayed_node` test where all the messages are buffered and then applied in reverse order. So the node gets round(5), TC(4), round(3)....

I think this covers the case you mentioned -- if not, we can use your TC testutil generator and make a specific test. Can do that in a follow-up PR if needed.

monad-state/src/lib.rs
```diff
@@ -742,6 +843,7 @@ mod test {
     }
 
     // 2f+1 votes for a VoteInfo leads to a QC locking -- ie, high_qc is set to that QC.
+    #[ignore]
```
ariqchowdhury, line 846: I'm fairly sure this condition isn't mandatory. But I want to think about it more before deleting it. When nodes vote, they lock (as part of processing the qc in the proposal), so if 2f+1 nodes are voting, that means that 2f+1 nodes have locked the qc. So I don't think it matters if the next leader locks before creating the next proposal as 2f+1 will have the highest locked qc.

monad-state/src/lib.rs
ariqchowdhury: The current implementation should work as is, albeit inefficiently in certain cases. Suppose we have round(C) with proposal C and an invalid proposal C'. 
if we receive in the order [C', C], then we have the correct proposal.
if we receive [C, C'], then its wrong. If proposal B arrives, it will be discarded for not extending C'. However, eventually we should get D which will get added, and then D's parent is checked and removed (C') and then C is sync requested. This argument should apply to any correct proposal arriving after C'. 

This relies on the sync request actually being implemented, so I think we should ignore this for now, and deal with testing and making it more efficient in that future PR

PR #94, opened 2023-05-03
Persistence integration

michael-yxchen:
Based on #100 
Tried breaking it up into smaller pieces but failed.... there's a lot refactoring changes because of the `State::init()` interface changes to support replay. so I'll highlight the interesting bits and examples here
- Changed the `State::init()` function. Added an extra events parameter to feed the replayed/genesis events to the State. So that the top-level knows all events passed to State and persists them.
- Modified `Nodes` (monad-executor/mock-swarm) to take in the persistence log file path, and instantiate the `PersistenceLogger`. It persists all events before calling `State.update`
- A replay test to assert that the nodes can recover their previous state and keep growing the ledger
- Changed `Nodes.states` to a BTreeMap so the stopping condition (`nodes.states().values().next().unwrap().1.ledger().len() > num_blocks`) is deterministic

### DIFF
--- a/monad-state/benches/two_node_benchmark.rs
+++ b/monad-state/benches/two_node_benchmark.rs
@@ -13,7 +13,6 @@ pub fn criterion_benchmark(c: &mut Criterion) {
 }
 
 criterion_group!(benches, criterion_benchmark);
-#[cfg(feature = "proto")]
 criterion_main!(benches);
 
 fn two_nodes() {
--- a/monad-state/src/message.rs
+++ b/monad-state/src/message.rs
@@ -122,11 +122,7 @@ where
         MessageActionPublish::new(peer, message)
     }
 
-    pub fn broadcast(&mut self, message: OM) -> Vec<MessageActionPublish<M, OM>>
-    where
-        <M as monad_executor::Message>::Id: Debug,
-        OM: Debug,
-    {
+    pub fn broadcast(&mut self, message: OM) -> Vec<MessageActionPublish<M, OM>> {
         let mut commands = Vec::new();
         for peer in self.peers.to_vec() {
             commands.push(self.send(peer, message.clone()));
@@ -140,10 +136,7 @@ where
         round: Round,
         peer: PeerId,
         id: M::Id,
-    ) -> Option<MessageActionUnpublish<M>>
-    where
-        <M as monad_executor::Message>::Id: Debug,
-    {
+    ) -> Option<MessageActionUnpublish<M>> {
         let max_round = self.max_round();
         if round >= self.min_round() && round <= max_round {
             let back_idx = self.max_rounds_cached() - 1;
--- a/monad-state/tests/msg_delays.rs
+++ b/monad-state/tests/msg_delays.rs
@@ -4,7 +4,6 @@ use monad_executor::mock_swarm::XorLatencyTransformer;
 
 mod base;
 
-#[cfg(feature = "proto")]
 #[test]
 fn two_nodes() {
     tracing_subscriber::fmt::init();
--- a/monad-state/tests/replay.rs
+++ b/monad-state/tests/replay.rs
@@ -26,7 +26,6 @@ mod test {
         recover_nodes_msg_delays(4, 10, 5);
     }
 
-    #[cfg(feature = "proto")]
     pub fn recover_nodes_msg_delays(
         num_nodes: u16,
         num_blocks_before: usize,
--- a/monad-viz/Cargo.toml
+++ b/monad-viz/Cargo.toml
@@ -17,6 +17,6 @@ macroquad = { version = "0.3.25", default-features = false }
 monad-executor = { path = "../monad-executor" }
 monad-consensus = { path = "../monad-consensus" }
 monad-crypto = { path = "../monad-crypto" }
-monad-state = { path = "../monad-state", features = ["proto"] }
+monad-state = { path = "../monad-state" }
 monad-testutil = { path = "../monad-testutil" }
 monad-wal = { path = "../monad-wal" }


monad-executor/src/state.rs
```diff
@@ -73,7 +73,10 @@ pub trait State: Sized {
     type OutboundMessage: Into<Self::Message> + AsRef<Self::Message>;
     type Message: Message<Event = Self::Event>;
 
-    fn init(config: Self::Config) -> (Self, Vec<Command<Self::Message, Self::OutboundMessage>>);
+    fn init(
+        config: Self::Config,
+        events: Vec<Self::Event>,
+    ) -> (Self, Vec<Command<Self::Message, Self::OutboundMessage>>);
```
michael-yxchen: Changed the `State::init()` function. Added an extra events parameter to feed the replayed/genesis events to the State. So that the top-level knows all events passed to State and persists them.

monad-state/src/lib.rs
```diff
@@ -247,9 +296,10 @@ where
             ),
         };
 
-        let init_cmds = monad_state.update(MonadEvent::ConsensusEvent(ConsensusEvent::Timeout(
-            PacemakerTimerExpire,
-        )));
+        let mut init_cmds = Vec::new();
+        for e in events {
+            init_cmds.extend(monad_state.update(e));
+        }
```
michael-yxchen: Modified `MonadState` to take in the genesis event (the round 1 timeout) instead of hard-coding it in `MonadState::init`

monad-executor/src/mock_swarm.rs
```diff
@@ -1,32 +1,52 @@
...
+                genesis_events
+            } else {
+                replay_events
+            };
+            let (state, init_commands) = S::init(config, init_events);
+
             executor.exec(init_commands);
-            states.insert(PeerId(pubkey), (executor, state));
+
+            states.insert(PeerId(pubkey), (executor, state, wal));
```
michael-yxchen: Modified `Nodes` (monad-executor/mock-swarm) to take in the persistence log file path and genesis event, and instantiate the `PersistenceLogger`

monad-executor/src/mock_swarm.rs
```diff
@@ -1,32 +1,52 @@
...
 use crate::{executor::mock::MockExecutor, Executor, PeerId, State};
 
-pub struct Nodes<S: State, L: Fn(&PeerId, &PeerId) -> Duration> {
-    states: HashMap<PeerId, (MockExecutor<S>, S)>,
+pub struct Nodes<
+    S: State,
+    L: Fn(&PeerId, &PeerId) -> Duration,
+    LGR: PersistenceLogger<Event = S::Event>,
+> {
+    states: BTreeMap<PeerId, (MockExecutor<S>, S, LGR)>,
```
michael-yxchen, line 101: Changed `Nodes.states` to a BTreeMap so the stopping condition (`nodes.states().values().next().unwrap().1.ledger().len() > num_blocks`) is deterministic

monad-state/src/lib.rs
michael-yxchen: The genesis events is passed to the top-level (e.g. `Nodes`) and forwarded to `MonadState` if the persistence log file is empty

monad-state/tests/replay.rs
michael-yxchen, line 1: A replay test to assert that the nodes can recover their previous state and keep growing the ledger

monad-executor/src/mock_swarm.rs
```diff
@@ -1,32 +1,41 @@
...
             let mut executor: MockExecutor<S> = MockExecutor::default();
-            let (state, init_commands) = S::init(config);
+            let (mut wal, replay_events) = LGR::new(file_path).unwrap();
+            let (state, init_commands) = S::init(config, replay_events);
+
             executor.exec(init_commands);
-            states.insert(PeerId(pubkey), (executor, state));
+
+            states.insert(PeerId(pubkey), (executor, state, wal));
         }
```
michael-yxchen: Modified Nodes (monad-executor/mock-swarm) to take in the persistence log file path, and instantiate the PersistenceLogger

monad-state/tests/replay.rs
michael-yxchen, line 1: A replay test to assert that the nodes can recover their previous state and keep growing the ledger

monad-executor/src/state.rs
omegablitz: I don't fully understand why `State::init()` needs an events parameter, vs calling `update()` directly with the replayed events.

Example: https://github.com/monad-crypto/monad-bft/blob/2ac7a273a499eda66f50541e718ff5aeacb330a6/monad-executor/src/lib.rs#L22-L25

michael-yxchen: yeah, I think the upper level should just call update instead of passing the events in now. I had a different interface for the upper level (e.g. Nodes) which composes with this interface slightly better. But I decided to move on from that, so should revert this interface as well.

michael-yxchen:
> does the file_path for the WAL need to be a separate parameter or can it go in MonadConfig. That seems like the more natural place for that to live to me.

I did this because MonadConfig is an associated type with MonadState and I think it would be better to separate WAL from  MonadState. It should definitely go in the top level config (e.g. [monad-node config](https://github.com/monad-crypto/monad-bft/blob/master/monad-node/src/main.rs#LL31C1-L42C2))

monad-executor/src/executor/mock.rs
```diff
@@ -618,6 +619,31 @@ mod tests {
...
+        }
+    }
+
+    impl Deserializable for SimpleChainEvent {
+        type ReadError = ReadError;
+
+        fn deserialize(message: &[u8]) -> Result<Self, Self::ReadError> {
+            unimplemented!()
+        }
+    }
```
omegablitz: If we remove the Serializable/Deserializable trait bounds from the PersistenceLogger trait definition, we won't need these dummy implementations here.

monad-executor/src/mock_swarm.rs
```diff
@@ -1,37 +1,50 @@
...
 where
     S: State,
     L: Fn(&PeerId, &PeerId) -> Duration,
+    LGR: PersistenceLogger<Event = S::Event>,
 
     MockExecutor<S>: Unpin,
 {
     pub fn new(
-        peers: Vec<(PubKey, S::Config)>,
+        peers: Vec<(PubKey, S::Config, PathBuf)>,
```
omegablitz: Instead of being a `PathBuf`, I think it makes more sense to make the configuration of `PersistenceLogger` be from a `LGR::Config` associated type. PathBuf doesn't really make sense for a mock logger.

monad-state/src/message.rs
```diff
@@ -112,7 +113,11 @@ where
         commands
     }
 
-    pub fn send(&mut self, peer: PeerId, message: OM) -> MessageActionPublish<M, OM> {
+    pub fn send(&mut self, peer: PeerId, message: OM) -> MessageActionPublish<M, OM>
+    where
+        <M as monad_executor::Message>::Id: Debug,
+        OM: Debug,
```
omegablitz: Ideally these Debug bounds shouldn't be necessary here - maybe they can be removed from the PersistenceLogger trait definition?

michael-yxchen:
Created an associated type PersistenceLogger::Config and moved the log_path parameter there for WALogger.

It makes swapping between the `MockWALogger` and `WALogger` a little more difficult because the two configs are different. But I think that's probably fine since we won't be switching between the two loggers for now

monad-state/benches/two_node_benchmark.rs
```diff
@@ -12,6 +13,7 @@ pub fn criterion_benchmark(c: &mut Criterion) {
 }
 
 criterion_group!(benches, criterion_benchmark);
+#[cfg(feature = "proto")]
```
omegablitz: nit: should this be necessary if we're only using the Mock WAL in `base::run_nodes`?

monad-state/src/message.rs
```diff
@@ -121,7 +122,11 @@ where
         MessageActionPublish::new(peer, message)
     }
 
-    pub fn broadcast(&mut self, message: OM) -> Vec<MessageActionPublish<M, OM>> {
+    pub fn broadcast(&mut self, message: OM) -> Vec<MessageActionPublish<M, OM>>
+    where
```
omegablitz: do we need the Debug trait bounds?

monad-state/src/message.rs
```diff
@@ -135,11 +140,15 @@ where
         round: Round,
         peer: PeerId,
         id: M::Id,
-    ) -> Option<MessageActionUnpublish<M>> {
+    ) -> Option<MessageActionUnpublish<M>>
+    where
+        <M as monad_executor::Message>::Id: Debug,
```
omegablitz: do we need the Debug trait bounds?

monad-state/tests/msg_delays.rs
```diff
@@ -4,6 +4,7 @@ use monad_executor::mock_swarm::XorLatencyTransformer;
 
 mod base;
 
+#[cfg(feature = "proto")]
```
omegablitz: nit: should this be necessary if we're using the Mock WAL?

monad-state/tests/replay.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    type SignatureType = SecpSignature;
+    type SignatureCollectionType = AggregateSignatures<SignatureType>;
+
+    #[test]
+    fn test_replay() {
+        recover_nodes_msg_delays(4, 10, 5);
+    }
+
+    #[cfg(feature = "proto")]
```
omegablitz: nit: this `cfg` feature guard may be redundant

monad-viz/Cargo.toml
```diff
@@ -17,5 +17,6 @@ macroquad = { version = "0.3.25", default-features = false }
 monad-executor = { path = "../monad-executor" }
 monad-consensus = { path = "../monad-consensus" }
 monad-crypto = { path = "../monad-crypto" }
-monad-state = { path = "../monad-state" }
+monad-state = { path = "../monad-state", features = ["proto"] }
```
omegablitz: nit: monad-viz probably shouldn't need the proto feature if we're using MockWAL. Maybe monad-proto should be an optional dep in monad-wal.... but we can always change this later if compile times are too annoying.

PR #106, opened 2023-05-10
basic counter using commands

michael-yxchen:
A basic counter service implemented using Commands. It prints the output when dropped, example output looks like:
```
   Running tests/msg_delays.rs (target/debug/deps/msg_delays-ec0573849b65bf2a)

running 1 test
Counter service: [("blocktree.add.success", 43), ("blocktree.prune.noop", 40), ("blocktree.prune.success", 41)]
Counter service: [("blocktree.add.success", 43), ("blocktree.prune.noop", 40), ("blocktree.prune.success", 41)]
Counter service: [("blocktree.add.success", 43), ("blocktree.prune.noop", 40), ("blocktree.prune.success", 41)]
Counter service: [("blocktree.add.success", 43), ("blocktree.prune.noop", 41), ("blocktree.prune.success", 41)]
test two_nodes ... ok
```


monad-executor/src/executor/parent.rs
```diff
@@ -4,35 +4,40 @@ use std::{
...
 
-impl<E, R, T> Stream for ParentExecutor<R, T>
+impl<E, R, T, C> Stream for ParentExecutor<R, T, C>
 where
     R: Stream<Item = E> + Unpin,
     T: Stream<Item = E> + Unpin,
+    C: Unpin,
 {
     type Item = E;
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
```
omegablitz, line 43: I think `this.counter` should be added to the `poll_next` - we don't need to fix this now though, especially if we're moving forward with the tracing approach.

michael-yxchen: I left this out because I thought the Counter Executor will not emit values. Should the `CounterService::poll_next` return `Poll:Ready(None)`?

omegablitz: Yeah, you can put it in and make it return `Poll::Ready(None)`

michael-yxchen:
Moving to `tracing-opentelemetry`

PR #108, opened 2023-05-15
testutil/ProposalGen: `next_tc`

michael-yxchen:
Adding `next_tc`: the function generates a `tc` for the next round, updates relevant state (last_tc), and returns the complete list of timeout messages required to recreate the tc.

The reason that `next_tc` function returns the timeout messages instead of the `TimeoutCertificate` is that the `ConsensusState` can only consume timeout messages. TimeoutCertificates are processed when they are included in a proposal block.

In tests, the return value can be ignored if we are not testing how `ConsensusState` is processing timeout messages. The next `next_proposal` call will contain the tc so pacemaker round will be updated.

### DIFF
--- a/monad-testutil/Cargo.toml
+++ b/monad-testutil/Cargo.toml
@@ -11,7 +11,7 @@ edition = "2021"
 bench = false
 
 [dependencies]
-monad-consensus = {path = "../monad-consensus"}
+monad-consensus = { path = "../monad-consensus" }
 monad-crypto = {path = "../monad-crypto"}
 monad-state = {path = "../monad-state"}
 monad-types = {path = "../monad-types"}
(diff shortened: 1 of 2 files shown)


michael-yxchen:
yeah that's a good point

The Safety state is indeed inconsistent with ProposalGen after we call `next_tc`. The consensus state should timeout itself. We can manually call `state.pacemaker.handle_event(&mut state.safety, &state.high_qc, PacemakerTimerExpire);` and pass in the keys that have a consistent state with propgen.

Alternatively, we can pass in the states as parameters and have `next_tc` call the handle event. I feel this way is cleaner, but provides less transparency/control into what's going on: should `next_tc` fail if the state passed in cannot make a timeout, or should it make one so long as there's enough stake in the remaining keys? So I'd prefer the first way

The inconsistency regarding `last_round_tc` is easier to solve. Passing the timeout messages to state would create a TC. 

Slightly unrelated: I should probably write a custom `PartialEq` implementation for TC that's independent of the ordering of `high_qc_rounds`

PR #109, opened 2023-05-15
consensus_state commit rule test

michael-yxchen:
Adding two tests to test the commit rule: specifically how `ledger_commit_info.commit_state_hash` is populated.

### DIFF
--- a/monad-state/src/lib.rs
+++ b/monad-state/src/lib.rs
@@ -1140,4 +1140,137 @@ mod test {
             perms
         );
     }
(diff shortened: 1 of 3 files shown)


michael-yxchen:
Changed round calculation on `next_tc` and some assertions on the commit rule test. 

So now instead of auto-incrementing the round number after generating the proposal, it is incremented
- when last round qc is included in a proposal on `next_proposal`
- when tc is generated

PR #111, opened 2023-05-16
`Transformer` abstraction for mock_swarm configuration

omegablitz:
Adds the concepts of a `Transformer` - something that can alter the delivery of messages.
Transformers can be composed together.
With the abstraction provided, it'll be easy for us to add stuff to control bandwidth, rearrange specific messages arbitrarily, etc.

Probably easiest to understand reading from the code... will mark out relevant sections.

### DIFF
--- a/monad-executor/src/executor/mock.rs
+++ b/monad-executor/src/executor/mock.rs
@@ -359,7 +359,7 @@ mod tests {
 
     use crate::{
         executor::mock::MockExecutor,
-        mock_swarm::Nodes,
+        mock_swarm::{LatencyTransformer, Nodes},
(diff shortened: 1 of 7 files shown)


omegablitz:
This is the interesting part:
https://github.com/monad-crypto/monad-bft/blob/0476dd65f83816adb59c5e105d3d4a6de2ea82d3/monad-executor/src/mock_swarm.rs#L26-L100

So for example, you could use a `LatencyTransformer` directly, or you could use a `vec![LatencyTransformer(...).boxed(), XorTransformer(...).boxed()]`, etc.

PR #114, opened 2023-05-18
Lift mempool out of monad-state

omegablitz:
This PR lifts the mempool out of monad-state, and fixes a couple small bugs I found.
It might be kinda confusing... happy to explain verbally if that's easier.

Will squash the bugfix down before merging.


monad-proto/proto/event.proto
```diff
@@ -26,10 +28,21 @@ message ProtoMessageWithSender {
...
 }
 
+message ProtoFetchedTxs {
+  monad_proto.basic.ProtoNodeId node_id = 1;
+  monad_proto.basic.ProtoRound round = 2;
+  monad_proto.quorum_certificate.ProtoQuorumCertificateAggSig high_qc = 3;
+
+  optional monad_proto.timeout.ProtoTimeoutCertificate last_round_tc = 4;
+
+  bytes txns = 5;
```
michael-yxchen: nit: there is `monad_proto.block.ProtoTransactionList` that we can reuse

PR #116, opened 2023-05-19
Handle out of order messages

ariqchowdhury:
- blocktree adds blocks, even if they don’t have a parent
- commit rule becomes consecutive blocks and path to root exists
- when block without parent is added, send request sync
- this allows a node to continue proposing and moving consensus forward, even if it skips

### DIFF
--- a/monad-blocktree/src/blocktree.rs
+++ b/monad-blocktree/src/blocktree.rs
@@ -1,4 +1,5 @@
 use monad_consensus::types::block::Block;
+use monad_consensus::types::quorum_certificate::QuorumCertificate;
 use monad_consensus::types::signature::SignatureCollection;
 use monad_types::BlockId;
 use monad_types::Round;
@@ -172,7 +173,7 @@ impl<T: SignatureCollection> BlockTree<T> {
         let children = self
             .tree
             .iter()
-            .filter(|(_k, ref v)| v.parent == Some(new_bid))
+            .filter(|(_k, v)| v.parent == Some(new_bid))
             .map(|k| k.0)
             .collect::<Vec<_>>();
 
@@ -185,7 +186,7 @@ impl<T: SignatureCollection> BlockTree<T> {
     pub fn path_to_root(&self, b: &BlockId) -> bool {
         let mut bid = b;
         loop {
-            let Some(i) = self.tree.get(&bid) else { return false };
+            let Some(i) = self.tree.get(bid) else { return false };
             if i.block.get_id() == self.root {
                 return true;
             }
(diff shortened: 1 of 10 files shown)


monad-executor/src/mock_swarm.rs
```diff
@@ -28,6 +30,8 @@ pub trait Transformer<M: Message> {
     // TODO smallvec? resulting Vec will almost always be len 1
     fn transform(&mut self, message: LinkMessage<M>) -> Vec<(Duration, LinkMessage<M>)>;
 
+    fn accumulated(&self) -> Option<&Vec<LinkMessage<M>>>;
```
omegablitz: I think this `accumulated` function shouldn't be necessary

monad-state/src/lib.rs
```diff
@@ -380,6 +380,11 @@ where
                         ConsensusCommand::Unschedule => {
                             cmds.push(Command::TimerCommand(TimerCommand::Unschedule))
                         }
+                        ConsensusCommand::RequestSync { blockid: _ } => {
+                            // TODO: respond with the Proposal matching the blockID.
```
omegablitz, line 422: super nit: maybe worth adding a trace log warn here

monad-state/tests/base.rs
```diff
@@ -98,3 +100,64 @@ pub fn run_nodes<T: Transformer<MM>>(
...
+        .expect("no filtered messages")
+        .clone();
+
+    let iter_direction = if reverse {
+        itertools::Either::Right(filtered_msgs.into_iter().rev())
+    } else {
+        itertools::Either::Left(filtered_msgs.into_iter())
+    };
+
+    for msg in iter_direction {
```
omegablitz: I think all of this could get folded into a transformer

monad-executor/src/state.rs
```diff
@@ -79,7 +79,7 @@ pub trait State: Sized {
 
 pub trait Message: Clone {
     type Event;
-    type Id: Eq + Hash + Clone;
+    type Id: Eq + Hash + Clone + std::fmt::Debug;
```
omegablitz: nit: is this necessary? it's fine if so, but just checking that it's not just a leftover from debugging

monad-state/src/lib.rs
```diff
@@ -609,16 +620,20 @@ where
...
+        if qc.info.ledger_commit.commit_state_hash.is_some()
+            && self
+                .pending_block_tree
+                .path_to_root(&qc.info.vote.parent_id)
+        {
             let blocks_to_commit = self
                 .pending_block_tree
                 .prune(&qc.info.vote.parent_id)
-                .unwrap();
+                .unwrap_or_else(|_| panic!("\n{:?}", self.pending_block_tree));
```
omegablitz, line 683: super nit: leftover custom unwrap, or is this still useful for debug?

if we see it happening in practice, may be nicer to make it a log fatal

monad-state/tests/base.rs
omegablitz: Btw, once this logic is encapsulated in a transformer, this whole function may be redundant - we might be able to have a single `run_node` function which we plug different combinations of transformers into.

We can save that for the future though

monad-blocktree/src/blocktree.rs
```diff
@@ -159,31 +160,62 @@ impl<T: SignatureCollection> BlockTree<T> {
...
+            let Some(parent_id) = &i.parent else {return false };
+            bid = parent_id;
+        }
+    }
+
+    pub fn has_parent(&self, qc: &QuorumCertificate<T>) -> bool {
+        self.tree.contains_key(&qc.info.vote.id)
+    }
+
     pub fn debug_print(&self) -> std::io::Result<()> {
```
michael-yxchen, line 202: nit: I think we would want to print all the blocks that don't trace back to the root. It sounds annoying to maintain all the "roots", so maybe we can delay until we need it.

PR #117, opened 2023-05-19
Tracing counter

michael-yxchen:
A counter implementation that composes with `tracing` and adopts the same interface as `tracing-opentelemetry::MetricsLayer`
- The counter is a `tracing-subscriber` layer that keeps track of the event counts
- The interface is the same as `tracing-opentelemetry::MetricsLayer` so it could be swapped with the `tracing-opentelemetry` backend. 
    - Only the `monotonic_counter` interface is implemented
- Instead of periodically pushing to a data upstream, the `counter_status!` macro print the current tally
 - Added a per-layer filter for the format layer to avoid printing all the counter logs
 - Level filter needs to be specified as a per-layer filter as well, can add support for passing that in through the `RUST_LOG` env var
 - `cargo test two_nodes --test single_node_metrics -- --nocapture`
 - Specifying span with name "node" and key "id" adds a prefix to the counter metrics collected within the span. (e.g. `info_span!("node", id = "1234")` prefixes "1234" to all metrics collected in the span)
 - Performance overhead: running two nodes for 10240 blocks
    - 0.47s -> 0.54s ~ 15% when blocktree is instrumented (emitting 2 metrics per node per block produced) if all the `println!` are suppressed

### DIFF
--- a/monad-blocktree/src/blocktree.rs
+++ b/monad-blocktree/src/blocktree.rs
@@ -1,12 +1,14 @@
 use monad_consensus::types::block::Block;
 use monad_consensus::types::quorum_certificate::QuorumCertificate;
 use monad_consensus::types::signature::SignatureCollection;
+use monad_tracing_counter::inc_count;
 use monad_types::BlockId;
 use monad_types::Round;
 use ptree::print_tree;
 use std::collections::HashMap;
 use std::fmt;
 use std::result::Result as StdResult;
+use tracing::trace;
 
 use ptree::builder::TreeBuilder;
 
@@ -125,6 +127,7 @@ impl<T: SignatureCollection> BlockTree<T> {
         // retain the new root block, remove all other committable block to avoid copying
         let mut commit: Vec<Block<T>> = Vec::new();
         if &self.root == new_root {
+            inc_count!(blocktree.prune.noop);
             return Ok(commit);
         }
         let new_root_block = &self
@@ -150,11 +153,13 @@ impl<T: SignatureCollection> BlockTree<T> {
             .retain(|_, b| b.block.round > new_root_round || b.block.get_id() == self.root);
 
         commit.reverse();
+        inc_count!(blocktree.prune.success);
         Ok(commit)
     }
 
     pub fn add(&mut self, b: Block<T>) -> Result<()> {
         if self.tree.contains_key(&b.get_id()) {
+            inc_count!(blocktree.add.duplicate);
             return Ok(());
         }
 
@@ -180,6 +185,7 @@ impl<T: SignatureCollection> BlockTree<T> {
         new_btb.children.extend(children);
         self.tree.insert(new_bid, new_btb);
         self.high_round = new_round;
+        inc_count!(blocktree.add.success);
         Ok(())
     }
 
--- a/monad-tracing-counter/src/counter.rs
+++ b/monad-tracing-counter/src/counter.rs
@@ -1,0 +1,223 @@
+use std::collections::HashMap;
+use std::sync::atomic::AtomicUsize;
+use std::sync::atomic::Ordering;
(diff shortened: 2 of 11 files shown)


monad-tracing-counter/src/counter.rs
```diff
@@ -0,0 +1,230 @@
...
+            {
+                return false;
+            }
+        }
+        true
+    }
+}
+
+lazy_static! {
+    pub static ref METRIC_FILTER: MetricFilter = MetricFilter {};
```
ariqchowdhury: why does there need to be a static ref for this? The place its used, with_filter(), just clones it -- why can't it just instantiate a MetricFilter

monad-tracing-counter/src/counter.rs
```diff
@@ -0,0 +1,230 @@
...
+}
+
+impl<S> tracing_subscriber::Layer<S> for CounterLayer
+where
+    S: CoreSubscriber,
+    Self: 'static,
+{
+    fn on_register_dispatch(&self, collector: &Dispatch) {
+        let _ = collector;
+    }
```
ariqchowdhury: what's the intent here?

monad-tracing-counter/src/counter.rs
```diff
@@ -0,0 +1,230 @@
...
+        interest
+    }
+
+    fn on_new_span(
+        &self,
+        attrs: &span::Attributes<'_>,
+        id: &span::Id,
+        ctx: tracing_subscriber::layer::Context<'_, S>,
+    ) {
+        let _ = (attrs, id, ctx);
```
ariqchowdhury: what's the intent here?

monad-blocktree/src/blocktree.rs
```diff
@@ -125,6 +126,7 @@ impl<T: SignatureCollection> BlockTree<T> {
         // retain the new root block, remove all other committable block to avoid copying
         let mut commit: Vec<Block<T>> = Vec::new();
         if &self.root == new_root {
+            trace!(monotonic_counter.blocktree.prune.noop = 1);
```
ariqchowdhury: because the name here "monotonic_counter." needs to match the static string in monad-tracing-counter, I wonder if its better to create a macro in that crate so it's easier to use. Provide something like `inc_count!(blocktree.prune.noop)`

monad-tracing-counter/src/counter.rs
michael-yxchen: sry these are the default implementations that suppresses the arguments unused warning. Will remove the method altogether.

monad-tracing-counter/src/counter.rs
michael-yxchen: yeah I should just instantiate a MetricFilter

michael-yxchen:
I was using the locks because the `set_global_default` function requires the subscriber to be `Send`, and `CounterLayer` need interior mutability because the `Layer` trait passes in an immutable reference of the layer. 

Used the `RwLock` over `Mutex` just in-case we are multithreading mock-swarm in the future.

monad-tracing-counter/src/counter.rs
ariqchowdhury: you can prefix an underscore in the argument name to suppress the warning

monad-blocktree/src/blocktree.rs
michael-yxchen: Let me see what I can do to concatenate two static strings in a macro

If we are adding the macro, we need to add the `monad-tracing-counter` dependency to every crate that's outputting metrics. Is that ok?

ariqchowdhury: I think that's okay...`monad-tracing-counter` should be a leaf dependency so won't cause any weird circular depends.

PR #120, opened 2023-05-31
optimizations for mock_swarm sims

ariqchowdhury:
Improving times for 100 nodes, 1024 blocks
Time with release build went from ~7.5s to ~2.7s

### DIFF
--- a/monad-executor/src/mock_swarm.rs
+++ b/monad-executor/src/mock_swarm.rs
@@ -239,7 +239,7 @@ where
     fn simulate_peer(&mut self, peer_id: &PeerId, tick: Duration) {
         let outbounds = {
             let mut outbounds: Vec<(Duration, LinkMessage<<S as State>::Message>)> = Vec::new();
-            let (mut executor, state, wal) = self.states.remove(peer_id).unwrap();
+            let (executor, _, _) = self.states.get_mut(peer_id).unwrap();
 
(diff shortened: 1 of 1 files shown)


monad-crypto/src/secp256k1.rs
```diff
@@ -32,6 +32,33 @@ impl std::fmt::Debug for PubKey {
...
+
+impl std::cmp::PartialOrd for PubKey {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl std::hash::Hash for PubKey {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        let slice = unsafe { std::mem::transmute::<Self, [u8; 64]>(*self) };
```
omegablitz, line 57: secp256kq::PublicKey implements `Hash` which calls `Hasher::write` [under the hood](https://docs.rs/secp256k1/0.12.2/src/secp256k1/ffi.rs.html#79), so we could also accomplish the same thing without unsafe via an ephemeral trait implementation.

It would be a little jank though because we'd [need to add](https://doc.rust-lang.org/std/hash/trait.Hasher.html#tymethod.finish) an `unreachable!()` in the `Hasher::finish` implementation.

Up to you if you want to change it, I see a case for both.

PR #123, opened 2023-06-02
Use `waker` in executors

omegablitz:
Previously we wouldn't actually call the `waker` in our futures - this PR fixes that.


omegablitz:
Will add more comprehensive test before review actually

PR #133, opened 2023-06-09
Replace Ledger trait with LedgerExecutor

ariqchowdhury:
Ledger commits can have side-effects so it should exist outside of the consensus state and be driven via ConsensusCommands.

### DIFF
--- a/monad-executor/src/state.rs
+++ b/monad-executor/src/state.rs
@@ -46,12 +46,16 @@ pub enum MempoolCommand<E> {
     FetchReset,
 }
 
+pub enum LedgerCommand<B> {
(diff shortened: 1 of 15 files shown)


monad-executor/src/state.rs
```diff
@@ -98,6 +106,7 @@ pub trait State: Sized {
 
 pub trait Message: Clone {
     type Event;
+    type Output;
```
omegablitz: `Message` corresponds to the p2p message type, so I don't think a `type Output` makes sense in that context.

zander-xyz: 🤔

PR #140, opened 2023-06-21
Move `(De)serializable` traits to new `monad-traits` crate

michael-yxchen:
The two traits were originally defined in `monad-types`, which depends on `monad-crypto` crate. The `monad-crypto` crate can use the traits, so lifting them to a separate crate to avoid circular dependencies.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3472,7 +3472,7 @@ dependencies = [
  "monad-consensus",
  "monad-crypto",
  "monad-executor",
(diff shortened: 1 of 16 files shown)


ariqchowdhury:
I'm not a huge fan of having a monad-traits crate. Its not really clear what should go in there. Feels like it could become a dumping ground for all sort of unrelated stuff. Maybe we can discuss offline what to do here.

PR #142, opened 2023-06-26
Visualize and replay the consensus logs

vicdunmin:
1. Implemented a _TimedEvent_ class for a timestamped write ahead logger.
2. Interactively visualize the logs
3. Integrated the new visualization with the old one using command line. (cargo run / cargo run -- logs)

### DIFF
--- a/monad-executor/src/executor/mock.rs
+++ b/monad-executor/src/executor/mock.rs
@@ -13,7 +13,6 @@ use super::mempool::MockMempool;
 use crate::{
     state::PeerId, Command, Executor, Message, RouterCommand, RouterTarget, State, TimerCommand,
 };
-
 use futures::{Stream, StreamExt};
 
 pub struct MockExecutor<S>
@@ -468,6 +467,21 @@ mod tests {
         Vote { peer: PeerId, round: u64 },
     }
 
+    use monad_types::{Deserializable, Serializable};
+
+    impl Serializable for SimpleChainEvent {
+        fn serialize(&self) -> Vec<u8> {
+            unreachable!("not used")
+        }
+    }
+
+    impl Deserializable for SimpleChainEvent {
+        type ReadError = ReadError;
+        fn deserialize(_buf: &[u8]) -> Result<Self, Self::ReadError> {
+            unreachable!("not used")
+        }
+    }
+
     #[derive(Debug)]
     struct ReadError {}
 
@@ -574,6 +588,7 @@ mod tests {
 
     #[test]
     fn test_nodes() {
+        use crate::timed_event::TimedEvent;
         let pubkeys = create_keys(NUM_NODES as u32)
             .iter()
             .map(KeyPair::pubkey)
@@ -589,10 +604,11 @@ mod tests {
             .zip(std::iter::repeat(MockWALoggerConfig {}))
             .map(|((a, b), c)| (a, b, c))
             .collect();
-        let mut nodes = Nodes::<SimpleChainState, _, MockWALogger<SimpleChainEvent>>::new(
-            peers,
-            LatencyTransformer(Duration::from_millis(50)),
-        );
+        let mut nodes =
+            Nodes::<SimpleChainState, _, MockWALogger<TimedEvent<SimpleChainEvent>>>::new(
+                peers,
+                LatencyTransformer(Duration::from_millis(50)),
+            );
 
         while let Some((duration, id, event)) = nodes.step() {
             println!("{duration:?} => {id:?} => {event:?}")
--- a/monad-executor/src/lib.rs
+++ b/monad-executor/src/lib.rs
@@ -4,8 +4,9 @@ mod state;
 pub use state::*;
 
 pub mod executor;
-
 pub mod mock_swarm;
+pub mod replay_nodes;
+pub mod timed_event;
 
 #[cfg(feature = "proto")]
 pub mod convert;
--- a/monad-executor/src/mock_swarm.rs
+++ b/monad-executor/src/mock_swarm.rs
@@ -7,13 +7,13 @@ use rand::Rng;
 use rand_chacha::rand_core::SeedableRng;
 use rand_chacha::ChaChaRng;
 
+use crate::timed_event::TimedEvent;
+use crate::RouterTarget;
+use crate::{executor::mock::MockExecutor, Executor, Message, PeerId, State};
 use futures::StreamExt;
 use monad_crypto::secp256k1::PubKey;
 use monad_wal::PersistenceLogger;
 use tracing::info_span;
-
-use crate::RouterTarget;
-use crate::{executor::mock::MockExecutor, Executor, Message, PeerId, State};
 
 #[derive(Clone)]
 pub struct LinkMessage<M: Message> {
@@ -181,7 +181,7 @@ pub struct Nodes<S, T, LGR>
 where
     S: State,
     T: Transformer<S::Message>,
-    LGR: PersistenceLogger<Event = S::Event>,
+    LGR: PersistenceLogger<Event = TimedEvent<S::Event>>,
 {
     states: BTreeMap<PeerId, (MockExecutor<S>, S, LGR)>,
     transformer: T,
@@ -192,7 +192,7 @@ impl<S, T, LGR> Nodes<S, T, LGR>
 where
     S: State,
     T: Transformer<S::Message>,
-    LGR: PersistenceLogger<Event = S::Event>,
+    LGR: PersistenceLogger<Event = TimedEvent<S::Event>>,
 
     MockExecutor<S>: Unpin,
 {
@@ -207,7 +207,7 @@ where
             let (mut state, mut init_commands) = S::init(state_config);
 
             for event in replay_events {
-                init_commands.extend(state.update(event));
+                init_commands.extend(state.update(event.event));
             }
 
             executor.exec(init_commands);
@@ -250,7 +250,11 @@ where
         {
             let id = *id;
             let event = futures::executor::block_on(executor.next()).unwrap();
-            wal.push(&event).unwrap(); // FIXME: propagate the error
+            let timed_event = TimedEvent {
+                timestamp: tick,
+                event: event.clone(),
+            };
+            wal.push(&timed_event).unwrap(); // FIXME: propagate the error
             let node_span = info_span!("node", id = ?id);
             let _guard = node_span.enter();
             let commands = state.update(event.clone());
--- a/monad-executor/src/replay_nodes.rs
+++ b/monad-executor/src/replay_nodes.rs
@@ -1,0 +1,151 @@
+use std::collections::{BTreeMap, HashSet};
+use std::time::Duration;
+
+use crate::{Command, LedgerCommand, Message, PeerId, RouterCommand, RouterTarget, State};
+use monad_crypto::secp256k1::PubKey;
+
+pub struct PendingMsg<S>
+where
+    S: State,
+{
+    pub send_id: PeerId,
+    pub send_tick: Duration,
+    pub event: S::Event,
+    pub message: <S as State>::Message,
+}
+
+pub struct NodesInfo<S>
+where
+    S: State,
+{
+    state: S,
+    blockchain: Vec<S::Block>,
+    pending_messages: Vec<PendingMsg<S>>,
+}
+
+impl<S> NodesInfo<S>
+where
+    S: State,
+{
+    pub fn state(&self) -> &S {
+        &self.state
+    }
+    pub fn mut_state(&mut self) -> &mut S {
+        &mut self.state
+    }
+
+    pub fn blockchain(&self) -> &Vec<S::Block> {
+        &self.blockchain
+    }
+
+    pub fn mut_blockchain(&mut self) -> &mut Vec<S::Block> {
+        &mut self.blockchain
+    }
+
+    pub fn pending_messages(&self) -> &Vec<PendingMsg<S>> {
+        &self.pending_messages
+    }
+
+    pub fn mut_pending_messages(&mut self) -> &mut Vec<PendingMsg<S>> {
+        &mut self.pending_messages
+    }
+}
+
+pub struct ReplayNodes<S>
+where
+    S: State,
+{
+    pub replay_nodes_info: BTreeMap<PeerId, NodesInfo<S>>,
+}
+
+impl<S> ReplayNodes<S>
+where
+    S: State,
+{
+    pub fn new(peers: Vec<(PubKey, S::Config)>) -> Self {
+        assert!(!peers.is_empty());
+        let mut replay_nodes_info = BTreeMap::new();
+        for (pubkey, state_config) in peers {
+            let (state, _) = S::init(state_config);
+            let blockchain = Vec::new();
+            let pending_messages = Vec::new();
+            let nodes_info = NodesInfo {
+                state,
+                blockchain,
+                pending_messages,
+            };
+            replay_nodes_info.insert(PeerId(pubkey), nodes_info);
+        }
+        Self { replay_nodes_info }
+    }
+
+    pub fn step(&mut self, node_id: &PeerId, event: S::Event, tick: Duration) {
+        let state = self.replay_nodes_info.get_mut(node_id).unwrap().mut_state();
+        let commands = state.update(event);
+        self.mutate_state(node_id, tick, commands);
+    }
+
+    fn mutate_state(
+        &mut self,
+        node_id: &PeerId,
+        tick: Duration,
+        cmds: Vec<
+            Command<<S as State>::Message, <S as State>::OutboundMessage, <S as State>::Block>,
+        >,
+    ) {
+        let mut to_publish = Vec::new();
+        let mut to_unpublish = HashSet::new();
+
+        for command in cmds {
+            match command {
+                Command::LedgerCommand(cmd) => match cmd {
+                    LedgerCommand::LedgerCommit(b) => {
+                        let block = self
+                            .replay_nodes_info
+                            .get_mut(node_id)
+                            .unwrap()
+                            .mut_blockchain();
+                        block.push(b);
+                    }
+                },
+                Command::RouterCommand(cmd) => match cmd {
+                    RouterCommand::Publish { target, message } => {
+                        to_publish.push((target, message));
+                    }
+                    RouterCommand::Unpublish { target, id } => {
+                        to_unpublish.insert((target, id));
+                    }
+                },
+                _ => {}
+            }
+        }
+        for (target, out_message) in to_publish {
+            let id = out_message.as_ref().id();
+            if to_unpublish.contains(&(target, id)) {
+                continue;
+            }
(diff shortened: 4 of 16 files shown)


monad-executor/src/executor/mock.rs
```diff
@@ -468,6 +467,34 @@ mod tests {
         Vote { peer: PeerId, round: u64 },
     }
 
+    fn peer_id_to_bytes(peer: &PeerId) -> Vec<u8> {
+        let peer_str = format!("{:?}", peer);
+        peer_str.into_bytes()
```
omegablitz: This implementation looks suspicious to me because it depends on the Debug trait. I think you can avoid all of this and just put `unreachable!()` inside the `serialize` and `deserialize` impls.

monad-executor/src/lib.rs
```diff
@@ -4,8 +4,9 @@ mod state;
 pub use state::*;
 
 pub mod executor;
-
+pub mod timedevent;
```
omegablitz: nit: `timed_event` to keep it consistent

monad-testutil/src/swarm.rs
```diff
@@ -9,22 +9,28 @@ use monad_consensus::{
...
     mock_swarm::{LinkMessage, Nodes, Transformer},
+    timedevent::TimedEvent,
     PeerId, State,
 };
 use monad_state::{MonadConfig, MonadEvent, MonadMessage, MonadState};
 use monad_wal::mock::{MockWALogger, MockWALoggerConfig};
 use monad_wal::PersistenceLogger;
 
-type SignatureType = NopSignature;
+type SignatureType = SecpSignature;
```
omegablitz: We probably shouldn't change the default `SignatureType` here - `SecpSignature` is very significantly slower. We might need to implement Ser/Des for NoopSignature but maybe that's fine? @ariqchowdhury @michael-yxchen thoughts?

monad-executor/src/timedevent.rs
```diff
@@ -0,0 +1,53 @@
...
+use std::fmt::Debug;
+use std::time::Duration;
+
+# [derive(Clone)]
+pub struct TimedEvent<M> {
+    pub timestamp: Duration, // ticks Duration in milliseconds
+    pub event: M,
+}
+
+impl<M> Serializable for TimedEvent<M>
```
omegablitz: nit: can simplify all the trait bounds down to `impl<M: Serializable> Serializable for TimedEvent<M>`

monad-executor/src/timedevent.rs
```diff
@@ -0,0 +1,53 @@
...
+{
+    fn serialize(&self) -> Vec<u8> {
+        let mut buf = Vec::new();
+        buf.extend_from_slice(&self.timestamp.as_secs_f64().to_be_bytes());
+        buf.extend_from_slice(&self.event.serialize());
+        buf
+    }
+}
+
+impl<M> Deserializable for TimedEvent<M>
```
omegablitz: nit: can simplify all the trait bounds down to `impl<M: Deserializable> Deserializable for TimedEvent<M>`

monad-executor/src/timedevent.rs
```diff
@@ -0,0 +1,53 @@
...
+    type ReadError = TryFromSliceError;
+    fn deserialize(buf: &[u8]) -> Result<Self, Self::ReadError> {
+        let (timestamp, buf) = buf.split_at(std::mem::size_of::<f64>());
+        let timestamp = f64::from_be_bytes(timestamp.try_into().unwrap());
+        let event = M::deserialize(buf).unwrap();
+        Ok(Self { timestamp: Duration::from_secs_f64(timestamp), event })
+    }
+}
+
+impl<M> Debug for TimedEvent<M>
```
omegablitz: nit: can simplify all the trait bounds down to `impl<M: Debug> Debug for TimedEvent<M>`

monad-executor/src/timedevent.rs
```diff
@@ -0,0 +1,53 @@
...
+{
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("TimedEvent")
+            .field("ticks", &self.timestamp)
+            .field("event", &self.event)
+            .finish()
+    }
+}
+
+impl<M> TimedEvent<M> {
```
omegablitz: super nit: probably no reason to have this constructor if both members you're initializing are public anyways

monad-testutil/src/main.rs
omegablitz, line 1: I think we should move the stuff in here somewhere else - maybe in an [example](http://xion.io/post/code/rust-examples.html)?

monad-testutil/src/main.rs
```diff
@@ -0,0 +1,67 @@
+use monad_state::{MonadEvent, MonadState};
+use std::time::Duration;
+type MS = MonadState<SignatureType, SignatureCollectionType>;
+type MM = <MS as State>::Message;
```
omegablitz: These imports could be put in groups better - they're kinda confusing as is

monad-executor/src/mock_swarm.rs
```diff
@@ -238,7 +240,10 @@ where
             .min()
     }
 
-    pub fn step(&mut self) -> Option<(Duration, PeerId, S::Event)> {
+    pub fn step(&mut self) -> Option<(Duration, PeerId, S::Event)>
+    where
+        <S as State>::Event: Serializable + Deserializable + Debug,
```
omegablitz: Are the trait bounds here necessary?

monad-executor/src/replay_nodes.rs
```diff
@@ -0,0 +1,130 @@
...
+use std::collections::BTreeMap;
+use std::collections::HashSet;
+use std::fmt::Debug;
+use std::time::Duration;
+
+pub struct ReplayNodes<S>
+where
+    S: State,
+{
+    states: BTreeMap<PeerId, S>,
```
omegablitz: Could we have a single map with a struct as a value per PeerId?

monad-executor/src/replay_nodes.rs
```diff
@@ -0,0 +1,130 @@
+use crate::{Command, LedgerCommand, Message, PeerId, RouterCommand, RouterTarget, State};
+use monad_crypto::secp256k1::PubKey;
+use monad_types::{Deserializable, Serializable};
+use std::collections::BTreeMap;
+use std::collections::HashSet;
```
omegablitz: Group imports, maybe std stuff on top, and monad stuff below?

monad-executor/src/replay_nodes.rs
```diff
@@ -0,0 +1,130 @@
...
+use std::time::Duration;
+
+pub struct ReplayNodes<S>
+where
+    S: State,
+{
+    states: BTreeMap<PeerId, S>,
+    blockchains: BTreeMap<PeerId, Vec<S::Block>>,
+    pending_messages: BTreeMap<PeerId, Vec<(PeerId, Duration, S::Event, <S as State>::Message)>>,
+    // ReceiverId: (SenderId, Sendtick, Event, Message)
```
omegablitz: I think it's more readable to split out this tuple into a struct instead of leaving it in a comment

monad-executor/src/replay_nodes.rs
```diff
@@ -0,0 +1,130 @@
...
+}
+
+impl<S> ReplayNodes<S>
+where
+    S: State,
+    S::Event: Serializable + Deserializable + Debug,
```
omegablitz: Are all these trait bounds necessary here?

monad-executor/src/replay_nodes.rs
```diff
@@ -0,0 +1,130 @@
...
+        }
+    }
+
+    pub fn step(&mut self, node_id: &PeerId, event: S::Event, tick: Duration) {
+        let state = self.states.get_mut(node_id).unwrap();
+        let commands = state.update(event);
+        self.parse_cmds(node_id, tick, commands);
+    }
+
+    fn parse_cmds(
```
omegablitz: I think this function could be named better - it seems weird to me that a function named `parse_cmds` mutates state.

monad-executor/src/replay_nodes.rs
```diff
@@ -0,0 +1,130 @@
...
+        for (target, out_message) in to_publish {
+            let id = out_message.as_ref().id();
+            if to_unpublish.contains(&(target, id)) {
+                continue;
+            }
+            let message = out_message.into();
+            let peer_ids: Vec<_> = self.states().keys().copied().collect();
+            let tos = match target {
+                RouterTarget::PointToPoint(to) => vec![to],
+                RouterTarget::Broadcast => peer_ids.clone(),
```
omegablitz: Can the `peer_ids` creation be moved into the `RouterTargeT::Broadcast` branch?

monad-executor/src/replay_nodes.rs
```diff
@@ -0,0 +1,130 @@
...
+            };
+            for to in tos {
+                if to != *node_id {
+                    let msg_queue = self.pending_messages.get_mut(&to).unwrap();
+                    msg_queue.push((
+                        *node_id,
+                        tick,
+                        message.clone().event(*node_id),
```
omegablitz: I think the first clone could be avoided

monad-executor/src/replay_nodes.rs
```diff
@@ -0,0 +1,130 @@
...
+                continue;
+            }
+            let message = out_message.into();
+            let peer_ids: Vec<_> = self.states().keys().copied().collect();
+            let tos = match target {
+                RouterTarget::PointToPoint(to) => vec![to],
+                RouterTarget::Broadcast => peer_ids.clone(),
+            };
+            for to in tos {
+                if to != *node_id {
```
omegablitz: super nit: It's probably more idiomatic to do something like `for to in tos.iter().filter(...)` - avoiding nesting whenever possible is always nice

monad-viz/src/main.rs
```diff
@@ -17,27 +19,50 @@ use iced::{
...
 };
+use monad_executor::timedevent::TimedEvent;
+use monad_executor::PeerId;
 use monad_executor::State;
 use monad_state::{MonadEvent, MonadState};
 use monad_wal::mock::MockWALogger;
+use replay_graph::{RepConfig, ReplayNodesSimulation};
 
-type SignatureType = NopSignature;
+type SignatureType = SecpSignature;
```
omegablitz: Shouldn't need to use SecpSignature here - it's significantly slower

monad-viz/src/main.rs
```diff
@@ -17,27 +19,50 @@ use iced::{
...
 type SignatureCollectionType = MultiSig<SignatureType>;
 type MS = MonadState<SignatureType, SignatureCollectionType>;
 type MM = <MS as State>::Message;
-type PersistenceLoggerType = MockWALogger<MonadEvent<SignatureType, SignatureCollectionType>>;
+type PersistenceLoggerType =
+    MockWALogger<TimedEvent<MonadEvent<SignatureType, SignatureCollectionType>>>;
 type Sim = NodesSimulation<MS, LayerTransformer<MM>, PersistenceLoggerType, SimConfig>;
+type ReplaySim = ReplayNodesSimulation<MS, RepConfig>;
+use monad_wal::wal::{WALogger, WALoggerConfig};
+use monad_wal::PersistenceLogger;
```
omegablitz: Organize these imports into the appropriate groups above

monad-viz/src/main.rs
```diff
@@ -17,27 +19,50 @@ use iced::{
...
+use std::cmp::max;
+use std::env;
+use std::path::PathBuf;
 
 pub fn main() -> iced::Result {
-    Viz::run(Settings::default())
+    let mut logs_enabled = false;
```
omegablitz: You can use the [clap](https://docs.rs/clap/latest/clap/) library for CLI flags - will be easier to extend and work with in the future

monad-viz/src/main.rs
```diff
@@ -49,75 +74,138 @@ enum Message {
 
 impl Application for Viz {
     type Executor = executor::Default;
-    type Flags = ();
+    type Flags = bool;
```
omegablitz: I think it'd be more readable to create a separate config/flags struct with params

monad-viz/src/main.rs
```diff
@@ -49,75 +74,138 @@ enum Message {
...
+        if flags {
+            let mut config = RepConfig {
+                num_nodes: 4,
+                delta: Duration::from_millis(101),
+                max_tick: Duration::from_secs_f32(0.0),
+            };
+            let pubkeys = config.nodes();
+            let mut replay_events = BTreeMap::new();
+            for (pk, _) in pubkeys.iter() {
+                let log_file_name = format!("../{:?}.log", pk);
```
omegablitz: I think it'd be a better UX if the location of these log files could be determined from the CLI flags

monad-viz/src/main.rs
```diff
@@ -49,75 +74,138 @@ enum Message {
...
+                        (sim.tick().as_secs_f32() + delta)
+                            .clamp(sim.min_tick().as_secs_f32(), sim.max_tick().as_secs_f32()),
+                    ));
+                }
+                Message::SetConfig(config) => {
+                    sim.update_config(config);
+                }
+            },
+            SimType::RS(ref mut sim) => match message {
+                Message::SetTick(tick) => {
```
omegablitz, line 175: Can the match branches for `SimType::S(Message::SetTick(_))` and `SimType::RS(Message::SetTick(_))` be unified? same for AddTick

monad-viz/src/main.rs
```diff
@@ -49,75 +74,138 @@ enum Message {
...
-                self.simulation.config(),
-                Message::SetConfig,
-            ))
-            .into()
+                .push(Canvas::new(sim).width(Length::Fill).height(Length::Fill))
+                .push(ConfigEditor::new(sim.config(), Message::SetConfig))
+                .into(),
+            SimType::RS(ref sim) => Row::new()
+                .push(
+                    VerticalSlider::new(
```
omegablitz, line 206: The `VerticalSlider` initialization could be lifted above the match to reduce code duplication

monad-viz/src/main.rs
```diff
@@ -140,6 +228,109 @@ impl Application for Viz {
     }
 }
 
+impl Program<Message> for &ReplaySim {
+    type State = ();
+
+    fn draw(
```
omegablitz, line 316: Didn't look at this too closely, but can any of the stuff in here be factored out so that it's not duplicated between here and the `&Sim` implementation?

monad-viz/src/replay_graph.rs
```diff
@@ -0,0 +1,163 @@
...
+use monad_executor::{
+    executor::mock::MockExecutor, replay_nodes::ReplayNodes, timedevent::TimedEvent, Message,
+    PeerId, State,
+};
+use monad_state::MonadConfig;
+use monad_testutil::signing::{create_keys, get_genesis_config};
+use monad_types::{Deserializable, Serializable};
+use std::collections::BTreeMap;
+use std::fmt::Debug;
+use std::time::Duration;
```
omegablitz: Group imports

monad-executor/src/mock_swarm.rs
victorzdm: They are not, removing them.

monad-executor/src/replay_nodes.rs
victorzdm: It will complain about `use of moved value`

monad-viz/src/main.rs
```diff
@@ -1,43 +1,96 @@
...
-type PersistenceLoggerType = MockWALogger<MonadEvent<SignatureType, SignatureCollectionType>>;
+type PersistenceLoggerType =
+    MockWALogger<TimedEvent<MonadEvent<SignatureType, SignatureCollectionType>>>;
 type Sim = NodesSimulation<MS, LayerTransformer<MM>, PersistenceLoggerType, SimConfig>;
+type ReplaySim = ReplayNodesSimulation<MS, RepConfig>;
+
+#[derive(Parser)]
+struct Arg {
+    log: Option<String>,
+    log_loc: Option<String>,
```
omegablitz: Would it be sufficient to only have one argument instead, something like `logdir: Option<PathBuf>`? The `log` arg seems redundant to me.

monad-viz/src/main.rs
```diff
@@ -1,43 +1,96 @@
...
+}
+
+#[derive(Default)]
+struct ConfigFlag {
+    pub log: bool,
+    pub log_loc: Option<String>,
```
omegablitz: Same as above, feels like we should only need a single arg here. Also, instead of creating a new struct, can probably just reuse the above `Arg` struct instead of defining a new struct here.

monad-viz/src/main.rs
```diff
@@ -1,43 +1,96 @@
...
+}
 
 pub fn main() -> iced::Result {
-    Viz::run(Settings::default())
+    let cli = Arg::parse();
+    if cli.log.is_some() && cli.log.unwrap() == "logs" {
+        Viz::run(Settings::<ConfigFlag> {
```
omegablitz: These two branches can probably be unified if the above comments are addressed

monad-viz/src/main.rs
```diff
@@ -49,75 +102,142 @@ enum Message {
...
+        if flags.log {
+            let mut config = RepConfig {
+                num_nodes: 4,
+                delta: Duration::from_millis(101),
+                max_tick: Duration::from_secs_f32(0.0),
+            };
+            let pubkeys = config.nodes();
+            let mut replay_events = BTreeMap::new();
+            for (pk, _) in pubkeys.iter() {
+                let log_file_name = format!(
```
omegablitz: If we parse the filepath into a  PathBuf`, can use https://doc.rust-lang.org/stable/std/path/struct.PathBuf.html#method.push directly instead

monad-viz/src/main.rs
```diff
@@ -49,75 +102,142 @@ enum Message {
...
+                    "{}/{:?}.log",
+                    flags.log_loc.clone().unwrap_or(String::from("..")),
```
omegablitz: Shouldn't need an unwrap here if the stuff mentioned above is changed

monad-viz/src/main.rs
```diff
@@ -1,43 +1,96 @@
...
+#[derive(Default)]
+struct ConfigFlag {
+    pub log: bool,
+    pub log_loc: Option<String>,
+}
 
 pub fn main() -> iced::Result {
-    Viz::run(Settings::default())
+    let cli = Arg::parse();
+    if cli.log.is_some() && cli.log.unwrap() == "logs" {
```
omegablitz: Once the above is changed, this could be `if let Some(logdir) = cli.logdir { ... }` to avoid the unwrap

monad-viz/src/main.rs
victorzdm: ended up using the `join` method cause the `Pathbuf` is immutable. To use `push`, we have to make a mutable clone of the `logdir` and push and change it back to immutable type. `Join` is clearer and requires less code

PR #143, opened 2023-06-27
Remove Validator type to simplify ValidatorSet

ariqchowdhury:
Previously, ValidatorSet kept a map of NodeId to Validator but the Validator type only has the NodeId and Stake. We should just map the NodeId to Stake in ValidatorSet and remove the Validator type.

If in the future, there is more information to track about a Validator, I think we should add a ValidatorProperty type that has Stake and whatever future information gets added, and the mapping stays the same in ValidatorSet.

I'm expecting this to simplify the proto serialization for ValidatorSet which I'm adding for another PR.


monad-validator/src/validator_set.rs
```diff
@@ -30,33 +29,30 @@ impl error::Error for ValidatorSetError {
...
         let mut vmap = HashMap::new();
         let mut vlist = Vec::new();
         for v in validators.into_iter() {
-            let entry = vmap.entry(NodeId(v.pubkey)).or_insert(v);
-            if entry.stake != v.stake {
-                return Err(ValidatorSetError::DuplicateValidator(format!(
-                    "{:?}",
-                    v.pubkey
-                )));
+            let entry = vmap.entry(v.0).or_insert(v.1);
```
omegablitz: super nit: I think `for (node_id, stake) in validators` is easier to read

monad-state/src/lib.rs
```diff
@@ -226,10 +226,7 @@ where
         let mut monad_state: MonadState<ST, SCT> = Self {
             message_state: MessageState::new(
                 10,
-                validator_list
-                    .into_iter()
-                    .map(|v| PeerId(v.pubkey))
-                    .collect(),
+                validator_list.into_iter().map(|v| PeerId(v.0 .0)).collect(),
```
omegablitz: super nit: I think `|(NodeId(node_id), _)| PeerId(node_id)` is easier to read

PR #145, opened 2023-06-28
Add initial skeleton for Checkpoints and Epochs

ariqchowdhury:
Checkpoints and Epochs involve side effects so they require an executor as well as Events and Commands. A Checkpoint is created via a command from consensus state when necessary. Loading checkpoints and reacting to Epoch changes are external input from Events.

This change adds the supporting types for the above but doesn't implement any behaviour so there should be no functional changes to how things work yet. Future changes will flesh out the particular details for checkpoint and epoch implementations.

Adding new events requires updating the serialization types as well.

### DIFF
--- a/monad-executor/src/executor/checkpoint.rs
+++ b/monad-executor/src/executor/checkpoint.rs
@@ -1,0 +1,25 @@
+use crate::{CheckpointCommand, Executor};
+
+pub struct MockCheckpoint<C> {
+    pub checkpoint: Option<C>,
+}
+
+impl<C> Default for MockCheckpoint<C> {
+    fn default() -> Self {
+        Self { checkpoint: None }
+    }
+}
+
+impl<C> Executor for MockCheckpoint<C> {
+    type Command = CheckpointCommand<C>;
+    fn exec(&mut self, commands: Vec<Self::Command>) {
+        for command in commands {
+            match command {
+                CheckpointCommand::Save(c) => {
+                    self.checkpoint = Some(c);
+                    println!("checkpoint saved");
+                }
+            }
+        }
+    }
+}
--- a/monad-proto/proto/validator_set.proto
+++ b/monad-proto/proto/validator_set.proto
@@ -1,0 +1,14 @@
+syntax = "proto3";
+
+package monad_proto.validator_set;
+
+import "basic.proto";
+
+message ValidatorMapEntry {
+    monad_proto.basic.ProtoNodeId key = 1;
+    monad_proto.basic.ProtoStake value = 2;
+}
+
+message ProtoValidatorSetData {
+    repeated ValidatorMapEntry validators = 1;
+}
--- a/monad-state/src/lib.rs
+++ b/monad-state/src/lib.rs
@@ -8,7 +8,7 @@ use monad_consensus::{
     validation::signing::{Unverified, Verified},
 };
 use monad_consensus_state::{
-    command::{ConsensusCommand, FetchedFullTxs, FetchedTxs},
+    command::{Checkpoint, ConsensusCommand, FetchedFullTxs, FetchedTxs},
     ConsensusProcess,
 };
 use monad_consensus_types::{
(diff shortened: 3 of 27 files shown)


monad-executor/src/executor/checkpoint.rs
```diff
@@ -0,0 +1,25 @@
...
+}
+
+impl<C> Executor for MockCheckpoint<C> {
+    type Command = CheckpointCommand<C>;
+    fn exec(&mut self, commands: Vec<Self::Command>) {
+        for command in commands {
+            match command {
+                CheckpointCommand::Save(c) => {
+                    self.checkpoint = Some(c);
+                    println!("checkpoint saved");
```
omegablitz, line 20: Is this for debug? If so, maybe worth changing to use the `log` crate

monad-state/src/lib.rs
```diff
@@ -65,6 +67,27 @@ where
     pub fn blocktree(&self) -> &BlockTree<SCT> {
         &self.consensus_state.pending_block_tree
     }
+
+    fn advance_epoch(&mut self, val_set: Option<ValidatorSet<LeaderElectionType>>) {
+        self.epoch = self.epoch + Epoch(1);
+
+        if let Some(vs) = &self.upcoming_validator_set {
+            self.validator_set = vs.clone();
```
omegablitz: super nit: I think we can dodge the clone, eg `if let Some(vs) = self.upcoming_validator_set.take()`

monad-executor/src/executor/checkpoint.rs
ariqchowdhury: that was just for debug, should be removed. When we know what we want the executor to do for that command we can log appropriately.

ariqchowdhury:
> LGTM

Rebasing onto the MonadState refactor changed the new events a little bit (We use the Validator Set trait in MonadState now) so might be worth a second look. The relevant changes post-rebase are in the second commit so you can probably just look at that.

monad-proto/proto/validator_set.proto
```diff
@@ -0,0 +1,32 @@
...
+    monad_proto.basic.ProtoNodeId address = 1;
+    monad_proto.basic.ProtoStake voting_power = 2;
+    monad_proto.basic.ProtoStake priority = 3;
+}
+
+message ProtoWeightedRoundRobin {
+    repeated ProtoVoter voters = 1;
+    uint64 leader = 2;
+    monad_proto.basic.ProtoStake total_voting_power = 3;
+}
```
michael-yxchen: nit: same here

monad-proto/proto/validator_set.proto
```diff
@@ -0,0 +1,32 @@
...
+message ValidatorMapEntry {
+    monad_proto.basic.ProtoNodeId key = 1;
+    monad_proto.basic.ProtoStake value = 2;
+}
+
+message ProtoValidatorSet {
+    repeated ValidatorMapEntry validators = 1;
+    repeated monad_proto.basic.ProtoNodeId validator_list = 2;
+    monad_proto.basic.ProtoStake total_stake = 3;
+}
```
michael-yxchen: nit: idk if it is used right now. If it's completely replaced by ValidatorSetData, maybe we can remove it.

monad-validator/src/validator_set.rs
```diff
@@ -44,13 +46,55 @@ pub trait ValidatorSetType {
...
+                .value
+                .ok_or(Self::Error::MissingRequiredField(
+                    "ValildatorMapEntry.value".to_owned(),
+                ))?
+                .try_into()?;
+
+            vlist.0.push((a, b));
+        }
+
+        Ok(Self(Vec::new()))
```
omegablitz: typo

monad-proto/proto/validator_set.proto
ariqchowdhury: yeah, you're right, its not used anywhere else, I'll remove it.

PR #146, opened 2023-06-28
Add BLS aggregate signature primitives

michael-yxchen:
Certificates (QC and TC) are implemented with MultiSig in our codebase. Its size scales linearly with the number of signers it carries - each additional signer is another ~64 bytes on the messages. It poses difficulty to scaling to more validators as throughput is bottlenecked by network bandwidth.

BLS signature scheme enables aggregation of signatures over the same or different messages. QC/TC messages can be made smaller when we aggregate the signatures on votes/timeout message.

### DIFF
--- a/monad-crypto/src/bls12_381.rs
+++ b/monad-crypto/src/bls12_381.rs
@@ -1,0 +1,885 @@
+use blst::min_pk as blst_core;
+use zeroize::Zeroize;
+
+// if the curve is switched to min_sig
+// the DST needs to be BLS_SIG_BLS12381G1_XMD:SHA-256_SSWU_RO_POP_
+// POP uses a separate pubkey validation step, enables fast verification for
+// signatures over the same message
+// https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-05#name-proof-of-possession
+const DST: &[u8] = MIN_PK_DST;
+#[allow(dead_code)]
+const MIN_PK_DST: &[u8] = b"BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_";
+#[allow(dead_code)]
+const MIN_SIG_DST: &[u8] = b"BLS_SIG_BLS12381G1_XMD:SHA-256_SSWU_RO_POP_";
+
+// TODO: simplify group swap with macro?
+const SIGNATURE_BYTE_LEN: usize = G2_BYTE_LEN;
+const SIGNATURE_COMPRESSED_LEN: usize = G2_COMPRESSED_LEN;
+const INFINITY_SIGNATURE: [u8; SIGNATURE_COMPRESSED_LEN] = G2_INFINITY;
+
+const PUBKEY_BYTE_LEN: usize = G1_BYTE_LEN;
+const PUBKEY_COMPRESSED_LEN: usize = G1_COMPRESSED_LEN;
+const INFINITY_PUBKEY: [u8; PUBKEY_COMPRESSED_LEN] = G1_INFINITY;
+
+const G1_BYTE_LEN: usize = 96;
+const G1_COMPRESSED_LEN: usize = 48;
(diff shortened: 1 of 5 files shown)


monad-crypto/src/bls12_381.rs
```diff
@@ -0,0 +1,842 @@
+use std::hash::Hash;
+
+use blst::min_pk as blst_core;
+use zeroize::Zeroize;
+
+// if the curve is switched to min_sig
+// the DST needs to be BLS_SIG_BLS12381G1_XMD:SHA-256_SSWU_RO_POP_
```
ariqchowdhury, line 5: If its possible we need this one day, can we define a const for this (DST_MIN_SIG or whatever), define the other one and then assign the one we are using to the DST const.
There's only one char difference between the strings so it looked to me like we were using the switched curve.

monad-crypto/src/bls12_381.rs
```diff
@@ -0,0 +1,842 @@
...
+        Self(value)
+    }
+}
+
+impl Hash for BlsPubKey {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        let slice = self.serialize();
+        slice.hash(state);
+    }
+}
```
ariqchowdhury: For the secp curve, I found that serialize was super slow -- the implementation does work to make sure that the serialized form can be compared across different library versions. It was much faster to just hash the raw underlying bytes. I'm not sure what its like for this library but we should measure it.

monad-crypto/src/bls12_381.rs
```diff
@@ -0,0 +1,842 @@
...
+        self.as_pubkey().hash(state)
+    }
+}
+
+impl PartialEq for BlsAggregatePubKey {
+    fn eq(&self, other: &Self) -> bool {
+        self.as_pubkey().eq(&other.as_pubkey())
+    }
+}
+
```
ariqchowdhury, line 116: similar to the hash thing, the secp library had a special `eq_fast_unstable` function -- not sure if that's relevant here, but something to check.

monad-crypto/src/bls12_381.rs
```diff
@@ -0,0 +1,842 @@
...
+impl Eq for BlsAggregatePubKey {}
+
+impl BlsAggregatePubKey {
+    pub fn infinity() -> Self {
+        Self::deserialize(&INFINITY_PUBKEY).unwrap()
```
ariqchowdhury: `expect()` ?

monad-crypto/src/bls12_381.rs
michael-yxchen: yeah make sense. I'll do that

monad-crypto/src/bls12_381.rs
michael-yxchen: Noted. I'll test transmute vs serialize. The pubkey/signature is a point in 2d. It looks like serialization is performing some deterministic operation in each coordinate and concat-ing the result together. Haven't figure out why it's doing it tho

monad-crypto/src/bls12_381.rs
michael-yxchen: the PartialEq implementations bind to these two functions `blst_p1_affine_is_equal, blst_p2_affine_is_equal`. They are almost comparing the underlying memory directly, so should be performant.

monad-crypto/src/bls12_381.rs
michael-yxchen: Hash with transmute is about 4x-6x faster (30ns vs 130ns-180ns) than serialize. Will switch to transmute implementation

ariqchowdhury: We should probably figure out why they do serialize the way they do

michael-yxchen: yeah working on that

PR #147, opened 2023-06-29
Implement Drop trait for ConsensusState for debug prints

ariqchowdhury:
Drop trait allows an implementer to run code on destruction. We can use this to implement a debug print to stderr that will be called on a panic as the stack unwinds.

### DIFF
--- a/monad-state/src/lib.rs
+++ b/monad-state/src/lib.rs
@@ -443,6 +443,26 @@ struct ConsensusState<S, T> {
     keypair: KeyPair,
 }
 
+struct ConsensusStateWrapper<S: Signature, T: SignatureCollection> {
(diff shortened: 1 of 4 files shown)


monad-state/src/lib.rs
```diff
@@ -443,6 +443,26 @@ struct ConsensusState<S, T> {
     keypair: KeyPair,
 }
 
+struct ConsensusStateWrapper<S: Signature, T: SignatureCollection> {
```
omegablitz, line 446: Did we want to rename this? Don't mind either way, just putting a comment here in case you were intending to but forgot about it

ariqchowdhury: I'm okay with this name for now

PR #152, opened 2023-07-07
Fetch txs from mempool and validate on proposal

abenedito-monad:
Currently, we have no way of validating the transactions provided in proposals. This is a draft for a proposed solution which involves creating a new TransactionCollection generic and adding an additional command to fetch the full transactions from the mempool so that they can be validated for proposals.

Todo:
- [x] Change Vec<TransactionCollection> to TransactionCollection
- [x] Finish protobuf serialization / deserialization

### DIFF
--- a/monad-executor/src/state.rs
+++ b/monad-executor/src/state.rs
@@ -1,5 +1,6 @@
 use std::hash::Hash;
 
+use monad_consensus_types::transaction::TransactionCollection;
 use monad_crypto::secp256k1::PubKey;
 
 #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
@@ -37,13 +38,15 @@ pub enum TimerCommand<E> {
     ScheduleReset,
 }
 
-pub enum MempoolCommand<E> {
+pub enum MempoolCommand<E, TC> {
     // TODO consider moving away from dynamic dispatch
     /// FetchReset should ALMOST ALWAYS be emitted by the state machine after handling E
     /// This is to prevent E from firing twice on replay
     // TODO create test to demonstrate faulty behavior if written improperly
     FetchTxs(Box<dyn (FnOnce(Vec<u8>) -> E) + Send + Sync>),
     FetchReset,
+    FetchFullTxs(Box<dyn (FnOnce(TC) -> E) + Send + Sync>),
+    FetchFullReset,
 }
 
 pub enum LedgerCommand<B> {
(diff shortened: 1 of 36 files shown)


monad-types/src/transaction.rs
ariqchowdhury, line 1: I think this should be moved into monad-consensus-types

monad-executor/src/state.rs
```diff
@@ -98,26 +103,27 @@ pub trait State: Sized {
...
     fn update(
         &mut self,
         event: Self::Event,
-    ) -> Vec<Command<Self::Message, Self::OutboundMessage, Self::Block>>;
+    ) -> Vec<Command<Self::Message, Self::OutboundMessage, Self::Block, Self::TransactionType>>;
 }
 
 pub trait Message: Clone {
-    type Event;
+    type Event<TC: TransactionCollection>;
```
ariqchowdhury, line 125: I'm not following why Event needs to be parameterized over TC. It's making other things look weird like,`Vec<TimerCommand<M::Event<TC>>` for example because of this, so just want to understand why its necessary and if there's an alternative.

monad-state/src/lib.rs
```diff
@@ -283,8 +297,17 @@ where
...
                         cmds
                     }
+                    ConsensusEvent::FetchedFullTxs(fetched_txs) => self
+                        .consensus_state
+                        .handle_proposal_message_full::<HasherType, LeaderElectionType>(
+                        fetched_txs.author,
+                        fetched_txs.p,
+                        fetched_txs.txns,
+                        &mut self.validator_set,
+                    ),
```
ariqchowdhury: do we need to check that the fetched_txs are for the current round of consensus state? Or at least have an assert.
But I think it might be possible for the fetched_txs to be expired -- proposal comes in, starts fetch, round times out, timeout certificate advances the round, fetch completes.

monad-state/src/lib.rs
```diff
@@ -397,6 +427,12 @@ impl<S: Debug, T: Debug> Debug for ConsensusEvent<S, T> {
                 .finish(),
             ConsensusEvent::Timeout(p) => p.fmt(f),
             ConsensusEvent::FetchedTxs(p) => p.fmt(f),
+            ConsensusEvent::FetchedFullTxs(p) => f
+                .debug_struct("FetchedFullTxs")
+                .field("author", &p.author)
+                .field("payload", &p.p)
```
ariqchowdhury: nit: payload -> proposal

monad-state/src/lib.rs
```diff
@@ -476,10 +524,25 @@ where
...
+        txns: T,
         validators: &mut ValidatorSet<V>,
-    ) -> Vec<ConsensusCommand<S, T>> {
-        let mut cmds = Vec::new();
+    ) -> Vec<ConsensusCommand<S, SC, T>> {
+        let mut cmds = vec![ConsensusCommand::FetchFullTxsReset];
+
+        if !txns.validate() {
+            return cmds;
+        }
```
ariqchowdhury, line 599: We need to write some tests to verify that if a node fails validate here while the rest of the nodes succeed (possible with shared mempool if txns don't propagate in time), that node still catches up properly when the next proposal comes. I added a task in Notion to track this.

monad-types/src/transaction.rs
abenedito-monad: Agreed, monad-consensus-types didn't exist when I started, just rebased and moved!

monad-state/src/lib.rs
abenedito-monad: Good catch!

monad-state/src/lib.rs
abenedito-monad: Yep just added, let me know if you think it looks good.

PR #154, opened 2023-07-10
Change the id of NopSignature

vicdunmin:
Previously, the id of the NopSignature was using a random number generated at runtime. This is not ideal, as it makes the MonadEvent impossible to compare when we replay the event log during the visualization. Thus, in order to provide a consistent way of doing equality comparisons, we now use the hash of the message as the id in the signature function.

### DIFF
--- a/monad-crypto/src/lib.rs
+++ b/monad-crypto/src/lib.rs
@@ -1,14 +1,13 @@
-use std::hash::Hash;
-
-use rand::Rng;
+use std::collections::hash_map::DefaultHasher;
+use std::hash::{Hash, Hasher};
 
 use crate::secp256k1::{Error, KeyPair, PubKey};
 
 #[cfg(feature = "proto")]
 pub mod convert;
 
-pub mod secp256k1;
 pub mod bls12_381;
+pub mod secp256k1;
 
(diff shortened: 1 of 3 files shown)


monad-crypto/src/lib.rs
```diff
@@ -31,12 +30,13 @@ pub struct NopSignature {
...
-    fn sign(_msg: &[u8], keypair: &KeyPair) -> Self {
-        let mut rng = rand::thread_rng();
+    fn sign(msg: &[u8], keypair: &KeyPair) -> Self {
+        let mut hasher = DefaultHasher::new();
+        hasher.write(msg);
 
         NopSignature {
             pubkey: keypair.pubkey(),
-            id: rng.gen::<u32>(),
+            id: hasher.finish() as u32,
```
omegablitz: You can change the `id` field to be a u64 instead of downcasting this

monad-crypto/src/lib.rs
```diff
@@ -1,14 +1,13 @@
-use std::hash::Hash;
-
-use rand::Rng;
```
omegablitz, line 3: Is `rand` used anywhere else in this crate? If not, you can remove it from the Cargo.toml also

PR #158, opened 2023-07-12
Add stake verification to QC and TC

michael-yxchen:
The current certificate verification only verifies the SignatureCollection and doesn't check whether +2/3 stake has signed the message. Any SignatureCollection is treated as valid so long as the signature is valid. This is wrong and this PR patches that, and adds two stake validation test. 

Verify_qc/tc now takes in the ValidatorSet for staking information.

### DIFF
--- a/monad-consensus-types/src/validation.rs
+++ b/monad-consensus-types/src/validation.rs
@@ -13,6 +13,8 @@ pub enum Error {
     AuthorNotSender,
     /// There are high qc rounds larger than the TC round
     InvalidTcRound,
+    /// The SignatureCollection doesn't have supermajority of the stake signed
+    InsufficientStake,
 }
 
 pub trait Hashable {
--- a/monad-consensus/src/validation/signing.rs
+++ b/monad-consensus/src/validation/signing.rs
@@ -17,6 +17,7 @@ use monad_proto::proto::message::{
     proto_unverified_consensus_message, ProtoUnverifiedConsensusMessage,
 };
 use monad_types::{Hash, NodeId, Stake};
+use monad_validator::validator_set::ValidatorSetType;
 
 #[cfg(feature = "proto")]
 use crate::convert::message::UnverifiedConsensusMessage;
@@ -104,9 +105,9 @@ where
 {
     // A verified proposal is one which is well-formed and has valid
     // signatures for the present TC or QC
-    pub fn verify<H: Hasher>(
+    pub fn verify<H: Hasher, VT: ValidatorSetType>(
         self,
-        validators: &ValidatorMember,
+        validators: &VT,
         sender: &PubKey,
     ) -> Result<Verified<S, ConsensusMessage<S, T>>, Error> {
         // FIXME this feels wrong... it feels like the enum variant should factor into the hash so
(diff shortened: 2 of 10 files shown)


monad-consensus-types/src/validation.rs
```diff
@@ -13,6 +13,8 @@ pub enum Error {
     AuthorNotSender,
     /// There are high qc rounds larger than the TC round
     InvalidTcRound,
+    /// The SignatureCollection doesn't have +2/3 of the stake signed
+    NoSuperMajStake,
```
ariqchowdhury: nit: InsufficientStake?

monad-consensus/src/validation/signing.rs
ariqchowdhury, line 1: we could also consider doing the weight verification at the monad-state level so that we don't pass the whole validator_set into this function. The verification could be split into a signature verification and the consensus level well-formed checks. And then later the BLS stuff is just checked at the sig verification function in monad-state. What do you think?

michael-yxchen: Discussed offline, splitting validator set from leader election. Verify functions takes the validators_set by reference

ariqchowdhury:
nit: the commit message is out of date after rebasing (leader election is now split from validator set)

michael-yxchen:
> nit: the commit message is out of date after rebasing (leader election is now split from validator set)

oops. fixed

monad-consensus/src/validation/signing.rs
```diff
@@ -275,20 +289,27 @@ where
...
 
         t.author_signature
             .verify(msg.as_ref(), &pubkey)
             .map_err(|_| Error::InvalidSignature)?;
+
+        node_ids.push(NodeId(pubkey));
+    }
+
+    if !validators.has_super_majority_votes(node_ids.iter()) {
+        return Err(Error::InsufficientStake);
```
michael-yxchen, line 302: TC stake verification

monad-consensus/src/validation/signing.rs
```diff
@@ -302,13 +323,19 @@ where
...
     qc.signatures
         .verify_signatures(qc_msg.as_ref())
         .map_err(|_| Error::InvalidSignature)?;
 
+    let node_ids = pubkeys.iter().map(|pk| NodeId(*pk)).collect::<Vec<_>>();
+
+    if !validators.has_super_majority_votes(node_ids.iter()) {
+        return Err(Error::InsufficientStake);
+    }
+
```
michael-yxchen, line 338: QC stake verification

monad-consensus-types/src/validation.rs
```diff
@@ -13,6 +13,8 @@ pub enum Error {
     AuthorNotSender,
     /// There are high qc rounds larger than the TC round
     InvalidTcRound,
+    /// The SignatureCollection doesn't have +2/3 of the stake signed
```
ariqchowdhury: nit: maybe, "+2/3" -> "supermajority"

monad-consensus/src/validation/signing.rs
ariqchowdhury, line 1: the type ValidatorMember was just an alias type used in this file, it should be deleted now

monad-consensus/tests/proposal.rs
```diff
@@ -1,16 +1,22 @@
...
 use monad_testutil::signing::{get_key, node_id, MockSignatures, TestSigner};
 use monad_types::*;
-
-fn setup_block(author: NodeId, block_round: u64, qc_round: u64) -> Block<MockSignatures> {
+use monad_validator::validator_set::{ValidatorSet, ValidatorSetType};
+
+fn setup_block(
+    author: NodeId,
+    block_round: u64,
+    qc_round: u64,
```
ariqchowdhury: nit: not your change, but just noticed these should probably be Round types

PR #161, opened 2023-07-14
Create traits for  validatorset, leaderelection and consensus_state

ariqchowdhury:
Monad state now depends on the generic traits

### DIFF
--- a/monad-state/src/lib.rs
+++ b/monad-state/src/lib.rs
@@ -9,11 +9,11 @@ use monad_consensus::{
 };
 use monad_consensus_state::{
     command::{ConsensusCommand, FetchedFullTxs, FetchedTxs},
-    ConsensusState,
+    ConsensusProcess,
 };
 use monad_consensus_types::{
     block::Block, quorum_certificate::QuorumCertificate, signature::SignatureCollection,
-    transaction_validator::TransactionValidator, validation::Sha256Hash, voting::VoteInfo,
+    validation::Sha256Hash, voting::VoteInfo,
 };
 use monad_crypto::{
     secp256k1::{KeyPair, PubKey},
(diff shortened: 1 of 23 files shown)


monad-validator/src/leader_election.rs
```diff
@@ -1,10 +1,7 @@
...
+use monad_types::{NodeId, Round};
 
 // VotingPower is i64
 pub trait LeaderElection {
     fn new() -> Self;
-    fn start_new_epoch(&mut self, voting_powers: Vec<(NodeId, Stake)>);
-    fn increment_view(&mut self, view: Round);
-    fn get_leader(&self) -> &NodeId;
-    fn update_voting_power(&mut self, addr: &NodeId, new_voting_power: Stake) -> bool;
+    fn get_leader(&self, round: Round, validator_list: &[NodeId]) -> NodeId;
```
omegablitz, line 6: Just to check my understanding, will this eventually need to be changed to `&[(NodeId, Stake)]` once we switch to weighted round-robin?

monad-state/src/lib.rs
```diff
@@ -39,44 +40,51 @@ pub mod convert;
...
+    TCT: TransactionCollection,
+    VT: ValidatorSetTrait,
+    LT: LeaderElection,
 {
     message_state: MessageState<MonadMessage<ST, SCT>, VerifiedMonadMessage<ST, SCT>>,
 
-    consensus_state: ConsensusState<ST, SCT, TC>,
-    validator_set: ValidatorSet<LeaderElectionType>,
+    consensus: CT,
+    transaction_type: PhantomData<TCT>,
```
omegablitz: Thinking out loud... can we avoid the PhantomData here? Maybe by removing the TCT generic param, and removing the trait bounds on the struct definition

Generally I think we should avoid PhantomData when possible unless it's unavoidable

monad-validator/src/leader_election.rs
ariqchowdhury: yeah it will need to be changed -- I think the whole weight round robin impl needs to be redone, so when that is happening, we can pick the correct interface here.

PR #164, opened 2023-07-18
Refactor `SignatureCollection` trait

michael-yxchen:
The old `SignatureCollection` trait doesn't fit the need of bls signature collection. When building the `SignatureCollection`, it keeps intermediate aggregation results for quicker search of the invalid signature. Once the `SignatureCollection` is created, i.e. the aggregate signature verified, the intermediate results are discarded to make the collection constant sized.

The previous trait is split into two parts, a SignatureBuilder struct that collects signatures in a vector, and a `SignatureCollection` trait that verifies. The leader appends the signatures received in the builder and constructs the `SignatureCollection` when quorum is reached. The `SigCol::new()` function emits the SigCol if it verifies or accountability information on err.

The `new()` function verifies the SigCol so a leader should never create invalid `SignatureCollection`. `SigCol::verify` doesn't produce any evidence that attributes the verification failure to specific validators even if the info is available (e.g. a `MultiSig`). The author should be slashed in the `Err` case.

The `add_signature` is changed slightly to carry the index of the node id in the validator set. It's used by bls signature collections to lookup the pubkey of the signing entity and verify the signature. The validator set is responsible for creating the ordered list of `(NodeId,BlsPubKey)` for such purpose (e.g. by sorting the list).

The ValidatorSet designed to only contain the staking information. The ValidatorSetProperty type stores any auxiliary data: blspubkey, validator index, etc.

### DIFF
--- a/monad-consensus-types/src/signature.rs
+++ b/monad-consensus-types/src/signature.rs
@@ -1,23 +1,61 @@
-use monad_crypto::{
-    secp256k1::{Error, PubKey},
-    Signature,
-};
-use monad_types::Hash;
+use std::fmt::Debug;
 
-pub trait SignatureCollection: Clone + Default + Send + Sync + std::fmt::Debug + 'static {
-    type SignatureType: Signature;
+use monad_crypto::{bls12_381::BlsPubKey, Signature};
+use monad_types::{Hash, NodeId};
 
-    fn new() -> Self;
+#[derive(Clone, Debug)]
+pub struct SignatureBuilder<SCT: SignatureCollection> {
+    sigs: Vec<(usize, SCT::SignatureType)>,
+}
+
+impl<SCT: SignatureCollection> Default for SignatureBuilder<SCT> {
+    fn default() -> Self {
+        Self::new()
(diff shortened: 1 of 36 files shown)


monad-validator/src/validator_property.rs
michael-yxchen, line 1: ValidatorSetProperty stores the NodeId->BlsPubKey mapping

monad-consensus-types/src/signature.rs
michael-yxchen, line 1: SignatureCollection trait refactor

ariqchowdhury:
> The old `SignatureCollection` trait doesn't fit the need of bls signature collection. When building the `SignatureCollection`, it keeps intermediate aggregation results for quicker search of the invalid signature. Once the `SignatureCollection` is created, i.e. the aggregate signature verified, the intermediate results are discarded to make the collection constant sized.
>
It feels like the implementation details are leaking into the interface -- intermediate aggregate results for quicker search doesn't seem like something a generic SignatureCollection should care about.

PR #165, opened 2023-07-18
Run clippy on everything

abenedito-monad:


omegablitz:
LGTM once clippy errors are fixed

PR #168, opened 2023-07-21
Add rustfmt toml

abenedito-monad:


.rustfmt.toml
```diff
@@ -0,0 +1,2 @@
+group_imports = "StdExternalCrate"
+imports_granularity = "Module"
```
ariqchowdhury: this should be "Crate", right?

PR #169, opened 2023-07-21
Split voting key from signing key

michael-yxchen:
`ConsensusState` holds only one keypair, which is used for both signing messages and votes. In a future PR on `BlsSignatureCollections`, we want to switch between two voting keypairs, based on the `SignatureCollection` we use.

This PR adds a `voting_key` field to `ConsensusState` that's dedicated to signing votes. VoteMessage holds the vote signature over the ledger commit info. The signature in `Verified<_>` is now over the entire VoteMessage.

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -20,7 +20,7 @@ use monad_consensus_types::{
 };
 use monad_crypto::{
     secp256k1::{KeyPair, PubKey},
-    Signature,
+    GenericSignature, Signature,
 };
 use monad_executor::{PeerId, RouterTarget};
 use monad_types::{NodeId, Round};
@@ -48,6 +48,7 @@ pub struct ConsensusState<ST, SCT: SignatureCollection, TV> {
 
     // TODO deprecate
     keypair: KeyPair,
+    vote_keypair: <SCT::SignatureType as GenericSignature>::KeyPairType,
 }
 pub trait ConsensusProcess<ST, SCT>
 where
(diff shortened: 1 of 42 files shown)


monad-crypto/src/lib.rs
michael-yxchen, line 1: Added GenericKeyPair and GenericSignature traits to associate SignatureCollection with the voting keypair

monad-consensus/src/messages/message.rs
michael-yxchen, line 1: VoteMessage contains the Vote (VoteInfo and LedgerCommitInfo) and a signature by the voting keypair

monad-consensus-state/src/lib.rs
```diff
@@ -218,17 +221,27 @@ where
...
 
         if let Some(v) = vote_msg {
+            // sign the Vote with vote keypair
+            let vote_hash = H::hash_object(&v);
+            let sig = <SCT::SignatureType as GenericSignature>::sign(
+                vote_hash.as_ref(),
+                &self.vote_keypair,
+            );
+            let vote_msg = VoteMessage { vote: v, sig };
+
```
michael-yxchen, line 231: Vote signing

monad-consensus-state/src/lib.rs
```diff
@@ -48,6 +48,7 @@ pub struct ConsensusState<ST, SCT: SignatureCollection, TV> {
 
     // TODO deprecate
     keypair: KeyPair,
+    vote_keypair: <SCT::SignatureType as GenericSignature>::KeyPairType,
```
michael-yxchen, line 51: vote keypair in consensus state

PR #170, opened 2023-07-21
Add evidence interface

victorzdm:
Currently, the `EvidencePool` is implemented as an `Executor`. There are now three `EvidenceCommands` in work that the executor can process:
`AddEvidence` is called when the consensus process found out any abnormalities;
`FetchEvidenceList` is to add the evidence into the buffer for block inclusion;
`DeleteEvidence` is for deleting the expired evidence before Round x.

### DIFF
--- a/monad-consensus-state/src/command.rs
+++ b/monad-consensus-state/src/command.rs
@@ -6,6 +6,7 @@ use monad_consensus::{
 };
 use monad_consensus_types::{
     block::Block,
+    evidence::Evidence,
     payload::{FullTransactionList, TransactionList},
     quorum_certificate::QuorumCertificate,
     signature::SignatureCollection,
@@ -39,6 +40,7 @@ pub enum ConsensusCommand<ST, SCT: SignatureCollection> {
     /// Checkpoints periodically can upload/backup the ledger and garbage clean
     /// persisted events if necessary
     CheckpointSave(Checkpoint<SCT>),
+    AddEvidence(Evidence),
     // TODO add command for updating validator_set/round
     // - to handle this command, we need to call message_state.set_round()
 }
@@ -87,4 +89,6 @@ pub struct FetchedFullTxs<ST, SCT> {
     pub author: NodeId,
     pub p: ProposalMessage<ST, SCT>,
     pub txns: Option<FullTransactionList>,
+    pub signature: Vec<u8>,
+    pub sender: NodeId,
 }
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -12,6 +12,7 @@ use monad_consensus::{
 };
 use monad_consensus_types::{
     block::Block,
+    evidence::EvidenceType,
     payload::FullTransactionList,
     quorum_certificate::{QuorumCertificate, Rank},
     signature::SignatureCollection,
@@ -78,6 +79,8 @@ where
         &mut self,
         author: NodeId,
         p: ProposalMessage<ST, SCT>,
+        signature: Vec<u8>,
+        sender: NodeId,
     ) -> Vec<ConsensusCommand<ST, SCT>>;
 
     fn handle_proposal_message_full<H: Hasher, VT: ValidatorSetType, LT: LeaderElection>(
@@ -87,7 +90,7 @@ where
         txns: FullTransactionList,
         validators: &VT,
         election: &LT,
-    ) -> Vec<ConsensusCommand<ST, SCT>>;
+    ) -> (Vec<ConsensusCommand<ST, SCT>>, Option<EvidenceType>);
 
     fn handle_vote_message<H: Hasher, VT: ValidatorSetType, LT: LeaderElection>(
         &mut self,
@@ -156,10 +159,18 @@ where
         &mut self,
         author: NodeId,
         p: ProposalMessage<ST, SCT>,
+        signature: Vec<u8>,
+        sender: NodeId,
     ) -> Vec<ConsensusCommand<ST, SCT>> {
         vec![ConsensusCommand::FetchFullTxs(
             p.block.payload.txns.clone(),
-            Box::new(move |txns| FetchedFullTxs { author, p, txns }),
+            Box::new(move |txns| FetchedFullTxs {
+                author,
+                p,
+                txns,
+                signature,
+                sender,
+            }),
         )]
     }
 
@@ -170,11 +181,11 @@ where
         txns: FullTransactionList,
         validators: &VT,
         election: &LT,
-    ) -> Vec<ConsensusCommand<ST, SCT>> {
+    ) -> (Vec<ConsensusCommand<ST, SCT>>, Option<EvidenceType>) {
         let mut cmds = vec![];
 
         if !self.transaction_validator.validate(&txns) {
-            return cmds;
+            return (cmds, None);
         }
 
         let process_certificate_cmds = self.process_certificate_qc(&p.block.qc);
@@ -201,9 +212,12 @@ where
                 blockid: p.block.get_parent_id(),
             });
         }
-
-        if p.block.round != round || author != leader || p.block.author != leader {
-            return cmds;
+        if p.block.round != round {
+            return (cmds, None);
+        }
+        // Non-leader nodes try to make a proposal
+        if author != leader || p.block.author != leader {
+            return (cmds, Some(EvidenceType::InvalidProposal));
         }
 
         let vote_msg = self
@@ -219,7 +233,7 @@ where
             cmds.push(send_cmd);
         }
 
-        cmds
+        (cmds, None)
     }
 
     fn handle_vote_message<H: Hasher, VT: ValidatorSetType, LT: LeaderElection>(
@@ -563,13 +577,15 @@ mod test {
         // check no vote commands result from receiving the proposal for round 1
 
         let (author, _, verified_message) = p1.destructure();
-        let cmds = state.handle_proposal_message_full::<Sha256Hash, _, _>(
-            author,
-            verified_message,
-            FullTransactionList(Vec::new()),
-            &valset,
-            &election,
-        );
+        let cmds = state
+            .handle_proposal_message_full::<Sha256Hash, _, _>(
+                author,
+                verified_message,
+                FullTransactionList(Vec::new()),
+                &valset,
+                &election,
+            )
+            .0;
         let result = cmds.iter().find(|&c| {
             matches!(
                 c,
@@ -598,13 +614,15 @@ mod test {
             ExecutionArtifacts::zero(),
         );
         let (author, _, verified_message) = p1.destructure();
-        let cmds = state.handle_proposal_message_full::<Sha256Hash, _, _>(
-            author,
-            verified_message,
-            FullTransactionList(Vec::new()),
-            &valset,
-            &election,
-        );
+        let cmds = state
+            .handle_proposal_message_full::<Sha256Hash, _, _>(
+                author,
+                verified_message,
+                FullTransactionList(Vec::new()),
+                &valset,
+                &election,
+            )
+            .0;
         let result = cmds.iter().find(|&c| {
             matches!(
                 c,
@@ -626,13 +644,15 @@ mod test {
             ExecutionArtifacts::zero(),
         );
         let (author, _, verified_message) = p2.destructure();
-        let cmds = state.handle_proposal_message_full::<Sha256Hash, _, _>(
-            author,
-            verified_message,
-            FullTransactionList(Vec::new()),
-            &valset,
-            &election,
-        );
+        let cmds = state
+            .handle_proposal_message_full::<Sha256Hash, _, _>(
+                author,
+                verified_message,
+                FullTransactionList(Vec::new()),
+                &valset,
+                &election,
+            )
+            .0;
         let result = cmds.iter().find(|&c| {
             matches!(
                 c,
@@ -692,13 +712,15 @@ mod test {
                 ExecutionArtifacts::zero(),
             );
             let (author, _, verified_message) = p.clone().destructure();
-            let cmds = state.handle_proposal_message_full::<Sha256Hash, _, _>(
-                author,
-                verified_message,
-                FullTransactionList(Vec::new()),
-                &valset,
-                &election,
-            );
+            let cmds = state
+                .handle_proposal_message_full::<Sha256Hash, _, _>(
+                    author,
+                    verified_message,
+                    FullTransactionList(Vec::new()),
+                    &valset,
+                    &election,
+                )
+                .0;
             let result = cmds.iter().find(|&c| {
                 matches!(
                     c,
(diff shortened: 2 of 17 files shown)


monad-consensus/src/evidence_pool.rs
```diff
@@ -0,0 +1,93 @@
...
+            evidence_store: HashMap::new(),
+        }
+    }
+}
+
+impl EvidencePool {
+    pub fn is_pending(&mut self, id: Hash) -> Option<bool> {
+        self.evidence_store
+            .get(&id)
+            .map(|evidence| evidence.status == EvidenceStatus::Pending)
```
abenedito-monad: Nit cool rust trick, I believe you can do
`self.evidence_store.get(&id)?.status == EvidenceStatus::Pending`
Probably a bit cleaner, and for is_committed below

monad-consensus/src/evidence_pool.rs
```diff
@@ -0,0 +1,93 @@
...
+        }
+        if self.verify(&ev) {
+            self.evidence_list.push_back(ev.clone());
+            self.add_pending_evidence(ev);
+            self.evidence_size += 1;
+        }
+    }
+
+    pub fn size(&self) -> u32 {
+        self.evidence_size
```
abenedito-monad: Hmm why do we have a separate evidence_size ... isn't this just evidence_list/store.len()?

monad-consensus-state/src/lib.rs
```diff
@@ -43,6 +45,7 @@ pub struct ConsensusState<S, SC, T> {
 
     // TODO deprecate
     pub keypair: KeyPair,
+    pub evidence_pool: EvidencePool,
```
michael-yxchen: I think we should split evidence collection from the core consensus state. ConsensusState should only contain the minimum fields to do consensus. 

It should be implemented as an Executor, with appropriate Events and Commands. [Reference](https://www.notion.so/monad-labs/Adding-new-MonadEvents-7fa26d51ce7041558d409e850f52c2d1?pvs=4)

monad-executor/src/state.rs
```diff
@@ -53,6 +57,12 @@ pub enum LedgerCommand<B> {
     LedgerCommit(B),
 }
 
+pub enum EvidenceCommand {
+    AddEvidence(Evidence),
+    FetchEvidenceList(usize),
+    DeleteEvidence(Round),
```
omegablitz: Do we need an evidence deletion mechanism right now? I'm wondering if we could handle aging out evidence in the evidence executor itself.

monad-executor/src/state.rs
```diff
@@ -53,6 +57,12 @@ pub enum LedgerCommand<B> {
     LedgerCommit(B),
 }
 
+pub enum EvidenceCommand {
+    AddEvidence(Evidence),
+    FetchEvidenceList(usize),
```
omegablitz: I think we can omit the `FetchEvidence` mechanism in this first PR, given that it'll be unused

monad-executor/src/executor/parent.rs
```diff
@@ -60,7 +62,7 @@ where
     }
 }
 
-impl<R, T, M, L> ParentExecutor<R, T, M, L> {
```
omegablitz: We should try to keep the generic param names consistent - so if we're using `EV` above, we should stick to that here

monad-executor/src/executor/evpool.rs
```diff
@@ -0,0 +1,80 @@
...
+impl Executor for EvidencePool {
+    type Command = EvidenceCommand;
+    fn exec(&mut self, commands: Vec<Self::Command>) {
+        for command in commands {
+            match command {
+                EvidenceCommand::AddEvidence(ev) => {
+                    let hash = ev.get_hash();
+
+                    if self.evidence_store.contains_key(&hash) {
+                        println!("Evidence already exists");
```
omegablitz: We shouldn't have a print statement in a "real" executor - if you want to record an err/warning, should use the `log` crate

monad-consensus-state/src/lib.rs
```diff
@@ -201,6 +202,13 @@ where
         }
 
         if p.block.round != round || author != leader || p.block.author != leader {
+            let ev = Evidence {
```
omegablitz: Can this branch happen if a node received a proposal out of order?

monad-consensus-types/src/evidence.rs
```diff
@@ -0,0 +1,44 @@
...
+        let mut hasher = sha2::Sha256::new();
+        hasher.update(self.round.as_bytes());
+        hasher.update([self.evidence_type as u8]);
+        for node in self.malicious_nodes.iter() {
+            hasher.update(&node.0.bytes());
+        }
+        Hash(hasher.finalize().into())
+    }
+
+    pub fn size(&self) -> usize {
```
omegablitz: What's this function for?

victorzdm: This function is for `FetchEvidence` where we specify the maximum number of bytes that we are reserving for evidence

monad-executor/src/executor/parent.rs
victorzdm: yes. I used `EV` here because `E` is taken by the `Event`. In that case, I'll change everything to `EV` then.

monad-executor/src/state.rs
victorzdm: Yeah I think that's probably better.

monad-consensus-types/src/evidence.rs
victorzdm: Actually, since we are not including `FetchEvidence` in this PR, it can actually be removed.

monad-state/src/lib.rs
```diff
@@ -241,6 +245,7 @@ where
     SCT: SignatureCollection<SignatureType = ST>,
     VT: ValidatorSetType,
     LT: LeaderElection,
+    VerifiedMonadMessage<ST, SCT>: monad_types::Serializable,
```
victorzdm, line 249: TODO: it's currently a workaround because the `VerifiedMonadMessage` with generic types have not implemented `Serializable`. Maybe rethink about this in the future.

monad-state/src/lib.rs
```diff
@@ -371,9 +401,14 @@ where
...
                         let (author, signature, verified_message) = verified_message.destructure();
                         match verified_message {
-                            ConsensusMessage::Proposal(msg) => self
-                                .consensus
-                                .handle_proposal_message::<HasherType>(author, msg),
+                            ConsensusMessage::Proposal(msg) => {
+                                self.consensus.handle_proposal_message::<HasherType>(
+                                    author,
+                                    msg,
+                                    signature.serialize(),
```
victorzdm, line 406: For proto purposes, I have to serialize the signature before passing in the function

monad-consensus-state/src/command.rs
```diff
@@ -86,4 +88,6 @@ pub struct FetchedFullTxs<ST, SCT> {
     pub author: NodeId,
     pub p: ProposalMessage<ST, SCT>,
     pub txns: Option<FullTransactionList>,
+    pub signature: Vec<u8>,
+    pub sender: PubKey,
```
omegablitz: I think `NodeId` would be more accurate; PubKey specifically represents a secp PubKey.

monad-consensus-state/src/lib.rs
```diff
@@ -86,7 +88,7 @@ where
         txns: FullTransactionList,
         validators: &VT,
         election: &LT,
-    ) -> Vec<ConsensusCommand<ST, SCT>>;
+    ) -> (Vec<ConsensusCommand<ST, SCT>>, bool);
```
omegablitz: The bool here isn't very readable, we could make this a custom enum instead

omegablitz (commented):
Left couple small nits, but more importantly, I think we should have tests that validate that the evidence pool is being updated as expected.

This could be done by creating an invalid proposal, passing it through the state machine, and then checking that the reported evidence is as expected.

monad-consensus-state/src/command.rs
```diff
@@ -86,4 +88,6 @@ pub struct FetchedFullTxs<ST, SCT> {
     pub author: NodeId,
     pub p: ProposalMessage<ST, SCT>,
     pub txns: Option<FullTransactionList>,
+    pub signature: Vec<u8>,
```
omegablitz, line 92: I think these names could be more descriptive. Maybe `pub original_message: (NodeId, Vec<u8>)` or something

monad-state/src/convert/event.rs
```diff
@@ -131,6 +133,13 @@ impl<S: Signature> TryFrom<ProtoConsensusEvent> for ConsensusEvent<S> {
                         ))?
                         .try_into()?,
                     txns: Some(FullTransactionList(fetched_full_txs.full_txs)),
+                    signature: fetched_full_txs.signature,
```
omegablitz: Should probably check if this is empty, because if the field doesn't exist the deserialization will still succeed silently with empty Vec<u8>.

monad-consensus-types/src/evidence.rs
```diff
@@ -0,0 +1,48 @@
...
+#[repr(u8)]
+pub enum EvidenceType {
+    Equivocation = 0,
+    DDoS = 1,
+    Amnesia = 2,
+    InvalidProposal = 3,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum EvidenceStatus {
```
omegablitz: What's the purpose of this enum?

monad-consensus-types/src/evidence.rs
```diff
@@ -0,0 +1,48 @@
...
+    pub round: Round,
+    pub malicious_nodes: HashSet<NodeId>,
+    pub signed_invalid_msg: Vec<u8>,
+}
+
+impl Evidence {
+    pub fn get_hash(&self) -> Hash {
+        let mut hasher = sha2::Sha256::new();
+        hasher.update(self.round.as_bytes());
+        hasher.update([self.evidence_type as u8]);
```
omegablitz, line 32: Is EvidenceStatus intentionally omitted in the hash function?

monad-consensus-types/src/evidence.rs
```diff
@@ -0,0 +1,48 @@
...
+        for node in self.malicious_nodes.iter() {
```
omegablitz: The Iteration order of `malicious_nodes` is not deterministic, which means that this hash function isn't deterministic. One possibility is to change it to a `BTreeSet<NodeId>`.

monad-consensus-types/src/evidence.rs
```diff
@@ -0,0 +1,48 @@
+use std::collections::HashSet;
+
+use monad_types::{Hash, NodeId, Round};
+use sha2::Digest;
+use zerocopy::AsBytes;
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum EvidencePresence {
```
omegablitz, line 6: I'm confused on why EvidencePresence is necessary. Why can't `Option<EvidenceType>` be used instead?

monad-consensus-state/src/lib.rs
```diff
@@ -86,7 +89,7 @@ where
         txns: FullTransactionList,
         validators: &VT,
         election: &LT,
-    ) -> Vec<ConsensusCommand<ST, SCT>>;
```
omegablitz, line 90: Can this return `Option<EvidenceType>` instead?

monad-executor/src/executor/evpool.rs
```diff
@@ -0,0 +1,109 @@
...
+    type Command = EvidenceCommand;
+    fn exec(&mut self, commands: Vec<Self::Command>) {
+        for command in commands {
+            match command {
+                EvidenceCommand::AddEvidence(ev) => {
+                    let hash = ev.get_hash();
+
+                    if self.evidence_store.contains_key(&hash) {
+                        info!("Evidence already exists");
+                        return;
```
omegablitz: Returning here will skip the rest of the commands, is that intended?

monad-executor/src/executor/evpool.rs
```diff
@@ -0,0 +1,109 @@
...
+use crate::{EvidenceCommand, Executor};
+
+#[derive(Default)]
+pub struct EvidencePool {
+    evidence_list: VecDeque<Evidence>,
+    evidence_store: HashMap<Hash, Evidence>,
+}
+
+impl EvidencePool {
+    fn verify(&mut self, ev: &Evidence) -> bool {
```
omegablitz: Verification here seems kinda pointless, because the state machine has presumably already verified the evidence prior to reaching this point.

monad-executor/src/executor/evpool.rs
```diff
@@ -0,0 +1,109 @@
+use std::collections::{HashMap, VecDeque};
+
+use log::info;
+use monad_consensus_types::evidence::{Evidence, EvidenceStatus};
+use monad_types::{Hash, Round};
+
+use crate::{EvidenceCommand, Executor};
+
+#[derive(Default)]
+pub struct EvidencePool {
```
omegablitz: Maybe rename this to MockEvidencePool? Because it doesn't actually write to disk

monad-executor/src/executor/evpool.rs
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    fn add_pending_evidence(&mut self, ev: Evidence) {
+        assert!(ev.status == EvidenceStatus::Pending);
+        let hash = ev.get_hash();
+        self.evidence_store.insert(hash, ev.clone());
+        self.evidence_list.push_back(ev);
+    }
+
+    fn delete_expired_evidence(&mut self, round: Round) {
```
omegablitz: I think we can just delete this function for now

monad-state/src/lib.rs
```diff
@@ -343,18 +348,40 @@ where
...
+                                    &self.leader_election,
+                                );
+                            if ev_presence == EvidencePresence::Found {
+                                let round = self.consensus.get_current_round();
+                                let signed_invalid_msg = Unverified::new(
+                                    ConsensusMessage::Proposal(fetched_txs.p.clone()),
+                                    ST::deserialize(&fetched_txs.signature).unwrap(),
+                                )
+                                .verify::<HasherType, _>(&self.validator_set, &fetched_txs.sender.0)
+                                .unwrap();
```
omegablitz, line 372: Can we just clone the verified message above so that we avoid needing to do the `verify().unwrap()`? It'll be an extra clone in the good case but that's probably fine for now.

monad-state/src/lib.rs
```diff
@@ -343,18 +348,40 @@ where
...
+                                    round,
+                                    malicious_nodes: HashSet::from([
                                         fetched_txs.author,
-                                        fetched_txs.p,
-                                        txns,
-                                        &self.validator_set,
-                                        &self.leader_election,
-                                    ),
-                            );
+                                        fetched_txs.p.block.author,
```
omegablitz: Are `fetched_txs.author` and `fetched_txs.p.block.author`. If so could we change evidence to just have a single `NodeId` instead of being a set?

PR #177, opened 2023-07-27
Implement monad-mempool executor

abenedito-monad:


monad-node/src/main.rs
```diff
@@ -276,7 +276,7 @@ async fn run(
     let mut executor = ParentExecutor {
         router,
         timer: TokioTimer::default(),
-        mempool: MockMempool::default(),
```
ariqchowdhury: I think we might want the choice of mempool to be a command line arg.

abenedito-monad: Sure, I can add an arg for that.

abenedito-monad: Decided offline to keep mempool as MockMempool.

PR #180, opened 2023-07-28
Fix `test_proposal_invalid_qc` test

michael-yxchen:
The previous test was a misnomer. The proposal has valid qc but the message was not signed by the sender.

This PR renames the test to
`test_proposal_author_not_sender` and adds a new test where the qc doesn't have enough stake. The MockSignature constructor and verify functions are corrected to produce the expected behavior

### DIFF
--- a/monad-consensus/tests/proposal.rs
+++ b/monad-consensus/tests/proposal.rs
@@ -98,14 +98,16 @@ fn test_proposal_missing_tc() {
 }
 
 #[test]
-fn test_proposal_invalid_qc() {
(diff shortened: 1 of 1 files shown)


michael-yxchen:
> So after these changes, it seems like we don't have a test that covers the InvalidAuthor error.
> InvalidAuthor is the case where the UnverifiedMessage is signed by a node not in the ValidatorSet. Can we add a test for that (or open an issue for now so we don't forget)

Test added

PR #182, opened 2023-08-01
Add ExecutionArtifacts to the consensus Block

ariqchowdhury:
The results of executing a block will be retrieved via executor at some point by consensus and will need to be included in block proposals. These are the execution artifacts.

### DIFF
--- a/monad-consensus-types/src/block.rs
+++ b/monad-consensus-types/src/block.rs
@@ -2,35 +2,17 @@ use monad_types::{BlockId, Hash as HashType, NodeId, Round};
 use zerocopy::AsBytes;
 
 use crate::{
+    payload::Payload,
     quorum_certificate::QuorumCertificate,
     signature::SignatureCollection,
     validation::{Hashable, Hasher},
 };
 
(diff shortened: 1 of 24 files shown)


monad-consensus-types/src/block.rs
```diff
@@ -50,7 +32,13 @@ impl<T: SignatureCollection> Hashable for Block<T> {
     fn hash<H: Hasher>(&self, state: &mut H) {
         state.update(self.author.0.bytes());
         state.update(self.round.as_bytes());
-        state.update(self.payload.0.as_bytes());
+        state.update(self.payload.txns.0.as_bytes());
```
omegablitz: Should we implement `Hashable` directly on `Payload` and `ExecutionArtifacts`? Just so that the hash implementations live closer to the members on the struct - I'm worried that down the line we'll add something to ExecutionArtifacts but then forget to add it to the `Hashable` implementation.

In the future, once we can derive Hashable, I won't be as worried.

ariqchowdhury: good point, made the change.

PR #187, opened 2023-08-02
RFC: SignatureCollection trait

michael-yxchen:


monad-consensus-types/src/signature.rs
```diff
@@ -35,24 +35,34 @@ impl<SCT: SignatureCollection> IntoIterator for SignatureBuilder<SCT> {
...
     /// the new() function verifies:
     ///   1. nodeId idx is in range
     ///   2. no conflicting signature
     ///   3. the signature collection built is valid
-    fn new(
-        sigs: SignatureBuilder<Self>,
-        validator_list: &[(NodeId, BlsPubKey)],
+    fn new<'a>(
+        sigs: Vec<(NodeId, Self::SignatureType)>,
+        validator_list: impl Iterator<
```
omegablitz: ~~Did you mean `&'a(NodeId, <Self::SignatureType as GenericSignature>::PubKeyType)`? Seems unlikely that we'll have `(&NodeId, &PubKeyType)`.~~ Nvm forgot about the map iterators

Regardless, I think it'd be more readable if we could create a new type called something like `ValidatorMapping` or `ValidatorKeys`, and use that in both these function signatures and monad-state.

monad-crypto/src/lib.rs
```diff
@@ -22,6 +22,30 @@ pub trait Signature: Copy + Clone + Eq + Hash + Send + Sync + std::fmt::Debug +
...
+
+    fn sign(&self, msg: &[u8]) -> Self::SignatureType;
+    fn from_bytes(secret: impl AsMut<[u8]>) -> Result<Self, Self::Error>;
+    fn pubkey(&self) -> <Self::SignatureType as GenericSignature>::PubKeyType;
+}
+
+pub trait GenericSignature:
+    Copy + Clone + Eq + Hash + Send + Sync + std::fmt::Debug + 'static
+{
+    type KeyPairType: GenericKeyPair; // do not restrict the associated type bc multiple signatures can be created from the same keypair
```
omegablitz: It's a little weird to me how GenericKeyPair and GenericSignature are cyclic. I think it'd be nice if it was acyclic...

Brainstorming, we could have something like `KeyPair` at the top level which has associated `PubKey` and `Signature` types, and then `PubKey` could just have an associated `Signature` type.

With that model, the sign function would exist in KeyPair, and the verify function would exist in PubKey.

What do you guys think @michael-yxchen @ariqchowdhury?

PR #188, opened 2023-08-04
Cleanup mempool controller and executor

abenedito-monad:


monad-mempool-controller/src/lib.rs
```diff
@@ -314,27 +280,68 @@ impl Controller {
...
+
+        encode_list::<Vec<u8>, _>(
+            proposal
+                .into_iter()
+                .map(|b| b.0.into())
+                .collect::<Vec<Vec<u8>>>()
+                .as_slice(),
+        )
+        .to_vec()
+    }
```
abenedito-monad, line 315: RLP encoding txs

PR #190, opened 2023-08-04
Added cron job in github action

Yimaha:
Number of round is controlled by a constant env, while the seed is randomly selected

### DIFF
--- a/monad-consensus/tests/proposal.rs
+++ b/monad-consensus/tests/proposal.rs
@@ -98,14 +98,16 @@ fn test_proposal_missing_tc() {
 }
 
 #[test]
-fn test_proposal_invalid_qc() {
+fn test_proposal_author_not_sender() {
     let mut vlist = Vec::new();
 
-    let keypair = get_key(6);
+    let author_keypair = get_key(6);
+    let sender_keypair = get_key(7);
 
-    vlist.push((NodeId(keypair.pubkey()), Stake(0)));
+    vlist.push((NodeId(author_keypair.pubkey()), Stake(0)));
+    vlist.push((NodeId(sender_keypair.pubkey()), Stake(0)));
 
-    let author = node_id();
+    let author = NodeId(author_keypair.pubkey());
     let proposal = ProposalMessage {
         block: setup_block(
             author,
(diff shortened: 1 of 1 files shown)


monad-state/tests/rand_lat.rs
```diff
@@ -1,25 +1,55 @@
...
+                for _ in 0..round {
+                    seed = generator.gen();
+                    println!("seed is set to be {}", seed);
+                    nodes_with_random_latency(seed);
+                }
+            }
+            Err(_e) => {
+                println!("RANDOM_TEST_SEED is not set, default it to 0");
+                nodes_with_random_latency(0);
+            }
```
ariqchowdhury: Because this is a separate test intended for CI, I think this error should fail loudly. Otherwise a problem in CI will lead to the same test being run over and over without us noticing.

.github/workflows/randomized.yml
```diff
@@ -0,0 +1,22 @@
+name: Randomized Rust Test
+
+on:
+  schedule:
+    - cron: '0 4 * * *' # run at 4 AM UTC
+env:
+  CARGO_TERM_COLOR: always
+  ALL_MESSAGES_DELAYED_ROUND: 40
```
Yimaha: I think 1 thing worth mentioning is if we have a fixed amount of rounds we want to run for each type of test. say each **round** we run all_message_delay 40 times always and Nodes_with_random_latency 40 times but in env we specify how many **round** in total we want to run, thus converting these env constant to be class constant (not sure how relevant is it though)

monad-state/tests/rand_lat.rs
```diff
@@ -1,25 +1,55 @@
...
+    fn nodes_with_random_latency_cron() {
+        let round = match env::var("NODES_WITH_RANDOM_LATENCY_ROUND") {
+            Ok(v) => v.parse().unwrap(),
+            Err(_e) => 1, // default to 1 if not found
+        };
+
+        match env::var("RANDOM_TEST_SEED") {
+            Ok(v) => {
+                let mut seed = v.parse().unwrap();
+                let mut generator: StdRng = StdRng::seed_from_u64(seed);
```
ariqchowdhury: nti: `let mut generator = StdRng::seed...` explicit type isn't needed

monad-state/tests/rand_lat.rs
```diff
@@ -1,25 +1,55 @@
...
+            Duration::from_millis(250),
+            RandLatencyTransformer::new(seed, 330),
+        );
+    }
+
+    #[test]
+    fn nodes_with_random_latency_cron() {
+        let round = match env::var("NODES_WITH_RANDOM_LATENCY_ROUND") {
+            Ok(v) => v.parse().unwrap(),
+            Err(_e) => 1, // default to 1 if not found
```
ariqchowdhury: Same comment as below, let's fail instead of silently continuing.

.github/workflows/randomized.yml
```diff
@@ -0,0 +1,22 @@
...
+  NODES_WITH_RANDOM_LATENCY_ROUND: 40
```
ariqchowdhury: maybe we should start with a lower number of iterations first, see how long each iteration of a test takes and costs I guess.

monad-state/tests/order.rs
```diff
@@ -1,41 +1,69 @@
...
-    run_one_delayed_node(
-        vec![
-            LatencyTransformer(Duration::from_millis(1)).boxed(),
-            PartitionThenReplayTransformer::new(filter_peers, 200, direction).boxed(),
-        ],
-        pubkeys,
-        state_configs,
-    );
+#[cfg(test)]
+mod test {
```
ariqchowdhury: is this necessary?

Yimaha: clippy failed if I have a function (including test) under an existing test, I checked the approach taken by protobuf.rs. Though I have to be honest I don't know if there is any other way to conclude mutliple test under the same file and not mess with clippy?

.github/workflows/randomized.yml
Yimaha: I think it is quiet fast based on the experiment I did, but I will lower it to something like 10?

ariqchowdhury: yeah, it might be fast locally, I'm just not sure how that translates to the github runner. 10 sounds fine to start with and we can bump it up later

Yimaha: I did tested on github action, let me pull up the actual log for that time (the most up to date result):
https://github.com/monad-crypto/monad-bft/actions/runs/5767137856/job/15636327892
if we look at run_all_message_delay, its kinda slow, because it had to compile everything, during the second round where you don't have to compile the test ran for 15 seconds. So roughly 40 rounds = 15 seconds? anyway this is just a really sketched bench mark

monad-state/tests/order.rs
ariqchowdhury: I see...I think its related to the test_case macro...it can also be fixed by moving the new test above so that the test macro is last.

.github/workflows/randomized.yml
```diff
@@ -5,8 +5,8 @@ on:
     - cron: '0 4 * * *' # run at 4 AM UTC
 env:
   CARGO_TERM_COLOR: always
-  ALL_MESSAGES_DELAYED_ROUND: 40
-  NODES_WITH_RANDOM_LATENCY_ROUND: 40
+  ALL_MESSAGES_DELAYED_ROUND: 100
+  NODES_WITH_RANDOM_LATENCY_ROUND: 100
```
Yimaha, line 9: follow up: clarified in slack round cost very little

PR #199, opened 2023-08-11
Block Sync Phase 1

Yimaha:
still work in progress, update as commit goes

### DIFF
--- a/monad-block-sync/src/lib.rs
+++ b/monad-block-sync/src/lib.rs
@@ -1,0 +1,67 @@
+use monad_consensus::messages::{
+    consensus_message::ConsensusMessage, message::RequestBlockSyncMessage,
+};
+use monad_consensus_state::command::ConsensusCommand;
+use monad_consensus_types::{
+    certificate_signature::CertificateSignatureRecoverable,
+    signature_collection::SignatureCollection,
+};
+use monad_executor::{PeerId, RouterTarget};
+use monad_types::{BlockId, NodeId};
+use monad_validator::validator_set::ValidatorSetType;
+
+pub struct BlockSyncState {}
+
+pub trait BlockSyncProcess<ST, SCT, VT>
+where
+    ST: CertificateSignatureRecoverable,
+    SCT: SignatureCollection,
+    VT: ValidatorSetType,
+{
+    fn new() -> Self;
+
+    fn request_block_sync(
+        &mut self,
+        blockid: BlockId,
+        validators: &VT,
+    ) -> (RouterTarget, ConsensusMessage<ST, SCT>);
+
+    fn handle_request_block_sync_message(
+        &mut self,
+        author: NodeId,
+        signature: SCT::SignatureType,
+        s: RequestBlockSyncMessage,
+        validators: &VT,
+    ) -> Vec<ConsensusCommand<ST, SCT>>;
+}
+
+impl<ST, SCT, VT> BlockSyncProcess<ST, SCT, VT> for BlockSyncState
+where
+    ST: CertificateSignatureRecoverable,
+    SCT: SignatureCollection,
+    VT: ValidatorSetType,
+{
+    fn new() -> Self {
+        BlockSyncState {}
+    }
+    fn request_block_sync(
+        &mut self,
+        blockid: BlockId,
+        validators: &VT,
+    ) -> (RouterTarget, ConsensusMessage<ST, SCT>) {
+        let target = RouterTarget::PointToPoint(PeerId(validators.get_list()[0].0));
+        let message =
+            ConsensusMessage::RequestBlockSync(RequestBlockSyncMessage { block_id: blockid });
+        (target, message)
+    }
+
+    fn handle_request_block_sync_message(
+        &mut self,
+        author: NodeId,
+        signature: SCT::SignatureType,
+        s: RequestBlockSyncMessage,
+        validators: &VT,
+    ) -> Vec<ConsensusCommand<ST, SCT>> {
+        vec![]
+    }
+}
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -4,7 +4,7 @@ use monad_blocktree::blocktree::BlockTree;
 use monad_consensus::{
     messages::{
         consensus_message::ConsensusMessage,
-        message::{ProposalMessage, TimeoutMessage, VoteMessage},
+        message::{BlockSyncMessage, ProposalMessage, TimeoutMessage, VoteMessage},
     },
     pacemaker::{Pacemaker, PacemakerCommand, PacemakerTimerExpire},
     validation::safety::Safety,
@@ -105,6 +105,11 @@ where
         tm: TimeoutMessage<ST, SCT>,
         validators: &VT,
         election: &LT,
+    ) -> Vec<ConsensusCommand<ST, SCT>>;
+
+    fn handle_block_sync_message(
+        &mut self,
+        b: BlockSyncMessage<SCT>,
     ) -> Vec<ConsensusCommand<ST, SCT>>;
 
     fn get_current_round(&self) -> Round;
@@ -306,6 +311,13 @@ where
         }
 
         cmds
+    }
+
+    fn handle_block_sync_message(
+        &mut self,
+        b: BlockSyncMessage<SCT>,
+    ) -> Vec<ConsensusCommand<ST, SCT>> {
+        Vec::new()
     }
 
     fn get_pubkey(&self) -> PubKey {
@@ -1333,7 +1345,7 @@ mod test {
         assert_eq!(second_state.high_qc.info.vote, votes[0].vote_info);
 
         // use the correct proposal gen to make next proposal and send it to second_state
-        // this should cause it to emit the RequestSync command because the the QC parent
+        // this should cause it to emit the RequestBlockSync Message because the the QC parent
         // points to a different proposal (second_state has the malicious proposal in its blocktree)
         let cp2 = correct_proposal_gen.next_proposal(
             &keys,
@@ -1355,8 +1367,8 @@ mod test {
             matches!(
                 c,
                 ConsensusCommand::RequestSync {
-                    blockid: BlockId(_),
-                }
+                    blockid: BlockId(_)
+                },
             )
         });
         assert!(res.is_some());
@@ -1372,9 +1384,10 @@ mod test {
         let res = cmds1.into_iter().find(|c| {
             matches!(
                 c,
-                ConsensusCommand::RequestSync {
-                    blockid: BlockId(_),
-                }
+                ConsensusCommand::Publish {
+                    target: RouterTarget::PointToPoint(_),
+                    message: ConsensusMessage::RequestBlockSync(_),
+                },
             )
         });
         assert!(res.is_none());
--- a/monad-consensus/src/messages/message.rs
+++ b/monad-consensus/src/messages/message.rs
@@ -7,6 +7,8 @@ use monad_consensus_types::{
     validation::{Hashable, Hasher},
     voting::VoteInfo,
 };
+use monad_types::BlockId;
+use zerocopy::AsBytes;
 
 #[derive(Copy, Clone, PartialEq, Eq)]
 pub struct VoteMessage {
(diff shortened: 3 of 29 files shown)


monad-consensus-state/src/lib.rs
```diff
@@ -4,7 +4,7 @@ use monad_blocktree::blocktree::BlockTree;
 use monad_consensus::{
     messages::{
         consensus_message::ConsensusMessage,
-        message::{ProposalMessage, TimeoutMessage, VoteMessage},
+        message::{ProposalMessage, RequestSyncMessage, TimeoutMessage, VoteMessage},
```
ariqchowdhury: RequestSyncMessage --> RequestBlockSyncMessage

monad-consensus-state/src/lib.rs
```diff
@@ -197,9 +199,18 @@ where
             .expect("Failed to add block to blocktree");
 
         if !self.pending_block_tree.has_parent(&p.block) {
-            cmds.push(ConsensusCommand::RequestSync {
-                blockid: p.block.get_parent_id(),
-            });
+            match validators.rand_member() {
```
ariqchowdhury: discussed offline, can't be random and should try and move logic into monad-state

monad-validator/src/validator_set.rs
```diff
@@ -36,7 +38,9 @@ pub trait ValidatorSetType {
         Self: Sized;
     fn get_members(&self) -> &HashMap<NodeId, Stake>;
     fn get_list(&self) -> &Vec<NodeId>;
+    fn get_len(&self) -> usize;
     fn is_member(&self, addr: &NodeId) -> bool;
+    fn rand_member(&self) -> Option<&NodeId>;
```
ariqchowdhury: as discussed, just pick a specific node for now

monad-consensus-state/src/lib.rs
Yimaha: follow up on this, which will be updated in the design doc appendix

The design goal of Monad_consensus_state & monad_state is to create a deterministic statemachine, thus introducing threaded random is not ideal from the design perspective.

As for the proposal we plan to just pick a default validator, but once in phase 2 we plan on introducing round-robin for now

Yimaha:
Phase 1 Update

Introduce the Types needed for proposal sync/recovery, the exact design detail is noted in the following notion docs

https://www.notion.so/monad-labs/Proposal-Sync-Recovery-Ver-0-9af591bd67e740329b6ddb6f2d7f899a?pvs=4

During Phase 1, our goal is to introduce the correct type and work flow needed to trigger missing proposal and follow by a firing of correct event/command for recovering the block from neighbour. Note that most function is not implemented, its just mostly types/interface so people are on the same page about the progress and goal.

oh and I will fix linters later :3

monad-consensus/src/convert/consensus_message.rs
```diff
@@ -0,0 +1,68 @@
+use std::fmt::Debug;
```
Yimaha: wait I don't remember this file being new, let me double check

Yimaha: I think this might be a bug with github, but I only modified this file instead of adding it? though the last edit is couple month ago so that might be the trigger?

ariqchowdhury: the contents of this file look like `monad-consensus/src/messages/consensus_message.rs` -- there isn't a consensus_message.rs file in the convert folder

Yimaha: let me correct that

monad-consensus-state/src/lib.rs
```diff
@@ -107,6 +107,10 @@ where
         election: &LT,
     ) -> Vec<ConsensusCommand<ST, SCT>>;
 
+    fn handle_request_block_sync_message(&mut self) -> Vec<ConsensusCommand<ST, SCT>>;
```
ariqchowdhury: I don't think that the consensus state machine should be involved with handling requests for a block from external nodes...
Can we have a separate BlockRequestHandler instantiated at MonadState that deals with fetching the block from the ledger and creating the response?

monad-consensus-state/src/lib.rs
```diff
@@ -107,6 +107,10 @@ where
...
+
+    fn handle_block_sync_message(&mut self) -> Vec<ConsensusCommand<ST, SCT>>;
```
ariqchowdhury: you should put in the input parameters to make it clear what the interface is intending.

monad-consensus-state/src/lib.rs
```diff
@@ -197,9 +201,19 @@ where
             .expect("Failed to add block to blocktree");
 
         if !self.pending_block_tree.has_parent(&p.block) {
-            cmds.push(ConsensusCommand::RequestSync {
-                blockid: p.block.get_parent_id(),
-            });
+            match validators.get_len() {
+                _ => {
+                    let sync_cmd = ConsensusCommand::Publish {
+                        // TODO: Introduce Round robin for picking validator in phase 2
```
ariqchowdhury: can you remove comments referring to 'phase 2', that notation doesn't refer to anything general in the codebase, it's just specific to your project

monad-consensus-state/src/lib.rs
```diff
@@ -197,9 +201,19 @@ where
...
+                        target: RouterTarget::PointToPoint(PeerId(validators.get_list()[0].0)),
+                        message: ConsensusMessage::RequestBlockSync(RequestBlockSyncMessage {
+                            block_id: p.block.get_parent_id(),
```
ariqchowdhury: is it possible to move this logic outside of consensus_state, into monad_state or something like the BlockRequestHandler I mentioned above? 
I think consensus_state should just output the minimal command, `Request(BlockID)` and something else should be responsible for determining what validator is chosen to fulfill the request.

monad-executor/src/state.rs
```diff
@@ -52,8 +53,11 @@ pub enum MempoolCommand<E> {
     FetchFullReset,
 }
 
-pub enum LedgerCommand<B> {
+pub enum LedgerCommand<B, E> {
     LedgerCommit(B),
+    // Q: I have a feeling block ID should be something like a unique identifier that is also generic,
+    // but this might work for now?
```
ariqchowdhury: delete

monad-consensus-state/src/lib.rs
Yimaha: that's a good point, I think handle_block_sync_message should still be invovled because that event is involved with block tree update, but handle_request_block_sync_message doesn't need to be involved with consensus too, will update

monad-consensus-state/src/lib.rs
Yimaha: agree, I will format this to simply emit a command that will be captured by monad state

monad-consensus/src/messages/message.rs
```diff
@@ -53,3 +55,25 @@ impl<S: Signature, T: SignatureCollection> Hashable for ProposalMessage<S, T> {
...
 }
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct RequestBlockSyncMessage {
+    pub block_id: BlockId,
+}
+
+impl Hashable for RequestBlockSyncMessage {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        state.update(self.block_id.0.as_bytes());
```
abenedito-monad, line 66: Hmm is there a reason we're not deriving Hash for RequestBlockSyncMessage? BlockId implements Hash, so you should be able to derive here ...

monad-executor/src/executor/ledger.rs
```diff
@@ -1,32 +1,93 @@
...
                 LedgerCommand::LedgerCommit(block) => {
                     self.blockchain.push(block);
                 }
+                LedgerCommand::LedgerFetch(block_id, cb) => {
+                    self.ledger_fetch_cb = Some(cb);
+                    wake = true;
+                }
             }
         }
+        match self.waker.take() {
```
abenedito-monad:
```
if let Some(waker) = self.waker.take() {
  ...
}
```

monad-consensus/src/messages/message.rs
Yimaha: hmm this is a good point, you mean just replace with 
some form of #[derive(Hash)] ?

monad-state/src/convert/event.rs
```diff
@@ -50,6 +51,14 @@ impl<S: Signature> From<&ConsensusEvent<S>> for ProtoConsensusEvent {
                         .unwrap_or_default(),
                 })
             }
+            TypeConsensusEvent::FetchedBlock(fetched_block) => {
+                proto_consensus_event::Event::FetchedBlock(ProtoFetchedBlock {
+                    block: match &fetched_block.block {
+                        Some(b) => Some((b).into()),
+                        None => None,
+                    },
```
abenedito-monad: `fetched_block.block.map(|b| b.into())`

monad-state/src/lib.rs
```diff
@@ -509,6 +527,7 @@ impl<S: Debug, SC: Debug> Debug for ConsensusEvent<S, SC> {
                 .field("proposal", &p.p)
                 .field("txns", &p.txns)
                 .finish(),
+            ConsensusEvent::FetchedBlock(b) => f.debug_struct("FetchedBlock").finish(),
```
abenedito-monad: No block field :(

monad-validator/src/validator_set.rs
```diff
@@ -126,6 +127,10 @@ impl ValidatorSetType for ValidatorSet {
         &self.validator_list
     }
 
+    fn get_len(&self) -> usize {
```
abenedito-monad: super nit but in rust the convention is to call the "size" function `len`

monad-consensus/src/messages/message.rs
abenedito-monad: Yep, that'll automatically generate the code you wrote out :)

abenedito-monad: Generally, you should always automatically derive because that avoids issues like in the future you add a field but forget to update the hash function definition. Automatically deriving the attribute would catch this. There are certain scenarios where you need to manually derive hash, but in those you do want the finer hash definition control.

Yimaha: we had a discussion on slack, and it seems like hashable not equal hash so this is not avoidable :3

monad-consensus-state/src/lib.rs
```diff
@@ -107,6 +109,14 @@ where
         election: &LT,
     ) -> Vec<ConsensusCommand<ST, SCT>>;
 
+    fn handle_block_sync_message<VT: ValidatorSetType>(
+        &mut self,
+        author: NodeId,
+        signature: ST,
+        b: BlockSyncMessage<SCT>,
+        validators: &VT,
```
ariqchowdhury: I don't think this function will require author, signature or validators.

monad-state/src/lib.rs
```diff
@@ -366,6 +374,11 @@ where
 
                         cmds
                     }
+                    ConsensusEvent::FetchedBlock(fetched_b) => {
+                        let mut cmds = vec![ConsensusCommand::FetchFullTxsReset];
```
ariqchowdhury: just return empty cmds if this isn't implemented yet

monad-block-sync/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+    fn request_block_sync(
+        &mut self,
+        blockid: BlockId,
+        validators: &VT,
+    ) -> (RouterTarget, ConsensusMessage<ST, SCT>);
+    fn handle_request_block_sync_message(
+        &mut self,
+        author: NodeId,
+        signature: SCT::SignatureType,
+        s: RequestBlockSyncMessage,
```
ariqchowdhury: nit: new line between the functions

monad-state/src/lib.rs
```diff
@@ -410,20 +439,22 @@ where
                         Vec::new()
                     }
                 };
-
+                let mut prepare_router_message =
```
ariqchowdhury: nit: newline?

monad-state/src/lib.rs
```diff
@@ -410,20 +439,22 @@ where
...
+                    |target: RouterTarget, message: ConsensusMessage<ST, SCT>| {
+                        let message = VerifiedMonadMessage(
+                            message.sign::<HasherType>(self.consensus.get_keypair()),
+                        );
+                        let publish_action = self.message_state.send(target, message);
+                        return Command::RouterCommand(RouterCommand::Publish {
+                            target: publish_action.target,
+                            message: publish_action.message,
+                        });
+                    };
```
ariqchowdhury, line 457: nit: newline?

monad-state/src/lib.rs
```diff
@@ -458,14 +489,29 @@ where
                         ConsensusCommand::FetchFullTxsReset => {
                             cmds.push(Command::MempoolCommand(MempoolCommand::FetchFullReset))
                         }
-                        ConsensusCommand::RequestSync { blockid: _ } => {
-                            // TODO: respond with the Proposal matching the blockID.
-                            //       right now, all 'missed' proposals are actually in-flight
-                            //       so we don't need to implement this yet for things to work
+
+                        ConsensusCommand::RequestSync { blockid } => {
+                            if self.validator_set.size() > 0 {
```
ariqchowdhury: I think this check is something that should be established elsewhere...Even in a network with just one node, that node should be in the validator set, so we would expect this to always be true. 
I think that the correct place to establish that the validator_set is non-zero is when the Epoch load (which loads the validator set) happens. Can be removed from here.

monad-validator/src/validator_set.rs
```diff
@@ -36,6 +36,7 @@ pub trait ValidatorSetType {
         Self: Sized;
     fn get_members(&self) -> &HashMap<NodeId, Stake>;
     fn get_list(&self) -> &Vec<NodeId>;
+    fn size(&self) -> usize;
```
ariqchowdhury: can remove this now.

monad-state/src/lib.rs
Yimaha: I see, will remove

PR #201, opened 2023-08-12
Remove unused deps and add lint rule

abenedito-monad:


.github/workflows/rust.yml
```diff
@@ -39,6 +39,8 @@ jobs:
           -A clippy::type_complexity
           -A clippy::int_plus_one
           -A clippy::uninlined-format-args
+      - name: Lint unused deps
+        run: cargo clippy --all-features -- -D unused_crate_dependencies
```
omegablitz, line 43: Can this lint rule be combined with the previous `Lint` step?

abenedito-monad: Unfortunately no, the unused_crate_dependencies lint is a little finicky at the moment. If it ever gets stabilized then we could combine them, but we need it to be a separate rule right now for some hacky behaviour when it comes to test and bench files which conflict with --all-targets in the previous lint command.

PR #202, opened 2023-08-14
`SignatureCollection` test cases

### LINKED ISSUE #171: `SignatureCollection` test cases
Will fix after the `SignatureCollection` refactor

michael-yxchen:
Add general `SignatureCollection` test cases and `MultiSig<S>` specific tests for invalid signatures.

Modified SignatureCollection trait to support general error reporting

Closes #171

### DIFF
--- a/monad-consensus-types/src/multi_sig.rs
+++ b/monad-consensus-types/src/multi_sig.rs
@@ -1,10 +1,13 @@
-use std::collections::{HashMap, HashSet};
+use std::collections::HashMap;
 
+use log::{error, warn};
 use monad_types::{Hash, NodeId};
 
 use crate::{
-    certificate_signature::{CertificateSignature, CertificateSignatureRecoverable},
-    signature_collection::SignatureCollection,
+    certificate_signature::CertificateSignatureRecoverable,
+    signature_collection::{
+        SignatureCollection, SignatureCollectionError, SignatureCollectionKeyPairType,
+    },
     validation::{Hashable, Hasher},
     voting::ValidatorMapping,
 };
@@ -20,43 +23,6 @@ impl<S: CertificateSignatureRecoverable> Default for MultiSig<S> {
     }
 }
 
-#[derive(Debug)]
-pub enum MultiSigError<S> {
-    NodeIdNotInMapping((NodeId, S)),
-    ConflictingSignatures((NodeId, S, S)),
-    PubKeyMismatch((NodeId, S)),
-    // verifications error
-    FailedToRecoverPubKey(S),
-    InvalidSignature(S),
-}
-
-impl<S: CertificateSignatureRecoverable> std::fmt::Display for MultiSigError<S> {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        match *self {
-            MultiSigError::NodeIdNotInMapping((node_id, sig)) => {
-                write!(f, "NodeId {node_id:?} not in validator mapping sig {sig:?}")
-            }
-            MultiSigError::ConflictingSignatures((node_id, s1, s2)) => {
-                write!(
-                    f,
-                    "Conflicting signatures from {node_id:?}\ns1: {s1:?}\ns2: {s2:?}"
-                )
-            }
-            MultiSigError::PubKeyMismatch((node_id, sig)) => {
-                write!(f, "PubKey mismatch NodeId {node_id:?} Sig {sig:?}")
-            }
-            MultiSigError::InvalidSignature(sig) => {
-                write!(f, "Invalid signature ({sig:?})")
-            }
-            MultiSigError::FailedToRecoverPubKey(sig) => {
-                write!(f, "Failed to recover pubkey sig ({sig:?})")
-            }
-        }
-    }
-}
-
-impl<S: CertificateSignatureRecoverable> std::error::Error for MultiSigError<S> {}
-
 impl<S: CertificateSignatureRecoverable> Hashable for MultiSig<S> {
     fn hash<H: Hasher>(&self, state: &mut H) {
         for s in self.sigs.iter() {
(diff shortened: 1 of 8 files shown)


michael-yxchen:
It changes the error reporting in `SignatureCollection` a little - instead of having different SCT defining their own error types, we have a general `SignatureCollectionError` struct. It makes it easier to test common behaviors across different implementations, but the errors are less expressive. What do you guys think? @ariqchowdhury @omegablitz

monad-consensus-types/tests/signature_collection.rs
```diff
@@ -0,0 +1,136 @@
...
+
+            #[test]
+            fn multi_sig_nop() {
+                invoke::<MultiSig<NopSignature>>();
+            }
+
+            // TODO: add module for bls signature collection
+        }
+    };
+}
```
michael-yxchen: The macro here runs the same test code on all SignatureCollection implementations. Thus requiring a uniform error type if we want to assert on the exact error produced

ariqchowdhury:
> instead of having different SCT defining their own error types, we have a general SignatureCollectionError struct. It makes it easier to test common behaviors across different implementations, but the errors are less expressive

I haven't looked at this in detail, but I don't know if I'm comfortable changing the error types we report to help make the testing easier...I think we will definitely want to know the specific errors we are getting if this fails.

michael-yxchen:
> > instead of having different SCT defining their own error types, we have a general SignatureCollectionError struct. It makes it easier to test common behaviors across different implementations, but the errors are less expressive
> 
> I haven't looked at this in detail, but I don't know if I'm comfortable changing the error types we report to help make the testing easier...I think we will definitely want to know the specific errors we are getting if this fails.

yeah I agree. Thinking from our code using the SignatureCollection, it doesn't have context on which implementation it's calling into, so I feel implementation specific errors won't provide that much context. Will discuss offline

monad-consensus-types/src/multi_sig.rs
```diff
@@ -65,54 +31,74 @@ impl<S: CertificateSignatureRecoverable> Hashable for MultiSig<S> {
...
-
-                let sig_entry = sig_map.entry(node_id).or_insert(sig);
-                if *sig_entry != sig {
-                    return Err(MultiSigError::ConflictingSignatures((
-                        node_id, *sig_entry, sig,
-                    )));
-                }
+                if let Ok(pubkey_recovered) = sig.recover_pubkey(msg) {
+                    if pubkey_recovered != *pubkey {
+                        info!(
```
ariqchowdhury: maybe this should be bumped up to warn?

monad-consensus-types/src/multi_sig.rs
```diff
@@ -65,54 +31,74 @@ impl<S: CertificateSignatureRecoverable> Hashable for MultiSig<S> {
...
+                            node_id, sig, *pubkey, pubkey_recovered,
+                        );
+                        invalid_sigs.push(sig);
+                        continue;
+                    }
 
-                <S as CertificateSignature>::verify(&sig, msg, pubkey)
-                    .map_err(|_| MultiSigError::InvalidSignature(sig))?;
+                    if sig.verify(msg, pubkey).is_err() {
+                        info!("Invalid sig NodeId={:?} sig={:?} ", node_id, sig);
```
ariqchowdhury: same thing, warn

monad-consensus-types/src/multi_sig.rs
```diff
@@ -65,54 +31,74 @@ impl<S: CertificateSignatureRecoverable> Hashable for MultiSig<S> {
...
+                    }
 
-                multi_sig.push(sig);
+                    let sig_entry = sig_map.entry(node_id).or_insert(sig);
+                    if *sig_entry != sig {
+                        // unreachable if signature algo is deteministic
+                        // no two valid signatures can be created with the same keypair
+                        return Err(SignatureCollectionError::ConflictingSignatures((
+                            node_id, *sig_entry, sig,
+                        )));
```
ariqchowdhury: "deterministic", and log error here?

monad-consensus-types/src/multi_sig.rs
```diff
@@ -65,54 +31,74 @@ impl<S: CertificateSignatureRecoverable> Hashable for MultiSig<S> {
...
+                    }
+                } else {
+                    info!(
```
ariqchowdhury: warn

monad-consensus-types/src/multi_sig.rs
```diff
@@ -121,35 +107,169 @@ impl<S: CertificateSignatureRecoverable> SignatureCollection for MultiSig<S> {
...
                 node_ids.push(*node_id);
-                pubkeys.remove(pk);
+                verified_sigs.remove(pk);
             }
         }
 
-        if !pubkeys.is_empty() {
-            // TODO: collect evidence: author included signatures from nodes not in the validator set
+        // verified_sigs contains valid but non-validator signatures
+        invalid_sigs.extend(verified_sigs.values());
```
ariqchowdhury: nit: can we change the name of verified_sigs to be more descriptive so that the comment isn't necessary?

monad-consensus-types/src/signature_collection.rs
```diff
@@ -11,15 +11,43 @@ pub type SignatureCollectionKeyPairType<SCT> =
 pub type SignatureCollectionPubKeyType<SCT> =
     <SignatureCollectionKeyPairType<SCT> as CertificateKeyPair>::PubKeyType;
 
+#[derive(Debug, PartialEq, Eq)]
+pub enum SignatureCollectionError<S> {
+    NodeIdNotInMapping(Vec<(NodeId, S)>),
+    ConflictingSignatures((NodeId, S, S)), // only possible for non-deterministic signature
```
ariqchowdhury: nit: comment on separate line

monad-consensus-types/src/multi_sig.rs
```diff
@@ -121,35 +107,169 @@ impl<S: CertificateSignatureRecoverable> SignatureCollection for MultiSig<S> {
...
+        certificate_signature::CertificateSignature,
+        signature_collection::{SignatureCollection, SignatureCollectionError},
+        test_utils::setup_sigcol_test,
+    };
+
+    type SignatureCollectionType = MultiSig<SecpSignature>;
+
+    #[test_case(7, 5, 1; "seed 7, 1/5 invalid")]
+    #[test_case(7, 32, 5; "seed 7, 5/32 invalid")]
+    #[test_case(16, 37, 7; "seed 16, 7/37 invalid")]
```
ariqchowdhury: Should we have a testcase for 100% invalid?
also, just curious, is there something special about these ratios of invalid to valid?

monad-consensus-types/src/multi_sig.rs
```diff
@@ -121,35 +107,169 @@ impl<S: CertificateSignatureRecoverable> SignatureCollection for MultiSig<S> {
...
+        }
+
+        let mut sig_col = SignatureCollectionType::new(sigs, &valmap, msg_hash.as_ref()).unwrap();
+        // add a signature from a non-validator signer
+        let sig = << SignatureCollectionType as SignatureCollection>::SignatureType as CertificateSignature>::sign(msg_hash.as_ref(), &non_validator_cert_key);
+        sig_col.sigs.push(sig);
+
+        assert_eq!(
+            sig_col.verify(&valmap, msg_hash.as_ref()).unwrap_err(),
+            SignatureCollectionError::InvalidSignatures(vec![sig])
```
ariqchowdhury, line 279: shouldn't non-validator signer lead to the `NodeIdNotInMapping` error?

monad-consensus-types/src/multi_sig.rs
michael-yxchen: sure

monad-consensus-types/src/multi_sig.rs
michael-yxchen: Sure.

The ratio was random.. It doesn't matter for Multisig. Changed the test_cases to remove any confusion

monad-consensus-types/src/multi_sig.rs
michael-yxchen: I put it under `InvalidSigantures` due to some quirks of the recoverable signature. With recoverable siganture, we can't distinguish between a signature created over a different message with a validator keypair, and a signature created over the correct message with a non-validator keypair. Both recover to non-validator pubkeys (with the correct msg) and verify correctly. I put it under the capture-all `InvalidSignatures` since we can't differentiate the cases.

ariqchowdhury: got it, makes sense

PR #205, opened 2023-08-15
Update MockLedger to support Ledger Fetch

Yimaha:
Provide Ledger with fetch historical block capability, 
as a consequence enforced a trait on Block

Preparation for Block-Sync

### DIFF
--- a/monad-executor/src/executor/ledger.rs
+++ b/monad-executor/src/executor/ledger.rs
@@ -1,4 +1,5 @@
 use std::{
+    collections::HashMap,
     marker::Unpin,
     ops::DerefMut,
     pin::Pin,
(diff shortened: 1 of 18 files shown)


monad-executor/src/executor/ledger.rs
```diff
@@ -1,57 +1,63 @@
...
         for command in commands {
             match command {
                 LedgerCommand::LedgerCommit(block) => {
+                    // it is assumed the consensus verified that block is valid, thus commiting shouldn't cause issue
+                    // only concern might be hash collision
+                    self.blockchain_kv
+                        .insert(block.get_id(), self.blockchain.len());
                     self.blockchain.push(block);
                 }
                 LedgerCommand::LedgerFetch(block_id, cb) => {
```
Yimaha, line 59: A good question to ask here is if we need some sort of reset mechanism to clear cb if needed. but personally thinking no? (similar to MempoolCommand::FetchFullReset) There is some part of code that talks about it but I thought between executor and state it should be sequential alwasy?

Yimaha: Talked to @omegablitz, it seems like lack of reset emit might cause duplicate event during replay, will update a clear function during next pr to fix the replay issue

monad-executor/src/executor/ledger.rs
```diff
@@ -1,57 +1,63 @@
...
+use monad_consensus_types::block::BlockType;
+use monad_types::BlockId;
 
 use crate::{Executor, LedgerCommand};
 
-pub struct MockLedger<O, E> {
+pub struct MockLedger<O: BlockType, E> {
     blockchain: Vec<O>,
-    ledger_fetch_cb: Option<Box<dyn (FnOnce(Option<O>) -> E) + Send + Sync>>,
+    blockchain_kv: HashMap<BlockId, usize>,
```
ariqchowdhury: nit: rename, block_index?

monad-executor/src/executor/ledger.rs
Yimaha: also this will not impact anything actively in the repo because nothing is calling this yet :3

PR #209, opened 2023-08-17
Refactor transformers to be managed by Pipeline

Yimaha:
once the format is approved test cases will be added in a separate pr

### DIFF
--- a/monad-executor/src/transformer.rs
+++ b/monad-executor/src/transformer.rs
@@ -1,0 +1,328 @@
+use std::{
+    collections::HashSet,
+    fmt::Debug,
+    mem,
+    ops::{Index, IndexMut},
+    time::Duration,
+};
+
+use rand::{prelude::SliceRandom, Rng};
+use rand_chacha::{rand_core::SeedableRng, ChaChaRng};
+
+use crate::{mock_swarm::LinkMessage, PeerId};
+
+/**
+ * TransformerStream carries (Duration, LinkedMessages) that emited
+ * after a transformer with Transform Trait complete their transform
+ *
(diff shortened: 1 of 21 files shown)


monad-state/tests/order.rs
```diff
@@ -52,10 +53,11 @@ fn all_messages_delayed(direction: TransformerReplayOrder) {
...
 
     run_one_delayed_node(
-        vec![
-            LatencyTransformer(Duration::from_millis(1)).boxed(),
-            PartitionThenReplayTransformer::new(filter_peers, 200, direction).boxed(),
-        ],
+        TransformerPipeline::new(vec![
+            Layer::Latency(LatencyTransformer(Duration::from_millis(1))),
+            Layer::Partition(PartitionTransformer(filter_peers)),
+            Layer::Replay(ReplayTransformer::new(50, direction)),
```
Yimaha: Note that I reduced to 50 because before it was listening to ALL message and wait until you hit 200, right now it just listen to 50 message that comes specifically from/to Node 1

monad-executor/src/transformer.rs
```diff
@@ -0,0 +1,311 @@
...
+use std::{
+    collections::HashSet,
+    fmt::Debug,
+    mem,
+    ops::{Index, IndexMut},
+    time::Duration,
+};
+
+use crate::{mock_swarm::LinkMessage, PeerId};
+pub struct TransformerStream<M> {
```
ariqchowdhury, line 25: can you add some comments to describe the fields and intent of TransformerStream

Yimaha: of course

PR #213, opened 2023-08-22
Add BlsSignatureCollection implementation

michael-yxchen:
BlsSignatureCollection uses a bitvec to represent all the signers and aggregate signatures from all signers into a single bls signature.

It optimistically aggregates the bls signatures without verifying them, to avoid the expesive verification on the bls12-381 curve. A binary tree stores intermediate aggregation results. If the aggregated signature is invalid, we can efficiently search the tree for invalid signatures and provide slashing evidence.

### DIFF
--- a/monad-consensus-types/src/bls.rs
+++ b/monad-consensus-types/src/bls.rs
@@ -1,0 +1,424 @@
+use std::collections::{HashMap, VecDeque};
+
+use bitvec::prelude::*;
+use monad_crypto::bls12_381::{
+    BlsAggregatePubKey, BlsAggregateSignature, BlsKeyPair, BlsSignature,
+};
+use monad_types::{Hash, NodeId};
+
+use crate::{
+    signature_collection::{
+        SignatureCollection, SignatureCollectionError, SignatureCollectionKeyPairType,
+    },
+    validation::{Hashable, Hasher},
+    voting::ValidatorMapping,
+};
+
+#[derive(Debug)]
+struct AggregationTree {
+    nodes: Vec<BlsSignatureCollection>,
+}
+
+impl AggregationTree {
+    fn new(
+        sigs: &Vec<(NodeId, BlsSignature)>,
+        validator_mapping: &ValidatorMapping<BlsKeyPair>,
+        validator_index: &HashMap<NodeId, usize>,
+    ) -> Self {
+        // build the binary heap represented as vector
+        if sigs.is_empty() {
+            return Self {
+                nodes: vec![BlsSignatureCollection::with_capacity(
+                    validator_mapping.map.len(),
+                )],
+            };
+        }
+
+        let n = sigs.len();
+        let total_nodes = n * 2 - 1;
+        let mut nodes =
+            vec![BlsSignatureCollection::with_capacity(validator_mapping.map.len()); total_nodes];
+
+        // copy all the signature as leaves
+        for (i, (node_id, sig)) in sigs.iter().enumerate() {
+            let cert = nodes.get_mut(n + i - 1).expect("node in range");
+            cert.signers.set(
+                *validator_index.get(node_id).expect("validator index"),
+                true,
+            );
+            cert.sig.add_assign(sig).expect("bls add assign");
+        }
+
+        // build the tree bottom-up. A node is the aggregation of its two children
+        if n > 1 {
+            for i in (0..n - 1).rev() {
+                let left = &nodes[2 * i + 1];
+                let right = &nodes[2 * i + 2];
+
+                *nodes.get_mut(i).unwrap() = merge_nodes(left, right);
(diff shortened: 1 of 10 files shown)


monad-consensus-types/src/bls.rs
```diff
@@ -0,0 +1,279 @@
...
+                        unverified_certs.push_back(cert_idx * 2 + 2);
+                    }
+                }
+            }
+        }
+
+        if invalid_sig.is_empty() {
+            Ok(self.nodes[0].clone())
+        } else {
+            println!("{:?}", invalid_sig.len());
```
ariqchowdhury: left over from debugging or should we log this?

monad-consensus-types/src/bls.rs
```diff
@@ -0,0 +1,279 @@
...
+        n1.num_signatures() + n2.num_signatures(),
+        cert.num_signatures()
+    );
+
+    cert
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct BlsSignatureCollection {
+    signers: BitVec<u8, Msb0>,
```
ariqchowdhury, line 117: we should check performance between lsb for local node stuff and considering encoding time for network messages.

monad-consensus-types/src/bls.rs
```diff
@@ -0,0 +1,279 @@
...
+        test_utils::setup_sigcol_test,
+    };
+
+    type SignatureCollectionType = BlsSignatureCollection;
+
+    #[test_case(7, 5, 1; "seed 7, 1/5 invalid")]
+    #[test_case(7, 27, 1; "seed 7, 1/27 invalid")]
+    #[test_case(7, 32, 5; "seed 7, 5/32 invalid")]
+    #[test_case(42, 32, 5; "seed 42, 5/32 invalid")]
+    #[test_case(16, 37, 7; "seed 16, 7/37 invalid")]
```
ariqchowdhury: left same comment in the other PR about the fractions of invalid sigs. And should we add a test for a collection of 1?

monad-consensus-types/src/bls.rs
```diff
@@ -0,0 +1,279 @@
...
+        }
+        for (node_id, sig) in sigs.iter() {
+            if !validator_mapping.map.contains_key(node_id) {
+                non_validator.push((*node_id, *sig));
+            }
+            let entry = sigs_map.entry(*node_id).or_insert(*sig);
+            if *entry != *sig {
+                return Err(SignatureCollectionError::ConflictingSignatures((
+                    *node_id, *entry, *sig,
+                )));
```
ariqchowdhury, line 160: do we have test cases to generate this error?

monad-consensus-types/src/bls.rs
```diff
@@ -0,0 +1,279 @@
...
+            }
+        }
+
+        if !non_validator.is_empty() {
+            return Err(SignatureCollectionError::NodeIdNotInMapping(non_validator));
+        }
```
ariqchowdhury, line 166: test case to generate this error?

monad-consensus-types/src/bls.rs
```diff
@@ -0,0 +1,279 @@
...
+        Ok(signers)
+    }
+
+    fn num_signatures(&self) -> usize {
+        self.signers.count_ones()
+    }
+}
+
+#[cfg(test)]
```
ariqchowdhury, line 217: are there tests that aggregate collections and merge different collections?

monad-consensus-types/src/bls.rs
michael-yxchen: will remove

monad-consensus-types/src/bls.rs
michael-yxchen: I will add tests for the aggregation tree and `merge_nodes` function. On the signature collection level, we don't currently support merging. I'm thinking implementing this feature when we want to do intermediate aggregation during gossiping

michael-yxchen:
> Are there supposed to be protobuf changes to handle the new SignatureCollection type? Or putting that in a different PR?

serialization and protobuf changes will come in a following PR

monad-consensus-types/src/bls.rs
michael-yxchen: Will add tests to cover collection of 1 and all invalid

The ratio doesn't matter either here. I'll clean it up so it's less confusing. The number of signatures matters. If the number is a power of 2, all signatures live on the leaf level of the binary tree. Otherwise, some may go up to the second to last level.

monad-consensus-types/src/bls.rs
michael-yxchen: this error is covered by `test_node_id_not_in_validator_mapping` in signature_collection.rs

PR #216, opened 2023-08-23
Add block-sync and associated testcases

Yimaha:
block-sync is a way for blocks to catch up from missing block, caused by either extreme network delay, lost node, or malicious leader censoring certain node by provide byzantine proposal.

Once a proposal is fully handled, if the pending block generated by the current proposal doesn't contain a path to root, it will try to request the closest (to the current pending block) block from a validator in round-robin fashion until either the block is retrieved or the original proposal finally arrived.

### DIFF
--- a/monad-block-sync/src/lib.rs
+++ b/monad-block-sync/src/lib.rs
@@ -1,7 +1,7 @@
 use monad_consensus::messages::{
     consensus_message::ConsensusMessage, message::RequestBlockSyncMessage,
 };
-use monad_consensus_state::command::ConsensusCommand;
+use monad_consensus_state::command::{ConsensusCommand, FetchedBlock};
 use monad_consensus_types::{
     message_signature::MessageSignature, signature_collection::SignatureCollection,
 };
@@ -9,12 +9,15 @@ use monad_executor::{PeerId, RouterTarget};
 use monad_types::{BlockId, NodeId};
 use monad_validator::validator_set::ValidatorSetType;
 
-pub struct BlockSyncState {}
+#[derive(Debug)]
+pub struct BlockSyncState {
+    round_robin_validator_selector: usize,
+}
 
-pub trait BlockSyncProcess<ST, SCT, VT>
+pub trait BlockSyncProcess<ST, SC, VT>
 where
     ST: MessageSignature,
-    SCT: SignatureCollection,
+    SC: SignatureCollection,
     VT: ValidatorSetType,
 {
     fn new() -> Self;
@@ -23,31 +26,36 @@ where
         &mut self,
         blockid: BlockId,
         validators: &VT,
-    ) -> (RouterTarget, ConsensusMessage<ST, SCT>);
+    ) -> (RouterTarget, ConsensusMessage<ST, SC>);
 
     fn handle_request_block_sync_message(
         &mut self,
         author: NodeId,
         s: RequestBlockSyncMessage,
-        validators: &VT,
-    ) -> Vec<ConsensusCommand<ST, SCT>>;
+    ) -> Vec<ConsensusCommand<ST, SC>>;
 }
 
-impl<ST, SCT, VT> BlockSyncProcess<ST, SCT, VT> for BlockSyncState
+impl<ST, SC, VT> BlockSyncProcess<ST, SC, VT> for BlockSyncState
 where
     ST: MessageSignature,
-    SCT: SignatureCollection,
+    SC: SignatureCollection,
     VT: ValidatorSetType,
 {
     fn new() -> Self {
-        BlockSyncState {}
+        BlockSyncState {
+            round_robin_validator_selector: 0,
+        }
     }
+
     fn request_block_sync(
         &mut self,
         blockid: BlockId,
         validators: &VT,
-    ) -> (RouterTarget, ConsensusMessage<ST, SCT>) {
-        let target = RouterTarget::PointToPoint(PeerId(validators.get_list()[0].0));
+    ) -> (RouterTarget, ConsensusMessage<ST, SC>) {
+        let target = RouterTarget::PointToPoint(PeerId(
+            validators.get_list()[self.round_robin_validator_selector % validators.len()].0,
+        ));
+        self.round_robin_validator_selector += 1;
         let message =
             ConsensusMessage::RequestBlockSync(RequestBlockSyncMessage { block_id: blockid });
         (target, message)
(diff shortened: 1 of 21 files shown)


monad-executor/src/replay_nodes.rs
```diff
@@ -115,7 +115,8 @@ where
                             .mut_blockchain();
                         block.push(b);
                     }
-                    LedgerCommand::LedgerFetch(b_id, cb) => {}
+                    // Need review, based on what I understood so far this is fine?
+                    _ => {}
```
Yimaha: Is this okay?

the intention of replay node seems  to be simply replaying event and you provide almost like a fake ledger for it, so ledger fetch is safe to be ignored?

monad-consensus-state/src/lib.rs
```diff
@@ -323,11 +323,29 @@ where
...
     ) -> Vec<ConsensusCommand<ST, SCT>> {
-        Vec::new()
+        let block = b.block;
+        if self.pending_block_tree.is_valid(&block) {
+            // you can now append to tree
+            self.pending_block_tree
+                .add(block)
+                .expect("failed to add block to tree during block sync");
+        }
+        // TODO, if tree does not have a parent fire another RequestSync
```
Yimaha: should there be a recursive request here? in case if you still cannot find a path to root

monad-executor/src/replay_nodes.rs
Yimaha: Confirmed its fine

monad-consensus-state/src/lib.rs
Yimaha: there should be :3

monad-block-sync/src/lib.rs
```diff
@@ -57,8 +71,154 @@ where
...
+                requester: author,
+                block,
+            }),
+        )]
+    }
+
+    // check if the response from block sync is actually a parent we are looking for
+    fn verify_block_is_found(&mut self, bid: &BlockId) -> bool {
+        if self.requested_blocks.contains(bid) {
+            self.requested_blocks.remove(bid);
```
ariqchowdhury: HashSet::remove returns a bool if the value was present

monad-block-sync/src/lib.rs
```diff
@@ -57,8 +71,154 @@ where
         &mut self,
         author: NodeId,
         s: RequestBlockSyncMessage,
-        validators: &VT,
-    ) -> Vec<ConsensusCommand<ST, SCT>> {
-        vec![]
+    ) -> Vec<ConsensusCommand<ST, SC>> {
+        let block_id = s.block_id;
+        vec![ConsensusCommand::LedgerFetch(
+            block_id,
```
ariqchowdhury: nit: s.block_id here

monad-block-sync/src/lib.rs
```diff
@@ -1,20 +1,25 @@
...
+use monad_consensus_state::command::{ConsensusCommand, FetchedBlock};
 use monad_consensus_types::{
     message_signature::MessageSignature, signature_collection::SignatureCollection,
 };
 use monad_executor::{PeerId, RouterTarget};
 use monad_types::{BlockId, NodeId};
 use monad_validator::validator_set::ValidatorSetType;
 
-pub struct BlockSyncState {}
+pub struct BlockSyncState {
```
ariqchowdhury, line 13: add Debug

monad-block-sync/src/lib.rs
```diff
@@ -57,8 +71,154 @@ where
...
+    use crate::{BlockSyncProcess, BlockSyncState};
+    type ST = NopSignature;
+    type SC = MultiSig<ST>;
+    type VT = ValidatorSet;
+
+    fn verify_target_and_message(
+        target: &RouterTarget,
+        message: &ConsensusMessage<ST, SC>,
+        desire_pubkey: PubKey,
+        desire_block_id: BlockId,
```
ariqchowdhury: nit: desired

monad-consensus-state/src/lib.rs
```diff
@@ -323,11 +323,28 @@ where
         cmds
     }
 
+    /**
+     * previous step should be ensured
```
ariqchowdhury: "previous step" is kind of confusing, maybe just "this function expects that..."
And then, "it has a valid qc" isn't necessary, incoming messages are Unverified<Message> and transformed to Message for that reason
If its expected that the block in this message was requested, I think that needs to be asserted here, can't be limited to a comment. If someone changes the code and ignores the comment then things will be broken. 
Will the pending block tree go into an invalid state if it is receiving block sync messages that it didn't request? 

Maybe the way to address this is to keep the `requested_blocks` set in consensus-state and let `handle_block_sync_message` do the checking that the block was actually asked for.

monad-block-sync/src/lib.rs
```diff
@@ -1,20 +1,25 @@
...
 use monad_validator::validator_set::ValidatorSetType;
 
-pub struct BlockSyncState {}
+pub struct BlockSyncState {
+    round_robin_validator_selector: usize,
+    requested_blocks: HashSet<BlockId>,
+}
 
-pub trait BlockSyncProcess<ST, SCT, VT>
+pub trait BlockSyncProcess<ST, SC, VT>
```
ariqchowdhury, line 17: nit:  can we keep this as SCT so its consistent with consensus-state? I realize there are other places in the code base where it may have changed, but we can at least avoid adding more inconsistent naming here

monad-consensus-state/src/lib.rs
Yimaha: The original plan is to limit the change to consensus-state but I do agree this approach where block request checking and block request adding is seperate is not well.

I don't believe this will break the blocktree but this might allow block tree to grow indefinitely if not managed well (or someone wrote a bug in future it might cause a lot of issue)

I will move the logic of checking requested block into consensus (and remove the top pr + more test to ensrue this behaviour is checked)

monad-block-sync/src/lib.rs
Yimaha: This is due to linter I believe, for some reason linter think SCT is some form of acronym and refuse to let it pass? I will double check to see if there is a way to get pass it

ariqchowdhury: oh, if its from the linter you can leave it...I think we had an issue like this somewhere else too, I'll figure out a more general solution so that we can have something uniform

monad-consensus-state/src/lib.rs
```diff
@@ -323,11 +325,32 @@ where
...
+        let mut cmds = vec![];
+        let block = b.block;
+        let bid = block.get_id();
+        if self.requested_blocks.remove(&bid) && self.pending_block_tree.is_valid(&block) {
+            self.pending_block_tree
+                .add(block)
+                .expect("failed to add block to tree during block sync");
+            if let Some(blockid) = self.pending_block_tree.get_missing_ancestor(&bid) {
+                cmds.push(ConsensusCommand::RequestSync { blockid });
+                self.requested_blocks.insert(blockid);
```
ariqchowdhury: I would consider adding a helper function, `create_request_sync(BlockId) -> ConsensusCommand` since these 2 statements are always need to go together and should always be the same blockid. 
Its used in 2 places right now and its definitely possible that it could be required in more places in the future.

PR #218, opened 2023-08-24
Allow for generic RouterScheduler in MockSwarm

omegablitz:
Previously MockSwarm would not allow for truly generic RouterScheduler - it would require RouterScheduler's message type to be equivalent to State's message type.

### DIFF
--- a/monad-testutil/src/swarm.rs
+++ b/monad-testutil/src/swarm.rs
@@ -12,7 +12,7 @@ use monad_crypto::{
     NopSignature,
 };
 use monad_executor::{
-    executor::mock::{MockExecutor, NoSerRouterScheduler},
+    executor::mock::{MockExecutor, NoSerRouterScheduler, RouterScheduler},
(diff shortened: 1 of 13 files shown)


monad-testutil/src/swarm.rs
```diff
@@ -46,6 +46,7 @@ type MS = MonadState<
 type MC = MonadConfig<SignatureCollectionType, TransactionValidatorType>;
 type MM = <MS as State>::Message;
 type RS = NoSerRouterScheduler<MM>;
+type Rsc = <RS as RouterScheduler>::Config;
```
ariqchowdhury, line 49: nit: RSC?

omegablitz: That's what I had originally but clippy complained lol

omegablitz: My next PR is anyway going to remove all of the top-level RSC instantiations... so I don't think it matters too much

PR #219, opened 2023-08-24
Add seeded randomness In mock_swarm's event

Yimaha:
When selecting external or internal events for next processing event, mock_swarm
always preferred internal. This is not a desirable because it limit the testing capability
of mock_swarm.

This commit allows mock_swarm to select event at random
(internal vs external) if arrived at the same time. This modification allows for some tougher testing environment

Currently, the random selection capability is disabled by setting preference to exclusively internal event, as lack of block-sync / checkpoint capability can cause certain integration tests to fail (which is a good)

As a side effect, too many parameter is added to the Nodes during initialization, and some may prefer default parameter too. To introduce better parameter management, NodesConfig is used to managing the init parameter passed to Nodes.

### DIFF
--- a/monad-executor/src/mock_swarm.rs
+++ b/monad-executor/src/mock_swarm.rs
@@ -1,6 +1,6 @@
 use std::{
     cmp::Reverse,
-    collections::{BTreeMap, BTreeSet, BinaryHeap},
+    collections::{BTreeMap, BinaryHeap},
     time::Duration,
 };
 
@@ -8,6 +8,8 @@ use futures::StreamExt;
 use monad_crypto::secp256k1::PubKey;
 use monad_types::{Deserializable, Serializable};
 use monad_wal::PersistenceLogger;
+use rand::Rng;
+use rand_chacha::{rand_core::SeedableRng, ChaChaRng};
 use tracing::info_span;
 
 use crate::{
@@ -39,7 +41,40 @@ where
         Some(self.cmp(other))
     }
 }
+pub enum NextTickEvent {
+    Internal { pid: PeerId, tick: Duration },
+    External { tick: Duration },
+}
 
+pub struct NodesConfig<S, RS, T, LGR>
+where
+    S: State,
+    RS: RouterScheduler,
+    T: Pipeline<RS::Serialized>,
+    LGR: PersistenceLogger<Event = TimedEvent<S::Event>>,
+{
+    pub peers: Vec<(PubKey, S::Config, LGR::Config, RS::Config)>,
+    pub pipeline: T,
+    pub seed: u64,
+    pub bias: f64,
+}
+
+impl<S, RS, T, LGR> Default for NodesConfig<S, RS, T, LGR>
+where
+    S: State,
+    RS: RouterScheduler,
+    T: Pipeline<RS::Serialized>,
+    LGR: PersistenceLogger<Event = TimedEvent<S::Event>>,
+{
+    fn default() -> Self {
+        Self {
+            peers: vec![],
+            pipeline: T::default(),
+            seed: 1,
+            bias: 1.0,
+        }
+    }
+}
 pub struct Nodes<S, RS, T, LGR, ME>
 where
     S: State,
@@ -51,6 +86,9 @@ where
     states: BTreeMap<PeerId, (MockExecutor<S, RS, ME>, S, LGR)>,
     pipeline: T,
     scheduled_messages: BinaryHeap<Reverse<(Duration, LinkMessage<RS::Serialized>)>>,
+    rng_gen: ChaChaRng,
+    ie_bias: f64,
+    pref_ie: bool,
 }
 
 impl<S, RS, T, LGR, ME> Nodes<S, RS, T, LGR, ME>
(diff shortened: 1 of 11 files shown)


monad-executor/src/mock_swarm.rs
```diff
@@ -90,15 +95,20 @@ where
...
         }
-
+        let mut gen = ChaChaRng::seed_from_u64(1);
+        let next_rng_pref = gen.gen_range(0.0..=1.0);
         Self {
             states,
             pipeline,
             scheduled_messages: Default::default(),
+            rng_gen: gen,
+            rng_leaned: 1.0,
```
Yimaha: by setting rng lean to 1.0, we make sure all internal event still get processed first

monad-executor/src/mock_swarm.rs
```diff
@@ -90,15 +95,20 @@ where
 
             states.insert(PeerId(pubkey), (executor, state, wal));
         }
-
+        let mut gen = ChaChaRng::seed_from_u64(1);
+        let next_rng_pref = gen.gen_range(0.0..=1.0);
```
Yimaha: An alternative init function will be provided for setting seed and rng related param instead of default

monad-executor/src/mock_swarm.rs
```diff
@@ -90,15 +95,20 @@ where
 
             states.insert(PeerId(pubkey), (executor, state, wal));
         }
-
+        let mut gen = ChaChaRng::seed_from_u64(1);
```
michael-yxchen: not introduced by your change. Some of our tests use `StdRng::seed_from_u64` and others use `ChaChaRng`. We should probably decide on one. Thoughts and preferences? @ariqchowdhury @omegablitz

monad-executor/src/mock_swarm.rs
```diff
@@ -90,15 +95,20 @@ where
...
             scheduled_messages: Default::default(),
+            rng_gen: gen,
+            rng_leaned: 1.0,
+            next_rng_pref,
         }
     }
 
-    pub fn next_tick(&self) -> Option<Duration> {
+    // this function merges both the up to date executor event and schedule event as one
+    pub fn peak_next_tick(&self) -> Option<(Option<PeerId>, Duration)> {
```
omegablitz: typo - peek

monad-executor/src/mock_swarm.rs
```diff
@@ -114,80 +124,85 @@ where
...
+                    Some((Some(min_event.unwrap().0.clone()), min_event_tick.clone()))
+                } else {
+                    Some((None, min_scheduled_tick.clone()))
+                }
+            }
+        }
+    }
+
+    pub fn next_tick(&self) -> Option<Duration> {
+        if let Some((_, min_tick)) = self.peak_next_tick() {
```
omegablitz: nit: `map` is more idiomatic

monad-executor/src/mock_swarm.rs
omegablitz: I think using `Option<PeerId>` to indicate that the next tick is an event tick is pretty weird, hard to tell what it's doing from the type signature.

monad-executor/src/mock_swarm.rs
```diff
@@ -114,80 +123,85 @@ where
...
-                            from: id,
-                            to,
-                            message: serialized,
-
-                            from_tick: tick,
-                        };
-                        let transformed = self.pipeline.process(lm);
-                        for (delay, msg) in transformed {
-                            self.scheduled_messages.push(Reverse((tick + delay, msg)));
+            if let Some((maybe_pid, tick)) = self.peak_next_tick() {
```
omegablitz: prefer `match` over `if let Some(_) { } else { }`

`match` lets you reduce nesting, because you can unwrap the nested `Option` as well in the same statement

monad-executor/src/mock_swarm.rs
omegablitz: Probably ChaCha bc it's reproducible across platforms

monad-executor/src/mock_swarm.rs
```diff
@@ -114,80 +123,85 @@ where
...
+                self.next_rng_pref = self.rng_gen.gen_range(0.0..=1.0);
```
omegablitz: [gen_bool](https://docs.rs/rand/latest/rand/trait.Rng.html#method.gen_bool)  may be helpful

monad-executor/src/mock_swarm.rs
Yimaha: the alternative in this case would probably be an enum that includes both types?

monad-executor/src/mock_swarm.rs
```diff
@@ -50,6 +56,9 @@ where
     states: BTreeMap<PeerId, (MockExecutor<S, RS>, S, LGR)>,
     pipeline: T,
     scheduled_messages: BinaryHeap<Reverse<(Duration, LinkMessage<RS::Serialized>)>>,
+    rng_event_pref: bool,
+    rng_gen: ChaChaRng,
+    rng_leaned: f64,
```
ariqchowdhury: `rng_leaned` isn't a good name

monad-testutil/src/swarm.rs
```diff
@@ -156,13 +160,47 @@ pub fn run_nodes<S, ST, SCT, RS, RSC, LGR, P, TVT, ME>(
     );
 }
 
-pub fn run_nodes_until_step<S, ST, SCT, RS, RSC, LGR, P, TVT, ME>(
+pub fn prep_peers<S, ST, SCT, RS, RSC, LGR, TVT>(
```
Yimaha, line 165: at this point the swarm should almost be a generic struct because how complex the requirement has become, its generic but there is a lot of duplication of code here for each function and all function are tightly coupled

Yimaha:
Converting to draft to wait until the parallel execution related changes to mockswarm is done @omegablitz

omegablitz:
My changes are all merged now

Yimaha:
nice :3

Yimaha:
Closing this pr because the new one after parallelized mock swarm is ready #306

PR #223, opened 2023-08-25
Modify TransformerStream from a struct to enum

Yimaha:
Better suited for the use case and cleaner code

### DIFF
--- a/monad-executor/src/transformer.rs
+++ b/monad-executor/src/transformer.rs
@@ -12,31 +12,18 @@ use rand_chacha::{rand_core::SeedableRng, ChaChaRng};
 use crate::{mock_swarm::LinkMessage, PeerId};
 
 /**
- * TransformerStream carries (Duration, LinkedMessages) that emited
+ * TransformerStream carries (Duration, LinkedMessages) that emitted
  * after a transformer with Transform Trait complete their transform
  *
- * remain_message indicate messages that should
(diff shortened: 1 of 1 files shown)


monad-executor/src/transformer.rs
```diff
@@ -15,28 +15,15 @@ use crate::{mock_swarm::LinkMessage, PeerId};
...
-        remain_messages: Vec<(Duration, LinkMessage<M>)>,
-        complete_messages: Vec<(Duration, LinkMessage<M>)>,
-    ) -> Self {
-        TransformerStream {
-            remain_messages,
-            complete_messages,
-        }
-    }
+type StreamMessage<M> = (Duration, LinkMessage<M>);
+pub enum TransformerStream<M> {
```
Yimaha, line 24: In current use case non of the transformer output both continue and complete type given a single message, in future this behaviour might change, but it can be improved by simply adding an extra enum

monad-executor/src/transformer.rs
```diff
@@ -15,28 +15,15 @@ use crate::{mock_swarm::LinkMessage, PeerId};
...
+    Continue(Vec<StreamMessage<M>>), // messages that continue getting processed
+    Complete(Vec<StreamMessage<M>>), // messages that should no longer get process
```
ariqchowdhury: nit: comments on their own line

PR #225, opened 2023-08-29
Add StateRootValidator trait and fix seq_num calculation

ariqchowdhury:
Get seq_num directly from the block being extended in the branch of the blocktree when creating a proposal instead of storing it in consensus-state. 

The state-root executor will at some point turn into, or merge with, the file ledger executor.

### DIFF
--- a/monad-consensus-state/src/command.rs
+++ b/monad-consensus-state/src/command.rs
@@ -13,7 +13,7 @@ use monad_consensus_types::{
     timeout::TimeoutCertificate,
 };
 use monad_executor::RouterTarget;
-use monad_types::{BlockId, Epoch, NodeId, Round};
+use monad_types::{BlockId, Epoch, Hash, NodeId, Round};
 
 pub enum ConsensusCommand<ST, SCT: SignatureCollection> {
     Publish {
@@ -47,6 +47,7 @@ pub enum ConsensusCommand<ST, SCT: SignatureCollection> {
     /// Checkpoints periodically can upload/backup the ledger and garbage clean
     /// persisted events if necessary
     CheckpointSave(Checkpoint<SCT>),
+    StateRootHash(Block<SCT>),
     // TODO add command for updating validator_set/round
     // - to handle this command, we need to call message_state.set_round()
 }
@@ -85,6 +86,7 @@ pub struct FetchedTxs<ST, SCT> {
     pub node_id: NodeId,
     pub round: Round,
     pub seq_num: u64,
+    pub state_root_hash: Hash,
     pub high_qc: QuorumCertificate<SCT>,
     pub last_round_tc: Option<TimeoutCertificate<ST>>,
 
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -13,7 +13,7 @@ use monad_consensus::{
 use monad_consensus_types::{
     block::{Block, BlockType},
     message_signature::MessageSignature,
-    payload::{FullTransactionList, StateRoot},
+    payload::{FullTransactionList, StateRootResult, StateRootValidator, TransactionList},
     quorum_certificate::{QuorumCertificate, Rank},
     signature_collection::{SignatureCollection, SignatureCollectionKeyPairType},
     timeout::TimeoutCertificate,
@@ -23,7 +23,7 @@ use monad_consensus_types::{
 };
 use monad_crypto::secp256k1::{KeyPair, PubKey};
 use monad_executor::{PeerId, RouterTarget};
-use monad_types::{BlockId, NodeId, Round};
+use monad_types::{BlockId, Hash, NodeId, Round};
 use monad_validator::{leader_election::LeaderElection, validator_set::ValidatorSetType};
 
 use crate::command::{ConsensusCommand, FetchedFullTxs, FetchedTxs};
@@ -31,12 +31,11 @@ use crate::command::{ConsensusCommand, FetchedFullTxs, FetchedTxs};
 pub mod command;
 pub mod wrapper;
 
-pub struct ConsensusState<ST, SCT: SignatureCollection, TV> {
+pub struct ConsensusState<ST, SCT: SignatureCollection, TV, SVT> {
     pending_block_tree: BlockTree<SCT>,
     vote_state: VoteState<SCT>,
     high_qc: QuorumCertificate<SCT>,
-    seq_num: u64,
-    state_root: StateRoot,
+    state_root_validator: SVT,
 
     pacemaker: Pacemaker<ST, SCT>,
     safety: Safety,
@@ -117,6 +116,8 @@ where
         b: BlockSyncMessage<SCT>,
     ) -> Vec<ConsensusCommand<ST, SCT>>;
 
+    fn handle_state_update(&mut self, seq_num: u64, root_hash: Hash);
+
     fn get_current_round(&self) -> Round;
 
     fn get_keypair(&self) -> &KeyPair;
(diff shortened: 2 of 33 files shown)


monad-executor/src/executor/state_root_hash.rs
```diff
@@ -0,0 +1,84 @@
...
+use futures::Stream;
+use monad_consensus_types::block::BlockType;
+use monad_types::Hash;
+use rand::RngCore;
+use rand_chacha::{rand_core::SeedableRng, ChaChaRng};
+
+use crate::{Executor, StateRootHashCommand};
+
+pub struct MockStateRootHash<O, E> {
+    update: Option<(u64, Hash)>,
```
omegablitz, line 17: super nit, only noting because I think it's cool:
You could reduce the state you need to keep here. It could be encoded directly into`generate_event: Option<Box<dyn (FnOnce() -> E) + Send + Sync>>`.  `generate_event` would then be set to a lambda that captures the parameters that you want to call it with.

This can be done whenever you have the pattern of storing a function and a parameter only to be used for calling that function.

monad-executor/src/executor/parent.rs
```diff
@@ -33,8 +33,14 @@ where
...
     fn exec(&mut self, commands: Vec<Command<M, OM, B, C>>) {
-        let (router_cmds, timer_cmds, mempool_cmds, ledger_cmds, checkpoint_cmds) =
-            Command::split_commands(commands);
+        let (
+            router_cmds,
+            timer_cmds,
+            mempool_cmds,
+            ledger_cmds,
+            checkpoint_cmds,
+            _state_root_hash_cmds,
```
omegablitz, line 42: I'm confused on how ParentExecutor works without handling StateRoot commands - why wouldn't the state roots never be set?

monad-executor/src/state.rs
```diff
@@ -66,6 +66,10 @@ pub enum CheckpointCommand<C> {
     Save(C),
 }
 
+pub enum StateRootHashCommand<B, E> {
+    LedgerCommit(B, Box<dyn (FnOnce(u64, ConsensusHash) -> E) + Send + Sync>),
```
omegablitz, line 70: On replay this will cause duplicate LedgerCommit events to be generated

eg:
On initial run, event E1 will generate a LedgerCommit command, which will in turn generate E2
On replay, events E1 and E2 will be replayed, but E1 will still generate a LedgerCommit command which will generate an extra event E3.

That's why we have the Reset commands in the other executor. Honestly, the whole command/replay situation seems a bit error-prone, maybe we just ignore fixing this for now and think further on how we can make it better...

monad-consensus-state/src/command.rs
```diff
@@ -47,6 +47,10 @@ pub enum ConsensusCommand<ST, SCT: SignatureCollection> {
     /// Checkpoints periodically can upload/backup the ledger and garbage clean
     /// persisted events if necessary
     CheckpointSave(Checkpoint<SCT>),
+    StateRootHash(
+        Block<SCT>,
+        Box<dyn (FnOnce(u64, Hash) -> (u64, Hash)) + Send + Sync>,
```
omegablitz: I'm forgetting... why is this callback necessary again? Why can't the top-level StateRootHashCommand directly construct a `ConsensusEvent::StateUpdate`?

monad-executor/src/executor/parent.rs
ariqchowdhury: only the mock test nodes are using a real StateRootValidator in ConsensusState right now. The executor in parent will be different, has to be a combination of the state root and upcoming file ledger executor, so this can be set when that work is done.

monad-executor/src/state.rs
ariqchowdhury: I'll just fix this for now and we can think about addressing this separately. We should come up with an improvement to cmd/replay

monad-consensus-state/src/command.rs
ariqchowdhury: didn't think of that, let me try that

monad-consensus-state/src/lib.rs
```diff
@@ -23,20 +23,19 @@ use monad_consensus_types::{
...
 pub mod wrapper;
 
-pub struct ConsensusState<ST, SCT: SignatureCollection, TV> {
+pub struct ConsensusState<ST, SCT: SignatureCollection, TV, SVT> {
     pending_block_tree: BlockTree<SCT>,
     vote_state: VoteState<SCT>,
     high_qc: QuorumCertificate<SCT>,
-    seq_num: u64,
-    state_root: StateRoot,
+    state_root: SVT,
```
michael-yxchen: super nit: maybe `state_root_validator`?

monad-consensus-state/src/lib.rs
```diff
@@ -445,20 +474,49 @@ where
...
+                    }
+                    None => {
+                        // Don't have the necessary state root hash ready so propose
+                        // a NULL block
+                        vec![ConsensusCommand::FetchTxs(
+                            0,
+                            Box::new(move |_txns| FetchedTxs {
+                                node_id,
+                                round,
+                                seq_num,
```
michael-yxchen, line 507: I can think of two ways to populate the sequence number for empty blocks

1. increase the seq num as usual, and commit them to the execution ledger. Execution basically performs a nop and outputs the same execution artifact as the previous block (I think that's the execution behavior, but I'm not 100% sure)
2. don't increase the seq num for empty blocks (use the same as its parent), and only increase them for normal proposal blocks. The empty blocks are thus not committed to the execution ledger.

The code is following option 1. I think option 2 makes more sense for the execution ledger. Empty blocks are to facilitate consensus, so maybe it makes sense to leave it out of execution ledger? 

Either option works. We can decide based on which one is more intuitive.

monad-consensus-state/src/lib.rs
```diff
@@ -174,10 +175,33 @@ where
         author: NodeId,
         p: ProposalMessage<ST, SCT>,
     ) -> Vec<ConsensusCommand<ST, SCT>> {
-        vec![ConsensusCommand::FetchFullTxs(
-            p.block.payload.txns.clone(),
-            Box::new(move |txns| FetchedFullTxs { author, p, txns }),
-        )]
+        // NULL blocks are not required to have state root hashes
```
michael-yxchen, line 178: should we also verify that the `seq_num` is correctly extending the parent block's

monad-consensus-state/src/lib.rs
```diff
@@ -406,10 +431,14 @@ where
...
-            self.seq_num += blocks_to_commit.len() as u64;
-            self.state_root.remove_old_roots(self.seq_num);
+            if let Some(b) = blocks_to_commit.last() {
+                self.state_root.remove_old_roots(b.payload.seq_num);
+            }
 
             if !blocks_to_commit.is_empty() {
+                cmds.extend(blocks_to_commit.iter().map(|b| {
+                    ConsensusCommand::StateRootHash(b.clone(), Box::new(move |a, b| (a, b)))
+                }));
```
michael-yxchen: just to check my understanding here. When we are integrating with execution, do we need the `StateRootHash` command to notify execution? or can execution read from the ledger directly?

monad-consensus-state/src/lib.rs
ariqchowdhury: Yeah, I went with the more simple approach on the consensus side, try and treat empty blocks the same as real blocks as much as possible and prevent too many places where different code/handling has to occur. 
I think your option 2 makes sense longer term, but lets do that change separately later

monad-consensus-state/src/lib.rs
ariqchowdhury: When we integrate with execution, I think we need to have 2 executors -- one for the consensus ledger and one for the execution-ledger. I think the execution-ledger will merge with the functionality of the StateRootHash executor, so these commands will cause the execution-ledger to write the blocks (to file or db), and execution can just monitor changes to the file on their own, and when it finishes a block it can send the hash update to that executor.

monad-consensus-state/src/lib.rs
ariqchowdhury: yes

michael-yxchen: was thinking about this... it seems complicated because we need the parent block to verify whether the sequence number is correctly extending. This makes block-sync block voting, and we wanted to avoid that.

The best idea I've come up with so far is to add sequence number to `QcInfo`. The receiver is convinced that the sequence number is correct by comparing the `seq_num` with `qc.seq_num`. Seems to be on the cutting edge of failure tho

ariqchowdhury: Initially I thought that voting/proposing with missing blocks was less complicated but thinking about it more, it might be making things more complicated. 
I'll figure something out to fix this so that we have an easy way of setting the behaviour we want when block-syncing and making sure its consistent across all the points where it matters. 
And then, I'll update this check.

PR #226, opened 2023-08-29
QuicRouterScheduler implementation

omegablitz:
This commit contains an implementation of a QuicRouterScheduler built on top of quinn-proto. QuicRouterScheduler is generic over Gossip - right now we only have a MockGossip that does a simple broadcast. Reconnect logic hasn't been implemented yet, so blackout tests will fail. After that's fixed, we can port further tests to use the QuicRouterScheduler.

### DIFF
--- a/monad-quic/src/lib.rs
+++ b/monad-quic/src/lib.rs
@@ -1,0 +1,420 @@
+use std::{
+    collections::{BTreeMap, BTreeSet, HashMap, VecDeque},
+    net::{Ipv4Addr, SocketAddr, SocketAddrV4},
+    sync::Arc,
+    time::{Duration, Instant},
+};
+
+use monad_crypto::rustls::UnsafeTlsVerifier;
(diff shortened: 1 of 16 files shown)


monad-quic/src/lib.rs
```diff
@@ -0,0 +1,420 @@
...
+    endpoint: quinn_proto::Endpoint,
+    connections: HashMap<ConnectionHandle, Connection>,
+    outbound_connections: HashMap<PeerId, ConnectionHandle>,
+    outbound_writeable_streams: HashMap<PeerId, StreamId>,
+    inbound_readable_streams: HashMap<PeerId, StreamId>,
+
+    peer_ids: HashMap<PeerId, SocketAddr>,
+    addresses: HashMap<SocketAddr, PeerId>,
+
+    algo: G,
```
ariqchowdhury: nit: change name to gossip?

PR #233, opened 2023-08-31
Change MockSwarm to have a Pipeline per-node

omegablitz:
These commits can be reviewed separately if that's easier; they stand alone.

---
Previously, there was a single Pipeline in MockSwarm. Every message
passed from any node to any other node would mutate the same single
Pipeline. This allowed for max flexibility, but was not parallization
friendly. Now, there exists a distinct Pipeline per node - so all
outbound messages from a given node pass through their own Pipeline. We
don't actually need to change any of our tests, because none of our
Pipeline Transformers depend on global mutable Pipeline state.

### DIFF
--- a/monad-executor/src/executor/mock.rs
+++ b/monad-executor/src/executor/mock.rs
@@ -807,6 +807,9 @@ mod tests {
                     NoSerRouterConfig {
                         all_peers: pubkeys.iter().copied().collect(),
                     },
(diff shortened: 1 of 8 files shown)


omegablitz:
I need to refactor the `PeriodicTransformer` and `ReplayTransformer` first, because they operate on message counts

PR #234, opened 2023-08-31
Refactor Partition and Replay transformers

omegablitz:
The Partition and Replay transformers worked by counting the # of messages that passed through, and then branching depending on that number. This commit changes their behavior to be more declarative - they operate on absolute ticks now, which is less brittle.

### DIFF
--- a/monad-executor/src/transformer.rs
+++ b/monad-executor/src/transformer.rs
@@ -125,25 +125,21 @@ impl<M> Transform<M> for DropTransformer {
 
 #[derive(Clone, Debug)]
 pub struct PeriodicTransformer {
-    pub start: u32,    // when period start
-    pub duration: u32, // how long does it last
-    cnt: u32,          // monotonically inreasing counter
+    pub start: Duration,
+    pub end: Duration,
 }
 
 impl PeriodicTransformer {
-    pub fn new(start: u32, duration: u32) -> Self {
-        PeriodicTransformer {
-            start,
-            duration,
-            cnt: 0,
-        }
+    /// [start, end)
+    pub fn new(start: Duration, end: Duration) -> Self {
+        assert!(start <= end);
+        PeriodicTransformer { start, end }
     }
 }
 
 impl<M> Transform<M> for PeriodicTransformer {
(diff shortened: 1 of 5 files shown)


monad-executor/src/transformer.rs
```diff
@@ -341,6 +333,7 @@ mod test {
         PeerId,
     };
 
+    /// FIXME these should take in from/to/from_tick as params, not have defaults
```
Yimaha, line 336: I can fix this in my pr, but idea is that most messages don't need to have a lot of specification until this point. Default() with a struct as param has a pretty nice solution like this

```
let p = cParams { iInsertMax: 10, ..Default::default() };
```

as an example from here https://stackoverflow.com/questions/19650265/is-there-a-faster-shorter-way-to-initialize-variables-in-a-rust-struct

Yimaha: This approach is very convenient, and is becoming my go-to solution for classes with actual default parameter usage and a lot of parameter to manage

omegablitz: Yeah, my only point was that:
```rust
let mut m = get_mock_message();
m.__ = ___;
```
is an anti-pattern.

It's not a big deal though, it's only in tests. We can change it later.

monad-executor/src/transformer.rs
```diff
@@ -517,41 +523,48 @@ mod test {
...
             mock_message.from = PeerId(keys[3].pubkey());
+            mock_message.from_tick = Duration::from_millis(idx);
             let result = pipe.process(mock_message.clone());
             assert_eq!(result.len(), 1);
             // since its not part of the id, it doesn't get selected and filter
             assert_eq!(result[0].0, Duration::from_millis(30));
             assert!(result[0].1 == mock_message);
         }
-        // not the previous message never made it to periodic, thus it doesn't get trigger the cnt
+        // note the previous message never made it to periodic, thus it doesn't get trigger the cnt
```
michael-yxchen, line 542: nit: this comment is irrelevant with the current `PeriodicTransformer`

PR #236, opened 2023-09-01
Allow FetchTx to determine the max number of txns in proposal

ariqchowdhury:
Removes the parameter on the mempool side so the only limit comes from consensus side via the fetch command.

The hardcoded 10000 tx limit will be put into config in future PR

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -447,16 +447,18 @@ where
         let high_qc = self.high_qc.clone();
         let seq_num = self.seq_num;
 
-        vec![ConsensusCommand::FetchTxs(Box::new(move |txns| {
-            FetchedTxs {
+        vec![ConsensusCommand::FetchTxs(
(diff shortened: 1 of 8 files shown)


monad-consensus-state/src/lib.rs
```diff
@@ -447,16 +447,18 @@ where
         let high_qc = self.high_qc.clone();
         let seq_num = self.seq_num;
 
-        vec![ConsensusCommand::FetchTxs(Box::new(move |txns| {
-            FetchedTxs {
+        vec![ConsensusCommand::FetchTxs(
+            // FIXME get from config
+            10000,
```
omegablitz, line 452: Do we want default to be 10k or 5k? Guess it doesn't matter anyways once we make it configurable

ariqchowdhury: yeah, it's actually still set to 10k in the mempool code anyways

PR #238, opened 2023-09-01
Parallelize MockSwarm

omegablitz:


monad-testutil/src/swarm.rs
```diff
@@ -106,65 +112,37 @@ pub fn run_nodes<S, ST, SCT, RS, RSC, LGR, P, TVT, ME>(
...
 }
 
 pub fn run_nodes_until<S, ST, SCT, RS, RSC, LGR, P, TVT, ME>(
     pubkeys: Vec<PubKey>,
     state_configs: Vec<MonadConfig<SCT, TVT>>,
     router_scheduler_config: RSC,
     logger_config: LGR::Config,
-
     pipeline: P,
+    parallelize: bool,
```
ariqchowdhury, line 142: super nit: if this was an enum instead, it would be a little easier to understand the config when reading test files

omegablitz: I thought the same, but I think a more scalable approach is to create a separate configuration struct to bundle all the params - as having like 8+ ordered params is unreasonable. When in a separate config struct, the bool param will be readable because it'll always be accompanied by the field name (parallelize).  I was too lazy to tag that onto this commit, but will create a separate commit for that eventually whenever it annoys us enough.

PR #243, opened 2023-09-06
Add logging and counters to ConsensusStateProcess

ariqchowdhury:
Only updated one of the unit tests with counter checking but we can follow up this PR and use checks in more of them .


monad-consensus-state/src/lib.rs
```diff
@@ -232,13 +263,15 @@ where
...
         if let Some(v) = vote {
+            inc_count!(created_vote);
             let vote_msg = VoteMessage::<SCT>::new::<H>(v, &self.cert_keypair);
 
             let next_leader = election.get_leader(round + Round(1), validators.get_list());
             let send_cmd = ConsensusCommand::Publish {
                 target: RouterTarget::PointToPoint(PeerId(next_leader.0)),
                 message: ConsensusMessage::Vote(vote_msg),
             };
+            debug!("Created Vote: vote={:?} next_leader={:?}", v, next_leader);
```
michael-yxchen: is there a reason to put the `inc_count` before the code block and `debug` statement after it?

ariqchowdhury: just a mistake, will move the count

PR #249, opened 2023-09-11
RFC: Switch `TimeoutCertificate` to `SignatureCollection`

### LINKED ISSUE #196: Switch `TimeoutMessage` to use the `CertificateSignatures`

michael-yxchen:
Review #256 instead

`TimeoutCertificate` was created with a vector of signatures over the current round and the high qc round. The size of the certificate is thus proportional to the number of signers on it.

We note that nodes are likely to be locked on a small number of high qcs, so we can leverage aggregate signatures to make TC more compact and potentially faster to verify.

`TimeoutMessage` now carries a signature created with the certificate keypair. We bucket the timeout signatures by `high_qc_round` and aggregate signatures in the same bucket when constructing the TC.

Closes #196

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -161,12 +161,20 @@ where
         }
     }
 
-    fn handle_timeout_expiry(
+    fn handle_timeout_expiry<H: Hasher>(
         &mut self,
         event: PacemakerTimerExpire,
     ) -> Vec<PacemakerCommand<ST, SCT>> {
         self.pacemaker
             .handle_event(&mut self.safety, &self.high_qc, event)
+            .into_iter()
+            .map(|cmd| match cmd {
+                PacemakerCommand::PrepareTimeout(tmo) => {
+                    let tmo_msg = TimeoutMessage::new::<H>(timeout, &self.cert_keypair);
+                    PacemakerCommand::Broadcast(tmo_msg)
+                }
+                _ => cmd,
+            })
     }
 
     fn handle_proposal_message<H: Hasher>(
--- a/monad-consensus-types/src/timeout.rs
+++ b/monad-consensus-types/src/timeout.rs
@@ -1,16 +1,54 @@
+use std::collections::HashMap;
+
 use monad_types::*;
 use zerocopy::AsBytes;
 
 use super::quorum_certificate::QuorumCertificate;
-use crate::validation::{Hashable, Hasher};
+use crate::{
+    signature_collection::{
+        SignatureCollection, SignatureCollectionError, SignatureCollectionKeyPairType,
+    },
+    validation::{Hashable, Hasher},
+    voting::ValidatorMapping,
+};
 
 #[derive(Clone, Debug, PartialEq, Eq)]
-pub struct TimeoutInfo<T> {
-    pub round: Round,
-    pub high_qc: QuorumCertificate<T>,
+pub struct Timeout<SCT: SignatureCollection> {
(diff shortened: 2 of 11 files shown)


monad-consensus-state/src/lib.rs
```diff
@@ -161,12 +161,20 @@ where
...
         self.pacemaker
             .handle_event(&mut self.safety, &self.high_qc, event)
+            .into_iter()
+            .map(|cmd| match cmd {
+                PacemakerCommand::PrepareTimeout(tmo) => {
+                    let tmo_msg = TimeoutMessage::new::<H>(timeout, &self.cert_keypair);
+                    PacemakerCommand::Broadcast(tmo_msg)
+                }
+                _ => cmd,
+            })
```
michael-yxchen, line 177: Pacemaker prepares `Timeout`. This function signs `Timeout` to create `TimeoutMessage`

monad-consensus-types/src/timeout.rs
```diff
@@ -1,16 +1,54 @@
...
-    pub high_qc: QuorumCertificate<T>,
+    pub high_qc: QuorumCertificate<SCT>,
+}
+
+impl<SCT: SignatureCollection> Hashable for TimeoutInfo<SCT> {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        state.update(self.round);
+        state.update(self.high_qc.info.vote.id.0.as_bytes());
+        state.update(self.high_qc.get_hash());
+    }
```
michael-yxchen, line 39: We use the `Hashable` trait for integrity hash when creating the message signature.

monad-consensus-types/src/timeout.rs
```diff
@@ -1,16 +1,54 @@
...
 
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+impl<SCT: SignatureCollection> TimeoutInfo<SCT> {
+    pub fn timeout_digest<H: Hasher>(&self) -> Hash {
+        let mut hasher = H::new();
+        hasher.update(self.round);
+        hasher.update(self.high_qc.info.vote.round);
+        hasher.hash()
+    }
+}
```
michael-yxchen: Use `timeout_digest` to obtain a hash over `(TimeoutRound, HighQCRound)`. The CertificateSignature is created over this hash (protocol hash).

monad-consensus-types/src/timeout.rs
```diff
@@ -22,18 +60,50 @@ impl Hashable for HighQcRound {
 }
 
 #[derive(Clone, Debug, PartialEq, Eq)]
-pub struct HighQcRoundSigTuple<S> {
+pub struct HighQcRoundSigColTuple<SCT> {
     pub high_qc_round: HighQcRound,
-    pub author_signature: S,
+    pub sigs: SCT,
 }
```
michael-yxchen, line 66: Signatures over the same `high_qc_round` are aggregated into one `SCT`

monad-consensus/src/messages/message.rs
```diff
@@ -51,15 +51,27 @@ impl<SCT: SignatureCollection> VoteMessage<SCT> {
...
+}
+
+impl<SCT: SignatureCollection> Hashable for TimeoutMessage<SCT> {
     fn hash<H: Hasher>(&self, state: &mut H) {
-        state.update(self.tminfo.round);
-        state.update(self.tminfo.high_qc.info.vote.round);
+        self.timeout.hash(state);
+        self.sig.hash(state);
     }
 }
```
michael-yxchen, line 76: Message integrity hash

PR #251, opened 2023-09-11
Allow retrieval of participants from SignatureCollection

Yimaha:
Prep for leader election

### DIFF
--- a/monad-consensus-types/src/bls.rs
+++ b/monad-consensus-types/src/bls.rs
@@ -1,4 +1,4 @@
-use std::collections::{HashMap, VecDeque};
+use std::collections::{HashMap, HashSet, VecDeque};
 
 use bitvec::prelude::*;
 use monad_crypto::bls12_381::{
@@ -214,6 +214,22 @@ impl SignatureCollection for BlsSignatureCollection {
                 .as_signature()]));
         }
         Ok(signers)
+    }
+
+    fn get_participants(
+        &self,
+        validator_mapping: &ValidatorMapping<SignatureCollectionKeyPairType<Self>>,
+        _msg: &[u8],
+    ) -> HashSet<NodeId> {
+        assert_eq!(self.signers.len(), validator_mapping.map.len());
+
+        let mut signers = HashSet::new();
+        for (bit, (node_id, _)) in self.signers.iter().zip(validator_mapping.map.iter()) {
+            if *bit {
+                signers.insert(*node_id);
+            }
+        }
+        signers
     }
 
     fn num_signatures(&self) -> usize {
--- a/monad-consensus-types/src/multi_sig.rs
+++ b/monad-consensus-types/src/multi_sig.rs
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 use log::{error, warn};
 use monad_proto::proto::signing::ProtoMultiSig;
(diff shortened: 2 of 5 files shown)


monad-consensus-types/src/bls.rs
```diff
@@ -440,4 +456,26 @@ mod test {
...
+        }
+
+        let sig_collection = SignatureCollectionType::new(sigs, &valmap, msg).unwrap();
+        let nodes = sig_collection.get_participants(&valmap, msg);
+
+        assert!(nodes.len() == voting_keys.len());
+        for (node_id, _key) in voting_keys.iter() {
+            assert!(nodes.contains(node_id))
+        }
+    }
```
michael-yxchen: This test can be moved to `monad-consensus-types/src/signature_collection` so we don't need to repeat the test code for different implementations

monad-consensus-types/src/multi_sig.rs
```diff
@@ -303,4 +326,42 @@ mod test {
...
+        let sig = << SignatureCollectionType as SignatureCollection>::SignatureType as CertificateSignature>::sign(msg_hash.as_ref(), &non_validator_cert_key);
+        sig_col.sigs.push(sig);
+
+        let participants = sig_col.get_participants(&valmap, msg_hash.as_ref());
+
+        assert!(participants.len() == 5);
+        for id in valmap.map.keys() {
+            assert!(participants.contains(id));
+        }
+    }
```
michael-yxchen: same here

monad-consensus-types/src/signature_collection.rs
```diff
@@ -65,6 +67,15 @@ pub trait SignatureCollection:
...
 
+    /**
+     * Get participants doesn't verify the validity of the certificate,
+     * but retrieve any valid nodeId participated given a validator mapping.
+     */
+    fn get_participants(
+        &self,
+        validator_mapping: &ValidatorMapping<SignatureCollectionKeyPairType<Self>>,
+        msg: &[u8],
+    ) -> HashSet<NodeId>;
```
michael-yxchen, line 78: nit: what's the reason to return a `HashSet<_>` instead of `Vec<_>`

monad-consensus-types/src/multi_sig.rs
```diff
@@ -153,6 +153,29 @@ impl<S: CertificateSignatureRecoverable> SignatureCollection for MultiSig<S> {
         Ok(node_ids)
     }
 
+    fn get_participants(
+        &self,
+        validator_mapping: &ValidatorMapping<SignatureCollectionKeyPairType<Self>>,
+        msg: &[u8],
+    ) -> HashSet<NodeId> {
+        let mut node_ids = HashSet::new();
+        let mut verified_unvalidated_sigs = HashSet::new();
```
michael-yxchen: nit: we don't verify the signature in the function, so the var name is less appropriate

monad-consensus-types/src/signature_collection.rs
Yimaha: if the aim is to keep interface consistent `Vec<_>` is great too, but I highly doubt there is a case where we will use index based look up like `participants[i]`, since there is naturally no much sense for ordering participants, and each participants is unique. I think majority of the use case should be either
1. iterate through all leaders
2. check if certain leader participated

Maybe another argument is `HashSet<_>` more expansive compared to `Vec<_>`? but I don't think this will be our bottleneck

PR #254, opened 2023-09-13
[WIP] Enable proactive block sync

Yimaha:
(WIP)
Previously, Block sync was only called when a handle_proposal_message_full detected missing blocks, and it potentially retrieves missing blocks without triggering any cleanup/follow up. its not a great design for the long run because it ties the engineering challenge (out of order delivery of blocks / msg lost / msg delay) with the consensus algorithm itself. In general, the logic should be change to "if I observe a proof that a block existed and is approved by 2f+1 validators, then its valid, and I should pause on any operation that relies on such proof to progress, try to retrieve blocks, and unblock all stuck operation"

So far, the only such proof is QC, as of right now, operation that can get stuck because lack of information to support QC is:
1. causing commit (a qc would have cause commit, but because no path to root this action is abandoned)
2. leader election (in future, reputation based leader election sometimes relies on QC to elect)
3. verifying transaction? (optional, but in future the verification of vote might relies on a complete pending_block_tree_history)

This pr introduces the following feature
1. when a QC is observed through forming qc from vote, proposal message, or timeout message, and such qc cannot connect back to the root of the pending_block_tree, then a block sync is triggered 
2. only 1 block sync per block id can happen at the same time, TODO: introducing timeout in future if no respond for too long [monad-crypto/monad-bft/pull/258]
3. when commit is triggered, pending_block_tree should also invalidate all in flight request with lower round number than root (they are no longer useful) [monad-crypto/monad-bft/pull/258 ]

[DEFERRED TO FUTURE]
1. On top of checking the ledger, consensus will also attempt to search the current pending_block_tree to retrieve blocks
2. when a block sync request successfully fixes a tree, one should attempt to cause commit given the original triggering condition.

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -1,4 +1,4 @@
-use std::{collections::HashSet, time::Duration};
+use std::{collections::HashSet, time::Duration, vec};
 
 use log::{debug, warn};
 use monad_blocktree::blocktree::BlockTree;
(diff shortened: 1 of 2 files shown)


Yimaha:
Closing and moving this to issue

PR #255, opened 2023-09-13
Add config param for proposal behaviour when missing blocks

ariqchowdhury:
In certain situations where we are missing blocks in the pending blocktree (don't have a chain of blocks back to the root), we still might have enough information to try and propose a block. The block may have a higher chance to fail, but proposing in that situation gives the validator a chance to try and push a block through.

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -1,7 +1,7 @@
 use std::{collections::HashSet, time::Duration};
 
 use log::{debug, warn};
-use monad_blocktree::blocktree::BlockTree;
+use monad_blocktree::blocktree::{BlockTree, RootKind};
 use monad_consensus::{
     messages::{
         consensus_message::ConsensusMessage,
@@ -44,6 +44,7 @@ pub struct ConsensusState<ST, SCT: SignatureCollection, TV, SVT> {
     safety: Safety,
 
     nodeid: NodeId,
+    config: ConsensusConfig,
 
     transaction_validator: TV,
     requested_blocks: HashSet<BlockId>,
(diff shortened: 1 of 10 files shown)


monad-consensus-state/src/lib.rs
```diff
@@ -534,56 +543,87 @@ where
...
+    #[must_use]
+    fn proposal_policy(&self, parent_bid: &BlockId) -> ConsensusAction {
+        match self.pending_block_tree.root {
+            RootKind::Unrooted(_) => return ConsensusAction::Abstain,
+            RootKind::Rooted(_) => (),
+        }
+
+        if !self.pending_block_tree.tree().contains_key(parent_bid) {
+            return ConsensusAction::Abstain;
+        }
```
michael-yxchen: Just thought of this while reading the PR. We can propose with the `seq_num` from `high_qc.info.vote.seq_num`
[Link to notion](https://www.notion.so/2023-09-11-missingblocks-behaviour-02b29bb655f44c25bb490517372ea1d3?d=a171d93302de4c8b99fa2788f2471df7&pvs=4#7f4e80fd60724f29803c4538b82eb718)

monad-consensus-state/src/lib.rs
```diff
@@ -534,56 +543,87 @@ where
...
-        match seq_num {
-            Some(s) => {
-                let seq_num = s + 1;
-                match self.state_root_validator.get_next_state_root(seq_num) {
-                    Some(h) => {
-                        inc_count!(creating_proposal);
-                        debug!("Creating Proposal: node_id={:?} round={:?} high_qc={:?}, seq_num={:?}, last_round_tc={:?}", 
+        match self.proposal_policy(&parent_bid) {
+            ConsensusAction::Propose => {
+                match self.pending_block_tree.get_seq_num(parent_bid) {
```
michael-yxchen: we can probably fetch `seq_num` from `high_qc` here, and move the decision making to `proposal_policy`

monad-consensus-state/src/lib.rs
ariqchowdhury: added the flag check

monad-consensus-state/src/lib.rs
ariqchowdhury: yup, moved fetching the state hash into the policy and added an action for empty proposal.

PR #257, opened 2023-09-14
Make `MonadState` generic over `Hasher`

### LINKED ISSUE #252: Make `MonadState` depend on a generic `Hasher` type

michael-yxchen:
`HasherType` determines the hashing algorithm when calculating `Block.Id`, `Qc.sigs` hash, etc. The top-level binary creates genesis config and pass it to `MonadState`. `MonadState` must use the same hashing algorithm as the top-level to verify the config.

Previously, `MonadState` is hardcoding the `HasherType` to `Sha256Hash`. So it's possible to use a different hasher for the top-level (`MonadNode`, `testground`), though unlikely because we only have a single Hasher implementation.

This PR address the issue by making `MonadState` generic over `Hasher`. The top-level can configure the hashing algo so it's always consistent with what's used to create the genesis config.

Closes #252

### DIFF
--- a/monad-state/src/lib.rs
+++ b/monad-state/src/lib.rs
@@ -1,4 +1,4 @@
-use std::{fmt::Debug, time::Duration};
+use std::{fmt::Debug, marker::PhantomData, time::Duration};
 
 use message::MessageState;
 use monad_block_sync::BlockSyncProcess;
@@ -24,7 +24,7 @@ use monad_consensus_types::{
     signature_collection::{
         SignatureCollection, SignatureCollectionKeyPairType, SignatureCollectionPubKeyType,
     },
-    validation::Sha256Hash,
+    validation::Hasher,
     voting::{ValidatorMapping, VoteInfo},
 };
 use monad_crypto::secp256k1::{KeyPair, PubKey};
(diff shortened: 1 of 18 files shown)


monad-state/src/lib.rs
```diff
@@ -59,15 +57,18 @@ where
     validator_mapping: ValidatorMapping<SignatureCollectionKeyPairType<SCT>>,
     upcoming_validator_set: Option<VT>,
     block_sync: BST,
+
+    hasher_phantom: PhantomData<HT>,
```
michael-yxchen, line 61: This PR might be debatable cuz the `hasher_phantom` can be inelegant , and inconsistent `Hasher` is nearly impossible.

Just writing it out to collect thoughts. We can close it if we don't think it's necessary.

ariqchowdhury: its unlikely, but its nice to guarantee its consistent. I'm okay with this

omegablitz: If you wanted to remove the PhantomData, you could create something like:
```rust
trait HashAlgo {
    type Hasher: Hasher;
    fn new(&self) -> Self::Hasher;
}
```

and make `hash_algo: H` a member of MonadState.


We did something similar for the `TransactionValidator` (TVT) to avoid the phantom member.

I don't feel strongly either way though, mostly just a matter of preference. The method I mentioned is a little more verbose.

michael-yxchen: That's good to know. I'm going to close this PR. It doesn't feel super necessary to add the `Hasher` to the already long generic list. Can revisit if (unlikely) we later want to experiment with different hashing algos.

PR #258, opened 2023-09-14
Allow Consensus to track in flight block sync

Yimaha:
When request block sync is triggered, it lacks
ability to keep track of blocks that are already
in flight and which validator is currently
processing corresponding block sync request.

such system can cause broadcast storm if given
frequent trigger condition (such as observing qc),
In order to avoid such case, now

1. BlockSyncProcess keeps track of in flight
requests for block sync, (both who are we
asking and block id that we are asking)

2. Nodes are capable to answer block sync
request in 2 different ways: I have block,
or block currently not available

3. requested_bock is being changed to a proper
class for potential future expansion

4. outdated request are also pruned 

5. block sync also relies on qc instead of blockId 
(5 probably belong in the next pr, but they are plan
to be merged right after each other)

More test cases supporting new feature added
to this pr will be added in anther pr

This PR partially complete https://github.com/monad-crypto/monad-bft/pull/254 
This PR is followed up by https://github.com/monad-crypto/monad-bft/pull/260 for styling and testing

### DIFF
--- a/monad-block-sync/src/lib.rs
+++ b/monad-block-sync/src/lib.rs
@@ -1,19 +1,27 @@
+use std::collections::{hash_map::Entry, HashMap};
+
 use monad_consensus::messages::{
-    consensus_message::ConsensusMessage, message::RequestBlockSyncMessage,
+    consensus_message::ConsensusMessage,
+    message::{BlockSyncMessage, RequestBlockSyncMessage},
 };
 use monad_consensus_state::command::{ConsensusCommand, FetchedBlock};
 use monad_consensus_types::{
-    message_signature::MessageSignature, signature_collection::SignatureCollection,
+    block::Block, message_signature::MessageSignature, signature_collection::SignatureCollection,
 };
-use monad_executor::{PeerId, RouterTarget};
+use monad_executor::PeerId;
 use monad_types::{BlockId, NodeId};
 use monad_validator::validator_set::ValidatorSetType;
-
+const DEFAULT_AUTHOR_INDEX: usize = 0;
+const DEFAULT_RETRY_INIT: usize = 0;
 #[derive(Debug)]
 pub struct BlockSyncState {
-    round_robin_validator_selector: usize,
+    request_mapping: HashMap<(PeerId, BlockId), usize>,
 }
-
+pub enum BlockRetrievalResult<ST, SC: SignatureCollection> {
+    Success(Block<SC>),                         // retrieved
+    Failed((PeerId, ConsensusMessage<ST, SC>)), // unable to retrieve
+    IllegalResponse(PeerId),                    // never requested from this peer (slash)
+}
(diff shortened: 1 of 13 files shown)


monad-block-sync/src/lib.rs
```diff
@@ -1,19 +1,27 @@
...
+const DEFAULT_RETRY_INIT: usize = 0;
 #[derive(Debug)]
 pub struct BlockSyncState {
-    round_robin_validator_selector: usize,
+    request_mapping: HashMap<(PeerId, BlockId), usize>,
+}
+pub enum BlockRetrievalResult<ST, SC: SignatureCollection> {
+    Success(Block<SC>),                         // retrieved
+    Failed((PeerId, ConsensusMessage<ST, SC>)), // unable to retrieve
+    IllegalResponse(PeerId),                    // never requested from this peer (slash)
```
Yimaha, line 23: This behaviour (as well as timeout) should be handled by network layer, previously non of our message req a call back, but block sync need a call back. 

right now it just reject any message that doesn't come from the right peer, but in future this should be resolved at router level

monad-state/src/lib.rs
```diff
@@ -435,13 +435,16 @@ where
...
                     ConsensusEvent::FetchedBlock(fetched_b) => {
                         let mut cmds = vec![ConsensusCommand::LedgerFetchReset];
-                        if let Some(b) = fetched_b.block {
-                            let m = BlockSyncMessage { block: b };
-                            cmds.push(ConsensusCommand::Publish {
-                                target: RouterTarget::PointToPoint(PeerId(fetched_b.requester.0)),
-                                message: ConsensusMessage::BlockSync(m),
-                            })
-                        }
+                        let m = BlockSyncMessage {
```
Yimaha, line 438: ugly, patch in test_case pr using into()

monad-state/src/lib.rs
```diff
@@ -483,7 +486,25 @@ where
                                 .block_sync
                                 .handle_request_block_sync_message(author, msg),
                             ConsensusMessage::BlockSync(msg) => {
-                                self.consensus.handle_block_sync_message(msg)
+                                match self.block_sync.handle_block_sync_message(
+                                    author,
+                                    msg,
+                                    &self.validator_set,
+                                ) {
+                                    BlockRetrievalResult::Failed((peer, message)) => {
```
Yimaha, line 494: ugly, patch in test_case pr directly return ConsensusCommand

Yimaha:
I think I'm going to refactor this pr a bit and make a new pr for it.

right now the biggest anit-pattern is having 2 places to keep track of requested block, I don't like that.

I plan to move the tracking logic to exclusively within the block-sync-manager, as suppose to letting it hang in consensus state. Earlier thought was to associated this with timeout but now looking at it, way better to keep it within one module.

PR #261, opened 2023-09-15
Clean up using automatic clippy fixes

Yimaha:
clippy --fix with all flag used in github action


omegablitz:
I think we don't actually want to fix some of these lints, eg converting unused variable names like `var` to `_var`. In a lot of these cases the warnings genuinely make sense. Whenever we properly fix all of these outstanding warnings, we can change CI to error instead of warning.

Yimaha:
sure, given this case then we will limit it to only the deserialize error? @omegablitz

PR #264, opened 2023-09-18
Track in-flight block-sync

Yimaha:
When request block sync is triggered, it lacks
ability to keep track of blocks that are already
in flight and which validator is currently
processing corresponding block sync request.

such system can cause broadcast storm if given
frequent trigger condition (such as observing qc), In order to avoid such case, now

1. BlockSyncProcess keeps track of in flight requests for block sync, (both who are we
asking and block id that we are asking)

2. Nodes are capable to answer block sync request in 2 different ways: I have block,
or block currently not available

3. requested_block is being changed to a proper class

4. block sync also relies on qc instead of blockId (4 probably belong in the next pr, but they are plan to be merged right after each other)

More test cases supporting new feature added from a follow up pr monad-crypto/monad-bft/pull/265

### DIFF
--- a/monad-block-sync/src/lib.rs
+++ b/monad-block-sync/src/lib.rs
@@ -1,16 +1,10 @@
-use monad_consensus::messages::{
-    consensus_message::ConsensusMessage, message::RequestBlockSyncMessage,
-};
+use monad_consensus::messages::message::RequestBlockSyncMessage;
 use monad_consensus_state::command::{ConsensusCommand, FetchedBlock};
 use monad_consensus_types::signature_collection::SignatureCollection;
-use monad_executor::{PeerId, RouterTarget};
-use monad_types::{BlockId, NodeId};
+use monad_types::NodeId;
 use monad_validator::validator_set::ValidatorSetType;
-
 #[derive(Debug)]
-pub struct BlockSyncState {
-    round_robin_validator_selector: usize,
-}
+pub struct BlockSyncState {}
 
 pub trait BlockSyncProcess<SCT, VT>
 where
@@ -18,12 +12,6 @@ where
     VT: ValidatorSetType,
 {
     fn new() -> Self;
-
-    fn request_block_sync(
-        &mut self,
-        blockid: BlockId,
-        validators: &VT,
-    ) -> (RouterTarget, ConsensusMessage<SCT>);
 
     fn handle_request_block_sync_message(
         &mut self,
@@ -38,23 +26,7 @@ where
     VT: ValidatorSetType,
 {
     fn new() -> Self {
-        BlockSyncState {
-            round_robin_validator_selector: 0,
-        }
-    }
-
-    fn request_block_sync(
-        &mut self,
-        blockid: BlockId,
-        validators: &VT,
-    ) -> (RouterTarget, ConsensusMessage<SCT>) {
-        let target = RouterTarget::PointToPoint(PeerId(
-            validators.get_list()[self.round_robin_validator_selector % validators.len()].0,
-        ));
-        self.round_robin_validator_selector += 1;
-        let message =
-            ConsensusMessage::RequestBlockSync(RequestBlockSyncMessage { block_id: blockid });
-        (target, message)
+        BlockSyncState {}
     }
 
     fn handle_request_block_sync_message(
@@ -66,6 +38,7 @@ where
             s.block_id,
             Box::new(move |block| FetchedBlock {
                 requester: author,
+                block_id: s.block_id,
                 block,
             }),
         )]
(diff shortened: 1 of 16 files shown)


monad-block-sync/src/lib.rs
```diff
@@ -42,23 +30,7 @@ where
...
-        validators: &VT,
-    ) -> (RouterTarget, ConsensusMessage<ST, SC>) {
-        let target = RouterTarget::PointToPoint(PeerId(
-            validators.get_list()[self.round_robin_validator_selector % validators.len()].0,
-        ));
-        self.round_robin_validator_selector += 1;
-        let message =
-            ConsensusMessage::RequestBlockSync(RequestBlockSyncMessage { block_id: blockid });
-        (target, message)
+        BlockSyncState {}
```
Yimaha, line 29: Proposer selection functionality is moved completely to the block_sync_manger within consensus

monad-blocktree/src/blocktree.rs
```diff
@@ -258,11 +259,25 @@ impl<T: SignatureCollection> BlockTree<T> {
         }
     }
 
-    pub fn get_missing_ancestor(&self, b: &BlockId) -> Option<BlockId> {
-        let mut bid = b;
+    pub fn get_missing_ancestor(&self, qc: &QuorumCertificate<T>) -> Option<QuorumCertificate<T>> {
```
Yimaha, line 262: get_missing_ancestor now require a qc, as suppose to arbitrary blockId to enforce every request to be backed by a qc in future

monad-consensus-state/src/manager.rs
```diff
@@ -0,0 +1,125 @@
...
+use crate::command::ConsensusCommand;
+
+const DEFAULT_AUTHOR_INDEX: usize = 0;
+const DEFAULT_RETRY: usize = 0;
+
+#[derive(Debug, Clone)]
+pub struct InFlightBlockSync<SCT> {
+    pub req_target: NodeId,
+    pub retry_cnt: usize,
+    pub qc: QuorumCertificate<SCT>, // qc responsible for this event
```
Yimaha: qc usage in next pr, depending on impl might remove in future

monad-consensus/src/validation/signing.rs
```diff
@@ -336,12 +336,12 @@ where
             &self.author_signature,
         )?;
 
-        verify_certificates::<S, H, _, _>(
-            validators,
-            validator_mapping,
-            &(None),
-            &self.obj.block.qc,
-        )?;
+        if let Some(b) = &self.obj.block {
```
Yimaha: the check to ensure the block returned is indeed in match with block_id of msg

monad-executor/src/state.rs
```diff
@@ -16,6 +16,12 @@ impl std::fmt::Debug for PeerId {
     }
 }
 
+impl From<&NodeId> for PeerId {
```
Yimaha, line 19: helper, thinking to add the reverse as well?

monad-blocktree/src/blocktree.rs
```diff
@@ -1575,30 +1598,30 @@ mod test {
...
-        assert!(blocktree.get_missing_ancestor(&b1.get_id()).unwrap() == b1.get_id());
+        assert!(blocktree.get_missing_ancestor(&g.qc).is_none()); // root naturally don't have missing ancestor
 
         assert!(blocktree.add(b2.clone()).is_ok());
-        assert!(blocktree.get_missing_ancestor(&b2.get_id()).unwrap() == b1.get_id());
+        assert!(blocktree.get_missing_ancestor(&b2.qc).unwrap() == b2.qc);
 
         assert!(blocktree.add(b3.clone()).is_ok());
-        assert!(blocktree.get_missing_ancestor(&b3.get_id()).unwrap() == b1.get_id());
+        assert!(blocktree.get_missing_ancestor(&b3.qc).unwrap() == b2.qc);
```
michael-yxchen: nit: `assert!(blocktree.get_missing_ancestor(&b3.qc).unwrap() == b3.qc)` is probably more intuitive

monad-consensus-state/src/lib.rs
```diff
@@ -262,12 +273,9 @@ where
...
             .add(p.block.clone())
             .expect("Failed to add block to blocktree");
-        if let Some(bid) = self
-            .pending_block_tree
-            .get_missing_ancestor(&p.block.get_id())
-        {
+        if let Some(bid) = self.pending_block_tree.get_missing_ancestor(&p.block.qc) {
             debug!("Block sync request: blockid={:?}", bid);
-            cmds.push(self.create_request_sync(bid));
+            cmds.extend(self.request_sync::<VT>(bid, validators));
```
michael-yxchen: nit: bid => qc

monad-consensus-state/src/lib.rs
```diff
@@ -397,39 +405,53 @@ where
...
-        b: BlockSyncMessage<SCT>,
+        author: NodeId,
+        msg: BlockSyncMessage<SCT>,
+        validators: &VT,
     ) -> Vec<ConsensusCommand<ST, SCT>> {
         let mut cmds = vec![];
-        let block = b.block;
-        let bid = block.get_id();
-        debug!("Block sync response: block={:?}", block);
+        // debug!("Block sync response: bid={:?}", bid);
```
michael-yxchen: un-comment the debug log?

monad-executor/src/state.rs
ariqchowdhury: https://doc.rust-lang.org/rust-by-example/conversion/from_into.html

monad-consensus-state/src/lib.rs
ariqchowdhury: move this to success case

monad-consensus/src/messages/message.rs
```diff
@@ -88,11 +88,15 @@ impl Hashable for RequestBlockSyncMessage {
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct BlockSyncMessage<T> {
-    pub block: Block<T>,
+    pub block_id: BlockId,
+    pub block: Option<Block<T>>,
```
ariqchowdhury: are you using Option here to signify the "block is not available" response? Can this be an enum like
`pub enum BlockSyncMessage<T> {
   HaveBlock(Block<T>),
   BlockNotAvailable(BlockId),
}
`
so we don't have redundant blockid and don't have to check that it matches when verifying

monad-blocktree/src/blocktree.rs
```diff
@@ -258,11 +259,25 @@ impl<T: SignatureCollection> BlockTree<T> {
         }
     }
 
-    pub fn get_missing_ancestor(&self, b: &BlockId) -> Option<BlockId> {
-        let mut bid = b;
+    pub fn get_missing_ancestor(&self, qc: &QuorumCertificate<T>) -> Option<QuorumCertificate<T>> {
+        match self.root {
+            RootKind::Rooted(b) => {
+                if self.tree.get(&b).unwrap().get_block().round >= qc.info.vote.round {
```
ariqchowdhury: should change unwrap() to expect() with msg

monad-consensus-state/src/manager.rs
ariqchowdhury, line 1: nit: I'm not sure about the name 'manager' for this file...probably just call it blocksync

monad-consensus-state/src/manager.rs
```diff
@@ -0,0 +1,125 @@
...
+        &mut self,
+        qc: &QuorumCertificate<SCT>,
+        validator_set: &VT,
+    ) -> Vec<ConsensusCommand<ST, SCT>> {
+        let id = &qc.info.vote.id;
+        match self.requests.entry(*id) {
+            Entry::Occupied(_) => vec![],
+            Entry::Vacant(entry) => {
+                inc_count!(block_sync_request);
+                let peer = validator_set.get_list()[DEFAULT_AUTHOR_INDEX];
```
michael-yxchen: nit: mod `DEFAULT_AUTHOR_INDEX` by `validator_set.len()` so we don't go out of bound

monad-consensus-state/src/manager.rs
```diff
@@ -0,0 +1,125 @@
...
+                None => {
+                    // block retrieve failed, re-request
+                    *retry_cnt += 1;
+
+                    *req_target = validator_set.get_list()[(*retry_cnt) % validator_set.len()];
+                    BlockSyncResult::Failed(ConsensusCommand::RequestSync {
+                        peer: *req_target,
+                        block_id: bid,
+                    })
+                }
```
michael-yxchen: when would it stop retrying?

monad-consensus-types/src/validation.rs
```diff
@@ -17,6 +17,8 @@ pub enum Error {
     InsufficientStake,
     /// Seq num in block proposal must be 1 higher than in the QC
     InvalidSeqNum,
+    /// Block retrieved doesn't match with the block_id requested (block sync)
+    InvalidBlock,
```
michael-yxchen: nit: maybe `InvalidBlockSyncReponse/InvalidBlockSync`? I'm not good with naming things either

monad-proto/proto/event.proto
```diff
@@ -42,7 +42,8 @@ message ProtoFetchedFullTxs {
 
 message ProtoFetchedBlock {
   monad_proto.basic.ProtoNodeId requester = 1;
-  monad_proto.block.ProtoBlock block = 2;
+  monad_proto.basic.ProtoBlockId block_id = 2;
+  monad_proto.block.ProtoBlock block = 3;
```
michael-yxchen: nit: can add an `optional` so it's more explicit

monad-consensus-state/src/manager.rs
Yimaha: The retry never stops until some block is retrieved.

In the next PR, I plan properly introduce a timeout mechanism that retires blocks. with it there will be a way similar to how pending_block_tree prunes the block, in_flight request will also be pruned if we received a block through other means (rn just handle_full_proposal). However I don't think its good to stop retrying? because that means certain block can be permanently lost (once we have state sync, maybe we can add a upper limit to retry attempts?)

michael-yxchen: I see. I thought the `DEFAULT_RETRY` parameter was to specify the maximum number of retires, but nvm.

michael-yxchen:
I'll take a look at #265 and come back to approve both

monad-consensus-state/src/manager.rs
Yimaha: i will rename it to default-retry-seed

_Update: it was deleted in next pr for prevent request to self_

PR #265, opened 2023-09-18
Add Test case and style fix for Inflight blocksync

Yimaha:
supplement for monad-crypto/monad-bft/pull/264

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,7 @@
 [workspace]
 
 members = [
+    "monad-block-sync",
     "monad-blocktree",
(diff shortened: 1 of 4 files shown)


Yimaha:
> lgtm
> 
> just want to note that you can use `assert!(matches!(_))` in `test_handle_retrieval` as an alternative to the `let Enum(_) = _ else {_}`. Up to you do decide which version is more readable.

ooo cool, I didn't know rust have this syntax.
In this test case sometime the output is re-iterated for future usage, but in case where output is not relevant except type match I guess its good alternative :3

PR #266, opened 2023-09-18
QC_based LeaderElection

Yimaha:
New leader election become mutable, thus changing the interface

1.  move the leader election related content to a new module, called monad-election
2.  Deleted current impl of weighted_round_robin, (plan is use weighted random choose, but maybe Interleave WRR? either way it will be patched later)
3.  Changed Monad-Config to contain a leader election module, the creation of election now depends on a lot of parameters, and a universal new() doesn't work well. we should have a enum based system for initialization in future.

This pr aims to solve https://github.com/monad-crypto/monad-bft/issues/287

### DIFF
--- a/monad-election/src/qc_reputation.rs
+++ b/monad-election/src/qc_reputation.rs
@@ -1,0 +1,1015 @@
+use std::collections::{btree_map::Entry, BTreeMap, HashSet};
+
+use monad_consensus_types::{
+    block::FullBlock,
+    quorum_certificate::QuorumCertificate,
+    signature_collection::{SignatureCollection, SignatureCollectionKeyPairType},
+    validation::Hasher,
+    voting::ValidatorMapping,
+};
+use monad_types::{NodeId, Round, Stake};
+use monad_validator::validator_set::ValidatorSetType;
+use rand::distributions::{Distribution, WeightedIndex};
+use rand_chacha::{rand_core::SeedableRng, ChaChaRng};
+
+use super::leader_election::{ElectionInfo, LeaderElection};
+
+struct QCReputationCache {
+    // TODO: convert to bit map
+    participants: HashSet<NodeId>,
+    parent_round: Round,
+    author: NodeId,
+}
+
+pub struct QCReputation {
+    // consist of Block, and a potentially cached set of participants.
+    qc_cache: BTreeMap<Round, QCReputationCache>,
+    leader_cache: BTreeMap<Round, Option<NodeId>>,
+    total_stake: Stake,
+
+    // MUST requirements
+    /// multiple of (f) leader that should be banned.
+    /// for example, 1.0 means minimum of f out of 3f+1
+    /// stake worth of leader must be banned
+    leader_ban_stake: i64,
+    // amount of block we should process in order to select recent participants
+    window_size: u64,
+    // amount of round to delay before leader election comes in effect.
+    leader_delay_round: Round,
+
+    // OPTIONAL requirements
+    /// multiple of (f) candidate needed to form a reputable candidate
+    /// for example, 1.0 means minimum of f out of 3f+1
+    /// stake worth of candidate must exists to trigger QC based leader election
+    min_candidate_stake: Option<i64>,
+    // maximum amount of round before observed qc is no longer valid.
+    max_observation_round: Option<Round>,
+}
+enum QCElection {
+    CanTry,  // MUST req are satisfied
+    Unknown, // leader unknown due to lack of info
+    Failed,  // MUST req are not satisfied
+}
+
+// record the environment for qc election
+struct QCElectionEnv {
+    observed_qc_round: Round,
+    // TODO: convert to bit map
+    banned_leader: HashSet<NodeId>,
+    pre_ban_candidates: HashSet<NodeId>,
+}
+
+impl QCReputation {
+    pub fn new<SCT, VT>(
+        leader_ban_stake: i64,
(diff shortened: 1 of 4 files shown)


monad-validator/src/leader_election.rs
```diff
@@ -1,7 +1,25 @@
...
+    validation::Hasher,
+    voting::ValidatorMapping,
+};
 use monad_types::{NodeId, Round};
 
-// VotingPower is i64
-pub trait LeaderElection {
+pub enum ElectionInfo<'a, SCT: SignatureCollection> {
+    Commits(Vec<Block<SCT>>),
+    QC(
```
Yimaha: the interface for submitting qc typically is linked with the retrieval of multiple leaders participants chained together, thus need to provide the hash function and the validator mapping

monad-validator/src/qc_reputation.rs
```diff
@@ -0,0 +1,94 @@
...
+    voting::ValidatorMapping,
+};
+use monad_types::{NodeId, Round};
+
+use super::leader_election::{ElectionInfo, LeaderElection};
+
+// used for both block cache and leader cache
+const QC_REP_CACHE_ROUND: u64 = 64;
+pub struct QCReputation<SCT> {
+    block_cache: BTreeMap<Round, Block<SCT>>,
```
Yimaha: In future the block<SCT> would be upgraded to a special struct that also allow caching of participants

monad-validator/src/leader_election.rs
```diff
@@ -1,7 +1,25 @@
...
+        &'a ValidatorMapping<SignatureCollectionKeyPairType<SCT>>,
+    ),
+}
+pub trait LeaderElection<SCT>
+where
+    SCT: SignatureCollection,
+{
     fn new() -> Self;
+    fn submit_election_info<H: Hasher>(&mut self, info: ElectionInfo<SCT>);
     fn get_leader(&self, round: Round, validator_list: &[NodeId]) -> NodeId;
```
Yimaha: next pr will allow leader election to return Option\<NodeId\> as suppose to NodeId.

ariqchowdhury:
If we're removing weighted round robin, will there be a way to run the reputation leader as a plain weighted rr (with the initial stake weights and no updates to the reputation)?

Yimaha:
> ariqchowdhury

I believe the default would run some form of Interchange weighted round robin

The reason why this file is deleted is because I want to rework the WRR s.t it that tries to not have repeated leader between rounds, rn the approach is priority based which can still trigger repeated leader (though in theory if any of the validator has more than 50% of stake within the current validator set, then there is no way to avoid consecutive leader, not sure if 3f+1 is always enforced here)

A question I have is how relevant is the repeated leader part? current approach can run into repeated leader due to how it updates priority, but if that is not that relevant I could just provide a simple, weighted randomized selection function with fixed seed from genesis

Yimaha:
merged in the refactor and impl in 2 Separate commit, can be viewed separately.

https://github.com/monad-crypto/monad-bft/pull/286

monad-election/src/leader_election.rs
```diff
@@ -0,0 +1,34 @@
...
+    block::FullBlock,
+    quorum_certificate::QuorumCertificate,
+    signature_collection::{SignatureCollection, SignatureCollectionKeyPairType},
+    validation::Hasher,
+    voting::ValidatorMapping,
+};
+use monad_types::{NodeId, Round};
+use monad_validator::validator_set::ValidatorSetType;
+
+pub enum ElectionInfo<'a, SCT, VT>
```
Yimaha, line 11: enum is only used as the medium to carry information, compiler can prove that enum is destroyed before any of the reference is destroyed, thus lifetime is valid for compiler

monad-validator/src/validator_set.rs
```diff
@@ -227,4 +252,28 @@ mod test {
...
+            validators.push((
+                NodeId(keypair.pubkey()),
+                // just a semi-random number
+                Stake((((i + 1) * 123123) as i64) % 552),
+            ));
+        }
+
+        // should be irrelevant, we always sort.
+        let mut rng = thread_rng();
+        validators.shuffle(&mut rng);
```
ariqchowdhury, line 271: lets use seeded rng even though it might not matter now

monad-election/src/qc_reputation.rs
```diff
@@ -0,0 +1,1015 @@
...
+use monad_validator::validator_set::ValidatorSetType;
+use rand::distributions::{Distribution, WeightedIndex};
+use rand_chacha::{rand_core::SeedableRng, ChaChaRng};
+
+use super::leader_election::{ElectionInfo, LeaderElection};
+
+struct QCReputationCache {
+    // TODO: convert to bit map
+    participants: HashSet<NodeId>,
+    parent_round: Round,
```
ariqchowdhury, line 20: can you add a comment to describe this field more

monad-election/src/qc_reputation.rs
```diff
@@ -0,0 +1,1015 @@
...
+    author: NodeId,
+}
+
+pub struct QCReputation {
```
ariqchowdhury, line 24: lets rename: maybe just ReputationSelection?

Yimaha:
The PR is followed by the design doc
https://www.notion.so/monad-labs/Leader-Election-Ver-1-1bd00e733945408cb481bd7188d84fe8?pvs=4

Issue #248, opened 2023-09-11
Proper error handling for signature errors


michael-yxchen:

michael-yxchen:
Signature collection error handling merged by #277

michael-yxchen:
#458

PR #267, opened 2023-09-19
Split `SignatureCollectionError::InvalidSignature`

michael-yxchen:
`SignatureCollection` used to produce the same `InvalidSigantures` error on creation and on verification. The error is a `Vec<S>` with all the invalid signatures.

This design doesn't provide enough context on creation. For example, Node1 can take Node2's vote signature (sig2) and submit it as its vote. When the leader creates the `SignatureCollection`, it returns an error saying `sig2` is invalid but doesn't provide info on who's accountable, because both nodes submits `sig2`.

This PR splits InvalidSigantures error into two: one for creation and the other for verification, so we can get accountability (NodeId) for `InvalidSiganture` on creation.

Related #248

### DIFF
--- a/monad-consensus-types/src/signature_collection.rs
+++ b/monad-consensus-types/src/signature_collection.rs
@@ -18,8 +18,8 @@ pub enum SignatureCollectionError<S> {
     NodeIdNotInMapping(Vec<(NodeId, S)>),
     // only possible for non-deterministic signature
     ConflictingSignatures((NodeId, S, S)),
-    // verification error
-    InvalidSignatures(Vec<S>),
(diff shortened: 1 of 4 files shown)


monad-consensus-types/src/signature_collection.rs
```diff
@@ -144,6 +147,42 @@ mod test {
...
+                    .collect::<HashSet<_>>();
+
+                let test_set = sigs.into_iter().collect::<HashSet<_>>();
+
+                assert_eq!(expected_set, test_set);
+            }
+            _ => unreachable!(),
+        }
+    });
+
```
michael-yxchen, line 185: > Can we add a test for this case?

Done

PR #268, opened 2023-09-19
Store full_txs in BlockTree

abenedito-monad:
Create FullBlock type which contains Block and FullTransactionList so that on LedgerCommit we don't need to re-fetch full_txs from mempool.

### DIFF
--- a/monad-consensus-state/src/blocksync.rs
+++ b/monad-consensus-state/src/blocksync.rs
@@ -2,9 +2,10 @@ use std::collections::{hash_map::Entry, HashMap};
 
 use monad_consensus::messages::message::BlockSyncMessage;
 use monad_consensus_types::{
-    block::{Block, BlockType},
+    block::{BlockType, FullBlock},
     quorum_certificate::QuorumCertificate,
     signature_collection::SignatureCollection,
+    transaction_validator::TransactionValidator,
 };
 use monad_tracing_counter::inc_count;
 use monad_types::{BlockId, NodeId};
@@ -21,7 +22,7 @@ pub struct InFlightBlockSync<SCT> {
     pub qc: QuorumCertificate<SCT>, // qc responsible for this event
 }
 pub enum BlockSyncResult<SCT: SignatureCollection> {
-    Success(Block<SCT>),           // retrieved
+    Success(FullBlock<SCT>),       // retrieved and validated
     Failed(ConsensusCommand<SCT>), // unable to retrieve
     IllegalResponse,               // never requested from this peer or never requested
 }
@@ -88,14 +89,15 @@ where
         }
     }
 
-    pub fn handle_retrieval<VT: ValidatorSetType>(
+    pub fn handle_retrieval<VT: ValidatorSetType, TV: TransactionValidator>(
         &mut self,
         author: &NodeId,
         msg: BlockSyncMessage<SCT>,
         validator_set: &VT,
+        transaction_validator: &TV,
     ) -> BlockSyncResult<SCT> {
         let bid = match &msg {
-            BlockSyncMessage::BlockFound(b) => b.get_id(),
+            BlockSyncMessage::BlockFound(b) => b.block.get_id(),
             BlockSyncMessage::NotAvailable(bid) => *bid,
         };
         if let Entry::Occupied(mut entry) = self.requests.entry(bid) {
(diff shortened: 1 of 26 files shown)


monad-consensus-state/src/command.rs
```diff
@@ -103,5 +103,5 @@ pub struct FetchedFullTxs<ST, SCT> {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct FetchedBlock<SCT> {
     pub requester: NodeId,
-    pub block: Option<Block<SCT>>,
+    pub unverified_full_block: Option<UnverifiedFullBlock<SCT>>,
```
abenedito-monad, line 125: We can make the FetchedBlock return a FullBlock instead of an unverified one since the ledger is the only component that returns FetchedBlocks so we would have previously verified the FullBlock before adding it to the ledger, but this construction is safer and the added overhead is likely small.

monad-consensus-state/src/lib.rs
```diff
@@ -410,7 +422,13 @@ where
         b: BlockSyncMessage<SCT>,
     ) -> Vec<ConsensusCommand<ST, SCT>> {
         let mut cmds = vec![];
-        let block = b.block;
+
+        let Some(block) = FullBlock::from_block(b.block, b.full_txs, &self.transaction_validator)
```
abenedito-monad: This is the key logic change, we try to build a FullBlock by having the TransactionValidator verify the block's payload matches the full_txs. If they don't match we just return.

monad-consensus-state/src/lib.rs
```diff
@@ -410,7 +422,13 @@ where
...
+        else {
+            return cmds;
+        };
+
+        // TODO: should we debug/inc before building fullblock?
```
abenedito-monad: Not sure if we want to debug / inc_count before or after validating the FullBlock

monad-consensus-state/src/lib.rs
```diff
@@ -252,17 +252,17 @@ where
...
         let mut cmds = vec![];
 
-        if !self.transaction_validator.validate(&txns) {
+        if !self.transaction_validator.validate_full_txs(&full_txs) {
             warn!("Transaction validation failed");
-            debug!("failed txns: {:?}", txns);
+            debug!("failed txns: {:?}", full_txs);
             inc_count!(failed_txn_validation);
             return cmds;
         }
```
abenedito-monad: Now that we are building FullBlocks from the Block and full_txs, I think we can remove validate_full_txs and have those validation checks run in the FullBlock construction.

monad-consensus-state/src/lib.rs
```diff
@@ -448,16 +459,26 @@ where
...
             .handle_retrieval(&author, msg, validators)
         {
-            BlockSyncResult::Success(block) => {
-                debug!("Block sync response: bid={:?}", block.get_id());
+            BlockSyncResult::Success(unverified_full_block) => {
+                let Some(full_block) = FullBlock::try_from_unverified(
+                    unverified_full_block,
+                    &self.transaction_validator,
+                ) else {
+                    return cmds;
```
Yimaha: it seems like if the validation failed, then there is no follow up anymore (we just give up on the missing block)

I think a better way is to move the try_from_unverified into the manager and make sure we produce a valid block, incase it doesn't produce a valid Fullblock, emit Failed(ConsensusCommand) which is just retry.

monad-consensus-state/src/blocksync.rs
```diff
@@ -21,9 +21,9 @@ pub struct InFlightBlockSync<SCT> {
     pub qc: QuorumCertificate<SCT>, // qc responsible for this event
 }
 pub enum BlockSyncResult<SCT: SignatureCollection> {
-    Success(Block<SCT>),           // retrieved
-    Failed(ConsensusCommand<SCT>), // unable to retrieve
-    IllegalResponse,               // never requested from this peer or never requested
+    Success(UnverifiedFullBlock<SCT>), // retrieved
```
Yimaha: Should return fullblock here

monad-consensus-state/src/command.rs
```diff
@@ -117,20 +117,20 @@ pub struct FetchedFullTxs<SCT> {
 pub struct FetchedBlock<SCT> {
     pub requester: NodeId,
     pub block_id: BlockId,
-    pub block: Option<Block<SCT>>,
+    pub unverified_full_block: Option<UnverifiedFullBlock<SCT>>,
```
Yimaha, line 125: Nit: Since ledger is local, I am not sure if we suppose to keep that as unverified_full_block or not

can we trust that the returned result from ledger is a always a valid full block?

(this is more a pattern issue, once it reaches the other party it always get converted to unverified full block anyway, but locally feels like it should just be fullblock?)

monad-consensus-state/src/lib.rs
abenedito-monad: Good catch, done!

monad-consensus-state/src/command.rs
abenedito-monad: Yep, this is definitely done intentionally for pattern reasons. FetchedBlocks should only be sent to network peers, so there is no need to have it be a "verified" FullBlock. UnverifiedFullBlock doesn't mean the block is invalid, so this also prevents the block from being used for "verified" reasons. Made a note about this in the code :)

monad-consensus-state/src/blocksync.rs
```diff
@@ -111,22 +113,26 @@ where
             }
 
             match msg {
-                BlockSyncMessage::BlockFound(block) => {
-                    entry.remove_entry();
-                    // block retrieve successful
-                    BlockSyncResult::Success(block)
+                BlockSyncMessage::BlockFound(unverified_full_block) => {
```
Yimaha, line 116: nit: if let syntax probably better?

Yimaha:
LGTM on block sync end

monad-consensus-state/src/blocksync.rs
abenedito-monad: Hehe yeah I did the match intentionally so that if we ever add another enum variant we need to explicitly think about it here.

PR #269, opened 2023-09-19
Refactor to allow use of concrete MonadEvents at Executors

ariqchowdhury:
Executor implementations are moved to monad-updaters (because they send events to state to update it). New crate monad-mock-swarm for mock-swarm stuff and tests (moved from monad-state)
monad-executor-glue is supposed to be the types that are required to allow updaters and state to interact. Basically commands and events. 

new crate names are kind of hit or miss. I think monad-updaters is good. monad-executor-glue is not so good.


monad-executor/src/lib.rs
```diff
@@ -1,10 +1,37 @@
...
 
 pub mod executor;
-pub mod mock_swarm;
 pub mod replay_nodes;
 pub mod timed_event;
-pub mod transformer;
 
-pub mod convert;
+// driver loop
+async fn run<S: State>(
```
omegablitz: This function can be deleted now

PR #271, opened 2023-09-20
Prevent blocksync from selecting self as peer

Yimaha:


monad-consensus-state/src/blocksync.rs
```diff
@@ -12,8 +12,7 @@ use monad_validator::validator_set::ValidatorSetType;
 
 use crate::command::ConsensusCommand;
 
-const DEFAULT_AUTHOR_INDEX: usize = 0;
-const DEFAULT_RETRY: usize = 0;
+const DEFAULT_PEER: usize = 0;
```
ariqchowdhury: DEFAULT_PEER_INDEX ?

PR #272, opened 2023-09-21
benches that output virtual time ticks

ariqchowdhury:


monad-executor/src/mock_swarm.rs
```diff
@@ -273,7 +282,8 @@ where
         None
     }
 
-    pub fn batch_step_until(&mut self, until: Duration) {
+    pub fn batch_step_until(&mut self, until: Duration, until_block: usize) -> Duration {
+        let mut total_duration = Duration::from_nanos(0);
```
omegablitz: Maybe `end_tick` or something is a better name? Because it's not actually the elapsed virtual time.

michael-yxchen:
The PR looks great.

We should have the runner accept the `--output-format bencher` argument passed in by the benchmark CI. 

Right now it's failing with
```
 Running unittests src/lib.rs (target/release/deps/monad_virtual_bench-f79cbe9946a52201)
error: Unrecognized option: 'output-format'
error: bench failed, to rerun pass `-p monad-virtual-bench --lib`
```

monad-executor/src/mock_swarm.rs
ariqchowdhury: makes sense

PR #273, opened 2023-09-21
Allow Lazy evaluation of Hash on get_participants

Yimaha:
BLS signature doesn't rely on having the original
msg, thus, instead of passing in the Hash, a
closure that can be used to evaluate to
hash is passed.

We might need more bench mark to see if the hash
is actually lazily evaluated or not, and if this
approach is actually an improvement

### DIFF
--- a/monad-consensus-types/src/bls.rs
+++ b/monad-consensus-types/src/bls.rs
@@ -226,10 +226,10 @@ impl SignatureCollection for BlsSignatureCollection {
         Ok(signers)
     }
 
-    fn get_participants(
+    fn get_participants<F: Fn() -> Hash>(
         &self,
         validator_mapping: &ValidatorMapping<SignatureCollectionKeyPairType<Self>>,
-        _msg: &[u8],
+        _: F,
     ) -> HashSet<NodeId> {
         assert_eq!(self.signers.len(), validator_mapping.map.len());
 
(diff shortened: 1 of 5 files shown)


omegablitz:
I think this is unnecessary, it adds noise to the types. If it turns out that eagerly/repeatedly computing hashes is expensive (which we should confirm with benchmarks), we should create a new type that lazily computes the hash of any given message, at most once.

Example pseudocode:
```rust
struct LazyHash<H: Hashable>(H, Option<Hash>);

impl<H: Hashable> LazyHash<H> {
    fn new(msg: H) -> Self {
        Self(msg, None)
    }
    fn fetch(&mut self) -> Hash {
        // if hash doesn't exist, compute and set hash
        // then return hash
    }
}
```

Yimaha:
> I think this is unnecessary, it adds noise to the types. If it turns out that eagerly/repeatedly computing hashes is expensive (which we should confirm with benchmarks), we should create a new type that lazily computes the hash of any given message, at most once.
> 
> Example pseudocode:
> 
> ```rust
> struct LazyHash<H: Hashable>(H, Option<Hash>);
> 
> impl<H: Hashable> LazyHash<H> {
>     fn new(msg: H) -> Self {
>         Self(msg, None)
>     }
>     fn fetch(&mut self) -> Hash {
>         // if hash doesn't exist, compute and set hash
>         // then return hash
>     }
> }
> ```

make sense, closing this

PR #276, opened 2023-09-21
Add Gossip tests

omegablitz:
Adds some basic tests for implementations of the Gossip trait - currently just MockGossip.


monad-quic/src/gossip/mock.rs
```diff
@@ -0,0 +1,171 @@
...
+use monad_executor_glue::{PeerId, RouterTarget};
+
+use super::{Gossip, GossipEvent};
+
+pub struct MockGossipConfig {
+    pub all_peers: Vec<PeerId>,
+}
+
+/// MockGossip is an implementation of Gossip that's not fully spec-compliant - it does not
+/// actually support retry
```
ariqchowdhury: nit: we don't expect retry on gossip anymore right?

monad-quic/src/lib.rs
```diff
@@ -229,23 +229,36 @@ impl<G: Gossip> RouterScheduler for QuicRouterScheduler<G> {
...
+                    self.poll_gossip(gossip_tick);
+                }
             }
         }
         None
     }
 }
 
 enum QuicEventType {
+    GossipTimeout,
```
ariqchowdhury, line 242: Why is this event called GossipTimeout? I might be missing something, but it looks like `gossip.peek_tick` checks if there's an event that needs to be handled in `Gossip` and then later poll_gossip handles the event. I don't get what's timing out.

monad-quic/src/gossip/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+/// `message` and `gossip_message` are both typed as bytes intentionally, because that's the atomic
+/// unit of transfer.
+pub trait Gossip {
+    type Config;
+
+    fn new(config: Self::Config) -> Self;
+    fn send(&mut self, time: Duration, to: RouterTarget, message: &[u8]);
+    fn handle_gossip_message(&mut self, time: Duration, from: PeerId, gossip_message: &[u8]);
+
+    fn peek_tick(&self) -> Option<Duration>;
```
ariqchowdhury: nit: maybe its a little weird that peek_tick is part of the `Gossip` trait when its not useful outside the mock swarm context? But probably easier to just leave it as is.

monad-quic/src/gossip/mod.rs
ariqchowdhury, line 1: does it make sense for Gossip trait to be in monad-quic? It should be independent of quic right?

monad-quic/src/gossip/mock.rs
omegablitz: Yup, good catch

monad-quic/src/gossip/mod.rs
omegablitz: Yeah it should be moved out

monad-quic/src/gossip/mod.rs
omegablitz: It's used in both contexts, peek_tick returns the next time that gossip should be polled at.

monad-quic/src/lib.rs
omegablitz: gossip.peek_tick should always return the tick after which gossip.poll should be called again. This should always be a timeout event, because all non-timeout events should be flushed out earlier via poll_gossip.

PR #278, opened 2023-09-21
Make `Pipeline` implementation more generic

omegablitz:
Implements `Pipeline` for a vector of generic transformers. This is in preparation for a specialized `Vec<Transformer<Vec<u8>>>` implementation.

### DIFF
--- a/monad-executor/src/transformer.rs
+++ b/monad-executor/src/transformer.rs
@@ -20,7 +20,7 @@ pub enum TransformerStream<M> {
     Complete(Vec<StreamMessage<M>>),
 }
 
-pub trait Transform<M> {
+pub trait Transformer<M> {
     #[must_use]
     /// note that the output Duration should be a delay, not an absolute time
     // TODO smallvec? resulting Vec will almost always be len 1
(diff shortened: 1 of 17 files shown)


omegablitz:
These commits can be reviewed independently; the latter one only includes renaming.

monad-executor/src/transformer.rs
```diff
@@ -315,34 +309,27 @@ impl<M> Pipeline<M> for TransformerPipeline<M> {
...
     fn min_external_delay(&self) -> Duration {
-        let delay = self
-            .transformers
-            .iter()
-            .map_while(|transformer| match transformer {
-                Transformer::Latency(LatencyTransformer(delay)) => Some(delay),
-                _ => None,
-            })
-            .sum();
+        let delay = self.iter().map_while(T::min_external_delay).sum();
```
michael-yxchen, line 325: just curious, why do we want to stop after the first `T::min_external_delay` returns `None`?

michael-yxchen: nvm, I think I understand now

PR #281, opened 2023-09-22
Add BytesTransformerPipeline Specialization

omegablitz:
Our current GenericTransformerPipeline contains transformers that are properly generic over any message type. BytesTransformerPipeline is a specialization for message bytes that has additional functionality, eg splitting/merging messages together.

### DIFF
--- a/monad-executor/src/transformer.rs
+++ b/monad-executor/src/transformer.rs
@@ -1,7 +1,12 @@
-use std::{collections::HashSet, fmt::Debug, mem, time::Duration};
+use std::{
+    collections::{BTreeMap, HashSet, VecDeque},
+    fmt::Debug,
+    mem,
+    time::Duration,
+};
 
 use rand::{prelude::SliceRandom, Rng};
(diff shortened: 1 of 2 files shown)


monad-executor/src/transformer.rs
```diff
@@ -262,6 +267,125 @@ impl<M> Transformer<M> for GenericTransformer<M> {
...
+    fn transform(&mut self, message: LinkMessage<Vec<u8>>) -> TransformerStream<Vec<u8>> {
+        let entry = self.buffers.entry(message.to).or_default();
+        entry.push_back(message);
+
+        let split_idx = self.rng.gen_range(0..entry.len());
+
+        let mut base_message = entry
+            .drain(0..split_idx)
+            .reduce(|mut base_message, mut merge_message| {
+                base_message.message.append(&mut merge_message.message);
```
Yimaha, line 303: curious if `from` matters anymore at this point? seems like bytes of different `from` would be merge together

omegablitz: It's not used in the pipeline, but it's used elsewhere

PR #282, opened 2023-09-23
Allow uncommitted block to be synced

Yimaha:
Now that QC backed block sync is in placed,
we can ensure that even uncommitted blocks
cannot be abused to cause fake commit

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -137,6 +137,8 @@ where
 
     fn get_keypair(&self) -> &KeyPair;
 
+    fn fetch_uncommitted_block(&self, bid: &BlockId) -> Option<&FullBlock<SCT>>;
+
     fn request_sync<VT: ValidatorSetType>(
         &mut self,
         qc: QuorumCertificate<SCT>,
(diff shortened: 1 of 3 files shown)


monad-consensus-state/src/command.rs
```diff
@@ -138,3 +138,12 @@ impl<SCT: SignatureCollection> From<FetchedBlock<SCT>> for ConsensusCommand<SCT>
         }
     }
 }
+
+impl<SCT: SignatureCollection> From<(&NodeId, Block<SCT>)> for ConsensusCommand<SCT> {
+    fn from((id, b): (&NodeId, Block<SCT>)) -> Self {
```
omegablitz: I don't think we should implement the `From` trait for this - doesn't make sense for a `(&NodeId, Block<SCT>)` to be coercible to a ConsensusCommand.

monad-consensus-state/src/lib.rs
```diff
@@ -469,6 +471,13 @@ where
         cmds
     }
 
+    fn fetch_uncommitted_block(&self, bid: &BlockId) -> Option<&Block<SCT>> {
+        if let Some(btree_block) = self.pending_block_tree.tree().get(bid) {
```
omegablitz: super nit: `map` is more idiomatic

Yimaha: so like self.pending_block_tree.map? or is it something else?

Yimaha:
Oh wow I did not know deleting the commit would cause the pr to be automatically closed

ariqchowdhury:
add tests

Yimaha:
test added

PR #284, opened 2023-09-23
Allow BlockSyncManager to Request in Round-Robin

Yimaha:
BlockSyncManager used to only request a default
user (except self), now it is able to request in
round-robin to all its neighbour.


ariqchowdhury:
can you update the commit message to be more detailed? Describe the bug being fixed and how does this change fix it.

PR #285, opened 2023-09-23
Enable qc observe driven blocksync

Yimaha:
Now every blocksync is backed by a qc, we are
able to retrieve blocks on all location where
a process_certificate_qc is called

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -262,7 +262,7 @@ where
         inc_count!(fetched_proposal);
         let mut cmds = vec![];
 
-        let process_certificate_cmds = self.process_certificate_qc(&p.block.qc);
+        let process_certificate_cmds = self.process_certificate_qc(&p.block.qc, validators);
         cmds.extend(process_certificate_cmds);
 
         if let Some(last_round_tc) = p.last_round_tc.as_ref() {
@@ -290,13 +290,6 @@ where
         self.pending_block_tree
             .add(full_block.clone())
             .expect("Failed to add block to blocktree");
-        if let Some(qc) = self
-            .pending_block_tree
-            .get_missing_ancestor(&full_block.get_block().qc)
-        {
-            debug!("Block sync request: blockid={:?}", qc);
-            cmds.extend(self.request_sync::<VT>(qc, validators));
-        }
 
         if full_block.get_round() != round || author != leader || full_block.get_author() != leader
         {
@@ -360,7 +353,7 @@ where
         if let Some(qc) = qc {
             debug!("Created QC {:?}", qc);
             inc_count!(created_qc);
-            cmds.extend(self.process_certificate_qc(&qc));
+            cmds.extend(self.process_certificate_qc(&qc, validators));
 
             if self.nodeid
                 == election.get_leader(self.pacemaker.get_current_round(), validators.get_list())
@@ -389,7 +382,7 @@ where
         debug!("Remote timeout msg: {:?}", tm);
         inc_count!(remote_timeout_msg);
 
-        let process_certificate_cmds = self.process_certificate_qc(&tm.tminfo.high_qc);
+        let process_certificate_cmds = self.process_certificate_qc(&tm.tminfo.high_qc, validators);
         cmds.extend(process_certificate_cmds);
 
         if let Some(last_round_tc) = tm.last_round_tc.as_ref() {
@@ -585,14 +578,18 @@ where
     }
 
     #[must_use]
-    fn process_certificate_qc(
+    fn process_certificate_qc<VT: ValidatorSetType>(
         &mut self,
         qc: &QuorumCertificate<SCT>,
+        validators: &VT,
     ) -> Vec<ConsensusCommand<SCT>> {
         let mut cmds = Vec::new();
         cmds.extend(self.process_qc(qc));
 
         cmds.extend(self.pacemaker.advance_round_qc(qc).map(Into::into));
+
+        // block sync
+        cmds.extend(self.get_blocks_if_missing(qc, validators));
         cmds
     }
 
@@ -683,6 +680,18 @@ where
             ConsensusAction::Abstain
         }
     }
+
+    fn get_blocks_if_missing<VT: ValidatorSetType>(
+        &mut self,
+        qc: &QuorumCertificate<SCT>,
+        validators: &VT,
+    ) -> Vec<ConsensusCommand<SCT>> {
+        if let Some(qc) = self.pending_block_tree.get_missing_ancestor(qc) {
+            self.request_sync::<VT>(qc, validators)
+        } else {
+            vec![]
+        }
+    }
 }
 
 pub enum ConsensusAction {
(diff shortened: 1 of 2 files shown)


monad-consensus-state/src/lib.rs
```diff
@@ -287,10 +286,6 @@ where
         self.pending_block_tree
             .add(p.block.clone())
             .expect("Failed to add block to blocktree");
-        if let Some(qc) = self.pending_block_tree.get_missing_ancestor(&p.block.qc) {
-            debug!("Block sync request: blockid={:?}", qc);
```
Yimaha, line 297: remove the old call

monad-consensus-state/src/lib.rs
```diff
@@ -2013,4 +2031,267 @@ mod test {
         });
         assert!(res.is_none());
     }
+
```
Yimaha: test case for all triggering condition

monad-consensus-state/src/lib.rs
```diff
@@ -2013,4 +2031,263 @@ mod test {
...
+    #[test_case(5; "5 participants")]
+    #[test_case(100; "100 participants")]
+    #[test_case(5235; "523 participants")]
```
Yimaha: oops will correct the naming

monad-consensus-state/src/lib.rs
```diff
@@ -576,14 +569,18 @@ where
...
     ) -> Vec<ConsensusCommand<SCT>> {
         let mut cmds = Vec::new();
         cmds.extend(self.process_qc(qc));
 
         cmds.extend(self.pacemaker.advance_round_qc(qc).map(Into::into));
+
+        // block sync
+        cmds.extend(self.get_blocks_if_missing(qc, validators));
         cmds
     }
```
michael-yxchen, line 594: I remember that you had some concerns about sending `RequestSync` here because of potential request storm. Is it ok now because the block sync module is aware of the pending requests?

monad-consensus-state/src/lib.rs
```diff
@@ -1719,11 +1731,9 @@ mod test {
         assert_eq!(p1_votes.vote, p3_votes.vote);
         assert_eq!(p1_votes.vote, p4_votes.vote);
         assert_ne!(p1_votes.vote, p2_votes.vote);
-
         let votes = vec![p1_votes, p2_votes, p3_votes, p4_votes];
-
         // Deliver all the votes to second_state, who is the leader for the next round
-        for i in 0..4 {
+        for i in 0..3 {
```
michael-yxchen: why?

Yimaha: in the last state, you actually formed a qc locally through collecting 2f+1 votes, thus triggered a block sync (mean while all other vote didn't trigger block sync because no qc is formed yet)

I was debaiting if observing a `vote` is enough to trigger a blocksync, the decision I went with is is no, because anyone can sent you a vote for a block that doesn't even exists, thus, a minimum of 2f+1 participants must all vote for something (and thus forming a qc) in order to trigger a blocksync

monad-consensus-state/src/lib.rs
Yimaha: yup, the reason why there are so many pr after #264 is that the proper manager now rejects request to identical blocks and ensure the requesting process is unique. (and thus moving this to QC observe should be fine)

monad-consensus-state/src/lib.rs
```diff
@@ -674,6 +671,19 @@ where
...
         }
     }
+
+    fn get_blocks_if_missing<VT: ValidatorSetType>(
+        &mut self,
+        qc: &QuorumCertificate<SCT>,
+        validators: &VT,
+    ) -> Vec<ConsensusCommand<SCT>> {
+        if let Some(qc) = self.pending_block_tree.get_missing_ancestor(qc) {
+            debug!("Block sync request: blockid={:?}", qc);
```
ariqchowdhury: do we have a counter for block sync requests? should increment

monad-consensus-state/src/lib.rs
```diff
@@ -2013,4 +2031,263 @@ mod test {
         });
         assert!(res.is_none());
     }
+
+    #[test_case(5; "5 participants")]
```
ariqchowdhury, line 2161: nit: 4 is better for min number of nodes

monad-consensus-state/src/lib.rs
michael-yxchen: ok I see what's going on here. `p2_vote` is actually voting for a malicious block, so after the loop, `second_state` has two votes for correct block, and one vote from malicious block, and no QC is formed for either.

Can we update the comment to make this more clear?

monad-consensus-state/src/lib.rs
```diff
@@ -2013,4 +2031,263 @@ mod test {
...
+            &certkeys,
+            &valset,
+            &election,
+            &valmap,
+            Default::default(),
+            ExecutionArtifacts::zero(),
+        );
+        let (author, _, verified_message) = cp.destructure();
+        let mut votes = vec![];
+        for (i, state) in states.iter_mut().enumerate().take(num_state) {
```
michael-yxchen: nit: redundant `take`

monad-consensus-state/src/lib.rs
```diff
@@ -2013,4 +2031,263 @@ mod test {
...
+                &certkeys,
+                &valset,
+                &election,
+                &valmap,
+                Default::default(),
+                ExecutionArtifacts::zero(),
+            );
+
+            let (author, _, verified_message) = cp.destructure();
+            for state in states.iter_mut().take(num_state) {
```
michael-yxchen: nit: redundant `take`

monad-consensus-state/src/lib.rs
```diff
@@ -2013,4 +2031,263 @@ mod test {
...
+                            peer: NodeId(_),
+                            block_id: BlockId(_)
+                        }
+                    )
+                });
+                // observing a qc that link to root should not trigger anything
+                assert!(bsync_reqest.is_none());
+            }
+        }
+
```
michael-yxchen: can we have an assertion on the round number here so it's more clear?

monad-consensus-state/src/lib.rs
```diff
@@ -2013,4 +2031,263 @@ mod test {
...
+                &certkeys,
+                &valset,
+                &election,
+                &valmap,
+                Default::default(),
+                ExecutionArtifacts::zero(),
+            );
+
+            let (author, _, verified_message) = cp.destructure();
+            for state in states.iter_mut().take(num_state).skip(1) {
```
michael-yxchen: same

monad-consensus-state/src/lib.rs
Yimaha: The counter is within block sync manager, namely `block_sync_request`. get_blocks_if_missing in particular will get spammed everytime process_certificate_qc is called, but if the requested qc is already part of the on going request then it will not be triggered.

moving the debug to within the blocksync.rs should make this more obvious

PR #286, opened 2023-09-25
Add QC reputation system for Leader election

Yimaha:
Currently, Leader election simply follows a
round robin schedule, which is not great for
detecting node that are actively participating.

This version of qc_reputation follows diem's
impl, with some extra feature / modification to
eliminate some potential edge cases. In general,
such leader election:
1. Can detect recent participants
2. avoid election of duplicate leader (though
some edge cases is not covered yet, will be added
soon)
3. very difficult to temper with unless
holding f+ stake (which we will have an even
bigger problem)

Its worth noting that this impl is not optimized.
1. you can change the participants to be bitvec
2. you can perform 2 pointer sliding window
but both are considered optimization.

Integration test using QC_rep will be added
in future pr. This pr contains only unit test.

This pr aims to solve https://github.com/monad-crypto/monad-bft/issues/287

### DIFF
--- a/monad-election/src/qc_reputation.rs
+++ b/monad-election/src/qc_reputation.rs
@@ -1,0 +1,1015 @@
+use std::collections::{btree_map::Entry, BTreeMap, HashSet};
+
+use monad_consensus_types::{
+    block::FullBlock,
+    quorum_certificate::QuorumCertificate,
+    signature_collection::{SignatureCollection, SignatureCollectionKeyPairType},
+    validation::Hasher,
+    voting::ValidatorMapping,
+};
+use monad_types::{NodeId, Round, Stake};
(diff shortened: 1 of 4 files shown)


monad-election/src/qc_reputation.rs
```diff
@@ -0,0 +1,999 @@
...
+    CanTry,  // MUST req are satisfied
+    Unknown, // leader unknown due to lack of info
+    Failed,  // MUST req are not satisfied
+}
+
+// record the environment for qc election
+struct QCElectionEnv {
+    observed_qc_round: Round,
+    // TODO: convert to bit map
+    banned_leader: HashSet<NodeId>,
```
Yimaha, line 58: bitvec is significantly better for our usage, but for this pr I think hashset is great for readability and understanding

PR #288, opened 2023-09-25
Remove txs in committed blocks from mempool

abenedito-monad:


.github/workflows/rust.yml
```diff
@@ -39,6 +39,7 @@ jobs:
           -A clippy::type_complexity
           -A clippy::int_plus_one
           -A clippy::uninlined-format-args
+          -A clippy::enum-variant-names
```
ariqchowdhury, line 42: can we keep cleanup changes and those kinds of changes separate from features?

abenedito-monad: Code format checks won't pass without this flag, seemed reasonable enough to add one line to rust.yml. I'll keep this in mind in the future.

PR #289, opened 2023-09-26
Deduplicate transactions in pending blocktree

abenedito-monad:
Provide TransactionList of each pending block in the extending branch to the mempool on FetchTxs so that it can avoid these transactions and return unused transactions.

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -620,12 +620,13 @@ where
         let seq_num_qc = high_qc.info.vote.seq_num;
         let proposed_seq_num = seq_num_qc + 1;
         match self.proposal_policy(&parent_bid, proposed_seq_num) {
-            ConsensusAction::Propose(h) => {
+            ConsensusAction::Propose(h, pending_txs) => {
                 inc_count!(creating_proposal);
                 debug!("Creating Proposal: node_id={:?} round={:?} high_qc={:?}, seq_num={:?}, last_round_tc={:?}", 
                                 node_id, round, high_qc, proposed_seq_num, last_round_tc);
                 vec![ConsensusCommand::FetchTxs(
                     self.config.proposal_size,
+                    pending_txs,
                     Box::new(move |txns| FetchedTxs {
                         node_id,
                         round,
@@ -650,6 +651,7 @@ where
                                 node_id, round, high_qc, proposed_seq_num, last_round_tc);
                 vec![ConsensusCommand::FetchTxs(
                     0,
+                    vec![],
                     Box::new(move |_txns| FetchedTxs {
                         node_id,
                         round,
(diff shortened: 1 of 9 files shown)


monad-consensus-state/src/lib.rs
```diff
@@ -654,30 +656,26 @@ where
...
                 None => ConsensusAction::ProposeEmpty,
-            }
-        } else {
-            ConsensusAction::Abstain
+            };
         }
+
+        if self.config.propose_with_missing_blocks {
+            return ConsensusAction::ProposeEmpty;
+        };
```
ariqchowdhury: the config name is probably confusing, but the intent of `propose_with_missing_blocks` is to allow a node to make a full proposal even if its blocktree is not complete.
The `ProposeEmpty` action is for the case where `get_next_state_root` fails.

monad-consensus-state/src/lib.rs
```diff
@@ -654,30 +656,26 @@ where
...
+
+        ConsensusAction::Abstain
     }
 }
 
 pub enum ConsensusAction {
-    Propose(Hash),
+    Propose(Hash, Vec<TransactionList>),
```
michael-yxchen, line 711: Not related this PR specifically. I think we should start thinking about the actual format/encoding of the `TransactionList`, and document it somewhere.

abenedito-monad: The point of TransactionList being a `Vec<u8>` is for consensus to be completely agnostic as to the underlying type of transactions. The actual format/encoding of the TransactionList data we'll be using in testnet/mainnet will be an RLP encoding of a list of ethereum transaction hashes.

michael-yxchen: oh. I understand the `Vec<Vec<u8>>` in `create_proposal` now. Thanks

monad-consensus-state/src/lib.rs
abenedito-monad: Fixed!

monad-consensus-state/src/lib.rs
```diff
@@ -655,30 +657,30 @@ where
...
+        if let Some(pending_txs) = self.pending_block_tree.get_txs_on_path_to_root(parent_bid) {
+            return ConsensusAction::Propose(h, pending_txs);
         }
+
+        // Still propose but with the chance of proposing duplicate txs
+        if self.config.propose_with_missing_blocks {
+            return ConsensusAction::Propose(h, vec![]);
+        };
+
+        ConsensusAction::Abstain
```
michael-yxchen, line 694: I think it's changing the outcome of one particular case. When we're missing blocks to the root, and `propose_with_missing_blocks == false`, the original code outputs `Abstain` and the PR outputs `ProposeEmpty`.

What's the reason for this change?

ariqchowdhury: andre and I talked about this offline before changing it, I think that it makes sense to propose empty in this case as its similar to the case of missing state_root_hashes. Basically, the node is live but missing "something" that allows it to properly propose, so it should propose empty to keep consensus on the happy path in both cases.

michael-yxchen: Cool, just wanted to make sure it was an intentional change.

That's reasonable, my only concern's that if the proposal comes after a quorum of the votes, then the leader would request block sync and immediately propose an empty block. That's not a huge deal as the network layer should prioritize sending to the next leader. Just wanted to bring it up so we're aware.

The rest of the PR looks good to me.

PR #290, opened 2023-09-27
Add Fuzz Testing for Mock ledger

Yimaha:
Previously, MockLedger lack consideration for the fact that requests for block might not be instant, and order of arrival of executor command is not sequential, this problem is addressed in pr #279.

This commit contains additional unit test to ensure such behaviour is tested, Its worth mentioning that if identical ledgerFetch is observed before the previous one is polled, then the new ledgerFetch is will not be answered. (when given identical request at the same time, only the first request is answered)


monad-executor/src/executor/ledger.rs
```diff
@@ -232,12 +252,126 @@ mod tests {
...
         assert_eq!(mock_ledger_event, None);
     }
+
+    /**
+     *  Fuzz testing on a set of request that is being shuffled into arbitrary order
+     */
+    #[test_case(123123, 12; "test 1")]
+    #[test_case(345345345, 23; "test 2")]
+    #[test_case(346346353, 100; "test 3")]
+    #[test_case(343413452, 200; "test 4")]
```
Yimaha: reduce the size to below 100

Yimaha:
provide a follow up with nightly test

monad-executor/src/executor/ledger.rs
```diff
@@ -232,12 +252,126 @@ mod tests {
...
+    #[test_case(23452345245, 500; "test 5")]
+    #[test_case(342342, 500; "test 6")]
+    #[test_case(346243632, 500; "test 7")]
+    #[test_case(23534366, 500; "test 8")]
+    #[test_case(457643563, 2; "test 9")]
+    #[test_case(965864809, 1; "test 10")]
+    fn test_ledger_of_arbitrary_order(seed: u64, node_req_range: i32) {
```
michael-yxchen: would it be a good idea to move this to the randomized nightly test?

Yimaha: yup, next pr would include a nightly test

Yimaha: actually you know what, I might as well add it in this pr

Yimaha:
nightly test added

PR #291, opened 2023-09-27
Add Blackout TestCases

Yimaha:
More Blackout Test cases with either bigger size of participants and/or longer blackout period is added.

_Note: Linter forces me to put #[test_case] below #[test], the real change is not that much, just mostly test_case update_

### DIFF
--- a/monad-state/tests/block_sync.rs
+++ b/monad-state/tests/block_sync.rs
@@ -1,138 +1,150 @@
-use std::{collections::HashSet, time::Duration};
+mod test {
+    use std::{collections::HashSet, time::Duration};
 
-use monad_block_sync::BlockSyncState;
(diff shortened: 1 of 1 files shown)


monad-state/tests/block_sync.rs
```diff
@@ -1,138 +1,150 @@
...
+    #[test_case(10, Duration::from_secs(0),Duration::from_secs(30),Duration::from_secs(60), 3; "test 6")]
 
-    assert!(num_nodes >= 2, "test requires 2 or more nodes");
+    fn black_out(
+        num_nodes: u16,
+        from: Duration,
+        to: Duration,
+        until: Duration,
+        black_out_cnt: usize,
+    ) {
```
Yimaha, line 96: convert these constant into parameters and check parameter is valid

PR #292, opened 2023-09-27
[PERMA DRAFT] TODO list for tmo for block sync

Yimaha:
This pr will never be converted to ready, simply a tool that help me keep track of things that needed to be done before making a new pr that is ready for review, will close once its ready

Kinda curious if there is a private way to do this, there is no reason why this pr should be public


Yimaha:
TODO: Add test cases for updated mocktimer

Yimaha:
ya you know what its probably better to just push but not making a public pr for it

PR #294, opened 2023-09-28
Add ReplayTest

michael-yxchen:
ReplayTest brings down f+1 nodes from the swarm, and then recover them with the persistence log. The chain lose liveness when f+1 nodes are down because there are enough votes to form a quorum. The chain should resume progress after the nodes are recovered.


michael-yxchen:
~~Wrote the last assertion wrong. The test is actually failing....~~

Fixed. The bug was in the testcase

monad-state/tests/replay_test.rs
ariqchowdhury, line 1: should we verify ledgers for all the nodes at the end of the test? Also, we should check counters to confirm that blocksync isn't being used to catch the nodes up.

michael-yxchen: yeah will do

monad-state/tests/replay_test.rs
```diff
@@ -0,0 +1,270 @@
...
+
+    MockExecutor<S, RS, ME>: Unpin,
+    S::Event: Unpin,
+    S::Block: Unpin,
+    Node<S, RS, P, LGR, ME>: Send,
+    RS::Serialized: Send,
+{
+    let mut max_tick = start_tick;
+
+    loop {
```
omegablitz: nit: `while let Some((tick, _, _)) = nodes. ...` is less verbose

monad-state/tests/replay_test.rs
ariqchowdhury, line 1: discussed offline, adding so we don't forget, check different leaders coming out of the outage

monad-state/src/lib.rs
```diff
@@ -549,4 +558,9 @@ where
             }
         }
     }
+    #[cfg(feature = "test_feature")]
+
+    fn consensus(&self) -> &Self::ConsensusState {
+        &self.consensus
+    }
```
michael-yxchen: Added this function to fetch the consensus state to compare. "test_feature" is added so we can conditionally compile this function when needed in tests. The cleanest way would be using `#[cfg(test)]` instead of the feature but that doesn't export to other crates. Had to use the feature as an workaround.

Related https://github.com/rust-lang/cargo/issues/8379

monad-state/src/lib.rs
```diff
@@ -213,9 +213,16 @@ pub struct MonadConfig<SCT: SignatureCollection, TV> {
...
-impl<CT, ST, SCT, VT, LT, BST> State for MonadState<CT, ST, SCT, VT, LT, BST>
+impl<
+        #[cfg(feature = "test_feature")] CT: ConsensusProcess<SCT> + Eq,
+        #[cfg(not(feature = "test_feature"))] CT: ConsensusProcess<SCT>,
+        ST,
+        SCT,
+        VT,
+        LT,
+        BST,
+    > State for MonadState<CT, ST, SCT, VT, LT, BST>
```
michael-yxchen: Somehow that conditional constraint only works in the `impl<>`, not in the where clause.

monad-consensus-state/Cargo.toml
```diff
@@ -8,6 +8,10 @@ edition = "2021"
 [lib]
 bench = false
 
+
+[features]
+test_feature = []
```
ariqchowdhury: nit: maybe rename to monad_test? To make it clear that the feature is for things required only in monad integration tests?

monad-consensus-state/src/lib.rs
```diff
@@ -58,6 +58,46 @@ pub struct ConsensusState<SCT: SignatureCollection, TV, SVT> {
...
+            .field("config", &self.config)
+            .field("block_sync_manager", &self.block_sync_manager)
+            .finish()
+    }
+}
+
+#[cfg(feature = "test_feature")]
+impl<SCT, TVT, SVT> Eq for ConsensusState<SCT, TVT, SVT> where SCT: SignatureCollection {}
+
+#[cfg_attr(feature = "test_feature", derive(PartialEq, Eq, Debug))]
```
ariqchowdhury: nit: its probably reasonable for this to derive debug for release, it should probably be logged or something

monad-mock-swarm/src/mock_swarm.rs
```diff
@@ -256,13 +238,14 @@ where
...
-                self.states
-                    .get_mut(&message.to)
-                    .expect("logic error, should be nonempty")
-                    .pending_inbound_messages
-                    .push(Reverse((sched_tick, message)));
+                // deliver message if node is still in the swarm
+                self.states.get_mut(&message.to).map(|node| {
+                    node.pending_inbound_messages
+                        .push(Reverse((sched_tick, message)))
+                });
```
ariqchowdhury, line 246: Just wanted to clarify, this change was required because we're sending messages to nodes that are "down"?

monad-mock-swarm/src/mock_swarm.rs
```diff
@@ -305,6 +288,7 @@ where
                 while let Some((_tick, _event)) = node.step_until(tick, &mut emitted) {}
                 emitted.into_iter()
             }));
+            self.tick = tick;
```
ariqchowdhury, line 289: what's `self.tick` used for?

monad-mock-swarm/tests/replay_test.rs
```diff
@@ -0,0 +1,298 @@
...
+use monad_testutil::swarm::{get_configs, node_ledger_verification};
+use monad_types::{Deserializable, Serializable};
+use monad_validator::{simple_round_robin::SimpleRoundRobin, validator_set::ValidatorSet};
+use monad_wal::{
+    mock::{MockMemLogger, MockMemLoggerConfig},
+    PersistenceLogger,
+};
+use tracing_test::traced_test;
+
+const DELTA: Duration = Duration::from_millis(10);
```
ariqchowdhury: nit: I think we should get into the habit of calling this consensus_delta for clarity

monad-mock-swarm/src/mock_swarm.rs
michael-yxchen: Yes. The `RouterScheduler` is not aware that some nodes are taken down, so it still emit send events to them.

monad-mock-swarm/src/mock_swarm.rs
michael-yxchen: Our simulation assumes `events.tick` is always monotonically increasing, as it resembles the global clock.

`self.tick` marks the global tick. When nodes are added to the swarm via `add_state`, their executor accepts `self.tick` as the current time. That way we have a consistent time line.

PR #299, opened 2023-10-02
Add TimeoutVariant and BlockSync Timeout

Yimaha:
In order to allow timeout in block sync system, we needed a way to correctly communicate timeout condition between monad-state and executors.

Previously, The TimerCommand is only responsible for setting a fixed timer for pacemaker timeout. we changed the logic such that Timeout out associated itself with a TimeoutVariant, which can be used to identify what type of Timeout are we dealing with and respond accordingly

Tokio timer has a built in sorting system through JointSet, so its naturally optimized for time complexity. For the mocks (mock_executor + mock_timer) I used a priority_queue for the 2 mock case.

Proto also changed to accommodate the new format


monad-consensus-state/src/lib.rs
```diff
@@ -178,23 +181,21 @@ where
             config,
 
             transaction_validator,
-            block_sync_manager: BlockSyncManager::new(NodeId(my_pubkey)),
+            // Q: is delta * 4 a good duration? need to check what is a good value for constant
```
Yimaha: what's a good timeout window?

monad-consensus-state/src/blocksync.rs
```diff
@@ -131,25 +142,59 @@ where
...
             BlockSyncResult::IllegalResponse
         }
     }
+
+    pub fn handle_timeout<VT: ValidatorSetType>(
+        &mut self,
+        bid: BlockId,
+        validator_set: &VT,
+    ) -> Vec<ConsensusCommand<SCT>> {
+        let mut cmds = vec![ConsensusCommand::ScheduleReset(
```
Yimaha: For the sake of design pattern, I emit the reset as well.

To be fair reset doesn't need to be emitted if you decided to re-timeout

monad-updaters/src/timer.rs
```diff
@@ -1,45 +1,104 @@
...
                     duration,
+                    variant,
                     on_timeout,
                 } => {
                     wake = true;
-                    Some((Box::pin(tokio::time::sleep(duration)), on_timeout))
+                    let handle = self
+                        .timers
+                        .spawn(TokioSleeper::<E>::new(duration, on_timeout));
+                    match self.aborts.entry(variant) {
```
Yimaha, line 48: when you schedule or schedule reset, you abort the process that was timing the previous clock if such exists.

such handle is generated when we first spawn the TokioSleeper

Yimaha:
Ready for review, the concept is just change timer so it can track any type of event (that you need to declare in TimoeoutVariant) follow by all the protocol change and where to emit different timeouts

PR #301, opened 2023-10-02
Consensus state tests for state root validator

ariqchowdhury:


monad-testutil/src/proposal.rs
michael-yxchen, line 1: Not related to this PR: the `seq_num` might be wrong for this `ProposalGen` util, because calling `next_tc` actually kills the proposal in the same round.

The more correct way of assigning `seq_num` would be
```
&Payload {
                txns,
                header: execution_header,
                seq_num: qc.info.vote.seq_num + 1,
            },
```
and we can remove `seq_num` from `ProposalGen`

monad-consensus-state/src/lib.rs
```diff
@@ -2012,6 +2012,231 @@ mod test {
...
+    #[traced_test]
+    #[test]
+    fn test_lagging_execution() {
+        let (keys, certkeys, valset, valmap, mut states) =
+            setup::<SignatureCollectionType, StateRoot, TransactionValidatorType>(4);
+        let election = SimpleRoundRobin::new();
+        let (state, _) = states.split_first_mut().unwrap();
+
+        let mut proposal_gen = ProposalGen::<SignatureType, _>::new(state.high_qc.clone());
+        proposal_gen.set_seq_num(2);
```
michael-yxchen: If we remove `seq_num` from `ProposalGen`, the alternative ways to achieve this is modifying the `ProposalMessage` directly.

monad-consensus-state/src/lib.rs
```diff
@@ -2012,6 +2012,231 @@ mod test {
...
+                .count()
+            {
+                1 => Ok(()),
+                n => Err(format!("Expected count 1 got {}", n)),
+            }
+        });
+    }
+
+    #[test]
+    fn test_state_root_updates() {
```
michael-yxchen, line 2142: I think the `seq_num`/`state_root round` would be slight offseted if we use `block.seq_num = qc.seq_num + 1`. 

The genesis block has `seq==0`, then `p0.seq_num==1`, etc.

state still needs `state_root_update` for `seq=0` the genesis block

monad-consensus-state/src/lib.rs
```diff
@@ -2012,6 +2012,231 @@ mod test {
...
+                state_root: Hash([0xaa; 32]),
+                transactions_root: Default::default(),
+                receipts_root: Default::default(),
+                logs_bloom: Bloom::zero(),
+                gas_used: Gas(0),
+            },
+        );
+
+        let (author, _, verified_message) = p2.destructure();
+        state.handle_state_root_update(1, Hash([0xaa; 32]));
```
michael-yxchen: nit: Shall we use a different `state_root` value for every block?

PR #302, opened 2023-10-02
Justin/block sync timeout integration

Yimaha:
In order to provide functionality for block sync timeout test, we wanted to have some form of OR/AND logic and slightly more separated test utilities, thus this commit also contains some extra changes to the transformer system which we have been running.

This is a supplement pr for #299, should be reviewed after it


Yimaha:
should not be merging to master

PR #303, opened 2023-10-02
Add Integration testing for block sync timeout

Yimaha:
In order to provide functionality for block sync timeout test, we wanted to have some form of OR/AND logic and slightly more separated test utilities, thus this commit also contains some extra changes to the transformer system which we have been running.

This is a supplement for #299, thus should be reviewed after


Yimaha:
This need relook (and divide into smaller section) after rebasing after the main code #299 is merged, closing right now

Yimaha:
now on #356

PR #304, opened 2023-10-02
Change Mempool updaters to use MonadEvent type

ariqchowdhury:


monad-executor-glue/src/lib.rs
```diff
@@ -106,20 +110,20 @@ pub enum StateRootHashCommand<B> {
...
 
-pub enum Command<M, OM, B, C>
+pub enum Command<M, OM, B, C, S>
 where
     M: Message,
 {
     RouterCommand(RouterCommand<M, OM>),
     TimerCommand(TimerCommand<M::Event>),
 
-    MempoolCommand(MempoolCommand<M::Event>),
```
omegablitz, line 116: nit: `MempoolCommand<SCT>`? When first reading this I got confused and thought it might mean `State`

ariqchowdhury: good point, I'll clean that up

PR #306, opened 2023-10-03
Allow fair&random selection of event in mock_swarm

Yimaha:
This PR for "mock_swarm" improves event selection by eliminating a fixed
preference for executor events or external messages when they occur
simultaneously at the same tick. Instead, it enables fair and random
selection between these two event types. This change may uncover hidden
issues, even though some were resolved by "blocksync." The selection
randomness is based on the first 32 bytes of the peerid's pubkey for
each node.

### DIFF
--- a/monad-mock-swarm/src/mock_swarm.rs
+++ b/monad-mock-swarm/src/mock_swarm.rs
@@ -1,10 +1,10 @@
 use std::{
     cmp::Reverse,
     collections::{BTreeMap, BinaryHeap},
-    marker::PhantomData,
     time::Duration,
 };
 
+use itertools::Itertools;
 use monad_consensus_types::{
     message_signature::MessageSignature, signature_collection::SignatureCollection,
 };
@@ -13,6 +13,8 @@ use monad_executor::{timed_event::TimedEvent, Executor, State};
 use monad_executor_glue::{MonadEvent, PeerId};
 use monad_types::{Deserializable, Serializable};
 use monad_wal::PersistenceLogger;
+use rand::{Rng, SeedableRng};
+use rand_chacha::{ChaCha20Rng, ChaChaRng};
 use rayon::prelude::*;
 use tracing::info_span;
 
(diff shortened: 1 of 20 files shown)


Yimaha:
alright, changes should be minimized

monad-mock-swarm/src/mock_swarm.rs
```diff
@@ -206,25 +211,30 @@ where
                     logger: wal,
                     pipeline,
                     pending_inbound_messages: Default::default(),
-                    _marker: PhantomData,
+                    // PeerId is used as a unique seed for generating random
+                    rng: ChaChaRng::from_seed(unsafe {
```
omegablitz: Why does we need unsafe to generate this? Can't we just use a fixed seed at the top of this function, and then use that for generating the seeds for each Node?

monad-mock-swarm/src/mock_swarm.rs
```diff
@@ -57,7 +60,7 @@ where
     MockExecutor<S, RS, ME, ST, SCT>: Unpin,
     S::Block: Unpin,
 {
-    fn peek_event(&self) -> Option<(Duration, SwarmEventType)> {
+    fn peek_event(&mut self) -> Option<(Duration, SwarmEventType)> {
```
omegablitz: If you call `peek_event` repeatedly, it'll toggle between different values - the function shouldn't take a `&mut`

monad-mock-swarm/src/mock_swarm.rs
Yimaha: sure, I will note to avoid unsafe more in future

monad-mock-swarm/src/mock_swarm.rs
```diff
@@ -206,7 +223,9 @@ where
                     logger: wal,
                     pipeline,
                     pending_inbound_messages: Default::default(),
-                    _marker: PhantomData,
+                    // PeerId is used as a unique seed for generating random
```
omegablitz: obsolete comment

monad-mock-swarm/src/mock_swarm.rs
```diff
@@ -58,7 +61,8 @@ where
     S::Block: Unpin,
 {
     fn peek_event(&self) -> Option<(Duration, SwarmEventType)> {
-        std::iter::empty()
+        // avoid modification of the original rng
```
omegablitz, line 64: nit: I think this comment is unnecessary, but up to you

monad-mock-swarm/src/mock_swarm.rs
Yimaha: I'm actually blind

monad-mock-swarm/src/mock_swarm.rs
Yimaha: This is almost like a mental note, if ever needed to modify the rng thing it would be readily apparent, I think it should be better kept (else lmk)

PR #310, opened 2023-10-04
Keep broadcasting timeout messages until TC formed

omegablitz:
Some replay tests don't pass because failed nodes miss the timeout messages broadcasted when they are down. Instead of asking gossip guarantees to fix the issue, we explicitly re-send the timeout messages on the application level.


monad-p2p/src/lib.rs
```diff
@@ -151,36 +144,12 @@ where
...
-        self.outbound_messages.insert(request_id, message);
-        self.outbound_messages_lookup
-            .insert((to_libp2p, id), request_id);
-        assert_eq!(
-            self.outbound_messages.len(),
-            self.outbound_messages_lookup.len()
-        );
-    }
-
-    pub fn unpublish_message(&mut self, to: &monad_executor_glue::PeerId, message_id: &M::Id) {
```
Yimaha, line 170: Just out of curiosity, have we ever used unpublished anywhere on the commit before this pr? (on the application level)

michael-yxchen: I don't think we were using it in the core state machine. 

Message state would use it to unpublish messages from earlier rounds but we weren't using message state.

monad-mock-swarm/tests/replay_test.rs
michael-yxchen, line 1: I want to take another look at this test. 

The rebroadcasted timeout carries a QC, and the replay recovered nodes shouldn't have seen the QC as the proposal was sent after they fail. So I'm expecting block sync requests from the two recovered nodes, but we're seeing none right now

michael-yxchen: Actually nvm, they have the previous block

PR #313, opened 2023-10-05
Add randao_reveal to block payload

abenedito-monad:


monad-consensus-types/src/payload.rs
```diff
@@ -85,13 +88,32 @@ impl std::fmt::Debug for FullTransactionList {
         f.debug_tuple("Txns").field(&self.0).finish()
     }
 }
+#[derive(Clone, Default, PartialEq, Eq)]
+pub struct RandaoReveal(pub Vec<u8>);
+
+impl RandaoReveal {
+    pub fn new<CS: CertificateSignature>(round: Round, keypair: &CS::KeyPairType) -> Self {
```
Yimaha, line 95: I'm not sure if this is a good idea to keep this generic, Secp don't satisfy the requirement for being a VRF, only BLS does

abenedito-monad: Ariq, Michael and I chatted about it and we agreed that this is the best path forwards. We are guaranteed going to be using BLS signatures outside of testing.

monad-consensus-types/src/payload.rs
```diff
@@ -85,13 +88,32 @@ impl std::fmt::Debug for FullTransactionList {
...
+    pub fn verify<CS: CertificateSignature>(
+        &self,
+        round: Round,
+        pubkey: &<CS::KeyPairType as CertificateKeyPair>::PubKeyType,
+    ) -> Result<(), CS::Error> {
+        let sig = CS::deserialize(&self.0)?;
+
+        sig.verify(&round.0.to_le_bytes(), pubkey)
+    }
+}
```
michael-yxchen, line 108: I'm thinking whether we should keep ser/de to the prototype conversions. So we have the flexibility to encode it differently for the `EthBlock` or leader election(?).  It also makes code cleaner as we only have ser/de on the edge. What do you think?

monad-consensus-state/src/lib.rs
```diff
@@ -336,6 +340,19 @@ where
...
+        };
+
+        if let Err(e) = p
+            .block
+            .payload
+            .randao_reveal
+            .verify::<SCT::SignatureType>(p.block.get_round(), author_pubkey)
+        {
+            return cmds;
+        };
```
michael-yxchen: warn and increment counter here?

monad-consensus-state/src/lib.rs
```diff
@@ -336,6 +340,19 @@ where
         let round = self.pacemaker.get_current_round();
         let leader = election.get_leader(round, validators.get_list());
 
+        let Some(author_pubkey) = validator_mapping.map.get(&author) else {
+            return cmds;
+        };
```
michael-yxchen: When we verify the proposal message, we should have verified that author is in the validator set. `expect()` would be appropriate here

monad-consensus-types/src/payload.rs
abenedito-monad: We could do that, but it would require making Block generic over a type which bubbles up everywhere. Ariq and I were thinking about the same thing with regards to beneficiary since it adds a dependency on eth types in monad-consensus-* but we decided to leave it for now and consider adding a ChainType generic later down the line.

monad-consensus-state/src/lib.rs
abenedito-monad: Yep good catch, I added them but they didn't make it into the commit :(

monad-consensus-types/src/payload.rs
michael-yxchen: Cool, makes sense

PR #315, opened 2023-10-06
Fix genesis block transaction lists

abenedito-monad:


Dockerfile
```diff
@@ -10,7 +10,7 @@ FROM chef AS builder
 
 WORKDIR /usr/src/monad-bft
 RUN apt update
-RUN apt install -y python3
+RUN apt install -y python3 clang
```
abenedito-monad, line 13: reth-primitives requires c-kzg which requires clang

monad-node/src/genesis/mod.rs
michael-yxchen, line 1: Trying to understand the reason for this fix. Is it to skip state root validation on the genesis block, or to fix the arbitrary value `0xc0`?

abenedito-monad: TransactionList is a generic blob data type that stores a list of transaction hashes. For our use cases, that blob will be an RLP encoded list of transaction hashes. The problem is that an empty byte array is not a valid RLP encoding of an empty array. The empty array is [0xc0]. This should have broken monad-testground a while back but ethers used to be more permissive in its decoding and would accept the empty array. After the switch to reth, RLP decoding an empty byte array produces an error. This is why monad-testground currently goes into a TC loop because the FetchTxs call fails on all nodes.

PR #316, opened 2023-10-06
Modify block sync Logs to be more precise

Yimaha:
block sync request report (bid, qc)

block sync response always report regardless of result, the report format is (bid, result), also each unique type of result is recorded by a different type of counter

### DIFF
--- a/monad-consensus-state/src/blocksync.rs
+++ b/monad-consensus-state/src/blocksync.rs
@@ -1,3 +1,4 @@
+use core::fmt;
 use std::collections::{hash_map::Entry, HashMap};
 
 use log::debug;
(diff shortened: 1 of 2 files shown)


monad-consensus-state/src/blocksync.rs
```diff
@@ -139,7 +151,11 @@ where
             })
         } else {
             BlockSyncResult::IllegalResponse
-        }
+        };
+
+        debug!("Block sync response: bid={:?}, result={:?}", bid, result);
+        inc_count!(block_sync_response);
```
Yimaha: should we have a separate counter for each type?

Yimaha: should

PR #320, opened 2023-10-11
Generate non-empty txn lists at MockMempool

Yimaha:
By providing the ability for mempool to return non-empty transactions, Some tests such as blackout, random delay are failing due to state_sync triggered by root_hash_delay. The behaviour is correct, but we don't have state sync implemented yet, thus nodes that failed to sync due to being too behind will never make progress.

Thus, mockswarm now accept state_root_delay as a parameter as suppose to always default to 4, thus making it possible to config delay related test's state_root_delay to a higher number and avoid trigger state_sync.

Also, empty txn lists coming from mempool will be RLP encoded so check the RLP encoded empty list value for empty block checks.

### DIFF
--- a/monad-mock-swarm/src/mock.rs
+++ b/monad-mock-swarm/src/mock.rs
@@ -15,6 +15,7 @@ use monad_consensus_types::{
     payload::{FullTransactionList, TransactionList},
     signature_collection::SignatureCollection,
 };
+use monad_eth_types::EMPTY_RLP_TX_LIST;
 use monad_executor::{Executor, State};
 use monad_executor_glue::{
     Command, MempoolCommand, Message, MonadEvent, PeerId, RouterCommand, RouterTarget, TimerCommand,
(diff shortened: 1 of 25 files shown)


monad-mock-swarm/src/mock.rs
```diff
@@ -515,6 +520,17 @@ impl<ST, SCT> Executor for MockMempool<ST, SCT> {
     }
 }
 
+impl<ST, SCT> MockMempool<ST, SCT> {
+    fn get_fetched_txs_list(&self) -> TransactionList {
+        if self.num_fetch_txs == 0 {
+            TransactionList(vec![EMPTY_RLP_TX_LIST])
+        } else {
+            // Random non-empty value
```
Yimaha, line 528: In a separate pr in near future, a way to provided seeded random generation of txs list will be added as it is also useful for twin testing

Yimaha: PR is here https://github.com/monad-crypto/monad-bft/pull/324

PR #321, opened 2023-10-11
Implement MonadFileLedger

abenedito-monad:


ariqchowdhury:
Sorry, I think there was some miscommunication about the FileLedger. The actual file ledger and eth block stuff looks fine, only thing is that this shouldn't replace the existing ledger in ParentExecutor, it needs to be an additional ledger. We need two ledgers, one for monad-bft blocks (blocksync between nodes etc) and one for eth blocks (input to execution). So ParentExecutor can have `ledger` and `execution_ledger` -- the same ledger commit command should be sent to both. Ledger fetch isn't needed on the execution ledger (unless one of the rpc calls requires it?)

PR #322, opened 2023-10-11
Abstract the node identifier in mock_swarm

Yimaha:
Nodes in mock_swarm relies on PeerId for identification, but with TwinBFT, its possible to get duplicate PeerIds within the same swarm.

This PR generalized the identification of nodes to a trait, and added an assert! within the add_state function to ensure each insert is unique.

The generalization applies until the stage where Nodes class is created within test-utils.

### DIFF
--- a/monad-mock-swarm/src/mock.rs
+++ b/monad-mock-swarm/src/mock.rs
@@ -1,6 +1,8 @@
 use std::{
     cmp::Ordering,
     collections::{BTreeSet, VecDeque},
+    fmt::Debug,
+    hash::Hash,
     marker::{PhantomData, Unpin},
     ops::DerefMut,
     pin::Pin,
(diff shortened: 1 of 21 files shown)


monad-quic/src/router_scheduler.rs
```diff
@@ -40,13 +40,14 @@ pub struct QuicRouterScheduler<G: Gossip> {
 
     timeouts: TimeoutQueue,
```
Yimaha, line 42: Decided to keep Quic using peerID, if quic router in future also want to be tested with twin, we can make it generic as well.

monad-mock-swarm/src/mock.rs
```diff
@@ -27,11 +29,22 @@ use monad_updaters::{
 };
 use rand::Rng;
 use rand_chacha::{rand_core::SeedableRng, ChaCha20Rng, ChaChaRng};
+pub trait RouterSchedulerID:
+    Clone + Copy + Debug + Ord + PartialOrd + Eq + PartialEq + Hash + Send + Sync
+{
+    fn get_peer_id(&self) -> PeerId;
+}
+
+impl RouterSchedulerID for PeerId {
```
Yimaha, line 38: might have to move it to peerId's file

monad-mock-swarm/src/mock.rs
```diff
@@ -42,19 +55,23 @@ pub trait RouterScheduler {
     /// transport-level message type - usually will be bytes
     type M;
     type Serialized;
+    type ID: RouterSchedulerID;
```
omegablitz, line 58: Was trying to think through alternatives... do you think it's possible to leave the `RouterScheduler` trait alone, and implement everything at the Pipeline/Transformer level? So RouterScheduler defines how gossip messages should be delivered between PeerIds, and differences between nodes with the same PeerId are handled externally, for example with the following:

```rust
pub struct LinkMessage<ID, M> {
    pub from: ID,
    pub to: ID,
    pub message: M,

    /// absolute time
    pub from_tick: Duration,
}
```

to:
```rust
pub struct OutboundLinkMessage<M> {
    pub from: (PeerId, PeerIdIdx), // (wouldn't actually be named PeerIdIdx, but you get the idea)
    pub to: PeerId,
    pub message: M,

    /// absolute time
    pub from_tick: Duration,
}
```
which gets transformed into:
```rust
pub struct InboundLinkMessage<M> {
    pub from: PeerId,
    pub to: (PeerId, PeerIdIdx), // (wouldn't actually be named PeerIdIdx, but you get the idea)
    pub message: M,

    /// absolute time
    pub from_tick: Duration,
}
```

Yimaha:
close because https://github.com/monad-crypto/monad-bft/pull/328

PR #325, opened 2023-10-13
consensus exposure draft

Yimaha:
This is just a draft for review purpose, 

right now, I want to be able to inspect into the state during mock swarm. There are 2 ways to deal with it.

The first way is describe in this PR, we add ConsensusProcess as a trait onto the type of State, making it possible to extra internal consensus machine while running twin bft related tests.

The **Cons** of such approach is that monad-wal has a test that doesn't really care about the consensus process but it still ran with the "monad_test" feature when cargo test is called, this forces us to have an some form of mock consensus process implemented, which is quiet ugly.

I was trying to see if I can disable the feature "monad_test" when running monad_wal, but no matter how I influence cargo.toml it seems like cargo-test always have monad_test enabled as feature. if anyone knows how to get around it please let me know.

Alternatively, I have to put up a big refactor pr like the following screen shot that makes the concrete impl of MonadState as suppose to state, and I like it even less.

Screenshot 2023-10-13 at 4 53 16 PM

I think the right plan here is find out a way for cargo to run that test with monad_test feature flag disabled, but if there is any other way around it please let me know!

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -45,4 +45,6 @@ jobs:
       - name: Check
         run: cargo check --all-targets --all-features --verbose
       - name: Run tests
-        run: cargo test --release --all-targets --all-features --verbose
+        run: |
+          cargo test --release --all-targets --all-features --verbose && \
+          cd monad-wal && cargo test --release --all-targets --verbose && cd ..
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4461,6 +4461,7 @@ dependencies = [
 name = "monad-executor"
 version = "0.1.0"
 dependencies = [
+ "monad-consensus-state",
  "monad-consensus-types",
  "monad-crypto",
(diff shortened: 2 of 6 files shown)


Yimaha:
Alright I think this version is probably better than any other solution.

The root of the problem is that test all will enable the feature "monad_test" within modules and that influence packages that doesn't rely on monad_test as well, this is really annoying for monad-wal because its not useful to have related features enabled.

I don't doubt there is a better way to configure it in CI such that the `cargo test` on the root directory will be able to compile twice (or something similar) and run for corresponding features, but a simple solution based on https://github.com/rust-lang/cargo/issues/6195 is actually just cd into the sub module and build (which allow you to ignore mandatory features applied by other modules)

This solution is cleaner, with the down side that CI gets a bit longer, and sometimes you will forget to test smaller tests locally (CI will catch it though), closing this PR to start an official one for review

PR #326, opened 2023-10-13
Enforce stricter consensus process type in state

Yimaha:
previously, state doesn't enforce consensus type to be consensus process even under test feature "monad_test". With introduction of Twins, its quiet useful to allow mockswarm to terminate based on consensus related condition (such as round number) as suppose just ledger or tick.

This change enforce the SignatureCollection type and Consensus type, and is fine for majority of test case due to how State is used in practice. However, monad-wal had some problem because its the only module that tests only component unrelated to consensus thus leaving consensus as ()

Some form of mass refactor would be able to avoid this issue by either impl ConsensusProcess for () or convert state => monad-state in Nodes. However, both approach drastically increase the complexity of the module in unnecessary way, making it difficult to read & manage in the future.

The alternative solution, is we compile twice, tests that relies on debug tagged function with "monad_test" would have to export those test as a specialized test with test features and add itself to the CI tool after -p, so they get compiled and tested in a different run from rest of the test that doesn't care about the "monad_test" feature flag

Change is possibly not idiomatic, so looking for suggestion&alternatives

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -12,7 +12,9 @@ env:
 jobs:
   build:
     runs-on: ubuntu-latest-32
-
+    strategy:
+      matrix: 
+          TARGET: [--all-targets, -p monad-mock-swarm --features monad_test]
     steps:
       - uses: actions/checkout@v3
       - run: rustup toolchain install nightly --profile minimal --component rustfmt
@@ -20,7 +22,7 @@ jobs:
         run: cargo +nightly fmt --all --check --verbose
       - uses: Swatinem/rust-cache@v2
       - name: Lint
-        run: cargo clippy --all-targets --all-features --
+        run: cargo clippy ${{ matrix.TARGET }} --
           -D clippy::suspicious
           -D clippy::style
           -D clippy::clone_on_copy
@@ -39,10 +41,10 @@ jobs:
           -A clippy::type_complexity
           -A clippy::int_plus_one
           -A clippy::uninlined-format-args
-          -A clippy::enum-variant-names
+          -A clippy::enum-variant-names 
       - name: Lint unused deps
         run: cargo clippy --all-features -- -D unused_crate_dependencies
       - name: Check
-        run: cargo check --all-targets --all-features --verbose
+        run: cargo check ${{ matrix.TARGET }} --verbose 
       - name: Run tests
-        run: cargo test --release --all-targets --all-features --verbose
+        run: cargo test --release ${{ matrix.TARGET }} --verbose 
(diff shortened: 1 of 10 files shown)


monad-mock-swarm/tests/replay_test.rs
```diff
@@ -1,301 +1,316 @@
...
-    Node<S, RS, P, LGR, ME, ST, SCT>: Send,
-    RS::Serialized: Send,
-{
-    let mut max_tick = start_tick;
-
-    while let Some((tick, _, _)) = nodes.step_until(until_tick, until_block) {
-        assert!(tick >= max_tick);
-        max_tick = tick;
-    }
+#[cfg(all(test, feature = "monad_test"))]
```
Yimaha, line 1: There is no real change in this file except now test is wrapped by [cfg(all(test, feature = "monad_test"))]

.github/workflows/rust.yml
```diff
@@ -12,15 +12,17 @@ env:
 jobs:
   build:
     runs-on: ubuntu-latest-32
-
+    strategy:
+      matrix: 
+          TARGET: [--all-targets, -p monad-mock-swarm --features monad_test]
```
michael-yxchen, line 17: I'm not sure about this change. Say we're adding new functional features (not testing related), then we'll need to add new feature flags to the first target. 

What's the down side of specializing State to MonadState in Nodes? Is it too verbose?

Yimaha: specializing state is quiet verbose, because monad state itself also has a lot of generic argument to begin with, this pr https://github.com/monad-crypto/monad-bft/pull/325 has a screen shot of roughly what would happen if we go down the specialization route, changes where every S:State need to be converted to CT, VT, LT, BST would be everywhere that uses mockswarm. 2 major problems:

1. its going to be a mass refactoring pr, and making mockswarm related class more complicated
2. MockSwarm would have even more complicated types, damaging readability by a lot

and yes, if we decided to add new function features we will need to add it to the first target. I can see like 2 ways to deal with this.

1. we use the reverse flag, that is, provide a feature that disable test so when we use --all-features it won't break anything. however, running just `cargo test` will fail, so people need to start running `cargo test --all-features` instead (essentially, adding features disable certain part of the code as suppose to enable)
2. we proceed with this and add flag to the default compile accordingly, allowing more control (which may even be a good thing, so we know exactly which feature are we compiling with)

Yimaha:
Had a conversation with @michael-yxchen , the result of the is that we want to keep the changes within mock swarm to begin with, so convert to concrete impl and remove the reliant on State for consensus is better in general. 

Closing this PR is reopening with a refactoring pr that convert State -> MonadState in Nodes

PR #328, opened 2023-10-17
Replace PeerId within LinkMessage to ID

Yimaha:
This change is the alternative suggested by Aashish and works quiet nice with limiting the amount of changes, essentially we don't touch router scheduler, and try to control twin-testing's filter mechanism from the transformer level.

With this change, there is no longer a need to refactor router scheduler and changes made to swarm/linkmessage/pipeline is minimized.

### DIFF
--- a/monad-mock-swarm/src/mock_swarm.rs
+++ b/monad-mock-swarm/src/mock_swarm.rs
@@ -8,9 +8,8 @@ use itertools::Itertools;
 use monad_consensus_types::{
     message_signature::MessageSignature, signature_collection::SignatureCollection,
 };
-use monad_crypto::secp256k1::PubKey;
 use monad_executor::{timed_event::TimedEvent, Executor, State};
-use monad_executor_glue::{MonadEvent, PeerId};
+use monad_executor_glue::MonadEvent;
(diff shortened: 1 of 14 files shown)


monad-mock-swarm/src/mock_swarm.rs
```diff
@@ -166,8 +168,9 @@ where
     LGR: PersistenceLogger<Event = TimedEvent<S::Event>>,
     ME: MockableExecutor<SignatureCollection = SCT>,
 {
-    states: BTreeMap<PeerId, Node<S, RS, P, LGR, ME, ST, SCT>>,
+    states: BTreeMap<ID, Node<S, RS, P, LGR, ME, ST, SCT>>,
     tick: Duration,
+    must_deliver: bool,
```
omegablitz, line 173: Is there a situation where we wouldn't want this to be `true`?

Yimaha: test where nodes are physically removed from swarm, in particular  `replay_tests.rs` contain couple test where nodes are removed from the mapping and delivery would fail

PR #329, opened 2023-10-18
Add bandwidth transformer

michael-yxchen:
Add bandwidth transformer implementation and a many node test using the quic router scheduler and the bw transformer.

### DIFF
--- a/monad-mock-swarm/src/transformer.rs
+++ b/monad-mock-swarm/src/transformer.rs
@@ -6,6 +6,7 @@ use std::{
 };
 
 use monad_executor_glue::PeerId;
+use monad_tracing_counter::inc_count;
 use rand::{prelude::SliceRandom, Rng};
 use rand_chacha::{rand_core::SeedableRng, ChaCha20Rng, ChaChaRng};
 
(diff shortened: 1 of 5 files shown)


monad-mock-swarm/src/mock.rs
```diff
@@ -31,6 +32,22 @@ use rand_chacha::{rand_core::SeedableRng, ChaCha20Rng, ChaChaRng};
 
 const MOCK_DEFAULT_SEED: u64 = 1;
 
+pub trait BitLen {
+    fn bit_len(&self) -> usize;
+}
```
ariqchowdhury: should we just require the unit for the transformer to be Bytes/s and then avoid having this trait?

michael-yxchen: I think we'd still need to throw some trait on `LinkMessage<M>::M` so `BwTransformer` can get the length of the message. Or there might be some tricks I don't know about.

Yimaha: I think Aashish had an interesting example for `Vec<u8>` essentially instead of define it as GenericTransformer, its exclusively  `BytesTransformer` which only serves `Vec<u8>`,

Is there any usage for `BwTransformer` to be generic?

michael-yxchen: Cool I'll try that.

No it doesn't need to be generic. It should only transform `Vec<u8>` messages

monad-mock-swarm/src/transformer.rs
```diff
@@ -354,6 +355,53 @@ impl Transformer<Vec<u8>> for BytesSplitterTransformer {
     }
 }
 
+#[derive(Clone, Debug)]
+pub struct BwTransformer {
+    // unit: bits per second
+    upload_bps: usize,
+    // upload bandwidth window
+    bandwidth_window: HashMap<PeerId, (VecDeque<(Duration, usize)>, usize)>,
```
ariqchowdhury: can you add more detail to this comment, its not quickly clear to me what this is.

michael-yxchen: Thought it's probably easier to explain with code than word. Hope it makes sense now

ariqchowdhury: yeah, thanks, that's more clear now.

PR #330, opened 2023-10-18
Make quic swarm fully deterministic

michael-yxchen:
The mock swarm tests with `QuicRouterScheduler` weren't fully deterministic because of two reasons:

1. The `quinn-proto` library calls `StdRng::from_entropy` in its implementation for connection and endpoint.

   The patching PR is pending review from quinn's people. In the meantime, the workspace dependency points to the patch commit in quinn.

2. The `make_certificate` function in `monad-crypto` generates a random key for TLS handshake. Depending on the actual value of the key, the QUIC handshake data might be off by 1 byte. The difference can cascade when mock swarm tests have a BwTransformer in place, causing different messages to be dropped across runs.

   An additional argument is added to `make_certificate` to optionally specify the tls_key bytes so the tests are fully deterministic.

In our nightly randomized tests, we should output the keys so we can debug the failures.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,10 +75,10 @@ prost-types = "0.11.8"
 protobuf-src = "1.1.0"
 ptree = { version = "0.4.0", default-features = false }
 
-# the important commit: "307d80b9398d4e1e305c0131f2c3989090ec9432"
+# the important commit: "dd34d57a0a184ae23707d5ae045ec62d8f0b6ecb"
 # we can use latest release once it includes that
-quinn = { git = "https://github.com/quinn-rs/quinn.git", rev = "307d80b9" }
-quinn-proto = { git = "https://github.com/quinn-rs/quinn.git", rev = "307d80b9" }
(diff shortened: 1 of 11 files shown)


michael-yxchen:
Will mark ready as review once the [quinn PR](https://github.com/quinn-rs/quinn/pull/1691) settles

Cargo.toml
```diff
@@ -75,10 +75,8 @@ prost-types = "0.11.8"
 protobuf-src = "1.1.0"
 ptree = { version = "0.4.0", default-features = false }
 
-# the important commit: "307d80b9398d4e1e305c0131f2c3989090ec9432"
-# we can use latest release once it includes that
-quinn = { git = "https://github.com/quinn-rs/quinn.git", rev = "307d80b9" }
-quinn-proto = { git = "https://github.com/quinn-rs/quinn.git", rev = "307d80b9" }
+quinn = { git = "https://github.com/quinn-rs/quinn.git" }
```
omegablitz: We should include a `rev` tag for versioning until a release is made

PR #334, opened 2023-10-19
Enforce more mandatory trait

Yimaha:
For certain trait like BlockType, MessageSignature, SignatureCollection, it might make more sense for trait to naturally impl Unpin / Send because they will always be used as part of the information send across threads by messages.

This commit enforce these trait and remove redundant where clauses that forces them to be unpin / send

### DIFF
--- a/monad-mock-swarm/src/mock.rs
+++ b/monad-mock-swarm/src/mock.rs
@@ -41,8 +41,8 @@ pub enum RouterEvent<M, Serialized> {
 pub trait RouterScheduler {
     type Config;
     /// transport-level message type - usually will be bytes
-    type M;
-    type Serialized;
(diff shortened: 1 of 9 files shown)


monad-mock-swarm/src/mock.rs
```diff
@@ -41,8 +41,8 @@ pub enum RouterEvent<M, Serialized> {
 pub trait RouterScheduler {
     type Config;
     /// transport-level message type - usually will be bytes
-    type M;
```
Yimaha, line 44: This trait I debait the most, but I think for a message to be part of the router, it should have these trait by default 
Clone for broadcasting related operation
Eq + PartialEq allow cmp organization
Send since its a message

PR #335, opened 2023-10-19
Add TwinsTransformer

Yimaha:
TwinsTransformer captures messages that target default PeerId and duplicate plus broadcast to the correct target(s).

The Vote & Proposal message gets filtered by Round, Timeout always gets broadcasted to everyone to ensure liveness, and all other form of messages that doesn't touch on the core of consensus protocol get drop by default.

TwinsTransformer comes with a specialized MonadMessagePipeline, only adding relevant transformer so far.

To help with spying on message internal, several associated module are granted the ability to spy and clone under test environment

### DIFF
--- a/monad-mock-swarm/src/transformer.rs
+++ b/monad-mock-swarm/src/transformer.rs
@@ -10,22 +10,27 @@ use monad_tracing_counter::inc_count;
 use rand::{prelude::SliceRandom, Rng};
 use rand_chacha::{rand_core::SeedableRng, ChaCha20Rng, ChaChaRng};
 
-const DEFAULT_ID: usize = 0;
+const UNIQUE_ID: usize = 0;
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct ID(usize, PeerId);
 
 impl ID {
     pub fn new(peer_id: PeerId) -> Self {
-        Self(DEFAULT_ID, peer_id)
+        Self(UNIQUE_ID, peer_id)
     }
 
-    pub fn with_identifier(mut self, identifier: usize) -> Self {
+    pub fn as_non_unique(mut self, identifier: usize) -> Self {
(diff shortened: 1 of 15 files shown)


monad-mock-swarm/src/utils.rs
```diff
@@ -0,0 +1,158 @@
+// test utils for mock swarm unit test
+#[cfg(all(feature = "monad_test", test))]
+pub mod test_tool {
```
Yimaha, line 3: Mostly test utils for creating dummy messages, content mostly not interesting, but if this file location and cfg looks right is debaitable

monad-consensus-types/src/certificate_signature.rs
michael-yxchen, line 1: Thinking that we'd want to minimize the `monad_test` feature footprint 

Using `CertificateSignature` as an example, one possible alternative is to duplicate the secret generated in the testutil functions. Since this is a twins testing specific feature, we can probably write specialized create_keys/certificate_keys util functions for it, so the core part of the code base isn't polluted

monad-mock-swarm/src/transformer.rs
```diff
@@ -428,6 +436,185 @@ impl Transformer<Vec<u8>> for BytesTransformer {
...
+pub struct TwinsTransformer {
+    // PeerId -> All Duplicate
+    dups: BTreeMap<PeerId, Vec<usize>>,
+    // Round -> Deliver target
+    partition: BTreeMap<Round, Vec<ID>>,
+    // when rounds cannot be found within partition
+    default_part: Vec<ID>,
+    // block sync and associated message is dropped by default
+    ban_block_sync: bool,
+}
```
michael-yxchen: The `monad_test` feature was introduced so we can inspect some otherwise private fields in `MonadState` in tests. I guess that applies to `Message` now as well. 

Things in mock-swarm are already test only, so we don't need to further gate them.

Yimaha:
@michael-yxchen the comments are addressed

PR #336, opened 2023-10-19
Add NodesTerminator

Yimaha:
Nodes used to terminate on simple condition such until (tick) and until_block (which terminates when any block reaches long enough ledger)

NodesTerminator provides a way for caller of Nodes to terminate on more complex condition, as well as combining assertion and checks within the same terminator as needed.

This change helps Twins to terminate on more precise state.

There are also some minor change to block_sync integration tests to avoid trigger state sync

### DIFF
--- a/monad-mock-swarm/benches/two_node_benchmark.rs
+++ b/monad-mock-swarm/benches/two_node_benchmark.rs
@@ -9,6 +9,7 @@ use monad_consensus_types::{
 use monad_crypto::NopSignature;
 use monad_mock_swarm::{
     mock::{MockMempool, MockMempoolConfig, NoSerRouterConfig, NoSerRouterScheduler},
+    mock_swarm::UntilTerminator,
     transformer::{GenericTransformer, LatencyTransformer},
 };
 use monad_state::{MonadMessage, MonadState};
@@ -41,6 +42,7 @@ fn two_nodes() {
         _,
         MockValidator,
         MockMempool<_, _>,
+        _,
     >(
         MockValidator,
         |all_peers, _| NoSerRouterConfig {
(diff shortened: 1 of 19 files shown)


monad-mock-swarm/src/mock_swarm.rs
```diff
@@ -158,6 +158,120 @@ where
     }
 }
 
+// can also just be a function
+// but might want to keep persistent state for termination condition, so keep as struct for now?
```
Yimaha: I will delete this comment after twins-test related job is finished or if this struct based approach is just okay regardless

monad-mock-swarm/benches/two_node_benchmark.rs
```diff
@@ -51,12 +53,11 @@ fn two_nodes() {
         vec![GenericTransformer::Latency(LatencyTransformer(
             Duration::from_millis(1),
         ))],
+        UntilTerminator::new().until_tick(Duration::from_secs(10)),
         SwarmTestConfig {
             num_nodes: 2,
             consensus_delta: Duration::from_millis(2),
             parallelize: false,
-            until: Duration::from_secs(10),
```
ariqchowdhury, line 58: should we keep this field as an overall timeout for all tests? And then let them supply the terminator as an additional end condition? 
I think for any test, we're going to want to have a timeout duration. So logic in the step functions is 
`if tick > until || terminator.should_terminate()`

Yimaha: There exists some test in log_gen and replay.rs where test defines a timeout parameter of infinite, which is why I kept this to be part of termination condition. However, I do agree having timeout on test is nice to avoid indefinitely long test case. 

My picture for terminator is that it's responsible for ensure behaviour when timeout happens. if you compare the UntilTransformer and ProgressTransformer, behaviour of tmo in Until is simple termination, while ProgressTransformer will panic because you are not suppose to run until this point. 

I think its better for the designer of test to set timeout behaviour as they wish? else you will need to terminate the test, but caller doesn't know if test is terminated by timeout, and need to re-evaluate logic to see if test properly terminated by terminator or terminated by timeout?

ariqchowdhury: okay, makes sense

PR #327, opened 2023-10-17
Abstract swarm relationship into SwarmRelation

Yimaha:
This pr abstract majority of "where" clauses and Generic to a single trait, thus allow easier modification in future and readability.

With the change of converting State in Nodes to MonadState, the amount of generic and the relationship between generic is pilling up, This pr would greatly reduce the difficulty for such change and also any future changes.

PR #338, opened 2023-10-20
Added test feature to check round of State

Yimaha:
Allow twins testing to inspect the internal of consensus's current round without mass refactoring


Yimaha:
closing this, had a discussion with Michael, plan to use #327 to achieve this behaviour

PR #340, opened 2023-10-23
Extract monad_test related features

Yimaha:
Allow monad_test to be abstracted to a mod and add impl as needed, thus making the code base more readable.

Also ordered the derive vs test_derive to be the same across


michael-yxchen:
yeah I think this version is more readable

PR #351, opened 2023-10-26
Make Node<S> in mock-swarm use concrete MonadState

michael-yxchen:
We used to need to add "monad_test" feature gated methods to the `State` trait to access consensus state. Writing the feature everywhere reduces the readability.

We are only using `MonadState` with mock swarm so it makes sense to use the concrete type than the generic `State` type.

### DIFF
--- a/monad-mock-swarm/src/mock.rs
+++ b/monad-mock-swarm/src/mock.rs
@@ -29,6 +29,8 @@ use monad_updaters::{
 use rand::Rng;
 use rand_chacha::{rand_core::SeedableRng, ChaCha20Rng, ChaChaRng};
 
+use crate::swarm_relation::{SwarmRelation, SwarmStateType};
+
 const MOCK_DEFAULT_SEED: u64 = 1;
 
 #[derive(Debug)]
@@ -140,24 +142,21 @@ pub trait MockableExecutor:
     fn ready(&self) -> bool;
 }
 
-pub struct MockExecutor<S, RS, ME, ST, SCT>
+pub struct MockExecutor<S>
 where
-    S: State,
-    ST: MessageSignature,
-    SCT: SignatureCollection,
-    ME: MockableExecutor<SignatureCollection = SCT>,
+    S: SwarmRelation,
 {
-    mempool: ME,
-    ledger: MockLedger<S::Block, S::Event>,
-    execution_ledger: MockExecutionLedger<SCT>,
-    checkpoint: MockCheckpoint<S::Checkpoint>,
-    epoch: MockEpoch<ST, SCT>,
-    state_root_hash: MockStateRootHash<S::Block, ST, SCT>,
+    mempool: S::ME,
+    ledger: MockLedger<<SwarmStateType<S> as State>::Block, <SwarmStateType<S> as State>::Event>,
+    execution_ledger: MockExecutionLedger<S::SCT>,
+    checkpoint: MockCheckpoint<<SwarmStateType<S> as State>::Checkpoint>,
+    epoch: MockEpoch<S::ST, S::SCT>,
+    state_root_hash: MockStateRootHash<<SwarmStateType<S> as State>::Block, S::ST, S::SCT>,
     tick: Duration,
 
-    timer: Option<TimerEvent<S::Event>>,
+    timer: Option<TimerEvent<<SwarmStateType<S> as State>::Event>>,
 
-    router: RS,
+    router: S::RS,
 }
 
 pub struct TimerEvent<E> {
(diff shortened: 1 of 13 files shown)


monad-mock-swarm/src/mock.rs
```diff
@@ -140,24 +142,21 @@ pub trait MockableExecutor:
...
+    S: SwarmRelation,
 {
-    mempool: ME,
-    ledger: MockLedger<S::Block, S::Event>,
-    execution_ledger: MockExecutionLedger<SCT>,
-    checkpoint: MockCheckpoint<S::Checkpoint>,
-    epoch: MockEpoch<ST, SCT>,
-    state_root_hash: MockStateRootHash<S::Block, ST, SCT>,
+    mempool: S::ME,
+    ledger: MockLedger<<SwarmStateType<S> as State>::Block, <SwarmStateType<S> as State>::Event>,
```
Yimaha, line 150: feel like this part might hurt readability in future, maybe we can do 
```
pub type SwarmBlock<S> = <SwarmStateType<S> as State>::Block;
...
// similar for other type that are set in place

```

and group all these pub type to swarm_relation.rs?

monad-mock-swarm/src/mock.rs
```diff
@@ -370,8 +377,10 @@ where
                         None => continue, // try next tick
                         Some(RouterEvent::Rx(from, message)) => {
                             let message =
-                                <S::Message as Deserializable<RS::M>>::deserialize(&message)
-                                    .expect("all messages should deserialize in mock executor");
+                                <<SwarmStateType<S> as State>::Message as Deserializable<
```
Yimaha: the `as` part is not necessary, we can do 


```

<<SwarmStateType<S> as State>::Message>::deserialize(&message)
```

monad-mock-swarm/src/mock_swarm.rs
```diff
@@ -19,13 +25,22 @@ use crate::{
     transformer::{LinkMessage, Pipeline, ID},
 };
 
+pub type SwarmStateType<S> = MonadState<
```
Yimaha: if we enforce `S: SwarmRelation` then `<S as SwarmRelation>` could be shorten to `S`

also calling it `SwarmState` instead of `SwarmStateType` maybe? 

This pub type should probably be placed in `swarm_relation.rs`?

monad-mock-swarm/src/mock_swarm.rs
```diff
@@ -19,13 +25,22 @@ use crate::{
...
+    ConsensusState<<S as SwarmRelation>::SCT, <S as SwarmRelation>::TVT, StateRoot>,
+    <S as SwarmRelation>::ST,
+    <S as SwarmRelation>::SCT,
+    ValidatorSet,
```
Yimaha: fields like these should probably be part of swarm relation instead of fixed, but it won't hurt until like a lot later (when variety of leader election / validator  is available, blocksyncstate being generic is kinda a mistake but I plan to fix it later)

monad-mock-swarm/tests/replay_test.rs
```diff
@@ -58,7 +62,12 @@ fn run_nodes_until<S>(
 where
     S: SwarmRelation,
 
-    MockExecutor<S::STATE, S::RS, S::ME, S::ST, S::SCT>: Unpin,
+    MonadMessage<<S as SwarmRelation>::ST, <S as SwarmRelation>::SCT>:
```
Yimaha: can be simplified to
```
    MonadMessage<S::ST, S::SCT>: Deserializable<S::Message>,
    VerifiedMonadMessage<S::ST, S::SCT>: Serializable<S::Message>,
```

monad-mock-swarm/src/mock_swarm.rs
Yimaha: also with this change we can remove the `STATE` field completely from Swarm relation

monad-mock-swarm/src/mock_swarm.rs
michael-yxchen: yeah, I can move it to `swarm_relation.rs`

The rust compiler doesn't support enforcing trait bounds on type aliases. 

> "[bounds on generic parameters are not enforced in type aliases](warning: bounds on generic parameters are not enforced in type aliases)"

We've been calling these type alias `...Type`, e.g. SignatureType & SignatureCollectionType. So I thought I can stick to that convention.

Yimaha: Oh wow rust doesn't allow generic type alias, that's kinda disappointing

monad-mock-swarm/tests/replay_test.rs
Yimaha: or alternatively similar to
```
   SwarmMessage<S>: Deserializable<S::Message>,
```

once we have 
```
pub type SwarmMessage<S> = <SwarmStateType<S> as State>::Message;
````

monad-mock-swarm/src/mock_swarm.rs
michael-yxchen: I can add these to `SwarmRelation` tho I prefer changing it when there's the need.

I remember that the purpose of making `BlockSyncState` generic was that we can mock it and test MonadState in isolation. So maybe it's ok too keep it this way?

As for removing `STATE` field completely, I don't feel strongly either way.

monad-mock-swarm/src/mock.rs
michael-yxchen: yeah sure

monad-mock-swarm/tests/replay_test.rs
michael-yxchen: yeah good point. I was too lazy and copy pasted

monad-mock-swarm/src/mock_swarm.rs
Yimaha: STATE was a field to kept when state are generic in Nodes, but that part is now concrete and its not useful to keep that around anymore. (SwarmStateType does that job of STATE, and better)

The BlockSyncState thing is really whatever, we can keep them all like that for sure, it won't hurt and when it hurt later its like 10 lines of change

monad-mock-swarm/src/mock.rs
michael-yxchen: Actually `SwarmEventType` was defined and renaming either one to `SwarmEvent` would be confusing. So I'm wanting to keep things as is.

PR #358, opened 2023-10-31
Add timeout limit for blocksync

Yimaha:
block sync will give up on a block when tracking cnt exceed a limit.

when limit is exceeded, instead of `Retry`, bsync manager emits `GiveUp` to consensus process.

limit is defined as 8 right now. can be added to MonadConfig in future.

This pr is debaitable because #359 might cover the last edge case for timeout (for now), and if we want to repeat request indefinitely. (as a back up mechanism in case consensus doesn't cancel the block sync request or 
something similar happens). an alternative in my mind to ensure that indefinite request never happens is to move the block sync module parallel to pending_block_tree, and it can check if the block is needed by checking the root height of pending_block_tree + current content of pending block tree

### DIFF
--- a/monad-consensus-state/src/blocksync.rs
+++ b/monad-consensus-state/src/blocksync.rs
@@ -30,16 +30,18 @@ pub struct InFlightBlockSync<SCT> {
 
 pub enum BlockSyncResult<SCT: SignatureCollection> {
     Success(FullBlock<SCT>),            // retrieved and validated
-    Failed(Vec<ConsensusCommand<SCT>>), // unable to retrieve
+    Retry(Vec<ConsensusCommand<SCT>>),  // unable to retrieve, retry
     IllegalResponse,                    // never requested from this peer or never requested
+    GiveUp(Vec<ConsensusCommand<SCT>>), // no more retry
 }
 
 impl<SCT: SignatureCollection> fmt::Debug for BlockSyncResult<SCT> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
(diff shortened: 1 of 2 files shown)


monad-consensus-state/src/blocksync.rs
```diff
@@ -213,15 +225,18 @@ where
...
             {
-                *req_target = peer;
-                *retry_cnt = cnt;
+                if cnt <= self.retry_limit {
+                    *req_target = peer;
+                    *retry_cnt = cnt;
+                    cmds.extend(vec![entry.get().into(), self.get_timeout(bid)]);
+                }
+            } else {
+                // give up on this block
```
Yimaha, line 243: this is a bug, let me fix this

monad-consensus-state/src/blocksync.rs
```diff
@@ -201,29 +217,37 @@ where
         &mut self,
         bid: BlockId,
         validator_set: &VT,
-    ) -> Vec<ConsensusCommand<SCT>> {
```
Yimaha, line 204: return become more complex, this used to only retry indefinitely

Yimaha:
closed on this:

```

The root of the problem is when should a block sync request be invalidated and pruned

The easiest, free version is simply don’t, because now the blocksync is safe guarded by qc, its very difficult for someone to provide me a proof that block exists s.t. I don’t already have the block or I can’t find it after some amount of request (at least 1 honest validator would have observed it and reply to you as long as block sync don't give up)

The only edge case is when blocksync started and we get the block through delayed message, yet before someone responded to us, we will still be block syncing. However, this is just an optimization that allow us to terminate early to reduce traffic

```

PR #359, opened 2023-10-31
Allow pruning of in-flight blocks

Yimaha:
if a proposal successfully added a missing block to the pending block tree, then block sync no longer cares about if such block is retrieved. retry and timeout will longer be happening

### DIFF
--- a/monad-consensus-state/src/blocksync.rs
+++ b/monad-consensus-state/src/blocksync.rs
@@ -225,6 +225,10 @@ where
         }
         cmds
     }
+
+    pub fn remove(&mut self, bid: BlockId) {
+        self.requests.remove(&bid);
+    }
 }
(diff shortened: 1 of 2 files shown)


Yimaha:
closed on this:

```

The root of the problem is when should a block sync request be invalidated and pruned

The easiest, free version is simply don’t, because now the blocksync is safe guarded by qc, its very difficult for someone to provide me a proof that block exists s.t. I don’t already have the block or I can’t find it after some amount of request (at least 1 honest validator would have observed it and reply to you as long as block sync don't give up)

The only edge case is when blocksync started and we get the block through delayed message, yet before someone responded to us, we will still be block syncing. However, this is just an optimization that allow us to terminate early to reduce traffic

```

PR #363, opened 2023-10-31
Make `proposal_size` configurable

michael-yxchen:
Only a small subset of all the tests needs (random) payload data, namely twins testing and tests involving `QuicRouterScheduler`. Empty payloads works fine for `NoSerRouterScheduler` test as the RS is moving the entire block atomically. Computing a random payload of 5k * 32 every test is wasteful.

This PR makes `proposal_size` configurable. It also fixes `get_fetched_txs_list`:
1. the total payload size `num_fetch_txs -> num_fetch_txs * 32`
2. use `fill_bytes` instead of calling `rng.gen` many times. This method is 4x faster from my benchmarks

### DIFF
--- a/monad-mock-swarm/benches/two_node_benchmark.rs
+++ b/monad-mock-swarm/benches/two_node_benchmark.rs
@@ -15,7 +15,11 @@ pub fn criterion_benchmark(c: &mut Criterion) {
     c.bench_function("two_nodes", |b| b.iter(two_nodes));
 }
 
-criterion_group!(benches, criterion_benchmark);
+criterion_group! {
+    name = benches;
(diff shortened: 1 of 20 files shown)


monad-mock-swarm/benches/two_node_benchmark.rs
```diff
@@ -37,6 +37,7 @@ fn two_nodes() {
             expected_block: 1024,
             state_root_delay: 4,
             seed: 1,
+            proposal_size: 0,
```
omegablitz: Should we make this one 5_000 because it's a bench?

michael-yxchen: yeah good point

PR #365, opened 2023-11-02
Move message serde to RouterScheduler

abenedito-monad:
Added easy to read Inbound, Outbound, and Transport Messages to RouterScheduler and propagated type names throughout SwarmRelations.

### DIFF
--- a/monad-mock-swarm/src/mock.rs
+++ b/monad-mock-swarm/src/mock.rs
@@ -31,35 +31,44 @@ use priority_queue::PriorityQueue;
 use rand::{Rng, RngCore};
 use rand_chacha::{rand_core::SeedableRng, ChaCha20Rng, ChaChaRng};
 
-use crate::swarm_relation::{SwarmRelation, SwarmStateType};
+use crate::swarm_relation::SwarmRelation;
 
 const MOCK_DEFAULT_SEED: u64 = 1;
 
 #[derive(Debug)]
-pub enum RouterEvent<M, Serialized> {
-    Rx(PeerId, M),
-    Tx(PeerId, Serialized),
+pub enum RouterEvent<InboundMessage, TransportMessage> {
+    Rx(PeerId, InboundMessage),
+    Tx(PeerId, TransportMessage),
 }
 
 /// RouterScheduler describes HOW gossip messages get delivered
 pub trait RouterScheduler {
     type Config;
-    /// transport-level message type - usually will be bytes
-    type M;
-    type Serialized;
+
+    // Transport level message type (usually bytes)
+    type TransportMessage;
+
+    // Application level data
+    type InboundMessage: Deserializable<Self::TransportMessage>;
+    type OutboundMessage: Serializable<Self::TransportMessage>;
 
     fn new(config: Self::Config) -> Self;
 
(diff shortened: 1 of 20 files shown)


monad-mock-swarm/src/mock.rs
```diff
@@ -367,51 +378,40 @@ where
...
+    ) -> Option<MockExecutorEvent<<S::State as State>::Event, S::TransportMessage>> {
         while let Some((tick, event_type)) = self.peek_event() {
             if tick > until {
                 break;
             }
+
             self.tick = tick;
-            let event = match event_type {
+
+            return Some(match event_type {
```
omegablitz: nit: IMO the control flow is actually more confusing with this change because the inner expression can also return/continue

monad-mock-swarm/src/mock_swarm.rs
```diff
@@ -189,9 +195,16 @@ impl UntilTerminator {
     }
 }
 
-impl<S> NodesTerminator<S> for UntilTerminator
+impl<S, CT, ST, SCT, VT, LT, BST> NodesTerminator<S> for UntilTerminator
 where
-    S: SwarmRelation,
+    S: SwarmRelation<State = MonadState<CT, ST, SCT, VT, LT, BST>>,
+
+    CT: ConsensusProcess<SCT> + PartialEq + Eq,
```
omegablitz, line 202: Just for my understanding, what changed to require the additional trait bounds now?

omegablitz:
Related: #351

michael-yxchen:
In #351, I made `Node<S>` depend on the concrete MonadState type because `UntilTerminator` needs to access consensus.round, which isn't available through the `State` trait. 

I think Andre was able to limit all the trait bounds to `UntilTerminator`, and thus those extra trait bounds on `UntilTerminator`. Node/MockExecutor is over the generic S::State again. 

I personally like the changes. My solution of making everything use the concrete type `StateType<S>` wasn't the most elegant

Yimaha:
related to #351, Michael & I previously did talk about the shrinking to only terminator approach.

I think in general, the design decision is if we are concreting the actual usage of `MonadState` within Nodes or allow arbitrary state to run. and the result of the discussion is that its unlikely for mockswarm to be used in any other way so its probably fine to concrete it.

This change looks pretty good, one thing I would say is useful is to have  `CT, VT, LT, BST` to be included within the swarm relation, as there is no need for repeated definition, although it might be redundant for states that doesn't use these (however, its that possible to have another state?)

for example, the following would be more like
```
impl<S, CT, ST, SCT, VT, LT, BST> NodesTerminator<S> for UntilTerminator
where
    S: SwarmRelation<State = MonadState<CT, ST, SCT, VT, LT, BST>>,

    CT: ConsensusProcess<SCT> + PartialEq + Eq,

// and more 
```

to 

```
impl<S> NodesTerminator<S> for UntilTerminator
where
    S: SwarmRelation<State = MonadState<S::CT, S::ST, S::SCT, S::VT, S::LT, S::BST>>,
```

and you don't have to carry all the generic everywhere

at the end of the day, SwarmRelation is just something that does the dirty work for you and bag all the annoying traits together + how should diff modules use each other

PR #366, opened 2023-11-02
Implement gossip trait for BroadcastTree

ariqchowdhury:
BroadcastTree divides application messages into chunked gossip messages and sends these parts to neighbours based on a k-ary tree topology. The trees are rooted by the original broadcaster and the remainder of the tree is ordered based on PRNG shuffle seeded by the application message ID.


monad-gossip/src/broadcasttree.rs
```diff
@@ -0,0 +1,737 @@
...
+            let read_buffer = self.read_buffers.entry(from).or_default();
+            let mut start: usize = 0;
+            let mut end: usize = OUTER_HEADER_SIZE;
+            if read_buffer.len() < end {
+                return;
+            }
+
+            let outer_header: OuterHeader = match bincode::deserialize(&read_buffer[start..end]) {
+                Ok(b) => b,
+                Err(e) => return, //TODO
```
omegablitz: Should we make these TODOs `todo!()` so they're loud?

PR #373, opened 2023-11-13
Switch monad-node to use monad-quic

abenedito-monad:


monad-node/Cargo.toml
```diff
@@ -17,8 +17,10 @@ monad-consensus-types = { path = "../monad-consensus-types" }
 monad-crypto = { path = "../monad-crypto", features = ["libp2p-identity"] }
```
omegablitz: We can stop using the `libp2p-identity` feature here if we create a `Keypair::from_der` function

PR #392, opened 2023-11-14
Delete monad-test feature

omegablitz:
The monad-test feature was being used for two things:
1. Clone implementation of a Secp KeyPair
2. Internal interospection of Unverified<_> messages

The observation made is that neither of these are actually necessary - we just change the NoSer RouterScheduler to use OutboundMessage instead of InboundMessage for its TransportMessage.


Yimaha:
Thank you, sorry for the mess ...  🙏

PR #399, opened 2023-11-15
Document and clean blocksync

ariqchowdhury:


monad-consensus-state/src/blocksync.rs
```diff
@@ -1,58 +1,77 @@
...
-    Success(FullBlock<SCT>),            // retrieved and validated
-    Failed(Vec<ConsensusCommand<SCT>>), // unable to retrieve
-    IllegalResponse,                    // never requested from this peer or never requested
+    /// retrieved and validated
+    Success(FullBlock<SCT>),
+
+    /// unable to retrieve
+    Failed(Vec<ConsensusCommand<SCT>>),
+
+    /// never requested from this node or never requested
```
Yimaha, line 51: this comment is quite outdated, now its probably more accurate to say "never requested or request timed-out"

monad-consensus-state/src/blocksync.rs
```diff
@@ -1,58 +1,77 @@
...
+    UnexpectedResponse,
 }
 
 impl<SCT: SignatureCollection> fmt::Debug for BlockSyncResult<SCT> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
             BlockSyncResult::Success(_) => write!(f, "successful"),
             BlockSyncResult::Failed(_) => write!(f, "failed"),
-            BlockSyncResult::IllegalResponse => write!(f, "illegal"),
+            BlockSyncResult::UnexpectedResponse => write!(f, "illegal"),
```
Yimaha: "illegal" > "unexpected" ?

monad-consensus-state/src/blocksync.rs
```diff
@@ -70,131 +89,127 @@ impl<SCT: SignatureCollection> InFlightBlockSync<SCT> {
...
+        self.requests.insert(id, req);
+
+        vec![req_cmd, self.create_timeout_command(id)]
     }
 
-    pub fn handle_retrieval<VT: ValidatorSetType, TV: TransactionValidator>(
+    /// Handle the response to a BlockSync request
+    /// If the request was not fulfilled, the request is tried again with
+    /// a different node
+    pub fn handle_request<VT: ValidatorSetType, TV: TransactionValidator>(
```
michael-yxchen: I think this handles block sync response rather than request. Would `handle_response` be more appropriate here?

We can also rename `BlockSyncMessage` to maybe `BlockSyncResponseMessage` so it's more clear from the type name.

monad-consensus-state/src/blocksync.rs
```diff
@@ -70,131 +89,127 @@ impl<SCT: SignatureCollection> InFlightBlockSync<SCT> {
...
-                choose_peer(self.id, validator_set.get_list(), *retry_cnt + 1)
-            {
-                *req_target = peer;
-                *retry_cnt = cnt;
-            }
-            BlockSyncResult::Failed(vec![entry.get().into(), self.get_timeout(bid)])
+            BlockSyncResult::Failed(self.request_helper(
+                &pending_req.qc,
+                validator_set,
+                pending_req.retry_cnt + 1,
```
michael-yxchen, line 204: I could be wrong, but it seems like the `retry_cnt` is not incremented. We can add a test where block sync keeps timing out and queries all other nodes in the set for the block

Yimaha: I think the current flow is:
1. remove entry if exists (thus we still care)
2. if retrieval failed, try with retry_cnt + 1
3. also skip yourself during retry

`test_never_request_to_self` is a test that check if block sync keeps requesting to everyone in the current valset

looks like retry_cnt is incremented here?

michael-yxchen: I see it now. Didn't realize it was removed and re-inserted with a new retry_cnt. thx

monad-consensus-state/src/blocksync.rs
ariqchowdhury: agree, changed both

PR #404, opened 2023-11-15
Add virtual benchmarks for monad-gossip

omegablitz:


omegablitz:
This shouldn't get merged until #405 is

monad-transformer/src/lib.rs
```diff
@@ -462,6 +462,75 @@ impl Transformer<Vec<u8>> for BwTransformer {
...
+        // new_scheduled_ts == last_scheduled_ts + extra_delay
+        // new_scheduled_ts == from_tick_2 + latency + delay_2
+        //
+        //
+        //
+        // last_scheduled_ts + extra_delay = from_tick_2 + latency + delay_2
+        // delay_2 = last_scheduled_ts + extra_delay - from_tick_2 - latency
+        // delay_2 = (from_tick_2 + latency + delay_offset) + extra_delay - from_tick_2 - latency
+        // delay_2 = delay_offset + extra_delay
+        // delay_2 = (from_tick_1 - from_tick_2 + delay_1) + extra_delay
```
ariqchowdhury: can be deleted?

omegablitz: The calculation for computing the next delay was nontrivial for me to think through, so the comment with how it was derived may be useful in the future if we need to change this transformer at all.

I'm ok deleting it though if we think it's too verbose.

omegablitz: I can also dump it in notion instead and delete it here if that's preferable

ariqchowdhury: oh okay, I wasn't sure what that was. My comment now is that its confusing and I don't understand it. Can you add more detail?

omegablitz: Yeah ur right, will do

omegablitz: Updated the comments, if it's still confusing let me know and I'll try and rephrase or simplify it - otherwise will merge this in later today.

ariqchowdhury: yeah, looks good now thanks

PR #405, opened 2023-11-15
Add sampling rate parameter to BwTransformer

michael-yxchen:
BwTransformer used to allow an unrealistic instant burst of 100mb for a 100mbps link. It now takes in a sampling frequency parameter to limit the burst size.

### DIFF
--- a/monad-transformer/src/lib.rs
+++ b/monad-transformer/src/lib.rs
@@ -399,25 +399,25 @@ impl Default for BwWindow {
 }
 
 impl BwWindow {
-    fn advance_to(&mut self, now: Duration) {
+    fn advance_to(&mut self, now: Duration, sampling_period: Duration) {
         while self
             .window
             .front()
-            .map_or(false, |&(tick, _)| tick + Duration::from_secs(1) < now)
+            .map_or(false, |&(tick, _)| tick + sampling_period < now)
         {
             let (_, bitlen) = self.window.pop_front().unwrap();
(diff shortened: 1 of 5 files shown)


monad-transformer/src/lib.rs
```diff
@@ -429,26 +429,30 @@ impl BwWindow {
 
 #[derive(Clone, Debug)]
 pub struct BwTransformer {
-    // unit: bits per second
-    upload_bps: usize,
+    // number of bits allowed in the sampling perdiod
```
omegablitz: nit: typo

monad-transformer/src/lib.rs
```diff
@@ -429,26 +429,30 @@ impl BwWindow {
...
+    burst_size: usize,
+    // sampling period
+    sampling_period: Duration,
     // upload bandwidth window
     window: BwWindow,
 }
 
 impl BwTransformer {
-    pub fn new(upload_mbps: usize) -> Self {
+    pub fn new(upload_mbps: usize, sampling_freq: u32) -> Self {
```
omegablitz: nit: I think the unit of `sampling_freq` is confusing, it's not obvious to me that it's Hz.

Separately, would it make sense to pass in a `sampling_period: Duration` in directly? That seems more intuitive to me, but probably up to personal preference.

michael-yxchen: yeah sampling_period should be more intuitive

PR #409, opened 2023-11-17
Evidence Interface

Yimaha:
first commit add some evidence interface for communication & storage between executor and monad-state
The format for Evidence is: `Evidence(node_id, Violation{proof_a, proof_b, ...})`

2nd commit includ evidence being collected at vote_state:
1. hash(vote_info) != vote_info_hash 
2. invalid vote signature

3rd commit include evidence being collected at Pacemaker (timeout related)

4th commit include evidence related unverified message failed to verify

(5th commit is mostly naming change)

commits can be reviewed separately
(I will probably squash the whole thing once approved)

### DIFF
--- a/monad-consensus/src/evidence.rs
+++ b/monad-consensus/src/evidence.rs
@@ -11,7 +11,7 @@ use crate::messages::{
 pub type SerializedSignature = Vec<u8>;
 // vote message itself carries a SCT signature, sufficient for evidence
 type ProofOfVote<SCT> = VoteMessage<SCT>;
-type ProofOfTimeOut<SCT> = TimeoutMessage<SCT>;
+type ProofOfTimeout<SCT> = TimeoutMessage<SCT>;
 type ProofOfMessage<SCT> = (ConsensusMessage<SCT>, SerializedSignature);
 /// The design of Evidence Interface is
 /// Evidence(node_id, Violation{proof_a, proof_b, ...})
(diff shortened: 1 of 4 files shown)


Yimaha:
As a side note, system at the moment havn't included design about how the evidence should be "shared" to other nodes

in general, I think its decent if we "cache" the evidence in consensus and just wait for a node's own turn to propose, and once a proposal went through (block committed), we officiall store the evidence in long-term storage.

This variant of change need some modification of the block&proposal format, thus will need further consideration, so not included early phase of pr, at the moment, we can consider all the evidence being collected **locally only.** if in future the need for evidence sharing rises, we can simply add a middle-man between executor and state that only commit evidence after observing a block is committed

monad-consensus/src/evidence.rs
```diff
@@ -0,0 +1,20 @@
...
+
+use crate::messages::message::VoteMessage;
+
+/// In order to decouple message signature, ConsensusViolation
+/// stores the message signature as bytes.
+/// Internally, individual nodes should never re-verify signature
+/// but when sharing/sending evidence with other nodes,
+/// original signature on the violation must be retained.
+pub type SerializedSignature = Vec<u8>;
+type ProofOfVote<SCT> = (VoteMessage<SCT>, SerializedSignature);
```
Yimaha: actually, for vote message & timeout message you don't need SerializedSignature because it naturally carries a SCT signature, but for proposal message its needed, fixing

monad-state/src/lib.rs
```diff
@@ -426,57 +426,70 @@ where
                         sender,
                         unverified_message,
                     } => {
-                        let verified_message = match unverified_message.verify::<HasherType, _>(
+                        match unverified_message.clone().verify::<HasherType, _>(
```
Yimaha, line 429: Some note about why this is a clone()

the current verification approach is very good, it enforced the creation of verified message only through calling verify() on unverified message, however, it does means handling error is semi-annoying because the (message, signature) has to flow every where if you wish to return (error, message, signature) for evidence collection.

Tried a couple approaches, seems like the cleanest way that introduce minimum change is just cloning. it shouldn't be expensive in most cases, and it retain high readability. if this become a performance issue in future we can refactor the verification interface too.

PR #414, opened 2023-11-21
Add more compression algo and decompression bench

michael-yxchen:
- Add lz4 and deflate compression algo
- Add decompression time benchmark


michael-yxchen:
Probably don't need to run the benchmarks in CI though. It tests every compression level for completeness. We can keep a few based on time to complete e.g. only keeping levels with compression time lower than 50ms

PR #423, opened 2023-11-29
Reduce copies throughout network layer 

omegablitz:
This removes all extraneous copies past the Protobuf serialization stuff. We accomplish this by using bytes::Bytes (ref-counted) instead of Vec<u8>. There are still extraneous copies happening at the Protobuf serialization layer - will address this next.

### DIFF
--- a/monad-quic/src/service.rs
+++ b/monad-quic/src/service.rs
@@ -129,7 +129,12 @@ where
         for command in commands {
             match command {
                 RouterCommand::Publish { target, message } => {
-                    let message = message.serialize();
+                    let message = {
+                        let mut _ser_span = tracing::info_span!("serialize_span").entered();
+                        message.serialize()
+                    };
+                    let mut _publish_span =
+                        tracing::info_span!("publish_span", message_len = message.len()).entered();
                     self.gossip.send(time, target, message);
 
                     if let Some(waker) = self.waker.take() {
@@ -159,6 +164,8 @@ where
         mut self: std::pin::Pin<&mut Self>,
         cx: &mut std::task::Context<'_>,
     ) -> Poll<Option<Self::Item>> {
+        let mut _router_poll_span = tracing::info_span!("router_poll_span").entered();
+
         let this = self.deref_mut();
         if this.waker.is_none() {
             this.waker = Some(cx.waker().clone());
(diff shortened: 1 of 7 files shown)


omegablitz:
Saves a couple ms per proposal - tested this before/after on both `monad-quic/examples/service.rs` and monad-testground

Before: 
image
After:
image

omegablitz:
Also, all of these changes are rooted in: `monad-gossip/src/lib.rs`. Can explain verbally if that's easier, this PR is kinda big...

monad-quic/src/service.rs
```diff
@@ -305,12 +330,12 @@ impl InboundConnection {
         Self::Pending(fut)
     }
     fn active(peer: NodeId, mut stream: RecvStream) -> Self {
-        const RX_MESSAGE_BUFFER_SIZE: usize = 64 * 1024;
+        const RX_MAX_SIZE_CHUNK: usize = 64 * 1024;
```
omegablitz: delete this

monad-gossip/src/broadcasttree.rs
```diff
@@ -481,17 +522,11 @@ mod tests {
...
             .iter()
-            .map(|s| s.as_bytes())
-            .collect::<Vec<&[u8]>>();
+            .map(|s| s.as_bytes().into())
+            .collect::<Vec<Bytes>>();
         for m in app_messages {
             g.send(Duration::ZERO, RouterTarget::Broadcast, m);
         }
-        // plus 1 for the emit to self
-        assert_eq!((num_routes + 1) * expected_messages.len(), g.events.len());
```
omegablitz, line 490: Add this assertion back, just need to multiply expected_messages length by appropriate amount

monad-gossip/src/broadcasttree.rs
```diff
@@ -545,32 +580,26 @@ mod tests {
...
-        assert_eq!(1, g.events.len());
 
-        let e = g.events.pop_front().unwrap();
-        match e {
-            GossipEvent::Send(from, msg) => {
-                g.handle_gossip_message(Duration::ZERO, from, msg.as_slice())
-            }
-            GossipEvent::Emit(_from, _msg) => panic!("expecting a GossipEvent::Send"),
-        }
-        assert_eq!(1, g.events.len());
```
omegablitz, line 563: Add this assertion back, multiplied by x

monad-gossip/src/broadcasttree.rs
```diff
@@ -545,32 +580,26 @@ mod tests {
...
+            match e {
+                GossipEvent::Send(_from, _msg) => {
+                    panic!("direct send should not result in any gossiped sends")
+                }
+                GossipEvent::Emit(_from, msg) => {
+                    assert_eq!(
+                        "appmessagepayload",
+                        String::from_utf8(msg.to_vec()).expect("bytes cannot have been corrupted")
+                    );
+                }
```
omegablitz: Only should be one event emitted, and then  add an assertion that no more events

monad-gossip/src/broadcasttree.rs
```diff
@@ -413,10 +456,8 @@ mod tests {
             num_routes,
         }
         .build();
-        let app_message = "appmessagepayload".as_bytes();
+        let app_message = "appmessagepayload".into();
         g.send(Duration::from_secs(1), RouterTarget::Broadcast, app_message);
-        // expecting num_routes number of sends plus 1 event for emit to self
-        assert_eq!(num_routes + 1, g.events.len());
```
ariqchowdhury, line 419: can this one be restored too?

omegablitz: Ah yes I missed that one, good catch

PR #424, opened 2023-11-29
Document Pacemaker and address todo

ariqchowdhury:


monad-consensus/src/pacemaker.rs
```diff
@@ -17,32 +17,58 @@ use crate::{
...
-    // used to not duplicate broadcast/tc
+    /// States for TimeoutMessage handling, ensures we only broadcast
+    /// one message at each phase of handling
     phase: PhaseHonest,
 }
 
+/// The different states of TimeoutMessage handling
 #[derive(Debug, PartialEq, Eq)]
 enum PhaseHonest {
+    /// No timeout messages have been received
```
omegablitz: This isn't exactly correct ( <F+1 && self hasn't timed out), but I can't think of a better way of wording it...
Same with PhaseHonest::One, because a local timeout triggers it, even if <F+1

It's straightforward enough to read the code and understand it though, so no actionable

monad-consensus/src/pacemaker.rs
```diff
@@ -17,32 +17,58 @@ use crate::{
...
+    /// None if we advanced to the current round via QC
+    /// Some(TC) if we advanced to the current round via TC
     last_round_tc: Option<TimeoutCertificate<SCT>>,
 
-    // only need to store for current round
+    /// map of the TimeoutMessages from other Nodes
+    /// only needs to be stored for the current round as timeout messages
+    /// carry a QC which will be processed and advance the node to the
+    /// highest known round
     pending_timeouts: HashMap<NodeId, TimeoutMessage<SCT>>,
```
omegablitz: Unrelated to this PR, but this should be a BTreeMap for deterministic ordering - as we're iterating over it to build TC.

ariqchowdhury: I'll fix that here

monad-consensus/src/pacemaker.rs
ariqchowdhury: right, I'll try to reword it

PR #426, opened 2023-11-29
Restore vote message test check

ariqchowdhury:


monad-consensus/tests/message.rs
```diff
@@ -239,47 +238,6 @@ fn max_high_qc() {
     assert_eq!(tc.max_round(), Round(3));
 }
 
-#[test]
-fn test_vote_message() {
```
michael-yxchen, line 243: Not sure what it was trying to test. Is it testing vote message signing and verification?

Also probably want to create `lci` using `LedgerCommitInfo::new` so the `vote_info_hash` value is valid.

ariqchowdhury: I think this was an older test just making sure that signing and verifying of a message worked. We can consider reviewing the unit tests later to remove redundant ones.

michael-yxchen: cool, we can review the redundant ones later then

PR #430, opened 2023-11-29
Document and clean payload file

ariqchowdhury:


monad-consensus-types/src/payload.rs
```diff
@@ -143,43 +166,76 @@ pub struct Payload {
...
+
+    /// State root hash for the sequence number is missing. Implies
+    /// this node has missed an update from execution
     Missing,
 }
 
+/// Special hash value used in proposals for the first delay-num blocks
+/// from genesis when there isn't a valid state root hash to include.
+const INITIAL_DELAY_STATE_ROOT_HASH: Hash = Hash([0; 32]);
+
```
michael-yxchen, line 222: Just an idea: I remember we were concerned that this initial value is a meaningful state root. A probably reasonable value to use is the state root hash after executing the genesis block.

ariqchowdhury: yeah, thought about this, in our usage, we check first that the incoming block's seqnum is in [0, delay-gap] before checking this special value for hash. We don't unconditionally allow this state root hash to skip validation, so if there ever is a block that legit creates this state root hash, it will still go through the normal validation. 
first hash after executing genesis block is reasonable too though, but the current way is simpler.

michael-yxchen: Oh, I see. So any magic value works here because we checks the seqnum first.

PR #435, opened 2023-11-30
Process QC/TC in `process_proposal_message`

michael-yxchen:
In the previous implementation, a faulty mempool can cause the consensus to stuck on a round because it never processes the certificates in the proposal.

Consensus should process the certificates before fetching the transactions from the mempool. This PR moves certificate handling earlier so mempool issues won't stall consensus.

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -132,10 +132,13 @@ where
 
     fn handle_timeout_expiry<H: Hasher>(&mut self) -> Vec<PacemakerCommand<SCT>>;
 
-    fn handle_proposal_message<H: Hasher>(
+    fn handle_proposal_message<H: Hasher, VT: ValidatorSetType, LT: LeaderElection>(
         &mut self,
         author: NodeId,
         p: ProposalMessage<SCT>,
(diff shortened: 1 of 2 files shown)


ariqchowdhury:
I need to think more about this one. My initial thoughts are that the existing behaviour isn't a problem. If mempool is faulty, I think its desirable for the node to get stuck and require a restart/recovery. If consensus just continues, it will just be blocksyncing on every single round. It won't be voting because mempool is faulty, so its not actually contributing to consensus. And it won't be able to propose when it comes time to be leader. 

In the case of a one-off mempool error, does the current behaviour get stuck? Shouldn't the next proposal be processed correctly and then trigger blocksync?

michael-yxchen:
The current behavior can deal with one-off mempool errors just fine. The next proposal will trigger block sync. That's the same as the behavior after this PR. Only difference is when the round is incremented, which honestly isn't a big deal.

Your point about faulty mempool makes sense, and the current behavior is arguably better as it doesn't keep wasting consensus resources.

It might be relevant if we're basing any consensus behavior on the round number, but I don't believe we have anything like that rn. So we can probably scratch this for now

PR #437, opened 2023-12-01
Document and cleanup blocktree

michael-yxchen:


monad-blocktree/src/blocktree.rs
```diff
@@ -109,7 +126,6 @@ pub enum RootKind {
 pub struct BlockTree<T> {
     pub root: RootKind,
     tree: HashMap<BlockId, BlockTreeBlock<T>>,
-    high_round: Round,
```
michael-yxchen, line 112: was used to assert that blocks are asserted in increasing rounds. Not used after we started accepting out-of-order proposals

monad-blocktree/src/blocktree.rs
```diff
@@ -279,80 +291,58 @@ impl<T: SignatureCollection> BlockTree<T> {
...
             }
 
-            let Some(parent_id) = &i.parent else {
-                return None;
-            };
+            txs.push(btb.block.get_block().payload.txns.clone());
 
-            bid = parent_id;
+            visit = &btb.parent;
         }
```
michael-yxchen, line 340: The current impl includes tx from the root block. The function is used to de-dup the transactions fetched from the mempool when making a proposal. We don't need to include the root blocks as it's committed, and the txns are removed from mempool.

monad-blocktree/src/blocktree.rs
```diff
@@ -33,27 +33,42 @@ impl std::error::Error for BlockTreeError {
...
+    parent: BlockId,
+
+    /// Link to the child blocks in the tree
+    ///
+    /// Only used to print the tree top-down for debugging
     children: Vec<BlockId>,
 }
 
 impl<T: SignatureCollection> BlockTreeBlock<T> {
     const CHILD_INDENT: &str = "    ";
```
ariqchowdhury: clippy has a warning for this, wants an explicit lifetime, will eventually be an error, `&'static`

PR #438, opened 2023-12-04
Use Bytes in prost to reduce copies

omegablitz:


omegablitz:
Saves around ~300us per block - further investigation necessary for additional improvement.
Before:
image
After:
image

PR #439, opened 2023-12-05
Document and cleanup monad-wal

ariqchowdhury:


monad-wal/src/lib.rs
```diff
@@ -4,13 +4,24 @@ pub mod wal;
 
 use std::error::Error;
 
-// the persistence layer only accepts one type of message
-// we can refactor M to Verified/Unverified type if we write to WAL after verifying the message
+/// PersistenceLogger is used to log all events input to MonadState and have them
+/// available for replay
+/// Use case is to track events and if a Node crashed, this logger would allow the Node
+/// to recover its state up to the point of the crash and then continue operating
```
michael-yxchen: maybe helpful to pull copy the explanation here: https://www.notion.so/monad-labs/WAL-replay-531e3fd7bfd84bccbd064b175a39c5d9?pvs=4 just to keep things in one place

PR #442, opened 2023-12-05
Document and clean up vote_state

omegablitz:


omegablitz:
https://github.com/monad-crypto/monad-bft/blob/02b34f110127e2f47b1e71c2e634840b8fac6c23/monad-consensus/src/vote_state.rs#L23
This previously didn't actually do anything, because it'd get wiped on start_new_round

monad-consensus/src/vote_state.rs
```diff
@@ -13,21 +14,42 @@ use monad_validator::validator_set::ValidatorSetType;
...
-// At the end of a round, older rounds can be cleaned up
+/// VoteState accumulates votes and creates a QC if enough votes are received
+/// Only one QC should be created in a round using the first supermajority of votes received
+/// At the end of a round, older rounds can be cleaned up
 #[derive(Debug, PartialEq, Eq)]
 pub struct VoteState<SCT: SignatureCollection> {
-    pending_votes:
-        BTreeMap<Round, HashMap<Hash, (Vec<(NodeId, SCT::SignatureType)>, HashSet<NodeId>)>>,
-    qc_created: BTreeSet<Round>,
+    /// Reeceived pending votes for rounds >= self.earliest_round
```
michael-yxchen: typo

monad-consensus/src/vote_state.rs
```diff
@@ -54,7 +76,7 @@ where
 
         let mut ret_commands = Vec::new();
 
-        if self.qc_created.contains(&round) {
+        if round < self.earliest_round {
```
michael-yxchen, line 80: do we want to make this an assertion, like `assert!(round >= self.earliest_round)` because `handle_vote_message` should be rejecting it first?

omegablitz: We can... but it makes the testing more complicated. Also, even though that shouldn't happen, it's not really dangerous as `process_vote` will end up discarding the vote anyways.

Opted to add a trace error message instead, so that if the invariant is broken it'll at least be noisy in the node logs.

PR #448, opened 2023-12-07
Move vote_info_hash from LedgerCommitInfo to VoteMessage Hash

ariqchowdhury:
PR is split into 2 commits putting the naming/refactoring updates in the second commit to make it easier to see the actual logic changes. 
This PR just focuses on removing the unnecessary vote_info_hash. There are further changes to make in this area that will come in the next PRs:
- remove the actual hash from ledger_commit_info since its only used as a bool in our code
- add some getters to make code more readable (block.qc.vote.vote_info.id...)


ariqchowdhury:
monad-testground and probably monad-node need to be updated too, they're still hashing the LedgerCommitInfo manually when generating the genesis sigs.

ariqchowdhury:
also, ignore the proto changes here, next PR cleans it up better

PR #450, opened 2023-12-08
Add MempoolSupervisor to enable switching MonadMempool to quic

abenedito-monad:


omegablitz:
We should create a test that demonstrates MonadMempool being used as a RouterScheduler, eg similar to the `many_nodes_quic` test we have right now.
https://github.com/monad-crypto/monad-bft/blob/5d78ee9cd0de0da33218c8b77f022dec5e51e663/monad-updaters/src/mempool.rs#L129
https://github.com/monad-crypto/monad-bft/blob/5d78ee9cd0de0da33218c8b77f022dec5e51e663/monad-updaters/src/mempool.rs#L192

Also separately, the goal is for monad-mempool-messenger to use the monad-quic Service backend as then we can remove the libp2p dependency completely - we can potentially roll that as a followup PR though.

PR #451, opened 2023-12-08
Make TlsVerifier safe

omegablitz:
Our previous TlsVerifier wouldn't actually do any cryptographic verification that the certificate was signed by the key corresponding to the pubkey attached.

The new TlsVerifier only supports ed25519 certificates, and only supports secp256k1 as the node's public identity. In the future, we can add support for other algorithms.

Removed the configurable tls_key for now for two reasons:
1) tls_key is not currently used in any tests (it eventually should be)
2) instead of embedding the tls_key option into TlsVerifier, we can
   create a separate SeedableTlsVerifier to reduce API surface area of
   the production one


omegablitz:
Good idea.. do you think we should put that in this TlsVerifier or in a separate DebugTlsVerifier? Put another way, would we ever want to set key_log in a production environment, or would that be something we'd only do in debug/devnet environments?

Just worried about adding options that could be set incorrectly to the "production" one.

omegablitz:
Discussed offline, will add a configuration option to the production TlsVerifier, as it may be useful for debug even in production.

Will do this in a followup PR, as this is more complicated than initially anticipated. Right now we're generating ephemeral client/server certificates for each connection, which makes the UX of configuring `key_log` poor.

PR #455, opened 2023-12-14
Replace hash in LedgerCommitInfo with enum

ariqchowdhury:
We don't create a hash on ledger commit because we are not executing right away so we can just use an enum to indicate whether commit condition was met or not


monad-consensus/src/validation/safety.rs
```diff
@@ -135,6 +133,6 @@ fn safe_to_extend<SCT>(
     }
 }
 
-fn commit_condition(block_round: Round, qc: QcInfo) -> bool {
-    consecutive(block_round, qc.vote.round)
+fn commit_condition(block_round: Round, qc_info: QcInfo) -> bool {
+    consecutive(block_round, qc_info.get_round())
```
michael-yxchen, line 136: Thinking we can check commit condition in VoteMessage validation as well. `VoteInfo.round == VoteInfo.parent_round + 1` <-> Committable. And if a Vote message says otherwise, we should reject it

monad-consensus/src/validation/signing.rs
```diff
@@ -603,25 +596,29 @@ mod test {
         let vset = ValidatorSet::new(stake_list).unwrap();
         let val_mapping = ValidatorMapping::new(voting_identity);
 
-        let msg = HasherType::hash_object(&lci);
+        let vote = Vote {
+            vote_info: vi,
+            ledger_commit_info: CommitResult::Commit,
```
michael-yxchen: super nit: a valid message shoule say NoCommit.

monad-proto/proto/voting.proto
```diff
@@ -15,5 +15,8 @@ message ProtoVoteInfo {
 
 message ProtoVote {
   ProtoVoteInfo vote_info = 1;
-  monad_proto.ledger.ProtoLedgerCommitInfo ledger_commit_info = 2;
-}
+  oneof ledger_commit_info {
+    monad_proto.ledger.ProtoNoCommit no_commit = 2;
+    monad_proto.ledger.ProtoCommit commit = 3;
+  }
```
michael-yxchen, line 21: Just curious what's the reason to use `oneof` over `enum`?

monad-consensus/src/validation/signing.rs
```diff
@@ -634,14 +631,16 @@ mod test {
...
             round: Round(0),
             parent_id: BlockId(Hash([0x00_u8; 32])),
             parent_round: Round(0),
             seq_num: SeqNum(0),
         };
-
-        let lci = LedgerCommitInfo::new::<HasherType>(Some(Hash([0xad_u8; 32])), &vi);
+        let vote = Vote {
+            vote_info: vi,
+            ledger_commit_info: CommitResult::Commit,
```
michael-yxchen: same here

monad-consensus/src/validation/signing.rs
```diff
@@ -848,11 +833,9 @@ mod test {
             parent_round: Round(0),
             seq_num: SeqNum(0),
         };
-        let lci = LedgerCommitInfo::new::<HasherType>(Some(Hash([0xad_u8; 32])), &vi);
-
         let v = Vote {
             vote_info: vi,
-            ledger_commit_info: lci,
+            ledger_commit_info: CommitResult::Commit,
```
michael-yxchen: ^

monad-consensus-types/src/voting.rs
```diff
@@ -111,39 +110,4 @@ mod test {
 
         assert_eq!(h1, h2);
     }
-
-    #[test_case(None ; "None commit_state")]
-    #[test_case(Some(Default::default()) ; "Some commit_state")]
-    fn vote_hash(cs: Option<Hash>) {
```
michael-yxchen, line 126: We can probably keep this test. I plan to revisit it when I review message hashing.

monad-consensus/src/validation/safety.rs
ariqchowdhury: https://github.com/monad-crypto/monad-bft/pull/455/commits/b34ac8ca19f592f807f313bc870f3814ad7a3d7f#diff-8259aa875d3a190cc815a89d73d01e7564cb3d3e3fc93ac26009d8df5e151f81R265-R278

I think this is what you meant. If there's a better way to word the comment too, let me know

monad-proto/proto/voting.proto
ariqchowdhury: We don't actually use the rust types that get generated from protobuf and we make conversion functions between our definitions and the proto. For proto enum, it just uses an int type but with oneof, it creates a rust enum so then when we do the convert we have to explicitly match the proto enum to our enum.

monad-consensus/src/validation/safety.rs
michael-yxchen: yeah that's what I was trying to say

monad-proto/proto/voting.proto
michael-yxchen: Prost provides a [getter function](https://docs.rs/prost/latest/prost/index.html#enumerations) to convert the int type to the enum type, then we can match that to our enum. 

Only raising this up because the `ProtoCommit` and `ProtoNoCommit` types feel redundant. With the protobuf enum approach, we need an extra field with zero value, as the getter function returns the default zero value enum when it fails to convert the i32 to the enum type. We can also avoid it by calling the  `from_i32` method directly, which returns an Option<EnumType>

PR #456, opened 2023-12-14
Split signature verification and protocol validation

michael-yxchen:
`Unverified::verified` was overloaded to do both consensus message signature verification and protocol validation (QC, well-form, etc.) Message types under MonadMessage won't fit in because they don't have a consensus message signature. Splitting the two so we always remember to verify.

### DIFF
--- a/monad-consensus/src/validation/signing.rs
+++ b/monad-consensus/src/validation/signing.rs
@@ -101,116 +101,29 @@ impl<S, M> From<Verified<S, M>> for Unverified<S, M> {
     }
 }
 
-impl<S, SCT> Unverified<S, ConsensusMessage<SCT>>
-where
-    S: MessageSignature,
-    SCT: SignatureCollection,
-{
-    // A verified proposal is one which is well-formed and has valid
-    // signatures for the present TC or QC
-    pub fn verify<VT: ValidatorSetType>(
-        self,
-        validators: &VT,
-        validator_mapping: &ValidatorMapping<SignatureCollectionKeyPairType<SCT>>,
-        sender: &PubKey,
-    ) -> Result<Verified<S, ConsensusMessage<SCT>>, Error> {
-        // FIXME-2 this feels wrong... it feels like the enum variant should factor into the hash so
-        //       signatures can't be reused across variant members
-        Ok(match self.obj {
-            ConsensusMessage::Proposal(m) => {
-                let verified = Unverified::new(m, self.author_signature).verify(
-                    validators,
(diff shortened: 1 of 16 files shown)


monad-consensus/src/validation/signing.rs
```diff
@@ -259,57 +243,46 @@ impl<S: MessageSignature, SCT: SignatureCollection> Unverified<S, VoteMessage<SC
...
     pub fn verify<VT: ValidatorSetType>(
         self,
         validators: &VT,
-        validator_mapping: &ValidatorMapping<SignatureCollectionKeyPairType<SCT>>,
         sender: &PubKey,
-    ) -> Result<Verified<S, TimeoutMessage<SCT>>, Error> {
-        self.well_formed_timeout()?;
-        let msg = HasherType::hash_object(&self.obj);
+    ) -> Result<Verified<S, Unvalidated<TimeoutMessage<SCT>>>, Error> {
+        let msg = HasherType::hash_object(&self.obj.obj);
```
michael-yxchen: make it more readable

monad-consensus/src/validation/signing.rs
```diff
@@ -220,35 +215,24 @@ where
...
-    fn valid_seq_num(&self) -> Result<(), Error> {
-        if self.obj.block.get_seq_num() != self.obj.block.qc.info.vote.seq_num + SeqNum(1) {
-            return Err(Error::InvalidSeqNum);
-        }
-        Ok(())
-    }
 }
 
-impl<S: MessageSignature, SCT: SignatureCollection> Unverified<S, VoteMessage<SCT>> {
+impl<S: MessageSignature, SCT: SignatureCollection> Unverified<S, Unvalidated<VoteMessage<SCT>>> {
```
ariqchowdhury: because validation is now separate from verification, does verify need to be specialized over the specific message type? The verification could be the same for all of them now

michael-yxchen: yeah, #457 removes all of them and keep only the one on ConsensusMessage.

monad-consensus/src/validation/signing.rs
```diff
@@ -111,81 +127,79 @@ where
     pub fn verify<VT: ValidatorSetType>(
```
omegablitz: Now that validation is split out from verification, can the implementation of `verify` be simpler?
More concretely, implement `verify` for all `Unverified<S, T: Hashable>`, and implement `Hashable` for `Unvalidated<T: Hashable>`

PR #457, opened 2023-12-14
Prefix enum hashing with a discriminant

michael-yxchen:
Add a discriminant prefix when hashing ConsensusMessage and BlockSyncResponse to avoid malleable hash/signatures.

### DIFF
--- a/monad-consensus/src/messages/consensus_message.rs
+++ b/monad-consensus/src/messages/consensus_message.rs
@@ -7,6 +7,7 @@ use monad_crypto::{
     hasher::{Hashable, Hasher},
     secp256k1::KeyPair,
 };
+use monad_types::EnumDiscriminant;
 
 use crate::{
     messages::message::{ProposalMessage, TimeoutMessage, VoteMessage},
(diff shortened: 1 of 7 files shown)


monad-consensus/src/messages/consensus_message.rs
```diff
@@ -43,14 +44,23 @@ where
 {
     fn hash(&self, state: &mut impl Hasher) {
         match self {
-            ConsensusMessage::Proposal(m) => m.hash(state),
+            ConsensusMessage::Proposal(m) => {
```
michael-yxchen, line 48: add type string prefix to hash

monad-consensus/src/messages/consensus_message.rs
```diff
@@ -42,15 +43,25 @@ where
     SCT: SignatureCollection,
 {
     fn hash(&self, state: &mut impl Hasher) {
+        state.update(std::any::type_name::<Self>().as_bytes());
```
michael-yxchen, line 46: Not sure if this is what you guys meant

omegablitz: Yup, that looks good. If you didn't already, could you just verify that the name it generates is actually unique?

michael-yxchen:
The prefix is a fully qualified type name, with the signature collection type name in their as well.

PR #461, opened 2023-12-19
Use genesis_prime_qc instead of genesis block

omegablitz:
This commit removes the genesis block. Instead, the initial state is rooted on genesis_prime_qc. This allows us to avoid needing to manually generate and collect signatures over the genesis block. Almost all of the changes needed to accomplish this are in the blocktree.


monad-blocktree/src/blocktree.rs
```diff
@@ -33,133 +32,19 @@ impl std::error::Error for BlockTreeError {
...
-                for (_, block) in self.tree.iter().filter(|a| a.1.block.get_round() == r) {
-                    block.tree_fmt(self, f, &"".to_owned())?;
-                }
-            }
-        }
-
-        Ok(())
-    }
+    /// The round and block_id of last committed block
+    root: (Round, BlockId),
```
ariqchowdhury: make struct

PR #464, opened 2023-12-19
Implement ValidatorSetUpdater (v2)

bamrith:


monad-types/src/epoch_manager.rs
michael-yxchen, line 1: Monad types holds the basic types that other crates depend on. I think it's best to move `EpochManager` out of `monad-types`, maybe into it's own crate.

monad-state/src/lib.rs
```diff
@@ -78,35 +86,13 @@ where
...
-            VT::new(upcoming_val_set.get_stakes())
+    fn update_next_val_set(&mut self, val_data: ValidatorData<SCT>) {
+        let next_epoch = self.epoch_manager.current_epoch + Epoch(1);
+        self.val_epoch_map.insert(
+            next_epoch,
+            VT::new(val_data.get_stakes())
                 .expect("ValidatorData should not have duplicates or invalid entries"),
+            ValidatorMapping::new(val_data.get_cert_pubkeys()),
         );
     }
```
michael-yxchen: I think it'd be helpful to have the updater send the epoch number associated with the validator data, and we can verify it here. In case the epoch_manager's epoch and execution updater go out-of-sync, whether it's starting up or the StateRoot/TxAsync part is buggy, at least we can detect it here and not use a different validator set from everyone else

monad-types/src/epoch_manager.rs
```diff
@@ -0,0 +1,65 @@
...
+
+    // validator set is updated every 'val_set_update_interval'
+    // blocks
+    pub val_set_update_interval: SeqNum,
+    // The start of next epoch is 'epoch_start_delay' rounds after
+    // the proposed block
+    pub epoch_start_delay: Round,
+
+    // A key-value (R, E) indicates that Epoch E starts on round R
+    pub epoch_starts: BTreeMap<Round, Epoch>,
```
michael-yxchen: Why is it mapping `R->E` but not the other way around? I think either way works, just wondering if there's a reason

monad-types/src/epoch_manager.rs
bamrith: Added a new crate `monad-epoch`. Currently just holds `EpochManager`.

monad-state/src/lib.rs
bamrith: Updated the event to send validator data and its associated epoch number

monad-types/src/epoch_manager.rs
bamrith: Mapping from `E->R` makes more sense semantically. Changed to that since the searching algorithm is the same both ways.

monad-consensus-types/src/validation.rs
```diff
@@ -14,4 +14,6 @@ pub enum Error {
     InsufficientStake,
     /// Seq num in block proposal must be 1 higher than in the QC
     InvalidSeqNum,
+    /// Required validator set (or cert pubkeys) not in validators epoch mapping
+    ValDataUnavailable,
```
ariqchowdhury: nit: ValidatorDataUnavailable

monad-consensus/src/validation/signing.rs
```diff
@@ -106,21 +108,39 @@ where
     S: MessageSignature,
     SCT: SignatureCollection,
 {
+    fn get_msg_epoch(&self, epoch_manager: &EpochManager, current_round: Round) -> Epoch {
+        match &self.obj {
+            ConsensusMessage::Proposal(m) => epoch_manager.get_epoch(m.block.round),
+            ConsensusMessage::Vote(m) => epoch_manager.get_epoch(m.vote.vote_info.round),
+            ConsensusMessage::Timeout(m) => epoch_manager.get_epoch(m.timeout.tminfo.round),
+            _ => epoch_manager.get_epoch(current_round),
```
ariqchowdhury: nit: would prefer not to have default match here so that new messages require people to explicitly think about what to do. There shouldn't be many messages here even in the future

monad-types/src/epoch_manager.rs
ariqchowdhury, line 1: I don't think monad-types is the correct crate for this file, its not really a fundamental type...monad-state or monad-consensus-state makes more sense to me.

monad-types/src/epoch_manager.rs
```diff
@@ -0,0 +1,76 @@
...
```
ariqchowdhury: nit: add comment for the EpochManager and follow the comment formatting of other structs

monad-consensus-state/src/lib.rs
```diff
@@ -150,26 +153,26 @@ where
         author: NodeId,
         p: ProposalMessage<SCT>,
         txns: FullTransactionList,
-        validators: &VT,
-        validator_mapping: &ValidatorMapping<SignatureCollectionKeyPairType<SCT>>,
+        epoch_manager: &mut EpochManager,
+        validators_epoch_mapping: &ValidatorsEpochMapping<VT, SCT>,
```
ariqchowdhury: nit: validators_epoch_mapping and val_epoch_map are both used, can it be consistent

monad-epoch/src/epoch_manager.rs
```diff
@@ -0,0 +1,70 @@
...
+
+/// Stores all the information related to epochs and its associated
+/// round numbers
+#[derive(Clone)]
+pub struct EpochManager {
+    pub current_epoch: Epoch,
+
+    // validator set is updated every 'val_set_update_interval'
+    // blocks
+    pub val_set_update_interval: SeqNum,
```
michael-yxchen: `epoch_length` maybe if you want a different name

monad-epoch/src/epoch_manager.rs
```diff
@@ -0,0 +1,70 @@
+use std::collections::BTreeMap;
+
+use monad_types::{Epoch, Round, SeqNum};
+
+/// Stores all the information related to epochs and its associated
+/// round numbers
+#[derive(Clone)]
+pub struct EpochManager {
+    pub current_epoch: Epoch,
```
michael-yxchen: just leaving a note: remove `current_epoch`

monad-epoch/src/epoch_manager.rs
```diff
@@ -0,0 +1,70 @@
...
+        let start_round = self.epoch_starts.insert(epoch, round);
+
+        assert!(start_round.is_none());
+    }
+
+    // Schedule next epoch start if the committed block is the last one in the current epoch
+    pub fn schedule_epoch_start(&mut self, block_num: SeqNum, block_round: Round) {
+        if block_num % self.val_set_update_interval == SeqNum(0) {
+            let epoch_start_round = block_round + self.epoch_start_delay;
+            self.insert_epoch_start(self.current_epoch + Epoch(1), epoch_start_round);
```
michael-yxchen: after removing `current_epoch`, we can use `block_num / self.val_set_update_interval + 1` as the new epoch number

Note for upgradability: this is assuming epoch length is constant. If some protocol upgrades change the value, we need to calculate from an "upgrade" offset

monad-epoch/src/epoch_manager.rs
michael-yxchen, line 1: Moving to monad-state

monad-epoch/src/epoch_manager.rs
bamrith: I avoided that intentionally since an epoch starts on a round and ends on another, and `epoch_length` sounds like it should be in terms of `Round` and not `SeqNum`. If that is easily understandable, I'll use that anyway.

monad-consensus/src/validation/signing.rs
bamrith: outdated commit. code no longer exists.

monad-types/src/epoch_manager.rs
bamrith: Moved to monad-validator

monad-consensus-state/src/lib.rs
bamrith: changed all occurrences to `val_epoch_map`

monad-epoch/src/epoch_manager.rs
bamrith: Moved to monad-validator because of dependency issues.

monad-epoch/src/epoch_manager.rs
bamrith: removed `current_epoch`.

Note: EpochManager is still passed as a mutable reference to consensus state since inserting new epoch starts requires mutability.

PR #465, opened 2023-12-19
Document monad-consensus/validation

michael-yxchen:


monad-consensus/src/validation/signing.rs
```diff
@@ -76,6 +91,9 @@ impl<S, M> AsRef<Unverified<S, M>> for Verified<S, M> {
     }
 }
 
+/// An unverified message is a messae with a signature, but the signature hasn't
```
ariqchowdhury: nit: typo, "message"

PR #475, opened 2023-12-27
Add connection retry support

omegablitz:
These changes are rooted on the additions made to the Gossip trait, namely the `Gossip::connected`, `Gossip::disconnected` functions and the `GossipEvent::RequestConnect` and `GossipEvent::RequestDisconnect` emittable events.

`monad_quic::Service` and `monad_quic::RouterScheduler` needed to be modified to accomodate these changes.

PR #480, opened 2024-01-03
Record WAL in monad-node

omegablitz:


omegablitz:
To test on top of #475, #476, #477, and #479 after merged

omegablitz:
Re-requesting review to go over the second commit I added here, which adds a separate replay mechanism to our `Executor` trait.

omegablitz:
Tested with the following:
```docker run -e RUST_LOG=monad_node=info -v $(pwd)/docker/devnet/monad:/monad --rm -it $(docker build -q -f docker/devnet/Dockerfile .) 2>&1 | grep ledger_len```

PR #481, opened 2024-01-03
Reorganize MonadEvent

michael-yxchen:
Move FetchedBlock and BlockSyncRequest out of ConsensusEvent, into BlockSyncEvent. It untangles the messages so we can delete the `PublishMessage` enum.

It's also easier now to fit sub-states (e.g. consensus, block sync) under MonadState.

### DIFF
--- a/monad-state/src/lib.rs
+++ b/monad-state/src/lib.rs
@@ -10,7 +10,7 @@ use monad_consensus::{
     validation::signing::{Unvalidated, Unverified, Validated, Verified},
 };
 use monad_consensus_state::{
-    command::{Checkpoint, ConsensusCommand, PublishMessage},
+    command::{Checkpoint, ConsensusCommand},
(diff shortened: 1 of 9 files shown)


monad-state/src/lib.rs
```diff
@@ -659,6 +601,44 @@ where
...
             }
+
+            MonadEvent::BlockSyncEvent(block_sync_event) => match block_sync_event {
+                BlockSyncEvent::BlockSyncRequest {
+                    sender,
+                    unvalidated_request,
+                } => {
+                    let validated_request = match unvalidated_request.validate() {
+                        Ok(req) => req,
+                        Err(e) => todo!("{e:?}"),
```
ariqchowdhury: nit: even though you're just moving the code, could you add the error handling here while we're touching this code?

PR #485, opened 2024-01-08
Add update function to all states under MonadState

michael-yxchen:
All MonadState fields that handles events should implement an update function that takes in events and outputs commands. The events should be registered under MonadEvent.

### DIFF
--- a/monad-state/src/consensus.rs
+++ b/monad-state/src/consensus.rs
@@ -1,0 +1,256 @@
+use std::marker::PhantomData;
+
+use monad_consensus::{
+    messages::{consensus_message::ConsensusMessage, message::RequestBlockSyncMessage},
+    validation::signing::Validated,
+};
+use monad_consensus_state::{
+    command::{Checkpoint, ConsensusCommand},
+    ConsensusProcess,
+};
+use monad_consensus_types::{
+    block::Block, signature_collection::SignatureCollection, txpool::TxPool,
+};
+use monad_crypto::certificate_signature::{
+    CertificateSignaturePubKey, CertificateSignatureRecoverable,
+};
+use monad_executor_glue::{
+    CheckpointCommand, Command, ConsensusEvent, ExecutionLedgerCommand, LedgerCommand, MonadEvent,
+    RouterCommand, StateRootHashCommand, TimerCommand,
+};
+use monad_types::{NodeId, RouterTarget, TimeoutVariant};
+use monad_validator::{
+    epoch_manager::EpochManager, leader_election::LeaderElection, validator_set::ValidatorSetType,
+    validators_epoch_mapping::ValidatorsEpochMapping,
+};
+
+use crate::{handle_validation_error, MonadState, VerifiedMonadMessage};
+
+pub(super) struct ConsensusChildState<'a, CP, ST, SCT, VT, LT, TT>
+where
+    ST: CertificateSignatureRecoverable,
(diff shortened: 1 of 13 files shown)


monad-state/src/lib.rs
```diff
@@ -344,316 +336,66 @@ where
...
+    type Event;
+    type Command;
+    type AuxData;
+
+    fn init(config: Self::Config) -> Self;
+
+    fn update(&mut self, event: Self::Event, aux_data: &'a mut Self::AuxData)
+        -> Vec<Self::Command>;
+
+    fn to_parent_command(
```
omegablitz: I'm not sure this function is necessary... I think this could be done with an `impl From<Vec<ChildCommand>> for Vec<ParentCommand>` outside the trait.
This way ChildState wouldn't need to be parametrized by ParentState.

monad-state/src/lib.rs
```diff
@@ -344,316 +336,66 @@ where
...
+            }
         }
     }
 }
+
+pub trait ChildState<'a, ParentState: State>: Sized {
+    type Config;
+    type Event;
+    type Command;
+    type AuxData;
```
omegablitz: I understand the motivation, but I'm still a bit suspicious of this AuxData type. It's weird to me that `update` takes in two mutable references instead of just one.

I'm wondering if we can get rid of `Config`/`AuxData`/`init`, and implement ChildState on `(<StateFromParent>, <StateForChild>)`.

Let's discuss this further as a group.

michael-yxchen: yeah the mutable reference is definitely not ideal. One way to fix it is to ask ChildState to write to other child states through commands instead of directly mutating. We can chat about alternatives as well

monad-state/src/consensus.rs
```diff
@@ -0,0 +1,298 @@
...
+    MempoolCommand, MonadEvent, RouterCommand, StateRootHashCommand, TimerCommand,
+};
+use monad_types::{NodeId, RouterTarget, TimeoutVariant};
+use monad_validator::{leader_election::LeaderElection, validator_set::ValidatorSetType};
+
+use crate::{handle_validation_error, MonadState, VerifiedMonadMessage};
+
+pub(super) struct ConsensusChildState<'a, CP, MS, SCT, VT, LT>(
+    pub(super) &'a mut MonadState<CP, MS, SCT, VT, LT>,
+)
```
michael-yxchen: Lmk your thoughts on this

omegablitz: This looks way nicer than before!

--


I would personally bias towards towards specifying the used subcomponents (along with mutability) explicitly. I think this would make it easier to audit what pieces of MonadState each substate uses.

For UX, the constructor for ConsensusChildState could remain taking a `&'mut MonadState`, and get destructured into the used fields in the implementation.

```rust
struct ConsensusChildState<'a>{
    state: &'a mut ConsensusState,
    a: &'a A,
    b: &'a B,
    c: &'a mut C,
}

impl <'a> ConsensusChildState<'a> {
    fn new(&'a mut MonadState) -> Self {
        Self {
            state: state.consensus_state,
            a: state.a,
            b: state.b,
            c: state.c,
        }
   }
}
```

I know @ariqchowdhury mentioned that he may feel differently about this, will defer to what you guys think.

ariqchowdhury: Having the constructor take in `&'a mut MonadState` like that looks good.

michael-yxchen: Cool will do

PR #486, opened 2024-01-09
Add virtual connection gossip abstraction

omegablitz:


monad-gossip/src/connection_manager.rs
```diff
@@ -0,0 +1,287 @@
...
+    max_message_size: u32,
+
+    current_tick: Duration,
+    connections: HashMap<NodeId, Connection>,
+    gossip: G,
+}
+
+type MessageLenType = u32;
+const MESSAGE_HEADER_LEN: usize = std::mem::size_of::<MessageLenType>();
+
```
ariqchowdhury, line 31: nit: maybe add some comments about connection being needed for framing over the quic connection

PR #487, opened 2024-01-10
Unix socket IPC executor

michael-yxchen:
Real IPC is ported from mempool-ipc. Added a placeholder mock ipc executor that doesn't nothing right now. We can use it to generate user transactions for testing.

### DIFF
--- a/monad-executor-glue/src/convert/event.rs
+++ b/monad-executor-glue/src/convert/event.rs
@@ -1,9 +1,12 @@
+use alloy_rlp::Decodable;
+use bytes::BytesMut;
 use monad_consensus_types::signature_collection::SignatureCollection;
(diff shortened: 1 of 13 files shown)


monad-executor-glue/src/convert/event.rs
```diff
@@ -7,8 +8,9 @@ use monad_proto::{
     error::ProtoError,
     proto::{basic::ProtoPubkey, event::*, validator_data::ProtoValidatorData},
 };
+use reth_primitives::TransactionSigned;
```
ariqchowdhury: nit: I added type-aliases in `monad-eth-types` to alias the reth-primitives I used (https://github.com/monad-crypto/monad-bft/blob/master/monad-eth-types/src/lib.rs#L11)

PR #495, opened 2024-01-16
Factor out secp and bls into separate crates

omegablitz:
This allows us to reduce some of our dependency spaghetti. Expanded out the macro-based SignatureCollection and CertificateSignature tests and put the tests in the new implementation crates. NopSignature also now matches behavior of SecpSignature, just without any cryptographic guarantees.


monad-secp/src/secp.rs
```diff
@@ -1,8 +1,7 @@
+use monad_crypto::hasher::{Hasher, HasherType};
 use secp256k1::Secp256k1;
 use zeroize::Zeroize;
 
-use crate::hasher::{Hashable, Hasher, HasherType};
-
 /// secp256k1 public key
 #[derive(Copy, Clone)]
 pub struct PubKey(secp256k1::PublicKey);
```
michael-yxchen, line 7: nit: now we have a trait named PubKey as well, shall we rename this to `SecpPubKey` to dedupe

omegablitz: Yes we definitely should.. I think there's other stuff that should also be renamed, eg CertificateSignature is no longer just used for certificates.

Am planning on doing the renaming stuff in batch in a future commit to reduce the amount of stuff here

PR #499, opened 2024-01-19
Delete State trait

omegablitz:
This is the initial step in simplifying the SwarmRelation trait. The observation is that we aren't ever using the State trait outside of MonadState.


monad-executor/src/replay_nodes.rs
ariqchowdhury, line 1: can you clarify why this is deleted now?

omegablitz: I just deleted it because it wasn't used anywhere

omegablitz: I assumed it was just a relic of the past and got replaced by something

PR #500, opened 2024-01-19
Loopback executor

michael-yxchen:
The loopback executor forwards events from one childstate to another in MonadState. Asking the lo executor to forward reduces potential inter-dependencies among child states. Due to replay, the handler of the event must be idempotent.

The executor prepares for async state verification (ASV). The ASV module is only tasked to collect state roots from execution and other peers. Once it reaches agreement, it forwards a StateRootUpdate to ConsensusState. There's minimal change to ConsensusState - we change the source of state root changes from execution to async state verification via loopback. That's it.

An alternative way to approach async state verification is to move state root validator to ASV. ConsensusState handlers take in an additional parameter and calls ASV to verify the state root. I don't think it's as elegant because it's nice to keep all consensus protocol under ConsensusState. If people have different/better ideas, lmk

### DIFF
--- a/monad-consensus-types/src/signature_collection.rs
+++ b/monad-consensus-types/src/signature_collection.rs
@@ -54,7 +54,7 @@ pub trait SignatureCollection:
     Clone + Hashable + Eq + Send + Sync + std::fmt::Debug + 'static
 {
     type NodeIdPubKey: PubKey;
-    type SignatureType: CertificateSignature;
+    type SignatureType: CertificateSignature + Unpin;
 
     fn new(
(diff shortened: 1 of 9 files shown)


ariqchowdhury:
To double-check my understanding of this PR. If there are sub-states A and B under MonadState and A wants to send an Event to B, it will output a LoopbackCommand::Forward(Event) to the LoopbackExecutor which then sends event to B. 
And for the ASV feature, the asv_state will receive StateRootHash update events from some executor and will be broadcasting out messages. Eventually it will receive a certificate at which point it will output a loopback command to forward the consensus_stateroothash_update event to consensus_state

michael-yxchen:
Yeah precisely

PR #504, opened 2024-01-22
Async state verify boilerplate & LocalAsyncStateVerify

michael-yxchen:
Added a new field `async_state_verify` to MonadState. It implements `AsyncStateVerifyProcess` to handle local state root updates from execution (StateRootHashUpdater in mock environment) and state roots from peers.

LocalAsyncStateVerify doesn't run any async state verify algorithm. It simply forwards the execution state root update to consensus. This is the previous behavior of our consensus algorithm.


monad-validator/src/epoch_manager.rs
```diff
@@ -55,4 +55,9 @@ impl EpochManager {
 
         *epoch_start.0
     }
+
+    /// Get the epoch of given sequence number
+    pub fn get_epoch_by_seq_num(&self, seq_num: SeqNum) -> Epoch {
+        Epoch((seq_num / self.val_set_update_interval).0 + 1)
```
bamrith: All the blocks whose sequence number is after the `val_set_update_interval` and are proposed before `epoch_start_delay` rounds are also in the same epoch.

michael-yxchen: Good catch. Is fetching Epoch number only possible through round number? I reckon it's possible to add another map to map sequence number to Epoch, but it'd make things much more complex here

bamrith: If there is a map for sequence number to Epoch, then it should be updated every time a block is committed (since branching can cause two blocks with the same sequence to be in 2 different Epochs and we can't be sure which Epoch the sequence number will belong to until one of the blocks is committed). If there is a way to get the round number from the actual block here, that would probably be easier.

michael-yxchen: Cool, I'll add a Round parameter to the event for now. We can revisit if it becomes an issue during integration. Eth block doesn't have consensus round number in them. Maybe we can add it if it's helpful in other places as well

michael-yxchen:
Will split into smaller PRs later

PR #505, opened 2024-01-23
Adding pipeline to build and publish docker image when commit to master

mhumeSF:
Publishes images to `ghcr.io/monad-crypto/monad-bft:<short-sha>` and `ghcr.io/monad-crypto/monad-bft:master`


.github/workflows/publish.yaml
```diff
@@ -0,0 +1,46 @@
+name: Create and publish a Docker image
+
+on:
+  push:
+    branches: ['release']
```
ariqchowdhury: should this be 'master'?

mhumeSF: Yes it should

PR #520, opened 2024-01-30
Augment state_root_hash_validation return value

michael-yxchen:
Switch the return value from a boolean to an enum StateRootAction. It targets the cases where local execution/state root verification is lagging/missing msgs.

The node doesn't have enough information to decide whether the state root is valid, so it can't vote on the block. However, it's safe to process other components in the block (X) - certificates, etc, and insert it to the block tree, instead of discarding it. If the next proposal (X+1) carrying the QC(X) arrives, it's proof that 2f+1 block have the state root and we can proceed without requesting block sync.

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -154,6 +154,21 @@ pub enum ConsensusAction {
     Abstain,
 }
 
+/// Actions after state root validation
+pub enum StateRootAction {
+    /// StateRoot validation is successful, proceed to next steps
+    Proceed,
+    /// StateRoot validation is unsuccessful - there's a mismatch of StateRoots.
+    /// Reject the proposal immediately
+    Reject,
+    /// StateRoot validation is undecided - we haven't collect enough
+    /// information. Either the state root is missing because we haven't
+    /// received a majority quorum on the state root, or the state root is
+    /// out-of-range. It's ok to insert to the block tree and observe if a QC
+    /// forms on the block. But we shouldn't vote on the block
+    Defer,
+}
+
 impl<ST, SCT, BVT, SVT> ConsensusState<ST, SCT, BVT, SVT>
 where
     ST: CertificateSignatureRecoverable,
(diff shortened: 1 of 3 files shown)


monad-consensus-state/src/lib.rs
```diff
@@ -776,34 +801,42 @@ where
...
-                false
+                StateRootAction::Defer
             }
-            // Don't vote and locally timeout if the proposed state root does not match
-            // or if state root is missing
-            StateRootResult::Missing | StateRootResult::Mismatch => {
+            // Don't vote and locally timeout if the proposed state root does
+            // not match
+            StateRootResult::Mismatch => {
                 metrics.consensus_events.rx_bad_state_root += 1;
```
ariqchowdhury, line 838: we should probably add debug logs to each of these cases while we're changing this

michael-yxchen:
Plan to add the mock-swarm test after we have the test framework

monad-consensus-state/src/lib.rs
```diff
@@ -776,34 +801,47 @@ where
...
-            // to try and catch up faster. For now, just wait
+            // TODO-1 execution lagging too far behind should be a trigger for
+            // something to try and catch up faster. For now, just wait
             StateRootResult::OutOfRange => {
+                debug!(
+                    "Proposal Message carries state root hash for a block higher than highest locally known, unable to verify"
+                );
                 metrics.consensus_events.rx_execution_lagging += 1;
-                false
+                StateRootAction::Defer
```
ariqchowdhury, line 832: Is Defer the correct action if execution is too far behind (D or more blocks behind)? The expectation here is that we should state-sync instead of blocksync, so does it make sense to still include the block into the blocktree?

michael-yxchen: I think it's a state sync design question. After async state verification is merged, the if statement is hit if it hasn't collected a quorum of state roots for the last D blocks, which might be caused by network drops. 

Say execution stops after executing block B, I think the earliest point to start state sync would be observing a QC on block B+D, and the latest point is when block B+D is committed. 

If we lean toward triggering state sync earlier, it doesn't hurt to insert the block to the block tree imo, because node will be listening for new blocks while state syncing. If we trigger state sync later, after block B+D is committed, then we should insert to block tree.

PR #523, opened 2024-01-30
Implement async state root verification

michael-yxchen:
This is a giant PR. It's split into hopefully digestible size commits. Each individual commit should pass CI, so lmk if it's easier to review separate PRs over separate commits in a giant PR

### DIFF
--- a/monad-async-state-verify/src/peer.rs
+++ b/monad-async-state-verify/src/peer.rs
@@ -1,0 +1,686 @@
+use std::{
+    collections::{BTreeMap, HashMap, HashSet},
+    marker::PhantomData,
+};
+
+use monad_consensus_types::{
+    signature_collection::{
+        SignatureCollection, SignatureCollectionError, SignatureCollectionKeyPairType,
+    },
+    state_root_hash::StateRootHashInfo,
+    voting::ValidatorMapping,
+};
+use monad_crypto::{
+    certificate_signature::CertificateSignature,
+    hasher::{Hash, Hasher, HasherType},
+};
+use monad_types::{NodeId, SeqNum, Stake};
+use monad_validator::validator_set::ValidatorSetType;
+use tracing::{info, warn};
+
+use crate::{
+    AsyncStateVerifyCommand, AsyncStateVerifyProcess, AsyncStateVerifyUpdateConfig,
+    StateRootCertificate,
+};
+
+pub fn majority_threshold(total_stake: Stake) -> Stake {
+    Stake(total_stake.0 / 2 + 1)
+}
+
+/// Collects votes on execution state root for a block
+#[derive(Debug, Clone)]
+struct ExecutionStateRootRecord<SCT: SignatureCollection> {
+    /// Execution state root computed by the local execution client
+    local_root: Option<StateRootHashInfo>,
+    /// Votes on state root collected from peers, indexed by the StateRootInfo
+    /// they voted on
+    pending_roots: HashMap<Hash, BTreeMap<NodeId<SCT::NodeIdPubKey>, SCT::SignatureType>>,
+    /// Tracks which state root each node has voted on, to detect double voting
+    node_vote_tracker: HashMap<NodeId<SCT::NodeIdPubKey>, HashSet<SCT::SignatureType>>,
+    /// Populated if a stake threshold of validators has voted on the same state
+    /// root
+    quorum: Option<StateRootCertificate<SCT>>,
+    /// Whether we have sent a state root update to consensus state
+    update_sent: bool,
+}
+
+impl<SCT: SignatureCollection> Default for ExecutionStateRootRecord<SCT> {
+    fn default() -> Self {
+        Self {
+            local_root: Option::None,
+            pending_roots: HashMap::new(),
+            node_vote_tracker: HashMap::new(),
+            quorum: None,
+            update_sent: false,
+        }
+    }
+}
+
+// TODO: garbage clean old state roots
+#[derive(Debug, Clone)]
+pub struct PeerAsyncStateVerify<SCT, VT>
+where
+    SCT: SignatureCollection,
+    VT: ValidatorSetType<NodeIdPubKey = SCT::NodeIdPubKey>,
+{
+    /// Collection of StateRoots after executing each block
+    state_roots: BTreeMap<SeqNum, ExecutionStateRootRecord<SCT>>,
+    /// The highest sequence number where a quorum of validators agree on the
+    /// execution state root
+    latest_certified: SeqNum,
+    /// StateRoot quorum threshold, takes in the total stake and computes the
+    /// threshold stake
+    quorum_threshold: fn(Stake) -> Stake,
+    /// Control when consensus update is emitted
+    consensus_update_config: AsyncStateVerifyUpdateConfig,
+    _phantom: PhantomData<(SCT, VT)>,
+}
+
+impl<SCT, VT> Default for PeerAsyncStateVerify<SCT, VT>
+where
+    SCT: SignatureCollection,
+    VT: ValidatorSetType<NodeIdPubKey = SCT::NodeIdPubKey>,
+{
+    fn default() -> Self {
+        Self {
+            state_roots: Default::default(),
+            latest_certified: SeqNum(0),
+            quorum_threshold: majority_threshold,
+            consensus_update_config: AsyncStateVerifyUpdateConfig::Local,
+            _phantom: Default::default(),
+        }
+    }
+}
+
+impl<SCT, VT> PeerAsyncStateVerify<SCT, VT>
+where
+    SCT: SignatureCollection,
+    VT: ValidatorSetType<NodeIdPubKey = SCT::NodeIdPubKey>,
+{
+    pub fn new(quorum_threshold: fn(Stake) -> Stake) -> Self {
+        Self {
+            quorum_threshold,
+            ..Default::default()
+        }
+    }
+}
+
+fn handle_invalid_signature_collection_creation<SCT: SignatureCollection>(
+    err: SignatureCollectionError<SCT::NodeIdPubKey, SCT::SignatureType>,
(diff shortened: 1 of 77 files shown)


monad-ledger/src/lib.rs
```diff
@@ -135,10 +135,10 @@ fn generate_header<SCT: SignatureCollection>(
     randao_reveal_hasher.update(monad_block.payload.randao_reveal.0);
 
     Header {
-        parent_hash: FixedBytes(parent_hash.0),
+        parent_hash: FixedBytes(parent_hash.0 .0),
```
ariqchowdhury: nit: can we add a helper function to get the underlying bytes directly? I don't like `foobar.0 .0`

monad-async-state-verify/src/peer.rs
```diff
@@ -0,0 +1,483 @@
...
+        let entry = self.state_roots.entry(info.seq_num).or_default();
+
+        // note the local execution result
+        if let Some(local_root) = &entry.local_root {
+            if local_root != &info {
+                warn!("conflicting local state root");
+            } else {
+                info!("duplicate local state root");
+            }
+        }
```
ariqchowdhury: This is handling the case that execution has sent the `StateRootHashInfo` more than once? Is there any other case this can happen?

monad-async-state-verify/src/peer.rs
```diff
@@ -0,0 +1,483 @@
...
+        entry.local_root = Some(info);
+
+        if let Some(cert) = &entry.quorum {
+            // TODO-2: handle the state root mismatch error
+            if cert.info != info {
+                warn!(
+                    "local execution result differ from quorum seq_num={:?}",
+                    info.seq_num
+                );
```
ariqchowdhury: I think its reasonable to panic right now. If we are getting state root mismatches in testing and testnet, it should fail loudly. Even in production that could be reasonable behaviour (restart node if the state root mismatches), but should leave the TODO so that we revisit the behaviour

monad-async-state-verify/src/peer.rs
```diff
@@ -0,0 +1,483 @@
...
+use monad_validator::validator_set::ValidatorSetType;
+use tracing::{info, warn};
+
+use crate::{AsyncStateVerifyCommand, AsyncStateVerifyProcess, StateRootCertificate};
+
+#[derive(Debug, Clone)]
+struct BlockStateRootState<SCT: SignatureCollection> {
+    local_root: Option<StateRootHashInfo>,
+    pending_roots: HashMap<Hash, BTreeMap<NodeId<SCT::NodeIdPubKey>, SCT::SignatureType>>,
+    node_roots: HashMap<NodeId<SCT::NodeIdPubKey>, HashSet<SCT::SignatureType>>,
```
ariqchowdhury: nit: could this be called pending_votes instead? or pending_sigs?

monad-async-state-verify/src/peer.rs
```diff
@@ -0,0 +1,483 @@
...
+            // it's correcting state root based on msgs received. Can bump the
+            // threshold slightly higher
+        }
+
+        // add to pending votes
+        let info_hash = HasherType::hash_object(&info);
+        let pending_roots = block_state.pending_roots.entry(info_hash).or_default();
+        pending_roots.insert(peer, sig);
+
+        while validators.has_majority_votes(&pending_roots.keys().copied().collect::<Vec<_>>()) {
```
ariqchowdhury: it might be better for this to configureable. we might want to set the threshold higher or maybe some validators want to receive more votes

monad-async-state-verify/src/peer.rs
```diff
@@ -0,0 +1,483 @@
...
+                }
+
+                Err(
+                    SignatureCollectionError::NodeIdNotInMapping(_)
+                    | SignatureCollectionError::ConflictingSignatures(_)
+                    | SignatureCollectionError::InvalidSignaturesVerify
+                    | SignatureCollectionError::DeserializeError(_),
+                ) => {
+                    unreachable!("InvalidSignaturesCreate is only expected error from creating SC");
+                }
```
ariqchowdhury: not related to this PR, but a future todo, maybe SIgnatureCollectionError should be split into creation and verification errors

monad-async-state-verify/src/peer.rs
```diff
@@ -0,0 +1,483 @@
...
+        self_id: NodeId<<Self::SignatureCollectionType as SignatureCollection>::NodeIdPubKey>,
+        cert_keypair: &SignatureCollectionKeyPairType<Self::SignatureCollectionType>,
+        info: StateRootHashInfo,
+    ) -> Vec<AsyncStateVerifyCommand<Self::SignatureCollectionType>> {
+        let entry = self.state_roots.entry(info.seq_num).or_default();
+
+        // note the local execution result
+        if let Some(local_root) = &entry.local_root {
+            if local_root != &info {
+                warn!("conflicting local state root");
```
ariqchowdhury: should log the conflicting values if we hit this

monad-async-state-verify/src/peer.rs
```diff
@@ -0,0 +1,483 @@
...
+            } else {
+                info!("duplicate local state root");
```
ariqchowdhury: should log the state root here. 
I think this should be a warning? I don't think its expected behaviour (unless I'm missing something)

monad-async-state-verify/src/peer.rs
```diff
@@ -0,0 +1,483 @@
...
+                pending_roots.iter().map(|(node, sig)| (*node, *sig)),
+                validator_mapping,
+                info_hash.as_ref(),
+            ) {
+                Ok(sigcol) => {
+                    // forward certified state root to consensus
+                    self.latest_certified = info.seq_num;
+                    if let Some(local_root) = block_state.local_root {
+                        if local_root != info {
+                            warn!("conflicting local state root");
```
ariqchowdhury: print the conflicting state roots

monad-async-state-verify/src/peer.rs
```diff
@@ -0,0 +1,483 @@
...
+            <Self::SignatureCollectionType as SignatureCollection>::NodeIdPubKey,
+            SignatureCollectionKeyPairType<Self::SignatureCollectionType>,
+        >,
+    ) -> Vec<AsyncStateVerifyCommand<Self::SignatureCollectionType>> {
+        // TODO: anti-spam: don't accept state root hash much higher than
+        // current consensus value
+        let mut cmds = Vec::new();
+        if info.seq_num <= self.latest_certified {
+            return cmds;
+        }
```
ariqchowdhury: if node A is on seq_num(N) and other nodes are at N+1 and the messages for N+1 arrive faster than N, node A will no longer gather sigs to create a certificate for N. And then if a user/client asks for proof of state-root hash at block-seq_num(N), it won't get it from this node (even though the node could have collected sigs and proved it for N)

monad-async-state-verify/src/peer.rs
```diff
@@ -0,0 +1,483 @@
...
+            ) {
+                Ok(sigcol) => {
+                    // forward certified state root to consensus
+                    self.latest_certified = info.seq_num;
+                    if let Some(local_root) = block_state.local_root {
+                        if local_root != info {
+                            warn!("conflicting local state root");
+                        }
+                    } else {
+                        info!("local execution delayed");
```
ariqchowdhury: should print the certified seq_num and this node's local_root seq_num so we can see how much its delayed.

monad-async-state-verify/src/peer.rs
```diff
@@ -0,0 +1,483 @@
...
+    hasher::{Hash, Hasher, HasherType},
+};
+use monad_types::{NodeId, SeqNum};
+use monad_validator::validator_set::ValidatorSetType;
+use tracing::{info, warn};
+
+use crate::{AsyncStateVerifyCommand, AsyncStateVerifyProcess, StateRootCertificate};
+
+#[derive(Debug, Clone)]
+struct BlockStateRootState<SCT: SignatureCollection> {
```
ariqchowdhury: nit: naming....maybe ExecutedStateRoots?

monad-mock-swarm/src/transformer.rs
```diff
@@ -75,6 +78,8 @@ pub struct TwinsTransformer<PT: PubKey> {
     default_part: Vec<ID<PT>>,
     // block sync and associated message is dropped by default
     ban_block_sync: bool,
+    // state root verificastion messages is forwarded by default
```
ariqchowdhury: typo

monad-async-state-verify/src/peer.rs
michael-yxchen: thanks for catching it. I'll remove the check and log if we receive multiple state root hashes from a peer

monad-crypto/src/hasher.rs
```diff
@@ -1,12 +1,22 @@
...
 
 use sha2::Digest;
 use zerocopy::AsBytes;
 
 /// A 32-byte/256-bit hash
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Default)]
+#[derive(Clone, Copy, PartialEq, Eq, Hash, Default)]
 pub struct Hash(pub [u8; 32]);
 
+impl Debug for Hash {
```
ariqchowdhury, line 10: nit: should we use this https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.alternate
so that we can still debug print the full hash when necessary

monad-crypto/src/lib.rs
```diff
@@ -12,9 +14,19 @@ pub struct NopKeyPair {
     pubkey: NopPubKey,
 }
 
-#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
+#[derive(Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub struct NopPubKey([u8; 32]);
 
+impl Debug for NopPubKey {
```
ariqchowdhury, line 20: same for pubkey

PR #525, opened 2024-01-31
Interleaved weighted round robin leader election

rgarc:
Implements an algorithm known as [interleaved weighted round robin](https://en.wikipedia.org/wiki/Weighted_round_robin#Interleaved_WRR) to be used in leader election. 

Traditionally, this algorithm is used in the context of network scheduling, but it can be adapted for leader election. We simply let each queue $q_N$ hold the same number of elements as the stake of validator $N$. Similarly, the weight of each queue $w_N$ is equivalent to the stake of validator $N$.

### Examples
Let $S_i$ be the stake of validator $V_i$ and the validator set represented as $\left[(V_1, S_1), (V_2, S_2), ..., (V_n, S_n)\right]$. Interleaved weighted round robin will yield the following schedule for the following validator sets.
- $[(A, 1), (B, 1), (C, 1)] \longrightarrow [A, B, C]$
- $[(A, 1), (B, 1), (C, 0)] \longrightarrow [A, B]$
- $[(A, 1), (B, 2), (C, 1)] \longrightarrow [A, B, C, B]$
- $[(A, 1), (B, 2), (C, 3)] \longrightarrow [A, B, C, B, C, C]$
- $[(A, 2), (B, 2), (C, 2)] \longrightarrow [A, B, C, A, B, C]$
- $[(A, 2), (B, 4)] \longrightarrow [A, B, A, B, B, B]$

### DIFF
--- a/monad-validator/Cargo.toml
+++ b/monad-validator/Cargo.toml
@@ -17,3 +17,4 @@ tracing = { workspace = true }
 
 [dev-dependencies]
 monad-testutil = { path = "../monad-testutil" }
+test-case = { workspace = true }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2920,6 +2920,7 @@ dependencies = [
  "monad-crypto",
  "monad-testutil",
  "monad-types",
+ "test-case",
  "tracing",
 ]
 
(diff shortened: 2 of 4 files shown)


monad-validator/Cargo.toml
```diff
@@ -14,6 +14,7 @@ monad-crypto = { path = "../monad-crypto" }
 monad-types = { path = "../monad-types" }
 
 tracing = { workspace = true }
+test-case = { workspace = true }
```
omegablitz: dev-dependencey

PR #528, opened 2024-01-31
Change sender in MonadEvent to NodeId

michael-yxchen:
MonadState uses NodeId to identify its peers/sender of the message. We are unwrap then immediately wrap the pubkeys with NodeId. The change removes that step.

### DIFF
--- a/monad-executor-glue/src/convert/event.rs
+++ b/monad-executor-glue/src/convert/event.rs
@@ -1,7 +1,6 @@
 use monad_consensus_types::signature_collection::SignatureCollection;
 use monad_crypto::certificate_signature::CertificateSignatureRecoverable;
 use monad_proto::{error::ProtoError, proto::event::*};
-use monad_types::convert::{proto_to_pubkey, pubkey_to_proto};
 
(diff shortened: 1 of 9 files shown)


monad-executor-glue/src/convert/event.rs
```diff
@@ -1,7 +1,6 @@
 use monad_consensus_types::signature_collection::SignatureCollection;
 use monad_crypto::certificate_signature::CertificateSignatureRecoverable;
 use monad_proto::{error::ProtoError, proto::event::*};
-use monad_types::convert::{proto_to_pubkey, pubkey_to_proto};
```
ariqchowdhury, line 4: can we also delete those functions now?

michael-yxchen: The function is still used in 2 places, converting consensus NodeId pubkey and certificate pubkey. I can remove it if it's ok to duplicate a little

ariqchowdhury: oh, nvm then, didn't realize it was still used

PR #530, opened 2024-01-31
Refactor LeaderElection trait.

rgarc:
- Introduce `fn update(&mut self, event: Vec<(NodeId<Self::NodeIdPubKey>, Stake)>)` to the `LeaderElection` trait to enable informed handling of validator set changes.
- Make `LeaderElection::get_leader` only a function of `Round`.

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -263,8 +263,7 @@ where
 
         // author, leader, round checks
         let round = self.pacemaker.get_current_round();
-        let block_round_leader =
-            election.get_leader(block.get_round(), epoch, validator_set.get_members());
+        let block_round_leader = election.get_leader(block.get_round());
         if block.get_round() > round
             || author != block_round_leader
             || block.get_author() != block_round_leader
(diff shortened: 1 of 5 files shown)


monad-consensus-state/src/lib.rs
```diff
@@ -316,15 +315,15 @@ where
...
                 let next_round = round + Round(1);
                 let next_epoch = epoch_manager.get_epoch(next_round);
                 let Some(next_validator_set) = val_epoch_map.get_val_set(&next_epoch) else {
                     todo!("handle non-existent validatorset for next round epoch");
                 };
-                (next_round, next_epoch, next_validator_set.get_members())
+                next_round
             };
-            let next_leader = election.get_leader(next_round, next_epoch, next_validator_set);
+            let next_leader = election.get_leader(next_round);
```
bamrith, line 326: The validator set used here may not be the same as the set used to get the leader of the current round. It happens when the next round is in a new epoch. So, the next validator set that was fetched from the `val_epoch_map` should be used instead.

rgarc: Thanks, that makes sense.

PR #533, opened 2024-02-01
Track running tx count in ledgewatch

ariqchowdhury:


monad-ledger/examples/ledgewatch.rs
```diff
@@ -26,6 +26,7 @@ fn main() -> io::Result<()> {
 
     let mut cursor = KECCAK_HDR_LEN;
     let mut running_hash = Hash::default();
+    let mut running_cnt = 0;
```
omegablitz: nit: 0_u64 just in case? Max i32 is ~2 billion, which is ~55 hours under 10k tps
Probably won't happen, but could be a future footgun

PR #536, opened 2024-02-03
Resolve IPC receiver starvation issues

ariqchowdhury:
IPC receiver gets its own task for polling the underlying UDS socket. Router is moved to the end of the future::select_all list Router poll has a budget to make sure it periodically yields

### DIFF
--- a/monad-ipc/src/lib.rs
+++ b/monad-ipc/src/lib.rs
@@ -1,7 +1,7 @@
-use std::{marker::PhantomData, path::Path, task::Poll};
+use std::{path::PathBuf, task::Poll};
 
 use alloy_rlp::Decodable;
-use futures::{stream::SelectAll, Stream, StreamExt};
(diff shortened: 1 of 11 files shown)


monad-ipc/src/lib.rs
```diff
@@ -44,22 +39,79 @@ pub fn generate_uds_path() -> String {
...
+                        };
+                        txns.push(bytes);
+                    }
+                    Some(Err(err)) => {
+                        warn!("framed reader error err={:?}", err);
+                        break;
+                    }
+                    None => {
+                        debug!("done reading");
+                        break;
```
omegablitz, line 109: Just noting that this only flushes to the event_channel once the IPC socket is closed, rather than after the stream yields.

This is fine for now for the purposes of ethtx-tool, but we may want to change this once we start testing with the RPC service. Otherwise, the RPC server will need to close/reopen its socket to flush transactions.

PR #545, opened 2024-02-07
Version check the consensus protocol messages

ariqchowdhury:
The consensus protocol and algorithm could change without any of the protocol messages acutally changing their structure so we should have a way of checking for versions. This aids in upgradability of the consensus algorithm.

The consensus protocol is neither backwards nor forwards compatible so we just use an exact string match to determine if a message should be validated.

This patch also includes the structure for a client major-minor version number but it is not checked yet.

### DIFF
--- a/monad-consensus/src/messages/consensus_message.rs
+++ b/monad-consensus/src/messages/consensus_message.rs
@@ -14,7 +14,7 @@ use crate::{
 
 /// Consensus protocol messages
 #[derive(Clone, PartialEq, Eq)]
-pub enum ConsensusMessage<SCT: SignatureCollection> {
+pub enum ProtocolMessage<SCT: SignatureCollection> {
     /// Consensus protocol proposal message
     Proposal(ProposalMessage<SCT>),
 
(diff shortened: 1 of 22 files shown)


monad-consensus/src/messages/consensus_message.rs
```diff
@@ -52,18 +52,34 @@ where
...
+    pub message: ProtocolMessage<SCT>,
+}
+
+impl<SCT> Hashable for ConsensusMessage<SCT>
+where
+    SCT: SignatureCollection,
+{
+    fn hash(&self, state: &mut impl Hasher) {
+        state.update(self.version.clone());
+        self.message.hash(state);
```
michael-yxchen, line 79: nit: can avoid cloning by taking reference

ariqchowdhury:
> You've probably thought about this: not being backward-compatible means that we can't replay events from a previous protocol version. Just want to note it down for upgradability

Yup, that's the intended behaviour. If a node upgrades to a new protocol version, I don't think that node would want to replay all the events from the previous version and potentially send out messages related to the older version.

PR #548, opened 2024-02-08
Add interface for metrics verification

bamrith:


monad-mock-swarm/src/verifier.rs
```diff
@@ -71,6 +135,40 @@ impl MockSwarmVerifier {
             ExpectedTick::None => {}
         }
 
+        for (node_id, fetch_metric, expected_metric) in self.metrics.iter() {
+            let node = swarm.states.get(node_id).unwrap();
+            let actual_metric = fetch_metric(node.state.metrics());
+            match expected_metric {
+                ExpectedMetric::Exact(metric) => {
+                    if actual_metric != *metric {
+                        eprintln!(
```
michael-yxchen, line 154: It's better to print the metric name here with the values

monad-mock-swarm/tests/single_node.rs
```diff
@@ -82,14 +82,13 @@ fn two_nodes_noser() {
         .is_some()
     {}
 
-    let verifier = MockSwarmVerifier::default().tick_exact(Duration::from_secs(10));
+    let n1 = *swarm.states().first_key_value().unwrap().0;
+    let verifier = MockSwarmVerifier::default()
+        .tick_exact(Duration::from_secs(10))
+        .metric_exact(n1, fetch_metrics!(consensus_events.commit_empty_block), 0);
```
michael-yxchen: Asserting a certain metrics for all nodes is useful as well

monad-mock-swarm/src/verifier.rs
```diff
@@ -71,6 +135,40 @@ impl MockSwarmVerifier {
...
+                            "Metric verify error expected={:?} actual={:?}",
+                            metric, actual_metric
+                        );
+                        return false;
```
michael-yxchen: I'm thinking that we should run through all the assertions instead of early returning, so we know everything that's failing

monad-mock-swarm/src/verifier.rs
bamrith: Added metric name and node ID to the error

monad-mock-swarm/tests/single_node.rs
bamrith: Changed input as vector of nodes instead of just one node

PR #550, opened 2024-02-09
Add tick assertions to mock-swarm tests

michael-yxchen:
More lenient about QUIC router scheduler tests. It takes a few more round trips to produce the same number of blocks comparing to NoserRouterScheduler. Some are handshake while others are harder to decipher. Leaving it there because the router scheduler isn't used in production.


monad-bls/src/lib.rs
```diff
@@ -17,6 +17,20 @@ impl Hashable for BlsAggregateSignature {
...
+impl std::fmt::Display for BlsPubKey {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let bytes = self.bytes();
+        write!(
+            f,
+            "{:>02x}{:>02x}..{:>02x}{:>02x}",
+            bytes[0],
+            bytes[1],
+            bytes[bytes.len() - 2],
+            bytes[bytes.len() - 2]
```
omegablitz: should this be `bytes[bytes.len() - 1]`?

michael-yxchen: oops yeah you're right

PR #551, opened 2024-02-09
Record metrics

rgarc:
We want to have a metrics executor that is responsible for recording [metrics](https://github.com/monad-crypto/monad-bft/blob/master/monad-consensus-types/src/metrics.rs#L64-L64). This PR lays the groundwork for that by
- adding a new event type for a metrics timeout event that fires when it is time to record metrics
- adding a new command type for recording metrics
- adding a no-op metrics executor owned by [`ParentExecutor`](https://github.com/monad-crypto/monad-bft/blob/master/monad-updaters/src/parent.rs#L18-L18) for testing
- adding an OpenTelemetry executor that is responsible for exporting metrics to a back-end
- adds a `--record-metrics-interval-seconds` CLI arg to the `monad-node` binary which allows controlling the interval in which metrics are sent to the OpenTelemetry back-end. 

### TODOs before merging
- [x] fix wasm build
wasm build is broken because the 2nd commit in this PR added a dependency that made `monad-debugger` depend on `mio` which can't be compiled on wasm
```
monad-debugger v0.1.0 (/home/rgarc/monad-bft/monad-debugger)
└── monad-mock-swarm v0.1.0 (/home/rgarc/monad-bft/monad-mock-swarm)
    └── monad-updaters v0.1.0 (/home/rgarc/monad-bft/monad-updaters)
        └── opentelemetry v0.20.0
            └── opentelemetry_sdk v0.20.0
                └── tokio v1.34.0
                    └── mio v0.8.9
```
- [ ] refactor the OpenTelemetry executor to use some sort of Gauge type instead of `Counter<u64>`. The latter causes metrics to be summed together in the OpenTelemetry back-end which leads to nonsensical metrics graphs.

### DIFF
--- a/monad-state/src/lib.rs
+++ b/monad-state/src/lib.rs
@@ -35,7 +35,7 @@ use monad_crypto::certificate_signature::{
 use monad_eth_types::EthAddress;
 use monad_executor_glue::{
     AsyncStateVerifyEvent, BlockSyncEvent, Command, ConsensusEvent, MempoolEvent, Message,
-    MonadEvent, ValidatorEvent,
+    MetricsCommand, MetricsEvent, MonadEvent, ValidatorEvent,
 };
 use monad_types::{Epoch, NodeId, Round, SeqNum, TimeoutVariant};
 use monad_validator::{
@@ -427,6 +427,8 @@ where
             ))),
         );
 
+        init_cmds.extend(monad_state.update(MonadEvent::MetricsEvent(MetricsEvent::Timeout)));
+
         (monad_state, init_cmds)
     }
 }
@@ -505,6 +507,13 @@ where
                     .flat_map(Into::<Vec<Command<_, _, _, _, _>>>::into)
                     .collect::<Vec<_>>()
             }
+            MonadEvent::MetricsEvent(metrics_event) => match metrics_event {
+                MetricsEvent::Timeout => {
+                    vec![Command::MetricsCommand(MetricsCommand::RecordMetrics(
+                        self.metrics,
+                    ))]
+                }
+            },
         }
     }
 }
--- a/monad-updaters/src/parent.rs
+++ b/monad-updaters/src/parent.rs
@@ -9,13 +9,13 @@ use monad_consensus_types::signature_collection::SignatureCollection;
 use monad_executor::Executor;
 use monad_executor_glue::{
     CheckpointCommand, Command, ExecutionLedgerCommand, LedgerCommand, LoopbackCommand,
-    RouterCommand, StateRootHashCommand, TimerCommand,
+    MetricsCommand, RouterCommand, StateRootHashCommand, TimerCommand,
 };
 
 /// Single top-level executor for all other required by a node.
 /// This executor will distribute commands to the appropriate sub-executor
 /// and will poll them for events
-pub struct ParentExecutor<R, T, L, EL, C, S, IPC, LO> {
+pub struct ParentExecutor<R, T, L, EL, C, S, IPC, LO, M> {
     pub router: R,
     pub timer: T,
     pub ledger: L,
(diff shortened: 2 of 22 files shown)


monad-updaters/src/metrics/nop_metrics.rs
```diff
@@ -0,0 +1,38 @@
...
+    fn replay(&mut self, mut _commands: Vec<Self::Command>) {}
+
+    fn exec(&mut self, _commands: Vec<Self::Command>) {}
+}
+
+impl<E> Stream for NopMetricsExecutor<E> {
+    type Item = E;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        todo!()
```
rgarc: `todo!` will obviously crash, but is it safe to return `Poll::Pending` for a no-op `impl Stream` ?

monad-types/src/lib.rs
```diff
@@ -277,6 +277,29 @@ pub enum TimeoutVariant {
...
 }
 
+#[derive(Hash, Debug, Clone, PartialEq, Eq, Copy)]
+/// This type only exists to extend the `TimeoutVariant` enum with a timeout
+/// that is to be handled by a metrics executor (for periodic measurements).
+/// Naively adding a `Metrics` variant to `TimeoutVariant` above forces us
+/// to handle the `Metrics` variant in a place where it doesn't make sense.
+/// i.e. `ConsensusCommand` uses `TimeoutVariant` but it doesn't make sense
+/// to have an instance of `ConsensusCommand` with `Metrics`
+pub enum ExecutorTimeout {
```
rgarc: please lmk if the doc comment for this `ExecutorTimeout` makes sense

monad-updaters/src/parent.rs
```diff
@@ -105,13 +112,14 @@ where
             this.ipc.next().boxed_local(),
             this.loopback.next().boxed_local(),
             this.router.next().boxed_local(),
+            // TODO(rene): poll the metrics executor stream here once implemented
```
rgarc: not polling metrics executor yet due to this comment
https://github.com/monad-crypto/monad-bft/pull/551/files#r1486789527

monad-state/src/lib.rs
```diff
@@ -505,6 +505,16 @@ where
                     .flat_map(Into::<Vec<Command<_, _, _, _, _>>>::into)
                     .collect::<Vec<_>>()
             }
+            MonadEvent::MetricsEvent(metrics_event) => match metrics_event {
+                MetricsEvent::Timeout => vec![
+                    Command::MetricsCommand(MetricsCommand::RecordMetrics(self.metrics)),
+                    Command::TimerCommand(TimerCommand::Schedule {
+                        duration: Duration::from_secs(30), /* TODO(rene): load from config */
```
rgarc: instead of hard-coding 30s, where would be a sensible place to load this from?

monad-updaters/src/metrics/nop_metrics.rs
michael-yxchen: yeah returning Poll::Pending is fine. `select_all` skips the future that return pending and continues to poll the next future

monad-types/src/lib.rs
michael-yxchen: make sense to me

monad-state/src/lib.rs
```diff
@@ -505,6 +505,16 @@ where
                     .flat_map(Into::<Vec<Command<_, _, _, _, _>>>::into)
                     .collect::<Vec<_>>()
             }
+            MonadEvent::MetricsEvent(metrics_event) => match metrics_event {
+                MetricsEvent::Timeout => vec![
+                    Command::MetricsCommand(MetricsCommand::RecordMetrics(self.metrics)),
+                    Command::TimerCommand(TimerCommand::Schedule {
```
michael-yxchen: should we also schedule the timeout in `init`?

monad-state/src/lib.rs
michael-yxchen: Two ideas
1. if we are scheduling the metrics timeout in monad-state, it can go into `MonadStateBuilder`
2. if we think that putting the config in `MonadStateBuilder` isn't ideal, because it's more like a parameter to the `MetricsExecutor`, we can have the `MetricsExecutor` schedule this timeout in its `exec` function. Then this would probably be a cli arg passed to the executor's constructor

monad-state/src/lib.rs
rgarc: you're right. good catch!

monad-state/src/lib.rs
rgarc: will opt for the 2nd approach in an upcoming PR

PR #552, opened 2024-02-10
Replace tokio channel in ipc receiver with flume

ariqchowdhury:
tested on linux with single monad-node running and inserting txns via ethtx tool

### DIFF
--- a/monad-ipc/src/lib.rs
+++ b/monad-ipc/src/lib.rs
@@ -1,7 +1,8 @@
 use std::{path::PathBuf, task::Poll};
 
 use alloy_rlp::Decodable;
-use futures::{FutureExt, Stream, StreamExt};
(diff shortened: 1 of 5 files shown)


monad-ipc/src/lib.rs
```diff
@@ -117,10 +104,63 @@ where
...
+        };
+
+        tokio::spawn(async move {
+            let mut txns = Vec::with_capacity(buf_size);
+            let batch_timeout = tokio::time::sleep(Duration::from_secs(100));
+            tokio::pin!(batch_timeout);
+
+            loop {
+                tokio::select! {
+                    read = reader.next() => {
```
ariqchowdhury, line 117: wrap this in timeout instead of using sleep

PR #553, opened 2024-02-12
Add Raptor broadcast algorithm implementation

omegablitz:
This adds support for the Raptor-based broadcast implementation. It's built on top of a generic Seeder abstraction. Further details are in comments in the respective files.

This required adding support for QUIC datagrams, hence the changes inside monad-quic.

The monad-gossip/src/seeder changes are still quite rough and subject to change, but it's functional and performs well.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2913,6 +2913,7 @@ dependencies = [
  "monad-virtual-bench",
  "rand",
  "rand_chacha",
+ "raptorq",
  "serde",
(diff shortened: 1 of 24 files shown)


omegablitz:
Everything inside `monad-gossip/src/seeder` stands on its own and can be reviewed separately. I can split these into two separate PRs, but the datagram changes in monad-quic probably won't make sense otherwise.

PR #554, opened 2024-02-14
Add metrics assertions to mock swarm tests

bamrith:


monad-mock-swarm/src/verifier.rs
```diff
@@ -76,44 +77,213 @@ impl<S: SwarmRelation> MockSwarmVerifier<S> {
...
+            // number of blocks authored in the ledger <= number of rounds as leader
+            // NOTE: '<=' is used since blocks can be rejected
+            let num_blocks_authored = blocks_proposed.len() as u64;
+
+            // should handle proposal for all blocks in ledger (except genesis block)
+            self.metric_minimum(
+                &vec![*node_id],
+                fetch_metric!(consensus_events.handle_proposal),
+                ledger_len - 1,
+            );
```
michael-yxchen: The genesis block was removed in https://github.com/monad-crypto/monad-bft/pull/461. We can use `ledger_len` here

monad-mock-swarm/src/verifier.rs
```diff
@@ -76,44 +77,213 @@ impl<S: SwarmRelation> MockSwarmVerifier<S> {
...
+                fetch_metric!(consensus_events.created_qc),
+                num_blocks_authored - 1,
+            );
+            // a node processes an old QC (generated by itself) when it receives it
+            // in the proposal for next round
+            self.metric_minimum(
+                &vec![*node_id],
+                fetch_metric!(consensus_events.process_old_qc),
+                num_blocks_authored,
+            );
```
michael-yxchen, line 271: would there be a max on this metric?

monad-mock-swarm/tests/epoch.rs
```diff
@@ -314,14 +329,38 @@ mod test {
...
+                &running_nodes_ids,
+                fetch_metric!(consensus_events.created_vote),
+                update_block_num.0 + 9,
+            )
+            .metric_maximum(
+                &vec![blackout_node_id],
+                fetch_metric!(blocksync_events.blocksync_request),
+                2,
+            )
+            // initial TC + account for TC whenever emmitted messages
```
michael-yxchen: nit: typo

monad-mock-swarm/tests/epoch.rs
```diff
@@ -314,14 +329,38 @@ mod test {
             Epoch(2),
         );
 
-        // TODO: add tick and metrics assertions, expected below but it's
```
michael-yxchen, line 317: can we leave the TODO in if we haven't figured out the tick assertion?

monad-mock-swarm/tests/order.rs
```diff
@@ -127,27 +128,97 @@ fn all_messages_delayed(direction: TransformerReplayOrder) {
...
+        );
+
+    assert!(verifier_before_delayed_messages.verify(&swarm));
+
     while swarm
         .step_until(&mut UntilTerminator::new().until_tick(Duration::from_millis(1000)))
         .is_some()
     {}
 
-    // TODO: assert block sync is not triggered too many times the replaying
```
michael-yxchen, line 142: have you looked at how many block sync does the first node requests?

monad-mock-swarm/tests/rand_lat.rs
```diff
@@ -114,13 +116,32 @@ fn nodes_with_random_latency(latency_seed: u64) {
         .step_until(&mut UntilTerminator::new().until_block(2000))
         .is_some()
     {}
-    swarm_ledger_verification(&swarm, 1998);
+
+    let min_ledger_len = 1998;
```
michael-yxchen: Can we add a comment explaining why difference is 2?

monad-mock-swarm/tests/many_nodes.rs
```diff
@@ -143,13 +148,35 @@ fn many_nodes_quic_latency() {
 
     let mut swarm = swarm_config.build();
     swarm.batch_step_until(&mut UntilTerminator::new().until_block(100));
-    swarm_ledger_verification(&swarm, 100);
+
+    let min_ledger_len = 100;
```
michael-yxchen, line 154: Related: https://github.com/monad-crypto/monad-bft/pull/554#discussion_r1496383701
min ledger is 2 blocks shorter there but not here

monad-mock-swarm/tests/two_nodes.rs
```diff
@@ -167,7 +163,9 @@ fn two_nodes_quic_latency() {
         .step_until(&mut UntilTerminator::new().until_block(1000))
         .is_some()
     {}
-    swarm_ledger_verification(&swarm, 998);
+
+    let min_ledger_len = 998;
```
michael-yxchen: Related https://github.com/monad-crypto/monad-bft/pull/554#discussion_r1496383701

monad-mock-swarm/src/verifier.rs
bamrith: The other places where a node processes old QCs (except during proposal) is when it receives votes and timeouts. There is a max on processing old QCs within vote messages. i.e. when it receives `f` extra votes for a block after creating a QC with `2f+1` votes, and hence max would be `num_blocks_authored * f`. But, there is no limit on how many timeout messages it may receive from other nodes, so there is no exact max value overall.

monad-mock-swarm/tests/order.rs
bamrith: added expected `blocksync_requests` metric for each order

monad-mock-swarm/tests/epoch.rs
```diff
@@ -314,14 +331,40 @@ mod test {
...
-        // let verifier = MockSwarmVerifier::default().tick_range(
-        //     happy_path_tick_by_block(update_block_num.0 as usize + 10, delta),
-        //     delta,
-        // );
-        // assert!(verifier.verify(&nodes));
+        // during blackout, if the blackout node is a leader for a round,
+        // it doesn't collect votes or propose a block. this causes TCs to
+        // be formed in two consecutive rounds
+        // TODO: add tick assertions. need to account for blackout node consequences
+        // and using `step_until` while updating pipelines
```
michael-yxchen: what does "using `step_until` while updating pipelines" mean?

bamrith: when the pipeline is updated to blackout a node, it doesn't take effect immediately because `pending_inbound_messages` in the blackout node will already have messages. The opposite effect when the blackout is removed (the blackout node won't get messages for another tick). That may cause an issue when calculating expected tick.

michael-yxchen: gotcha thanks

PR #555, opened 2024-02-15
Commit one file per block

ariqchowdhury:
monad-ledger has two executor implementations to either commit all blocks into one file or commit each block into its own file.

blockwatch tool does the same thing as ledgewatch but for a folder of block files

### DIFF
--- a/monad-ledger/examples/blockwatch.rs
+++ b/monad-ledger/examples/blockwatch.rs
@@ -1,0 +1,123 @@
+use std::{
+    fs::File,
+    io::{self, Read, Seek},
+    path::{Path, PathBuf},
+};
+
+use alloy_rlp::Decodable;
+use clap::Parser;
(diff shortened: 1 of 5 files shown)


monad-ledger/examples/blockwatch.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    Ok(())
+}
+
+fn process_block(path: &PathBuf) -> (u64, u64, Hash) {
+    let mut f = File::open(path).unwrap();
+
+    let sz = f.metadata().unwrap().len() - KECCAK_HDR_LEN;
+    f.seek(io::SeekFrom::Start(KECCAK_HDR_LEN)).unwrap();
```
ariqchowdhury, line 106: this can probably be removed on the encoding side when we're using individual files

PR #558, opened 2024-02-16
Flush ipc receiver stream before batch is full after timeout

ariqchowdhury:


monad-ipc/src/lib.rs
```diff
@@ -137,3 +175,35 @@ where
         self.read_events_recv.poll_recv(cx)
     }
 }
+
+struct ArmedSleep {
+    sleep: Pin<Box<tokio::time::Sleep>>,
```
omegablitz, line 180: It doesn't really matter, but we represent this in the quic service executor with an `Option<Sleep>`, where it's armed if it's `Some(_)`.

This is fine as well though

PR #637, opened 2024-02-28
Add max retry count for blocksync

bamrith:


monad-consensus-state/src/blocksync.rs
```diff
@@ -181,7 +189,7 @@ where
     {
         let id = qc.get_block_id();
 
-        if self.requests.contains_key(&id) {
+        if self.requests.contains_key(&id) || req_cnt > self.max_retry_cnt {
```
michael-yxchen: let's log the two cases separately

PR #641, opened 2024-02-29
Flexnet changes after removing hash from ledger blocks

michael-yxchen:
Based on: #640


docker/flexnet/nets/net0/scripts/inspect-block.py
```diff
@@ -22,7 +22,7 @@
 
         with open(block_file, "rb") as f:
             # skip the block hash
```
ariqchowdhury: remove the comment

PR #643, opened 2024-03-01
Add JSON-RPC batch support.

rgarc:
This PR adds support to the JSON-RPC server for [batch requests](https://www.jsonrpc.org/specification#batch). This was encountered when pointing the RPC server to a local instance of an open-source block explorer [blockscout](https://github.com/blockscout/blockscout). Most of the initial web requests were batch requests which failed, ultimately breaking the front-end.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3159,6 +3159,7 @@ dependencies = [
  "alloy-rlp",
  "bytes",
  "clap 4.4.10",
+ "env_logger",
  "flume",
  "futures",
  "heed",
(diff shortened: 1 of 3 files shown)


monad-rpc/src/jsonrpc.rs
```diff
@@ -13,6 +13,34 @@ pub struct Request {
     pub id: Value,
 }
 
+#[derive(Deserialize, Serialize, Debug, PartialEq, Eq)]
+#[serde(untagged)]
+pub enum RequestWrapper<T> {
+    /// To be JSON-RPC spec-compliant, `Batch(Vec<T>)` needs to be the first variant in this enum.
```
rgarc, line 19: Let me know if this comment is unclear.

PR #651, opened 2024-03-05
Allow ethtx tool to send over JSON-RPC

rgarc:
This PR adds a new feature to the ethtx tool, enabling the sending of transactions over HTTP/JSON-RPC with customizable behavior. Two new options are introduced: `batch_size`, which divides the number of transactions into chunks for submission over JSON-RPC batch, and `concurrent`, which is a flag that allows issuing JSON-RPC HTTP requests concurrently.

### DIFF
--- a/monad-ipc/Cargo.toml
+++ b/monad-ipc/Cargo.toml
@@ -21,7 +21,7 @@ tokio = { workspace = true, features = [
     "net",
     "rt",
     "time",
-]}
+] }
 tokio-util = { workspace = true, features = ["codec"] }
 tracing = { workspace = true }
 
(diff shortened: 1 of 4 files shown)


monad-ipc/Cargo.toml
```diff
@@ -21,17 +21,22 @@ tokio = { workspace = true, features = [
...
 monad-secp = { path = "../monad-secp" }
+tokio = { workspace = true, features = ["macros", "net", "rt-multi-thread", "time"] }
 
 clap = { workspace = true, features = ["derive"] }
+hex = { workspace = true }
+itertools = { workspace = true }
 reth-primitives = { workspace = true }
 tracing-subscriber = { workspace = true }
+reqwest = { version = "0.11", features = ["json"] }
+serde_json = { workspace = true }
```
michael-yxchen: super nit: dependency order. Move version to workspace Cargo.toml

monad-ipc/examples/ethtx.rs
```diff
@@ -65,43 +67,137 @@ impl Sink<TransactionSigned> for MempoolTxIpcSender {
...
+}
 
-    #[arg(long, default_value_t = u32::MAX)]
-    tps: u32,
+async fn send_requests<'a, C: Iterator<Item = &'a TransactionSigned>>(
+    transaction_chunk: C,
+    rpc_addr: &String,
+    rpc_port: u16,
+) -> String {
+    let client = reqwest::Client::new();
```
michael-yxchen, line 114: just curious, is there a benefit to reuse the client vs creating new one for every request?

rgarc: not that I know of. perhaps there's a higher chance of different requests reusing the same connection, but I didn't really write this with any of that in mind.

PR #652, opened 2024-03-06
Pace ethtx tool

michael-yxchen:


monad-ipc/examples/ethtx.rs
```diff
@@ -84,8 +88,14 @@ async fn main() -> Result<(), Box<dyn Error>> {
     let mut sender = MempoolTxIpcSender::new(args.ipc_path).await?;
 
     let txs: Vec<_> = (0..args.num_tx).map(|_| make_tx(args.input_len)).collect();
+    let interval = time::Duration::from_secs(1) / args.tps;
     for tx in txs {
+        let start_time = time::Instant::now();
         sender.send(tx).await?;
+        let elapsed = start_time.elapsed();
+        if let Some(sleep_duration) = interval.checked_sub(elapsed) {
+            std::thread::sleep(sleep_duration);
```
omegablitz: This should be a tokio-aware sleep - or it'll break on a single-threaded runtime.

PR #654, opened 2024-03-06
Cleaner debug messages

michael-yxchen:


monad-secp/src/secp.rs
```diff
@@ -25,7 +25,7 @@ impl std::error::Error for Error {}
 
 impl std::fmt::Debug for PubKey {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        let ser = self.bytes();
+        let ser = self.bytes_compressed();
```
michael-yxchen, line 28: Changed to match the format in config file `node.toml` to match NodeIds in logs to the nodes. Can change everything to the uncompressed serialization if performance suffers

PR #671, opened 2024-03-08
Add state sync threshold

bamrith:


monad-consensus-state/src/lib.rs
```diff
@@ -943,6 +948,19 @@ where
         VT: ValidatorSetType<NodeIdPubKey = SCT::NodeIdPubKey>,
     {
         if let Some(qc) = self.pending_block_tree.get_missing_ancestor(qc) {
+            let max_seq_num_to_request =
+                self.pending_block_tree.get_root_seq_num() + self.config.state_sync_threshold;
+
+            if qc.get_seq_num() >= max_seq_num_to_request {
```
michael-yxchen, line 1001: Should this be something like `qc.seq_num() - block_tree.root.seq_num() >= max_seq_num_to_request`?

bamrith: The condition should be if `qc.get_seq_num()` is `self.config.state_sync_threshold` blocks over `block_tree.root.get_seq_num()`, then statesync. Which is `qc.get_seq_num() - block_tree.root.get_seq_num() >= self.config.state_sync_threshold`. Is that what you meant ?

michael-yxchen: oh yeah. I confused `max_seq_num_to_request` with the `config.state_sync_threshold`. Make sense now

PR #676, opened 2024-03-15
Create NodeState to group parameters to consensus state

ariqchowdhury:
This change groups common state parameters into consensus state handlers. This also simplifies consensus state unit tests

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -1,4 +1,4 @@
-use std::time::Duration;
+use std::{marker::PhantomData, time::Duration};
 
 use monad_blocktree::blocktree::BlockTree;
 use monad_consensus::{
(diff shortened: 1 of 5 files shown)


monad-consensus-state/src/lib.rs
```diff
@@ -176,6 +176,22 @@ pub enum StateRootAction {
     Defer,
 }
 
+pub struct NodeState<'a, ST, SCT, VTF, LT, TT>
```
michael-yxchen, line 179: nit: monad-node also defines a [NodeState](https://github.com/monad-crypto/monad-bft/blob/f9c4204e53396c1c0afba7a883a9eef032241cee/monad-node/src/state.rs#L18). Though we won't be using the consensus NodeState outside of monad-state, renaming this one can help avoid ambiguity. One idea I have is `ConsensusContext`. 

I don't feel strong about it. Feel free to leave as is or pick other names.

monad-consensus-state/src/lib.rs
```diff
@@ -1353,20 +1553,57 @@ mod test {
...
+        let (env, mut ctx) =
+            setup::<SignatureType, SignatureCollectionType, _, StateRootValidatorType, _, _>(
+                num_state as u32,
+                ValidatorSetFactory::default(),
+                SimpleRoundRobin::default(),
+                MockTxPool::default(),
+                || NopStateRoot,
+            );
+
+        let (consensus_state, mut node_state) = ctx[0].get_state();
```
michael-yxchen, line 1605: I see that some tests splits `NodeContext` to `ConsensusState` and others call the utility function from `NodeContext`. Just for my understanding, how does a test choose one way or the other?

monad-consensus-state/src/lib.rs
ariqchowdhury: I'll leave it for now, if it ends up being annoying can rename later

monad-consensus-state/src/lib.rs
ariqchowdhury: probably related to borrow checking

PR #682, opened 2024-03-20
Adding publish blockwatch step

mhumeSF:
~HOLD for now... monad-rpc docker file not building because of library issues~ Ready to merge


michael-yxchen:
Pieces from [flexnet builder](https://github.com/monad-crypto/monad-bft/blob/112d4bae25853d2abfda3dd68bf3f67e522f0dc9/docker/flexnet/images/monad-bft-builder/Dockerfile) might help with the lib issues

PR #683, opened 2024-03-21
Remove old blocks from blocktree

bamrith:


monad-consensus-state/src/lib.rs
```diff
@@ -966,17 +966,23 @@ where
...
                 );
 
                 metrics.consensus_events.trigger_state_sync += 1;
-            }
 
-            self.block_sync_requester
-                .request::<VT>(&qc, validators, metrics)
+                // Remove older blocks from the blocktree
+                self.pending_block_tree
+                    .remove_old_blocks(max_seq_num_to_request);
```
michael-yxchen: I think we want to keep only blocks within the state sync threshold. So something like `self.pending_block_tree.remove_old_blocks(qc.get_seq_num() - self.config.state_sync_threshold)`

bamrith: Got confused about the direction of sequence numbers. Fixed

monad-consensus-state/src/lib.rs
```diff
@@ -966,17 +966,23 @@ where
                 self.pending_block_tree.get_root_seq_num() + self.config.state_sync_threshold;
 
             if qc.get_seq_num() >= max_seq_num_to_request {
-                // TODO: This should trigger statesync. Remove blocksync request.
```
michael-yxchen, line 969: Did the comment mean removing pending block sync requests in block_sync module, or is it referring to something else? I think the former still might be helpful so we ignore the response when it comes back

bamrith: It was there because the node would request blocksync even if this condition is met (I assumed that the node would keep using blocksync regardless of how far it is until state sync was implemented). Now, it would request blocksync only if the condition is not met (with the added else clause).

PR #684, opened 2024-03-21
eth call Rust<->CXX 

rgarc:


monad-cxx/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+        &rlp_encoded_sender,
+        0,
+        &triedb_path);
+    }
+
+    let evmc_result = EvmcResult {
+        evmc_raw_result: *result.deref(),
+    };
+
+    let res = unsafe {
```
rgarc: Is there a way to avoid unsafe here?

monad-cxx/include/evmc.h
rgarc, line 1: This C header is needed for automatically generating bindings for the `evmc_result` type (and all downstream types) even in the mock implementation of eth_call. I had to manually copy this header into our repo to avoid requiring cloning submodules in CI.

monad-cxx/src/dummy.cpp
rgarc, line 1: This file was necessary to avoid a missing definition linking error.

monad-cxx/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+
+    impl Default for evmc_bytes32 {
+        fn default() -> Self {
+            evmc_bytes32 { bytes: [0u8; 32] }
+        }
+    }
+}
+
+autocxx::include_cpp! {
+    // #include "monad/rpc/eth_call.hpp"
```
rgarc: What's a good way of actually switching which file to include here? I couldn't get anything to work with `#[cfg(...)]`.

monad-rpc/src/call.rs
```diff
@@ -0,0 +1,194 @@
...
+use serde::{Deserialize, Serialize};
+use serde_json::{json, Value};
+
+use crate::{
+    blockdb::BlockDbEnv,
+    eth_json_types::{deserialize_block_tags, BlockTags, EthAddress, FixedData},
+    jsonrpc::JsonRpcError,
+};
+
+fn default_eth_address() -> EthAddress {
```
rgarc: remove

monad-cxx/Cargo.lock
omegablitz, line 1: This file shouldn't exist, should it?

monad-cxx/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+monad-eth-tx = { path = "../monad-eth-tx" }
+
+
+[dev-dependencies]
+tempfile = { workspace = true }
+
+[build-dependencies]
+autocxx-build = { workspace = true }
+bindgen = { workspace = true }
+cc = "1.0"
```
omegablitz: Specify versions in workspace Cargo.toml

monad-cxx/include/evmc.h
omegablitz, line 1: Is it possible for something to break silently if execution bumps their evmc version, but we don't bump this header version?

Ideally we could dodge needing this header at all for the dummy case. Right now we're only extracing status_code and output_data from the evmc result - is that all we think we'll need in the near future?
If so, we should just create our own wrapped eth_call result type (that doesn't depend on evmc).

monad-cxx/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+
+[lib]
+bench = false
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+autocxx = { workspace = true }
+alloy-primitives = { workspace = true }
+cxx = { workspace = true }
```
omegablitz, line 14: Is autocxx replacing cxx? If so, remove cxx from deps (here and in workspace Cargo.toml)

monad-cxx/src/dummy.cpp
omegablitz, line 1: What's this used for?

monad-cxx/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+    #include "mock_eth_call.hpp"
+    safety!(unsafe)
+    extern_cpp_type!("evmc_result", crate::bindings::evmc_result)
+    generate!("monad::rpc::eth_call")
+}
+
+/// A RAII-wrapper around evmc_result, analogous to the evmc::Result C++ type.
+#[derive(Debug)]
+pub struct EvmcResult {
+    evmc_raw_result: bindings::evmc_result,
```
omegablitz, line 48: Can you move the err, output data, and status code unsafe stuff to the eth_call function? And change this to just be a simple Rust type.

Unless there was an intentional reason for keeping the bindings::evmc_result type around

monad-cxx/src/lib.rs
omegablitz: Did you try `cfg` guarding the entire `autocxx::include_cpp` block?

monad-cxx/Cargo.lock
rgarc: no!

monad-cxx/src/dummy.cpp
rgarc: without this `dummy.cpp` file, the program fails at load time with this error
```
  = note: /usr/bin/ld: /home/rgarc/monad-bft/target/debug/build/monad-cxx-b6b200f967c7b0db/out/build/libmonad_shared.so: undefined reference to `monad::tracer'
          collect2: error: ld returned 1 exit status
```

monad-cxx/src/lib.rs
rgarc: tried this
```rust
#[cfg(feature = "production")]
autocxx::include_cpp! {
    #include "monad/rpc/eth_call.hpp"
    safety!(unsafe)
    extern_cpp_type!("evmc_result", crate::bindings::evmc_result)
    generate!("monad::rpc::eth_call")
}

#[cfg(not(feature = "production"))]
autocxx::include_cpp! {
    #include "mock_eth_call.hpp"
    safety!(unsafe)
    extern_cpp_type!("evmc_result", crate::bindings::evmc_result)
    generate!("monad::rpc::eth_call")
}
```
and it fails to compile with this error
```
  thread 'main' panicked at monad-cxx/build.rs:66:14:
  autocxx failed: ParseError(ConflictingModNames)
```
similarly trying something like this
```rust
autocxx::include_cpp! {
    #[cfg(feature = "production")]
    #include "monad/rpc/eth_call.hpp"
    #[cfg(not(feature = "production"))]
    #include "mock_eth_call.hpp"
    safety!(unsafe)
    extern_cpp_type!("evmc_result", crate::bindings::evmc_result)
    generate!("monad::rpc::eth_call")
}
```
does not compile either

michael-yxchen: I'm guessing that `autocxx::include_cpp` generates more than a block and the `cfg` thing is only attached to the first block. You can expand the macro to make sure.

Maybe try putting it in a `mod`?

rgarc: I tried both of the following
```rust
#[cfg(feature = "production")]
mod ffi {
    autocxx::include_cpp! {
        #include "monad/rpc/eth_call.hpp"
        safety!(unsafe)
        extern_cpp_type!("evmc_result", crate::bindings::evmc_result)
        generate!("monad::rpc::eth_call")
    }
}

#[cfg(not(feature = "production"))]
mod ffi {
    autocxx::include_cpp! {
        #include "mock_eth_call.hpp"
        safety!(unsafe)
        extern_cpp_type!("evmc_result", crate::bindings::evmc_result)
        generate!("monad::rpc::eth_call")
    }
}
```

and renaming the second one to `ffi2`
```rust
#[cfg(feature = "production")]
mod ffi {
    autocxx::include_cpp! {
        #include "monad/rpc/eth_call.hpp"
        safety!(unsafe)
        extern_cpp_type!("evmc_result", crate::bindings::evmc_result)
        generate!("monad::rpc::eth_call")
    }
}

#[cfg(not(feature = "production"))]
mod ffi {
    autocxx::include_cpp! {
        #include "mock_eth_call.hpp"
        safety!(unsafe)
        extern_cpp_type!("evmc_result", crate::bindings::evmc_result)
        generate!("monad::rpc::eth_call")
    }
}
```
neither of which work

rgarc: solution ended up being just having one code block
```rust
autocxx::include_cpp! {
    #include "eth_call.hpp"
    safety!(unsafe)
    extern_cpp_type!("evmc_result", crate::bindings::evmc_result)
    generate!("monad::rpc::eth_call")
}
```
as this doesn't affect whether the mock or prod version is called. that'll be controlled by just linking against the mock or prod library where necessary

PR #693, opened 2024-03-26
Docker compose to start local test instance

eerkaijun:
Was doing some metamask testing locally and realized the devnet dockerfile is not working so went ahead to write a docker compose such that we can start an instance of node + RPC in a single command. 

@ariqchowdhury @omegablitz @michael-yxchen Let me know if that's alright / any feedback!

### DIFF
--- a/docker/devnet/README.md
+++ b/docker/devnet/README.md
@@ -1,5 +1,18 @@
 ## Devnet
 
-To run a local devnet node, run the following commands in the root directory of `monad-bft`
-1. To build the docker image, run `docker build -t devnet -f docker/devnet/Dockerfile .`
-2. To run the docker container, run `docker run -d devnet`
+A local devnet consists of running a Monad node and a RPC server (which is the interface to send requests to the node). 
+
+**Option 1: Docker Compose (Recommended)**
+
+Run the docker compose script to start a local instance:
+1. Navigate to the current directory (`docker/devnet`)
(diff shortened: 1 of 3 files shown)


michael-yxchen:
We have a docker compose setup just like this under `docker/flexnet`. Notion doc [here](https://www.notion.so/monad-labs/FlexNet-Design-0385966cfe3944a89ef46a6ac40c6209?pvs=4). It comes with more tooling to generate node configs, configure the network, send RPC requests, etc. I'm working on jenkins CI to make sure the tests there always pass so we don't end up with broken docker images/nodes. It doesn't use the images under `docker/devnet`, which is the annoying part. We get two copies of basically the same images for RPC server and monad-node. 

I think one way of unifying things is to create one flexnet test referencing the devnet images, and add it to CI. I'm also ok with this docker compose and can help add it to CI

eerkaijun:
Ok so if I'm understanding correctly flexnet configures multiple nodes and basically defines the tests for this multi-node setup? 

I think either way that you suggested works so whichever that works better since you probably have more insights on this than me.

michael-yxchen:
That's right. 

Let's do it in flexnet to keep the integrations test in one place. Let me know if you run into any issues

eerkaijun:
Closing in favor of https://github.com/monad-crypto/monad-bft/pull/695

PR #697, opened 2024-03-29
Add nonce checking when creating proposal

bamrith:


monad-eth-txpool/src/lib.rs
```diff
@@ -62,9 +102,30 @@ impl TxPool for EthTxPool {
...
     }
+
+    fn handle_committed_txns(&mut self, committed_txns: FullTransactionList) {
+        let eth_tx_list = EthFullTransactionList::rlp_decode(committed_txns.bytes().clone())
+            .expect(
+                "transactions in blocks must have been verified and rlp decoded \
+            before being put in the pending blocktree",
+            );
+
+        // update account nonces
```
omegablitz, line 118: I think this can be set from the [transaction itself](https://paradigmxyz.github.io/reth/docs/src/reth_primitives/transaction/mod.rs.html#223-233)? Roughly:
```rust
let old_nonce = self.account_nonces.entry(sender).or_default();
*old_nonce = eth_tx.nonce().max(old_nonce);
```

monad-eth-txpool/src/lib.rs
```diff
@@ -26,24 +35,55 @@ impl TxPool for EthTxPool {
...
+            }
         }
 
         let pending_blocktree_txs: HashSet<EthTxHash> = HashSet::from_iter(pending_tx_hashes);
 
         let mut txs = Vec::new();
         let mut total_gas = 0;
 
-        for tx in self.0.values() {
+        let mut txs_to_propose: Vec<_> = self.tx_map.values().collect();
```
omegablitz, line 63: I think we'll need to store these in sorted order at some point, but we can wait until we have more concrete benchmarks before we implement that.

ariqchowdhury: I think we can merge in now and @rgarc can work on optimizations after

PR #700, opened 2024-04-03
Weighted random leader selection

MForensic:
This update introduces a predictable random leader selection mechanism, enhancing the system's ability to accurately identify the source of continuous leader failures, thereby improving leader accountability.


omegablitz:
#1169

PR #706, opened 2024-04-08
Integrate Raptor 🦖 broadcast into monad-node

omegablitz:


omegablitz:
This is what's currently running on devnet

monad-gossip/src/seeder/mod.rs
```diff
@@ -216,11 +218,29 @@ impl<'k, C: Chunker<'k>> Gossip for Seeder<'k, C> {
         self.update_tick(time);
         match to {
             RouterTarget::Broadcast => {
+                self.events
+                    .push_back(GossipEvent::Emit(self.me, message.clone()));
+
+                // TODO should we set this bound more empirically? This is arbitrary right now
+                if message.len() <= 2 * MAX_DATAGRAM_SIZE {
+                    let messages = self
```
ariqchowdhury, line 270: not sure I understand the purpose of this check. 
Why is 2x max datagram size fine? And is it intended that the message is processed by self regardless of whether its broadcast out or not?

monad-node/src/main.rs
```diff
@@ -245,7 +245,7 @@ async fn run(
         epoch_start_delay: Round(50),
         beneficiary: node_state.node_config.beneficiary,
         consensus_config: ConsensusConfig {
-            proposal_txn_limit: 5000,
+            proposal_txn_limit: 1000,
```
ariqchowdhury, line 248: why is this needed?

monad-gossip/src/seeder/mod.rs
omegablitz: If we're broadcasting a small message, naively broadcasting it is lower latency and uses less bandwidth. This is because there's overhead (at least 1 additional chunk) in sending "meta" messages.

monad-node/src/main.rs
omegablitz: It makes it easier to reason about performance when debugging devnet, because there's no possibility of cascading failures (eg small proposal fails, then next proposal is slightly larger, etc).

Once the current performance issues are fixed, we'll bump both this and ethtx up.

monad-gossip/src/seeder/mod.rs
omegablitz: And yes, broadcast always should send to self

PR #707, opened 2024-04-09
eth call Rust<->CXX

rgarc:
Does the same thing as #684 but gets rid of the dependency on evmc by creating our own type in C++
```C++
struct monad_evmc_result
{
    int status_code;
    std::vector<uint8_t> output_data;

    int get_status_code() const;
    std::vector<uint8_t> get_output_data() const;
};
```
and using that as the return type

### DIFF
--- a/monad-rpc/Cargo.toml
+++ b/monad-rpc/Cargo.toml
@@ -12,6 +12,7 @@ bench = false
 
 [dependencies]
 monad-blockdb = { path = "../monad-blockdb" }
+monad-cxx = { path = "../monad-cxx" }
 monad-triedb = { path = "../monad-triedb" }
 
 actix = { workspace = true }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3697,6 +3697,7 @@ dependencies = [
  "heed",
  "log",
  "monad-blockdb",
+ "monad-cxx",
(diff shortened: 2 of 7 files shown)


monad-cxx/src/eth_call.cpp
```diff
@@ -0,0 +1,28 @@
...
+}
+
+monad_evmc_result eth_call(
+    std::vector<uint8_t> const &rlp_encoded_transaction,
+    std::vector<uint8_t> const &rlp_encoded_block_header,
+    std::vector<uint8_t> const &rlp_encoded_sender,
+    block_num_t const block_number, std::string const &triedb_path,
+    std::string const &block_db_path)
+{
+    return monad_evmc_result{};
```
rgarc: this will call into execution but leaving empty for now

monad-rpc/Cargo.toml
```diff
@@ -13,12 +13,13 @@ bench = false
 [dependencies]
 monad-blockdb = { path = "../monad-blockdb" }
 monad-triedb = { path = "../monad-triedb" }
+monad-cxx = { path = "../monad-cxx" }
```
omegablitz, line 15: nit: sort

PR #708, opened 2024-04-09
Consolidate SeqNum to Epoch conversion

michael-yxchen:
Both the state root hash updater and epoch_manager check the epoch boundary conditions. The checkpoint code will do the same check. This PR consolidates SeqNum to Epoch conversion and lifts it to monad-types. It's easier to maintain and reason about.

### DIFF
--- a/monad-types/src/lib.rs
+++ b/monad-types/src/lib.rs
@@ -134,6 +134,32 @@ impl std::fmt::Debug for SeqNum {
     }
 }
 
+impl SeqNum {
+    /// Compute the epoch that the sequence number belong to. It does NOT mean
+    /// that the block is proposed in the epoch
+    ///
+    /// [0, val_set_update_interval] belongs to Epoch 1
+    /// [val_set_update_interval + 1, 2 * val_set_update_interval ] belongs to Epoch 2
+    /// ...
+    /// The first epoch is one block longer than all other ones
+    pub fn to_epoch(&self, val_set_update_interval: SeqNum) -> Epoch {
+        Epoch((self.0.saturating_sub(1) / val_set_update_interval.0) + 1)
+    }
+
+    pub fn is_epoch_end(&self, val_set_update_interval: SeqNum) -> bool {
+        *self % val_set_update_interval == SeqNum(0)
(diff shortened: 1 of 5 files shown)


monad-types/src/lib.rs
```diff
@@ -134,6 +134,16 @@ impl std::fmt::Debug for SeqNum {
     }
 }
 
+impl SeqNum {
+    pub fn to_epoch(&self, val_set_update_interval: SeqNum) -> Epoch {
+        Epoch((*self / val_set_update_interval).0 + 1)
```
bamrith: This might confuse readers into thinking that a block with that `SeqNum` is proposed in this `Epoch`. I think a comment explaining this function might be useful so it is not misused later.

monad-types/src/lib.rs
```diff
@@ -134,6 +134,16 @@ impl std::fmt::Debug for SeqNum {
...
+    }
+
+    pub fn is_epoch_start(&self, val_set_update_interval: SeqNum) -> bool {
```
bamrith: nit: Should this be `is_epoch_end` ?

monad-updaters/src/state_root_hash.rs
```diff
@@ -144,9 +144,15 @@ where
...
                     });
 
-                    if block.get_seq_num() % self.val_set_update_interval == SeqNum(0) {
+                    if block
+                        .get_seq_num()
+                        .is_epoch_end(self.val_set_update_interval)
+                    {
+                        let next_epoch =
+                            block.get_seq_num().to_epoch(self.val_set_update_interval) + Epoch(1);
+                        assert_eq!(next_epoch, self.epoch + Epoch(1));
```
ariqchowdhury: should we also put this in a helper function under impl SeqNum? ie get_next_epoch(SeqNum)

michael-yxchen: Created a slightly different helper `get_next_block_epoch` to return the epoch number of the next sequence number, so it only returns the next epoch if the sequence number is the end of an epoch.

PR #712, opened 2024-04-11
Add execution ledger path to RPC dockerfile

eerkaijun:


docker/devnet/Dockerfile
```diff
@@ -12,12 +12,12 @@ RUN --mount=type=cache,target=/usr/local/cargo/registry \
     mv target/release/monad-node node
 
 # Runner
-FROM debian:buster-slim
+FROM debian:bookworm-slim
 WORKDIR /usr/src/monad-bft
 
 COPY --from=builder /usr/src/monad-bft/node /usr/local/bin/monad-node
 
-ENV RUST_LOG=info
```
michael-yxchen, line 20: I'd prefer leaving the info log level in for anyone running the container itself. In docker-compose, we can override it to avoid spamming the terminal.

PR #713, opened 2024-04-11
Start ledger from SeqNum 0

michael-yxchen:
Exteral tools expect block number to start with 0. Changing sequence number logic to match that


michael-yxchen:
@bamrith Please take a look at the epoch switching mock-swarm test and lmk if the changes make sense

PR #715, opened 2024-04-12
Update RPC response format

eerkaijun:


monad-rpc/src/blockdb_handlers.rs
```diff
@@ -45,33 +47,94 @@ pub async fn monad_eth_chainId(blockdb_env: &BlockDbEnv) -> Result<Value, JsonRp
...
+    };
+
+    // parse block uncles, this is for legacy transactions, we can just set this as None?
+    let uncles = value
+        .block
+        .clone()
+        .ommers
+        .into_iter()
+        .map(|h| h.hash_slow())
+        .collect();
```
eerkaijun: Can we safely assume that there will no uncle blocks (I think that's for Ethereum pre-merge)? If so we can just set this as None.

michael-yxchen: yeah we won't have uncle blocks

PR #724, opened 2024-04-22
Add flexnet2-1

michael-yxchen:
Flexnet2-1 follows the same setup as flexnet2, except the bandwidth limits are lowered to 50Mbps. This is because raptor shard generation can move too fast with higher bandwidth configurations and cause serious backlog. Commit 0d7167d2 (pre raptor) provides the baseline for expected ledger length and txns count.

Other improvements
- Move verify ledger and count-tx script to common
- Correct path for cargo build cache with rootless docker
- Update README.md with current workflows


ariqchowdhury:
is this still needed @omegablitz ? Devnet was having the same issue with bandwidth but it has been resolved, so the bandwidth limit shouldn't be necessary, correct?

omegablitz:
Yes this shouldn't be necessary anymore; I still need to clean up my changes and create a PT for them, so let's defer on this for now.

omegablitz:
Changes are in #739

PR #728, opened 2024-04-25
Add check to prevent duplicated transaction hash to overwrite previous hash

eerkaijun:
.

### DIFF
--- a/monad-blockdb/src/lib.rs
+++ b/monad-blockdb/src/lib.rs
@@ -105,13 +105,21 @@ impl BlockDb {
 
         for (i, eth_tx) in block.body.iter().enumerate() {
             let key = EthTxKey(eth_tx.recalculate_hash());
-            let value = EthTxValue {
-                block_hash: block_table_key.clone(),
-                transaction_index: i as u64,
-            };
-            self.txn_hash_dbi
-                .put(&mut txn_hash_table_txn, &key, &value)
-                .expect("txn_hash_dbi put failed");
+            let existing_entry = self
+                .txn_hash_dbi
+                .get(&txn_hash_table_txn, &key)
+                .expect("failed to read from txn db");
+
+            // check that transaction hash does not exist before inserting
+            if existing_entry.is_none() {
+                let value = EthTxValue {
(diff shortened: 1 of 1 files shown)


omegablitz:
What's the motivation of this? If two transactions with the same hash are committed, the first one isn't necessarily the canonical one is it?

monad-blockdb/src/lib.rs
```diff
@@ -105,13 +105,21 @@ impl BlockDb {
...
+            // check that transaction hash does not exist before inserting
+            if existing_entry.is_none() {
+                let value = EthTxValue {
+                    block_hash: block_table_key.clone(),
+                    transaction_index: i as u64,
+                };
+                self.txn_hash_dbi
+                    .put(&mut txn_hash_table_txn, &key, &value)
+                    .expect("txn_hash_dbi put failed");
+            }
```
eerkaijun, line 122: To be honest, I'm not too sure if this is the desired behavior? Right now if someone send two (or more) exact same `rawTransaction`, the second transaction hash will overwrite the first hash in blockdb. I think this is not the expected behavior? As an attacker can just repeat a raw transaction and it will overwrite the information of the first transaction in the blockdb (which is usually the successful one). 

So I added a check to make sure there's no previous same transaction hash, but I think there's some edge cases though. For example, the first transaction could be the reverting transaction (e.g. nonce too high), and the second transaction is the valid transaction, which in this case it will be unable to be updated. 

Are there any account nonce checking before a transaction is added to a block?

ariqchowdhury: monad-blockdb should not be responsible for nonce checking. Nonce checking should be done at a higher level -- its just not currently implemented. 
Blockdb is just inserting committed blocks into the db, committed blocks are already considered valid from the protocol level.

eerkaijun: In that case, how should we deal with duplicated transaction hash in blockdb? 

In my opinion, the first transaction hash should take priority over subsequent transaction hashes in the `txnhash` table. Other table like `block` table wouldn't be affected, they can store duplicated transaction hashes.

PR #731, opened 2024-04-27
Return full transaction receipt object

eerkaijun:
Metamask needs some of these information and to make to make it fully compatible to ethereum json specs

### DIFF
--- a/monad-rpc/src/eth_txn_handlers.rs
+++ b/monad-rpc/src/eth_txn_handlers.rs
@@ -6,6 +6,7 @@ use reth_primitives::{BlockHash, TransactionSigned};
 use reth_rpc_types::{AccessListItem, Parity, Signature, Transaction};
 use serde::Deserialize;
 use serde_json::Value;
+use std::cmp::min;
 
 use crate::{
     blockdb::BlockDbEnv,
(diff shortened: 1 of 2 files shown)


monad-rpc/src/eth_txn_handlers.rs
```diff
@@ -179,9 +178,37 @@ pub async fn monad_eth_getTransactionReceipt(
...
+                    let tx_receipt = TransactionReceipt {
                         transaction_type: tx_type,
+                        transaction_hash: Some(tx.hash()),
+                        transaction_index: U64::from(txn_index),
+                        block_hash: Some(block.block.hash_slow()),
+                        block_number: Some(U256::from(block_num)),
+                        from: transaction.signer(),
+                        to: tx.to(),
+                        // TODO: read from triedb to determine whether a contract is deployed
+                        contract_address: None,
```
eerkaijun, line 204: Will somehow need this information from triedb eventually.

monad-rpc/src/eth_txn_handlers.rs
```diff
@@ -179,9 +178,37 @@ pub async fn monad_eth_getTransactionReceipt(
...
+                        // TODO: gas_used = cumulative_gas_used[txn_index] - cumulative_gas_used[txn_index-1]
+                        gas_used: U128::from(21000),
```
eerkaijun, line 206: For this one, we can read receipt at index i and index i-1 and get the difference, but I feel like it will be easier to just store `gas_used` in triedb if possible? (Since it's already calculated during execution, and will prevent from redoing the calculation at rpc side again)

ariqchowdhury: yeah this was discussed before, can you just note it in that notion page and we'll address it later.

eerkaijun: okay sounds good

PR #743, opened 2024-05-02
Pace block production for single node

michael-yxchen:
Pace block production by adding self delay for unicast message (vote)

When we have more single node debug configs, we should group everything together and pick MockGossip/SeederGossip depending on debug mode or not

### DIFF
--- a/docker/devnet/Dockerfile
+++ b/docker/devnet/Dockerfile
@@ -12,6 +12,7 @@ ENV PATH="${CARGO_ROOT}/bin:$PATH"
 # Builder
 COPY . .
 RUN --mount=type=cache,target=${CARGO_ROOT}/registry    \
+    --mount=type=cache,target=${CARGO_ROOT}/git          \
     --mount=type=cache,target=/usr/src/monad-bft/target \
     cargo build --release --bin monad-node && \
     mv target/release/monad-node node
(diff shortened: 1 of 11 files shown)


michael-yxchen:
output

monad-node/src/config/mod.rs
```diff
@@ -18,6 +18,10 @@ use util::{deserialize_bls12_381_pubkey, deserialize_secp256k1_pubkey};
 pub struct NodeConfig {
     #[serde(deserialize_with = "deserialize_eth_address_from_str")]
     pub beneficiary: EthAddress,
+    /// TODO: move under single node debug mode
+    /// Single node debug: block time in ms, ignored if genesis validator set
+    /// size is larger than one
+    pub single_node_block_time_ms: Option<u64>,
```
omegablitz: Could we overload max_rtt_ms to provide this functionality?
If node count is 1, set all sent messages delay in MockGossip to max_rtt_ms / 2 (minus some epsilon if necessary).

Justification for overloading this on max_rtt_ms:
- for the single-node case, setting all sent outbound message delay to 0.5 rtt is equivalent to (send proposal -> rx ack) taking 1 rtt.
- reduces monad-node config surface area
- can have a more general "message_delay" param instead of a "single_node_delay"

michael-yxchen: yeah it'd make the config much simpler. That block_time config right now need to be smaller than 3 * delta so node doesn't timeout

Block time will be (2 * `message_delay` + epsilon) when `message_delay` is applied to all outbound messages, proposal and vote

PR #745, opened 2024-05-02
eth_call support

zander-xyz:


monad-cxx/build.rs
```diff
@@ -7,7 +7,7 @@ fn main() {
     println!("cargo:rerun-if-changed=src");
     println!("cargo:rerun-if-changed=include");
     println!("cargo:rerun-if-changed=monad-execution");
-    let target = env::var("ETH_CALL_TARGET").unwrap_or("mock_eth_call".to_owned());
+    let target = env::var("ETH_CALL_TARGET").unwrap_or("monad_shared".to_owned());
```
omegablitz: Should be mock by default

monad-cxx/.clang-format
omegablitz, line 1: We already have a .clang-format that's copied from execution in `monad-triedb/triedb-driver/.clang-format`. Maybe we move that one to the root of this repo and use that for both `monad-cxx` and `monad-triedb`?

That way we can still run a formatter linter in CI without needing the submodule initialized

monad-cxx/src/lib.rs
```diff
@@ -83,20 +90,74 @@ mod test {
...
+            hex::encode(result.unwrap()),
+            "0000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000100000000000000000000000001020304050102030405010203040501020304050000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
         )
-        .unwrap();
-        dbg!("{}", &eth_call);
+    }
+
+    #[test]
+    fn test_callcontract() {
+        let db = ffi::make_testdb();
```
omegablitz, line 135: nit: wrap the testdb ffi into its own struct with a Drop implementation, so that the unsafe stuff is hidden. There's an example of this in `monad-triedb/src/lib.rs`

omegablitz:
Deferring on this until read-only triedb issues are diagnosed
cc @Chen-Yifan

omegablitz:
> Deferring on this until read-only triedb issues are diagnosed cc @Chen-Yifan

JK, this isn't a regression so will merge this now

PR #747, opened 2024-05-03
store transaction hashes in pending block tree

rgarc:


rgarc:
closed moving to #758

PR #751, opened 2024-05-06
Docker compose file to start single devnet node

eerkaijun:


eerkaijun:
Integrated execution + triedb into docker compose, such that we can now start a local devnet node (consensus + execution + rpc) in a single command. See instructions here: https://github.com/monad-crypto/monad-bft/tree/kai/update-dockerfile?tab=readme-ov-file#using-docker

When this PR https://github.com/monad-crypto/monad/pull/706 is merged, we can update the execution submodule.

@michael-yxchen I integrated this into the flexnet docker compose file so we can eventually add this into CI too, but currently I think it will break some of the RPC blaster tests as the nonce of the blaster are not set I think.

michael-yxchen:
Thanks for integrating them into flexnet.

The transactions from rpc blaster have non-zero nonces as I grabbed them fairly recently from eth history. They are working fine with our consensus client as we don't check for account nonces, only sorting them. CI won't fail after this PR because the transaction will make it into the ledger. 

I'm not familiar with execution but I assume that it fails the transactions as they are not consistent with the account nonces. We can fix them later when we add more assertions on triedb states, etc. 

I'll go over the PR one more time and we can merge it

eerkaijun:
> Can you update "monad-scripts/jenkins/flexnet-ci.sh:19"

Okie added that.

> but I assume that it fails the transactions as they are not consistent with the account nonces.

Yeap, and we probably also need to add these accounts to the genesis file to set some initial gas balance for that, but as you said, can be left for later

PR #756, opened 2024-05-07
Use single TrieDB handle for RPC calls

bamrith:


eerkaijun:
aaaa is this what is causing rpc to crash when there are multiple triedb reads at the same time? might have encounter similar issues too

PR #758, opened 2024-05-08
blocktree hash policy

rgarc:
A hash policy in this context is a function
```
fn(Block) -> HashSet<TxHash>
```
that RLP decodes the transactions in the block and returns a set of their hashes. This is used to ensure we don't propose a transaction that is already pending in the block tree. 

### Problem
Currently this RLP decoding is being done on the critical path of proposal creation. Naively moving the RLP decoding to when we append to the block tree tightly couples the hash policy with the block tree which is undesirable.
### Solution
RLP decode the transactions when adding a block to the block tree. Parameterize the block tree over this hash policy to enable opting out of this behavior for tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3515,6 +3515,7 @@ dependencies = [
  "futures",
  "itertools 0.10.5",
  "monad-async-state-verify",
+ "monad-blocktree",
  "monad-bls",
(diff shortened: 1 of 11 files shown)


rgarc:
closed moving to #759

PR #759, opened 2024-05-09
blocktree hash policy

rgarc:
A hash policy in this context is a function
```
fn(Block) -> HashSet<TxHash>
```
that RLP decodes the transactions in the block and returns a set of their hashes. This is used to ensure we don't propose a transaction that is already pending in the block tree. 

### Problem
Currently this RLP decoding is being done on the critical path of proposal creation. Naively moving the RLP decoding to when we append to the block tree tightly couples the hash policy with the block tree which is undesirable.
### Solution
RLP decode the transactions when adding a block to the block tree. Parameterize the block tree over this hash policy to enable opting out of this behavior for tests.

### DIFF
--- a/monad-blocktree/src/blocktree.rs
+++ b/monad-blocktree/src/blocktree.rs
@@ -5,6 +5,7 @@ use monad_consensus_types::{
     payload::FullTransactionList,
     quorum_certificate::QuorumCertificate,
     signature_collection::SignatureCollection,
+    txpool::{HashPolicy, HashPolicyOutput},
 };
 use monad_tracing_counter::inc_count;
 use monad_types::{BlockId, Round, SeqNum};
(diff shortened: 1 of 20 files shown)


monad-blocktree/src/blocktree.rs
```diff
@@ -178,21 +210,21 @@ impl<SCT: SignatureCollection> BlockTree<SCT> {
     }
 
     /// Fetches transactions on the path in [`b`, root)
-    pub fn get_txs_on_path_to_root(&self, b: &BlockId) -> Option<Vec<FullTransactionList>> {
-        let mut txs = Vec::default();
+    pub fn get_tx_hashes_on_path_to_root(&self, b: &BlockId) -> Option<HashPolicyOutput> {
```
ariqchowdhury, line 236: 2 separate functions

PR #760, opened 2024-05-09
End-to-end integration test with simple transfer

michael-yxchen:
This replaces the `rpc_blaster` test as we've added execution to integration.


michael-yxchen:
genesis.json allocates balances to the first 10 addresses in `monad-orderbook-web/python/benchmarks/constants.py`

PR #762, opened 2024-05-13
Update execution submodule and fix corresponding RPC

eerkaijun:
Execution submodule recent update allows multiple access to triedb at the same time, hence fixing RPC crash issue. There's also some schema change to triedb account schema, which is updated on RPC side.

### DIFF
--- a/monad-cxx/src/eth_call.cpp
+++ b/monad-cxx/src/eth_call.cpp
@@ -48,7 +48,7 @@ namespace
         ro.set_block_number(block_number);
         ro.load_latest();
         BlockState block_state{ro};
-        Incarnation incarnation{0, 0};
+        Incarnation incarnation{block_number, Incarnation::LAST_TX};
         State state{block_state, incarnation};
 
         // nonce validation hack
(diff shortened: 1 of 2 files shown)


eerkaijun:
Using the latest run loop that is on devnet1 https://github.com/monad-crypto/monad/pull/706

monad-cxx/src/eth_call.cpp
```diff
@@ -47,7 +48,8 @@ namespace
         ro.set_block_number(block_number);
         ro.load_latest();
         BlockState block_state{ro};
-        State state{block_state};
+        Incarnation incarnation{0, 0};
```
bamrith: Should this be `{block_number, Incarnation::LastTX}` ?

eerkaijun: Incarnation is the {block_number, txn_number} that the account is created at , while the `block_number` here is value passed in by the RPC call, so probably not the same. Not sure if incarnation matters here in eth_call though, cc @zander-xyz might know better

eerkaijun: ok alex suggested to follow bharath's suggestion

PR #763, opened 2024-05-13
Use single TrieDB handle for RPC calls

bamrith:


monad-rpc/src/triedb.rs
```diff
@@ -13,6 +16,7 @@ use crate::eth_json_types::{BlockTags, EthAddress, EthStorageKey};
 #[derive(Clone)]
 pub struct TriedbEnv {
     triedb_path: PathBuf,
+    db_handle: Arc<Handle>,
```
ariqchowdhury: this will be unsafe in the future if things change. Should give each thread in the threadpool its own db_ptr

omegablitz:
nit for future reference: https://doc.rust-lang.org/std/option/enum.Option.html#method.get_or_insert_with

PR #767, opened 2024-05-15
Compute EthBlock.gas_limit

michael-yxchen:


monad-ledger/src/lib.rs
```diff
@@ -171,6 +176,15 @@ fn generate_header<SCT: SignatureCollection>(
 
     randao_reveal_hasher.update(monad_block.payload.randao_reveal.0.clone());
 
+    // FIXME: decoding involves signature recovery, which is expensive. We
+    // should only do it once in our hot path. Before voting on the block, we
+    // need to decode the transaction payload for nonce and reserve balance
+    // checking. When that's implemented, we should calculate and store the gas
+    // limit to avoid re-decode and computing it here
```
ariqchowdhury, line 183: can you open an issue for this? There's multiple places that now require decoding txns (getting txn hashes). We should figure out the correct place to do it once and capture all the necessary data.

michael-yxchen:
Closed in favor of #773

PR #773, opened 2024-05-17
Assign protocol param to gas_limit

michael-yxchen:
`gas_limit` is the maximum gas allowed in this block. Eth sets block's gas limit to the protocol parameter 30M, instead of calculating the total sum of transaction gas limits.

See [Block Size](https://ethereum.org/en/developers/docs/gas/#block-size) for current limits. [Blocks page](https://ethereum.org/en/developers/docs/blocks/) provides definitions for different fields

This is an alternative to #767 , more aligned with what Eth is doing


ariqchowdhury:
can you update the commit message to explain how its more like eth's behaviour. add a reference.

PR #774, opened 2024-05-17
feat(rpc): basic eth_estimateGas support

philipglazman:
Notes for reviewer:
- change `eth_call` from Result to enum since pattern matching different exec results will be useful later 
- simple conversion from `CallRequest` to a typed ethereum transaction based on available fields
- basic functionality for gas estimation, lots of todos to actually improve reliability and skip eth_calls for transfer transactions

Qs:
- ~~Is there an easy way to seed triedb with populated accounts so I can test behavior?~~

### DIFF
--- a/monad-cxx/src/eth_call.cpp
+++ b/monad-cxx/src/eth_call.cpp
@@ -89,6 +89,11 @@ std::string monad_evmc_result::get_message() const
     return message;
 }
 
+int64_t monad_evmc_result::get_gas_used() const
+{
+    return gas_used;
+}
+
 monad_evmc_result eth_call(
     std::vector<uint8_t> const &rlp_txn, std::vector<uint8_t> const &rlp_header,
     std::vector<uint8_t> const &rlp_sender, uint64_t const block_number,
@@ -150,11 +155,13 @@ monad_evmc_result eth_call(
         ret.message = result.error().message().c_str();
     }
     else {
+        int64_t const gas_used = txn.gas_limit - result.assume_value().gas_left;
         ret.status_code = result.assume_value().status_code;
         ret.output_data = {
             result.assume_value().output_data,
             result.assume_value().output_data +
                 result.assume_value().output_size};
+        ret.gas_used = gas_used;
     }
     return ret;
 }
--- a/monad-rpc/src/call.rs
+++ b/monad-rpc/src/call.rs
@@ -36,6 +36,83 @@ pub struct CallRequest {
     pub transaction_type: Option<U8>,
 }
 
+/// Optimistically create a typed Ethereum transaction from a CallRequest based on provided fields.
+/// TODO: add support for other transaction types.
+impl TryFrom<CallRequest> for reth_primitives::transaction::Transaction {
+    type Error = JsonRpcError;
+    fn try_from(call_request: CallRequest) -> Result<Self, JsonRpcError> {
+        match call_request {
+            CallRequest {
+                gas_price: Some(gas_price),
+                ..
+            } => {
+                // Legacy
+                Ok(reth_primitives::transaction::Transaction::Legacy(
+                    reth_primitives::TxLegacy {
+                        chain_id: Some(1),
+                        nonce: call_request
+                            .nonce
+                            .unwrap_or_default()
+                            .try_into()
+                            .map_err(|_| JsonRpcError::invalid_params())?,
+                        gas_price: gas_price
+                            .try_into()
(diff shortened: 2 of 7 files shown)


monad-cxx/src/eth_call.cpp
```diff
@@ -89,6 +89,11 @@ std::string monad_evmc_result::get_message() const
     return message;
 }
 
+int64_t monad_evmc_result::get_gas_used() const
+{
+    return gas_used;
+}
+
```
eerkaijun, line 96: might want to include this in `mock_eth_call.cpp` too

monad-rpc/src/call.rs
```diff
@@ -36,6 +36,71 @@ pub struct CallRequest {
...
+                        .try_into()
+                        .map_err(|_| JsonRpcError::invalid_params())?,
+                    to: reth_primitives::TransactionKind::Call(call_request.to.unwrap_or_default()),
+                    value: reth_primitives::TxValue::from(call_request.value.unwrap_or_default()),
+                    input: call_request.input.data.unwrap_or_default(),
+                }))
+            }
+            CallRequest { .. } => {
+                // EIP-1559
+                Ok(reth_primitives::transaction::Transaction::Eip1559(reth_primitives::TxEip1559 {
```
eerkaijun: Tried running a quick test and seems like `eth_call` fails at `MONAD_ASSERT(!txn_result.has_error());` when it comes to eip1559 transactions, probably rlp decoding issue.

In `monad-cxx/monad-execution/libs/execution/src/monad/core/rlp/transaction_rlp.cpp`, the `decode_transaction` function requires the first byte as the typed transaction eip2718.

philipglazman: Yep, it was an rlp encoding issue for enveloped txns. Fixed now. I also, tested using the orderbook itest.

monad-rpc/src/gas_handlers.rs
```diff
@@ -21,20 +23,52 @@ struct MonadEthEstimateGasParams {
...
-    // Hardcoded as 200,000 gas for now
-    serialize_result(format!("0x{:x}", 2000000))
+    // TODO: use the provided block from user
+    let Some(block_header) = blockdb_env
+        .get_block_by_tag(BlockTags::Default(BlockTagKey::Latest))
+        .await
+    else {
+        debug!("blockdb did not have latest block header");
+        return Err(JsonRpcError::internal_error());
+    };
```
eerkaijun: I think we might want to use the latest executed block rather than the latest consensus block. Execution is always slightly lagging behind consensus, so if we use the latest consensus block, the states for that block are not initialized yet - so things like user balance will be 0 and I think the gas estimation will fail due to insufficient balance. 

There's a `get_latest_block` api in `triedb.rs` to get the latest executed block.

eerkaijun: Did a quick test `curl localhost:8080 -X POST --data '{"id":1,"jsonrpc":"2.0","method":"eth_estimateGas","params":[{"from":"0x6e4b6128516329707953a2067fc29e8adfb26c5e","value":"0xa688906bd8b00000","gasPrice":"0x77359400","data":"0x","to":"0x1c068220845bfcd05632a11ebc6976813b6c25db"}]}'` and seems like it's indeed returning an insufficient balance error. I'm *guessing* this might be the reason?

philipglazman: Good catch on fetching latest block from triedb, just updated. However, the root issue is that we set gasLimit to max int which causes `validate_transaction` to fail for  both `eth_call` and `eth_estimateGas`
```
9223372036854775807 * 2000000000
=18446744073709551614000000000 # bigger than sender's balance
```

eerkaijun: gotcha i see, any workaround for that?

philipglazman: We can either (1) manually set the gas price really low (2) fetch the account's balance from triedb and set the gas limit to the maximum they can actually afford with their provided gas_price. 

I think 2 is better since it fits expected user behavior and is more accurate.

eerkaijun: sounds good, alternatively seems like many rpc providers just set the gas limit to a high value: https://docs.alchemy.com/reference/gas-limits-for-eth_call-and-eth_estimategas

philipglazman: Cool, will commit the change! fwiw i think those alchemy gas limits are for monthly billing since the gas limit in a block is 30million

PR #776, opened 2024-05-19
Add coherency check to blocktree entries

bamrith:


monad-blocktree/src/blocktree.rs
```diff
@@ -46,14 +46,114 @@ struct Root {
...
+    Unvalidated(Block<SCT>),
+    Validated(BP::ValidatedBlock),
+}
+
+impl<SCT: SignatureCollection, BP: BlockPolicy<SCT>> Clone for BlockTreeEntry<SCT, BP> {
+    fn clone(&self) -> Self {
+        use BlockTreeEntry::*;
+        match self {
+            Unvalidated(block) => BlockTreeEntry::Unvalidated(block.clone()),
+            Validated(block) => BlockTreeEntry::Validated(block.clone()),
```
ariqchowdhury: delete the use;
`BlockTreeEntry::Unvalidated()`...

monad-blocktree/src/blocktree.rs
```diff
@@ -46,14 +46,114 @@ struct Root {
...
+        }
+    }
+}
+
+impl<SCT: SignatureCollection, BP: BlockPolicy<SCT>> fmt::Debug for BlockTreeEntry<SCT, BP> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        use BlockTreeEntry::*;
```
ariqchowdhury: same

monad-blocktree/src/blocktree.rs
```diff
@@ -46,14 +46,114 @@ struct Root {
...
+        match self {
+            Unvalidated(block) => write!(f, "Unvalidated({:?})", block),
+            Validated(block) => write!(f, "Validated({:?})", block),
+        }
+    }
+}
+
+impl<SCT: SignatureCollection, BP: BlockPolicy<SCT>> PartialEq<Self> for BlockTreeEntry<SCT, BP> {
+    fn eq(&self, other: &Self) -> bool {
+        use BlockTreeEntry::*;
```
ariqchowdhury: same

monad-consensus-state/src/lib.rs
```diff
@@ -314,14 +329,16 @@ where
             node.version,
         ));
 
+        let block = p.block;
```
ariqchowdhury: remove this, use p.block until it goes through the block_validator

monad-blocktree/src/blocktree.rs
```diff
@@ -46,14 +46,114 @@ struct Root {
     block_id: BlockId,
 }
 
+pub enum BlockTreeEntry<SCT: SignatureCollection, BP: BlockPolicy<SCT>> {
+    Unvalidated(Block<SCT>),
+    Validated(BP::ValidatedBlock),
```
ariqchowdhury: it looks strange to have Block and ValidatedBlock mixed throughout the blocktree code.
I was thinking that the block_validator returns and ValidatedBlock and everything after that point operates on that validated block -- in this case the validity of the block captures whether it is valid in isolation. Then, when we try to add to the blocktree, there's another kind of validity which is checked that validates that the branch is valid. 
Maybe we can call this second property "coherency". So blocktree only deals with ValidatedBlocks and tracks whether a block is Coherent. 
A block is Coherent if it can point to a parent block that is Coherent and it passes some branch rules (nonce checks etc). The root of the blocktree is always Coherent.

monad-blocktree/src/blocktree.rs
```diff
@@ -46,14 +46,114 @@ struct Root {
...
+            (Unvalidated(left), Unvalidated(right)) => left == right,
+            (Validated(left), Validated(right)) => left == right,
+            _ => false,
+        }
+    }
+}
+
+impl<SCT: SignatureCollection, BP: BlockPolicy<SCT>> Eq for BlockTreeEntry<SCT, BP> {}
+
+impl<SCT: SignatureCollection, BP: BlockPolicy<SCT>> BlockTreeEntry<SCT, BP> {
```
ariqchowdhury: this looks strange too...Maybe its cleaner if BlockTreeEntry is a struct with a ValidatedBlock field and a coherency field?

monad-blocktree/src/blocktree.rs
```diff
@@ -126,8 +223,20 @@ impl<SCT: SignatureCollection, BP: BlockPolicy<SCT>> BlockTree<SCT, BP> {
         // traverse up the branch from new_root, removing blocks and pushing to
         // the commit list.
         loop {
-            let parent_id = block_to_commit.get_parent_id();
-            commit.push(block_to_commit);
+            assert!(
+                matches!(block_to_commit, BlockTreeEntry::Validated(_)),
+                "should be validated"
+            );
```
ariqchowdhury: this is being checked again when matching.

monad-blocktree/src/blocktree.rs
```diff
@@ -205,6 +314,18 @@ impl<SCT: SignatureCollection, BP: BlockPolicy<SCT>> BlockTree<SCT, BP> {
         false
     }
 
+    pub fn is_validated(&self, b: &BlockId) -> bool {
```
ariqchowdhury: is_validated (or is_coherent) and path_to_root are the same thing right? The code has a mix of checking/using validated and path to root, we should just check this flag now. Whenever a block is added to the blocktree we should check if there is a path to root and set its coherency state appropriately, and then moving forward only check that state. (can have a debug_assert to confirm that path-to-root exists)

monad-consensus-state/src/lib.rs
```diff
@@ -330,47 +347,66 @@ where
...
-            node.metrics.consensus_events.failed_txn_validation += 1;
-            return cmds;
-        };
+        let validated_block =
+            match self.try_validate_block(block) {
+                BlockValidationResult::Validated(validated_block) => {
+                    let author_pubkey = node
+                        .val_epoch_map
+                        .get_cert_pubkeys(&epoch)
+                        .expect("proposal message was verified")
```
ariqchowdhury: I think you should keep the original code with `let Some(block) = self.block_validator.validate(p.block) else...` and then add that block to the blocktree and let the blocktree figure out if the block is coherent yet. 
the try_add_blocktree function can return whether the block was coherent or incoherent and based on that, we can skip voting

monad-consensus-state/src/lib.rs
```diff
@@ -418,6 +456,70 @@ where
...
+
+        // Parent is not validated yet. Add to blocktree to validate later
+        BlockValidationResult::DelayValidation(block)
+    }
+
+    /// This function tries to validate all the blocks on its path from root and update the
+    /// blocktree
+    /// Returns true if there is a path to root and its parent is validated
+    /// Returns false if there is no path to root (missing ancestor)
+    fn try_validate_ancestors(&mut self, block: &Block<SCT>) -> bool {
```
ariqchowdhury: this function and its usage seem strange to me.
If a block is added to the blocktree and there is a path-to-root, then the entire path must be coherent already and we don't need to validate any ancestors. If there isn't a path-to-root, then we can't validate the path. 

When we are missing blocks and waiting for them (via blocksync or out-of-order arrival), when a block arrives that connects a path-to-root, at that time we should be able to go through that entire path and validate each block and discard any that are invalid. But that validation needs to start from the root and move down the chain. 

so when would this function ever be needed?  

edit: 
I think I understand, if blocks A, D are in the blocktree and B, C arrive later, A, B, C, D are a chain but are not validated until E arrives which then validates all the ancestors. 
I think that we should fix this so that we always have the property of coherent iff path-to-root
There should be the invariant that if a path-to-root exists, the blocks in that path are all coherent. 
This means we need to change the blocktree to immediately check the blocks A, B, C, D when B arrives instead of waiting for E.

monad-blocktree/src/blocktree.rs
ariqchowdhury, line 1: there should be new unit tests added to the blocktree to cover the new functionality we're adding here. From what I can see, only existing tests were updated

michael-yxchen, line 1: In your design, when does validation/coherence check happen?

bamrith: In this implementation, it is done in `try_validate_ancestors` in the consensus state. But, as Ariq mentioned in his example, that validation happens only when a new block extends the branch and not immediately after a path to root is created.

Edit: I'm updating it to do the coherence checking immediately after the path to root is created.

monad-blocktree/src/blocktree.rs
```diff
@@ -110,42 +148,50 @@ impl<SCT: SignatureCollection, BP: BlockPolicy<SCT>> BlockTree<SCT, BP> {
...
             .expect("new root must exist in blocktree");
-        let mut block_to_commit = new_root_block.clone();
+        let mut entry_to_commit = new_root_entry.clone();
 
         // traverse up the branch from new_root, removing blocks and pushing to
         // the commit list.
         loop {
-            let parent_id = block_to_commit.get_parent_id();
-            commit.push(block_to_commit);
+            assert!(entry_to_commit.is_coherent, "should be coherent");
```
ariqchowdhury: nit: "must be coherent"

monad-blocktree/src/blocktree.rs
```diff
@@ -162,11 +208,80 @@ impl<SCT: SignatureCollection, BP: BlockPolicy<SCT>> BlockTree<SCT, BP> {
...
 
+    pub fn update_coherency(&mut self, block_id: BlockId) {
+        let mut block_ids_to_update: VecDeque<BlockId> = vec![block_id].into();
+
+        while !block_ids_to_update.is_empty() {
+            // Next block to check coherency
+            let next_block = block_ids_to_update.pop_front().unwrap();
+
+            let blocktree_entry = self.tree.entry(next_block);
+            blocktree_entry.and_modify(|entry| entry.is_coherent = true);
```
ariqchowdhury: nit: `self.tree.entry(next_block).and_modify()...`

monad-blocktree/src/blocktree.rs
```diff
@@ -162,11 +208,80 @@ impl<SCT: SignatureCollection, BP: BlockPolicy<SCT>> BlockTree<SCT, BP> {
...
+            is_coherent,
+            children_blocks,
+        };
+
+        self.tree.insert(new_block_id, new_block_entry);
+
+        // Retrieve the parent block's coherency and children blocks
+        let (is_parent_coherent, children_blocks_to_update) = if parent_id == self.root.block_id {
+            (true, Some(&mut self.root.children_blocks))
+        } else if let Some(parent_entry) = self.tree.get_mut(&parent_id) {
```
ariqchowdhury, line 235: do we need to special case the root == parent case? If the parent happens to be the root, its still returning the same stuff (is_coherent and children blocks)

monad-consensus-state/src/lib.rs
```diff
@@ -343,7 +343,7 @@ where
             return cmds;
         }
 
-        let Some(block) = self.block_validator.validate(p.block) else {
+        let Some(validated_block) = self.block_validator.validate(p.block) else {
```
ariqchowdhury: nit: I'd rather keep the name block since the type of it shows its validated

monad-consensus-state/src/lib.rs
```diff
@@ -788,7 +752,87 @@ where
...
+            let state_root_action =
+                state_root_action.expect("should exist if proposal was handled this round");
+            cmds.extend(self.try_vote(block, state_root_action, node));
+        } else {
+            cmds.extend(self.try_propose(node));
+        }
+
+        cmds
+    }
+
```
ariqchowdhury, line 768: #[must_use]

monad-blocktree/src/blocktree.rs
bamrith: I added that case because root is its own struct `Root` and not a `BlockTreeEntry`. Also, since root is committed, it is always coherent, so it doesn't have a `is_coherent` flag.

ariqchowdhury: makes sense

PR #784, opened 2024-05-27
Add log rotation to RPC 

eerkaijun:
Adding log rotation as the size of log file becomes large very quickly


omegablitz:
Let's do log rotation externally (cc @mhumeSF)

PR #785, opened 2024-05-28
Make mempool ipc sender more robust

philipglazman:
- Adds basic retry to initial mempool sender connection
- Handles broken pipe inside `sink` by waiting for new socket file to be created. We can't undo a`poll_ready` inside a `send` so broken pipe is handled during flush.

### DIFF
--- a/monad-rpc/src/mempool_tx.rs
+++ b/monad-rpc/src/mempool_tx.rs
@@ -1,12 +1,34 @@
-use std::{path::Path, task::Poll};
+use std::{
+    ffi::OsStr,
+    io::ErrorKind,
+    path::{Path, PathBuf},
+    pin::Pin,
+    task::Poll,
+};
 
-use futures::{Sink, SinkExt};
+use futures::{executor::block_on, ready, Future, FutureExt, Sink, SinkExt};
+use log::debug;
+use notify::{Event, RecursiveMode, Watcher};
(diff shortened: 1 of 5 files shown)


monad-rpc/src/mempool_tx.rs
```diff
@@ -15,44 +26,172 @@ impl MempoolTxIpcSender {
...
 impl Sink<TransactionSigned> for MempoolTxIpcSender {
     type Error = std::io::Error;
 
     fn poll_ready(
-        mut self: std::pin::Pin<&mut Self>,
+        self: std::pin::Pin<&mut Self>,
         cx: &mut std::task::Context<'_>,
     ) -> Poll<Result<(), Self::Error>> {
-        self.writer.poll_ready_unpin(cx)
+        block_on(async { self.writer.lock().await.poll_ready_unpin(cx) })
```
ariqchowdhury: should be able to use this instead of block_on?
https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#method.blocking_lock

philipglazman: right, this works perfectly. thanks !

omegablitz:
Does this PR also address monad-crypto/monad-internal#284? If so, could you update the Dockerfile?

PR #787, opened 2024-05-29
keystore key creation and decryption

eerkaijun:
in reference to https://github.com/monad-crypto/monad-internal/issues/138

actual integration to be in a separate pr: https://github.com/monad-crypto/monad-bft/pull/788

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@
 /docker/devnet/monad/ledger
 /docker/devnet/monad/wal
 /docker/devnet/monad/triedb
+
+test-keys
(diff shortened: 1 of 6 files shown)


bamrith:
This keystore has a counterpart implemented under [monad-scripts/keystore/](https://github.com/monad-crypto/monad-bft/tree/master/monad-scripts/keystore). Those scripts were meant to be used to generate the BLS and SECP keys and this was meant to be used for decryption. The reason checksum and cipher messages were in their respective modules was to follow the [EIP-2335 schema](https://eips.ethereum.org/EIPS/eip-2335), and the python scripts follow this. If we are going to deviate, could you also update the scripts to make it consistent ?

eerkaijun:
Integrating the key generation into `monad-keystore` crate: https://github.com/monad-crypto/monad-bft/pull/791

PR #788, opened 2024-05-29
Keystore module integration into monad node

eerkaijun:
I think this will break the integration test though as they seem to be using base64 encoded private keys, might need to update those (or otherwise we can still have a fallback option such that if no keystore file is provided, it will fallback to using the base64 file)


eerkaijun:
Will merge this in after I fix the relevant integration tests

PR #791, opened 2024-05-31
keystore generation

eerkaijun:


monad-keystore/src/main.rs
```diff
@@ -0,0 +1,74 @@
...
+}
+
+fn main() {
+    let args = Args::parse();
+    let mode = args.mode;
+    let keystore_path = args.keystore_path;
+
+    match mode.as_str() {
+        "create" => {
+            println!("It is recommended to generate key in air-gapped machine to be secure.");
```
ariqchowdhury, line 40: this warning should be much louder. Not a recommendation -- no one should ever use this in production

PR #794, opened 2024-06-04
Lock n+2 validator set in epoch n

michael-yxchen:
Our previous implementation locks n+1 validator set in epoch n, giving validators a small window to compute and validate that they all agree on the same set. This PR changes it to n+2. Validators have at least an entire epoch to prepare for their validating duties, e.g. syncing etc.

Eth adopts the same [policy](https://eth2book.info/capella/part2/building_blocks/randomness/#lookahead)

### DIFF
--- a/monad-updaters/src/state_root_hash.rs
+++ b/monad-updaters/src/state_root_hash.rs
@@ -18,7 +18,7 @@ use monad_executor_glue::{MonadEvent, StateRootHashCommand};
 use monad_types::{Epoch, SeqNum, Stake};
 use rand::RngCore;
 use rand_chacha::{rand_core::SeedableRng, ChaChaRng};
-use tracing::debug;
+use tracing::{debug, error};
 
 pub trait MockableStateRootHash:
     Executor<Command = StateRootHashCommand<Block<Self::SignatureCollection>>>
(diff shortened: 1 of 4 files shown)


monad-updaters/src/state_root_hash.rs
```diff
@@ -50,9 +50,8 @@ pub struct MockStateRootHashNop<ST, SCT: SignatureCollection> {
     state_root_update: Option<StateRootHashInfo>,
 
     // validator set updates
-    epoch: Epoch,
     genesis_validator_data: ValidatorData<SCT>,
-    next_val_data: Option<ValidatorData<SCT>>,
+    next_val_data: Option<(Epoch, ValidatorData<SCT>)>,
```
ariqchowdhury: can this tuple be changed to a struct?

monad-updaters/src/state_root_hash.rs
```diff
@@ -145,12 +143,15 @@ where
...
+                            error!("Validator set data is not consumed");
+                        }
+                        let locked_epoch = block
                             .get_seq_num()
-                            .get_next_block_epoch(self.val_set_update_interval);
-                        assert_eq!(next_epoch, self.epoch + Epoch(1));
-                        self.epoch = next_epoch;
-                        self.next_val_data = Some(self.genesis_validator_data.clone());
+                            .get_locked_epoch(self.val_set_update_interval);
+
```
ariqchowdhury: should we assert here that locked_epoch is self.epoch+2?

PR #795, opened 2024-06-04
txpool improvements

rgarc:
This PR replaces with existing proposal creation algorithm with a new one that sorts transactions by effective tip while preserving the nonce order per account.

See this [linked comment](https://github.com/monad-crypto/monad-internal/issues/143#issuecomment-2164849103) to see how the algorithm creates a block for a small example txpool.

more context: https://github.com/monad-crypto/monad-internal/issues/143


monad-eth-txpool/src/lib.rs
```diff
@@ -10,46 +13,219 @@ use monad_consensus_types::{
...
+};
 use monad_types::{BlockId, Epoch, NodeId, Round, SeqNum};
+use reth_primitives::{Transaction, TxEip1559, TxEip2930, TxEip4844, TxLegacy};
+use sorted_vector_map::SortedVectorMap;
+
+type VirtualTimestamp = u64;
+
+/// Needed to have control over Ord implementation
+#[derive(Debug, Default, Eq, PartialEq)]
+struct WrappedTransaction((EthTransaction, EthAddress, VirtualTimestamp));
```
ariqchowdhury: prefer named fields rather than tuple, easier to read

PR #797, opened 2024-06-05
Raptorcast Implementation

omegablitz:
Raptorcast is a replacement for monad_quic::Service, built on top of monad_dataplane.

### DIFF
--- a/monad-raptorcast/src/lib.rs
+++ b/monad-raptorcast/src/lib.rs
@@ -1,0 +1,1097 @@
+use std::{
+    collections::{hash_map::Entry, BTreeMap, HashMap, VecDeque},
+    marker::PhantomData,
+    net::SocketAddr,
+    ops::{DerefMut, Range},
+    pin::Pin,
+    task::{Context, Poll, Waker},
+};
+
+use bytes::{Bytes, BytesMut};
+use futures::{Stream, StreamExt};
+use itertools::Itertools;
+use monad_crypto::{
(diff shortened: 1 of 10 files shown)


omegablitz:
I'll squash this before merging. Future TODOs:
- Parallelize signatures/hashing/coding on both rx and tx sides
  - Create smaller batches per sendmmsg to lower latency?
- Implement rebroadcast batching. Right now, if we receive a sequence of `[valid_message, invalid_message, valid_message, ...]`, we will call 1 sendmmsg per valid_message.
- Add Round as a parameter to RouterCommand to be used as a timestamp
  - Clear out message cache for earlier obsolete rounds
- Figure out pacing (cc @buytenh)
- Integrate into monad-node

monad-raptorcast/src/lib.rs
```diff
@@ -0,0 +1,1096 @@
...
+use monad_executor_glue::{Message, RouterCommand};
+use monad_merkle::{MerkleHash, MerkleProof, MerkleTree};
+use monad_types::{Deserializable, NodeId, Round, RouterTarget, Serializable, Stake};
+use raptor_code::SourceBlockDecoder;
+
+pub struct RaptorCastConfig<ST>
+where
+    ST: CertificateSignatureRecoverable,
+{
+    // TODO support dynamic updating
```
omegablitz, line 30: Add to TODO: we need to GC these

PR #801, opened 2024-06-06
Add `triedb_async_read()` and `triedb_poll()` to the triedb driver.

ned14:
Depends on https://github.com/monad-crypto/monad/pull/746

Implements https://github.com/monad-crypto/monad-internal/issues/232


ned14:
https://github.com/monad-crypto/monad/pull/746 has been merged, so rebased and updated this PR to match.

PR #802, opened 2024-06-06
figure out why test fails. it really shouldnt

rgarc:
PR only exists for tracking purposes
https://github.com/monad-crypto/monad-internal/issues/247


rgarc:
we should be using 
```rust
TransactionSigned::from_transaction_and_signature(transaction, signature)
```

PR #804, opened 2024-06-06
Parsing access list more gracefully

eerkaijun:
Prevents RPC from crashing if access list is None


eerkaijun:
Some integrations estimate `maxFeePerGas` in their transactions based on previous blocks, so we need to set a non-zero base fee. Hardcoded to a small value for now until we do the proper calculation.

PR #806, opened 2024-06-07
Use sender balance to improve gas estimation

philipglazman:


monad-rpc/src/gas_handlers.rs
```diff
@@ -55,6 +61,57 @@ pub async fn monad_eth_estimateGas(
...
+    let allowance = if params.tx.from.is_some() && params.tx.gas.is_none() {
+        let from = params.tx.from.expect("sender address");
+        let TriedbResult::Account(_, balance, _) = triedb_env
+            .get_account(from.into(), BlockTags::Number(Quantity(block_number)))
+            .await
+        else {
+            return Err(JsonRpcError::internal_error());
+        };
+
+        if params.tx.gas_price.is_some() {
```
philipglazman: Would be nice to use flatten enum for gas prices to support both legacy + eip1559. Will do that next, but wanted to get the PR to unblock quickly

monad-rpc/src/gas_handlers.rs
```diff
@@ -55,6 +61,57 @@ pub async fn monad_eth_estimateGas(
         return Err(JsonRpcError::internal_error());
     };
 
+    // If a `from` address is provided, subtract the effective gas price from the balance to get an accurate gas limit.
+    let allowance = if params.tx.from.is_some() && params.tx.gas.is_none() {
```
eerkaijun: I'm thinking perhaps the following logic can be put into the `try_from` function in https://github.com/monad-crypto/monad-bft/blob/master/monad-rpc/src/call.rs#L43 

`eth_call` will also face the same issue on gas estimation

philipglazman: We can't do this in `try_from` since we need access to sender's `balance` but a helper function makes sense for `eth_call`

monad-rpc/src/gas_handlers.rs
tongzhi1998: I was thinking along the same line. (In execution code, we just have max_fee_per_gas instead of 2 separate fields). Also just curious if we are going to support eip2930 type txn?

philipglazman: > Also just curious if we are going to support eip2930 type txn?

eip2930 usage is quite low, but we should for completeness

PR #809, opened 2024-06-11
Add block timestamp in block proposal

eerkaijun:
Will need milliseconds Unix timestamp since our blocks might be subseconds but wanted to check Blockscout integration first

Edit: currently setting the block header timestamp to Unix milliseconds for validation, but on RPC side parsing back to in Unix seconds so that external integrations work as expected.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3521,6 +3521,7 @@ version = "0.1.0"
 dependencies = [
  "alloy-primitives",
  "alloy-rlp",
+ "chrono",
  "clap 4.4.10",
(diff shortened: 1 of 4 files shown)


omegablitz:
Just noting that this doesn't add the block timestamp to the proposal - it only adds it to the ledger - so multiple nodes in the network won't have consistent ledgers.

eerkaijun:
Will add this commit to devnet but keeping it off the master branch for now. Let's revisit when we are setting up the multi nodes

PR #810, opened 2024-06-11
Add validator set change support to RaptorCast

omegablitz:
This involved modifying RouterCommand. This lays the groundwork for supporting blocksync/statesync from nodes that aren't in the validator set. The testing here is notably light - we need to backport the RouterCommand changes into the mock routers to be able to test this. Saving this for a future PR.


monad-gossip/src/broadcasttree.rs
ariqchowdhury, line 1: we can probably delete this now and just keep gossipsub around for baseline comparisons.

monad-state/src/consensus.rs
```diff
@@ -216,6 +217,9 @@ where
         let mut parent_cmds: Vec<Command<_, _, _, _, _>> = Vec::new();
 
         match cmd.0 {
+            ConsensusCommand::EnterRound(epoch, round) => parent_cmds.push(Command::RouterCommand(
+                RouterCommand::UpdateCurrentRound(epoch, round),
+            )),
             ConsensusCommand::Publish { target, message } => {
```
ariqchowdhury, line 223: Wondering if there is an off-by-one issue here. The EnterRound command is generated by pacemaker.
If consensus state is in round N-1 and is the leader for N, it collects the votes from N-1 and then creates a QC in handle_vote for N. It calls `process_certificate_qc` which then calls `pacemaker.advance_round_qc` which increments the pacemaker to round N. But it does this directly, not via command, so we're still in the consensus state machine and we call try_propose and create the router raptor cast command. When the router gets the command, its internal current_round is still N-1, right? Because it hasn't got the pacemaker newround command yet. 
not sure if I missed anything there.

omegablitz: `process_certificate_qc` gets called here: https://github.com/monad-crypto/monad-bft/blob/47a1a5c0df7d1b775866aa7bd850f1c813f2155a/monad-consensus-state/src/lib.rs#L425-L431
which calls `pacemaker.advance_round_qc` here: https://github.com/monad-crypto/monad-bft/blob/47a1a5c0df7d1b775866aa7bd850f1c813f2155a/monad-consensus-state/src/lib.rs#L720-L731
which calls `pacemaker.enter_round` here: https://github.com/monad-crypto/monad-bft/blob/47a1a5c0df7d1b775866aa7bd850f1c813f2155a/monad-consensus/src/pacemaker.rs#L303

So at this point, the top-level cmds contains EnterRound. After that, try_propose is called:
https://github.com/monad-crypto/monad-bft/blob/47a1a5c0df7d1b775866aa7bd850f1c813f2155a/monad-consensus-state/src/lib.rs#L433

Raptorcast will process the EnterRound command first, and then the Raptorcast command (or at least, that's the intention so let me know if I missed something).

----
Sidenote, the Broadcast/Raptorcast commands are tagged with the epoch/round that they belong to: https://github.com/monad-crypto/monad-bft/blob/47a1a5c0df7d1b775866aa7bd850f1c813f2155a/monad-consensus-state/src/lib.rs#L958
As long as those are set correctly, the right validator set will be picked for rebroadcasting. If there's an off-by-one error between the EnterRound and the Raptorcast command, there'll be some warnings in the logs telling us that.

ariqchowdhury: okay, that sounds right...keep in mind, there might be a mock-swarm test that shuffles the order of commands. I recall Justin working on that or something similar.

PR #811, opened 2024-06-11
Separate triedb utilities from monad-rpc into crate

philipglazman:
Move triedb utilities from `monad-rpc` into a separate crate. Blocks `monad-indexer` project. These utilities do not go into `monad-triedb` in order to avoid unnecessary dependencies.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,7 @@ members = [
     "monad-validator",
     "monad-virtual-bench",
     "monad-wal",
+    "monad-triedb-utils",
 ]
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -46,7 +46,7 @@ members = [
     "monad-updaters",
     "monad-validator",
     "monad-virtual-bench",
-    "monad-wal",
+    "monad-wal", "monad-triedb-utils",
```
omegablitz: format

PR #812, opened 2024-06-12
delay mempool clearing

rgarc:
https://github.com/monad-crypto/monad-internal/issues/142

This PR aims to decouple the clearing of the mempool from the proposal creation path by introducing a new event and command pair, enabling the mempool to be cleared via the loopback executor.

### DIFF
--- a/monad-consensus-state/src/command.rs
+++ b/monad-consensus-state/src/command.rs
@@ -57,6 +57,8 @@ where
     StateRootHash(Block<SCT>),
     // TODO-2 add command for updating validator_set/round
     // - to handle this command, we need to call message_state.set_round()
+    /// Issue to clear mempool
+    ClearMempool,
 }
 
 impl<ST, SCT> ConsensusCommand<ST, SCT>
--- a/monad-executor-glue/src/lib.rs
+++ b/monad-executor-glue/src/lib.rs
@@ -238,6 +238,8 @@ pub enum ValidatorEvent<SCT: SignatureCollection> {
 pub enum MempoolEvent {
     /// Txns that are incoming via RPC (users)
     UserTxns(Vec<Bytes>),
+    /// Remove transactions that were not included in proposal
+    Clear,
 }
 
 impl Debug for MempoolEvent {
@@ -250,6 +252,7 @@ impl Debug for MempoolEvent {
                     &txns.iter().map(Bytes::len).sum::<usize>(),
                 )
                 .finish(),
+            Self::Clear => f.debug_struct("Clear").finish(),
         }
     }
 }
@@ -335,6 +338,7 @@ where
             MonadEvent::MempoolEvent(MempoolEvent::UserTxns(txns)) => {
                 format!("MempoolEvent::UserTxns -- number of txns: {}", txns.len())
             }
+            MonadEvent::MempoolEvent(MempoolEvent::Clear) => "CLEARMEMPOOL".to_string(),
             MonadEvent::AsyncStateVerifyEvent(AsyncStateVerifyEvent::LocalStateRoot(root)) => {
                 format!(
                     "AsyncStateVerifyEvent::LocalStateRoot -- round:{} seqnum:{} hash:{}",
--- a/monad-proto/proto/message.proto
+++ b/monad-proto/proto/message.proto
@@ -57,6 +57,6 @@ message ProtoMonadMessage {
     ProtoUnverifiedConsensusMessage consensus = 1;
     ProtoRequestBlockSyncMessage block_sync_request = 2;
     ProtoBlockSyncMessage block_sync_response = 3;
-    ProtoPeerStateRootMessage peer_state_root = 5;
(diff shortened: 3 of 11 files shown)


monad-consensus-state/src/command.rs
```diff
@@ -56,12 +55,7 @@ where
     StateRootHash(Block<SCT>),
     // TODO-2 add command for updating validator_set/round
     // - to handle this command, we need to call message_state.set_round()
-    //
-    /// Sends transactions remaining in TxPool to another validator
-    CascadeTxns {
-        peer: NodeId<SCT::NodeIdPubKey>,
-        txns: Bytes,
-    },
+    ClearMempool,
```
michael-yxchen, line 61: nit: update comment

monad-executor-glue/src/lib.rs
```diff
@@ -357,6 +344,7 @@ where
             }
             MonadEvent::AsyncStateVerifyEvent(_) => "ASYNCSTATEVERIFY".to_string(),
             MonadEvent::MetricsEvent(_) => "METRICS".to_string(),
+            MonadEvent::MempoolEvent(MempoolEvent::Clear) => "CLEARMEMPOOL".to_string(),
```
michael-yxchen: nit: group it with the other mempool event

monad-proto/proto/message.proto
```diff
@@ -61,7 +57,6 @@ message ProtoMonadMessage {
     ProtoUnverifiedConsensusMessage consensus = 1;
     ProtoRequestBlockSyncMessage block_sync_request = 2;
     ProtoBlockSyncMessage block_sync_response = 3;
-    ProtoCascadeTxMessage cascade_txns = 4;
     ProtoPeerStateRootMessage peer_state_root = 5;
```
michael-yxchen: update `peer_state_root` field number to 4

rgarc:
> while the txpool clear event loopsback, new txns could have been inserted by rpc and could get into txpool first and then get cleared. one approach, txpool could have an active and secondary set of txns and at the end of proposal creation, swap active-secondary and then the clear command only clears secondary.

Thanks for bringing that up, didn't really think of that. Implemented what I think you meant in e0674a1058e6142b239f1c4c65fccf8ed971ada3 but now I'm wondering if, at swap time, secondary can already contain transactions (in other words, the assert below failing). I don't think that would be desirable.
https://github.com/monad-crypto/monad-bft/blob/e0674a1058e6142b239f1c4c65fccf8ed971ada3/monad-eth-txpool/src/lib.rs#L68

rgarc:
> > while the txpool clear event loopsback, new txns could have been inserted by rpc and could get into txpool first and then get cleared. one approach, txpool could have an active and secondary set of txns and at the end of proposal creation, swap active-secondary and then the clear command only clears secondary.
> 
> Thanks for bringing that up, didn't really think of that. Implemented what I think you meant in [e0674a1](https://github.com/monad-crypto/monad-bft/commit/e0674a1058e6142b239f1c4c65fccf8ed971ada3) but now I'm wondering if, at swap time, secondary can already contain transactions (in other words, the assert below failing). I don't think that would be desirable.
> 
> https://github.com/monad-crypto/monad-bft/blob/e0674a1058e6142b239f1c4c65fccf8ed971ada3/monad-eth-txpool/src/lib.rs#L68

> > while the txpool clear event loopsback, new txns could have been inserted by rpc and could get into txpool first and then get cleared. one approach, txpool could have an active and secondary set of txns and at the end of proposal creation, swap active-secondary and then the clear command only clears secondary.
> 
> Thanks for bringing that up, didn't really think of that. Implemented what I think you meant in [e0674a1](https://github.com/monad-crypto/monad-bft/commit/e0674a1058e6142b239f1c4c65fccf8ed971ada3) but now I'm wondering if, at swap time, secondary can already contain transactions (in other words, the assert below failing). I don't think that would be desirable.
> 
> https://github.com/monad-crypto/monad-bft/blob/e0674a1058e6142b239f1c4c65fccf8ed971ada3/monad-eth-txpool/src/lib.rs#L68

Did some more thinking. ~Another downside to this active/secondary approach is that it's non-trivial to make generic over the txpool data structure(s), which is very likely to change in the future.~

Might be easier if we defer destruction to a separate thread. But I wasn't sure if that was reasonable 

something like
```rust
struct TxPool(/* ... */);

fn create_proposal(&mut self) {
    // create_proposal logic

    let mut old_pool = std::mem::take(&mut self.0);
    std::thread::spawn(move || {
        drop(old_pool);
    });

    FullTransactionList::empty()
}
```
Would need to find something such that thread spawning is cheap. This would also obviate the need for mempool clearing events/commands.

rgarc:
### Change Since Previous Benchmark

image

### Understanding this report:

The plot on the left shows the probability of the function taking a certain amount of time. The red curve represents the saved measurements from the last time this benchmark was run, while the blue curve shows the measurements from this run. The lines represent the mean time per iteration. Click on the plot for a larger view.

PR #813, opened 2024-06-12
Add nonce checking

bamrith:


monad-eth-block-policy/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+
+impl<SCT: SignatureCollection> BlockPolicy<SCT> for EthBlockPolicy {
+    type ValidatedBlock = EthValidatedBlock<SCT>;
+
+    fn check_coherency(
+        &self,
+        block: &Self::ValidatedBlock,
+        extending_blocks: Vec<&Self::ValidatedBlock>,
+    ) -> bool {
+        let mut latest_nonces = self.latest_nonces.clone();
```
ariqchowdhury: isn't this going to be very expensive?

monad-eth-tx/src/utils.rs
ariqchowdhury, line 1: is this just used for testing? we used monad-testutils as a place for test helpers so that crates can include it under dev-dependencies

monad-eth-txpool/src/lib.rs
```diff
@@ -40,16 +35,35 @@ impl<SCT: SignatureCollection> TxPool<SCT, EthBlockPolicy> for EthTxPool {
         // TODO: when sorting by gas fees is implemented, txs should be grouped by accounts
         txs_to_propose.sort_by(|a, b| a.nonce().cmp(&b.nonce()));
 
-        for tx in txs_to_propose {
-            if pending_blocktree_txs.contains(&tx.hash()) {
-                continue;
+        // Get the latest nonces at the parent block (block to extend)
+        let mut latest_nonces = block_policy.latest_nonces.clone();
```
ariqchowdhury: same thing about cloning

monad-eth-block-policy/src/lib.rs
bamrith: yes, I was worried about that too. It would be better to fetch the nonce when verifying each transaction to avoid redundant fetches initially here. I'll change it to that.

monad-eth-tx/src/utils.rs
bamrith: yes, just for testing. there was WASM build error when using `alloy-rlp` as a dependency in `monad-testutils` which is why I put it here. I'll find a way to move it there.

ariqchowdhury: its not super high priority so if its not trivial, feel free to keep as is now and we can address in a later pr

bamrith: I'll refactor this along with the mock swarm tests

monad-eth-txpool/src/lib.rs
```diff
@@ -153,12 +146,21 @@ impl<SCT: SignatureCollection> TxPool<SCT, EthBlockPolicy> for EthTxPool {
...
-        >,
+        block_policy: &EthBlockPolicy,
+        extending_blocks: Vec<&EthValidatedBlock<SCT>>,
     ) -> FullTransactionList {
-        self.remove_pending_transactions::<SCT>(&pending_blocktree_txs);
-        self.remove_nonce_gaps();
+        // Get the latest nonce deltas at the parent block (block to extend)
+        let mut blocktree_nonce_deltas = BTreeMap::new();
+        for extending_block in extending_blocks {
+            let block_nonces = extending_block.get_nonces();
```
rgarc, line 155: This `extending_block.get_nonces()` call performs signature recovery https://github.com/monad-crypto/monad-bft/blob/90b3320fbf019f18919b8584a5a8766a0721e1b0/monad-eth-block-policy/src/lib.rs#L29-L29

Do you know if it's possible to avoid this somehow? It's an expensive operation that we wanted to avoid during `create_proposal`.

bamrith: Moved this map creation to block validation to avoid creating it every time. We need to recover signer there anyway to validate the signature.

PR #814, opened 2024-06-12
Try commit blocks when path_to_root is repaired

michael-yxchen:
Previous implementation only checks commit condition in handle proposal. When a block sync message or out-of-order proposal is received, we add the block to block tree, but do not attempt to commit.

This leads to issues with consensus restarts near epoch boundaries. Consider a scenario where, by the time the restarting node finishes syncing via block sync (requested the boundary block but not committing it), the network has moved on to the next epoch. The node never gets the chance to commit the boundary block and is stuck. This PR addresses that

### DIFF
--- a/monad-blocktree/src/blocktree.rs
+++ b/monad-blocktree/src/blocktree.rs
@@ -289,6 +289,40 @@ impl<SCT: SignatureCollection, BP: BlockPolicy<SCT>> BlockTree<SCT, BP> {
         }
     }
 
+    /// Iterate the block tree and return highest QC that have path to block tree
+    /// root and is committable, if exists
+    pub fn get_high_committable_qc(&self) -> Option<QuorumCertificate<SCT>> {
+        let mut high_commit_qc: Option<QuorumCertificate<SCT>> = None;
+        let mut iter: VecDeque<BlockId> = self.root.children_blocks.clone().into();
+        while !iter.is_empty() {
+            let bid = iter.pop_front().expect("iter non-empty");
+            let qc = self
+                .tree
+                .get(&bid)
+                .expect("block in tree")
+                .validated_block
+                .get_qc();
+            if qc.info.vote.ledger_commit_info.is_commitable()
+                && high_commit_qc
+                    .as_ref()
+                    .map(|high_commit_qc| high_commit_qc.get_round() < qc.get_round())
+                    .unwrap_or(true)
+            {
+                high_commit_qc = Some(qc.clone());
+            }
+
+            iter.extend(
+                self.tree
+                    .get(&bid)
+                    .expect("should be in tree")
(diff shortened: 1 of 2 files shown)


monad-consensus-state/src/lib.rs
```diff
@@ -599,6 +601,24 @@ where
...
+    /// Try commit any committable blocks in the block tree
+    #[must_use]
+    fn try_commit_blocktree(
+        &mut self,
+        epoch_manager: &mut EpochManager,
+        metrics: &mut Metrics,
+    ) -> Vec<ConsensusCommand<ST, SCT>> {
+        let mut cmds = Vec::new();
+        let commit_qc = self.pending_block_tree.get_committable_qc();
+        for qc in commit_qc.iter().rev() {
```
bamrith: won't the loop always break on the second QC since the first QC will commit everything ? is there a case where this doesn't happen ?

michael-yxchen: Oh yeah you're right. I shouldn't write a loop in the first place..

monad-blocktree/src/blocktree.rs
```diff
@@ -277,10 +277,40 @@ impl<SCT: SignatureCollection, BP: BlockPolicy<SCT>> BlockTree<SCT, BP> {
                     .cloned(),
             );
         }
+    }
 
-        for block_id in block_ids_to_update {
-            self.update_coherency(block_id);
```
ariqchowdhury: was this supposed to be deleted?

monad-consensus-state/src/lib.rs
```diff
@@ -365,6 +365,7 @@ where
 
         // at this point, block is valid and can be added to the blocktree
         cmds.extend(self.try_add_blocktree(&block, Some(state_root_action), node));
+        cmds.extend(self.try_commit_blocktree(node.epoch_manager, node.metrics));
```
ariqchowdhury: do we always expect to call `try_commit_blocktree` after try add? If so, I think try_add should call this new function so that it can't be forgotten.

monad-blocktree/src/blocktree.rs
michael-yxchen: It's irrelevant to this PR. I discussed with @bamrith and we agree that these lines are never run

monad-consensus-state/src/lib.rs
michael-yxchen: good point. I'm merging the two functions

michael-yxchen:
Addressed comments in #821 here, re-requesting review
Ref:
https://github.com/monad-crypto/monad-bft/pull/821#discussion_r1647896631
https://github.com/monad-crypto/monad-bft/pull/821#discussion_r1647948165

PR #815, opened 2024-06-12
Include libcgroup to fix build

philipglazman:


docker/rpc/Dockerfile
```diff
@@ -58,7 +58,8 @@ RUN apt-get install -y \
   libboost-graph1.83-dev \
   libboost-json1.83-dev \
   libboost-stacktrace1.83-dev \
-  libboost1.83-dev
+  libboost1.83-dev \
+  libcgroup-dev
```
michael-yxchen: Let's group it with other non-boost packages, after line 29

PR #818, opened 2024-06-13
Fill gas prices when missing from request

philipglazman:


monad-rpc/src/call.rs
```diff
@@ -133,7 +179,7 @@ impl TryFrom<CallRequest> for reth_primitives::transaction::Transaction {
                             .map_err(|_| JsonRpcError::invalid_params())?,
                         gas_limit: call_request
                             .gas
-                            .unwrap_or(Uint::from(10_000_000))
+                            .unwrap_or_default()
```
eerkaijun: Seems like `eth_call` fails if it defaults to 0 (I presume it will be the same for gas estimation if no `gas_limit` is set). `'intrinsic gas greater than limit'` - basically I think the gas needed is higher than the gas limit

 I feel like we can just set it to a high value that is more than sufficient for most transactions?

tongzhi1998: I would tend to agree. I had this hack in my PR. 
Also I probably should mention this earlier. I saw there are some existing types that takes care of some of the stuff for us. https://github.com/monad-crypto/monad-internal/issues/271

philipglazman: I just updated this to mimic the gas estimation algorithm found in geth and reth. I think this should solve your issue @eerkaijun  - I also updated the tx-generator to use our gas estimation (with the exception of the main tx gen loop).

PR #821, opened 2024-06-17
Forkpoint/consensus restart

michael-yxchen:
Forkpoint captures necessary state to restart a consensus node from the persistent state. Node operator can generate forkpoint from the on-disk state (ledger, validator sets, epochs, etc) and bootstrap the consensus node with the file. The consensus node then catches up with the network via statesync/blocksync.

Closes: https://github.com/monad-crypto/monad-internal/issues/120


monad-state/src/lib.rs
```diff
@@ -143,6 +155,159 @@ pub struct Forkpoint<SCT: SignatureCollection> {
...
+                .collect::<HashSet<SeqNum>>();
+
+            for state_root in self.state_roots.iter() {
+                expected_state_roots.remove(&state_root.seq_num);
+            }
+
+            if !expected_state_roots.is_empty() {
+                return Err(ForkpointValidationError::InsufficientStateRoots);
+            }
+        }
```
ariqchowdhury: I think these conditions should be put in helper methods. Should make the validate function read better.

docker/devnet/monad/config/forkpoint.toml
omegablitz, line 1: Should docket/devnet/monad/config/genesis.toml be deleted?

monad-node/Cargo.toml
```diff
@@ -54,3 +57,6 @@ tracing = { workspace = true, features = ["log"] }
 tracing-opentelemetry = { workspace = true }
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
 zeroize = { workspace = true }
+
+[dev-dependencies]
+toml = { workspace = true }
```
omegablitz: nit: toml is already a dependency

docker/devnet/monad/config/forkpoint.toml
```diff
@@ -0,0 +1,33 @@
+state_roots = []
```
omegablitz, line 1: Can we relax this to only include the validator sets and QC?

I think our init condition will be us having forkpoint + [block_n-delay, ..., block_n], so the state roots will be embedded in those blocks.

We need to have those blocks so that after we statesync to block_n-delay, we can blocksync [block_n-delay to block_n], and also so that our only trust assumption is the validator set in the forkpoint file.

monad-consensus/src/validation/signing.rs
```diff
@@ -171,10 +171,12 @@ where
         // If the node is lagging too far behind, it wouldn't know when the
         // next epoch is starting. The epoch retrieved here may be incorrect.
         // TODO: Need to check that case. Should trigger statesync.
-        let epoch = epoch_manager.get_epoch(self.obj.obj.get_round());
+        let epoch = epoch_manager
+            .get_epoch(self.obj.obj.get_round())
+            .expect("epoch verified");
```
omegablitz: Where is this verified? The first thing called in `ConsensusChildState::update` is `Unverified<ST, Unvalidated<ConsensusMessage<SCT>>>::verify` if I'm reading the code correctly.

We call `verify_epoch` in the `validate` codepaths (which comes after).

monad-consensus/src/validation/signing.rs
```diff
@@ -476,10 +478,12 @@ impl<SCT: SignatureCollection> Unvalidated<PeerStateRootMessage<SCT>> {
         // If the node is lagging too far behind, it wouldn't know when the
         // next epoch is starting. The epoch retrieved here may be incorrect.
         // TODO: Need to check that case. Should trigger statesync.
-        let epoch = epoch_manager.get_epoch(self.obj.info.round);
+        let epoch = epoch_manager
+            .get_epoch(self.obj.info.round)
+            .expect("epoch verified");
```
omegablitz: Same as above, it's not clear to me where this is verified.

monad-consensus/src/vote_state.rs
```diff
@@ -66,6 +66,13 @@ impl<SCT> VoteState<SCT>
 where
     SCT: SignatureCollection,
 {
+    pub fn new(round: Round) -> Self {
```
omegablitz, line 60: nit: might be worth deleting the `VoteState::default()` impl now that we aren't always initializing earliest_round to 0.

monad-state/src/async_state_verify.rs
```diff
@@ -100,7 +100,10 @@ where
 
                 let PeerStateRootMessage { peer, info, sig } = validated;
 
-                let epoch = self.epoch_manager.get_epoch(info.round);
+                let epoch = self
+                    .epoch_manager
+                    .get_epoch(info.round)
```
omegablitz, line 105: Question unrelated to this PR: if info.round is from some round far in the future, how do we know that the `epoch` computed here is valid?
It makes sense for the consensus messages, because we cross-reference the epoch tagged in the message with the epoch information we have locally.

monad-consensus-state/src/blocksync.rs
```diff
@@ -246,6 +246,7 @@ where
     /// Handle the response to a BlockSync request
     /// If the request was not fulfilled, the request is tried again with
     /// a different node
+    /// FIXME: does it reset timeout?
```
omegablitz, line 249: Good catch - can you create an issue for this (so we don't forget)

monad-blocktree/src/blocktree.rs
```diff
@@ -277,10 +277,40 @@ impl<SCT: SignatureCollection, BP: BlockPolicy<SCT>> BlockTree<SCT, BP> {
...
+    }
 
-        for block_id in block_ids_to_update {
-            self.update_coherency(block_id);
+    /// Iterate the block tree and return highest QC that have path to block tree
+    /// root and is committable, if exists
+    pub fn get_high_committable_qc(&self) -> Option<QuorumCertificate<SCT>> {
+        let mut high_commit_qc: Option<QuorumCertificate<SCT>> = None;
+        let mut iter: VecDeque<BlockId> = self.root.children_blocks.clone().into();
+        while !iter.is_empty() {
```
omegablitz: nit: `while let Some(bid) = iter.pop_front()` avoids the unwrap

monad-blocktree/src/blocktree.rs
```diff
@@ -277,10 +277,40 @@ impl<SCT: SignatureCollection, BP: BlockPolicy<SCT>> BlockTree<SCT, BP> {
...
+            let bid = iter.pop_front().expect("iter non-empty");
+            let qc = self
+                .tree
+                .get(&bid)
+                .expect("block in tree")
+                .validated_block
+                .get_qc();
+            if qc.info.vote.ledger_commit_info.is_commitable() {
+                // We are visiting the commit QCs in increasing rounds due to
```
omegablitz: Trying to convince myself that the commit QCs are always increasing.

image

if `k` is root, with my (likely wrong) understanding, we'd iterate over blocks `k+3` (qc_round = k), `k+1` (qc_round = k), `k+4` (qc_round = k+3), and then `k+2` (qc_round = k+1).

monad-state/src/lib.rs
```diff
@@ -117,6 +121,178 @@ impl MonadVersion {
...
+    /// A valid forkpoint must satisfy the following requirements
+    /// 1. Enough state roots to validate the next proposal it accepts
+    /// 2. Carries the validator set to verify QC
+    /// 3. QC is valid against the validator set
+    /// 4. QC epoch-round and validator set epoch-round are set correctly
+    /// 5. Has n+1 validator set locked
+    /// 6. If the last ledger block is on epoch boundary, n+2 validator set is
+    ///    locked
+    /// 7. Every validator set except the last one is scheduled to start on a
+    ///    round
```
omegablitz, line 194: Current epoch validator set invariants:

1. qc.epoch to validate qc
2. qc.epoch + 1 the validator set locked in at the end of qc.epoch - 1 , due to our n+2 epoch switching
3. qc.epoch + 2 if the ledger tip is the last block of an epoch. It locks qc.epoch + 2

I think in addition, we need the epoch of qc.round-state_root_delay - because that'll be the round that we statesync (and maybe blocksync) on.


sidenote: should we also include state_root_delay in forkpoint.toml?

monad-blocktree/src/blocktree.rs
```diff
@@ -277,10 +277,40 @@ impl<SCT: SignatureCollection, BP: BlockPolicy<SCT>> BlockTree<SCT, BP> {
                     .cloned(),
             );
         }
+    }
 
-        for block_id in block_ids_to_update {
-            self.update_coherency(block_id);
+    /// Iterate the block tree and return highest QC that have path to block tree
+    /// root and is committable, if exists
+    pub fn get_high_committable_qc(&self) -> Option<QuorumCertificate<SCT>> {
```
ariqchowdhury, line 294: add unit tests for this

monad-blocktree/src/blocktree.rs
omegablitz: We can remove the assertion and just set high_commit_qc to the highest qc

monad-consensus/src/validation/signing.rs
michael-yxchen: Yeah good catch. Will fix this and below. Note: add test for this

monad-blocktree/src/blocktree.rs
michael-yxchen: Fixed in https://github.com/monad-crypto/monad-bft/pull/814

monad-blocktree/src/blocktree.rs
michael-yxchen: Fixed in https://github.com/monad-crypto/monad-bft/pull/814

monad-state/src/async_state_verify.rs
michael-yxchen: We should revisit this when testing async state root verification. I think `epoch` should go into the messages so we can cross-check. 

Another question is which validator set should build the state root proof. It determines the epoch number to put in the async state verify messages.

Tracked here: https://github.com/monad-crypto/monad-internal/issues/301

docker/devnet/monad/config/forkpoint.toml
michael-yxchen: I'm wanting to keep it just in case we're changing the forkpoint serialization. We can generate the new format with this genesis file. Will remove in the long term

PR #823, opened 2024-06-18
Wait for consensus node to come up in RPC server

rgarc:
https://github.com/monad-crypto/monad-internal/issues/284


rgarc:
was already being handled in 
#785

PR #831, opened 2024-06-20
price gas limit ratio sorting

rgarc:
https://github.com/monad-crypto/monad-internal/issues/287

This PR also has a good side effect of helping performance. Previously we were sorting by `u128`, but now we sort by `f64`, which our boxes have a much easier time doing.

### Change Since Previous Benchmark
image
image

|  | Lower bound | Estimate |  Upper bound|
-- | -- | -- | --
Change in time | -15.649% | -14.662% | -13.748% | (p = 0.00 < 0.05)

<span style="color: rgb(0, 0, 0); font-family: &quot;Helvetica Neue&quot;; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial; display: inline !important; float: none;">Performance has improved.</span>

### Understanding this report:
The plot on the top shows the probability of the function taking a certain amount of time. The red curve represents the saved measurements from the last time this benchmark was run, while the blue curve shows the measurements from this run. The lines represent the mean time per iteration. Click on the plot for a larger view.

The plot on the bottom shows the iteration times for the two measurements. Again, the red dots represent the previous measurement while the blue dots show the current measurement.

### DIFF
--- a/monad-eth-txpool/src/lib.rs
+++ b/monad-eth-txpool/src/lib.rs
@@ -17,27 +17,31 @@ use sorted_vector_map::SortedVectorMap;
 type VirtualTimestamp = u64;
 
 /// Needed to have control over Ord implementation
-#[derive(Debug, Eq, PartialEq)]
+#[derive(Debug, PartialEq)]
 struct WrappedTransaction<'a> {
     inner: &'a EthTransaction,
-    sender: EthAddress,
     insertion_time: VirtualTimestamp,
+    price_gas_limit_ratio: f64,
+}
+
+fn effective_tip_per_gas(transaction: &EthTransaction) -> u128 {
+    match transaction.transaction {
+        Transaction::Legacy(TxLegacy { gas_price, .. })
+        | Transaction::Eip2930(TxEip2930 { gas_price, .. }) => gas_price,
+        Transaction::Eip1559(TxEip1559 {
+            max_priority_fee_per_gas,
+            ..
+        })
+        | Transaction::Eip4844(TxEip4844 {
+            max_priority_fee_per_gas,
+            ..
+        }) => max_priority_fee_per_gas,
+    }
 }
 
 impl<'a> WrappedTransaction<'a> {
     pub fn effective_tip_per_gas(&self) -> u128 {
-        match self.inner.transaction {
-            Transaction::Legacy(TxLegacy { gas_price, .. })
-            | Transaction::Eip2930(TxEip2930 { gas_price, .. }) => gas_price,
-            Transaction::Eip1559(TxEip1559 {
-                max_priority_fee_per_gas,
-                ..
-            })
-            | Transaction::Eip4844(TxEip4844 {
-                max_priority_fee_per_gas,
-                ..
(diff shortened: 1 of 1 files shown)


monad-eth-txpool/src/lib.rs
```diff
@@ -53,11 +57,14 @@ impl<'a> WrappedTransaction<'a> {
...
 
+impl<'a> Eq for WrappedTransaction<'a> {}
+
 impl<'a> Ord for WrappedTransaction<'a> {
     fn cmp(&self, other: &Self) -> Ordering {
         match self
-            .effective_tip_per_gas()
-            .cmp(&other.effective_tip_per_gas())
+            .price_gas_limit_ratio
+            .partial_cmp(&other.price_gas_limit_ratio)
```
omegablitz: nit: this could be simplified to `(self.price_gas_limit_ratio, self.insertion_time).partial_cmp(...)`

monad-eth-txpool/src/lib.rs
```diff
@@ -195,11 +214,8 @@ impl<SCT: SignatureCollection> TxPool<SCT, EthBlockPolicy> for EthTxPool {
         // loop invariant: `max_heap` contains one transaction per account (lowest nonce).
         // the root of the heap will be the best priced eligible transaction
         while !max_heap.is_empty() {
-            let WrappedTransaction {
-                inner: best_tx,
-                sender: address,
-                ..
-            } = max_heap.pop().unwrap();
+            let WrappedTransaction { inner: best_tx, .. } = max_heap.pop().unwrap();
```
omegablitz: nit: you could remove the unwrap with `while let Some(wrapped_tx) = max_heap.pop()`

monad-eth-txpool/src/lib.rs
```diff
@@ -53,11 +57,14 @@ impl<'a> WrappedTransaction<'a> {
...
+impl<'a> Eq for WrappedTransaction<'a> {}
+
 impl<'a> Ord for WrappedTransaction<'a> {
     fn cmp(&self, other: &Self) -> Ordering {
         match self
-            .effective_tip_per_gas()
-            .cmp(&other.effective_tip_per_gas())
+            .price_gas_limit_ratio
+            .partial_cmp(&other.price_gas_limit_ratio)
+            .unwrap()
```
omegablitz, line 69: nit: add comment noting that `partial_cmp` is guaranteed to return `Some(_)` if neither of the operands is NaN
> Four mutually exclusive relations are possible: less than, equal, greater than, and unordered. The last case arises when at least one operand is NaN. Every NaN shall compare unordered with everything, including itself. Comparisons shall ignore the sign of zero (so +0 = −0). Infinite operands of the same sign shall compare equal.

PR #832, opened 2024-06-20
use weighted round robin in monad-node

rgarc:
https://github.com/monad-crypto/monad-internal/issues/250


ariqchowdhury:
@michael-yxchen @rgarc 
we should create a test in flexnet for this first. use very un-even staking, make sure the committed blocks authors have the same ratio

PR #838, opened 2024-06-24
Virtual Mempool

omegablitz:
This contains an implementation of the virtual mempool. Transactions from RPC are forwarded to the next K leaders.

Ideally we can test this in mock-swarm, but need to add some infrastructure in mock-swarm for that. Might be easier to test this in flexnet in the interim.

### DIFF
--- a/monad-state/src/mempool.rs
+++ b/monad-state/src/mempool.rs
@@ -1,52 +1,129 @@
 use std::marker::PhantomData;
 
-use monad_consensus_state::command::Checkpoint;
+use bytes::Bytes;
+use itertools::Itertools;
+use monad_consensus_state::{command::Checkpoint, ConsensusState};
 use monad_consensus_types::{
     block::{Block, BlockPolicy},
     block_validator::BlockValidator,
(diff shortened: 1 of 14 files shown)


monad-state/src/mempool.rs
```diff
@@ -1,52 +1,129 @@
...
+        &mut self,
+        event: MempoolEvent<CertificateSignaturePubKey<ST>>,
+    ) -> Vec<MempoolCommand<CertificateSignaturePubKey<ST>>> {
         match event {
             MempoolEvent::UserTxns(txns) => {
+                let txs = txns
+                    .into_iter()
+                    .filter_map(|tx| {
+                        if let Err(err) = self.txpool.insert_tx(tx.clone()) {
+                            tracing::warn!("failed to insert rpc tx: {:?}", err);
```
ariqchowdhury, line 101: nit: I don't think this needs to be logged. external users could be spamming garbage txns etc

omegablitz: Garbage spam should be dropped earlier at the IPC receiver level

PR #839, opened 2024-06-24
Eth_call state overrides

tongzhi1998:


monad-rpc/src/call.rs
```diff
@@ -260,11 +260,30 @@ pub struct CallInput {
     pub data: Option<alloy_primitives::Bytes>,
 }
 
+#[derive(Debug, Default, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct StateOverride {
```
ariqchowdhury: rename to `StateOverrideObject`

monad-rpc/src/call.rs
```diff
@@ -260,11 +260,30 @@ pub struct CallInput {
...
+    pub nonce: Option<U64>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub code: Option<Bytes>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub state: Option<HashMap<B256, B256>>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub state_diff: Option<HashMap<B256, B256>>,
+}
+
+pub type StateOverrides = HashMap<Address, StateOverride>;
```
ariqchowdhury: then rename this to `StateOverrideSet`

monad-rpc/src/gas_handlers.rs
```diff
@@ -70,6 +72,13 @@ pub async fn monad_eth_estimateGas(
         (None, data) | (data, None) => data,
     };
 
+    let state_override_set = &params.state_override_set;
+    for state_override in state_override_set.values() {
+        if state_override.state.is_some() && state_override.state_diff.is_some() {
+            return Err(JsonRpcError::invalid_params());
+        }
+    }
+
```
philipglazman: Alternatively, state and state_diff can be an enum in `StateOverrideSet` to avoid writing this. And can use `serde(flatten)` to deserialize.

monad-cxx/src/lib.rs
```diff
@@ -75,14 +98,105 @@ pub fn eth_call(
...
+            cxx_address.pin_mut().push(*byte);
+        }
+        cxx_state_override_set
+            .as_mut()
+            .add_override_address(&cxx_address);
+
+        if state_override_object.balance.is_some() {
+            let mut cxx_balance = cxx::CxxVector::new();
+
+            if let Some(balance) = state_override_object.balance {
```
eerkaijun: The first `if state_override_object.balance.is_some()` is redundant - it's equivalent to `if let Some(balance) = state_override_object.balance`

monad-cxx/src/lib.rs
```diff
@@ -75,14 +98,105 @@ pub fn eth_call(
...
+            cxx_state_override_set.as_mut().set_override_nonce(
+                &cxx_address,
+                &state_override_object.nonce.unwrap().as_limbs()[0],
+            );
+        }
+
+        if state_override_object.code.is_some() {
+            let mut cxx_code = cxx::CxxVector::new();
+
+            if let Some(code) = &state_override_object.code {
```
eerkaijun: Same as above

monad-cxx/src/lib.rs
```diff
@@ -75,14 +98,105 @@ pub fn eth_call(
...
+            cxx_state_override_set
+                .as_mut()
+                .set_override_balance(&cxx_address, &cxx_balance);
+        }
+
+        if state_override_object.nonce.is_some() {
+            cxx_state_override_set.as_mut().set_override_nonce(
+                &cxx_address,
+                &state_override_object.nonce.unwrap().as_limbs()[0],
+            );
```
eerkaijun: I think in general we should use the syntax format `let Some(nonce) = state_override_object.nonce` and use `nonce` subsequently rather than unwrapping

monad-cxx/src/lib.rs
```diff
@@ -38,13 +43,37 @@ pub struct FailureCallResult {
     pub data: Option<String>,
 }
 
+// ensure that only one of {State, StateDiff} can be set
```
eerkaijun, line 46: Is this also a strict requirement is geth?

tongzhi1998: Yeah! And I think intuitively, if we set both, it doesn't make much sense since State would overwrite StateDiff.

monad-rpc/src/gas_handlers.rs
```diff
@@ -23,6 +24,7 @@ struct MonadEthEstimateGasParams {
     tx: CallRequest,
     #[serde(default, deserialize_with = "deserialize_block_tags")]
     block: BlockTags,
+    state_override_set: StateOverrideSet,
```
eerkaijun, line 28: Add `#[serde(default)]` for `state_override_set`

PR #842, opened 2024-06-25
Nonce checking tests

bamrith:


docker/flexnet/nets/consensus-nodes-nonce-test-1/scripts/net-run.sh
```diff
@@ -0,0 +1,142 @@
...
+set -ex
+# Define the function to show usage
+usage() {
+    echo "Usage: $0 --output-dir <dir_path> --net-dir <net_path> --flexnet-root <flexnet_root> --monad-bft-root <monad_bft_root>"
+    exit 1
+}
+
+
+# Initialize variables
+num_nodes=2
```
michael-yxchen, line 12: Is there a reason why we need a second node for this test?

docker/flexnet/nets/net0/rpc_blaster/scripts/run.sh
```diff
@@ -2,4 +2,7 @@
 
 mkdir /monad/logs
 
+# Wait until RPC comes up
+sleep 5
```
michael-yxchen: is this needed because rpc is taking longer than the retry period in `blaster.py` to come up?

monad-mock-swarm/tests/nonces.rs
```diff
@@ -0,0 +1,462 @@
...
+                            "Unexpected transaction in block {}. NodeID: {}, TxnHash: {}",
+                            block.get_seq_num().0,
+                            node_id,
+                            txn_hash
+                        );
+                        return false;
+                    }
+                }
+            }
+        }
```
michael-yxchen, line 176: should we assert `txns_to_see` is empty here? or is this only checking if txns in ledger are in the set of allowable txns?

monad-updaters/src/ipc.rs
```diff
@@ -26,10 +39,15 @@ where
...
     fn poll_next(
-        self: std::pin::Pin<&mut Self>,
+        mut self: std::pin::Pin<&mut Self>,
         _cx: &mut std::task::Context<'_>,
     ) -> Poll<Option<Self::Item>> {
-        // TODO: generate user tx
-        Poll::Pending
+        let txn_bytes = self.transactions.clone();
+
+        self.as_mut().transactions.clear();
```
michael-yxchen: nit: `mem::take` can achieve the same without cloning

monad-mock-swarm/tests/nonces.rs
```diff
@@ -0,0 +1,462 @@
...
+            &swarm,
+            swarm.states().keys().cloned().collect_vec(),
+            expected_txns
+        ));
+
+        swarm_ledger_verification(&swarm, 2);
+    }
+
+    #[test]
+    fn duplicate_nonces_multi_nodes() {
```
michael-yxchen, line 225: just want to note that with the virtual mempool changes, UserTxns received on the ipc updater are forwarded to the next k leaders. When leader 1 is proposing, it might have all 20 transactions in the mempool. That might be something we want to avoid if the test intend to nonce check against the pending blocks

monad-mock-swarm/tests/nonces.rs
```diff
@@ -0,0 +1,462 @@
...
+            &swarm,
+            swarm.states().keys().cloned().collect_vec(),
+            expected_txns
+        ));
+
+        swarm_ledger_verification(&swarm, 2);
+    }
+
+    #[test]
+    fn sequential_nonces_multi_nodes() {
```
michael-yxchen: similar here regarding virtual mempool

monad-updaters/src/ipc.rs
```diff
@@ -26,10 +39,15 @@ where
     type Item = MonadEvent<ST, SCT>;
 
     fn poll_next(
-        self: std::pin::Pin<&mut Self>,
+        mut self: std::pin::Pin<&mut Self>,
         _cx: &mut std::task::Context<'_>,
     ) -> Poll<Option<Self::Item>> {
-        // TODO: generate user tx
-        Poll::Pending
+        let txn_bytes = self.transactions.clone();
```
michael-yxchen: This updater yields an event even if it has no transactions. I can block events from other updaters if this one always gets polled. I think it's not happening right now because MockExecutor peek events from this updater last

Can we change it to only return Poll::Ready when self.transactions is not empty, and Poll::Pending otherwise?

docker/flexnet/nets/consensus-nodes-nonce-test-1/scripts/net-run.sh
bamrith: not really. just wanted to be sure that the node that didn't create the proposal also validates the blocks the same way.

docker/flexnet/nets/net0/rpc_blaster/scripts/run.sh
bamrith: yeah, it happened a few times locally. removed it.

monad-mock-swarm/tests/nonces.rs
bamrith: good catch. This doesn't happen immediately, so leader 1 only has 10 transactions in the mempool. But, the transactions sent to leader 2 by the test gets overwritten by the transactions forwarded from leader 1. I added a test in `monad-consensus-state` to test that behavior of checking against pending blocks. I can't really enforce that here.

monad-mock-swarm/tests/nonces.rs
bamrith: similarly, tested in `monad-consensus-state` for checking against pending blocks. removed this test since the same behavior is also tested in `commited_nonces` test right after this.

docker/flexnet/nets/full-nodes-nonce-test-1/node0/scripts/run.sh
michael-yxchen, line 1: I remember Rene added a control panel IPC argument. It would be great if you can re-run the tests before merging to confirm everything is working

monad-updaters/src/ipc.rs
```diff
@@ -26,10 +39,17 @@ where
...
     fn poll_next(
-        self: std::pin::Pin<&mut Self>,
+        mut self: std::pin::Pin<&mut Self>,
         _cx: &mut std::task::Context<'_>,
     ) -> Poll<Option<Self::Item>> {
-        // TODO: generate user tx
-        Poll::Pending
+        if self.transactions.is_empty() {
+            return Poll::Pending;
+        }
```
michael-yxchen: When the futures returns pending, it's put to sleep by the runtime and won't be waken up until the waker is called. Can you take a look at what other updaters is doing with the waker?

bamrith: Overlooked because this is only used in mock swarm tests which peeks on events (like you mentioned previously) and doesn't require wakers. Added waker now.

michael-yxchen: oh that makes sense. Thanks

PR #845, opened 2024-06-26
Add state root hash executor that polls triedb

ariqchowdhury:
This change requires monad-bft to have access to the triedb driver.

testing was done with flexnet and monad-node changed to use the real executor -- keeping the mock version in monad-node for now until we figure out a feature flag to control whether we have triedb_driver set


docker/flexnet/nets/full-nodes/scripts/net-run.sh
michael-yxchen, line 1: should we rebase this on top of https://github.com/monad-crypto/monad-bft/pull/841

monad-consensus/src/validation/signing.rs
```diff
@@ -471,9 +471,8 @@ impl<SCT: SignatureCollection> Unvalidated<PeerStateRootMessage<SCT>> {
         // If the node is lagging too far behind, it wouldn't know when the
         // next epoch is starting. The epoch retrieved here may be incorrect.
         // TODO: Need to check that case. Should trigger statesync.
-        let epoch = epoch_manager
-            .get_epoch(self.obj.info.round)
-            .ok_or(Error::InvalidEpoch)?;
+        let num_epochs = self.obj.info.seq_num / epoch_manager.val_set_update_interval;
+        let epoch = Epoch(num_epochs.0);
```
michael-yxchen: we can use `SeqNum::to_epoch` for consistency

docker/flexnet/nets/full-nodes/scripts/net-run.sh
ariqchowdhury: yeah probably

PR #849, opened 2024-06-28
Update validators at runtime

rgarc:
https://github.com/monad-crypto/monad-internal/issues/121

Builds on https://github.com/monad-crypto/monad-bft/pull/846. Adds an `update-validators` command where the user provides a TOML file of the following form which gets serialized and sent over IPC socket to an executor that creates an `ValidatorEvent::<SCT>::UpdateValidators` event.

```toml
epoch = 10
[[validators]]
node_id = "0x02093852f554d6196025b0bebe21ea646f73e34f1fd6f154e6ae77c890363ea5be"
stake = 1
cert_pubkey = "0xab1011a17be0921c122fc7362f6e0401f80623c644bb9d9e9eedff0410a4de2259f0d3a36414d4dd9dd748ace56ffb5e"
```

demo

output

there is also a stateless CLI that is more amenable to scripting
```
CLI program to manage validators and metrics

Usage: debug-node --control-panel-ipc-path <CONTROL_PANEL_IPC_PATH> <COMMAND>

Commands:
  validators         Displays the list of validators
  update-validators  Updates the validators using the provided TOML file
  clear-metrics      Clears the metrics
  help               Print this message or the help of the given subcommand(s)

Options:
  -c, --control-panel-ipc-path <CONTROL_PANEL_IPC_PATH>  
  -h, --help                                             Print help
```

```
Updates the validators using the provided TOML file

Usage: debug-node --control-panel-ipc-path <CONTROL_PANEL_IPC_PATH> update-validators --path <FILE>

Options:
  -p, --path <FILE>  Path to the TOML file
  -h, --help         Print help
```

### DIFF
--- a/monad-updaters/src/state_root_hash.rs
+++ b/monad-updaters/src/state_root_hash.rs
@@ -185,10 +185,10 @@ where
                             locked_epoch,
                             block.get_seq_num().to_epoch(self.val_set_update_interval) + Epoch(2)
                         );
-                        self.next_val_data = Some(ValidatorSetUpdate {
-                            epoch: locked_epoch,
-                            validator_data: self.genesis_validator_data.clone(),
-                        });
+                        // self.next_val_data = Some(ValidatorSetUpdate {
+                        //     epoch: locked_epoch,
+                        //     validator_data: self.genesis_validator_data.clone(),
+                        // });
                     }
 
                     wake = true;


monad-control-panel/src/ipc.rs
```diff
@@ -142,6 +145,27 @@ where
...
                         m => error!("unhandled message {:?}", m),
                     },
+                    WriteCommand::UpdateValidatorSet(update_validator_set) => {
+                        match update_validator_set {
+                            UpdateValidatorSet::Request(parsed_validator_set) => {
+                                let ParsedValidatorData::<SCT> { epoch, validators } =
+                                    parsed_validator_set;
+                                let validators =
+                                    validators.into_iter().map(Into::into).collect::<Vec<_>>();
+                                let event = MonadEvent::<ST, SCT>::ValidatorEvent(
```
rgarc, line 155: This is where we send the event that updates the validator set. Is it the correct one?

michael-yxchen: yes this is the correct event!

PR #850, opened 2024-06-28
monitoring metrics

eerkaijun:
Adding a watcher for execution metrics


omegablitz:
I'm gonna close this, because we deleted `monad-opentelemetry-executor`.
Also, if we're polling triedb for the latest block number and want to send these to an OTEL backend, I think it makes more sense to embed an OTEL metrics reporter in RPC, so that we don't need to be running another container.

PR #851, opened 2024-07-01
human readable node_name CLI argument for observability

eerkaijun:
We can pass in a name when starting the node - this should allow easier mapping on the Grafana dashboard https://github.com/monad-crypto/monad-internal/issues/124


eerkaijun:
setting the name in `node.toml` instead (it's an optional field)

PR #858, opened 2024-07-03
Reserve balance validation

mikety:
Implements a reserve balance validation per [this spec](https://www.notion.so/monad-labs/Reserve-Balance-Spec-1f35cc8a808d429d919782a36bfbabff) with carriage cost computing per [this spec](https://www.notion.so/monad-labs/Carriage-cost-spec-b3e95dc391714182a460930dc3a72696)
The transaction cost calculation is intentionally returns 0 for this commit , effectively disabling the reserve balance validation, so that all existing tests will produce the same results as before, but will be running the reserve balance validation codepath.

### DIFF
--- a/monad-eth-block-policy/src/lib.rs
+++ b/monad-eth-block-policy/src/lib.rs
@@ -1,4 +1,4 @@
-use std::collections::BTreeMap;
+use std::{cmp::min, collections::BTreeMap};
 
 use monad_consensus_types::{
     block::{Block, BlockPolicy, BlockType},
@@ -7,12 +7,141 @@ use monad_consensus_types::{
     state_root_hash::StateRootHash,
 };
 use monad_crypto::hasher::{Hashable, Hasher};
+use monad_eth_reserve_balance::{ReserveBalanceCacheResult, ReserveBalanceCacheTrait};
 use monad_eth_tx::{EthTransaction, EthTxHash};
 use monad_eth_types::{EthAddress, Nonce};
 use monad_types::{BlockId, Epoch, NodeId, Round, SeqNum};
+use sorted_vector_map::SortedVectorMap;
+use tracing::debug;
 
 pub trait AccountNonceRetrievable {
     fn get_account_nonces(&self) -> BTreeMap<EthAddress, Nonce>;
+}
+
+pub enum ComputeReserveBalanceResult {
+    Val(u128),
+    TrieDBNone, // No entry in Triedb
+    NeedSync,   // Triedb needs sync
+    Spent,      // Spent by the last execution-delay blocks carriage
+}
+
+pub enum ReserveBalanceCheck {
+    Insert,
+    Propose,
+    Validate,
+}
+
+pub fn compute_intrinsic_gas(txn: &EthTransaction) -> u128 {
+    // TODO implement intrinsic_gas formula
+    0
+}
+
+pub fn compute_txn_carriage_cost(txn: &EthTransaction) -> u128 {
+    let max_fee_per_gas = txn.max_fee_per_gas();
+    let intrinsic_gas = compute_intrinsic_gas(txn);
+    intrinsic_gas * max_fee_per_gas
+}
+
+// Computes reserve balance available for the account
+pub fn compute_reserve_balance<SCT: SignatureCollection, RBCT: ReserveBalanceCacheTrait>(
+    consensus_block_seq_num: SeqNum,
+    cache: &mut RBCT,
+    block_policy: &EthBlockPolicy,
+    extending_blocks: Option<&Vec<&EthValidatedBlock<SCT>>>,
+    eth_address: &EthAddress,
+) -> ComputeReserveBalanceResult {
+    let mut reserve_balance: u128;
+    let mut block_seq_num: SeqNum;
+
+    match cache.get_account_balance(consensus_block_seq_num, eth_address) {
+        ReserveBalanceCacheResult::Val(seq_num, acc_balance) => {
+            reserve_balance = min(acc_balance, block_policy.max_reserve_balance);
+            block_seq_num = seq_num;
+            debug!(
+                "ReserveBalance compute 1: \
+                        balance from cache: {:?} \
+                        for TDB block: {:?} \
+                        for address: {:?}",
+                reserve_balance, block_seq_num, eth_address
+            );
+        }
+        ReserveBalanceCacheResult::NeedSync => {
+            debug!(
+                "ReserveBalance compute 2: \
+                        triedb needs sync \
+                        consensus block seq num: {:?} \
+                        for address: {:?}",
+                consensus_block_seq_num, eth_address
+            );
+            return ComputeReserveBalanceResult::NeedSync;
+        }
+        ReserveBalanceCacheResult::None => {
+            //Error
+            return ComputeReserveBalanceResult::TrieDBNone;
+        }
+    }
+
+    // Apply Carriage Cost for the txns from committed blocks
+    let carriage_cost_committed: u128 =
+        block_policy.compute_committed_txns_carriage_cost(eth_address, &block_seq_num);
+
+    if reserve_balance < carriage_cost_committed {
+        debug!(
+            "ReserveBalance compute 3: \
+                    Not sufficient balance: {:?} \
+                    Carriage Cost Committed: {:?} \
+                    consensus block:seq num {:?} \
+                    for address: {:?}",
+            reserve_balance, carriage_cost_committed, consensus_block_seq_num, eth_address
+        );
+        return ComputeReserveBalanceResult::Spent;
+    } else {
+        reserve_balance -= carriage_cost_committed;
+        debug!(
+            "ReserveBalance compute 4: \
+                    updated balance to: {:?} \
+                    Carriage Cost Committed: {:?} \
+                    consensus block:seq num {:?} \
+                    for address: {:?}",
+            reserve_balance, carriage_cost_committed, consensus_block_seq_num, eth_address
+        );
+    }
+
+    // Apply Carriage Cost for txns in extending blocks
+    let mut carriage_cost_pending: u128 = 0;
+    if let Some(blocks) = extending_blocks {
+        for extending_block in blocks {
+            for txn in &extending_block.validated_txns {
+                if EthAddress(txn.signer()) == *eth_address {
+                    carriage_cost_pending += compute_txn_carriage_cost(txn);
+                }
+            }
+        }
+    }
+
+    if reserve_balance < carriage_cost_pending {
+        debug!(
+            "ReserveBalance compute 5: \
+                    Not sufficient balance: {:?} \
+                    Carriage Cost Pending: {:?} \
+                    consensus block:seq num {:?} \
+                    for address: {:?}",
+            reserve_balance, carriage_cost_pending, consensus_block_seq_num, eth_address
+        );
+        return ComputeReserveBalanceResult::Spent;
+    } else {
+        reserve_balance -= carriage_cost_pending;
+        debug!(
+            "ReserveBalance compute 6: \
+                    updated balance to: {:?} \
+                    Carriage Cost Pending: {:?} \
+                    consensus block:seq num {:?} \
+                    for address: {:?}",
+            reserve_balance, carriage_cost_pending, consensus_block_seq_num, eth_address
+        );
+    }
+
+    ComputeReserveBalanceResult::Val(reserve_balance)
 }
 
 /// A consensus block that has gone through the EthereumValidator and makes the decoded and
@@ -149,6 +278,21 @@ pub struct EthBlockPolicy {
 
     /// SeqNum of last committed block
     pub last_commit: SeqNum,
+
+    // last execution-delay committed blocks
+    pub txn_cache: SortedVectorMap<SeqNum, Vec<EthTransaction>>,
+
+    /// Maximum reserve balance enforced by execution
+    pub max_reserve_balance: u128,
+
+    /// Cost for including transaction in the consensus
+    pub execution_delay: u64,
+
+    /// lowest-order bit 0 set: enable check for insert_tx
+    /// lowest-order bit 1 set: enable check for create_proposal
+    /// lowest-order bit 2 set: enable check for validation
+    /// i.e. 0b00000111 all reserve balance checks are ehabled
+    pub reserve_balance_check_mode: u8,
 }
 
 impl EthBlockPolicy {
@@ -170,15 +314,67 @@ impl EthBlockPolicy {
             todo!("triedb read: nonce in triedb is always next nonce");
         }
     }
+
+    pub fn get_committed_block_seq_num(&self) -> Option<SeqNum> {
+        self.txn_cache.last_key_value().map(|seq_num| *seq_num.0)
+    }
+
+    pub fn update_txn_cache(&mut self, seq: SeqNum, txns: &[EthTransaction]) {
+        let txn_cache_size: usize = self.execution_delay.try_into().unwrap();
+        if self.txn_cache.len() >= txn_cache_size * 2 {
+            let mut idx = 0;
+            let mut divider: Option<SeqNum> = None;
+            for (key, _) in &self.txn_cache {
+                if idx == txn_cache_size {
+                    divider = Some(*key);
+                    break;
+                }
+                idx += 1;
+            }
+            if divider.is_some() {
+                self.txn_cache.split_off(&divider.unwrap());
+            }
+        }
+        self.txn_cache.insert(seq, txns.to_owned());
+    }
+
+    pub fn compute_committed_txns_carriage_cost(
+        &self,
+        eth_address: &EthAddress,
+        triedb_block_seq_num: &SeqNum,
+    ) -> u128 {
+        let mut carriage_cost: u128 = 0;
+        for item in &self.txn_cache {
+            if item.0 > triedb_block_seq_num {
+                for txn in item.1 {
+                    if EthAddress(txn.signer()) == *eth_address {
+                        carriage_cost += compute_txn_carriage_cost(txn);
+                    }
+                }
+            }
+        }
+        carriage_cost
+    }
+
+    pub fn reserve_balance_check_enabled(&self, mode: ReserveBalanceCheck) -> bool {
+        match mode {
+            ReserveBalanceCheck::Insert => self.reserve_balance_check_mode & 0b00000001 > 0,
+            ReserveBalanceCheck::Propose => self.reserve_balance_check_mode & 0b00000010 > 0,
+            ReserveBalanceCheck::Validate => self.reserve_balance_check_mode & 0b00000100 > 0,
+        }
+    }
 }
 
-impl<SCT: SignatureCollection> BlockPolicy<SCT> for EthBlockPolicy {
+impl<SCT: SignatureCollection, RBCT: ReserveBalanceCacheTrait> BlockPolicy<SCT, RBCT>
+    for EthBlockPolicy
+{
     type ValidatedBlock = EthValidatedBlock<SCT>;
 
     fn check_coherency(
         &self,
         block: &Self::ValidatedBlock,
         extending_blocks: Vec<&Self::ValidatedBlock>,
+        reserve_balance_cache: &mut RBCT,
     ) -> bool {
         assert_eq!(
             extending_blocks
@@ -197,10 +393,85 @@ impl<SCT: SignatureCollection> BlockPolicy<SCT> for EthBlockPolicy {
             let txn_nonce = txn.nonce();
 
             let expected_nonce = self.get_account_nonce(&eth_address, &pending_account_nonces);
+
             if txn_nonce != expected_nonce {
                 return false;
             }
-            pending_account_nonces.insert(eth_address, txn_nonce + 1);
+            let res = compute_reserve_balance::<SCT, RBCT>(
+                block.get_seq_num(),
+                reserve_balance_cache,
+                self,
+                Some(&extending_blocks),
+                &eth_address,
+            );
+            let reserve_balance = match res {
+                ComputeReserveBalanceResult::Val(val) => {
+                    debug!(
+                        "ReserveBalance - check_coherency 1: \
+                                reserve balance {:?} \
+                                consensus block:seq num {:?} \
+                                for address: {:?}",
+                        val,
+                        block.get_seq_num(),
+                        eth_address
+                    );
+                    val
+                }
+                ComputeReserveBalanceResult::TrieDBNone => {
+                    /* reserve balance is 0 */
+                    debug!(
+                        "ReserveBalance - check_coherency 2: \
+                                TrieDBNone \
+                                consensus block:seq num {:?} \
+                                for address: {:?}",
+                        block.get_seq_num(),
+                        eth_address
+                    );
+                    0
+                }
+                ComputeReserveBalanceResult::NeedSync => {
+                    /* TODO implement waiting, reserve balance is 0 for now */
+                    debug!(
+                        "ReserveBalance - check_coherency 3: \
+                                NeedSync \
+                                consensus block:seq num {:?} \
+                                for address: {:?}",
+                        block.get_seq_num(),
+                        eth_address
+                    );
+                    0
+                }
+                ComputeReserveBalanceResult::Spent => {
+                    debug!(
+                        "ReserveBalance - check_coherency 4: \
+                                Not sufficient balance: spent. \
+                                consensus block:seq num {:?} \
+                                for address: {:?}",
+                        block.get_seq_num(),
+                        eth_address
+                    );
+                    0
+                }
+            };
+
+            let txn_carriage_cost = compute_txn_carriage_cost(txn);
+
+            if reserve_balance >= txn_carriage_cost {
+                pending_account_nonces.insert(eth_address, txn_nonce + 1);
+            } else {
+                debug!(
+                    "ReserveBalance - check_coherency 5: \
+                            Not sufficient balance: {:?} \
+                            Txn Carriage Cost: {:?} \
+                            consensus block:seq num {:?} \
+                            for address: {:?}",
+                    reserve_balance,
+                    txn_carriage_cost,
+                    block.get_seq_num(),
+                    eth_address
+                );
+                return false;
+            }
         }
 
         true
@@ -213,5 +484,6 @@ impl<SCT: SignatureCollection> BlockPolicy<SCT> for EthBlockPolicy {
         for (address, account_nonce) in committed_block_account_nonces {
             self.account_nonces.insert(address, account_nonce);
         }
+        self.update_txn_cache(block.get_seq_num(), &block.validated_txns);
     }
 }
--- a/monad-eth-reserve-balance/src/lib.rs
+++ b/monad-eth-reserve-balance/src/lib.rs
@@ -1,0 +1,45 @@
+use std::path::PathBuf;
+
+use monad_eth_types::EthAddress;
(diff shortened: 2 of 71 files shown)


monad-eth-txpool/src/lib.rs
```diff
@@ -90,20 +93,182 @@ impl TransactionGroup {
...
+}
+
+fn compute_reserve_balance_for_propose<SCT: SignatureCollection>(
+    block_policy: &EthBlockPolicy,
+    extending_blocks: &Vec<&EthValidatedBlock<SCT>>,
+    eth_address: &EthAddress,
+    tdb_path: &Path) -> ComputeReserveBalanceResult {
+
+    let handle = TriedbHandle::try_new(tdb_path).unwrap();
+    let triedb_block_id = handle.latest_block();
```
bamrith: Just a quick note here. I think the block id we use here should be exactly "execution delay" blocks behind the proposing block number instead of fetching the latest executed block id from TrieDB. This will ensure that all the nodes that validate this proposal will also fetch the same block id from TrieDB since if some nodes have faster execution, the latest block id will be different for them.

monad-eth-txpool/src/lib.rs
```diff
@@ -90,20 +93,182 @@ impl TransactionGroup {
...
+                eth_address,
+                reserve_balance);
+    }
+    else {
+        debug!("ReserveBalance propose 2: \
+                 Cant get account balance from TDB for address: {:?}", eth_address);
+        return ComputeReserveBalanceResult::TrieDBNone;
+    }
+    // Apply Carriage Cost for the txns from committed blocks
+    let txn_cnt: u128 = block_policy.count_txns_for_address(eth_address);
```
bamrith: Adding onto that, we should also make sure that the block numbers here don't overlap with the block id used to retrieve account balance from TrieDB. 
As a quick example, say the execution delay is 10 blocks, the last committed block is 55, and the proposing block is 60. Then the TrieDB account balance should be retrieved from block (60 - 10) = block 50, then add carriage cost for transactions in committed blocks 51-55, then add carriage cost for transactions in extending blocks 56-59.

monad-eth-txpool/src/lib.rs
mikety: Done

monad-eth-txpool/src/lib.rs
mikety: Done

monad-eth-block-policy/src/lib.rs
```diff
@@ -8,7 +10,156 @@ use monad_consensus_types::{
...
+
+pub enum ReserveBalanceCheck {
+    Insert,
+    Propose,
+    Validate,
+}
+
+pub fn compute_intrinsic_gas(txn: &EthSignedTransaction) -> u128 {
+   // TODO implement intrinsic_gas formula
+   0
```
mikety: 0 txn cost means all txn will pass as before, intending to isolate non cost related issues

monad-eth-block-policy/src/lib.rs
```diff
@@ -130,6 +299,73 @@ impl EthBlockPolicy {
...
     }
+
+    pub fn compute_triedb_block_id(&self, triedb_latest_block_id: u64) -> u64 {
+        let mut triedb_block_id = 0;
+        if let Some(committed_seq_num) = self.get_committed_block_seq_num() {
+            if committed_seq_num.0 > self.execution_delay {
+                triedb_block_id = committed_seq_num.0 - self.execution_delay;
+            }
+            if (triedb_block_id > triedb_latest_block_id) {
+                triedb_block_id = triedb_latest_block_id;
```
mikety: This will be changed to just wait for triedb to sync up and do no propose or validation until then

monad-eth-block-policy/src/lib.rs
```diff
@@ -8,7 +10,156 @@ use monad_consensus_types::{
...
+    intrinsic_gas * max_fee_per_gas
+}
+
+// Computes reserve balance available for the account
+pub fn compute_reserve_balance<SCT: SignatureCollection>(
+        cache: &mut ReserveBalanceCache,
+        block_policy: &EthBlockPolicy,
+        extending_blocks: &Vec<&EthValidatedBlock<SCT>>,
+        eth_address: &EthAddress) -> (SeqNum, ComputeReserveBalanceResult) {
+        let mut reserve_balance: u128 = 0;
```
michael-yxchen: I think this function should use the sequence number of the validating block so it knows which triedb state to query. Looks like it's fetching the latest triedb block right now, which may not be synchronized among all nodes. Maybe a different commit is pushed?

mikety: It uses the latest committed block SeqNum known to the node. Per the offline discussion changed to use SeqNum of the block being validated and proposed.

monad-eth-block-policy/src/lib.rs
```diff
@@ -8,7 +10,159 @@ use monad_consensus_types::{
...
+                        latest block_id: {:?} \
+                        for address: {:?}",
+                        block_id,
+                        eth_address);
+                cache.purge();
+            }
+            _ => ()
+        }
+
+        let handle = TriedbHandle::try_new(block_policy.triedb_path.as_path()).unwrap();
```
bamrith: This might be slow. I think it's better to store the handle instead of creating it every time.

monad-eth-block-policy/src/lib.rs
```diff
@@ -8,7 +10,159 @@ use monad_consensus_types::{
...
+
+pub fn compute_txn_carriage_cost(txn: &EthSignedTransaction) -> u128 {
+    let max_fee_per_gas = txn.max_fee_per_gas();
+    let intrinsic_gas = compute_intrinsic_gas(txn);
+    intrinsic_gas * max_fee_per_gas
+}
+
+// Computes reserve balance available for the account
+pub fn compute_reserve_balance<SCT: SignatureCollection>(
+        block_id: SeqNum,
```
michael-yxchen: nit: every block is associated with a block_id, which is a hash over the block. I'd change the variable name to avoid confusion

monad-eth-block-policy/src/lib.rs
```diff
@@ -130,6 +302,83 @@ impl EthBlockPolicy {
...
+                // TODO consensus node needs to wait for the triedb blocks to catch up.
+                debug!("Compute TDB block_id: the latest TDB block_id: {:?} is less than the computed TDB block_id: {:?} :\
+                       will use the latest TDB block_id available",
+                    triedb_latest_block_id,
+                    triedb_block_id);
+                triedb_block_id = triedb_latest_block_id;
+            }
+        }
+        else {
+            triedb_block_id = triedb_latest_block_id;
```
michael-yxchen: if `block.seq_num < execution_delay`, we should use the genesis block sequence number. `triedb_latest` may not be synchronized across nodes

monad-eth-reserve-balance/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+
+use tracing::{info, debug};
+
+#[derive(Clone, Debug, Default)]
+pub struct ReserveBalanceCache {
+    cache: BTreeMap<EthAddress, u128>,
+    // None - the cache is purged.
+    last_block_id: Option<SeqNum>
+}
+
```
michael-yxchen: I'm thinking we can cache the account/reserve balances at the last few blocks, instead of one, as we might not be validating higher blocks always, w/ blocktree forks. The cache for block `N` would be purged when block `N + delay` is committed. After that block `N` reserve balance is never queried. I may be off by 1 here for mempool insert check, i.e. clear cache when `N + delay + 1` is committed

monad-eth-block-policy/src/lib.rs
```diff
@@ -8,7 +10,159 @@ use monad_consensus_types::{
...
+    let intrinsic_gas = compute_intrinsic_gas(txn);
+    intrinsic_gas * max_fee_per_gas
+}
+
+// Computes reserve balance available for the account
+pub fn compute_reserve_balance<SCT: SignatureCollection>(
+        block_id: SeqNum,
+        cache: &mut ReserveBalanceCache,
+        block_policy: &EthBlockPolicy,
+        extending_blocks: Option<&Vec<&EthValidatedBlock<SCT>>>,
```
michael-yxchen: nit: this doesn't need to be in an `Option<_>`

monad-eth-block-policy/src/lib.rs
```diff
@@ -130,6 +302,83 @@ impl EthBlockPolicy {
...
+                idx = idx + 1;
+            }
+            if divider.is_some() {
+                self.txn_cache.split_off(&divider.unwrap());
+            }
+        }
+        self.txn_cache.insert(seq, txns.clone());
+    }
+
+    pub fn compute_committed_txns_carriage_cost(&self, eth_address: &EthAddress, triedb_block_id: u64) -> u128 {
```
michael-yxchen: nit: use SeqNum type for `triedb_block_id`, also change the variable name

monad-eth-block-policy/src/lib.rs
```diff
@@ -114,6 +268,24 @@ pub struct EthBlockPolicy {
     /// SeqNum of next block to be committed
     /// Not last_commit because first block is SeqNum(0)
     pub next_commit: SeqNum,
+
+    // pub committed_block_seq_num: Option<SeqNum>,
+    // last K committed blocks
+    pub txn_cache: SortedVectorMap<SeqNum, Vec<EthSignedTransaction>>,
```
michael-yxchen: the cache should use `EthTransaction`, which holds the signer. This avoids the expensive `recover_sender` call. Might need to rebase to include changes from https://github.com/monad-crypto/monad-bft/pull/865

monad-eth-block-policy/src/lib.rs
```diff
@@ -8,7 +10,159 @@ use monad_consensus_types::{
...
+        else {
+            debug!("ReserveBalance compute 4: \
+                    Cant get account balance from TDB for address: {:?} block_id {:?}", eth_address, block_id);
+            return ComputeReserveBalanceResult::TrieDBNone;
+        }
+
+        // Apply Carriage Cost for the txns from committed blocks
+        let carriage_cost_committed: u128 = block_policy.compute_committed_txns_carriage_cost(eth_address, triedb_block_id);
+
+        if reserve_balance < carriage_cost_committed {
```
michael-yxchen: should be asserted

monad-eth-block-policy/src/lib.rs
```diff
@@ -8,7 +10,159 @@ use monad_consensus_types::{
...
+                    if EthAddress(txn.recover_signer().unwrap()) == *eth_address {
+                        carriage_cost_pending = carriage_cost_pending + compute_txn_carriage_cost(txn);
+                    }
+                }
+            }
+        }
+
+        if reserve_balance < carriage_cost_pending {
+            debug!("ReserveBalance compute 7: \
+                    Not sufficient balance: {:?} \
```
michael-yxchen, line 125: should also assert, because the extending blocks are coherent

monad-eth-block-policy/src/lib.rs
```diff
@@ -8,7 +10,159 @@ use monad_consensus_types::{
...
+                    latest block_id: {:?} \
+                    for address: {:?}",
+                    reserve_balance,
+                    carriage_cost_pending,
+                    block_id,
+                    eth_address);
+        }
+        // Balalnce cache  keeps track of the latest _committed_ block
+        cache.update(&block_id, eth_address, reserve_balance);
+
```
michael-yxchen, line 143: reserve balance at this point is dependent on committed blocks and extending blocks. I think we should cache the raw triedb results instead, to maximize reuse. Traversing the committed block cache and extending blocks should be cheap

monad-eth-txpool/src/lib.rs
```diff
@@ -96,12 +100,16 @@ type Pool = SortedVectorMap<EthAddress, TransactionGroup>;
 pub struct EthTxPool {
     pool: Pool,
     garbage: Vec<Pool>,
+    reserve_balance_cache: ReserveBalanceCache,
```
michael-yxchen: if we only cache triedb values, this cached can be shared between TxPool and BlockPolicy

monad-eth-txpool/src/lib.rs
```diff
@@ -117,11 +125,44 @@ impl EthTxPool {
...
+                    ComputeReserveBalanceResult::TrieDBNone => { /* reserve balance is 0 */}
+                    _ => ()
+                }
+
+                for (nonce, txn) in &transaction_group.transactions {
                     if high_nonce.is_some() && *nonce != high_nonce.unwrap() + 1 {
                         maybe_nonce_to_remove = Some(*nonce);
                         break;
                     }
+                    let signed_txn = txn.0.clone();
```
michael-yxchen: nit: unnecessary clone

monad-eth-txpool/src/lib.rs
```diff
@@ -96,12 +100,16 @@ type Pool = SortedVectorMap<EthAddress, TransactionGroup>;
 pub struct EthTxPool {
     pool: Pool,
     garbage: Vec<Pool>,
+    reserve_balance_cache: ReserveBalanceCache,
 }
 
 impl EthTxPool {
-    fn remove_invalid_nonces(
+
+    fn remove_invalid_nonces<SCT: SignatureCollection>(
```
michael-yxchen: not sure about the performance impact. Lifting the reserve balance evictions to a different functions might improve readability

monad-eth-block-policy/src/lib.rs
```diff
@@ -149,9 +398,10 @@ impl<SCT: SignatureCollection> BlockPolicy<SCT> for EthBlockPolicy {
                 .get_seq_num(),
             self.next_commit
         );
+        let mut validation_cache = ReserveBalanceCache::new();
```
michael-yxchen: If caching triedb accnt balances, this cache can be reused by multiple check_coherency calls and potentially shared with txpool

monad-eth-txpool/src/lib.rs
```diff
@@ -117,11 +125,44 @@ impl EthTxPool {
...
+                for (nonce, txn) in &transaction_group.transactions {
                     if high_nonce.is_some() && *nonce != high_nonce.unwrap() + 1 {
                         maybe_nonce_to_remove = Some(*nonce);
                         break;
                     }
+                    let signed_txn = txn.0.clone();
+                    let txn_carriage_cost = compute_txn_carriage_cost(&signed_txn);
+
+                    if reserve_balance >= txn_carriage_cost { 
+                        self.reserve_balance_cache.update(&proposed_seq_num, &eth_address, reserve_balance - txn_carriage_cost);
```
michael-yxchen: I'm not quite understanding the cache use in EthTxPool. Both `insert_tx` and `remove_invalid_nonces`, which is called from `create_proposal`, accesses and updates the cache. `create_proposal` is called with a higher sequence number. Doesn't that invalidates the cache?

monad-eth-block-policy/src/lib.rs
```diff
@@ -1,14 +1,164 @@
...
+            debug!(
+                "ReserveBalance compute 2: \
+                        cache is stale. \
+                        latest block_id: {:?} \
+                        for address: {:?}",
+                block_id, eth_address
+            );
+            cache.purge();
+        }
+        _ => (),
```
ariqchowdhury: `ReserveBalanceCacheResult::Mismatch` comment says it should never happen so we should match on that and log. Should also consider panicking for now until more testing is complete and can then remove the panic

monad-eth-block-policy/src/lib.rs
```diff
@@ -1,14 +1,164 @@
...
+            reserve_balance, carriage_cost_committed, block_id, eth_address
+        );
+    }
+
+    // Apply Carriage Cost for txns in extending blocks
+    let mut carriage_cost_pending: u128 = 0;
+    if let Some(blocks) = extending_blocks {
+        for extending_block in blocks {
+            for txn in &extending_block.validated_txns {
+                if EthAddress(txn.recover_signer().unwrap()) == *eth_address {
```
ariqchowdhury: should be able to use `txn.signer()` instead of recover_signer. The type of txn in validated_txns has already done the signature recovery as part of the validation and stores it in a field along-side the raw txn. 
Calling recover_signer will do the cryptographic operation and is expensive

monad-eth-block-policy/src/lib.rs
ariqchowdhury: agreed, it seems more intuitive to me for the cache to be raw triedb results

monad-eth-block-policy/src/lib.rs
```diff
@@ -1,14 +1,164 @@
...
+        debug!(
+            "ReserveBalance insert 8: \
+                    updated balance to: {:?} \
+                    Carriage Cost Pending: {:?} \
+                    latest block_id: {:?} \
+                    for address: {:?}",
+            reserve_balance, carriage_cost_pending, block_id, eth_address
+        );
+    }
+    // Balalnce cache  keeps track of the latest _committed_ block
```
ariqchowdhury: nit: typo

monad-eth-block-policy/src/lib.rs
```diff
@@ -1,14 +1,164 @@
...
 use monad_eth_tx::{EthTransaction, EthTxHash};
 use monad_eth_types::{EthAddress, Nonce};
+use monad_triedb::Handle as TriedbHandle;
 use monad_types::{BlockId, Epoch, NodeId, Round, SeqNum};
+use sorted_vector_map::SortedVectorMap;
+use tracing::debug;
+
+pub enum ComputeReserveBalanceResult {
+    Val(u128),
+    TrieDBNone, // No entrhy in TrieDB
```
ariqchowdhury: nit: typo

monad-eth-block-policy/src/lib.rs
```diff
@@ -113,6 +263,24 @@ pub struct EthBlockPolicy {
 
     /// SeqNum of last committed block
     pub last_commit: SeqNum,
+
+    // pub committed_block_seq_num: Option<SeqNum>,
```
ariqchowdhury: delete?

monad-eth-block-policy/src/lib.rs
```diff
@@ -113,6 +263,24 @@ pub struct EthBlockPolicy {
...
+    // last K committed blocks
```
ariqchowdhury: nit: "last execution-delay committed blocks"

monad-eth-block-policy/src/lib.rs
```diff
@@ -113,6 +263,24 @@ pub struct EthBlockPolicy {
...
+    pub txn_cache: SortedVectorMap<SeqNum, Vec<EthTransaction>>,
+
+    /// Maximum reserve balance enforced by execution
+    pub max_reserve_balance: u128,
+
+    /// Cost for including transaction in the consensus
```
ariqchowdhury, line 288: comment is out of date?

monad-eth-block-policy/src/lib.rs
```diff
@@ -129,6 +297,79 @@ impl EthBlockPolicy {
 
         None
     }
+
+    pub fn compute_triedb_block_id(&self, block_id: SeqNum, triedb_latest_block_id: u64) -> u64 {
```
ariqchowdhury: nit: execution code might refer to this as block_id but in consensus code it becomes confusing because block_id is the hash of the block header; instead we would call this block_num.

monad-eth-block-policy/src/lib.rs
ariqchowdhury: I think this function should return an `Option<u64>` and if triedb is lagging behind, return None. The caller of this function should have to decide what to do in that situation (avoid voting, proposing etc.)

monad-eth-block-policy/src/lib.rs
```diff
@@ -1,14 +1,164 @@
...
+                block_id, eth_address
+            );
+            cache.purge();
+        }
+        _ => (),
+    }
+
+    let handle = TriedbHandle::try_new(block_policy.triedb_path.as_path()).unwrap();
+    let triedb_latest_block_id = handle.latest_block();
+    let triedb_block_id = block_policy.compute_triedb_block_id(block_id, triedb_latest_block_id);
```
ariqchowdhury: related to comment left below -- if this function could return an option instead. And if it returns None here, I think it makes sense for this function to return TrieDBNone.

monad-eth-block-policy/src/lib.rs
```diff
@@ -129,6 +297,79 @@ impl EthBlockPolicy {
...
+    pub fn compute_committed_txns_carriage_cost(
+        &self,
+        eth_address: &EthAddress,
+        triedb_block_id: u64,
+    ) -> u128 {
+        let mut carriage_cost: u128 = 0;
+        for item in &self.txn_cache {
+            if item.0 .0 > triedb_block_id {
+                for txn in item.1 {
+                    if EthAddress(txn.recover_signer().unwrap()) == *eth_address {
```
ariqchowdhury: same as above, should be `txn.signer()`

monad-eth-block-policy/src/lib.rs
```diff
@@ -171,7 +412,38 @@ impl<SCT: SignatureCollection> BlockPolicy<SCT> for EthBlockPolicy {
             if txn_nonce != expected_nonce {
                 return false;
             }
-            blocktree_nonce_deltas.insert(eth_address, txn_nonce);
+            let mut reserve_balance: u128 = 0;
```
ariqchowdhury: nit: can write this as 
`let reserve_balance = match res {
...
}`
also has benefit that it doesn't have to mut and the value set in the TrieDBNone case is explicit

monad-eth-reserve-balance/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+#[derive(Clone, Debug, Default)]
+pub struct ReserveBalanceCache {
+    cache: BTreeMap<EthAddress, u128>,
+    // None - the cache is purged.
+    last_block_id: Option<SeqNum>,
+}
+
+// exists for the current block and address - return
+// does not existst for the current block and address - need to compute
```
ariqchowdhury: nit: typo

monad-eth-txpool/src/lib.rs
```diff
@@ -117,11 +126,58 @@ impl EthTxPool {
...
-                for (nonce, _) in &transaction_group.transactions {
+                let mut reserve_balance: u128 = 0;
+                let res = compute_reserve_balance::<SCT>(
+                    proposed_seq_num,
+                    &mut self.reserve_balance_cache,
+                    block_policy,
+                    Some(&extending_blocks),
+                    eth_address,
+                );
+                match res {
```
ariqchowdhury: nit: same thing about `let reserve_balance = match...`

ariqchowdhury:
Thinking out loud about how the code is structured...I think there could be a struct in monad_state that holds the triedb related stuff -- the triedb_handle and the reserve_balance_cache. If the cache is just the raw triedb values as Michael suggested above, then this new struct can be passed to txpool or block_policy as needed and the triedb_handle doesn't have to be created each time either.

The struct would basically abstract reading reserve_balance from triedb so txpool and block_policy don't have to be aware of triedb handles or that there is a cache -- they just get an interface to read reserve balance.

monad-eth-block-policy/src/lib.rs
mikety: Done

monad-eth-block-policy/src/lib.rs
mikety: Done

monad-eth-txpool/src/lib.rs
mikety: Done

monad-eth-block-policy/src/lib.rs
mikety: Done

monad-eth-txpool/src/lib.rs
mikety: Updated cache implementation

monad-eth-block-policy/src/lib.rs
mikety: done

monad-eth-block-policy/src/lib.rs
mikety: done

monad-eth-block-policy/src/lib.rs
mikety: done

monad-eth-block-policy/src/lib.rs
mikety: done, note that triedb bindings return u64 for triedb block ids so there is need to convert SeqNum to u64.

monad-eth-block-policy/src/lib.rs
mikety: done

monad-eth-block-policy/src/lib.rs
mikety: done

monad-eth-block-policy/src/lib.rs
mikety: done

monad-eth-block-policy/src/lib.rs
mikety: done

monad-eth-block-policy/src/lib.rs
mikety: done

monad-eth-block-policy/src/lib.rs
mikety: updated

monad-eth-block-policy/src/lib.rs
mikety: Its used in the insert_tx that does not pass extanding blocks

monad-eth-block-policy/src/lib.rs
mikety: I will follow the stricter checks in the follow up txn, that will also enable intrinsic gas compute

monad-eth-reserve-balance/src/lib.rs
mikety: Implemented a bit more conservative cache purging at purge_old_blocks - will keep at least "execution_delay +1"  blocks _in the cache_

monad-eth-block-policy/src/lib.rs
mikety: I plan to assert in the follow up txn

monad-eth-txpool/src/lib.rs
mikety: done

monad-eth-block-policy/src/lib.rs
mikety: Removed Mismatch, not relevant anymore

monad-eth-block-policy/src/lib.rs
mikety: done

monad-eth-block-policy/src/lib.rs
mikety: done

monad-eth-block-policy/src/lib.rs
mikety: done

monad-eth-block-policy/src/lib.rs
mikety: done

monad-eth-txpool/src/lib.rs
mikety: done

monad-cxx/monad-execution
omegablitz, line 1: Looks like the execution version was reverted to a prior version - I assume this wasn't intentional?

mikety: Thank you! Fixed

PR #861, opened 2024-07-05
Initial Flexnet Testing Library

eclark0426:
There's still more to be done with this, since we have the meeting about other features coming up on Monday, but I figured I'd create a PR for what I've done so far, in order to get some feedback on it thus far.


docker/flexnet/nets/net0/scripts/net-run.py
```diff
@@ -0,0 +1,31 @@
...
+
+    TX_AMOUNT = 1000
+    GAS_AMOUNT = 21000
+
+    for i in range(5):
+        print(f'Creating account {i}')
+        receiver = net.create_account()
+        sender = net.create_account()
+        print(f'Creating transaction {i}')
+        net.create_transaction(Transaction(sender, receiver, TX_AMOUNT, GAS_AMOUNT))
```
michael-yxchen, line 19: with `net0` that has no execution, this call prints error because `wait_for_transaction_receipt` is not supported. `get_transaction_count` seems to return something, but I'm not sure if the value is correct, because there's no triedb backend

we probably want different variant of `create_transaction` for consensus only vs full-node setup

docker/flexnet/testing-library/monad_flexnet/flexnet.py
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        signed_transfer = transaction.sender.sign_transaction(transfer_tx)
+        tx_hash = self.w3.eth.send_raw_transaction(signed_transfer.raw_transaction)
+        self.transactions.append({'hash': tx_hash.hex(), 'submitted': True})
+
+        try:
+            self.w3.eth.wait_for_transaction_receipt(tx_hash, timeout=5)
+        except Exception as e:
+            print('wait_for_transaction_receipt exception ', e, file=sys.stderr)
```
michael-yxchen: we should fail the test if an exception is caught

docker/flexnet/testing-library/monad_flexnet/flexnet.py
```diff
@@ -0,0 +1,158 @@
...
+
+        self._create_empty_triedb(self.root_dir / 'node'  / 'triedb' / 'test.db')
+
+        self.topology = Topology.from_json(self.topology_path)
+
+        # Build the containers that will be needed
+        docker.build(
+            './images/dev',
+            tags='monad-python-dev'
+        )
```
michael-yxchen: we should probably install the testing-library when building this image, so we don't need to `pip install` at every use

docker/flexnet/testing-library/monad_flexnet/flexnet.py
```diff
@@ -0,0 +1,158 @@
...
+class Flexnet:
+    def __init__(self, topology_path: str | os.PathLike):
+        self.topology_path = pathlib.Path(topology_path)
+        self.topology = None
+        self.transactions = []
+        self.root_dir = None
+    
+    def create_topology(self):
+        # Create output directory
+        self.run_name = f'{self.topology_path.stem}-{time.strftime("%Y%m%d_%H%M%S")}-{os.urandom(8).hex()}'
```
michael-yxchen: `Path.stem` returns `topology` for topology path `./nets/net0/topology.json`. We can probably assert that `topology.json` exists and pass the net directory name instead

docker/flexnet/testing-library/monad_flexnet/flexnet.py
```diff
@@ -0,0 +1,158 @@
...
+            '../..',
+            file='./images/image0/Dockerfile',
+            tags='image0:latest'
+        )
+        docker.build(
+            '../..',
+            file='./images/rpc/Dockerfile',
+            tags='flexnet-rpc:latest'
+        )
+
```
michael-yxchen: Do you plan to keep these hardcoded path? We shoud update the readme file if so

docker/flexnet/testing-library/monad_flexnet/topology/node.py
```diff
@@ -0,0 +1,45 @@
...
+        self.name = name
+        self.has_execution = has_execution
+        self.has_rpc = has_rpc
+        self.rpc_ip = None
+        self.upload_speed = upload_speed
+        self.download_speed = download_speed
+
+    def start(self, root_dir: str | os.PathLike, network_name: str):
+        docker.run(
+            image='image0:latest',
```
michael-yxchen, line 18: Now that docker compose is removed and we're building images manually, maybe we can use the `devnet` image everywhere. Can deal with this when you migrate `devnet-integration` to python

docker/flexnet/testing-library/monad_flexnet/topology/node.py
```diff
@@ -0,0 +1,45 @@
...
+    def __init__(self, name: str, has_execution: bool = False, has_rpc: bool = False, upload_speed: int = 100, download_speed: int = 100):
+        self.name = name
+        self.has_execution = has_execution
+        self.has_rpc = has_rpc
+        self.rpc_ip = None
+        self.upload_speed = upload_speed
+        self.download_speed = download_speed
+
+    def start(self, root_dir: str | os.PathLike, network_name: str):
+        docker.run(
```
michael-yxchen: It'd be nice if we can print the container names in the topology layout in case we need to connect to a certain container for debug. I'm thinking about something like
```
<net name>
    <region name>
        <node name>
            node: <container id>
            rpc: <container id>
        <node name>
        ...
```
feel free to use layout as you see fit

docker/flexnet/testing-library/monad_flexnet/topology/node.py
```diff
@@ -0,0 +1,45 @@
...
+                image='flexnet-rpc:latest',
+                remove=True,
+                name=f'{self.name}-rpc',
+                volumes=[(f'{root_dir}/{self.name}', '/monad')],
+                publish=[(8080,)],
+                detach=True,
+                command=['bash', '/monad/scripts/rpc.sh'],
+                networks=[network_name]
+            )
+            self.rpc_port = int(rpc_container.network_settings.ports['8080/tcp'][0]['HostPort'])
```
michael-yxchen: This is perfect. Resolves our port conflict issue. Unlikely to happen, but we should plan for the host running out of ports

docker/flexnet/testing-library/monad_flexnet/topology/node.py
```diff
@@ -0,0 +1,45 @@
...
+        self.has_rpc = has_rpc
+        self.rpc_ip = None
+        self.upload_speed = upload_speed
+        self.download_speed = download_speed
+
+    def start(self, root_dir: str | os.PathLike, network_name: str):
+        docker.run(
+            image='image0:latest',
+            remove=True,
+            name=self.name,
```
michael-yxchen: The name might conflict with other tests running on the same machine. Names are definitely helpful, but we should address (when stopping) containers with their unique id

docker/flexnet/testing-library/monad_flexnet/transaction.py
```diff
@@ -0,0 +1,11 @@
+import eth_account
+
+from dataclasses import dataclass
+
+@dataclass
+class Transaction:
+    sender: eth_account.Account
+    recipient: eth_account.Account
+    amount: int
+    gas: int = 10000
```
michael-yxchen: nit: we can set the default value to 21000, which is the base fee per EIP-1559

docker/flexnet/testing-library/monad_flexnet/flexnet.py
```diff
@@ -0,0 +1,158 @@
...
+    def get_account_balance(self, account: eth_account.Account) -> int:
+        return self.w3.eth.get_balance(account.address)
+
+    def expect_account_balance(self, account: eth_account.Account, balance: int):
+        print(self.get_account_balance(account))
+        assert(self.get_account_balance(account) == balance)
+
+    def check_block_ledger_equivalence(self) -> bool:
+        nodes = self.topology.get_all_nodes()
+        combinations = itertools.combinations(nodes, 2)
```
michael-yxchen: nit: we don't need to compare every combination. Comparing al lledgers with the first should do

docker/flexnet/testing-library/monad_flexnet/flexnet.py
eclark0426: As-is, yes. My intention is that we shouldn't need a directory of scripts for every single net though, since pretty much everything in them is duplicated between different nets. At that point I imagine having a single directory that has multiple topology files, at which point the filename here becomes more significant.

docker/flexnet/testing-library/monad_flexnet/topology/node.py
eclark0426: Yeah, I've been coming across things now that have `image0` vs `monad-node` (the `devnet` Dockerfile). Good to know that `monad-node` is the one I should move everything over to

docker/flexnet/testing-library/monad_flexnet/flexnet.py
eclark0426: I think I remember seeing somewhere that things should typically be run from `monad-bft/docker/flexnet`, so I wrote things based on that. If you'd rather not make that assumption, I can try to make this more generic, probably use keyword-only args in the flexnet constructor that would make the default correct if running from `monad-bft/docker/flexnet`

docker/flexnet/testing-library/monad_flexnet/topology/node.py
eclark0426: I don't think running out of host ports is going to be a huge issue; the bigger issue I think is going to be running out of subnets allocated for Docker bridge networks. I've already had that happen when I was missing some cleanup and not removing the networks. I want to say it was 16 networks when it ran out - which would give us upwards of 4000 nodes per network before we'll run out of host ports

eclark0426:
re your overall comments:

 > Improve on error capture and reporting.

I think this is something which should get better with future PRs. One of the things I've already done that's not in this PR is having output from all the containers get output to log files - this was done for some commands/containers, but not for everything. Even just that has already helped with debugging things like incorrect arguments.

> As we scale to hundreds of tests, we will probably want to run them in parallel.

It might be worth looking into external testing libraries to help facilitate this. It's the type of thing that has certainly been done before, and it'd be way easier to make use of the existing implementations than trying to roll our own testing framework.

> Handle potential resource/naming conflicts

This one doesn't seem like it'd be too hard to handle. We're already randomly generating parts of the output log path, we can probably just add the same random string to things like container names, so we have e.g. `node0-5ec05dc4509cdc1b` instead of just `node0`

docker/flexnet/testing-library/monad_flexnet/flexnet.py
eclark0426: This is one of the things we were talking about on Monday, and how we sometimes might want to stop the test, and other times we might want to ignore the error and continue. In one of my other comments, I mentioned also using an external testing framework - this is another thing that might already exist in external tools, is the option to just let the exception be thrown, and decide what to do with it on a per-test basis

monad-scripts/jenkins/flexnet-ci.sh
```diff
@@ -18,8 +18,10 @@ fi
 
 git submodule update --init --recursive
 
+pip install ./testing-library
```
michael-yxchen, line 30: Be mindful that this might break on our CI server because we don't have a python virtual env configured for the job. It's probably missing other packages (e.g. web3py and python-on-whales) as well

docker/flexnet/nets/consensus-nodes-nonce-test-1/scripts/net-run.sh
```diff
@@ -99,7 +99,7 @@ fi
 # Create node volume directory
 net_name=$(basename $(realpath "$net_dir"))
 rand_hex=$(od -vAn -N8 -tx1 /dev/urandom | tr -d " \n" | cut -c 1-16)
-vol_root="$output_dir/$net_name-$(date +%Y%m%d_%H%M%S)-$rand_hex"
+export vol_root="$output_dir/$net_name-$(date +%Y%m%d_%H%M%S)-$rand_hex"
```
michael-yxchen: super nit: cap env variable

docker/flexnet/nets/devnet-integration/e2e_tester/scripts/run.sh
```diff
@@ -2,4 +2,8 @@
 
 mkdir /monad/logs
 
+pushd /monad-lib
+pip install -e .
+popd
+
```
michael-yxchen: nit: we can install the testing library on the monad-python image so this isn't needed. Can fix this in the next PR

docker/flexnet/nets/devnet-integration/scripts/net-run.sh
michael-yxchen, line 1: I know this is annoying, but can you keep the devnet-integration changes to another PR?

monad-scripts/jenkins/flexnet-ci.sh
eclark0426: Is there a way to check I haven't broken anything with this before merging or anything?

PR #862, opened 2024-07-07
Add smart contract revert message in rpc response

eerkaijun:
If a smart contract transaction revert, we want to be able to pass the revert message in `eth_call` and `eth_estimateGas`. Tested and made sure it returns with the same format as geth. 

Closes https://github.com/monad-crypto/monad-internal/issues/337

### DIFF
--- a/monad-cxx/src/lib.rs
+++ b/monad-cxx/src/lib.rs
@@ -23,13 +23,18 @@ pub const EVMC_SUCCESS: i32 = 0;
 
 pub enum CallResult {
     Success(SuccessCallResult),
-    Failure(String),
+    Failure(FailureCallResult),
 }
 
 pub struct SuccessCallResult {
     pub gas_used: u64,
     pub gas_refund: u64,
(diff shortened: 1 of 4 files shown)


monad-cxx/src/lib.rs
```diff
@@ -91,10 +96,54 @@ pub fn eth_call(
...
+        }
     }
 }
 
+pub fn decode_revert_message(output_data: &[u8]) -> String {
+    // if there is an error message to be decoded, output_data will be the following form:
+    // 4 bytes function signature
+    // 32 bytes data offset
+    // 32 bytes error message length (let's call it x)
+    // x bytes error message (padded to multiple of 32 bytes)
```
ariqchowdhury, line 128: is this format specified somewhere, like an eip? should link to it if so

eerkaijun: Don't think it's an eip but have now added the documentation link in the comment

PR #866, opened 2024-07-10
Suggest gas tips using a gas oracle

philipglazman:
Depends on `monad-stream` which feeds blocks into the gas oracle to suggest a gas tip.


philipglazman:
replaced with https://github.com/monad-crypto/monad-bft/pull/1253

PR #868, opened 2024-07-11
Update blocktree parent coherency

omegablitz:
Whenever a new block is added to the blocktree, if there's a path to root, incoherent parents should be rechecked for coherency. This re-checking is necessary in the case that execution is delayed.


omegablitz:
`update_coherency` will check all descendents of the block you pass it, so we're not skipping checking the current block

PR #872, opened 2024-07-11
Flexnet Execution Testing

eclark0426:
Ports `full-nodes` and `devnet-integration` to the new testing framework.


docker/flexnet/testing-library/monad_flexnet/rpcconnection.py
```diff
@@ -0,0 +1,68 @@
...
+            'maxFeePerGas': transaction.maxFeePerGas,
+            'maxPriorityFeePerGas': 0,
+            'chainId': 1
+        }
+
+        signed_transfer = transaction.sender.sign_transaction(transfer_tx)
+        tx_hash = self.w3.eth.send_raw_transaction(signed_transfer.raw_transaction)
+        self.transactions.append({'hash': tx_hash.hex(), 'submitted': True})
+        return tx_hash
+
```
michael-yxchen, line 50: maybe it's already implemented in the next PRs, I think we should decouple creating transaction from sending transactions. Creating transaction can fit into a helper function, or be completely user controlled. RpcConnection should have the send_transaction method

docker/flexnet/testing-library/monad_flexnet/rpcconnection.py
```diff
@@ -0,0 +1,68 @@
...
+            'nonce': sender_tx_count,
+            'gas': transaction.gas,
+            'maxFeePerGas': transaction.maxFeePerGas,
+            'maxPriorityFeePerGas': 0,
+            'chainId': 1
+        }
+
+        signed_transfer = transaction.sender.sign_transaction(transfer_tx)
+        tx_hash = self.w3.eth.send_raw_transaction(signed_transfer.raw_transaction)
+        self.transactions.append({'hash': tx_hash.hex(), 'submitted': True})
```
michael-yxchen, line 48: not sure if "submitted" is still meaningful here

docker/flexnet/nets/devnet-integration/scripts/net-run.sh
michael-yxchen, line 1: Can you cherry-pick the "re-adding devnet integration" commit? https://github.com/monad-crypto/monad-bft/commit/13af1bc163732f7d232624dee827fa9ae62807d2#diff-c78b1a4f30c0b96f1c18e7761ce3668eba511da0df0678fa11d947d38db3bd29

We need to make sure that dev/rpc Dockerfiles are always tested, and keep the utility to easily spin up a single node

docker/flexnet/testing-library/monad_flexnet/rpcconnection.py
```diff
@@ -0,0 +1,68 @@
...
+        adapter = requests.adapters.HTTPAdapter(max_retries=urllib3.util.retry.Retry(connect=3, backoff_factor=1))
+        session = requests.Session()
+        session.mount('http://', adapter)
+        session.mount('https://', adapter)
+        self.w3 = Web3(Web3.HTTPProvider(rpc_server, session=session))
+
+    def create_account(self, key: str | None = None) -> eth_account.Account:
+        if key is not None:
+            return eth_account.Account.from_key(key)
+        return eth_account.Account.create()
```
michael-yxchen, line 30: Same thing here, this should be an utility function separate from the RPC connection

docker/flexnet/nets/full-nodes/scripts/net-run.py
```diff
@@ -0,0 +1,54 @@
...
+
+    with net_config.start_topology(gen_config=True) as net:
+        sender_account = None
+        clients = {}
+
+        for rpc_node in rpc_nodes:
+            client = net.connect(rpc_node)
+            clients[rpc_node] = client
+            if sender_account is None:
+                sender_account = client.create_account(sender_privkey)
```
michael-yxchen, line 30: this should be a one time setup

docker/flexnet/nets/full-nodes/scripts/net-run.py
```diff
@@ -0,0 +1,54 @@
...
+
+        for i in range(5):
+            # Pick different RPC to service the transaction
+            send_client: RpcConnection = clients[rpc_nodes[i % len(rpc_nodes)]]
+            receiver_account = send_client.create_account()
+            print(f'receiver account address {receiver_account.address}')
+            tx_hash = send_client.create_transaction(Transaction(sender_account, receiver_account, TRANSACTION_AMOUNT, GAS_AMOUNT))
+
+            time.sleep(5)
```
michael-yxchen: what is this sleep for?

docker/flexnet/nets/full-nodes/scripts/net-run.py
```diff
@@ -0,0 +1,54 @@
...
+            print(f'sender balance {client.get_account_balance(sender_account)}')
+            print(f'receiver balance {client.get_account_balance(receiver_account)}')
+
+            for other_client_name, other_client in clients.items():
```
michael-yxchen: nit: "other_client_name" & "other_client" may not be the most accurate as we're checking every client

docker/devnet/Dockerfile
michael-yxchen, line 1: We should avoid changing this file as it's used in deployment, unless it's breaking tests and deployment

docker/flexnet/testing-library/monad_flexnet/flexnet.py
```diff
@@ -27,15 +25,22 @@ def __init__(self, topology_path: str | os.PathLike):
...
         self.run_id = os.urandom(8).hex()
         self.run_name = f'{self.topology_path.stem}-{time.strftime("%Y%m%d_%H%M%S")}-{self.run_id}'
         self.root_dir = pathlib.Path(f'{os.getcwd()}/logs/{self.run_name}')
         shutil.copytree(self.topology_path.parent, self.root_dir)
 
-        self._create_empty_triedb(self.root_dir / 'node'  / 'triedb' / 'test.db')
-
         self.topology = Topology.from_json(self.topology_path)
 
         # Build the containers that will be needed
```
michael-yxchen, line 46: nit: I think this step can be parallelized

docker/flexnet/nets/full-nodes/scripts/net-run.py
eclark0426: I'm not sure if I understand what you're getting at, this is already only being done once

docker/devnet/Dockerfile
eclark0426: I think there had been an issue when trying to build the container - it was related to Docker caching the `apt-get update` lines, so when it tries to do the `apt install`, it was getting 404 errors because the server list would've been updated from `apt-get update` but it was cached.

docker/flexnet/nets/full-nodes/scripts/net-run.py
eclark0426: It was just to wait for the transaction to go through. I think I was having issues with the `wait_for_transaction_receipt` calls - will change it to a `create_and_wait_for_transaction`

docker/flexnet/testing-library/monad_flexnet/flexnet.py
eclark0426: Probably can - are you okay with it if I update that in a later PR so that I won't have to fix a ton of conflicts from different things touching this code?

docker/flexnet/testing-library/monad_flexnet/rpcconnection.py
eclark0426: Probably not all that much, although I think there was still a check somewhere that verifies things have `submitted == True`

docker/flexnet/testing-library/monad_flexnet/rpcconnection.py
eclark0426: I wonder a bit how much the `RpcConnection` class is actually needed. Largely it's just a wrapper around web3 calls, so maybe it'd make more sense to just give the user a web3 client, so they can use any functions that are available in w3, instead of being limited to the wrappers we provide. That'd then also get rid of the combined creation and sending of transactions, as well as removing account creation from rpcconnection

michael-yxchen: i agree. I think the connection itself probably doesn't need a class. We can provide a web3 helper library if the tests repeat code

docker/flexnet/testing-library/monad_flexnet/rpcconnection.py
michael-yxchen: yeah one test was verifying `submitted == True` because it's reading from a transaction data file, which has `submitted: False`, and only set it to true after submitting it to RPC. Not sure if that's still what the test is doing now

docker/flexnet/nets/full-nodes/scripts/net-run.py
michael-yxchen: calls to `create_account` should be moved out of the for loop

docker/devnet/Dockerfile
michael-yxchen: yeah I remember seeing that as well. I had to ask Mike to prune docker cache on peach22

docker/flexnet/testing-library/monad_flexnet/flexnet.py
michael-yxchen: yes ofc

docker/devnet/Dockerfile
```diff
@@ -4,7 +4,7 @@ WORKDIR /usr/src/monad-bft
 RUN apt update
 RUN apt upgrade -y
 RUN apt-get update
-RUN apt install -y clang curl make ca-certificates pkg-config gnupg software-properties-common wget
+RUN apt install -y clang curl iproute2 make ca-certificates pkg-config gnupg software-properties-common wget
```
michael-yxchen: is there a step in the build that depends on this package, or is it for things like `ping` in testing? we should remove it if it's the latter case

eclark0426: I think this change must have been a remnant from before I switched to using `image0` for everything. It was needed for the `ip` commands in the traffic control scripts, but I'll remove it from the devnet container

docker/flexnet/nets/full-nodes/scripts/net-run.py
eclark0426: With using a separate `RpcConnection` class, you need a client to exist before you can create an account, so that's why I have it in the loop with a check against none

michael-yxchen: yeah I understand why it's needed now with `create_account` under `RpcConnection`. Just another reason to move `create_account` out. You don't need to make the change in this PR, considering how much is based on top

eclark0426: Was in the process of working on that, but if you're okay with it waiting until a different PR, then I'll move the changes over to later. Definitely will make the other PRs easier on my end if I hold off on those changes to later

michael-yxchen: yeah we can put it in a later PR, probably after the move to monad-integrations

docker/flexnet/testing-library/monad_flexnet/rpcconnection.py
eclark0426: The ledger still checks that submitted is true. At least at this point in the PRs, there's still a `net1` test that starts out with submitted=False

docker/flexnet/nets/full-nodes/scripts/net-run.py
eclark0426: Sounds good. In that case I _think_ your comments have all been addressed, or are planned for future changes

michael-yxchen: cool i'll give it another look

monad-scripts/jenkins/flexnet-ci.sh
```diff
@@ -33,8 +33,8 @@ pip install ./testing-library
 rm -rf logs && mkdir -p logs
 python nets/net0/scripts/net-run.py
 nets/net1/scripts/net-run.sh --output-dir logs --net-dir nets/net1/ --flexnet-root . --monad-bft-root ../.. 
-nets/devnet-integration/scripts/net-run.sh test --output-dir logs --net-dir nets/devnet-integration/ --image-root images --monad-bft-root ../..
-nets/full-nodes/scripts/net-run.sh test --output-dir logs --net-dir nets/full-nodes/ --flexnet-root . --monad-bft-root ../..
+python nets/devnet-integration/scripts/net-run.py devnet test
```
michael-yxchen, line 37: let's keep the tests with the old net-run.sh script, and add in the script with python

docker/flexnet/nets/devnet-integration/scripts/net-run.py
```diff
@@ -0,0 +1,37 @@
...
+        TRANSACTION_COUNT = 5
+
+        sender_account = client.create_account(sender_privkey)
+        print(f'sender account address {sender_account.address}')
+
+        for i in range(TRANSACTION_COUNT):
+            receiver_account = client.create_account()
+            client.create_transaction(Transaction(sender_account, receiver_account, TRANSACTION_AMOUNT, GAS_AMOUNT))
+
+            time.sleep(5)
```
michael-yxchen: Can we add a FIXME here so we don't forget?

docker/flexnet/testing-library/monad_flexnet/flexnet.py
```diff
@@ -98,45 +139,8 @@ def stop(self):
...
-    def get_account_balance(self, account: eth_account.Account) -> int:
-        return self.w3.eth.get_balance(account.address)
-
-    def expect_account_balance(self, account: eth_account.Account, balance: int):
-        print(self.get_account_balance(account))
-        assert(self.get_account_balance(account) == balance)
+        
+        return RpcConnection(f'http://localhost:{rpc_port}', not rpc_node.has_execution)
 
     def check_block_ledger_equivalence(self) -> bool:
```
michael-yxchen, line 145: Something I didn't catch in the previous PR. The old test (net0) asserts two things
- The shortest ledger is longer than a number e.g. 300, after running for ~30 seconds
- All ledgers agree up to the shortest ledger. We only check up to shortest because the protocol only guarantees eventual consistency. The last few (2) blocks in the longest ledger can be un-committed on other ledgers

The current verification misses length assertion, and isn't comparing ledgers correctly, e.g. I got this error when running net0
```
  File "/home/mchen/github/monad-bft/docker/flexnet/nets/net0/scripts/net-run.py", line 26, in <module>
    main()
  File "/home/mchen/github/monad-bft/docker/flexnet/nets/net0/scripts/net-run.py", line 22, in main
    assert net_config.check_block_ledger_equivalence()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mchen/github/monad-bft/docker/flexnet/.venv/lib/python3.11/site-packages/monad_flexnet/flexnet.py", line 150, in check_block_ledger_equivalence
    if ledger1 != ledger2:
       ^^^^^^^^^^^^^^^^^^
  File "/home/mchen/github/monad-bft/docker/flexnet/.venv/lib/python3.11/site-packages/monad_flexnet/ledger.py", line 49, in __eq__
    if not filecmp.cmp(f, pathlib.Path(other.blockpath) / f.name):
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mchen/.pyenv/versions/3.11.9/lib/python3.11/filecmp.py", line 54, in cmp
    s2 = _sig(os.stat(f2))
              ^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/home/mchen/github/monad-bft/docker/flexnet/logs/topology-20240719_134729-ec4e38b35895ed8c/node1/ledger/140'
```

We should fix that before migrating more tests. It'd be easier to review if it's another PR based on master, and we can merge that first. Or if you want to build it into this PR, that's fine too.

docker/flexnet/testing-library/monad_flexnet/flexnet.py
michael-yxchen: let's add a TODO: here so we don't forget

docker/flexnet/testing-library/monad_flexnet/flexnet.py
eclark0426: That error is one I know about, and is fixed later on - it has to do with containers not being stopped before checking the ledger on disk, so things haven't always been fully flushed to disk when trying to read. The ledger length one I don't think I realized

docker/flexnet/testing-library/monad_flexnet/flexnet.py
eclark0426: I don't tend to make a habit of checking todo's, because half the time they never get done, so here's this instead: https://github.com/monad-crypto/monad-internal/issues/376

docker/flexnet/testing-library/monad_flexnet/flexnet.py
michael-yxchen: fyi, I think the ledger length one is uncommon. It happens if one/a few node processed the proposal which commits a block, while other nodes were killed before processing the proposal

docker/flexnet/testing-library/monad_flexnet/flexnet.py
michael-yxchen: Works too, thanks

docker/flexnet/nets/devnet-integration/scripts/net-run.py
eclark0426: I assume this is meant to be a similar comment to what you asked about with sleep on devnet integration? If so that's a simple enough change that I'll make now

eclark0426: Just went ahead and assumed you wanted this to not have a sleep, removed in last push

docker/flexnet/testing-library/monad_flexnet/flexnet.py
eclark0426: Per offline conversation - the error has been fixed in my last push. Ledger length check can be added later, since this test isn't being run for a specific amount of time

docker/flexnet/testing-library/monad_flexnet/rpcconnection.py
eclark0426: This conversation was related to https://github.com/monad-crypto/monad-bft/pull/872#discussion_r1684510106 and https://github.com/monad-crypto/monad-bft/pull/872#discussion_r1683529356 - we decided to remove the RpcConnection class entirely since it's mostly just wrappers around web3 functions. I've already completed this in one of the later PRs

michael-yxchen:
Thanks, reminder to manual rebase before merging

PR #873, opened 2024-07-11
Adds Byzantine nodes to Flexnet testing

eclark0426:
Essentially, ports `net1` to the testing library. This is based on my two previous PRs, so those should probably get merged first.


docker/flexnet/testing-library/monad_flexnet/topology/node.py
```diff
@@ -122,6 +140,7 @@ def stop(self):
             Node._try_stop(self.rpc_container)
         if self.has_execution:
             Node._try_stop(self.execution_container)
+        self.node_root = None
```
rgarc: Why are you setting `self.node_root = None` here?

docker/flexnet/testing-library/monad_flexnet/flexnet.py
```diff
@@ -35,6 +36,9 @@ def start_topology(self, gen_config: bool = True):
             self.stop()
 
     def create_topology(self, gen_config: bool = True):
```
rgarc: Is there a reason for this `create_topology` function to exist? Why can't it all be moved into the `Flexnet.__init__` function above? This would also obviate the need for the `self.created` field.

docker/flexnet/testing-library/monad_flexnet/topology/node.py
eclark0426: I think the goal with that was so the same node could be started again later, but as part of a different run so it's using a different output directory

docker/flexnet/testing-library/monad_flexnet/flexnet.py
eclark0426: Similar answer here, it would let you reuse the same topology for different tests while having separate output directories. More importantly on this one though, this lets you make additional configuration changes outside of the topology file itself - e.g. the `net1` test which runs with `test-mode --byzantine-execution`

docker/flexnet/testing-library/monad_flexnet/topology/node.py
rgarc: Why would we want to re-use the same node for a different run? Why not just create a new one? Does a node own a resource that was expensive to create and is that what we want to re-use?

The reason why I'm asking is because when I was porting a test that needed to access a file in each node's directory, I expected to be able to do something like this
```python
from monad_flexnet import Flexnet

import time

def main():
    net_config = Flexnet('./nets/update-validators-new/topology.json')
    net_config.set_test_mode(True)

    with net_config.start_topology() as net:
        # do some stuff with net here
        time.sleep(10)

    for node in net_config.topology.get_all_nodes():
        with open(node.node_root / 'wal.json', 'r') as wal:
            # do something with `wal` here 
            pass


if __name__ == '__main__':
    main()
```
But I was surprised to see that `node.node_root` was `None`. It wasn't very intuitive. I think in general, we should try to avoid re-using the same object because if we add new fields for example, we need to remember which fields need to be set to `None` inside this `stop` function.

eclark0426: The node itself probably can be recreated without much issue, and not need to reset `node_root` to `None`. Though I will say that my initial thought is that the `# do something with wal here` interactions might be better fit inside the library itself, since that's probably something that can be reused with other tests

eclark0426: Without knowing the specifics of the test and how it differs from what I've done, my instinct would be that directly interacting with output files shouldn't be necessary from tests, and can be abstracted a bit in the library

rgarc: Oh I think I see now. The intended way of going about this is to add a getter for the `wal` at the `Flexnet` class level similar to how `Flexnet.get_block_ledger` works. 

For more context, each node serializes events in binary format to disk in a file called `wal`. So my test parses that `wal` file, computes statistics on it and writes it as JSON to `wal.json` in each node directory. Then the test will parse that JSON and assert some conditions depending on the needs of the test.

docker/devnet/Dockerfile
eclark0426, line 1: I coulda sworn these changes were in the previous PR - these are what fixed the 404 because of a cached `apt-get update`

docker/flexnet/nets/net1/scripts/net-run.py
eclark0426, line 1: I was mistaken on the last PR when I said net1 checks ledger length. The test I was thinking of is actually net2, which just runs a transaction generator for 10 minutes

docker/flexnet/testing-library/monad_flexnet/topology/node.py
eclark0426: Removed the `self.node_root = None` from `stop()` in the commit I just added

docker/devnet/Dockerfile
michael-yxchen, line 1: Pls revert: https://github.com/monad-crypto/monad-bft/pull/872#discussion_r1684516154

docker/flexnet/testing-library/monad_flexnet/flexnet.py
```diff
@@ -35,6 +36,9 @@ def start_topology(self, gen_config: bool = True):
...
+        if self.created:
+            return
+
```
michael-yxchen: I think this `created` flag should be avoided. It's hard to reason about which fields are mutable and which context (topology v. instantiated run) do they belong to. In the current layout, some are tied to the topology configuration while some are tied to a run

I think of a topology as a network layout and associated docker images, and a test run as the instantiated docker network/containers with artifact directories. Just one way to split `Flexnet` and avoid this created flag

docker/devnet/Dockerfile
eclark0426: Pruning the cache seems more like a workaround than a fix. If I revert that commit, the same 404 error is bound to pop up again at some point in the future when the available package list changes

docker/flexnet/testing-library/monad_flexnet/flexnet.py
eclark0426: Pushed a change to this

docker/devnet/Dockerfile
eclark0426: If you'd like me to pull this commit into a separate PR or something I can, but I think this is a worthwhile change to include, even if it is changing the Dockerfiles that are used in deployments. If we don't change it now, it's just going to become a problem again at some point down the line

michael-yxchen: Can you look around to see if this is the common practice for Dockerfiles? It feels counter-intuitive that you need to run `apt update` before every install, but maybe that's common to combat docker cache

Related:
https://github.com/moby/moby/issues/3313 but i didn't read closely

eclark0426: It is, or at least it's one of the common solutions to things like this - that issue is where I got this pattern from

docker/flexnet/testing-library/monad_flexnet/flexnet.py
```diff
@@ -20,29 +20,23 @@
...
         # Create output directory
-        self.run_id = os.urandom(8).hex()
-        self.run_name = f'{self.topology_path.stem}-{time.strftime("%Y%m%d_%H%M%S")}-{self.run_id}'
-        self.root_dir = pathlib.Path(f'{os.getcwd()}/logs/{self.run_name}')
         shutil.copytree(self.topology_path.parent, self.root_dir)
 
-        self.topology = Topology.from_json(self.topology_path)
-
         # Build the containers that will be needed
         docker.build(
```
michael-yxchen, line 41: nit TODO: we should build a subset of these images conditionally on the topology. Maybe add to https://github.com/monad-crypto/monad-internal/issues/376

docker/devnet/Dockerfile
eclark0426: Specifically this comment from the issue was what made it seem like this was the best solution: https://github.com/moby/moby/issues/3313#issuecomment-32599795

docker/flexnet/testing-library/monad_flexnet/flexnet.py
michael-yxchen: Topology reuse is a TODO then. Is that correct?

docker/devnet/Dockerfile
michael-yxchen: Are there other examples/guideline supporting that pattern?

eclark0426: Plenty, put in a Google search for `"run apt update && apt install" site:github.com`. Several from just the first page:
- https://github.com/fluent/fluent-bit/blob/574a69af744535b6e016965f02eef9f739a5df1e/dockerfiles/Dockerfile#L37-L57
- https://github.com/Kong/kong-pongo/blob/ec497140fe30d3edcd3d8713d98047980b2bda72/assets/Dockerfile#L49-L50
- https://github.com/rameau-fr/MC-Calib/blob/ef63759523c81c10b556d5c570b18405a7ce4fde/Dockerfile#L4-L8
- https://github.com/jselbie/stunserver/blob/2049da33990b66d71a0f42f32d70b75fe4911374/Dockerfile#L4-L11
- https://github.com/google/genc/blob/26ffa1568f5efc49e6c5132d6a19807f5cb76cc7/Dockerfile#L18-L22

docker/flexnet/testing-library/monad_flexnet/flexnet.py
eclark0426: Yeah, you'd have to create a new topology instance if you want to run an additional test

docker/flexnet/testing-library/monad_flexnet/flexnet.py
eclark0426: Added to that issue (currently working on that one)

docker/devnet/Dockerfile
michael-yxchen: https://docs.docker.com/build/building/best-practices/#apt-get
> Always combine RUN apt-get update with apt-get install in the same RUN statement

michael-yxchen, line 1: nit: I don't think it matters for correctness, but can you change it so we don't have `apt install` after `apt-get update` and vice versa?

docker/flexnet/nets/net1/scripts/net-run.py
michael-yxchen: TODO: add those length assertions back to the tests

eclark0426: https://github.com/monad-crypto/monad-internal/issues/382
Issue to add those back, and make it more generically applicable than just a simple check, more something that can be checked by default when a net is stopped

docker/devnet/Dockerfile
```diff
@@ -22,7 +18,7 @@ RUN apt-get install -y \
   libboost-regex1.83.0 \
   libboost-stacktrace1.83.0
 
-RUN apt install -y \
+RUN apt-get update && apt-gwt install -y \
```
michael-yxchen: typo

docker/devnet/Dockerfile
```diff
@@ -1,18 +1,14 @@
 FROM ubuntu:24.04 AS base
 
 WORKDIR /usr/src/monad-bft
-RUN apt update
-RUN apt upgrade -y
-RUN apt-get update
-RUN apt install -y clang curl make ca-certificates pkg-config gnupg software-properties-common wget
+RUN apt-get update && apt upgrade -y
```
michael-yxchen: apt-get upgrade

docker/devnet/Dockerfile
```diff
@@ -48,13 +44,13 @@ RUN echo 'deb [signed-by=/usr/share/keyrings/kitware-archive-keyring.gpg] https:
   | tee /etc/apt/sources.list.d/kitware.list >/dev/null
 RUN apt-get update
```
michael-yxchen: remove

PR #874, opened 2024-07-11
Remove empty blocks from Eth ledger

michael-yxchen:
A node proposes an empty block if don't have the execution state root, delay blocks ago. Empty blocks now use the same sequence number and state root as their parent. They are committed to consensus ledger, but not to eth ledger. This provides a nice property that if we observe QC(N), block N-delay is guaranteed to be committed. We can then state sync to N-delay.

Some implications of the change
- Epoch is now measured in number of non-empty blocks
- We wouldn't issue reward for empty blocks, as they are not committed to execution. Otherwise nodes can always produce empty blocks, get rewards, and never move in to a new epoch

Closes https://github.com/monad-crypto/monad-internal/issues/302

### DIFF
--- a/monad-eth-block-policy/src/lib.rs
+++ b/monad-eth-block-policy/src/lib.rs
@@ -136,8 +136,8 @@ impl<SCT: SignatureCollection> BlockType<SCT> for EthValidatedBlock<SCT> {
         self.validated_txns.len()
     }
 
-    fn is_txn_list_empty(&self) -> bool {
-        self.validated_txns.is_empty()
+    fn is_empty_block(&self) -> bool {
+        self.block.is_empty_block()
     }
 }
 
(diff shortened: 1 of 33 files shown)


ariqchowdhury:
while we have this PR that's changing things, maybe we should come up with a different name for empty blocks since its confusing to distinguish between a block with no payload and the special empty block. Maybe something along the lines of KeepAliveBlock? LivelinessBlock? SubsistenceBlock?

ariqchowdhury:
One interaction we need to consider now -- how does rpc deal with the empty blocks that are in blockdb? If we are retrieving blocks by block_number, there will be multiple blocks with the same block_num since they have the same seq_num. I'm pretty sure there is a table in blockdb keyed by block_num (which is equal to seq_num). So right now this code is overwriting and will end up with the latest empty block and the actual executed block would be lost. 
Likely most users would only want the executed blocks when looking up by block num in rpc. But we probably want to keep empty blocks around in blockdb and still make them retrievable -- maybe there's a use case for finding empty blocks.

omegablitz:
Noting that this might change forkpoint validation
Specifically, anywhere we check for is_epoch_end, do we also need to check that the block isn't an empty-block?

michael-yxchen:
https://github.com/monad-crypto/monad-bft/pull/874/files#diff-7f422a2cd73b3262c684d2aff2aaa417c777cbaf8db4cfadd34693088862b92bR644-R648
this check was added so we only trigger epoch scheduling on non-empty block

monad-eth-block-policy/src/lib.rs
```diff
@@ -532,6 +532,11 @@ where
         state_backend: &SBT,
     ) -> Result<(), BlockPolicyError> {
         trace!(?block, "check_coherency");
+        // if first_block.is_empty_block() {
+        //     assert_eq!(first_block.get_seq_num(), self.last_commit);
+        // } else {
+        //     assert_eq!(first_block.get_seq_num(), self.last_commit + SeqNum(1));
+        // }
```
ariqchowdhury: nit: delete

PR #880, opened 2024-07-12
Only include root_qc_N-delay root in forkpoint

omegablitz:
The ledger is required to have blocks (N - delay, N], so those roots are all fetchable. The root_qc_N-delay root is only necessary for statesyncing and validating the initial execution state.

### DIFF
--- a/monad-mock-swarm/src/node.rs
+++ b/monad-mock-swarm/src/node.rs
@@ -7,7 +7,6 @@ use std::{
 use itertools::Itertools;
 use monad_async_state_verify::BoxedAsyncStateVerifyProcess;
 use monad_consensus_types::{
-    block::BlockType,
     payload::StateRootValidator,
     quorum_certificate::QuorumCertificate,
     signature_collection::SignatureCollection,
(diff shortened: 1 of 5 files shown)


monad-mock-swarm/src/node.rs
```diff
@@ -343,16 +343,8 @@ impl<S: SwarmRelation> Node<S> {
...
-        let mut state_roots = Vec::new();
-
-        for i in ((root_qc.get_seq_num().0 - delay.0 + 1)..=(root_qc.get_seq_num().0 + 1)).rev() {
-            let b = &blocks[i as usize];
-            assert_eq!(b.payload.seq_num, SeqNum(i));
-            let state_root = state_root_executor.compute_state_root_hash(&b.get_seq_num());
-            state_roots.push(state_root);
-        }
+        let state_root =
+            state_root_executor.compute_state_root_hash(&(root_qc.get_seq_num() - delay));
```
michael-yxchen: nit: should we take `max(root_qc.get_seq_num(), delay) - delay` to avoid underflow?

PR #881, opened 2024-07-15
Change EthBlockPolicy to cache account nonce

michael-yxchen:
EthBlockPolicy was caching nonce of last transaction. It was inconsistent with eth/triedb account nonces

### DIFF
--- a/monad-eth-block-policy/src/lib.rs
+++ b/monad-eth-block-policy/src/lib.rs
@@ -10,6 +10,10 @@ use monad_eth_tx::{EthTransaction, EthTxHash};
 use monad_eth_types::{EthAddress, Nonce};
 use monad_types::{BlockId, Epoch, NodeId, Round, SeqNum};
 
+pub trait AccountNonceRetrievable {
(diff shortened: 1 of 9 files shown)


monad-eth-block-policy/src/lib.rs
```diff
@@ -105,29 +109,64 @@ impl<SCT: SignatureCollection> BlockType<SCT> for EthValidatedBlock<SCT> {
...
+            let acc_nonce = txn_nonce + 1;
+            accnt_nonces.insert(address, acc_nonce);
+        }
+        accnt_nonces
+    }
+}
+
+impl<SCT: SignatureCollection> AccountNonceRetrievable for Vec<&EthValidatedBlock<SCT>> {
+    fn get_account_nonces(&self) -> BTreeMap<EthAddress, Nonce> {
+        let mut accnt_nonces = BTreeMap::new();
```
ariqchowdhury: nit: i would prefer using the full name `account_nonces` or if shortening go further and make it meaningfully shorter.

PR #885, opened 2024-07-16
Save and restart from checkpoint instead of WAL

omegablitz:
This is tested to save and load from the generated forkpoint successfully. However, block production resumption won't work until state-sync is merged, because state roots, eth-block-policy, blocktree, etc. aren't seeded properly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@
 /docker/flexnet/logs
 /docker/devnet/monad/blockdb
 /docker/devnet/monad/ledger
-/docker/devnet/monad/wal
+/docker/devnet/monad/wal_*
 /docker/devnet/monad/triedb
 /docker/devnet/monad/config/forkpoint.toml
 
--- a/docker/devnet/clean.sh
+++ b/docker/devnet/clean.sh
@@ -29,9 +29,7 @@ if [ -S "$controlpanel_sock" ]; then
     rm "$controlpanel_sock"
 fi
 
-if [ -f "$wal" ]; then
-    rm "$wal"
-fi
+rm ${wal}_* || true
 
 if [ -f "$forkpoint" ]; then
     rm "$forkpoint"
--- a/monad-node/src/state.rs
+++ b/monad-node/src/state.rs
@@ -78,6 +78,22 @@ impl NodeState {
         let forkpoint_config: ForkpointConfig =
             toml::from_str(&std::fs::read_to_string(&cli.forkpoint_config)?)?;
 
+        let wal_path = cli.wal_path.with_file_name(format!(
+            "{}_{}_{}_{}",
+            cli.wal_path
+                .file_name()
+                .expect("no wal file name")
+                .to_owned()
+                .into_string()
+                .expect("invalid wal path"),
+            forkpoint_config.root.round.0,
+            forkpoint_config.root.block_id.0,
+            std::time::UNIX_EPOCH
+                .elapsed()
+                .expect("time went backwards")
+                .as_millis()
+        ));
+
         let run_mode = cli.run_mode.unwrap_or_default();
 
         Ok(Self {
@@ -89,7 +105,7 @@ impl NodeState {
             bls12_381_identity: bls_key,
 
             forkpoint_path: cli.forkpoint_config,
-            wal_path: cli.wal_path,
+            wal_path,
             execution_ledger_path: cli.execution_ledger_path,
             blockdb_path: cli.blockdb_path,
             triedb_path: cli.triedb_path,


monad-consensus-state/src/lib.rs
```diff
@@ -610,9 +613,69 @@ where
...
+                epoch,
+                round,
+                validators: ValidatorSetData(validators),
+            }
+        };
+
+        let base_epoch = self
+            .epoch_manager
+            .get_epoch(self.consensus.pending_block_tree.root().round + Round(1))
+            .expect("epoch for next round should exist");
```
michael-yxchen: Could be wrong. Should this be `epoch_manager.get_epoch(high_qc.round)`

monad-state/src/lib.rs
```diff
@@ -191,14 +188,11 @@ impl<SCT: SignatureCollection> Forkpoint<SCT> {
...
     /// 3. QC epoch-round and validator set epoch-round are set correctly
     /// 4. Has n+1 validator set locked
-    /// 5. If the last ledger block is on epoch boundary, n+2 validator set is
-    ///    locked
-    /// 6. Every validator set except the last one is scheduled to start on a
-    ///    round
+    /// 5. First validator set is scheduled to start on a round
     pub fn validate<VTF>(
         &self,
         state_root_delay: SeqNum,
```
michael-yxchen, line 243: nit: remove `state_root_delay`?

monad-state/src/lib.rs
```diff
@@ -208,32 +202,36 @@ impl<SCT: SignatureCollection> Forkpoint<SCT> {
...
-        // 1. Validator set exists for qc.epoch
-        let qc_vset_data = self
-            .get_validator_set(self.high_qc.get_epoch())
+        assert_eq!(self.validator_sets.len(), 2);
+        // 1. Validator set exists for N+1 where N is last committed block
+        let base_epoch = if self.root.seq_num.is_epoch_end(val_set_update_interval) {
+            self.root.epoch + Epoch(1)
+        } else {
+            self.root.epoch
+        };
```
michael-yxchen: Related here. I'm wondering if we should only include high_qc.epoch and high_qc.epoch + 1 validator set, and trust the root

monad-state/src/lib.rs
```diff
@@ -251,15 +249,11 @@ impl<SCT: SignatureCollection> Forkpoint<SCT> {
         VTF: ValidatorSetTypeFactory<NodeIdPubKey = SCT::NodeIdPubKey>,
     {
         assert_eq!(qc_vset_data.epoch, self.high_qc.get_epoch());
-        // 6.
-        if qc_vset_data.round.is_none() {
-            return Err(ForkpointValidationError::InvalidValidatorSet);
-        }
 
         // 3. Check QC carries consistent round numbers
         if self.high_qc.get_round() < qc_vset_data.round.expect("asserted before")
```
michael-yxchen: I think it's no longer asserted? not sure

monad-updaters/src/parent.rs
```diff
@@ -29,22 +29,22 @@ pub struct ParentExecutor<R, T, L, EL, C, S, IPC, CP, LO> {
...
     EL: Executor<Command = ExecutionLedgerCommand<SCT>>,
     SE: Executor<Command = StateRootHashCommand>,
     CPE: Executor<Command = ControlPanelCommand<SCT>>,
     LOE: Executor<Command = LoopbackCommand<E>>,
 {
-    type Command = Command<E, OM, B, C, SCT>;
+    type Command = Command<E, OM, B, SCT>;
 
-    fn exec(&mut self, commands: Vec<Command<E, OM, B, C, SCT>>) {
+    fn exec(&mut self, commands: Vec<Command<E, OM, B, SCT>>) {
```
michael-yxchen, line 47: nit: I think WAL replay was removed

monad-consensus-state/src/lib.rs
omegablitz: Talked offline, forkpoint validator sets will be based on `root` and not `high_qc`.

omegablitz:
Tested that the `#[ignore]`'d  forkpoint tests pass with these changes.

monad-consensus-state/src/lib.rs
```diff
@@ -231,8 +233,9 @@ where
                 config.max_blocksync_retries,
             ),
 
-            // initial value here doesn't matter; it just must increase monotonically
-            last_proposed_round: Round(0),
+            // we initialize this to high_qc_round to guarantee we don't propose twice for the same
+            // round (even on restart), provided that restart.high_qc is >= crash.high_qc
+            last_proposed_round: high_qc_round,
```
michael-yxchen: should we make this `high_qc_round + 1`? After processing high_qc, the validator will propose on `high_qc_round + 1`

monad-node/src/main.rs
```diff
@@ -236,6 +236,7 @@ async fn run(
     };
     assert!(wal_events.is_empty(), "wal must be cleared after restart");
 
+    let mut last_ledger_len = node_state.forkpoint_config.root.seq_num.0 as usize;
```
michael-yxchen: We're using `ledger().get_num_commits()` as ledger len. Is that value initialized to `root.seq_num` on restart?

monad-state/src/lib.rs
```diff
@@ -179,101 +174,189 @@ impl<SCT: SignatureCollection> Forkpoint<SCT> {
...
+        // 3.
+        if self.validator_sets[0].epoch != self.0.root.epoch {
+            return Err(ForkpointValidationError::InvalidValidatorSetStartEpoch);
+        }
+
+        // 4.
+        if self.validator_sets[0].round.is_none() {
+            return Err(ForkpointValidationError::InvalidValidatorSetStartRound);
+        }
+
```
michael-yxchen, line 277: I might have missed it somewhere. Are we checking `root.round >= validator_sets[0].round.unwrap()` and `root.round < validator_sets[1].round.unwrap_or(Round::MAX)`?

omegablitz: We aren't, I omitted those intentionally. These are the documented trust assumptions:
1. Root is fully trusted
2. Each validator set is trusted to have been valid at some point in time

Because of assumption 2, it's assumed that if a validator_set's round is specified, then it must be correct. Trust assumption 2 would ideally be covered by some proof eventually.

I do think it makes sense to add an assertion here (to be a louder error that indicates that the "trusted" source shouldn't be).

monad-node/src/main.rs
omegablitz: Good catch

PR #886, opened 2024-07-16
Port final tests and restructure testing files

eclark0426:


.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "monad-cxx/monad-execution"]
 	path = monad-cxx/monad-execution
 	url = git@github.com:monad-crypto/monad.git
+[submodule "docker/flexnet/images/monad-orderbook-web"]
+	path = docker/flexnet/images/monad-orderbook-web
+	url = git@github.com:monad-crypto/monad-orderbook-web.git
```
eerkaijun: we should probably have an alternative rather than submoduling this repo, we can chat offline with @ariqchowdhury @michael-yxchen

michael-yxchen: yeah, we should have more flexibility if moving into a separate repo. @eerkaijun mentioned that we don't want to open source the orderbook repo, so we should move integration tests into a new one

@eclark0426 
There's [monad-integration](https://github.com/monad-crypto/monad-integration) that we can take over. I'll review https://github.com/monad-crypto/monad-bft/pull/872 and https://github.com/monad-crypto/monad-bft/pull/873, and merge it into monad-bft. Then we'd only need to migrate one PR over. Probably need to do some set up work there. We can hop on a meeting if needed

eclark0426: Removed the submodule from this repo. Will need additional changes so that this won't break the nightly tests, since those run `net2`, which no longer has an analog without the orderbook. But there's more changes that will be needed to make sure CI works anyway, since everything is ultimately moving to `monad-integration` now.

docker/flexnet/testing-library/monad_flexnet/flexnet.py
```diff
@@ -26,37 +26,43 @@ def __init__(self, topology_path: str | os.PathLike):
...
             yield self
         finally:
             self.stop()
 
-    def run(self, runtime: int | None = None, blocking: bool = False, gen_config: bool = True):
+    def run(self, runtime: int | None = None, blocking: bool = False, gen_config: bool = True, mock_triedb_driver: bool = False):
         # Create output directory
         shutil.copytree(self.topology_path.parent, self.root_dir)
 
+        triedb_driver = 'triedb_driver_mock' if mock_triedb_driver else 'triedb_driver'
```
michael-yxchen: I would suggest coupling eth call target with triedb driver. They should be mock/real triedb version together

docker/flexnet/tests/full-nodes-1.py
michael-yxchen, line 1: I'm assuming these tests are ported from full-node-nonce-test-... Can you copy the description from the original test?

monad-scripts/jenkins/flexnet-ci.sh
```diff
@@ -31,15 +31,13 @@ pip install -r testing-library/requirements.txt
...
-python nets/full-nodes/scripts/net-run.py
-nets/consensus-nodes-nonce-test-1/scripts/net-run.sh --output-dir logs --net-dir nets/consensus-nodes-nonce-test-1/ --flexnet-root . --monad-bft-root ../..
-nets/full-nodes-nonce-test-1/scripts/net-run.sh --output-dir logs --net-dir nets/full-nodes-nonce-test-1/ --flexnet-root . --monad-bft-root ../..
-nets/full-nodes-nonce-test-2/scripts/net-run.sh --output-dir logs --net-dir nets/full-nodes-nonce-test-2/ --flexnet-root . --monad-bft-root ../..
-nets/full-nodes-nonce-test-3/scripts/net-run.sh --output-dir logs --net-dir nets/full-nodes-nonce-test-3/ --flexnet-root . --monad-bft-root ../..
+python tests/net1.py
+python tests/devnet-integration.py
+python tests/full-nodes-1.py
+python tests/full-nodes-2.py
+python tests/reserve-balance.py
```
michael-yxchen, line 40: Are any of the tests removed?

eclark0426: What do you mean by if they're removed? The nonce tests are consolidated slightly into two test files, and I added the reserve balance test, which hadn't yet been added to the CI script. This should run all of the tests which exist, except for `net2` which is run by the nightly CI. The port of `net2` got pushed from this PR to `monad-integration` since it makes use of `monad-orderbook-web`

docker/flexnet/tests/full-nodes-1.py
eclark0426: I can figure something out here if you want, but if I do also separate them out again a bit in the `monad-integration` PRs. I mainly combined things here since the tests seemed related enough that we could save some time setting up/tearing down tests by combining them. Once in `monad-integration`, tests are split into individual functions per test, which has better/more explicit naming

monad-scripts/jenkins/flexnet-ci.sh
michael-yxchen: Ok cool, I just saw that there're fewer test and wondering if any were removed

docker/flexnet/tests/full-nodes-1.py
michael-yxchen: Ok, please coordinate with @bamrith so we have the same coverage for nonce checking

monad-scripts/jenkins/flexnet-ci.sh
eclark0426: Nope, nothing got removed, just that I combined the nonce tests with the other full nodes tests, since they were all just copies of each other with different test functions

docker/flexnet/tests/full-nodes-1.py
eclark0426: I don't think there should be any difference in what is being covered by nonce tests. Just that it's being done in one place instead of spread across multiple tests

PR #887, opened 2024-07-16
Add high_qc to forkpoint

omegablitz:


monad-blocktree/src/blocktree.rs
```diff
@@ -35,11 +37,17 @@ impl std::error::Error for BlockTreeError {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct RootInfo {
+    pub round: Round,
+    pub seq_num: SeqNum,
+    pub block_id: BlockId,
+    pub state_root: StateRootHash,
```
michael-yxchen, line 45: When do we use this state root?

monad-state/src/lib.rs
```diff
@@ -189,14 +191,13 @@ impl<SCT: SignatureCollection> Forkpoint<SCT> {
...
+    /// 1. Carries the validator set to verify QC
+    /// 2. QC is valid against the validator set
+    /// 3. QC epoch-round and validator set epoch-round are set correctly
+    /// 4. Has n+1 validator set locked
+    /// 5. If the last ledger block is on epoch boundary, n+2 validator set is
     ///    locked
-    /// 7. Every validator set except the last one is scheduled to start on a
+    /// 6. Every validator set except the last one is scheduled to start on a
     ///    round
     pub fn validate<VTF>(
```
michael-yxchen, line 202: do we also want n-1 validator set to validate blocks between high_qc and root?

monad-blocktree/src/blocktree.rs
omegablitz: It's only used at genesis - because there's no genesis consensus block we can look the state root up from.

monad-state/src/lib.rs
omegablitz: My thought was that the parent_id chain would be sufficient, and that we wouldn't need to verify the block QCs.

PR #888, opened 2024-07-16
Fix state machine benchmark config

michael-yxchen:


monad-consensus-state/benches/state_machine.rs
```diff
@@ -535,6 +535,12 @@ where
 
 #[allow(clippy::useless_vec)]
 pub fn criterion_benchmark(c: &mut Criterion) {
+    // hardware requirement: CPU (16 cores/assuming 32 threads)
+    // assume half of those are allocated to consensus
+    rayon::ThreadPoolBuilder::new()
+        .num_threads(16)
```
omegablitz: We run with hyperthreading disabled, so did you mean to set this to 8?

michael-yxchen: oh didn't know that we have hyperthreading disabled. Will set it to 8

PR #893, opened 2024-07-18
Flexnet lib fix

michael-yxchen:
Ignore venv and build files in git. Add setuptools to lib requirements or it complains about
```
ModuleNotFoundError: No module named 'pkg_resources'
```
when running the net-run.py script

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,5 +4,6 @@
 /docker/devnet/monad/ledger
 /docker/devnet/monad/wal
 /docker/devnet/monad/triedb
+/docker/flexnet/.venv
 
 test-keys
--- a/docker/flexnet/testing-library/.gitignore
+++ b/docker/flexnet/testing-library/.gitignore
@@ -1,2 +1,3 @@
 *.egg-info/
 __pycache__/
+build
--- a/docker/flexnet/testing-library/requirements.txt
+++ b/docker/flexnet/testing-library/requirements.txt
@@ -1,0 +1,1 @@
+setuptools
--- a/monad-scripts/jenkins/flexnet-ci.sh
+++ b/monad-scripts/jenkins/flexnet-ci.sh
@@ -27,6 +27,7 @@ if [ ! -d "./.venv" ]; then
 fi
 source ./.venv/bin/activate
 
+pip install -r testing-library/requirements.txt
 pip install ./testing-library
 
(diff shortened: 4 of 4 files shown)


docker/flexnet/testing-library/setup.py
```diff
@@ -1,18 +1,19 @@
...
-        'py-evm',
-        'python-on-whales',
-        'requests',
-        'rlp',
-        'urllib3',
-        'web3'
-    ]
+        "eth_account",
+        "gitpython",
+        "setuptools",
```
eclark0426: How would someone be able to run this `setup.py` script in the first place if they don't already have `setuptools` installed?

michael-yxchen: I'm not sure. My guess is that `pip install` works differently?

To reproduce, install python and activate python with
```
pyenv install 3.12
pyenv global 3.12
```
`pip freeze` should output nothing

Then run the flexnet-ci script

eclark0426: Could it be something to do with not being in a venv? It seems to be working fine when CI runs it, and I've had plenty of weird things happen if I use the peach servers without running in a venv. Heck, I can't even run `pyenv install 3.12` without getting an error

eclark0426: > My guess is that pip install works differently?

`pip install` will internally run the setup.py file. I don't think I've ever directly run `setup.py`, I always run it by way of `pip install`

michael-yxchen: same thing happens if I create a venv with a bare python, and run the script from that venv. CI server probably isn't running python with no packages at all

were you able to reproduce the error on master

eclark0426: Using `pyenv global 3.11`, no, everything works fine. I can't install 3.12, I get an error about sqlite

michael-yxchen: it's not a real error. Just that an extension isn't built

eclark0426: Doesn't seem to replicate on 3.12 either

EDIT: That was probably because I had a venv left over from the 3.11 run

michael-yxchen: Discussed offline, moving dependency to a requirements.txt file

PR #917, opened 2024-07-24
Update execution version

omegablitz:


omegablitz:
note to self: remove `load_latest`, no longer exists

omegablitz:
Tested on devnet-staging

PR #918, opened 2024-07-24
Dynamic log level

rgarc:
https://github.com/monad-crypto/monad-internal/issues/386

Adds new events/commands to support changing the logging directive at runtime via a [reload](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/reload/index.html) layer.

Adds a `debug-node` binary to make it easier to send debug commands
```
CLI program to manage validators and metrics

Usage: debug-node --control-panel-ipc-path <CONTROL_PANEL_IPC_PATH> <COMMAND>

Commands:
  validators         Displays the list of validators of the current epoch
  update-validators  Updates the validators using the provided TOML file
  clear-metrics      Clears the metrics
  update-log-filter  Update the logging directive
  help               Print this message or the help of the given subcommand(s)
```

```
Update the logging filter

Usage: debug-node --control-panel-ipc-path <CONTROL_PANEL_IPC_PATH> update-log-filter <--filter <FILTER>|--file <FILE>>

Options:
      --filter <FILTER>  
      --file <FILE>      
  -h, --help      
```

### String Example
This will set the logging directive for all crates to `trace`
```
debug-node  --control-panel-ipc-path controlpanel.sock update-log-filter --filter "trace"
```
But this also supports the same syntax as the env filter so you can set specific crates like so
```
debug-node  --control-panel-ipc-path controlpanel.sock update-log-filter --filter "monad_consensus_state=trace,monad_updaters=info"
```
Setting the directive to an empty string will disable logging entirely
```
debug-node  --control-panel-ipc-path controlpanel.sock update-log-filter --filter ""
```

### File example
You can also specify a complicated directive via a file. For example assume the file `filter.txt` contains the following.
```
monad_triedb_utils=info
monad_proto=info
monad_eth_tx=trace
monad_quic=trace
monad_secp=info
monad_router_scheduler=debug
monad_eth_types=info
monad_async_state_verify=trace
```
Then doing
```
debug-node  --control-panel-ipc-path controlpanel.sock update-log-filter --file filter.txt
```
is equivalent to doing
```
debug-node  --control-panel-ipc-path controlpanel.sock update-log-filter \
    --filter "monad_triedb_utils=info,monad_proto=info,monad_eth_tx=trace,monad_quic=trace,monad_secp=info,monad_router_scheduler=debug,monad_eth_types=info,monad_async_state_verify=trace"
```

### DIFF
--- a/monad-control-panel/src/ipc.rs
+++ b/monad-control-panel/src/ipc.rs
@@ -23,6 +23,9 @@ use tokio::{
 };
 use tokio_util::codec::{FramedRead, FramedWrite, LengthDelimitedCodec};
 use tracing::{debug, error, warn};
+use tracing_subscriber::{reload::Handle, EnvFilter, Registry};
+
+pub type ReloadHandle = Handle<EnvFilter, Registry>;
 
 pub struct ControlPanelIpcReceiver<ST, SCT>
 where
@@ -33,6 +36,8 @@ where
     client_sender: broadcast::Sender<ControlPanelCommand<SCT>>,
 
     metrics: ExecutorMetrics,
+
+    reload_handle: ReloadHandle,
 }
 
 impl<ST, SCT> Stream for ControlPanelIpcReceiver<ST, SCT>
@@ -52,7 +57,11 @@ where
     ST: CertificateSignatureRecoverable,
     SCT: SignatureCollection<NodeIdPubKey = CertificateSignaturePubKey<ST>>,
 {
-    pub fn new(bind_path: PathBuf, buf_size: usize) -> Result<Self, std::io::Error> {
+    pub fn new(
+        bind_path: PathBuf,
+        reload_handle: ReloadHandle,
+        buf_size: usize,
+    ) -> Result<Self, std::io::Error> {
         let (sender, receiver) = mpsc::channel(buf_size);
         let (client_sender, _client_receiver) =
             broadcast::channel::<ControlPanelCommand<SCT>>(buf_size);
@@ -61,6 +70,7 @@ where
         let r = Self {
             receiver,
             client_sender,
+            reload_handle,
 
             metrics: Default::default(),
         };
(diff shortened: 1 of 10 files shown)


ariqchowdhury:
is it possible to pass a file to update-log-filter? 
It would also be useful if there was a "default" setting so if the user changes a bunch of stuff while debugging, they can get back to sensible levels. maybe if we supported files, there could be a default file

rgarc:
> is it possible to pass a file to update-log-filter? It would also be useful if there was a "default" setting so if the user changes a bunch of stuff while debugging, they can get back to sensible levels. maybe if we supported files, there could be a default file

Yea sure. Updated the CLI for `update-log-filter` to accept either a file or a string
```
Update the logging filter

Usage: debug-node --control-panel-ipc-path <CONTROL_PANEL_IPC_PATH> update-log-filter <--filter <FILTER>|--file <FILE>>

Options:
      --filter <FILTER>  
      --file <FILE>      
  -h, --help      
```

monad-control-panel/src/ipc.rs
```diff
@@ -186,6 +208,12 @@ where
 
     fn exec(&mut self, commands: Vec<Self::Command>) {
         for command in commands {
+            if let ControlPanelCommand::Write(WriteCommand::UpdateLogFilter(filter)) = &command {
+                self.reload_handle
+                    .reload(EnvFilter::builder().parse_lossy(filter))
+                    .expect("failed to update logging filter");
```
omegablitz: What can cause `reload` to fail? We could also log a warning and drop the update if a failure here is possible

rgarc: reload only fails in exceptional circumstances
- the subscriber created in main goes out of scope
- the underlying lock gets poisoned 

that being said, we can print a more helpful error if parsing the logging directive fails

PR #921, opened 2024-07-25
enable carriage cost

mikety:
Enable carriage cost validation,  for testing purposes  max_reserve_balance is set to u64::max so triedb account balances will be used as a reserve balance on full nodes.


michael-yxchen:
https://github.com/monad-crypto/monad-bft/blob/85b93054a99a8d38e9444abdee650047ddf4f065/monad-eth-block-policy/src/lib.rs#L350
Something I noticed during the rebase. The code as is throws away elements after the divider, which if I understood correctly , is not the behavior we want from the cache

mikety:
fixed the issue pointed out by Michael, rebased on top of the main branch.

PR #922, opened 2024-07-25
Disable CI benchmarks

omegablitz:
Benchmarks will run in jenkins on commit to master.


omegablitz:
Nvm, you can disable these manually in the Github UI without deleting the workflow - disabled now.

PR #930, opened 2024-07-29
RPC returns block timestamp in unix seconds

eerkaijun:
For integration compatibility.

### DIFF
--- a/docker/devnet/Dockerfile
+++ b/docker/devnet/Dockerfile
@@ -90,7 +90,7 @@ CMD monad-node \
     --secp-identity /monad/config/id-secp \
     --bls-identity /monad/config/id-bls \
     --node-config /monad/config/node.toml \
-    --forkpoint-config /monad/config/forkpoint.toml \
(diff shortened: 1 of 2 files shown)


docker/devnet/Dockerfile
```diff
@@ -90,7 +90,7 @@ CMD monad-node \
     --secp-identity /monad/config/id-secp \
     --bls-identity /monad/config/id-bls \
     --node-config /monad/config/node.toml \
-    --forkpoint-config /monad/config/forkpoint.toml \
+    --forkpoint-config /monad/config/forkpoint.genesis.toml \
```
omegablitz, line 93: We still expect the forkpoint path to be `forkpoint.toml`, but now we have an additional `cp forkpoint.genesis.toml forkpoint.toml` step. This is because monad-bft will overwrite the passed in forkpoint.toml as it's running.

eerkaijun: got it, these arguments will be overwritten in devnets. setting this here so that we can still start the node in a single docker command

PR #933, opened 2024-07-29
Add randao validation to blocksync

eclark0426:


omegablitz:
What was the motivation for this? Asking because relevant to #941

ariqchowdhury:
> What was the motivation for this? Asking because relevant to #941

https://github.com/monad-crypto/monad-internal/issues/304
This issue was what flagged that this was missing, the block policy checks are split and they shouldn't be. Randao is not necessarily special and needing to be checked for blocksync but we should be consistent with block-policy checks and do all of them. Txns need to be validated because we need to get the nonces and gas-params

omegablitz:
Ok, rolled these changes into https://github.com/monad-crypto/monad-bft/pull/941

eclark0426:
Closing this PR in favor of #941 which now incorporates these changes

PR #934, opened 2024-07-29
Add wal-tool and debug-node to flexnet

rgarc:
The [`wal-tool`](https://github.com/monad-crypto/monad-bft/blob/master/monad-wal/examples/wal-tool.rs) is a binary that can parse the write-ahead log written by a monad node and compute useful statistics for asserting in tests.

The [`debug-node`](https://github.com/monad-crypto/monad-bft/blob/master/monad-control-panel/src/bin/debug_node.rs) is a binary that sends useful debug commands to a running node.

This PR builds both binaries and wraps them in docker images so that they can be invoked in flexnet.

### DIFF
--- a/docker/flexnet/images/control-panel/Dockerfile
+++ b/docker/flexnet/images/control-panel/Dockerfile
@@ -1,0 +1,13 @@
+# control-panel: a tool to send debug commands to a running node over UNIX sockets
+
+FROM ubuntu:24.04
+WORKDIR /usr/src/monad-bft
+
+RUN apt update && apt install -y openssl
+
+ENV LD_LIBRARY_PATH="/usr/local/lib:$LD_LIBRARY_PATH"
+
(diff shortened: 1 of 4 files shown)


docker/flexnet/images/control-panel/Dockerfile
```diff
@@ -0,0 +1,13 @@
+# ethtx: a tool to send random eth transactions over UNIX sockets
```
eclark0426: Comment left over from ethtx isn't accurate here

docker/flexnet/images/wal/Dockerfile
```diff
@@ -0,0 +1,13 @@
+# ethtx: a tool to send random eth transactions over UNIX sockets
```
eclark0426: Same for this comment

eclark0426:
Sorry I haven't gotten to looking at this again - have been in the process of moving everything over to `monad-integration`, and that's pretty much all done now, pending updates to CI and removing things from `monad-bft`

PR #936, opened 2024-07-29
Support configurable validator stakes

rgarc:
Right now, all validators are configured with stake of 1. To test weighted leader election, it is necessary to make this configurable.

### DIFF
--- a/docker/flexnet/testing-library/monad_flexnet/topology/node.py
+++ b/docker/flexnet/testing-library/monad_flexnet/topology/node.py
@@ -57,7 +57,7 @@ class Node:
             '> /monad/logs/client.log 2>&1'
         ]
 
-    def __init__(self, name: str, has_execution: bool = False, has_rpc: bool = False, upload_speed: int = 100, download_speed: int = 100):
+    def __init__(self, name: str, has_execution: bool = False, has_rpc: bool = False, upload_speed: int = 100, download_speed: int = 100, stake: int = 1):
         self.name = name
         self.has_execution = has_execution
         self.has_rpc = has_rpc
@@ -70,6 +70,7 @@ class Node:
         self.is_byzantine = False
         self.node_root = None
         self.test_mode = False
+        self.stake = stake
 
     def set_byzantine(self, is_byzantine: bool = True):
         self.test_mode = True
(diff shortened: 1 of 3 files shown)


docker/flexnet/testing-library/monad_flexnet/topology/node.py
```diff
@@ -53,8 +53,9 @@ def _run_cmd(self):
             '> /monad/logs/client.log 2>&1'
         ]
 
-    def __init__(self, name: str, has_execution: bool = False, has_rpc: bool = False, upload_speed: int = 100, download_speed: int = 100):
+    def __init__(self, name: str, stake: int, has_execution: bool = False, has_rpc: bool = False, upload_speed: int = 100, download_speed: int = 100):
```
eclark0426: Since stake doesn't have to be unique, I'd suggest giving it the default value of 1 here so that manual calls to create a node aren't required to explicitly set the stake value

docker/flexnet/testing-library/monad_flexnet/topology/region.py
```diff
@@ -9,8 +9,8 @@ def __init__(self, name: str, intraregion_latency: LatencyProfile = LatencyProfi
         self.intraregion_latency = intraregion_latency
         self.interregion_latencies = {}
 
-    def create_node(self, node_name: str, **kwargs):
-        new_node = Node(node_name, **kwargs)
+    def create_node(self, node_name: str, stake: int, **kwargs):
```
eclark0426: At which point it doesn't need to be explicitly passed through here, since it's included in kwargs

docker/flexnet/testing-library/monad_flexnet/topology/topology.py
```diff
@@ -23,6 +23,7 @@ def from_json(cls, topology_file: str | os.PathLike):
             for node_dict in region_dict['nodes']:
                 region.create_node(
                     node_dict['name'],
+                    node_dict['stake'] or 1,
```
eclark0426: Eventually these all change to `node_dict.get()` - for consistency with changes-to-come, I'd recommend `node_dict.get('stake', 1)` instead of the `or` syntax

PR #938, opened 2024-07-30
Configurable chain id

eerkaijun:
Will also need to fix `eth_call.cpp` interface change.


eerkaijun:
I can add the chain id validation on consensus in a separate PR
https://github.com/monad-crypto/monad-internal/issues/381

PR #941, opened 2024-07-30
Refactor blocksync

omegablitz:
Lift blocksync out into MonadState. This is in preparation for statesync, as we need to be able to blocksync before consensus is live.


monad-executor-glue/src/lib.rs
```diff
@@ -209,11 +195,10 @@ pub enum ConsensusEvent<ST, SCT: SignatureCollection> {
...
         unverified_message: Unverified<ST, Unvalidated<ConsensusMessage<SCT>>>,
     },
-    Timeout(TimeoutVariant),
-    BlockSyncResponse {
-        sender: NodeId<SCT::NodeIdPubKey>,
-        unvalidated_response: Unvalidated<BlockSyncResponseMessage<SCT>>,
-    },
+    Timeout,
+    /// a block that was previously requested
+    /// this is an invariant; if the block is not recognized, this will panic
```
omegablitz: log error instead of panic

omegablitz:
Tested that this works locally with the real executors by running two nodes.
Verified that one node is able to catch up via the other servicing blocksync requests.

PR #942, opened 2024-07-31
Check chain id during tx insertion and proposal validation

eerkaijun:
This prevents the execution client from crashing if someone send a transaction with an invalid nonce. Closes https://github.com/monad-crypto/monad-internal/issues/381


monad-node/src/config/mod.rs
```diff
@@ -27,6 +27,7 @@ pub type SignatureCollectionType =
 pub struct NodeConfig {
     pub name: Option<String>,
     pub network_name: Option<String>,
+    pub chain_id: u64,
```
eerkaijun, line 30: Should we set this as an optional (with a default value) or required field in the node config file?

ariqchowdhury: I think make it required. Don't think there is a sensible default value that client should use.

monad-eth-block-validator/src/lib.rs
```diff
@@ -57,6 +60,10 @@ impl<SCT: SignatureCollection, RBCT: ReserveBalanceCacheTrait>
         let mut validated_txns: Vec<EthTransaction> = Vec::with_capacity(eth_txns.len());
 
         for (eth_txn, signer) in eth_txns.into_iter().zip(signers) {
+            if eth_txn.chain_id().unwrap_or_default() != self.chain_id {
```
ariqchowdhury: this one too

eerkaijun: oops missed that

PR #946, opened 2024-07-31
Remove wait and spammy logs from full-nodes-1 tests

eclark0426:


docker/flexnet/tests/full-nodes-1.py
```diff
@@ -54,18 +47,10 @@ def main():
...
-            time.sleep(5)
-            print(f'sender balance {client.get_account_balance(sender_account)}')
-            print(f'receiver balance {client.get_account_balance(receiver_account)}')
-
+            time.sleep(2)
             for other_client_name, other_client in clients.items():
-                try:
-                    other_client.wait_for_transaction_receipt(tx_hash)
-                except Exception as e:
-                    print(f'{other_client_name}: wait_for_transaction_receipt exception {e}')
```
michael-yxchen, line 68: Printing in other places are fine. I just think we shouldn't print error here, as it's expected to fail

docker/flexnet/tests/full-nodes-1.py
```diff
@@ -75,18 +60,10 @@ def main():
...
```
michael-yxchen, line 89: and here

eclark0426:
Will drop the second commit that removes the additional logs, then

michael-yxchen:
Thanks

PR #948, opened 2024-08-01
Rpc gas allowance capped at block gas limit

philipglazman:
Fixes eth_call error when a large account attempts to estimate gas or use eth_call. Noticed that after making large balance genesis accounts, `eth_call` and `eth_estimateGas` were failing.


eerkaijun:
btw yesterday I temporarily decrease the balance in `docker/devnet/monad/config/genesis.json` to make things work, feel free to increase it back up if this solve the issue.

philipglazman:
> Can you also update the state root in forkpoint.toml to the root after genesis?

Already updated the state root in a commit yesterday. The genesis was changed since then w/o updating the state root. This commit updates the genesis accounts to my commit from yesterday with the matching state root.

michael-yxchen:
Please update the INITIAL_ACCOUNT_BAL in flexnet so the tests have a consistent view

PR #949, opened 2024-08-01
Update execution submodule

eerkaijun:


omegablitz:
Let's get this one in as well: https://github.com/monad-crypto/monad/pull/793

PR #950, opened 2024-08-01
Add TxPool IPC ingest watermark

omegablitz:


monad-ipc/src/lib.rs
```diff
@@ -84,23 +92,34 @@ where
...
-                Ok(_) => debug!("bytes received from IPC and sent to channel"),
-                Err(mpsc::error::TrySendError::Full(_)) => todo!(
-                    "IPC recv channel full, max_capacity={}",
-                    event_channel.max_capacity()
-                ),
+            match tx_batch_channel.try_send(tx.to_vec()) {
+                Ok(()) => {
+                    trace!("bytes received from IPC and sent to channel");
+                    let capacity = tx_batch_channel.capacity();
+                    if capacity < tx_batch_channel.max_capacity() / 2 {
```
ariqchowdhury: might be useful to make this configurable

omegablitz: Just made the executor configurable, but hard-coded the value in monad-node.
Were you thinking that we should also expose the tx_batch_size/max_queued_batches/watermark value in node.toml?

ariqchowdhury: yeah, I think we might want that flexibility during testing

PR #951, opened 2024-08-01
Recover incoming txpool txn batches in parallel

ariqchowdhury:


monad-state/src/mempool.rs
```diff
@@ -138,19 +129,24 @@ where
...
-                    } else {
-                        self.metrics.txpool_events.external_inserted_txns += 1;
-                    }
+                let num_txns = txns.len() as u64;
+                let valid_encoded_txs =
+                    self.txpool
+                        .insert_tx(txns, self.block_policy, self.reserve_balance_cache);
+
+                let num_valid_txns = valid_encoded_txs.len() as u64;
+                self.metrics.txpool_events.external_inserted_txns += 1;
```
omegablitz: Counter not incremented correctly

omegablitz:
One thing I just thought of... should we make the number of threads allocated to rayon configurable? Execution spins on a bunch of threads right now, so the default rayon behavior of running num_cpu threads will be pathological.

PR #956, opened 2024-08-02
Precompute carriage cost usage

michael-yxchen:
Closes: https://github.com/monad-crypto/monad-internal/issues/409

### DIFF
--- a/monad-eth-block-policy/src/lib.rs
+++ b/monad-eth-block-policy/src/lib.rs
@@ -185,6 +185,8 @@ struct CommittedTxnBuffer {
     // TODO: create a reserve balance map for transactions when validating
     // Block, and store it in EthValidatedBlock
     txns: SortedVectorMap<SeqNum, (Vec<EthTransaction>, BlockAccountNonce)>,
+    /// cache carriage cost usage by each account in each block
+    carriage_costs: SortedVectorMap<SeqNum, BTreeMap<EthAddress, Balance>>,
     size: usize, // should be execution delay
 }
 
@@ -197,6 +199,7 @@ impl CommittedTxnBuffer {
     fn new(size: usize) -> Self {
         Self {
             txns: Default::default(),
+            carriage_costs: Default::default(),
             size,
         }
     }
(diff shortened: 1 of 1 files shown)


monad-eth-block-policy/src/lib.rs
```diff
@@ -268,6 +274,16 @@ impl CommittedTxnBuffer {
...
+        // compute carriage cost usage in the block
+        let mut block_carriage_cost = BTreeMap::new();
+        for txn in block.validated_txns.iter() {
+            let entry = block_carriage_cost
+                .entry(EthAddress(txn.signer()))
+                .or_insert(0_u128);
+            *entry += compute_txn_carriage_cost(txn);
+        }
+        self.carriage_costs
+            .insert(block_number, block_carriage_cost);
```
mikety: Assert that None is returned - i.e. no such blocks was in the cache before?

monad-eth-block-policy/src/lib.rs
```diff
@@ -221,15 +224,14 @@ impl CommittedTxnBuffer {
...
+        for (&cache_seq_num, block_carriage_costs) in self.carriage_costs.iter() {
             if cache_seq_num > base_seq_num {
                 assert_eq!(next_validate, cache_seq_num);
-                for txn in txns {
-                    if EthAddress(txn.signer()) == *eth_address {
-                        carriage_cost += compute_txn_carriage_cost(txn);
-                    }
+                if let Some(block_carriage_cost) = block_carriage_costs.get(eth_address) {
+                    carriage_cost += block_carriage_cost;
                 }
```
mikety, line 235: assert/warn that the expected eth_address not found in cache?
nit: its an account_carriage_cost for the committed block

michael-yxchen: this cache stores carriage costs usage for accounts with transactions in the block. It's ok if the account we're querying doesn't have transaction in previous blocks

michael-yxchen:
Thanks, I can add caching for pending block carriage cost in a separate PR

PR #959, opened 2024-08-04
State backend refactor

omegablitz:
1. Unify StateBackend/ReserveBalanceCacheTrait
2. Expose get_earliest_block_id
3. Add batching interface to StateBackend
   - This doesn't actually use the async triedb interface as yet
4. Clean up NeedsSync error handling


omegablitz:
Important changes are in:
- monad-eth-block-policy/src/lib.rs
- monad-eth-txpool/src/lib.rs
- monad-state-backend/src/lib.rs
- monad-triedb-cache/src/lib.rs

monad-eth-block-policy/src/lib.rs
```diff
@@ -334,40 +333,56 @@ impl EthBlockPolicy {
...
+        // when purging, we never purge nonces newer than last_commit - delay
+
+        let base_seq_num = consensus_block_seq_num.max(self.execution_delay) - self.execution_delay;
+        let cache_miss_statuses =
+            state_backend.get_account_statuses(base_seq_num, cache_misses.iter().copied())?;
+        account_nonces.extend(
+            cache_misses
+                .into_iter()
+                .zip_eq(cache_miss_statuses)
+                .map(|(address, status)| (address, status.map_or(0, |status| status.nonce))),
```
michael-yxchen, line 382: If `get_account_statuses` returns none for an account address, it means account doesn't exist and block is not coherent

monad-eth-block-policy/src/lib.rs
```diff
@@ -383,161 +398,138 @@ impl EthBlockPolicy {
...
-            ReserveBalanceCacheResult::None => {
-                return Err(CarriageCostValidationError::AccountNoExist);
-            }
-        }
+        let addresses = addresses.unique().collect_vec();
+        let account_balances = state_backend
+            .get_account_statuses(base_seq_num, addresses.iter().copied())?
+            .into_iter()
+            .map(|maybe_status| {
+                maybe_status.map_or(0, |status| status.balance.min(self.max_reserve_balance))
```
michael-yxchen, line 422: Similar here, I think we should return error if account not found

monad-eth-block-policy/src/lib.rs
omegablitz: Discussed offline

monad-eth-block-policy/src/lib.rs
omegablitz: Discussed offline

PR #960, opened 2024-08-05
debug_traceTransaction rpc

tongzhi1998:


monad-rpc/src/trace_handlers.rs
```diff
@@ -0,0 +1,138 @@
...
+    pub input: Bytes,
+
+    pub output: Bytes,
+
+    pub status: U8,
+
+    pub depth: U64,
+}
+
+impl Decodable for CallFrame {
```
philipglazman, line 36: I think you can derive Decodable trait using `RlpDecodable` macro

monad-rpc/src/trace_handlers.rs
```diff
@@ -0,0 +1,138 @@
...
+use monad_blockdb_utils::BlockDbEnv;
+use monad_triedb_utils::{TriedbEnv, TriedbResult};
+use reth_rpc_types::Transaction;
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+use tracing::{debug, trace};
+
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+pub struct CallFrame {
+    pub typ: U8,
```
philipglazman: `serde(alias="type")`

monad-triedb-utils/src/lib.rs
```diff
@@ -347,4 +371,31 @@ impl TriedbEnv {
 
         (key, num_nibbles)
     }
+
+    fn create_call_frame_key(txn_index: u64) -> (Vec<u8>, u8) {
```
eerkaijun: i refactored some triedb utils recently. after rebasing, this can go into `monad-triedb-utils/src/key.rs`

philipglazman:
> if execution does not enable call tracer, what happens when someone calls trace transaction in rpc? it will just return null right?

Yes, it will return null

philipglazman:
rebased @eerkaijun

eerkaijun:
@philipglazman I rebased this against latest master. when running against the anvil diff test though https://github.com/monad-crypto/monad-orderbook-web/pull/102 I got an error

```
 assert top_level_trace[key] == top_level_trace_anvil[key], (
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I tested with Blockscout and it works well, so I guess probably just the slightly discrepancy here and we can move forward with it

philipglazman:
> @philipglazman I rebased this against latest master. when running against the anvil diff test though [monad-crypto/monad-orderbook-web#102](https://github.com/monad-crypto/monad-orderbook-web/pull/102) I got an error
> 
> ```
>  assert top_level_trace[key] == top_level_trace_anvil[key], (
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> ```
> 
> I tested with Blockscout and it works well, so I guess probably just the slightly discrepancy here and we can move forward with it

Fixed in last commit. Issue was the code result being returned

monad-rpc/src/trace_handlers.rs
```diff
@@ -0,0 +1,443 @@
...
+        .await?
+        .iter()
+        .enumerate()
+        .map(|(idx, tx)| (idx, tx.hash()))
+        .collect::<Vec<_>>();
+
+    let mut resp = Vec::new();
+    for (idx, tx_hash) in tx_ids {
+        let Some(traces) =
+            fetch_trace_for_tx(triedb_env, idx as u64, block_num, &params.tracer).await?
```
eerkaijun: we might want to traverse the call frame tree rather than fetch traces for each transaction separately. there's a `traverse_triedb` helper function in `monad-triedb` which can help with this

monad-rpc/src/trace_handlers.rs
```diff
@@ -0,0 +1,443 @@
...
```
eerkaijun: same as above

monad-rpc/src/main.rs
```diff
@@ -52,6 +51,9 @@ use crate::{
         monad_trace_block, monad_trace_call, monad_trace_callMany, monad_trace_get,
         monad_trace_transaction,
     },
+    trace_handlers::{
+        monad_debugTraceBlockByHash, monad_debugTraceBlockByNumber, monad_debugTraceTransaction,
```
eerkaijun: for naming consistency should these be `monad_debug_traceBlockByHash`, `monad_debug_traceBlockByNumber` and `monad_debug_traceTransaction`?

monad-rpc/src/trace_handlers.rs
```diff
@@ -0,0 +1,443 @@
...
+        else {
+            return Err(JsonRpcError::internal_error("traces not found".to_string()));
```
eerkaijun, line 265: just to double check here, does every transaction guaranteed to have traces?

philipglazman: Yes, every transaction should have a call trace. Even if it is a simple native transfer.

monad-rpc/src/trace_handlers.rs
philipglazman: ah, great idea. I'll add this

philipglazman: @eerkaijun do you know if order of transactions is preserved if I call `get_transactions` then `get_call_frames` ? I need to include the txid for each top call frame, and that is not part of the rlp encoding for the call frame

eerkaijun: yes, the transactions and call frames array are both sorted by txid when we traverse triedb

PR #963, opened 2024-08-05
Compute block carriage cost in EthValidator

michael-yxchen:
Ensures we only compute carriage costs once per block. Resolves pending block carriage cost recalculation bug

Closes: https://github.com/monad-crypto/monad-internal/issues/419

### DIFF
--- a/monad-eth-block-policy/src/lib.rs
+++ b/monad-eth-block-policy/src/lib.rs
@@ -59,7 +59,7 @@ fn compute_intrinsic_gas(tx: &EthSignedTransaction) -> u64 {
     intrinsic_gas
 }
 
-pub fn compute_txn_carriage_cost(txn: &EthTransaction) -> u128 {
+pub fn compute_txn_carriage_cost(txn: &EthSignedTransaction) -> u128 {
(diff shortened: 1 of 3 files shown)


monad-eth-block-policy/src/lib.rs
```diff
@@ -312,49 +319,38 @@ impl CommittedTxnBuffer {
...
     fn update_committed_block<SCT: SignatureCollection>(&mut self, block: &EthValidatedBlock<SCT>) {
         let block_number = block.get_seq_num();
-        if let Some((&last_block_num, _)) = self.txns.last_key_value() {
+        if let Some((&last_block_num, _)) = self.blocks.last_key_value() {
             assert_eq!(last_block_num + SeqNum(1), block_number);
         }
 
-        if self.txns.len() >= self.size * 2 {
-            let (&first_block_num, _) = self.txns.first_key_value().expect("txns non-empty");
+        // FIXME: overflow if self.size == usize::MAX
```
omegablitz: This _shouldn't_ ever overflow anymore (delay is never SeqNum::MAX), although maybe we should assert that eventually

PR #966, opened 2024-08-07
Switch monad-raptorcast over to custom Raptor decoder

buytenh:
This PR imports the in-house developed Raptor code decoder corresponding
to the previously merged Raptor encoder, and switches monad-raptorcast
and raptor_bench over to use this decoder.

There is some work left to be done, but the initial performance numbers
look good.  With the old decoder, raptor_bench gets these numbers on my
laptop:

```
encoder/decoder/Decoding
                        time:   [77.529 ms 78.731 ms 79.760 ms]
                        thrpt:  [47.827 MiB/s 48.453 MiB/s 49.203 MiB/s]
```

And with the new decoder:

```
encoder/decoder/Decoding
                        time:   [53.722 ms 53.814 ms 53.910 ms]
                        thrpt:  [70.760 MiB/s 70.887 MiB/s 71.008 MiB/s]
                 change:
                        time:   [-32.555% -31.648% -30.584%] (p = 0.00 < 0.05)
                        thrpt:  [+44.059% +46.302% +48.269%]
                        Performance has improved.
```

raptor_bench seems to work, but the monad-raptorcast/src/lib.rs changes
are untested -- and neither of these verify that the decoder properly
decodes the output of the encoder, but there is a separate test that
does that, and a few more tests will be submitted shortly.

Some notes about the decoder:

- The decoder implements the belief-propagation decoding and inactivation
  decoding algorithms described in the "Raptor Codes" monograph by Amin
  Shokrollahi and Michael Luby, except that where the monograph describes
  a decoding procedure that is run once, to completion, once all of the
  encoded symbols that were going to be received have been received, this
  decoder uses an incremental version of that procedure, where we try to
  perform belief-propagation decoding every time we receive an encoded
  symbol, and where we initiate inactivation decoding once we have a
  reasonable belief that we will be able to finish the decoding process
  with our current set of received encoded symbols.

  One advantage of this decoding strategy is that if the inactivation
  decoding process fails (for example, if there are linear dependencies
  between the received encoded symbols), we do not need to restart the
  decoding procedure from the beginning once more encoded symbols are
  received, but instead, we can re-use the decoding work that has
  already been performed in the previous decoding attempt(s).

  Another property of using an incremental decoding strategy is that it
  leads to spreading the CPU-intensive decoding work that would only be
  initiated at the end of the packet reception process in the monograph
  version of the decoding process out more evenly over the duration of
  the packet reception process.  If there is a large time interval between
  the reception of the initial encoded symbol and the final encoded symbol,
  the conjectured (but as of yet unverified) advantage of this approach
  is that it will hopefully cause the time interval between the reception
  of the final symbol and the completion of the decoding process to be
  smaller than in the monograph version of the decoding process, in
  other words, it will hopefully improve our decoding latency.

- The core Raptor decoder added by this PR is implemented such that it
  does not touch any of the encoded or decoded data.  You invoke a
  decoder method every time a new encoded symbol is received, and in
  response, it provides you with a list of XOR operations that you have
  to perform on the data ("XOR the buffer that encoded symbol N was
  received in with the buffer that encoded symbol M was received in"),
  and once the decoding process is done, it gives you the source data
  fragment to buffer mapping, which you can then use to produce a
  linearized version of the source data if you need one.

  This interface is a bit too bare-bones for RaptorCast use, so there is
  also a managed wrapped around this decoder (ManagedDecoder), which
  does the buffer management and source data linearization for you, and
  RaptorCast and raptor_bench use the decoder via this wrapper.

Some future work involves:

- The decoder has a decision to make regarding at which number of
  received encoded symbols to initiate inactivation decoding, and this
  involves a trade-off between allowing recovery of the source data with
  fewer received encoded symbols by starting the decoding process early,
  and saving CPU by allowing more of the intermediate symbol recovery to
  be done using belief propagation and less of it using inactivation
  decoding by starting the decoding process late.

  We currently start inactivation decoding when we will have received
  a number of encoded symbols that is 1.5 times the number of source
  symbols in the application message, and this seems to give relatively
  good decoding performance based on some quick-and-dirty testing, but
  there is no proper justification for this particular choice of
  threshold, and we should spend some effort on picking a good threshold.

- Since the new decoder is more efficient, a larger proportion of the
  spent time is now used by the cryptographic primitives, so while
  further decoder optimizations are possible, it might also be worth
  looking into e.g. increasing the RaptorCast Merkle tree depth to see
  what effect that will have on performance.

- The managed decoder currently enforces that all received encoded
  symbols have the same length by asserting if that is not the case,
  and this could potentially be exploited by a malicious proposer, and
  we need to detect this case and potentially report it as a slashable
  offense via some standard API.

  The other case I can think of where a proposer can potentially also
  cause assertion failures in the decoder is by sending an oversized
  proposal, i.e. consisting of more than the Raptor-10 specified maximum
  number of 8192 source symbols.  The Raptor(Cast) code needs to be
  audited for more such cases.

- When performing inactivation decoding, we currently semi-randomly
  choose an intermediate symbol to inactivate, but there is a better
  way to do this, described in section 5.5.2.2. of RFC 5053, and I
  need to try out this algorithm.

- There are a few more tests I want to import, and a pure Raptor
  encoding/decoding bench I want to add.

### DIFF
--- a/monad-raptorcast/src/lib.rs
+++ b/monad-raptorcast/src/lib.rs
@@ -20,8 +20,8 @@ use monad_dataplane::event_loop::{BroadcastMsg, Dataplane, UnicastMsg};
 use monad_executor::{Executor, ExecutorMetrics, ExecutorMetricsChain};
 use monad_executor_glue::{Message, RouterCommand};
 use monad_merkle::{MerkleHash, MerkleProof, MerkleTree};
+use monad_raptor::ManagedDecoder;
 use monad_types::{Deserializable, Epoch, NodeId, Round, RouterTarget, Serializable, Stake};
-use raptor_code::SourceBlockDecoder;
 
 pub struct RaptorCastConfig<ST>
 where
@@ -59,7 +59,7 @@ where
     // generate a bunch of linearly dependent chunks and cause unbounded memory usage.
     // TODO GC these based on time elapsed since first chunk? Right now we do no GC - we can't do
     // this by round because some node may be trying to sync to us with a much older round.
-    message_cache: BTreeMap<MessageCacheKey<CertificateSignaturePubKey<ST>>, SourceBlockDecoder>,
+    message_cache: BTreeMap<MessageCacheKey<CertificateSignaturePubKey<ST>>, ManagedDecoder>,
     signature_cache:
         HashMap<[u8; HEADER_LEN as usize - 65 + 20], NodeId<CertificateSignaturePubKey<ST>>>,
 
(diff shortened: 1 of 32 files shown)


monad-raptorcast/src/lib.rs
```diff
@@ -437,20 +437,31 @@ where
...
+                        let symbol_len = parsed_message.chunk.len();
+
                         // data_size is always greater than zero, so this division is safe
-                        let num_source_symbols =
-                            app_message_len.div_ceil(parsed_message.chunk.len());
-                        SourceBlockDecoder::new(num_source_symbols)
+                        let num_source_symbols = app_message_len.div_ceil(symbol_len);
+
+                        // TODO: verify unwrap
+                        ManagedDecoder::new(num_source_symbols, symbol_len).unwrap()
```
omegablitz, line 446: If this errors, we can `tracing::warn` and continue. It's fine if we drop any app_message.

PR #971, opened 2024-08-07
Handle eth_call edge case from zero address

eerkaijun:
A small edge case which is currently not handled well by eth_call

### DIFF
--- a/monad-rpc/src/call.rs
+++ b/monad-rpc/src/call.rs
@@ -311,7 +311,7 @@ pub async fn monad_eth_call(
         BlockTags::Number(block_number) => block_number.0,
     };
 
-    let Some(block_header) = blockdb_env
+    let Some(mut block_header) = blockdb_env
         .get_block_by_tag(monad_blockdb_utils::BlockTags::Number(block_number))
         .await
     else {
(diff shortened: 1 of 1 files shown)


monad-rpc/src/call.rs
```diff
@@ -311,33 +311,46 @@ pub async fn monad_eth_call(
...
-        params.transaction.gas = allowance.map(U256::from);
-    };
+    let sender = params.transaction.from.unwrap_or_default();
+    match sender {
+        Address::ZERO => {
+            // for eth_call from a zero address, we want to override the block base fee to be zero
+            // so that reading from the smart contract does not require the zero address
+            // to have any gas balance
+            block_header.block.header.base_fee_per_gas = Some(0);
+            params.transaction.fill_gas_prices(U256::from(0))?;
```
philipglazman, line 329: I don't think you need to call `fill_gas_prices` since we use min(max_fee_per_gas,base_fee)

eerkaijun: we still need to set the `max_fee_per_gas` to 0 in the case where the user does not specify it?

philipglazman: The default is 0 when the user does not specify it

eerkaijun: in the event that the user does specify a non-zero `max_fee_per_gas`, we still want to override that to 0, otherwise it will still run into insufficient balance by execution

PR #973, opened 2024-08-07
NullBlock mock-swarm test

michael-yxchen:


monad-mock-swarm/tests/null_block.rs
```diff
@@ -0,0 +1,278 @@
...
+    // propose null blocks
+    fn generate_swarm() -> (
+        Nodes<NullBlockSwarm>,
+        NodeId<
+            SignatureCollectionPubKeyType<
+                <NullBlockSwarm as SwarmRelation>::SignatureCollectionType,
+            >,
+        >,
+    ) {
+        let execution_delay = SeqNum(4);
```
ariqchowdhury, line 104: I think this value is coming from the impl of `MissingNextStateRoot`?
Should probably add a constructor for `MissingNextStateRoot` so that we can config the execution_delay.
Having to know that its the same will probably cause a debugging headache in the future

michael-yxchen: I think a delay of zero makes more logical sense for `MissingNextStateRoot`. Will fix

PR #974, opened 2024-08-08
TrieDB async read

bamrith:


monad-triedb/src/lib.rs
```diff
@@ -150,6 +188,42 @@ impl Handle {
         Some(value)
     }
 
+    pub fn triedb_poll(&self) {
```
omegablitz: nit: make this private

monad-triedb/src/lib.rs
```diff
@@ -150,6 +188,42 @@ impl Handle {
...
+        let _io_count = unsafe { bindings::triedb_poll(self.db_ptr, true, usize::MAX) };
+    }
+
+    pub fn read_async(
```
omegablitz: nit: make this private

maybe add comment that the receiver must be resolved before key is dropped?

monad-triedb/src/lib.rs
```diff
@@ -197,6 +271,39 @@ impl Handle {
...
+            num_accounts += 1;
+            let (triedb_key, key_len_nibbles) = create_addr_key(eth_address.as_ref());
+            self.read_async(
+                triedb_key.as_ref(),
+                key_len_nibbles,
+                block_id,
+                completed_counter.clone(),
+            )
+            .map(|receiver_result| {
+                let result = receiver_result.unwrap();
```
omegablitz: nit: `let maybe_account = receiver_result.expect("receiver can't be canceled");`

monad-triedb/src/lib.rs
```diff
@@ -197,6 +271,39 @@ impl Handle {
 
         Handle::rlp_decode_account(account_rlp)
     }
+
+    pub fn get_accounts<'a>(
+        &self,
+        eth_addresses: impl Iterator<Item = &'a EthAddress>,
+        block_id: u64,
+    ) -> Vec<Option<EthAccount>> {
+        let completed_counter = Arc::new(Mutex::new(0));
```
omegablitz: Can you switch the mutex to [AtomicUsize](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicUsize.html) with [Ordering::SeqCst](https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html#variant.SeqCst)?

Asking only because we're spinning on completed_counter below.

PR #979, opened 2024-08-09
Fix AsyncStateVerfiy memory leak

michael-yxchen:


monad-async-state-verify/src/peer.rs
```diff
@@ -56,14 +56,14 @@ impl<SCT: SignatureCollection> Default for ExecutionStateRootRecord<SCT> {
...
 
-// TODO: garbage clean old state roots
 #[derive(Debug, Clone)]
 pub struct PeerAsyncStateVerify<SCT, VT>
 where
     SCT: SignatureCollection,
     VT: ValidatorSetType<NodeIdPubKey = SCT::NodeIdPubKey>,
 {
     /// Collection of StateRoots after executing each block
+    /// FIXME: memory leak here
```
omegablitz: remove FIXME?

PR #982, opened 2024-08-09
Copy only first matching .so to devnet container

eclark0426:


omegablitz:
Closed in favor of #983

PR #984, opened 2024-08-09
Data collection from logs

eerkaijun:
Currently collect a few things: 

1. shows the duration to collect the votes and the total number of votes collected

vote_collection

2. flag the round numbers where the node does not receive a proposal or create a vote. prints out missing rounds

3. duration between committed blocks

block_time

4. duration between receiving consecutive proposals

proposal_time

5. time taken for collecting transactions and creating a proposal

proposal_creation

6. duration between timeout (or timeouts) to the next received proposal

timeout_duration

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -1043,7 +1043,7 @@ where
                 target: RouterTarget::PointToPoint(next_leader),
                 message: msg,
             };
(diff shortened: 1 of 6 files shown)


eerkaijun:
I think these are the main data for collection, good to be reviewed and merged. Will make the CLI to be interactive and also collecting periodic summary data in a separate PR

eerkaijun:
Good point, will make the change

PR #985, opened 2024-08-10
Handle both input and data fields for backward compatibility

eerkaijun:
Closes https://github.com/monad-crypto/monad-internal/issues/437

### DIFF
--- a/monad-rpc/src/call.rs
+++ b/monad-rpc/src/call.rs
@@ -99,6 +99,16 @@ impl CallRequest {
     }
 }
 
+#[derive(Debug, Clone, Default, Eq, PartialEq, Serialize, Deserialize)]
+pub struct CallInput {
+    /// Transaction data
+    pub input: Option<alloy_primitives::Bytes>,
+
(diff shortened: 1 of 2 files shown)


monad-rpc/src/call.rs
```diff
@@ -292,6 +292,16 @@ pub async fn monad_eth_call(
...
+    if let (Some(input), Some(data)) = (
+        &params.transaction.input.input,
+        &params.transaction.input.data,
+    ) {
+        if input != data {
+            return Err(JsonRpcError::invalid_params());
+        }
+    } else if params.transaction.input.input.is_none() && params.transaction.input.data.is_some() {
+        params.transaction.input.input = params.transaction.input.data.clone();
+    }
```
ariqchowdhury: This is difficult to parse. I think its more clear with a match statement
```
let input = match (params.input, params.data) {
   (Some(a), Some(b)) => {
      if a != b return;
   (None, Some(b)) => b,
    ...
   }
}
```

eerkaijun: sounds good, done

PR #991, opened 2024-08-13
Initial reserve balance unit tests

bamrith:


monad-consensus-state/src/lib.rs
```diff
@@ -5102,4 +5070,323 @@ mod test {
...
+            SimpleRoundRobin::default(),
+            || EthBlockPolicy::new(GENESIS_SEQ_NUM, u128::MAX, EXECUTION_DELAY.0, 0, 1337),
+            || {
+                InMemoryStateInner::new(
+                    u128::MAX,
+                    EXECUTION_DELAY,
+                    InMemoryBlockState::genesis(
+                        std::iter::once((
+                            sender_1_address,
+                            EthAccount::new(0, total_carriage_cost, None),
```
michael-yxchen: It's more clear to use `txn_nonce_zero_carriage_cost` directly

monad-consensus-state/src/lib.rs
```diff
@@ -5102,4 +5070,323 @@ mod test {
...
+        // state receives block 2 which is incoherent
+        let cp = env.next_proposal(
+            generate_full_tx_list(vec![txn_nonce_one]),
+            StateRootHash::default(),
+        );
+        let (author_2, _, proposal_message_2) = cp.destructure();
+        let block_2_id = proposal_message_2.block.get_id();
+
+        let cmds = n1.handle_proposal_message(author_2, proposal_message_2);
+        // shoult not vote for block 2
```
michael-yxchen: nit: typo

monad-eth-ledger/src/lib.rs
```diff
@@ -77,18 +77,24 @@ where
...
                                     .into_iter()
-                                    .map(|tx| (EthAddress(tx.signer()), tx.nonce() + 1))
+                                    // TODO: For any transfer transaction, update the balance of sender and receiver account
+                                    .map(|tx| {
+                                        (
+                                            EthAddress(tx.signer()),
+                                            EthAccountDelta::new(tx.nonce() + 1, 0, None),
+                                        )
+                                    })
                                     // collecting into a map will handle a sender sending multiple
```
michael-yxchen: What's the reason for tracking account balance in memory? From a test perspective, I think it's less complicated to set account balance at a block number explicitly, than relying on some simplified execution

monad-consensus-state/src/lib.rs
michael-yxchen, line 1: The tests are good. To be more accurate, we can use `[traced_test]` to pinpoint the reason why block is rejected. Specifically, assert that https://github.com/monad-crypto/monad-bft/blob/0cb78cf980be4c73cd78d95a6821bf0e7b6a1da8/monad-eth-block-policy/src/lib.rs#L665-L675 is in the logs

monad-eth-txpool/src/lib.rs
```diff
@@ -502,13 +507,62 @@ mod test {
         EthBlockPolicy::new(GENESIS_SEQ_NUM, u128::MAX, EXECUTION_DELAY, 0, 1337)
     }
 
+    pub trait TestPool {
+        fn insert_tx_unchecked(&mut self, txns: Vec<Bytes>) -> Vec<Bytes>;
+    }
```
michael-yxchen: Since test is in same file as EthTxPool, I think you can directly impl EthTxPool, without another trait

monad-eth-txpool/src/lib.rs
```diff
@@ -1275,4 +1418,119 @@ mod test {
         // transaction should not be inserted into the pool
         assert!(result.is_empty());
     }
+
+    #[test]
+    fn test_insufficient_reserve_balance_genesis_proposal() {
+        let sender_1_key = B256::random();
+        let txn_nonce_one = make_tx(sender_1_key, TRANSACTION_BASE_FEE, GAS_LIMIT, 0, 10);
+        let txn_nonce_two = make_tx(sender_1_key, TRANSACTION_BASE_FEE, GAS_LIMIT, 1, 10);
```
michael-yxchen: should `txn_nonce_one` be `txn_nonce_zero` and `two` be `one`?

monad-eth-ledger/src/lib.rs
bamrith: Yeah, this is not necessary for the tests in EthTxPool and consensus state. I added this so that the mock swarm tests we write are able to run without the user having to update the StateBackend after every block.

michael-yxchen: I think for memory backend to work in mockswarm, we need to refresh all account state for every block committed. Currently only accounts that have sent txns are refreshed. 

That said, I don't think this needs to be addressed in this PR. It may matter for https://github.com/monad-crypto/monad-bft/pull/1008

bamrith: The in memory backend clones the state from the previous block and refreshes the accounts that have sent the txns in the current block. Agreed that it's overkill, removed it.

The StateBackend now has a `set_account_state` to update the account balance as necessary in tests. Let me know if that is more understandable.

PR #992, opened 2024-08-13
monad-node: Switch over to use monad-raptorcast

buytenh:


monad-dataplane/src/event_loop.rs
```diff
@@ -230,7 +230,7 @@ impl DataplaneEventLoop {
             match self.ingress_sender.producer.push(rx) {
                 Err(e) => {
                     debug!("send failed on ingress sender {e}");
```
omegablitz, line 232: Should this be a warn?

buytenh: Perhaps.  I'll address this in a subsequent PR.

PR #1000, opened 2024-08-14
Add debug info to statesync

omegablitz:


monad-raptorcast/src/lib.rs
```diff
@@ -233,6 +233,10 @@ where
                 }
                 RouterCommand::Publish { target, message } => {
                     let app_message = message.serialize();
+                    tracing::debug!(
```
omegablitz: lower log level?

PR #1006, opened 2024-08-15
Transaction validation during tx insertion and block validation

eerkaijun:
Closes https://github.com/monad-crypto/monad-internal/issues/442

Currently there are still a few ways to crash the execution client. This commit adds transaction validation on the consensus client to make sure invalid transactions do not get included in a block. 

There's also currently a hardcoded `base_fee` parameter in the block header. Eventually, `base_fee` will need to be included in the consensus proposal


monad-eth-block-validator/Cargo.toml
```diff
@@ -15,4 +15,5 @@ monad-eth-tx = { path = "../monad-eth-tx" }
 monad-state-backend = { path = "../monad-state-backend" }
 
 alloy-rlp = { workspace = true }
+reth-primitives = { workspace = true }
```
ariqchowdhury: remove

monad-eth-block-validator/src/lib.rs
```diff
@@ -11,8 +11,49 @@ use monad_eth_block_policy::{EthBlockPolicy, EthValidatedBlock};
 use monad_eth_tx::{EthSignedTransaction, EthTransaction};
 use monad_eth_types::EthAddress;
 use monad_state_backend::StateBackend;
+use reth_primitives::TransactionSigned;
```
ariqchowdhury: remove this and use the `monad_eth_tx::EthSignedTransaction` type

monad-eth-block-validator/src/lib.rs
```diff
@@ -66,10 +107,15 @@ where
...
-                        .ok_or(BlockValidationError::TxnError)?;
+                    if static_validate_transaction(&eth_txn, self.chain_id).is_err() {
+                        return Err(BlockValidationError::TxnError);
+                    }
+
+                    // TODO(kai): currently block base fee is hardcoded to 1000 in monad-ledger
+                    // update this when base fee is included in consensus proposal
+                    if eth_txn.max_fee_per_gas() < 1000 {
+                        return Err(BlockValidationError::TxnError);
+                    }
```
ariqchowdhury, line 118: is there a reason this check is not included with the other static validate checks?

monad-eth-txpool/src/lib.rs
```diff
@@ -228,9 +229,7 @@ impl EthTxPool {
         let sender = EthAddress(eth_tx.signer());
         let txn_hash = eth_tx.hash();
 
-        // Do not add transaction with incorrect or missing chain id
-        let valid_chain_id = eth_tx.chain_id().is_some_and(|cid| cid == valid_chain_id);
-        if !valid_chain_id {
+        if static_validate_transaction(&eth_tx, valid_chain_id).is_err() {
             return Err(TxPoolInsertionError::NotWellFormed);
         }
```
ariqchowdhury, line 235: should the below gas_limit check be moved into the static validation?

ariqchowdhury:
also, forgot to mention, can we add flexnet tests that cover these invalid txns?

monad-eth-block-validator/src/lib.rs
eerkaijun: We want to also check these transactions during `insert_tx` in eth-txpool. Base fee however wouldn't be known during transaction insertion (will only be known during proposal creation), hence separating out the base fee check from the function

monad-eth-txpool/src/lib.rs
eerkaijun: I think we can remove the `gas_limit == 0` check altogether since we already check that gas limit must be larger than intrinsic gas in the static validation. Will remove it

monad-eth-block-validator/src/lib.rs
```diff
@@ -13,6 +13,46 @@ use monad_eth_types::EthAddress;
...
+    }
+
+    // EIP-3860
+    const DATA_SIZE_LIMIT: usize = 2 * 0x6000;
+    if tx.to().is_some() && tx.input().len() > DATA_SIZE_LIMIT {
+        return Err(TransactionError::InitCodeLimitExceeded);
+    }
+
+    // YP eq. 62
+    if tx.gas_limit() < 21000 {
```
ariqchowdhury: use named const please

PR #1012, opened 2024-08-16
Smarter raptorcast cache eviction

omegablitz:


monad-raptorcast/src/lib.rs
```diff
@@ -160,8 +170,9 @@ where
             current_epoch: Epoch(0),
             current_round: Round(0),
 
-            message_cache: Default::default(),
-            signature_cache: Default::default(),
+            pending_message_cache: LruCache::unbounded(),
```
ariqchowdhury, line 173: what's the reason this is unbounded?

omegablitz: https://github.com/monad-crypto/monad-bft/pull/1012/files#diff-a7a0c418338285b73aa56490b3f74e7fad73f870617ea93822f8cb78b2a6b7e9R382

All evictions from pending_message_cache are logged, so that unfinished message decoding is noisy.

omegablitz: Happy path: https://github.com/monad-crypto/monad-bft/pull/1012/files#diff-a7a0c418338285b73aa56490b3f74e7fad73f870617ea93822f8cb78b2a6b7e9R496

PR #1013, opened 2024-08-16
Fix CLI for flexnet and Jenkins

tongzhi1998:


michael-yxchen:
This change looks good. Is the CI working for you after this change? Or is it still breaking, possibly due to other issues?

tongzhi1998:
> This change looks good. Is the CI working for you after this change? Or is it still breaking, possibly due to other issues?

Sorry just saw this comment. It's still breaking. I replied in the consensus thread that we are on. I think we might need to force rebuild or something but I am not sure.

PR #1014, opened 2024-08-17
Intrinsic gas calculation

eerkaijun:
I realized I made a mistake in my previous PR https://github.com/monad-crypto/monad-bft/pull/1006 when defining the intrinsic gas, my bad, apologies. Made the fixes in this PR - this is now consistent with our execution client. Tested manually to make sure intrinsic gas calculation is the same as execution client

### DIFF
--- a/monad-eth-block-policy/src/lib.rs
+++ b/monad-eth-block-policy/src/lib.rs
@@ -8,7 +8,7 @@ use monad_consensus_types::{
     state_root_hash::StateRootHash,
 };
 use monad_crypto::hasher::{Hashable, Hasher};
-use monad_eth_tx::{EthTransaction, EthTxHash};
+use monad_eth_tx::{EthSignedTransaction, EthTransaction, EthTxHash};
 use monad_eth_types::{Balance, EthAddress, Nonce};
 use monad_state_backend::{StateBackend, StateBackendError};
 use monad_types::{BlockId, Epoch, NodeId, Round, SeqNum, GENESIS_SEQ_NUM};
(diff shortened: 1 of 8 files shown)


ariqchowdhury:
I think these changes need to be coordinated with the reserve balance carriage costs
cc: @mikety

monad-eth-block-validator/src/lib.rs
```diff
@@ -22,6 +22,40 @@ pub enum TransactionError {
...
+        .unwrap_or(&[]);
+    let accessed_slots: usize = access_list.iter().map(|item| item.storage_keys.len()).sum();
+    // each address in access list costs 2400 gas
+    initial_gas += access_list.len() as u64 * 2400;
+    // each storage key in access list costs 1900 gas
+    initial_gas += accessed_slots as u64 * 1900;
+
+    initial_gas
+}
+
```
mikety: This should be moved to https://github.com/monad-crypto/monad-bft/blob/ab1dcdf38f858c534892ed2b66fdc2208e373a98/monad-eth-block-policy/src/lib.rs#L28.

eerkaijun: aaaa cool, didn't realize there is an existing function. will move

eerkaijun: done

monad-eth-block-policy/src/lib.rs
```diff
@@ -25,15 +25,85 @@ pub enum ReserveBalanceCheck {
...
+    // EIP-2930
+    let access_list = tx
+        .access_list()
+        .map(|list| list.0.as_slice())
+        .unwrap_or(&[]);
+    let accessed_slots: usize = access_list.iter().map(|item| item.storage_keys.len()).sum();
+    // each address in access list costs 2400 gas
+    initial_gas += access_list.len() as u64 * 2400;
+    // each storage key in access list costs 1900 gas
+    initial_gas += accessed_slots as u64 * 1900;
```
mikety: nit: the initial_gas name seems to mean initial 21,000 , changing variable name to intrinsic_gas will be more consistent naming.

eerkaijun: done

PR #1015, opened 2024-08-17
Update triedb decode

eerkaijun:
The latest execution update seems to have a few triedb schema change that breaks things. https://github.com/monad-crypto/monad/pull/771

Updating the decoding in bft


ariqchowdhury:
can you describe where this was tested? Flexnet rpc tests?

eerkaijun:
The python integration test where we check against Anvil RPC (which is en route to be migrated to flexnet)

PR #1022, opened 2024-08-21
Additional accounts at genesis

philipglazman:
Onboarded validators can use these additional accounts until we have a faucet


philipglazman:
@omegablitz what is an easy way to identify the new `state_root` ?

eerkaijun:
Btw I think the change might be more suited at https://github.com/monad-crypto/monad-infra/blob/main/ansible/chain/genesis.json which is what we use for the devnet boxes. 

The `genesis.json` file in this repo is only used for starting a local single node so might not need these addresses

PR #1027, opened 2024-08-22
Log -> tracing

omegablitz:


omegablitz:
They probably should be removed; will batch that with a future change (maybe @buytenh can do that together with TCP stuff?)

buytenh:
> They probably should be removed; will batch that with a future change (maybe @buytenh can do that together with TCP stuff?)

I folded these into https://github.com/monad-crypto/monad-bft/pull/1076

PR #1036, opened 2024-08-27
Block validation makes sure no nonce gap

eerkaijun:
I _think_ in block validation, we also want to make sure that there is no nonce gap between transactions, otherwise a malicious proposer can create a block with nonce gap which passes consensus validation but will crash execution


eerkaijun:
> We should also make sure our EthTxPool and EthValidator are enforcing the same set of rules.

yeap I think this is the only discrepancy.

PR #1045, opened 2024-08-29
Incorporate execution statesync changes

omegablitz:


omegablitz:
Tested on the statesync boxes

omegablitz:
Note to self: wait for monad-crypto/monad#848 to hit master before merging

monad-proto/proto/message.proto
```diff
@@ -66,10 +66,17 @@ message ProtoStateSyncRequest {
...
 }
 
+enum ProtoStateSyncUpsertType {
+  CODE = 0;
+  ACCOUNT = 1;
+  STORAGE = 2;
+  ACCOUNT_DELETE = 3;
+  STORAGE_DELETE = 4;
+}
+
```
michael-yxchen, line 76: protobuf recommends reserving 0 for an unspecified value. It's used as the default value when deserializing an unknown enum.

ref: https://protobuf.dev/programming-guides/dos-donts/#unspecified-enum

omegablitz: You're right, but I'm lazy and will fix this later when we add versioning support for state-sync messages

PR #1047, opened 2024-08-30
monad-raptorcast: Reorder post-proof header fields to improve alignment

buytenh:
Without the changes in this PR, the part of the RaptorCast header
after the Merkle proof section looks like this:

```
             0        4        8        c
    0x0000:  xx                                   chunk_merkle_leaf_idx
               xxxxxx xxxxxxxx xxxxxxxx xxxxxxxx  chunk_recipient
    0x0010:  xxxxxxxx xx
                        xxxx                      chunk_id
                            xx xx                 chunk_len
                                 xxxxxx xx......  data
```

Note the lack of proper alignment of several of the fields, and how the
data field has a length that is congruent to 3 mod 4.

This PR moves the chunk_merkle_leaf_idx byte to be after the
chunk_recipient field, and removes the chunk_len field (since we can
infer the chunk length from the length of the packet) and replaces it
with a single padding byte.  This makes the layout look like this:

```
             0        4        8        c
    0x0000:  xxxxxxxx xxxxxxxx xxxxxxxx xxxxxxxx  chunk_recipient
    0x0010:  xxxxxxxx
                      xx                          chunk_merkle_leaf_idx
                        xx                        reserved
                          xxxx                    chunk_id
                               xxxxxxxx ........  data
```

This looks a bit better, and this way, all fields are naturally aligned,
and the data field is a multiple of 4 bytes long.

### DIFF
--- a/monad-raptorcast/src/udp.rs
+++ b/monad-raptorcast/src/udp.rs
@@ -202,13 +202,13 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
 /// - 20 bytes * (merkle_tree_depth - 1) => merkle proof (leaves include everything that follows,
 ///   eg hash(chunk_recipient + chunk_byte_offset + chunk_len + payload))
 ///
-/// - 1 byte => Chunk's merkle leaf idx
 /// - 20 bytes => first 20 bytes of hash of chunk's first hop recipient
 ///   - we set this even if broadcast bit is not set so that it's known if a message was intended
 ///     to be sent to self
+/// - 1 byte => Chunk's merkle leaf idx
+/// - 1 byte => reserved
 /// - 2 bytes (u16) => This chunk's id
-/// - 2 bytes => Merkle chunk payload len
-/// - (merkle_chunk_payload_len bytes) => data
+/// - rest => data
 ///
 //
 //
@@ -225,10 +225,10 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
 //
 //     merkle_proof: Vec<[u8; 20]>,
 //
-//     chunk_merkle_leaf_idx: u8,
 //     chunk_recipient: [u8; 20],
+//     chunk_merkle_leaf_idx: u8,
+//     reserved: u8,
 //     chunk_id: u16,
-//     chunk_len: u16,
 //
 //     data: Bytes,
 // }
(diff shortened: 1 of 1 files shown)


monad-raptorcast/src/udp.rs
```diff
@@ -228,7 +228,6 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
 //     chunk_merkle_leaf_idx: u8,
```
omegablitz: Forgot to move this?

monad-raptorcast/src/udp.rs
```diff
@@ -620,32 +617,23 @@ where
...
-    let cursor_payload_len = split_off(2)?;
-    let payload_len = u16::from_le_bytes(
-        cursor_payload_len
-            .as_ref()
-            .try_into()
-            .expect("u16 is 2 bytes"),
-    );
-    if payload_len == 0 {
-        // handle the degenerate case
-        return Err(MessageValidationError::TooShort);
```
omegablitz, line 645: Is this condition handled somewhere else now?Just checking that a zero payload won't break; specifically worried about this:
https://github.com/monad-crypto/monad-bft/blob/316f00f46a8ae739b3edbaf6cf8219140245b480/monad-raptorcast/src/udp.rs#L144-L146

monad-raptorcast/src/udp.rs
buytenh: Thanks, fixed.

monad-raptorcast/src/udp.rs
buytenh: Good catch -- I added back the length check:
```
     let cursor_payload = cursor;
+    if cursor_payload.is_empty() {
+        // handle the degenerate case
+        return Err(MessageValidationError::TooShort);
+    }
```

monad-raptorcast/src/udp.rs
```diff
@@ -225,10 +225,9 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
 //
 //     merkle_proof: Vec<[u8; 20]>,
 //
-//     chunk_merkle_leaf_idx: u8,
 //     chunk_recipient: [u8; 20],
+//     chunk_merkle_leaf_idx: u8,
```
omegablitz, line 229: one last nit: could you also add the `reserved: u8` here for completeness?

buytenh: done

PR #1051, opened 2024-08-30
Optional traces and lower log levels

omegablitz:


omegablitz:
Yep, they were happening regularly

PR #1053, opened 2024-08-31
Statesync progress estimate metric

omegablitz:


monad-statesync/src/ffi.rs
```diff
@@ -171,12 +181,39 @@ impl<PT: PubKey> StateSync<PT> {
...
+                .as_ref()
+                .map_or(response.request.old_target * num_prefixes, |progress| {
+                    progress.current_progress
+                });
+            self.progress = Some(Progress {
+                current_progress: old_progress + response_progress,
+                total_progress,
+                num_prefixes,
+            });
+
```
ariqchowdhury: nit: I think its more clear when small blocks of code like this are in helper functions, keeps the overall handle_* functions more clear

PR #1055, opened 2024-08-31
Move statesync client out of main thread

omegablitz:


ariqchowdhury:
is this for handling requests from other nodes or making requests or both?

omegablitz:
This is for making requests

PR #1060, opened 2024-09-03
Move long-running triedb job to dedicated thread

omegablitz:
The triedb state root hash poller is currently running on a long-running rayon task, which permanently eats up a thread from the Rayon pool. The tredb state root hash poller is IO (and not CPU) bound.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4509,7 +4509,6 @@ dependencies = [
  "ntest",
  "rand",
  "rand_chacha",
- "rayon",
  "serde_json",
  "sorted-vec",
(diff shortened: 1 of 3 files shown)


zander-xyz:
unclear if related but we're trying to move towards not having that polling thread for the readonly db in bft. thats already configurable today

omegablitz:
> unclear if related but we're trying to move towards not having that polling thread for the readonly db in bft. thats already configurable today

Discussed offline, this is unrelated

omegablitz:
Will revisit later once we clean up our thread usages

PR #1061, opened 2024-09-04
Switch RPC to use async triedb

eerkaijun:
wip - opening a draft pr for easier discussion later


monad-triedb/src/lib.rs
```diff
@@ -24,6 +24,9 @@ pub struct Handle {
     db_ptr: *mut bindings::triedb,
 }
 
+unsafe impl Send for Handle {}
+unsafe impl Sync for Handle {}
```
bamrith: I don't think this is right. Async reads should only be initiated and polled from the thread that constructs the handle. Also, we shouldn't pass around references of `Handle` between threads.

eerkaijun: Yeah, this was from a previous implementation. I forgot to remove it

monad-rpc/src/triedb.rs
```diff
@@ -22,9 +42,84 @@ pub enum TriedbResult {
...
+        // Check polling status
+        if completed_counter.load(SeqCst) >= 1 {
+            break;
+        }
+
+        // poll the triedb handle periodically
+        triedb_handle.triedb_poll();
+
+        // yield to allow other tasks to run
+        tokio::task::yield_now().await;
```
bamrith: Not sure if this would affect performance since all tasks are yielding to each other just to poll the same handle. You could try initiating reads and polling from `polling_thread` so there's no overhead, and have these sub-tasks just await on the receiver to see if it makes a difference.

eerkaijun: switched to poll at regular intervals instead so we can have larger batch

eerkaijun:
Overview: 
- Use of mpsc channel to send rpc requests to a single dedicated thread used for async triedb reads
- Use non-blocking triedb poll at intervals (interval length will determine batch size of triedb reads)

Some benchmarking data here: https://www.notion.so/monad-labs/RPC-Async-Triedb-Benchmark-cb8814d53230460bad2ce8627690f635

To follow up in a subsequent PR: 
- Now that we switched triedb reads to async, some of the helper functions in `monad-triedb-utils` which perform blocking reads are not used anymore. To clean them up and consolidate the helper functions
- See if it's possible to pass a waker to the `read_async` callback instead of using message passing

monad-rpc/src/main.rs
```diff
@@ -640,11 +640,6 @@ async fn main() -> std::io::Result<()> {
         }
     });
 
-    rayon::ThreadPoolBuilder::new()
```
omegablitz, line 666: Still using rayon in blockdb-utils

monad-rpc/src/triedb.rs
```diff
@@ -22,69 +55,179 @@ pub enum TriedbResult {
     BlockNum(u64),
 }
 
+async fn polling_thread(triedb_path: PathBuf, mut receiver: mpsc::Receiver<TriedbRequest>) {
```
omegablitz: This should shouldn't be an async function, and it should take in an [std::sync::mpsc::Receiver](https://doc.rust-lang.org/nightly/std/sync/mpsc/index.html) instead of a tokio::mpsc::Receiver. We don't need a tokio runtime for polling triedb.

monad-rpc/src/triedb.rs
```diff
@@ -22,69 +55,179 @@ pub enum TriedbResult {
...
+    // create a new triedb handle for the polling thread
+    let triedb_handle: Rc<TriedbHandle> =
+        Rc::new(TriedbHandle::try_new(&triedb_path).expect("triedb should exist in path"));
+
+    // create an interval that ticks every 200 microseconds
+    let mut interval = tokio::time::interval(tokio::time::Duration::from_micros(200));
+
+    loop {
+        tokio::select! {
```
omegablitz: Remove `tokio::select!`. This should spin on [mpsc::Receiver::try_recv](https://doc.rust-lang.org/nightly/std/sync/mpsc/struct.Receiver.html#method.try_recv) and triedb_poll(false, usize::MAX)

monad-rpc/src/triedb.rs
```diff
@@ -22,69 +55,179 @@ pub enum TriedbResult {
...
+                            process_request(triedb_handle_clone, async_request).await;
+                        });
+                    }
+                }
+            }
+            _ = interval.tick() => {
+                // call triedb_poll periodically
+                triedb_handle.triedb_poll(false, usize::MAX);
+            }
+            else => break,
```
omegablitz: This function _should_ never exit, correct? Switch to `unreachable!` if so.

monad-rpc/src/triedb.rs
```diff
@@ -22,69 +55,179 @@ pub enum TriedbResult {
...
+        BlockTags::Default(t) => match t {
+            BlockTagKey::Latest => triedb_handle.latest_block(),
+            BlockTagKey::Finalized => triedb_handle.latest_block(),
+        },
+    };
+
+    // counter which is updated when TrieDB processes a single async read to completion
+    let completed_counter = Arc::new(AtomicUsize::new(0));
+
+    let triedb_receiver = triedb_handle.read_async(
```
omegablitz: It makes more sense to modify read_async to pass in the `oneshot::Sender` so that you don't need to spawn the tokio task below

monad-rpc/src/triedb.rs
```diff
@@ -22,69 +55,179 @@ pub enum TriedbResult {
...
+        // create a mpsc channel where sender are incoming requests, and the receiver is the triedb poller
+        let (sender, receiver) = mpsc::channel::<TriedbRequest>(1024);
+
+        // spawn the polling thread in a dedicated thread
+        let triedb_path_cloned = triedb_path.to_path_buf();
+        std::thread::spawn(move || {
+            let local_set = tokio::task::LocalSet::new();
+
+            // create a new runtime to block on the LocalSet in this thread
+            let rt = tokio::runtime::Runtime::new().unwrap();
```
omegablitz: Triedb polling thread shouldn't use tokio.

monad-rpc/src/triedb.rs
```diff
@@ -22,69 +55,179 @@ pub enum TriedbResult {
...
+        let (request_sender, request_receiver) = oneshot::channel();
+
+        // clone a sender from mpsc
+        if let Err(e) = self
+            .mpsc_sender
+            .clone()
+            .send(TriedbRequest::BlockNumberRequest(BlockNumberRequest {
+                request_sender,
+            }))
+            .await
```
omegablitz: Switch to [try_send](https://doc.rust-lang.org/nightly/std/sync/mpsc/struct.SyncSender.html#method.try_send), and log err + drop request if channel is full.

monad-rpc/src/triedb.rs
```diff
@@ -22,69 +55,179 @@ pub enum TriedbResult {
...
-        });
-        recv.await.expect("rayon panic get_account")
+        // create a one shot channel to retrieve the triedb result from the polling thread
+        let (request_sender, request_receiver) = oneshot::channel();
+
+        let (triedb_key, key_len_nibbles) = create_addr_key(&addr);
+
+        // clone a sender from mpsc
+        if let Err(e) = self
+            .mpsc_sender
```
omegablitz, line 180: Same as above

monad-rpc/src/triedb.rs
```diff
@@ -93,81 +236,120 @@ impl TriedbEnv {
...
+        // create a one shot channel to retrieve the triedb result from the polling thread
+        let (request_sender, request_receiver) = oneshot::channel();
+
+        let (triedb_key, key_len_nibbles) = create_storage_at_key(&addr, &at);
+
+        // clone a sender from mpsc
+        if let Err(e) = self
+            .mpsc_sender
+            .clone()
+            .send(TriedbRequest::AsyncRequest(AsyncRequest {
```
omegablitz: Same as above

monad-rpc/src/triedb.rs
```diff
@@ -93,81 +236,120 @@ impl TriedbEnv {
...
+        // create a one shot channel to retrieve the triedb result from the polling thread
+        let (request_sender, request_receiver) = oneshot::channel();
+
+        let (triedb_key, key_len_nibbles) = create_code_key(&code_hash);
+
+        // clone a sender from mpsc
+        if let Err(e) = self
+            .mpsc_sender
+            .clone()
+            .send(TriedbRequest::AsyncRequest(AsyncRequest {
```
omegablitz: Same as above

monad-rpc/src/triedb.rs
```diff
@@ -93,81 +236,120 @@ impl TriedbEnv {
...
+        // create a one shot channel to retrieve the triedb result from the polling thread
+        let (request_sender, request_receiver) = oneshot::channel();
+
+        let (triedb_key, key_len_nibbles) = create_receipt_key(txn_index);
+
+        // clone a sender from mpsc
+        if let Err(e) = self
+            .mpsc_sender
+            .clone()
+            .send(TriedbRequest::AsyncRequest(AsyncRequest {
```
omegablitz: Same as above

monad-rpc/src/triedb.rs
eerkaijun: what's the main motivation for not using Tokio mpsc? Wouldn't non-blocking messages be desirable?

omegablitz: Tokio is an async runtime that's a wrapper around epoll. This thread's events should be driven by `triedb_poll()` and not an epoll loop.

Stuff will get communicated back to the rpc server's tokio runtime via a [tokio::sync::oneshot::Receiver](https://docs.rs/tokio/latest/tokio/sync/oneshot/struct.Receiver.html) that's sent to by the triedb poller thread.

monad-rpc/src/triedb.rs
eerkaijun: removed

monad-rpc/src/triedb.rs
eerkaijun: gonna start consolidating some triedb helper functions in a followup PR: https://github.com/monad-crypto/monad-bft/pull/1110 perhaps we can use future and waker and remove the one shot sender altogether

monad-rpc/src/triedb.rs
```diff
@@ -22,69 +60,171 @@ pub enum TriedbResult {
...
+    // Parse block tag
+    let block_num = match async_request.block_tag {
+        BlockTags::Number(q) => q,
+        BlockTags::Default(t) => match t {
+            BlockTagKey::Latest => triedb_handle.latest_block(),
+            BlockTagKey::Finalized => triedb_handle.latest_block(),
+        },
+    };
+
+    let future_result = triedb_handle.read_async(
```
omegablitz: You'll want to pass the oneshot::Sender in directly here.

Concretely, pass the oneshot::Sender in an arg to this function, and pass that Sender to the `SenderContext` instead of creating a redundant oneshot channel inside.
https://github.com/monad-crypto/monad-bft/blob/cf77d38520a7e61c21b1bc0c550e6bd0ee7bd7c0/monad-triedb/src/lib.rs#L123-L129.

You'll also need to relax this assertion:
https://github.com/monad-crypto/monad-bft/blob/cf77d38520a7e61c21b1bc0c550e6bd0ee7bd7c0/monad-triedb/src/lib.rs#L60

monad-rpc/src/triedb.rs
```diff
@@ -1,19 +1,57 @@
...
+struct BlockNumberRequest {
+    // a sender to send the block number back to the request handler
+    request_sender: oneshot::Sender<u64>,
+}
+
+// struct that is sent from the request handler to the polling thread
+struct AsyncRequest {
+    // a sender for the polling thread to send the result back to the request handler
+    // after polling is completed
+    request_sender: oneshot::Sender<Receiver<Option<Vec<u8>>>>,
```
omegablitz: This should just be a `oneshot::Sender<Option<Vec<u8>>`

monad-rpc/src/triedb.rs
```diff
@@ -22,69 +60,171 @@ pub enum TriedbResult {
...
+    // create a new triedb handle for the polling thread
+    let triedb_handle: Rc<TriedbHandle> =
+        Rc::new(TriedbHandle::try_new(&triedb_path).expect("triedb should exist in path"));
+
+    // create a polling interval of 200 microseconds
+    let poll_interval = Duration::from_micros(200);
+    let mut last_poll = Instant::now();
+
+    loop {
+        if last_poll.elapsed() >= poll_interval {
```
omegablitz, line 81: Should either be busy-polling `Handle::triedb_poll`, or sleeping. Doesn't make sense to spin on `last_poll.elapsed()`.

omegablitz: https://github.com/monad-crypto/monad/blob/89ef0dfe7964274fe358a62dbe473faa216a3e21/libs/async/src/monad/async/io.hpp#L245-L256

Probably simplest to just busy-poll here, and not worry about any of the `last_poll` stuff

omegablitz: Nvm, discussed with Niall, I understand the async triedb interface better.

You'll want the loop to look something like this:
```rust
// read-only triedb thread
loop {
    if let Ok(request) = request_channel.recv_timeout(Duration::from_millis(1)) {
        triedb_handle.triedb_async_read(...)
    }
    triedb_handle.poll(blocking=false)
}
```

omegablitz: The tokio version would look like this:
```rust
// read-only triedb task
loop {
    let mut interval = time::interval(Duration::from_millis(1));
    tokio::select! {
        Ok(request) = request_channel.recv() => {
            triedb_handle.triedb_async_read(...)
        }
        _ = interval.tick() => {
            triedb_handle.poll(blocking=false)
        }        
    }
}
```

This is similar to your original code, except that for this to be useful, we'd want this to run within our _original_ tokio runtime, and not a new one spawned just for this. I think this is more annoying to implement, so let's stick with the std::thread version for now.

monad-rpc/src/triedb.rs
eerkaijun: I have a commit in a followup refactor PR https://github.com/monad-crypto/monad-bft/pull/1110/commits/f7e1dae4b00dc5dbc6d81c3fae97ad088b5eceba that does this, but I'll cherry-pick it over

monad-rpc/src/triedb.rs
eerkaijun: Will cherry pick it over as per above comment

monad-rpc/src/triedb.rs
eerkaijun: > ```rust
> // read-only triedb thread
> loop {
>     if let Ok(request) = request_channel.recv_timeout(Duration::from_millis(1)) {
>         triedb_handle.triedb_async_read(...)
>     }
>     triedb_handle.poll(blocking=false)
> }
> ```

This does one poll per request though, which impact performance. Benchmark performance is better when we batch requests and send them to `async_read` before polling once, which is the reason for only polling once at each time interval.

omegablitz: Ok, we can revisit this later

monad-rpc/src/triedb.rs
```diff
@@ -20,78 +69,161 @@ pub enum TriedbResult {
...
 
-#[derive(Clone, Debug, PartialEq, Eq)]
-pub enum BlockTags {
-    Number(u64),
-    Latest,
-}
+fn polling_thread(triedb_path: PathBuf, receiver: mpsc::Receiver<TriedbRequest>) {
+    // create a new triedb handle for the polling thread
+    let triedb_handle: Rc<TriedbHandle> =
+        Rc::new(TriedbHandle::try_new(&triedb_path).expect("triedb should exist in path"));
```
omegablitz: Does this need to be in an Rc?

omegablitz:
Btw, you can remove rayon from monad-rpc now after #1125

monad-rpc/src/triedb.rs
eerkaijun: good catch, removed it

PR #1064, opened 2024-09-05
Initcode limit bugfix

eerkaijun:


monad-eth-block-policy/src/lib.rs
```diff
@@ -94,7 +94,7 @@ pub fn static_validate_transaction(
 
     // EIP-3860
     const DATA_SIZE_LIMIT: usize = 2 * 0x6000;
-    if tx.to().is_some() && tx.input().len() > DATA_SIZE_LIMIT {
+    if tx.to().is_none() && tx.input().len() > DATA_SIZE_LIMIT {
```
ariqchowdhury: maybe `let is_contract_creation = tx.to().is_none()` so that we can write `if is_contract_create && ...`
will help readability

monad-eth-block-validator/src/lib.rs
```diff
@@ -260,4 +261,42 @@ mod test {
         let result = block_validator.validate_payload(&payload);
         assert!(matches!(result, Err(BlockValidationError::TxnError)));
     }
+
+    #[test]
+    fn test_invalid_block_over_initcode_limit() {
```
michael-yxchen: This can be a unit test on `static_validate_transaction` as well. There, you can assert on specific validation error there

monad-eth-block-policy/src/lib.rs
eerkaijun: realized there's an existing function for that `tx.kind().is_create()`

monad-eth-block-validator/src/lib.rs
eerkaijun: great point, moved the test to get more granular error

PR #1070, opened 2024-09-06
Strip symbols

omegablitz:


kkuehlz:
@omegablitz should we update execution to the version with the new release dockerfile?

PR #1071, opened 2024-09-09
Add compression to raptor broadcast

rgarc:
This PR enables no-op compression support on the raptor broadcast router so it shouldn't break anything.


rgarc:
Closing because we don't want to compression at consensus state level instead of the transport layer.

PR #1074, opened 2024-09-09
Fix confusing monad-raptor debug tracing output

buytenh:
When running the monad-raptorcast service example with `RUST_LOG=debug`,
we get a debug message for every encoded symbol received that has its
buffer index congruent to 99 mod 100:

```
    Finished `release` profile [optimized + debuginfo] target(s) in 4.38s
     Running `target/release/examples/service --addresses '127.0.0.1:10000' '127.0.0.1:10001' '127.0.0.1:10002' '127.0.0.1:10003'`
2024-09-09T22:37:56.345550Z DEBUG monad_raptor::r10::nonsystematic::decoder::receive_symbol: received_encoded_symbol buffer_index=199
2024-09-09T22:37:56.346791Z DEBUG monad_raptor::r10::nonsystematic::decoder::receive_symbol: received_encoded_symbol buffer_index=299
2024-09-09T22:37:56.348231Z DEBUG monad_raptor::r10::nonsystematic::decoder::receive_symbol: received_encoded_symbol buffer_index=399
2024-09-09T22:37:56.349277Z DEBUG monad_raptor::r10::nonsystematic::decoder::receive_symbol: received_encoded_symbol buffer_index=499
2024-09-09T22:37:56.350447Z DEBUG monad_raptor::r10::nonsystematic::decoder::receive_symbol: received_encoded_symbol buffer_index=599
[...]
```

But notice how `buffer_index=99` is missing from the debug output.  Is
the initial burst of encoded symbols getting lost somewhere?

Changing the debug message to trace for every received buffer seems to
suggest that we are indeed losing the first burst of encoded symbols:

```
    Finished `release` profile [optimized + debuginfo] target(s) in 0.22s
     Running `target/release/examples/service --addresses '127.0.0.1:10000' '127.0.0.1:10001'`
2024-09-09T22:41:53.929164Z DEBUG monad_raptor::r10::nonsystematic::decoder::receive_symbol: received_encoded_symbol buffer_index=141
2024-09-09T22:41:53.929210Z DEBUG monad_raptor::r10::nonsystematic::decoder::receive_symbol: received_encoded_symbol buffer_index=142
2024-09-09T22:41:53.929218Z DEBUG monad_raptor::r10::nonsystematic::decoder::receive_symbol: received_encoded_symbol buffer_index=143
2024-09-09T22:41:53.929223Z DEBUG monad_raptor::r10::nonsystematic::decoder::receive_symbol: received_encoded_symbol buffer_index=144
[...]
```

Is still there a packet loss issue when transmitting over the loopback
interface, perhaps caused by a pacing bug?  Or is there a bug where we
fail to transmit the first few superframes of a Raptor-encoded message
in monad-dataplane?

As it turns out, we aren't actually losing any packets/symbols, and the
tracing output is confusing because the decoder allocates the first set
of decoding buffers, 141 buffers in this case, to the constraint symbols
corresponding to the LDPC/Half pre-code symbols, and therefore, the first
symbol received from the network will land in buffer 141 and not in
buffer 0.

Make the tracing output less confusing by taking the number of constraint
symbols into account and subtracting that from the buffer index that is
being logged.

Also change the tracing to log the number of received buffers including
the current one, and to trace for all multiples of 100 buffers, which
will thus not include a trace for 0.

The output will now look like this:

```
    Finished `release` profile [optimized + debuginfo] target(s) in 4.39s
     Running `target/release/examples/service --addresses '127.0.0.1:10000' '127.0.0.1:10001'`
2024-09-09T22:50:58.867969Z DEBUG monad_raptor::r10::nonsystematic::decoder::receive_symbol: received_encoded_symbol num_buffers_received=100
2024-09-09T22:50:58.868472Z DEBUG monad_raptor::r10::nonsystematic::decoder::receive_symbol: received_encoded_symbol num_buffers_received=200
2024-09-09T22:50:58.869131Z DEBUG monad_raptor::r10::nonsystematic::decoder::receive_symbol: received_encoded_symbol num_buffers_received=300
2024-09-09T22:50:58.870911Z DEBUG monad_raptor::r10::nonsystematic::decoder::receive_symbol: received_encoded_symbol num_buffers_received=400
[...]
```

### DIFF
--- a/monad-raptor/src/r10/nonsystematic/decoder/receive_symbol.rs
+++ b/monad-raptor/src/r10/nonsystematic/decoder/receive_symbol.rs
@@ -11,11 +11,16 @@ impl Decoder {
         encoding_symbol_id: usize,
         mut xor_buffers: impl FnMut(BufferId, BufferId),
     ) {
-        let buffer_index: u16 = self.buffer_state.len().try_into().unwrap();
+        {
+            let num_buffers_received =
+                self.buffer_state.len() - self.num_redundant_intermediate_symbols() + 1;
 
-        if (buffer_index % 100) == 99 {
-            tracing::debug!(?buffer_index, "received_encoded_symbol");
+            if (num_buffers_received % 100) == 0 {
+                tracing::debug!(?num_buffers_received, "received_encoded_symbol");
+            }
         }
+
+        let buffer_index: u16 = self.buffer_state.len().try_into().unwrap();
 
         let mut buffer = Buffer::new();
 


monad-raptor/src/r10/nonsystematic/decoder/receive_symbol.rs
```diff
@@ -11,12 +11,17 @@ impl Decoder {
...
+                self.buffer_state.len() - self.num_redundant_intermediate_symbols() + 1;
 
-        if (buffer_index % 100) == 99 {
-            tracing::debug!(?buffer_index, "received_encoded_symbol");
+            if (num_buffers_received % 100) == 0 {
+                tracing::debug!(?num_buffers_received, "received_encoded_symbol");
+            }
         }
 
+        let buffer_index: u16 = self.buffer_state.len().try_into().unwrap();
```
ariqchowdhury, line 23: is buffer_state's len guaranteed to be less than 2^16? Can someone keep sending stuff so that it grows beyond that?

buytenh: (Note that that line was here before, and that this PR doesn't change it, although the diff makes it look like it does.)

Yes, that is one of the remaining vulnerabilities in the decoder -- a malicious proposer can send the same encoded symbol repeatedly or choose a pathological combination of encoded symbols to transmit, either of which may induce excessive CPU usage in the decoder, and it can send more than 2^16 symbols, which would cause the decoder to panic.

(It also may or may not be possible for an on-path attacker to replay the same symbol repeatedly, which would cause similar effects in the decoder -- I need to take another look at the RaptorCast receive path to decide whether or not that is possible.)

I have patches to address this in two different ways:

- Maintain a bitmap of encoded symbol IDs we have already received, and drop received symbols that we have seen before.

  (We need to do this even if RaptorCast protects against replays, as a malicious proposer could send the same encoded symbol over and over in different chunks with different message headers, such that this wouldn't be prevented by replay protection.)

- Reject all received encoded symbol IDs that have indices that are >= 7 * num_source_symbols.

  As the maximum number of R10 symbols in a message is 8192, and the maximum number of different encoded fountain symbols is 65521, it doesn't make sense to allow for a sender redundancy level higher than 65521 / 8192 = 7, and restricting the number of encoded symbols we will accept for a given application message length bounds the amount of extra work a malicious proposer or an on-path attacker might cause us to perform.

Even with these two countermeasures, we'd probably have to sit down with some of the people in the Research team to think about whether it might still be possible for a malicious proposer to transmit carefully chosen sets of encoded symbols that might induce pathological behavior in the decoder.

I'll write up a Notion page about the known and theorized vulnerabilities in our current RaptorCast receive path and R10 decoder combination, as we discussed earlier today.

omegablitz: We could also just hard cap the size of buffer_state - if we receive _x_ times the original number of symbols and still can't decode the original, we should be able to safely assume that the broadcaster is behaving pathologically.

buytenh: > We could also just hard cap the size of buffer_state - if we receive _x_ times the original number of symbols and still can't decode the original, we should be able to safely assume that the broadcaster is behaving pathologically.

I think that if we do that but don't cap the range of ESIs that we accept from a peer, the peer can carefully select and send us a large number of encoded symbols that have many linear dependencies, such that the set is not decodable, but such that we'll spend a lot of time trying to decode it.

https://github.com/monad-crypto/monad-bft/pull/1104 implements the points from my previous comment, and I believe that it sufficiently deals with the issue that Ariq referred to, by not allowing the same encoded symbol to be sent to the decoder twice, and to cap the maximum ESI we accept from a peer to a multiple (currently 7) of the number of source symbols in the message.

PR #1078, opened 2024-09-10
Switch away from using a const generic symbol length in the R10 encoder

buytenh:
The R10 encoder as originally implemented and imported into the monad-bft
repository required the encoding symbol length to be a const generic.

This makes the assembly code generated for the buffer XORing functions
look clean and optimal, but unfortunately, it is too restrictive for
real-world use, as the payload (symbol) portion of RaptorCast packets
doesn't have a fixed size known at compile time, mainly for the reason
that the RaptorCast header length depends on the chosen Merkle tree depth
for this message, and the Merkle tree depth isn't necessarily fixed.

To be able to use the current version of the R10 encoder in RaptorCast,
we've temporarily fixed the Merkle tree depth for all outgoing messages
to 6 when we imported the encoder, and we would really like to lift
this restriction again.

This patch changes the encoder to make the symbol length a run-time
determinable variable, and adapts all users of the encoder to match.
A follow-up patch will restore the previous RaptorCast logic to
determine the Merkle tree depth at run time based on the packet type.

### DIFF
--- a/monad-raptor/src/r10/nonsystematic/encoder.rs
+++ b/monad-raptor/src/r10/nonsystematic/encoder.rs
@@ -14,58 +14,59 @@ use crate::{
 };
 
 #[derive(Debug)]
-pub struct Encoder<'a, const SYMBOL_LEN: usize> {
-    // Reference to the provided source data.
+pub struct Encoder<'a> {
+    // Reference to the provided source data, and the requested symbol length.
     src: &'a [u8],
+    symbol_len: usize,
 
-    // R10 parameters for K = ceil(src.len() / SYMBOL_LEN).max(SOURCE_SYMBOLS_MIN).
+    // R10 parameters for K = ceil(src.len() / symbol_len).max(SOURCE_SYMBOLS_MIN).
     params: CodeParameters,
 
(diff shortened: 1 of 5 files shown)


omegablitz:
If you didn't already, maybe run benches before/after this change just to confirm that perf remains the sameish

buytenh:
> If you didn't already, maybe run benches before/after this change just to confirm that perf remains the sameish

raptor_bench.rs encoding performance with the current master branch on my laptop:

```
encoder/decoder/Encoding
                        time:   [22.283 ms 22.314 ms 22.348 ms]
                        thrpt:  [170.70 MiB/s 170.96 MiB/s 171.20 MiB/s]
```

And with this commit on top:

```
encoder/decoder/Encoding
                        time:   [22.373 ms 22.455 ms 22.576 ms]
                        thrpt:  [168.97 MiB/s 169.88 MiB/s 170.51 MiB/s]
                 change:
                        time:   [+0.2281% +0.6313% +1.2085%] (p = 0.01 < 0.05)
                        thrpt:  [-1.1941% -0.6273% -0.2276%]
```

In other words, performance does not appear to be significantly affected.

PR #1083, opened 2024-09-11
Switch monad-raptorcast build_messages() back to a dynamic gso_size

buytenh:
Commit 7e18cf414442 ("Switch monad-raptorcast over to custom Raptor
encoder") made the gso_size used by monad-raptorcast a const, as the
Raptor encoder we imported required the desired encoded symbol length
to be a const generic.  This unfortunately involved the loss of the
ability to dynamically choose a per-message Merkle tree depth.

Now that our Raptor encoder no longer requires a const generic symbol
length, we can revert the part of the abovementioned commit that made
the gso_size a const, thereby regaining the ability to flexibly choose
a Merkle tree depth specifically for each transmitted message.

### DIFF
--- a/monad-raptorcast/src/udp.rs
+++ b/monad-raptorcast/src/udp.rs
@@ -253,26 +253,9 @@ const CHUNK_HEADER_LEN: u16 = 20 // Chunk recipient hash
             + 1  // reserved
             + 2; // Chunk idx
 
-// We compute these as consts so that the desired Raptor symbol length is also a const, which
-// then allows passing it into the Raptor encoder as a const generic.
-const _: () = assert!(monad_dataplane::network::MONAD_GSO_SIZE <= (u16::MAX as usize));
-const GSO_SIZE: u16 = monad_dataplane::network::MONAD_GSO_SIZE as u16;
-
-const BODY_SIZE: u16 = GSO_SIZE - HEADER_LEN - CHUNK_HEADER_LEN;
-
-// TODO make this more sophisticated
-const TREE_DEPTH: u8 = 6;
-const _: () = assert!(
-    TREE_DEPTH & (1 << 7) == 0,
-    "tree depth doesn't fit in 7 bits"
-);
-
-const PROOF_SIZE: u16 = 20 * ((TREE_DEPTH as u16) - 1);
-
-const DATA_SIZE: u16 = BODY_SIZE - PROOF_SIZE;
-
 pub fn build_messages<ST>(
     key: &ST::KeyPairType,
+    gso_size: u16,
     app_message: Bytes,
     redundancy: u8, // 2 == send 1 extra packet for every 1 original
     epoch_no: u64,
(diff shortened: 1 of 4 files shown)


monad-raptorcast/src/udp.rs
```diff
@@ -284,22 +267,39 @@ pub fn build_messages<ST>(
...
     let app_message_len: u32 = app_message.len().try_into().expect("message too big");
 
     let is_broadcast = matches!(
         build_target,
         BuildTarget::Broadcast(_) | BuildTarget::Raptorcast(_)
     );
 
+    // calculate tree depth as a function of app_message_len and body_size
+    let tree_depth: u8 = if !is_broadcast && app_message_len <= body_size.into() {
+        1 // corresponds to 1 chunk (2^(h-1))
```
omegablitz: I think this reversion actually leads to pathological behavior because of SOURCE_SYMBOLS_MIN.

If SOURCE_SYMBOLS_MIN is set to 4, then with redundancy=3, we always send a minimum of 12 unique chunks per recipient for BuildTarget::Broadcast, even if app_message_len < body_size.

Each of these unique chunks will have a distinct signature tagged onto them if tree_depth=1.

Does this make sense?

omegablitz: If this is the case, we should be able to observe this behavior in a benchmark that tries to send a small app_message.

omegablitz: For the purposes of this PR, we could keep the TREE_DEPTH to be always set to 6.

buytenh: Can we then maybe set the tree depth to 5 (= 2^(5-1) = 16 chunks) for small messages until we fix the 4 symbol lower limit?

omegablitz: Yeah we could do that.

I'll also note that if chunks_per_merkle_batch exceeds the number of chunks sent to each recipient, signatures will start to get batched across recipients.

Concretely, with tree_depth=6, chunks_per_merkle_batch=2^(tree_depth-1)=32, and 12 sent symbols per recipient, there will be approximately 1 signature for every 3 recipients.

Because of this, in retrospect, maybe it makes sense for us to fix tree_depth=6 for signature generation bandwidth purposes? There's no benefit rx-side, but it improves cpu utilization tx-side at the cost of some wasted bandwidth.

buytenh: OK, I applied the following change:

```
-    // calculate tree depth as a function of app_message_len and body_size
-    let tree_depth: u8 = if !is_broadcast && app_message_len <= body_size.into() {
-        1 // corresponds to 1 chunk (2^(h-1))
-    } else {
-        // TODO make this more sophisticated
-        6 // corresponds to 32 chunks (2^(h-1))
-    };
+    // TODO make this more sophisticated
+    let tree_depth: u8 = 6; // corresponds to 32 chunks (2^(h-1))
```

I also updated the commit message to explain why we keep the tree depth at 6 for now.  Does this look OK?

PR #1085, opened 2024-09-12
Fix broadcast chunk ids

omegablitz:
For BuildTarget::Broadcast, each recipient's assigned raptor chunk_id should start at 0.

### DIFF
--- a/monad-raptorcast/src/udp.rs
+++ b/monad-raptorcast/src/udp.rs
@@ -313,7 +313,7 @@ where
 
     let mut chunk_datas = message
         .chunks_mut(gso_size.into())
-        .map(|chunk| &mut chunk[(HEADER_LEN + proof_size).into()..])
+        .map(|chunk| (None, &mut chunk[(HEADER_LEN + proof_size).into()..]))
         .collect_vec();
     assert_eq!(chunk_datas.len(), num_packets as usize);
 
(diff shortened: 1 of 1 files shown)


buytenh:
I tested this (commit 5dd7c7f) and it seems to behave as I'd expect.  Thanks!

monad-raptorcast/src/udp.rs
```diff
@@ -410,17 +417,24 @@ where
...
         let (cursor_chunk_id, cursor) = cursor.split_at_mut(2);
         cursor_chunk_id.copy_from_slice(&chunk_id.to_le_bytes());
         let (cursor_chunk_payload, _cursor) = cursor.split_at_mut(chunk_len.into());
+
+        // for BuildTarget::Broadcast, we will be encoding each chunk_id once per recipient
+        //
+        // we could cache these as an optimization, but probably doesn't make a big difference in
+        // practice, because we're generally using BuildTarget::Broadcast for small messages.
+        //
+        // can revisit this later
```
ariqchowdhury, line 432: can we make an issue to track things that need to be revisited

buytenh: Made an issue for this: https://github.com/monad-crypto/monad-internal/issues/559

PR #1087, opened 2024-09-12
Use gossip router in a 2-node setup

rgarc:
We want to add a 0-stake node to `devnet_prime`. Instead of patching raptorcast to handle 2-nodes, we default to the gossip router when using 2 nodes.

### DIFF
--- a/monad-gossip/src/mock.rs
+++ b/monad-gossip/src/mock.rs
@@ -14,11 +14,6 @@ pub struct MockGossipConfig<PT: PubKey> {
 
 impl<PT: PubKey> MockGossipConfig<PT> {
     pub fn build(self) -> MockGossip<PT> {
-        assert!(
-            self.all_peers.len() == 1 || self.message_delay.is_zero(),
-            "Message delay only enabled for single node"
-        );
-
         MockGossip {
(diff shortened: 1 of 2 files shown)


monad-node/src/main.rs
```diff
@@ -148,7 +148,7 @@ async fn run(
         .validators
         .0
         .len()
-        > 1
+        > 2
```
omegablitz: Can you change this to enable MockGossip iff there's <= 1 **non-zero staked** validator?

monad-gossip/src/mock.rs
```diff
@@ -15,8 +15,8 @@ pub struct MockGossipConfig<PT: PubKey> {
 impl<PT: PubKey> MockGossipConfig<PT> {
     pub fn build(self) -> MockGossip<PT> {
         assert!(
-            self.all_peers.len() == 1 || self.message_delay.is_zero(),
-            "Message delay only enabled for single node"
+            self.all_peers.len() <= 2 || self.message_delay.is_zero(),
```
omegablitz: This seems super jank, I think it's better to just delete the assertion.

PR #1088, opened 2024-09-13
Fix TrieDB poll count on async reads

bamrith:


monad-triedb-utils/src/lib.rs
```diff
@@ -84,8 +84,9 @@ impl TriedbReader {
         let mut poll_count = 0;
         // Poll TrieDB until the completed_counter reaches num_accounts
         while completed_counter.load(SeqCst) < num_accounts && poll_count < MAX_TRIEDB_ASYNC_POLLS {
-            self.handle.triedb_poll();
-            poll_count += 1;
+            // blocking = true => wait to do more work
+            // max_completions = usize::MAX => process as many completions before returning
+            poll_count += self.handle.triedb_poll(true, usize::MAX);
```
ariqchowdhury, line 91: is it possible that this gets called with true when work hasn't been submitted and gets stuck? I guess if we submit a read in this function and it somehow doesn't get submitted properly, there's already an error.

bamrith: you're right. the function to initiate the read doesn't return/throw anything, so we don't know if there was an error when submitting.

ariqchowdhury: add a todo and create an issue to revisit this, maybe there should be a timeout

bamrith: done

PR #1092, opened 2024-09-14
Add support for updating statesync target

omegablitz:


omegablitz:
To be tested for parity on statesync boxes before getting opened up for review.

omegablitz:
Tested on statesync boxes along with #1093.

PR #1093, opened 2024-09-16
Track consensus tip

omegablitz:


omegablitz:
Tested on statesync boxes.

Will clean up commit messages, but opening for review because it's functional.

PR #1103, opened 2024-09-18
Fix eth_getBlockReceipts

philipglazman:
incorrect use of serde flatten on sequence

Fixes https://github.com/monad-crypto/monad-internal/issues/569

### DIFF
--- a/monad-rpc/src/block_handlers.rs
+++ b/monad-rpc/src/block_handlers.rs
@@ -274,10 +274,7 @@ pub struct MonadEthGetBlockReceiptsParams {
 }
 
 #[derive(Serialize, Debug, schemars::JsonSchema)]
-pub struct MonadEthGetBlockReceiptsResult {
-    #[serde(flatten)]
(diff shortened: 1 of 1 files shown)


michael-yxchen:
Test at https://github.com/monad-crypto/monad-integration/pull/78

monad-rpc/src/block_handlers.rs
```diff
@@ -293,11 +290,12 @@ pub async fn monad_eth_getBlockReceipts(
...
-    let block_receipts = block_receipts(triedb_env, block).await?;
-    Ok(Some(MonadEthGetBlockReceiptsResult {
-        receipts: block_receipts
-            .into_iter()
-            .map(MonadTransactionReceipt)
-            .collect(),
-    }))
+    let receipts = block_receipts(triedb_env, block).await?;
+    if receipts.is_empty() {
+        return Ok(None);
```
michael-yxchen: I think we don't need to special case here. "Ok(None)" is converted to "result: null", which means block not found

PR #1105, opened 2024-09-18
Handle SIGTERM gracefully

omegablitz:


omegablitz:
The statesync task panics when shutting down, but that should be fine.
```
thread 'tokio-runtime-worker' panicked at /usr/src/monad-bft/monad-statesync/src/ipc.rs:71:64:
request_tx dropped
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

PR #1110, opened 2024-09-20
consolidate triedb utils

eerkaijun:
wip - gonna consolidate some triedb utils and remove some unused triedb blocking reads helper function


eerkaijun:
initially wanted to switch `read_async` to use a custom future instead of a oneshot channel in this PR, but seems to have slightly worse performance, so will leave it as a oneshot channel for now

PR #1120, opened 2024-09-23
Return fake FeeHistory response

philipglazman:
Also fixes request params for `eth_feeHistory` 

**merging to devnet-prime**


philipglazman:
> do we need to have mock data returned for the `reward` field? https://www.quicknode.com/docs/ethereum/eth_feeHistory

my impression was that it was optional, but I'll include it for completeness

PR #1122, opened 2024-09-23
Gas oracle for rpc

philipglazman:
Implements a gas oracle by polling blockdb periodically. Uses blockdb atm, so leaving in draft until we migrate away. Ideal impl is streaming from execution


philipglazman:
replaced with https://github.com/monad-crypto/monad-bft/pull/1170

PR #1123, opened 2024-09-23
Full-node support

michael-yxchen:
- Added router message filter so full nodes are excluded from consensus
- Validator rebroadcasts all chunks to full nodes after servicing other
  validators


docker/devnet/fullnode.Dockerfile
```diff
@@ -0,0 +1,123 @@
...
+    mv target/release/debug-node debug-node && \
+    mv target/release/examples/wal2json wal2json && \
+    mv target/release/examples/wal-tool wal-tool && \
+    cp `ls -Lt $(find target/release | grep -e "libtriedb_driver.so") | awk -F/ '!seen[$NF]++'` . && \
+    cp `ls -Lt $(find target/release | grep -e "libmonad_statesync.so") | awk -F/ '!seen[$NF]++'` . && \
+    cp `ls -Lt $(find target/release | grep -e "libevmone.so") | awk -F/ '!seen[$NF]++'` . && \
+    cp `ls -Lt $(find target/release | grep -e "libquill.so") | awk -F/ '!seen[$NF]++'` . && \
+    cp `ls -Lt $(find target/release | grep -e "libkeccak.so") | awk -F/ '!seen[$NF]++'` .
+
+RUN apt install -y binutils && strip \
```
eclark0426: Probably don't want an install this late in the dockerfile, after everything has been built - suggest adding it to line 5 for consistency with https://github.com/monad-crypto/monad-bft/commit/54d6b65930580fa82c811718e89326c2fc51607c

docker/devnet/fullnode.Dockerfile
```diff
@@ -0,0 +1,123 @@
...
+# install cargo
+ARG CARGO_ROOT="/root/.cargo"
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | bash -s -- -y
+ENV PATH="${CARGO_ROOT}/bin:$PATH"
+
+ARG TRIEDB_TARGET=triedb_driver
+
+# Builder
+COPY . .
+RUN --mount=type=cache,target=${CARGO_ROOT}/registry    \
```
eclark0426, line 69: Just a comment on something I'm noticing, and not necessarily anything to change immediately - there's a lot of duplication between the base+builder parts of all our dockerfiles which may or may not be worth refactoring into a separate, common base/builder. This new file is almost identical to the devnet dockerfile with only about 3 or 4 relevant diffs, and seems to me like it might be better off with some of the initial steps separated out

docker/devnet/fullnode.Dockerfile
michael-yxchen: Right, I copied from a older version of devnet dockerfile. Will fix

monad-router-filter/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+use itertools::Itertools;
+use monad_consensus_types::signature_collection::SignatureCollection;
+use monad_crypto::certificate_signature::{
+    CertificateSignaturePubKey, CertificateSignatureRecoverable,
+};
+use monad_executor::Executor;
+use monad_executor_glue::{ConsensusEvent, MonadEvent, RouterCommand};
+use monad_state::VerifiedMonadMessage;
+
+pub struct FullNodeRouterFilter<ST, SCT, R> {
```
michael-yxchen, line 13: rename to FullNodeRouter

monad-eth-txpool/src/lib.rs
```diff
@@ -31,14 +31,32 @@ mod utils;
 const MAX_TXPOOL_SIZE: usize = 30_000;
 const MAX_PROPOSAL_CHECKED_TXS: usize = 10_000;
 
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug)]
 pub struct EthTxPool {
     /// pool is transient, garbage collected after creating a proposal
     pool: Pool,
     garbage: Vec<Pool>,
+    do_local_insert: bool,
```
andr-dev, line 39: I think it might be better to make a separate NoopEthTxPool which doesn't store/do anything and use that when instantiating the node.

monad-raptorcast/src/lib.rs
```diff
@@ -320,24 +327,49 @@ where
...
+                        |(from, decoded, raw_payloads)| match M::deserialize(&decoded) {
+                            Ok(app_message) => {
+                                let full_node_targets: Vec<SocketAddr> = this
+                                    .full_nodes
+                                    .list
+                                    .iter()
+                                    .filter_map(|full_node| {
+                                        this.known_addresses.get(full_node).copied()
+                                    })
+                                    .collect();
```
andr-dev: We can construct this outside the `filter_map` right?

monad-router-filter/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+                            MonadEvent::MempoolEvent(_) => continue,
+                            _ => maybe_event,
+                        };
+                        return Poll::Ready(filtered);
+                    } else {
+                        return Poll::Ready(None);
+                    }
+                }
+                Poll::Pending => return Poll::Pending,
+            }
```
andr-dev: This was difficult to read, can we unnest after the `if let Some(event)` and put the `filtered` stuff afterwards?

monad-eth-txpool/src/lib.rs
michael-yxchen: I agree. That will be a decent amount of refactoring to separate txn validation and insertion logic from EthTxPool now, to reuse existing validation logic. This will change soon when we merge full node logic into consensus so there's only one binary

michael-yxchen:
Locally verified that CI passes with: https://github.com/monad-crypto/monad-integration/pull/117

PR #1124, opened 2024-09-23
Add raptorcast known_addresses peer dynamically

rgarc:
Introduce AddPeer router command

Peer discovery needs the router to learn of new addresses at runtime so
we introduce a new command to allow this. Clearing stale addresses in
the raptorcast router is left as a TODO.

It is to be decided how the following routers will handle the AddPeer
command.

- SyncEndpoint (monad-quic) might be deprecated soon
- MockExecutor (monad-mock-swarm) more thinking needs to go into how to
  support dynamic peers in mock swarm
- LocalPeerRouter (monad-testground) was replaced by the raptorcast
  router in https://github.com/monad-crypto/monad-bft/commit/af3f39ded846c3ebd42e120f02ee94a745cfe6b8

### DIFF
--- a/monad-executor-glue/src/lib.rs
+++ b/monad-executor-glue/src/lib.rs
@@ -1,6 +1,6 @@
 pub mod convert;
 
-use std::fmt::Debug;
+use std::{fmt::Debug, net::SocketAddr};
 
 use bytes::{BufMut, Bytes, BytesMut};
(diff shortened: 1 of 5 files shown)


michael-yxchen:
I'm reviewing everything together in #1184 if that's ok

PR #1125, opened 2024-09-24
Remove blockdb persist bft blocks in file ledger

ariqchowdhury:


monad-rpc/src/block_util.rs
```diff
@@ -0,0 +1,98 @@
...
+    }
+}
+
+pub enum BlockResult {
+    Block(EthBlock),
+    NotFound,
+    DecodeFailed(AlloyError),
+}
+
+pub fn get_block_from_num(reader: &FileBlockReader, block_num: u64) -> BlockResult {
```
ariqchowdhury: make async

monad-rpc/src/cli.rs
```diff
@@ -11,11 +11,15 @@ pub struct Cli {
...
 
-    /// Set the monad blockdb path
+    /// Set the path where the bft block headers will be stored
     #[arg(long)]
-    pub blockdb_path: Option<PathBuf>,
+    pub bft_block_header_path: PathBuf,
+
+    /// Set the path where the bft block payloads will be stored
+    #[arg(long)]
+    pub bft_block_payload_path: PathBuf,
```
eerkaijun, line 22: out of curiosity, are we using these for rpc?

monad-rpc/src/eth_txn_handlers.rs
```diff
@@ -538,24 +535,37 @@ pub struct MonadEthGetTransactionByBlockNumberAndIndexParams {
...
     };
 
     let retval = parse_tx_content(&block, transaction, params.index.0).map(MonadTransaction);
 
     Ok(retval)
 }
+
+fn get_block_hash(block: &EthBlock) -> FixedBytes<32> {
+    let mut block_hash = [0; 32];
+    block.extra_data.clone().copy_to_slice(&mut block_hash);
```
eerkaijun, line 569: are we now storing the block hash in the `extra_data` field in the header?

monad-rpc/src/cli.rs
ariqchowdhury: no, right now they would only be used from consensus to retrieve blocks for blocksync requests

monad-rpc/src/eth_txn_handlers.rs
ariqchowdhury: storing the bft block_id so that there is a link from the eth block to the bft block

PR #1126, opened 2024-09-24
Fix eth_getLogs

philipglazman:
Use generic ValueOrArray for topics and address


philipglazman:
still need to add corresponding flexnet tests, this was not caught earlier

PR #1127, opened 2024-09-25
Add random tests

eclark0426:
It's not a whole lot more coverage, but here's a two thirds of a percent of extra code coverage!

I think the biggest things that aren't currently widely covered by unit tests are:
- overarching monad-node crate
- random things in monad-consensus-types
- monad-quic
- RPC
- statesync
- triedb utils

I was trying for a while to get instrumentation through Flexnet tests, since much of that is probably tested through Flexnet, but unfortunately it didn't want to cooperate

### DIFF
--- a/monad-types/src/lib.rs
+++ b/monad-types/src/lib.rs
@@ -439,6 +439,8 @@ where
 
 #[cfg(test)]
 mod test {
+    use monad_crypto::NopPubKey;
+    use serde_json::from_str;
     use test_case::test_case;
 
     use super::*;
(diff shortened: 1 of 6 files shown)


ariqchowdhury:
monad-quic isn't used anymore (replace by raptorcast), should probably be removed at some point.

monad-types/src/lib.rs
```diff
@@ -455,4 +457,108 @@ mod test {
...
+        let val_set_update_interval = SeqNum(100);
+        let seq_num = SeqNum(0);
+        seq_num.get_locked_epoch(val_set_update_interval);
+    }
+
+    #[test]
+    fn test_deserialize_missing_hex_prefix() {
+        let json = r#"{"node_id": "0011223344556677"}"#;
+        let result = from_str::<NodeId<NopPubKey>>(json);
+        assert!(result.is_err());
```
andr-dev, line 527: Can we have a separate test or add another assert here to show how a `node_id` with a 0x prefix succeeds? It's unclear from the test why from_str is failing. As an example, it could be because we rename `node_id` to `id_node`.

monad-types/src/lib.rs
```diff
@@ -455,4 +457,108 @@ mod test {
...
+        let deserialized = NopPubKey::deserialize(&test_pubkey).unwrap();
+        assert_eq!(deserialized, NopPubKey::from_bytes(&test_bytes).unwrap());
+    }
+
+    #[test]
+    fn test_drop_timer_drops() {
+        let timer = DropTimer::start(Duration::from_millis(10), |_| {
+            panic!();
+        });
+    }
```
andr-dev, line 554: How about a test for `Duration::ZERO`?

PR #1132, opened 2024-09-27
A few edits to rpc+bft Dockerfiles

mhumeSF:
- Dropping `software-properties-common` because we're not using external apt repos anymore.
- `s/apt-get/apt/g`

### DIFF
--- a/monad-rpc/src/cli.rs
+++ b/monad-rpc/src/cli.rs
@@ -45,9 +45,15 @@ pub struct Cli {
     #[arg(long, default_value_t = 25_000_000)]
     pub max_response_size: u32,
 
+    /// Otel endpoint to collect metrics data
     #[arg(long)]
     pub otel_endpoint: Option<String>,
 
+    /// Allow pre EIP-155 transactions
     #[arg(long, default_value_t = false)]
     pub allow_unprotected_txs: bool,
+
+    /// Set the max concurrent requests for eth_call and eth_estimateGas
+    #[arg(long, default_value_t = 1000)]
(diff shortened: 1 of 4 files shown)


docker/devnet/Dockerfile
```diff
@@ -1,14 +1,12 @@
...
+RUN apt install -y ca-certificates curl pkg-config gnupg wget cgroup-tools
+RUN apt install -y libstdc++-13-dev
+RUN apt install -y gcc-13 g++-13
 
-RUN add-apt-repository -y ppa:ubuntu-toolchain-r/test
-RUN add-apt-repository -y ppa:mhier/libboost-latest
-RUN apt-get update && apt-get install -y libstdc++-13-dev gcc-13 g++-13
-
-RUN apt-get update && apt-get install -y \
+RUN apt install -y \
```
eclark0426: This line (and the other similar diffs) should be left as it was - doing an apt install without an apt update is prone to 404 errors during installs because of cached results from apt update

ariqchowdhury:
is this PR still necessary?

docker/devnet/Dockerfile
mhumeSF: @eclark0426 I've updated to keep the `apt update` updates

mhumeSF:
@ariqchowdhury Yea, this PR does to things... standardize on `apt` and with moving to ubuntu-24.04 we have moved off of additional repos like kitware and ubuntu-toolchain/r. Both sort of nit-picky, but I think worth it.

PR #1141, opened 2024-10-02
Skip ClearMempool when node is leader in upcoming round

andr-dev:
If the current node will be the leader in the next few rounds, we should not clear the mempool.

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -42,6 +42,9 @@ use tracing::{debug, info, trace, warn};
 
 use crate::{command::ConsensusCommand, timestamp::BlockTimestamp};
 
+// TODO: Configurable
+const MAX_UPCOMING_LEADER: u64 = 3;
+
 pub mod command;
 pub mod timestamp;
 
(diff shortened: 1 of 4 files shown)


omegablitz:
I think the single node devnet txpool will grow unboundedly if ingress rate exceeds block production rate

ariqchowdhury:
> I think the single node devnet txpool will grow unboundedly if ingress rate exceeds block production rate

That should be fine on `devnet-prime`
But we can also add a hard-cap to the txpool size in case -- just clear everything no matter what after a proposal if there's X number of txns

michael-yxchen:
I think it's helpful to have a hard cap. Reasoning about leader schedule and `NUM_LEADERS_FORWARD_SKIP_MEMPOOL_CLEAR` can get confusing. i.e. 3 equal-stake nodes with MEMPOOL_CLEAR configured as 3 will never clear their mempool

andr-dev:
New idea from @ariqchowdhury is to tell the txpool in `create_proposal` whether we'll also be an upcoming validator so it can decide whether to move the entire pool to the `garbage`, effectively clearing itself, or to just remove the txs that get proposed. I pushed up the changes needed outside of the txpool, now working through the txpool itself.

monad-consensus-state/src/lib.rs
```diff
@@ -1205,13 +1208,33 @@ where
...
+                    })
+                    .any(|(epoch, round)| {
+                        &self.election.get_leader(
+                            round,
+                            self.val_epoch_map
+                                .get_val_set(&epoch)
+                                .expect("handle non-existent validatorset for upcoming round epoch")
+                                .get_members(),
+                        ) == self.nodeid
+                    });
```
andr-dev, line 1228: Will become a helper function in a bit.

ariqchowdhury:
new txpool design doesn't get rid of anything (unless at capacity) so this PR can be closed

PR #1142, opened 2024-10-02
Verify block_id during block validation

bamrith:


michael-yxchen:
what would cause this validation to fail?

bamrith:
@michael-yxchen 
> what would cause this validation to fail?

if any of the fields in the block were changed after it was created.

for example, a malicious node could create a block, clone it and change some fields, and propose two different blocks with the same `block_id`. without this check, two honest nodes will vote for both of them when they could be looking at two different payloads.

michael-yxchen:
block_id is not sent over the wire. It's reconstructed by the receiver, so in your example two honest nodes will compute different block_ids with different payloads

bamrith:
oh, i didn't know it was reconstructed by the receiver. this isn't required then.

PR #1143, opened 2024-10-03
Fix crash when parsing messages of length 0

eclark0426:


monad-dataplane/src/event_loop.rs
```diff
@@ -441,6 +441,10 @@ impl DataplaneEventLoop {
             total_recv_bytes += len;
             let src_sock_addr = result.src_addr;
 
+            if len == 0 {
```
eclark0426, line 444: Not sure if a 0-length message would ever be valid; if it is valid to get a message with 0 length, then the crash ultimately is because stride is still set to 0 on line 449 for a message with length 0; so the crash could instead be fixed by changing that to `len.max(1)`

buytenh:
Good catch! Could you fold something like the following reproducer into your PR?

```
diff --git a/monad-raptorcast/tests/malformed_inputs.rs b/monad-raptorcast/tests/malformed_inputs.rs
index b89686c4..ba11c3d8 100644
--- a/monad-raptorcast/tests/malformed_inputs.rs
+++ b/monad-raptorcast/tests/malformed_inputs.rs
@@ -194,6 +194,31 @@ pub fn oversized_message() {
     std::thread::sleep(Duration::from_millis(100));
 }
 
+// Try to crash the RaptorCast receive path by feeding it a zero-sized packet.  A previous
+// version of the RaptorCast receive path would crash via handle_message() when receiving
+// a zero-sized packet due to that causing it to be invoked with message.payload.len() ==
+// message.stride == 0 which would then call .step_by(0) on (0..0), which panics.
+#[test]
+pub fn zero_sized_packet() {
+    let rx_addr = "127.0.0.1:10006";
+    let tx_addr = "127.0.0.1:10007";
+
+    let (_rx_nodeid, _tx_nodeid, _tx_keypair, _known_addresses) = set_up_test(rx_addr, tx_addr);
+
+    let message = vec![0; 10];
+
+    let tx_socket = UdpSocket::bind(tx_addr).unwrap();
+
+    // Sending a single zero-sized packet is sufficient to crash the
+    // receiver if it is vulnerable to this issue.
+    tx_socket
+        .send_to(&message[0..0], rx_addr)
+        .unwrap();
+
+    // Wait for RaptorCast instance to catch up.
+    std::thread::sleep(Duration::from_millis(100));
+}
+
 static ONCE_SETUP: Once = Once::new();
 
 pub fn set_up_test(
```

The slightly unexpected (for me) cause of this issue is .step_by(0) panicking even if it is called on (0..0), but it makes sense why it would do that. I don't think that we will ever need to support a use case where dataplane will need to pass zero-sized received packets up to raptorcast, so we can drop such packets in either of these two places.

I tried finding out what happens when we receive an aggregate (GRO) packet consisting of multiple zero-byte sub-packets, but this isn't entirely trivial to do, as such packets cannot be generated by transmitting over the loopback interface (which is how we'd usually test this) as the size argument you pass to a GSO write call is num_packets * packet_size, where the information about num_packets gets lost when packet_size is zero. I don't know what happens if multiple zero-sized UDP packets are received over the network with a NIC that supports GRO in hardware. We don't seem to get a UDP_GRO cmsg if only a single packet is received, and I suppose that we would get a UDP_GRO cmsg with value 0 if multiple zero-sized packets had been aggregated, but I don't know if such aggregation would actually happen in practice.

PR #1146, opened 2024-10-07
Clear obsolete extending block txs from txpool

omegablitz:
This lowers the amount of unnecessary state accesses we need to do. This is because a nontrivial amount of txs in the proposer's txpool likely had already been included in the prior block due to the virtual mempool.

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -788,6 +788,7 @@ where
                     // epoch manager records
                     self.metrics.consensus_events.commit_block += 1;
                     self.block_policy.update_committed_block(block);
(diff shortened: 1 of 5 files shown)


michael-yxchen:
if txns were recently committed (included last execution delay blocks), they should hit committed_cache on nonce fetches. I think it's easier to do a nonce filtering before querying reserve balances, i.e. move [this block](https://github.com/monad-crypto/monad-bft/blob/2ff6da0cb2a5f4a5be9239e3c5347fa4df5e0b99/monad-eth-txpool/src/lib.rs#L152-L173) before `compute_account_base_reserve_balances`

PR #1148, opened 2024-10-08
Allow pre-eip155 transactions at consensus level

eerkaijun:
Closes https://github.com/monad-crypto/monad-internal/issues/589

Allows pre EIP-155 transactions at consensus level, but additional check on the rpc level which is configurable by the rpc provider. This is consistent with geth

### DIFF
--- a/monad-eth-block-policy/src/lib.rs
+++ b/monad-eth-block-policy/src/lib.rs
@@ -81,9 +81,12 @@ pub fn static_validate_transaction(
     chain_id: u64,
 ) -> Result<(), TransactionError> {
     // EIP-155
-    tx.chain_id()
-        .and_then(|cid| (cid == chain_id).then_some(()))
-        .ok_or(TransactionError::InvalidChainId)?;
+    // We allow legacy transactions without chain_id specified to pass through
+    if let Some(tx_chain_id) = tx.chain_id() {
+        if tx_chain_id != chain_id {
+            return Err(TransactionError::InvalidChainId);
+        }
+    }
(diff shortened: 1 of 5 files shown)


monad-eth-block-policy/src/lib.rs
```diff
@@ -81,9 +81,12 @@ pub fn static_validate_transaction(
     chain_id: u64,
 ) -> Result<(), TransactionError> {
     // EIP-155
-    tx.chain_id()
```
ariqchowdhury, line 84: is there a check to see directly if the txn is a legacy type?

should add a test that tries legacy txns and non-legacy txns

eerkaijun: pre-EIP155 transaction is only a subset of legacy transaction. added the test

monad-rpc/src/eth_txn_handlers.rs
```diff
@@ -400,6 +401,13 @@ pub async fn monad_eth_sendRawTransaction(
                 return Err(JsonRpcError::custom(error_message.to_string()));
             }
 
+            // drop pre EIP-155 transactions if disallowed by the rpc (for user protection purposes)
+            if !allow_unprotected_txs && txn.chain_id().is_none() {
+                return Err(JsonRpcError::custom(
+                    "Unprotected transactions are not allowed".to_string(),
```
ariqchowdhury: maybe "Unprotected transactions (pre-EIP155) are not allowed over RPC"

PR #1169, opened 2024-10-15
fix performance issues with leader election

maxkozlovsky:
Leader election was O(stake value) which could lead to performance issues when stake values were high. Modify the algorithm to provide O(log(len(validators)) performance not depending on particular stake values.

The new algorithm involves pseudo random
number generation to provide approximate interleaving of leaders so it does not result in perfect round robin distribution.

### DIFF
--- a/monad-validator/src/weighted_round_robin.rs
+++ b/monad-validator/src/weighted_round_robin.rs
@@ -1,6 +1,9 @@
 use std::{collections::BTreeMap, marker::PhantomData};
 
-use monad_crypto::certificate_signature::PubKey;
+use monad_crypto::{
+    certificate_signature::PubKey,
+    hasher::{Hasher, HasherType},
+};
 use monad_types::{NodeId, Round, Stake};
 
 use crate::leader_election::LeaderElection;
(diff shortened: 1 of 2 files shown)


monad-validator/src/weighted_round_robin.rs
```diff
@@ -18,48 +22,47 @@ impl<PT: PubKey> Default for WeightedRoundRobin<PT> {
     }
 }
 
+fn randomize(x: u64, m: u64) -> u64 {
```
omegablitz, line 24: We'll want to use a cryptographic pseudorandom function here, like ChaCha. https://docs.rs/rand_core/0.6.2/rand_core/trait.SeedableRng.html#tymethod.from_seed

maxkozlovsky: Why a cryptographic hash is necessary here? It seems we just need good enough random distribution and standard hash (siphash) does that.

omegablitz: Hmm, you may be right that it doesn't need to be cryptographic.

Regardless, don't think we should be using DefaultHasher here, because the hash algorithm is something that the rust compiler is able to change at any time. The leader schedule needs to be stable across builds.

maxkozlovsky: I agree, being explicit about the hash algorithm will be better. I can use blake3 since it looks like it is used as default hash elsewhere in the code. Is this ok?

omegablitz: Yeah, that'll work

maxkozlovsky: updated to use blake3 hash

omegablitz:
LGTM

PR #1170, opened 2024-10-15
Blockwatcher and eip1559 gas oracle 

philipglazman:
Blockwatcher will stream blocks to virtual mempool in order to evict or promote transactions. It is currently a basic implementation that does not yet use triedb.

Gas Oracle provides methods for calculating the base fee for the next block according to eip1559 rules.

### DIFF
--- a/monad-rpc/src/block_watcher.rs
+++ b/monad-rpc/src/block_watcher.rs
@@ -1,0 +1,163 @@
+use std::{
+    future::Future,
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+use futures::{FutureExt, Stream};
+use pin_project::pin_project;
+use reth_primitives::Block;
+use reth_rpc_types::TransactionReceipt;
+
+#[pin_project::pin_project(project = StateProj)]
+enum State {
+    GetBlock(u64),
+    GetBlockResult {
+        task: std::pin::Pin<Box<dyn futures::Future<Output = Block> + Send + Sync>>,
+    },
+    BlockAndReceipts {
+        block: Block,
+        task:
+            std::pin::Pin<Box<dyn futures::Future<Output = Vec<TransactionReceipt>> + Send + Sync>>,
+    },
+}
+
+// TODO: remove this trait when triedb contains blocks
+pub trait BlockState: Send {
+    fn get_block(&self, block_num: u64) -> impl Future<Output = Block> + Send + Sync;
+    fn get_receipts(
+        &self,
+        block_num: u64,
+    ) -> impl Future<Output = Vec<TransactionReceipt>> + Send + Sync;
+}
+
+#[derive(Clone)]
+pub struct MockBlockState {
+    blocks: Vec<Block>,
+}
+
+impl MockBlockState {
+    pub fn new() -> Self {
+        Self { blocks: Vec::new() }
+    }
+
+    pub fn set_blocks(&mut self, blocks: Vec<Block>) {
+        self.blocks = blocks;
+    }
+}
+
+impl BlockState for MockBlockState {
+    async fn get_block(&self, block_num: u64) -> Block {
+        if self.blocks.is_empty() || block_num > self.blocks.len() as u64 {
+            Block::default()
+        } else {
+            self.blocks
+                .get(block_num as usize)
+                .cloned()
+                .unwrap_or_default()
+        }
+    }
+
+    async fn get_receipts(&self, _block_num: u64) -> Vec<TransactionReceipt> {
+        vec![]
+    }
+}
+
+#[pin_project]
+pub struct BlockWatcher<B: BlockState + Clone> {
+    triedb: B,
+    interval: tokio::time::Interval,
+    #[pin]
+    state: State,
+}
+
+impl<B: BlockState + Clone> BlockWatcher<B> {
+    pub fn new(triedb: B, current_height: u64) -> Self {
+        Self {
+            triedb,
+            interval: tokio::time::interval(std::time::Duration::from_secs(1)),
+            state: State::GetBlock(current_height),
+        }
+    }
+}
+
+pub struct BlockWithReceipts {
+    block: Block,
+    receipts: Vec<TransactionReceipt>,
+}
+
+impl<B: BlockState + Clone + Send + Sync + 'static> Stream for BlockWatcher<B> {
+    type Item = BlockWithReceipts;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let mut this = self.as_mut().project();
+        let mut state = this.state.as_mut();
+
+        match state.as_mut().project() {
+            StateProj::GetBlock(height) => match this.interval.poll_tick(cx) {
+                Poll::Ready(_) => {
+                    let triedb = this.triedb.clone();
+                    let height = *height;
+                    let task = async move { triedb.get_block(height).await };
+                    let task = Box::pin(task);
+                    state.set(State::GetBlockResult { task });
+                    cx.waker().wake_by_ref();
+                    Poll::Pending
+                }
+                Poll::Pending => Poll::Pending,
+            },
+            StateProj::GetBlockResult { task } => match task.poll_unpin(cx) {
+                Poll::Ready(block) => {
+                    let triedb = this.triedb.clone();
+                    let height = block.header.number;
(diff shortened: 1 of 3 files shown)


monad-rpc/src/gas_oracle.rs
```diff
@@ -0,0 +1,356 @@
...
+        .collect::<Vec<TransactionSigned>>();
+    transactions.sort_by_cached_key(|tx| tx.effective_tip_per_gas(Some(base_fee)));
+
+    let mut prices = Vec::new();
+    let mut rewards = Vec::new();
+    let mut gas_used_ratios = Vec::new();
+    for (idx, tx) in transactions.iter().enumerate() {
+        let tip = tx
+            .effective_tip_per_gas(Some(base_fee))
+            .expect("missing base fee");
```
eerkaijun: can we remove the `.expect()` here and log a `warn` or `error` instead?

monad-rpc/src/block_watcher.rs
```diff
@@ -0,0 +1,160 @@
...
+    }
+}
+
+impl BlockState for MockBlockState {
+    async fn get_block(&self, block_num: u64) -> Block {
+        if self.blocks.is_empty() || block_num > self.blocks.len() as u64 {
+            Block::default()
+        } else {
+            self.blocks.get(block_num as usize).unwrap().clone()
+        }
```
andr-dev: `self.blocks.get(block_num as usize).unwrap_or_default()`

monad-rpc/src/gas_oracle.rs
```diff
@@ -0,0 +1,365 @@
...
+        };
+        if cache.len() < self.block_sample_size {
+            return None;
+        }
+
+        let mut prices: Vec<u64> = cache
+            .iter()
+            .take(self.block_sample_size)
+            .cloned()
+            .flat_map(|block| block.sampled_tips)
```
andr-dev: We should cloned() after the flat_map.

monad-rpc/src/gas_oracle.rs
```diff
@@ -0,0 +1,365 @@
...
+}
+
+fn process_block(
+    block: Block,
+    receipts: Vec<TransactionReceipt>,
+) -> Result<ProcessedBlock, GasOracleError> {
+    let base_fee = match block.base_fee_per_gas {
+        Some(base_fee) => base_fee,
+        None => return Err(GasOracleError::MissingBaseFee),
+    };
```
andr-dev: `let base_fee = block.base_fee_per_gas.ok_or(GasOracleError::MissingBaseFee)?;`

monad-rpc/src/gas_oracle.rs
```diff
@@ -0,0 +1,365 @@
...
+        cache.push_front(processed_block);
+
+        if cache.len() > CACHE_CAPACITY {
+            cache.pop_back();
+        };
+
+        Ok(())
+    }
+
+    fn base_fee(&self) -> Option<u64> {
```
andr-dev, line 104: Can we add a comment linking to the EIP where this computation is defined?

monad-rpc/src/gas_oracle.rs
```diff
@@ -0,0 +1,365 @@
...
+    let base_fee = match block.base_fee_per_gas {
+        Some(base_fee) => base_fee,
+        None => return Err(GasOracleError::MissingBaseFee),
+    };
+
+    let mut transactions = block
+        .body
+        .clone()
+        .into_iter()
+        .collect::<Vec<TransactionSigned>>();
```
andr-dev: I think we can just .iter() here and let transactions be a `Vec<&TransactionSigned>` to save us the clone cost.

monad-rpc/src/gas_oracle.rs
```diff
@@ -0,0 +1,365 @@
...
+            continue;
+        }
+
+        let tip = tip.unwrap();
+
+        // For each receipt, calculate gas_used_ratio
+        let receipt = match receipts.get(idx) {
+            Some(r) => r,
+            None => return Err(GasOracleError::TransactionReceiptMissing),
+        };
```
andr-dev:
```rust
let receipt = receipts.get(idx).ok_or(GasOracleError::TransactionReceiptMissing)?;
```

monad-rpc/src/gas_oracle.rs
```diff
@@ -0,0 +1,365 @@
...
+            continue;
+        };
+
+        let gas_used: f64 = gas_used.into();
+        let gas_used_ratio = gas_used / block.gas_limit as f64;
+        gas_used_ratios.push(gas_used_ratio);
+
+        if tip < IGNORE_PRICE {
+            continue;
+        }
```
andr-dev, line 198: What's the rationale behind the IGNORE_PRICE being const? I think I understand that we want to skip txs that have low tips, but what "low" means doesn't seem like a fixed value over time.

monad-rpc/src/gas_oracle.rs
```diff
@@ -0,0 +1,365 @@
...
+        });
+        blocks.push(Block {
+            header: Header {
+                base_fee_per_gas: Some(1000),
+                number: 2,
+                ..Default::default()
+            },
+            body: vec![make_tx(1103), make_tx(1104), make_tx(1105)],
+            ..Default::default()
+        });
```
andr-dev: Probably cleaner to just
```rust
let blocks = [
    Block {
        ...
    }
    ...
];

File: monad-rpc/src/gas_oracle.rs [hunk 10]
```diff
@@ -0,0 +1,365 @@
...
+            TransactionReceipt {
+                gas_used: Some(U256::from(21_000)),
+                ..Default::default()
+            },
+        ];
+
+        oracle.process_block(block1, receipts).unwrap();
+        assert_eq!(oracle.base_fee().unwrap(), 1011); // Base fee increases
+    }
+}
```
andr-dev, line 362: We should add a test that shows how txs with tip above/below IGNORE_PRICE impact the oracle.

monad-rpc/src/gas_oracle.rs
```diff
@@ -0,0 +1,365 @@
...
+        } else {
+            warn!("could not calculate effective tip for {:?}", tx.hash());
+            None
+        };
+
+        if tip.is_none() {
+            continue;
+        }
+
+        let tip = tip.unwrap();
```
andr-dev:
```rust
let Some(tip) = tx.effective_tip_per_gas(Some(base_fee)) else {
    warn!("could not calculate effective tip for {:?}", tx.hash());
    continue;
};

rewards.push(tip);
```

monad-rpc/src/gas_oracle.rs
philipglazman: We could add it as a cli flag to rpc, but did not want to clutter the rpc with cli flags unless we need it. I don't feel strongly about this so can add it as a param if we find it useful. I borrowed the same default from geth.

andr-dev: Got it, I agree that we should minimize the number of rpc cli flags. Can we add a doc comment linking to the const in geth so that if we revisit it in the future we know where it came from?

monad-rpc/src/gas_oracle.rs
```diff
@@ -0,0 +1,355 @@
...
+/// Gas price percentile
+const GAS_PRICE_PERCENTILE: f64 = 0.60;
+
+const IGNORE_PRICE: u128 = 2;
+
+/// Number of recent blocks to cache
+const CACHE_CAPACITY: usize = 100;
+
+pub struct Oracle {
+    cache: Arc<std::sync::Mutex<LinkedList<ProcessedBlock>>>,
```
ariqchowdhury: vecdeque instead of linkedlist? Its fixed capacity so ringbuffer should work

monad-rpc/src/block_watcher.rs
ariqchowdhury, line 1: can think about this in a future PR, but block_watcher seems like a useful generic thing to have outside of RPC. Can imagine other tools/processes on a full-node wanting to watch blocks and do something.

monad-rpc/src/gas_oracle.rs
philipglazman: fair!

monad-rpc/src/gas_oracle.rs
philipglazman: added

PR #1172, opened 2024-10-15
Rewrite txpool to dynamically track account nonces

andr-dev:
Previously, the txpool would validate reserve balance at insertion time and validate both nonce and reserve balance at create_proposal time. This led to significant triedb read pressure as possibly thousands of accounts need to be looked up all at once and at a critical point in the consensus process.

This changeset rearchitects the txpool to keep a collection of untracked addresses for which the pool knows nothing about and a collection of nonce-tracked addresses. The txpool can then selectively promote addresses from untracked to tracked, allowing the txpool to select when to promote txs and bounding the number of reads to triedb through each promotion.

The final goal is to move the txpool and the promotion process to a separate thread so that these operations can happen in parallel to consensus. This would then enable create_proposal to be bounded in time and have close to zero triedb reads, keeping the hotpath as efficient as possible.

Bundled into this changeset is a set of additional features requested for testnet. These include the following:
1. Time based tx expiry
2. Don't clear txpool after each round
    - Increases average TPS under normal load
    - Saturates TPS when hitting TPS limit
3. Reduce triedb account lookups
    - Accounts with no available transactions for inclusion, given pending blocktree, are removed from lookup set

These changes have been running smoothly in stressnet for the past 2-3 weeks cherrypicked into the release-testnet branch.

#### Clear
```
txpool/clear-simple
                        time:   [286.03 µs 288.12 µs 290.34 µs]
                        change: [-83.309% -82.967% -82.578%] (p = 0.00 < 0.05)
                        Performance has improved.
txpool/clear-random
                        time:   [283.84 µs 284.58 µs 285.33 µs]
                        change: [-83.945% -83.806% -83.664%] (p = 0.00 < 0.05)
                        Performance has improved.
txpool/clear-duplicate_nonce
                        time:   [27.775 µs 27.886 µs 28.027 µs]
                        change: [-9.0070% +2.0040% +15.454%] (p = 0.74 > 0.05)
                        No change in performance detected.
txpool/clear-increasing_nonce
                        time:   [30.868 µs 31.974 µs 33.082 µs]
                        change: [+1448.2% +1948.4% +2465.0%] (p = 0.00 < 0.05)
                        Performance has regressed.
txpool/clear-nonce_gaps
                        time:   [27.318 µs 28.322 µs 29.355 µs]
                        change: [+9244.1% +10139% +11201%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

#### Create Proposal
```
txpool/create_proposal-simple
                        time:   [25.870 ms 26.014 ms 26.152 ms]
                        change: [+38.288% +39.178% +39.967%] (p = 0.00 < 0.05)
                        Performance has regressed.
txpool/create_proposal-random
                        time:   [25.693 ms 25.903 ms 26.118 ms]
                        change: [+35.832% +37.494% +39.100%] (p = 0.00 < 0.05)
                        Performance has regressed.
txpool/create_proposal-duplicate_nonce
                        time:   [1.9111 ms 1.9418 ms 1.9712 ms]
                        change: [+48.007% +50.635% +53.192%] (p = 0.00 < 0.05)
                        Performance has regressed.
txpool/create_proposal-increasing_nonce
                        time:   [1.0409 ms 1.0448 ms 1.0489 ms]
                        change: [+0.9978% +1.9989% +2.9528%] (p = 0.00 < 0.05)
txpool/create_proposal-nonce_gaps
                        time:   [766.61 µs 769.20 µs 771.86 µs]
                        change: [-18.075% -17.272% -16.521%] (p = 0.00 < 0.05)
                        Performance has improved.
```

#### Insert
```
txpool/insert-simple
                        time:   [41.806 ms 41.937 ms 42.088 ms]
                        change: [-27.742% -27.449% -27.116%] (p = 0.00 < 0.05)
                        Performance has improved.
txpool/insert-random
                        time:   [41.841 ms 41.971 ms 42.123 ms]
                        change: [-26.046% -25.745% -25.432%] (p = 0.00 < 0.05)
                        Performance has improved.
txpool/insert-duplicate_nonce
                        time:   [26.571 ms 26.733 ms 26.916 ms]
                        change: [-11.990% -11.023% -10.177%] (p = 0.00 < 0.05)
                        Performance has improved.
txpool/insert-increasing_nonce
                        time:   [27.194 ms 27.633 ms 28.225 ms]
                        change: [+1.5038% +3.0907% +5.3692%] (p = 0.00 < 0.05)
                        Performance has regressed.
txpool/insert-nonce_gaps
                        time:   [26.709 ms 26.973 ms 27.293 ms]
                        change: [-0.8147% +0.2026% +1.4349%] (p = 0.74 > 0.05)
                        No change in performance detected.
```

### Flexnet Benchmark (1node-debug, Native Token Transfers 256k x 256k, tx_limit = 10_000)

flexnet_stress_256k

### Flexnet Benchmark (1node-debug, Native Token Transfers 256k x 256k, tx_limit = 3_000)

flexnet_stress_saturated_256k

### Flexnet Benchmark (2nodes-debug, Native Token Transfers 256k x 256k, tx_limit = 10_000)

flexnet_stress_256k


ariqchowdhury:
add benchmarking for flexnet so we get comparisons with real triedb reads

monad-eth-txpool/src/pending/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    pub fn split_off(&mut self, count: usize) -> IndexMap<EthAddress, EthTxPendingList> {
+        if count >= self.txs.len() {
+            return std::mem::take(&mut self.txs);
+        }
+
+        let mut split = self.txs.split_off(count);
+        std::mem::swap(&mut split, &mut self.txs);
+
+        resize_indexmap(&mut self.txs);
```
JoeHowarth: Why resize here?

monad-eth-txpool/src/tracked/mod.rs
```diff
@@ -0,0 +1,376 @@
...
+
+        let pending_account_nonces = extending_blocks.get_account_nonces();
+
+        #[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
+        struct VirtualValidEthTransaction<'a> {
+            tx: &'a ValidEthTransaction,
+            virtual_time: u64,
+        }
+
+        impl<'a> VirtualValidEthTransaction<'a> {
```
JoeHowarth: I would either remove this impl and construct it using punning or move the struct outside this function

monad-eth-txpool/src/tracked/list.rs
```diff
@@ -0,0 +1,91 @@
...
+        if tx.nonce() < self.account_nonce {
+            return Err(TxPoolInsertionError::NonceTooLow);
+        }
+
+        match self.txs.entry(tx.nonce()) {
+            btree_map::Entry::Vacant(v) => {
+                v.insert(tx);
+            }
+            btree_map::Entry::Occupied(mut existing_tx) => {
+                if tx.cmp(existing_tx.get()).is_lt() {
```
JoeHowarth: nit: why not use a '<' for readability?

monad-eth-txpool/src/tracked/list.rs
```diff
@@ -0,0 +1,91 @@
...
+        mut this: indexmap::map::OccupiedEntry<'_, EthAddress, TrackedTxList>,
+        account_nonce: &u64,
+    ) {
+        this.get_mut().account_nonce = *account_nonce;
+
+        let Some((lowest_nonce, _)) = this.get().txs.first_key_value() else {
+            return;
+        };
+
+        if lowest_nonce.ge(account_nonce) {
```
JoeHowarth: ditto

monad-eth-txpool/src/tracked/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+        &mut self,
+        tx: ValidEthTransaction,
+    ) -> Result<(), Either<ValidEthTransaction, TxPoolInsertionError>> {
+        match self.txs.entry(tx.sender()) {
+            IndexMapEntry::Vacant(_) => Err(Either::Left(tx)),
+            IndexMapEntry::Occupied(mut o) => o.get_mut().try_add_tx(tx).map_err(Either::Right),
+        }
+    }
+
+    pub fn create_proposal<SCT, SBT>(
```
JoeHowarth: Consider breaking this function into multiple helpers as it's complexity is high

monad-eth-txpool/src/tracked/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+            }) = tx_heap.pop()
+            else {
+                break;
+            };
+
+            let sender = best_tx.sender();
+
+            // If this transaction will take us out of the block limit, we cannot include it.
+            // This will create a nonce gap, so we no longer want to consider this account, but
+            // we want to continue considering other accounts.
```
JoeHowarth: then shouldn't we remove this address from tx_iters?

monad-eth-txpool/src/transaction.rs
```diff
@@ -1,52 +1,104 @@
...
+
+        Ok(Self {
+            tx,
+            gas_per_gas_limit,
+            carriage_cost,
+        })
+    }
+
+    pub fn apply_reserve_balance(
+        self,
```
JoeHowarth: Why does this take self instead of &self? The balance is not included in the struct and no mutation occurs?

monad-eth-txpool/src/tracked/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+            // update this when base fee is included in consensus proposal
+            if best_tx.max_fee_per_gas() < 1000 {
+                continue;
+            }
+
+            let reserve_balance = reserve_balances
+                .get_mut(&sender)
+                .expect("address in reserve_balances");
+
+            match best_tx.to_owned().apply_reserve_balance(reserve_balance) {
```
JoeHowarth: If apply_reserve_balance didn't take self, we wouldn't have to clone here

monad-eth-txpool/src/tracked/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+
+        let mut tx_heap = BinaryHeap::<
+            VirtualValidEthTransaction,
+            heapless::binary_heap::Max,
+            MAX_ADDRESSES,
+        >::new();
+
+        let mut virtual_time = 0u64;
+
+        let mut tx_iters = self
```
JoeHowarth: This feels like it should be a helper with a descriptive name

monad-eth-txpool/src/tracked/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+        if self.is_empty() {
+            return Ok(());
+        }
+
+        let addresses = self.txs.len();
+        let _timer = DropTimer::start(Duration::ZERO, |elapsed| {
+            debug!(?elapsed, addresses, "txpool update_nonces")
+        });
+
+        let addresses = self.txs.keys().cloned().collect_vec();
```
JoeHowarth: why are we collecting and then only calling iter() again

monad-eth-txpool/src/pending/mod.rs
andr-dev: The newly allocated map returned from split_off is not guaranteed to be of the same capacity as the original map.

monad-eth-txpool/src/transaction.rs
andr-dev: There was a previous idea to "consume" the tx if it doesn't satisfy the reserve balance so passing `self` guaranteed this. No longer in the codebase, good catch.

monad-eth-txpool/src/tracked/mod.rs
andr-dev: `tx_heap` will no longer contain a transaction with this address since the loop invariant is that there is one transaction in the heap per account. I agree that the comment is a bit misleading, I'll update it soon.

monad-eth-txpool/src/tracked/mod.rs
andr-dev: `block_policy.get_account_base_nonces(...)` two lines below takes in an iterator of address references for which it will produce a `BTreeMap<&EthAddress, u64>`. Unfortunately, this means that if we want to mutate `self.txs` while also iterating over its keys through the `BTreeMap` then we need to clone all of its keys so that we're not holding onto an immutable reference. 

I've seen this pattern in a few places so far and am not super happy with it. I've considered making another version of get_account_base_nonces that uses some iterator chaining and lifetime magic to avoid this issue but that would be in a separate PR and likely after all the txpool work is done.

monad-eth-txpool/src/lib.rs
```diff
@@ -186,12 +117,12 @@ impl EthTxPool {
...
+            .map(|(tx, sender)| {
                 self.validate_and_insert_tx(
-                    block_policy.get_chain_id(),
-                    eth_tx,
+                    tx,
+                    block_policy,
                     sender_account_balances
-                        .get(sender)
+                        .get(&sender)
                         .expect("sender_account_balances should be populated"),
```
eerkaijun: I think we should remove the expect here - better to just default to zero

monad-eth-txpool/src/tracked/mod.rs
```diff
@@ -0,0 +1,390 @@
...
+            block_policy.get_last_commit(),
+            state_backend,
+            &Vec::<&EthValidatedBlock<SCT>>::default(),
+            addresses.iter(),
+        )?;
+
+        for (address, tx_list) in to_insert {
+            let account_nonce = account_nonces
+                .get(&address)
+                .expect("address exists in nonce map");
```
eerkaijun: maybe return a StateBackendError here instead?

monad-eth-txpool/src/pending/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+            return std::mem::take(&mut self.txs);
+        }
+
+        let mut split = self.txs.split_off(count);
+        std::mem::swap(&mut split, &mut self.txs);
+
+        self.num_txs = self
+            .num_txs
+            .checked_sub(split.values().map(EthTxPendingList::num_txs).sum())
+            .expect("num txs does not underflow");
```
eerkaijun: ideally we remove `expect` statement, perhaps just return empty map

monad-eth-txpool/src/pending/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+use crate::transaction::ValidEthTransaction;
+
+mod list;
+
+const MAX_ADDRESSES: usize = 16 * 1024;
+const MAX_TXS: usize = 32 * 1024;
+
+#[derive(Clone, Debug, Default)]
+pub struct PendingEthTxMap {
+    txs: IndexMap<EthAddress, EthTxPendingList>,
```
eerkaijun: not sure if indexmap is the ideal data structure here - when we promote transactions, that means we can have many transactions from just a few addresses (regardless of insertion time and priority fee)? feels like promotion should also be related to priority fee

andr-dev: I agree that there should be some ordering to how we're promoting, but I'm planning on addressing these "nice to have" sort of features later once we have the txpool thread working because then we'll have a lot more compute budget and time to make complex decisions outside of the bft hotpath.

monad-eth-testutil/src/lib.rs
```diff
@@ -98,7 +98,18 @@ pub fn generate_block_with_txs(
     let nonces = validated_txns
         .iter()
         .map(|t| (EthAddress(t.signer()), t.nonce()))
-        .collect();
+        .fold(BTreeMap::default(), |mut map, (address, nonce)| {
```
JoeHowarth, line 101: nit: leave a comment describing what's the idea is or use a helper fn

monad-eth-txpool/src/lib.rs
```diff
@@ -1,181 +1,119 @@
...
         txs: Vec<EthTransaction>,
-    ) -> Result<Vec<Result<(), TxPoolInsertionError>>, StateBackendError>
-    where
-        SCT: SignatureCollection,
-    {
+    ) -> Result<Vec<Result<(), TxPoolInsertionError>>, StateBackendError> {
         let senders = txs.iter().map(|tx| EthAddress(tx.signer())).collect_vec();
+
         let block_seq_num = block_policy.get_last_commit() + SeqNum(1); // ?????
+
```
JoeHowarth, line 132: nit: drive by let's remove those ??? or provide a better comment

monad-eth-txpool/src/pending/list.rs
```diff
@@ -0,0 +1,50 @@
...
+    /// Returns true when the list does not contain a transaction with the same nonce and false when
+    /// an existing tx was updated.
+    pub fn try_add(&mut self, tx: ValidEthTransaction) -> Result<bool, TxPoolInsertionError> {
+        match self.nonce_map.entry(tx.nonce()) {
+            Entry::Vacant(v) => {
+                v.insert(tx);
+                Ok(true)
+            }
+            Entry::Occupied(mut existing_tx) => {
+                if existing_tx.get().cmp(&tx).is_gt() {
```
JoeHowarth: nit: prefer `>` to `cmp(..).is_gt()`

monad-eth-txpool/src/tracked/list.rs
```diff
@@ -0,0 +1,146 @@
...
+use monad_consensus_types::txpool::TxPoolInsertionError;
+use monad_eth_types::{EthAddress, Nonce};
+use tracing::error;
+
+use crate::{pending::PendingTxList, transaction::ValidEthTransaction};
+
+/// Stores byte-validated transactions alongside the an account_nonce to enforce at the type level
+/// that all the transactions in the txs map have a nonce at least account_nonce. Similar to
+/// PendingTxList, this struct also enforces non-emptyness in the txs map to guarantee that
+/// TrackedTxMap has a transaction for every address.
```
JoeHowarth, line 17: nit: "TrackedTxMap has a transaction for every address." should this be for every nonce?

monad-eth-txpool/src/tracked/list.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub fn get_queued(
+        &self,
+        pending_account_nonce: Option<u64>,
+    ) -> impl Iterator<Item = &ValidEthTransaction> {
+        let mut account_nonce = pending_account_nonce.unwrap_or(self.account_nonce);
+
+        self.txs
+            .range(account_nonce..)
+            .map_while(move |(tx_nonce, (tx, _))| {
+                debug_assert_eq!(*tx_nonce, tx.nonce());
```
JoeHowarth, line 59: Confirming this invariant is only supposed to be checked for tests?

monad-eth-txpool/src/tracked/mod.rs
```diff
@@ -0,0 +1,326 @@
...
+// proposal. This will be removed once we move the txpool into its own thread.
+const MAX_PROMOTABLE_ON_CREATE_PROPOSAL: usize = 1024 * 10;
+
+/// Stores transactions using a "snapshot" system by which each address has an associated
+/// account_nonce stored in the TrackedTxList which is guaranteed to be the correct
+/// account_nonce for the seqnum stored in last_commit_seq_num.
+#[derive(Clone, Debug)]
+pub struct TrackedTxMap<SCT, SBT> {
+    last_commit_seq_num: SeqNum,
+    txs: IndexMap<EthAddress, TrackedTxList>,
```
JoeHowarth, line 41: General comment that you can choose to disregard: When using variations on standard datastructures it can be helpful to leave an explanation of why that was used over the standard. 

For my own code, I try to use standard structures unless there's a clear reason not to since it adds cognitive overhead when reading the code. It's not clear to me why we prefer a Map<address, usize> + Vec<TrackedTxList> here instead of a Map<address, TrackedTxList>, though there could be a very good reason for it!

monad-eth-txpool/src/tracked/mod.rs
```diff
@@ -0,0 +1,326 @@
...
+            error!(
+                block_policy_last_commit = block_policy.get_last_commit().0,
+                txpool_last_commit = self.last_commit_seq_num.0,
+                "last commit update does not match block policy last commit"
+            );
+
+            return Ok(FullTransactionList::empty());
+        }
+
+        let _timer = DropTimer::start(Duration::ZERO, |elapsed| {
```
JoeHowarth, line 113: nit: remove the let (unless this causes a clippy warn otherwise)

monad-eth-txpool/src/tracked/mod.rs
```diff
@@ -0,0 +1,326 @@
...
+    }
+
+    fn create_proposal_tx_list(
+        &self,
+        tx_limit: usize,
+        proposal_gas_limit: u64,
+        tx_heap: TrackedTxHeap<'_>,
+        mut account_balances: BTreeMap<&EthAddress, u128>,
+    ) -> Result<(u64, Vec<EthTransaction>), StateBackendError> {
+        assert!(tx_limit > 0);
```
JoeHowarth, line 237: elsewhere we use a debug_assert, should that be used here too?

monad-eth-txpool/src/pending/list.rs
```diff
@@ -0,0 +1,50 @@
...
+
+use monad_consensus_types::txpool::TxPoolInsertionError;
+use monad_eth_types::Nonce;
+
+use crate::transaction::ValidEthTransaction;
+
+/// This struct ensures at the type level that nonce_map is never empty, a property used to enforce
+/// that every address in the PendingTxMap has an associated pending transaction.
+#[derive(Clone, Debug, Default)]
+pub struct PendingTxList {
```
JoeHowarth, line 11: General comment, can disregard:
When making custom datastructures it's usually a good idea to write some simple unit tests in the file.

monad-eth-txpool/src/transaction.rs
```diff
@@ -1,52 +1,106 @@
...
 
-pub type VirtualTimestamp = u64;
+impl ValidEthTransaction {
+    pub fn validate(
+        tx: EthTransaction,
+        block_policy: &EthBlockPolicy,
+    ) -> Result<Self, TxPoolInsertionError> {
+        // TODO(andr-dev): Block base fee is hardcoded to 1000 in monad-ledger, we need to update
+        // this logic once its included in the consensus proposal
+        if tx.max_fee_per_gas() < 1000 {
```
JoeHowarth, line 25: Base fee is no longer hardcoded as 1000

monad-eth-txpool/src/transaction.rs
```diff
@@ -1,52 +1,106 @@
...
+        &self,
+        account_balance: &Balance,
+    ) -> Result<Balance, TxPoolInsertionError> {
+        let Some(new_account_balance) = account_balance.checked_sub(self.txn_fee) else {
+            trace!(
+                "AccountBalance insert_tx 2 \
+                            do not add txn to the pool. insufficient balance: {:?} \
+                            txn_fee: {:?} \
+                            for address: {:?}",
+                account_balance,
```
JoeHowarth: drive by: please include these as params instead

monad-eth-txpool/src/tracked/mod.rs
```diff
@@ -0,0 +1,326 @@
...
+    pub fn promote_pending(
+        &mut self,
+        block_policy: &EthBlockPolicy,
+        state_backend: &SBT,
+        pending: &mut PendingTxMap,
+        max_promotable: usize,
+    ) -> Result<(), StateBackendError> {
+        if self.last_commit_seq_num.0 == 0 {
+            return Ok(());
+        }
```
eerkaijun: why do we need to exit if last commit seqnum is 0?

monad-eth-txpool/src/tracked/list.rs
andr-dev: For every key (address) in the TrackedTxMap, there is at least one transaction in the value (TrackedTxList).

monad-eth-txpool/src/tracked/list.rs
andr-dev: We store txs keyed by nonce, so this is just debug checking that the key nonce matches the value nonce. I would be happy changing this to an assert_eq!(...) to enforce correctness.

monad-eth-txpool/src/tracked/mod.rs
andr-dev: We use `last_commit_seq_num == 0` as a marker that the tracked pool is not in a ready state. This is required for statesync and reboots. I'm considering making last_commit_seq_num an Option<u64> as that would be clearer.

monad-eth-txpool/src/transaction.rs
andr-dev: This change currently lives in the release-testnet branch. I think it's best to merge it in afterwards as it introduces changes across multiple files that aren't directly related to core txpool function.

monad-eth-txpool/src/tracked/mod.rs
```diff
@@ -0,0 +1,340 @@
...
+
+        let addresses = to_insert.len();
+        DropTimer::start(Duration::ZERO, |elapsed| {
+            debug!(?elapsed, addresses, "txpool promote_pending")
+        });
+
+        let addresses = to_insert.keys().cloned().collect_vec();
+
+        let account_nonces = block_policy.get_account_base_nonces(
+            last_commit_seq_num + SeqNum(1),
```
omegablitz, line 201: add comment explaining the SeqNum(1), similar to how you did so above

monad-eth-txpool/src/tracked/mod.rs
```diff
@@ -0,0 +1,340 @@
...
+    ) -> Result<(u64, Vec<EthTransaction>), StateBackendError> {
+        assert!(tx_limit > 0);
+
+        let mut txs = Vec::new();
+        let mut total_gas = 0u64;
+
+        tx_heap.drain_in_order_while(|sender, tx| {
+            if total_gas
+                .checked_add(tx.gas_limit())
+                .expect("total gas does not overflow")
```
omegablitz: Is this assert safe? Can someone pass an arbitrarily high gas limit?

andr-dev: Good catch, changed to skip the tx if it can't add the gas limit to the total gas.

monad-consensus-types/src/txpool.rs
```diff
@@ -61,6 +62,8 @@ where
 
     /// Reclaims memory used by internal TxPool datastructures
     fn clear(&mut self);
+
+    fn reset(&mut self, last_delay_committed_blocks: Vec<&BPT::ValidatedBlock>);
```
ariqchowdhury: what does reset mean for this trait? What's the difference between clearing a txpool and resetting a txpool

monad-eth-txpool/src/pending/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+use monad_eth_types::EthAddress;
+
+pub use self::list::PendingTxList;
+use crate::transaction::ValidEthTransaction;
+
+mod list;
+
+const MAX_ADDRESSES: usize = 16 * 1024;
+const MAX_TXS: usize = 64 * 1024;
+const PROMOTE_TXS_WATERMARK: usize = 48 * 1024;
```
ariqchowdhury: should TXS_WATERMARK be a function of MAX_TXS? At least it should be compile time checked to be less than it.

monad-eth-txpool/src/pending/mod.rs
```diff
@@ -0,0 +1,91 @@
+use indexmap::IndexMap;
+use monad_consensus_types::txpool::TxPoolInsertionError;
+use monad_eth_types::EthAddress;
+
+pub use self::list::PendingTxList;
+use crate::transaction::ValidEthTransaction;
+
+mod list;
+
+const MAX_ADDRESSES: usize = 16 * 1024;
```
ariqchowdhury, line 12: where does this restriction come from? How should we think about choosing this value

monad-eth-txpool/src/tracked/list.rs
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        let now = Instant::now();
+
+        match self.txs.entry(tx.nonce()) {
+            btree_map::Entry::Vacant(v) => {
+                v.insert((tx, now));
+            }
+            btree_map::Entry::Occupied(mut entry) => {
+                let (existing_tx, existing_tx_insert) = entry.get();
```
ariqchowdhury: nit: existing_tx_insert_time

monad-eth-txpool/src/tracked/list.rs
```diff
@@ -0,0 +1,146 @@
...
+
+                if !tx_expired(existing_tx_insert, &now) && &tx < existing_tx {
+                    return Err(TxPoolInsertionError::ExistingHigherPriority);
+                }
```
ariqchowdhury: maybe subjective, but I would find the following more readable
```
if tx_expired(existing) || tx > existing_tx {
   insert()
} else {
   return Err(ExistingHigher...)
}
```

monad-eth-txpool/src/tracked/list.rs
```diff
@@ -0,0 +1,146 @@
...
+            return None;
+        }
+
+        Some(this)
+    }
+}
+
+fn tx_expired(tx_insert: &Instant, now: &Instant) -> bool {
+    &tx_insert
+        .checked_add(Duration::from_secs(5))
```
ariqchowdhury: 5s tx expiry should probably be a config parameter for the TrackedList. We can use a const for now but it should probably become a top level node config.

monad-eth-txpool/src/tracked/mod.rs
```diff
@@ -0,0 +1,344 @@
...
+use tracing::{debug, error, info, trace};
+use tx_heap::TrackedTxHeapDrainAction;
+
+use self::{list::TrackedTxList, tx_heap::TrackedTxHeap};
+use crate::{pending::PendingTxMap, transaction::ValidEthTransaction};
+
+mod list;
+mod tx_heap;
+
+const MAX_ADDRESSES: usize = 16 * 1024;
```
ariqchowdhury: where does this restriction come from? How should we think about choosing this value

monad-consensus-types/src/txpool.rs
andr-dev: Reset is used similar to block_policy.reset which is called whenever the node falls out of sync to ensure that the nonce tracking logic in the txpool is correct. I'll add a comment to clarify.

ariqchowdhury:
https://github.com/category-labs/category-internal/issues/1069
Let's resolve this after this merges. Its already released as is into testnet 0.7.0 so can resolve after merged to master

PR #1173, opened 2024-10-16
Graceful rpc error handling

eerkaijun:
Closes https://github.com/monad-crypto/monad-internal/issues/553 We want to make sure RPC doesn't crash in production, even if there are potential issues. Forwarding error logs to Loki instead.

Remaining crashing points listed here: https://www.notion.so/monad-labs/Scenario-where-RPC-could-crash-1211413dee6e80d6abadcec127f34484 which are initialization issues, which I think make sense to not let RPC start properly if certain parameters are not well formed.

### DIFF
--- a/monad-rpc/src/main.rs
+++ b/monad-rpc/src/main.rs
@@ -36,7 +36,7 @@ use trace::{
     monad_trace_block, monad_trace_call, monad_trace_callMany, monad_trace_get,
     monad_trace_transaction,
 };
-use tracing::{debug, info};
+use tracing::{debug, info, warn};
 use tracing_subscriber::{
     fmt::{format::FmtSpan, Layer as FmtLayer},
     layer::SubscriberExt,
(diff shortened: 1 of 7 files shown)


monad-rpc/src/main.rs
```diff
@@ -665,7 +665,9 @@ async fn main() -> std::io::Result<()> {
             .expect("failed to create ipc sender");
 
         while let Ok(tx) = ipc_receiver.recv_async().await {
-            sender.send(tx).await.expect("IPC send failed");
+            if let Err(e) = sender.send(tx).await {
+                warn!("IPC send failed: {}", e);
+            }
```
eerkaijun: I think if the IPC sender fails, the RPC server should still function as usual for other read-only requests, but would love to get people's thought on this.

monad-rpc/src/mempool_tx.rs
```diff
@@ -91,7 +91,9 @@ impl SocketWatcher {
         let mut watcher = notify::INotifyWatcher::new(
             move |res| {
                 block_on(async {
-                    tx.send(res).expect("inotify channel");
+                    if let Err(e) = tx.send(res) {
+                        warn!("inotify channel: {}", e);
+                    }
```
eerkaijun: Same as above

monad-rpc/src/main.rs
ariqchowdhury: maybe extend the log message to add that monad-bft has likely crashed

PR #1174, opened 2024-10-16
Parallelize blocksync

bamrith:


monad-state/Cargo.toml
```diff
@@ -11,6 +11,7 @@ bench = false
 [dependencies]
 monad-async-state-verify = { path = "../monad-async-state-verify" }
 monad-blocktree = { path = "../monad-blocktree" }
+monad-blocksync = { path = "../monad-blocksync" }
```
omegablitz, line 13: sort

monad-state/src/statesync.rs
```diff
@@ -161,27 +162,45 @@ impl<SCT: SignatureCollection> BlockBuffer<SCT> {
...
             return None;
         }
 
         let chain = self.root_parent_chain(root);
 
         let Some(last) = chain.last() else {
-            return Some(root.block_id);
+            return Some(BlockRange {
+                last_block_id: root.block_id,
+                num_blocks: 1,
```
omegablitz: revisit this, request all at once

monad-state/src/statesync.rs
```diff
@@ -161,27 +162,45 @@ impl<SCT: SignatureCollection> BlockBuffer<SCT> {
...
+            num_blocks = num_blocks.min(BLOCKSYNC_MAX_BATCH_SIZE);
+            Some(BlockRange {
+                last_block_id: block_parent_id,
+                num_blocks,
+            })
+        } else if (block_parent_id != GENESIS_BLOCK_ID) && block_is_empty {
+            // if block is empty, keep requesting until we hit non-empty
+            Some(BlockRange {
+                last_block_id: block_parent_id,
+                num_blocks: 1,
```
omegablitz: revisit this when we switch add sequence number range

monad-state/src/lib.rs
omegablitz, line 1: There's a `for (sender, proposal) in cached_proposals.into_iter().rev() {` in this file
After your changes, _should_ be able to remove the `.into_iter().rev()` part, because that was special-cased specifically for timestamp validation

monad-executor-glue/src/lib.rs
```diff
@@ -242,8 +258,8 @@ pub enum ConsensusEvent<ST, SCT: SignatureCollection> {
     /// a block that was previously requested
     /// this is an invariant
     BlockSync {
-        block: Block<SCT>,
-        payload: Payload,
+        block_range: BlockRange,
+        full_blocks: Vec<FullBlock<SCT>>,
```
omegablitz, line 262: Maybe could be changed to EthValidated block type? Might be annoying though if needs to be parameterized by BVT/BPT...

Can revisit later if so

monad-ledger/src/lib.rs
```diff
@@ -296,9 +329,6 @@ where
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         self.fetches.poll_recv(cx).map(|response| {
             let response = response.expect("fetches_tx never dropped");
-            if let BlockSyncResponseMessage::BlockFound(block) = &response {
```
omegablitz, line 299: If I'm reading this right, we no longer will hydrate the cache on init with this change. Is this expected?

monad-consensus-state/src/lib.rs
```diff
@@ -1016,10 +999,29 @@ where
         &mut self,
         validated_block: &BPT::ValidatedBlock,
         state_root_action: StateRootAction,
+        original_round: Round,
```
omegablitz: remove original_round, move TimestampUpdate branching to handle_proposal?

monad-consensus-state/src/lib.rs
```diff
@@ -1316,19 +1318,28 @@ where
...
             return Vec::new();
         }
 
-        vec![ConsensusCommand::RequestSync {
-            block_id: qc.get_block_id(),
-        }]
+        let root_seq_num = self.consensus.pending_block_tree.get_root_seq_num();
+
+        let mut num_blocks = (qc.get_seq_num() - root_seq_num).0 as u32;
+        num_blocks = num_blocks.min(BLOCKSYNC_MAX_BATCH_SIZE);
```
omegablitz: move BLOCKSYNC_MAX_BATCH_SIZE logic to blocksync impl, whole range can be requested ahead-of-time

omegablitz:
Verify that requests and repsonses are TCP

omegablitz:
Make sure blocksync timeout is sensible

monad-blocksync/src/blocksync.rs
```diff
@@ -0,0 +1,1648 @@
...
+use rand_chacha::ChaCha8Rng;
+use tracing::trace;
+
+use crate::messages::message::{
+    BlockSyncHeadersResponse, BlockSyncPayloadResponse, BlockSyncRequestMessage,
+    BlockSyncResponseMessage,
+};
+
+// TODO configurable
+pub const BLOCKSYNC_MAX_BATCH_SIZE: u32 = 10;
```
omegablitz: this should be an impl detail of this module instead of used externally

monad-blocksync/src/blocksync.rs
```diff
@@ -0,0 +1,1648 @@
...
+                self.handle_payload_response_for_self(Some(sender), payload_response)
+            }
+        }
+    }
+
+    pub fn handle_timeout(
+        &mut self,
+        request: BlockSyncRequestMessage,
+    ) -> Vec<BlockSyncCommand<SCT>> {
+        trace!("blocksync timeout: {:?}", request);
```
omegablitz: prefer trace!(?request,"blocksync timeout");

monad-blocksync/src/blocksync.rs
```diff
@@ -0,0 +1,1648 @@
...
+        }
+
+        cmds
+    }
+
+    pub fn handle_self_response(
+        &mut self,
+        response: BlockSyncResponseMessage<SCT>,
+    ) -> Vec<BlockSyncCommand<SCT>> {
+        trace!("self response from ledger: {:?}", response);
```
omegablitz: prefer trace!(?response,"self response from ledger");

monad-blocksync/src/blocksync.rs
```diff
@@ -0,0 +1,1648 @@
...
+            ));
+        }
+
+        let self_requester = self_request.requester;
+        match payload_response {
+            BlockSyncPayloadResponse::Found((payload_id, payload)) => {
+                assert_eq!(self_requester, self.block_sync.self_request_mode);
+
+                if payload_id == payload.get_id() {
+                    trace!("received valid payload response for {:?}", payload_id);
```
omegablitz: ?payload_id

monad-blocksync/src/blocksync.rs
```diff
@@ -0,0 +1,1648 @@
...
+                    });
+                    cmds.push(BlockSyncCommand::ScheduleTimeout(
+                        BlockSyncRequestMessage::Headers(block_range),
+                    ));
+                }
+            }
+            BlockSyncHeadersResponse::NotAvailable(block_range) => {
+                self.metrics.blocksync_events.headers_response_failed += 1;
+                trace!(
+                    "headers response not available: from: {:?}, block range {:?}",
```
omegablitz: ?sender, ?block_range

monad-blocksync/src/blocksync.rs
```diff
@@ -0,0 +1,1648 @@
...
+
+                // verify the headers
+                // TODO: do we need to validate headers if response is from self ledger ?
+                if Self::verify_block_headers(
+                    self.epoch_manager,
+                    self.val_epoch_map,
+                    block_range,
+                    block_headers.as_slice(),
+                ) {
+                    trace!(
```
omegablitz, line 429: ?sender, ?block_range

monad-blocksync/src/blocksync.rs
```diff
@@ -0,0 +1,1648 @@
...
+                        .entry(block_range)
+                        .or_default();
+                    entry.insert(sender, Vec::new());
+                    cmds.push(BlockSyncCommand::FetchHeaders(block_range));
+                }
+            }
+            BlockSyncRequestMessage::Payload(payload_id) => {
+                self.metrics.blocksync_events.peer_payload_request += 1;
+                trace!(
+                    "peer payload request: from {:?}, payload id: {:?}",
```
omegablitz: ?sender, ?payload_id

monad-blocksync/src/blocksync.rs
```diff
@@ -0,0 +1,1648 @@
...
+        &mut self,
+        sender: NodeId<SCT::NodeIdPubKey>,
+        request: BlockSyncRequestMessage,
+    ) -> Vec<BlockSyncCommand<SCT>> {
+        let mut cmds = Vec::new();
+
+        match request {
+            BlockSyncRequestMessage::Headers(block_range) => {
+                self.metrics.blocksync_events.peer_headers_request += 1;
+                trace!(
```
omegablitz: ?sender, ?block_range

monad-blocksync/src/blocksync.rs
```diff
@@ -0,0 +1,1648 @@
...
+    BPT: BlockPolicy<SCT, SBT>,
+    SBT: StateBackend,
+    VTF: ValidatorSetTypeFactory<NodeIdPubKey = CertificateSignaturePubKey<ST>>,
+{
+    pub fn handle_self_request(
+        &mut self,
+        requester: BlockSyncSelfRequester,
+        block_range: BlockRange,
+    ) -> Vec<BlockSyncCommand<SCT>> {
+        trace!(
```
omegablitz: ?requester, ?block_range

monad-blocksync/src/blocksync.rs
```diff
@@ -0,0 +1,1648 @@
...
+        if block_range.last_block_id != block_headers.last().unwrap().get_id() {
+            return false;
+        }
+
+        // verify that the headers form a chain by verifying the QCs and
+        // their parent block ids
+        for (index, block_header) in block_headers.iter().enumerate() {
+            let qc = &block_header.qc;
+            // verify the QC
+            if let Err(_err) = verify_certificates(epoch_manager, val_epoch_map, &None, qc) {
```
omegablitz: Are we certain that we want to be verifying QCs here? Intuitively, I would think blocksync is only responsible for returning back a valid chain of blocks given a BlockId

monad-consensus/src/validation/signing.rs
```diff
@@ -477,7 +477,7 @@ fn validate_commit_result(vote: &Vote) -> Result<(), Error> {
     }
 }
 
-fn verify_certificates<SCT, VTF, VT>(
+pub fn verify_certificates<SCT, VTF, VT>(
```
omegablitz: are we sure we want to do this in blocksync?

monad-blocksync/src/blocksync.rs
```diff
@@ -0,0 +1,1648 @@
...
+                    block_range
+                );
+
+                if block_range.num_blocks > BLOCKSYNC_MAX_BATCH_SIZE {
+                    // BLOCKSYNC_MAX_BATCH_SIZE can be different for peers
+                    // TODO: respond with as many blocks as self allows
+                    return cmds;
+                }
+
+                if let Some(blocktree) = self.maybe_blocktree {
```
omegablitz: i think this branch could be consolidated to simplify the code and make it easier to read.

something like:
```
let mut cached_blocks = self.maybe_blocktree
    .map(|blocktree| blocktree.get_parent_block_chain(&block_range.last_block_id))
    .iter()
    .flatten()
    .map(|validated_block| validated_block.get_unvalidated_block_ref().clone())
    .collect_vec();
```

monad-blocksync/src/blocksync.rs
```diff
@@ -0,0 +1,1648 @@
...
+
+impl<'a, ST, SCT, BPT, SBT, VTF> BlockSyncWrapper<'a, ST, SCT, BPT, SBT, VTF>
+where
+    ST: CertificateSignatureRecoverable,
+    SCT: SignatureCollection<NodeIdPubKey = CertificateSignaturePubKey<ST>>,
+    BPT: BlockPolicy<SCT, SBT>,
+    SBT: StateBackend,
+    VTF: ValidatorSetTypeFactory<NodeIdPubKey = CertificateSignaturePubKey<ST>>,
+{
+    pub fn handle_self_request(
```
omegablitz, line 167: Add `#[must_use]` around all the pub functions returning Vec<BlockSyncCommand<_>>

monad-blocksync/src/blocksync.rs
```diff
@@ -0,0 +1,1648 @@
...
+                }
+                assert_eq!(self_request.requester, self.block_sync.self_request_mode);
+
+                // verify the headers
+                // TODO: do we need to validate headers if response is from self ledger ?
+                if Self::verify_block_headers(
+                    self.epoch_manager,
+                    self.val_epoch_map,
+                    block_range,
+                    block_headers.as_slice(),
```
omegablitz: verify_block_headers asserts this stuff:
```
// atleast one block header was requested and received
assert!(block_range.num_blocks > 0);
// TODO: should be relaxed when nodes can respond with partial ranges
assert_eq!(block_range.num_blocks, num_blocks);
```
header_response is untrusted though - it's called in handle_peer_response

monad-blocksync/src/blocksync.rs
```diff
@@ -0,0 +1,1648 @@
...
+        val_epoch_map: &ValidatorsEpochMapping<VTF, SCT>,
+        block_range: BlockRange,
+        block_headers: &[Block<SCT>],
+    ) -> bool {
+        let num_blocks = block_headers.len() as u32;
+
+        // atleast one block header was requested and received
+        assert!(block_range.num_blocks > 0);
+        // TODO: should be relaxed when nodes can respond with partial ranges
+        assert_eq!(block_range.num_blocks, num_blocks);
```
omegablitz: input to this is untrusted?

monad-blocksync/src/blocksync.rs
```diff
@@ -0,0 +1,1648 @@
...
+
+        // The id of the last header must be block_range.last_block_id
+        if block_range.last_block_id != block_headers.last().unwrap().get_id() {
+            return false;
+        }
+
+        // verify that the headers form a chain by verifying the QCs and
+        // their parent block ids
+        for (index, block_header) in block_headers.iter().enumerate() {
```
omegablitz: perhaps simpler to use `block_headers.iter().zip(block_headers[1..].iter())`

monad-blocksync/src/blocksync.rs
```diff
@@ -0,0 +1,1648 @@
...
+        true
+    }
+
+    fn initiate_payload_requests(
+        &mut self,
+        block_range: BlockRange,
+        block_headers: Vec<Block<SCT>>,
+    ) -> Vec<BlockSyncCommand<SCT>> {
+        let mut cmds = Vec::new();
+        trace!("self initiating {} payload requests", block_headers.len());
```
omegablitz: trace!(num_block_headers = block_headers.len(), "self initiating payload requests");

monad-blocksync/src/blocksync.rs
```diff
@@ -0,0 +1,1648 @@
...
+        let Entry::Occupied(mut entry) = self.block_sync.self_headers_requests.entry(block_range)
+        else {
+            // unexpected respose. could be because the self request was cancelled
+            return cmds;
+        };
+        let self_request = entry.get_mut();
+
+        match headers_response {
+            BlockSyncHeadersResponse::Found((block_range, block_headers)) => {
+                if self_request.to != sender {
```
omegablitz, line 420: remove, maybe can still bump metric though

monad-blocksync/src/blocksync.rs
```diff
@@ -0,0 +1,1648 @@
...
+    ) -> Vec<BlockSyncCommand<SCT>> {
+        let mut cmds = Vec::new();
+        trace!("self initiating {} payload requests", block_headers.len());
+
+        if self
+            .block_sync
+            .self_completed_headers_requests
+            .contains_key(&block_range)
+        {
+            // payload requests already initiated, nothing to do
```
omegablitz: when does this happen?

monad-blocksync/src/blocksync.rs
```diff
@@ -0,0 +1,1648 @@
...
+            return cmds;
+        }
+
+        for block_header in block_headers.iter() {
+            let payload_id = block_header.get_payload_id();
+
+            // TODO: check blocktree and not request existing payloads
```
omegablitz: is this todo easy to implement? it would be nice to roll that into this PR, this will happen commonly

monad-blocksync/src/blocksync.rs
omegablitz: is this if the same range is requested twice? if that should be dedup'd externally (eg consensus_state), add a warn here

monad-blocksync/src/blocksync.rs
```diff
@@ -0,0 +1,1648 @@
...
+                    cmds.push(BlockSyncCommand::ResetTimeout(
+                        BlockSyncRequestMessage::Headers(block_range),
+                    ));
+
+                    // start payload requests
+                    cmds.extend(self.initiate_payload_requests(block_range, block_headers));
+                } else {
+                    // failed header verification
+                    trace!(
+                        "headers response verification failed: sender: {:?}, {:?}",
```
omegablitz: ?sender, ?block_range

monad-blocksync/src/blocksync.rs
```diff
@@ -0,0 +1,1648 @@
...
+
+        let Entry::Occupied(mut entry) = self.block_sync.self_payload_requests.entry(payload_id)
+        else {
+            // unexpected response
+            self.metrics.blocksync_events.payload_response_unexpected += 1;
+            return cmds;
+        };
+
+        let self_request = entry.get_mut();
+        if self_request.to != sender {
```
omegablitz, line 551: consider removing

monad-proto/proto/blocksync.proto
```diff
@@ -0,0 +1,49 @@
...
+  oneof headers_response {
+    ProtoBlockSyncHeaders headers_found = 1;
+    ProtoBlockRange not_available = 2;
+  }
+}
+  
+// no rust type
+// conversion wrapped into ProtoBlockSyncPayloadResponse
+message ProtoBlockSyncPayload {
+  monad_proto.basic.ProtoPayloadId payload_id = 1;
```
omegablitz: can you remove payload_id?

omegablitz: i guess you could make the argument that this helps clarify the sender's intent in the case of version upgrades/downgrades

monad-blocksync/src/blocksync.rs
```diff
@@ -0,0 +1,1648 @@
...
+                            if block.payload_id == payload_id && maybe_payload.is_none() {
+                                // clone incase there are multiple requests that require the same payload
+                                *maybe_payload = Some(payload.clone());
+                            }
+                        }
+                    }
+                } else {
+                    // self response shouldn't fail payload validation
+                    assert!(sender.is_some());
+                    trace!("received inavlid payload response for {:?}", payload_id);
```
omegablitz: typo "inavlid", ?payload_id

monad-blocksync/src/blocksync.rs
```diff
@@ -0,0 +1,1648 @@
...
+                let block_range = headers_response.get_block_range();
+
+                // reply to the requested peers
+                let requesters = self
+                    .block_sync
+                    .headers_requests
+                    .remove(&block_range)
+                    .unwrap_or_default();
+                for (requester, cached_blocks) in requesters {
+                    let headers_response = if cached_blocks.is_empty() {
```
omegablitz: i think this branch can be consolidated

monad-blocksync/src/blocksync.rs
bamrith: I think you're right. changing it to verify just the parent block_id. will take a closer look when writing more tests.

monad-blocksync/src/blocksync.rs
bamrith: yes, it's a check for when the same range is requested twice. this should never happen. added a warning. (it could possibly be asserted too)

monad-consensus/src/validation/signing.rs
bamrith: removed

monad-proto/proto/blocksync.proto
bamrith: could remove it and generate the payload_id when converting to rust type ? (same as block_id)

monad-ledger/src/lib.rs
bamrith: we never hydrated the cache on init, we used to add a block to the cache when a read was done (i.e. when a request was made for it). this change just moved hydrating the cache when the block is committed (and keep a window of the most recent committed blocks in cache).

there aren't any issues on init because if the block isn't in the cache, it will go read it from blockpersist anyway. we can remove the "read from blockpersist" part if we hydrate the cache on init. will make an issue for that

omegablitz: What I meant is that on init, we blocksync the last ~256 blocks (hitting disk if it exists locally). With the old code, if the block exists in disk locally, it then gets added to the blocksync cache. Whereas now it only gets added to the cache via the commit path.

Maybe that's fine though... I think statesync re-commits the last ~256 blocks after it's done statesync'ing, so it might end up getting added to the cache in a roundabout way anyways

bamrith: ahh, I misunderstood. okay, I'll check that as well

monad-consensus-types/src/block.rs
```diff
@@ -14,6 +14,21 @@ use crate::{
     state_root_hash::StateRootHash,
 };
 
+/// Represent a range of blocks the last of which is `last_block_id` and includes
+/// `num_blocks` that extended to the last_block_id including itself
```
omegablitz: update comment

monad-blocksync/src/messages/message.rs
```diff
@@ -0,0 +1,152 @@
...
+                EnumDiscriminant(2).hash(state);
+                block_range.hash(state);
+            }
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum BlockSyncPayloadResponse {
+    Found((PayloadId, Payload)),
```
omegablitz: Can you change the type of this to `Found(Payload)`?

Seems less error-prone - right now the PayloadId here does not necessarily match up with the Payload, and must be checked.

monad-proto/proto/blocksync.proto
omegablitz: Let's just remove it so that it's consistent with how we do BlockId. I can see this causing confusion down the line otherwise.

monad-state/src/statesync.rs
```diff
@@ -161,27 +161,37 @@ impl<SCT: SignatureCollection> BlockBuffer<SCT> {
...
+    pub fn needs_blocksync(&self, root: &RootInfo) -> Option<BlockRange> {
         if root.seq_num == GENESIS_SEQ_NUM {
             return None;
         }
 
         let chain = self.root_parent_chain(root);
 
         let Some(last) = chain.last() else {
-            return Some(root.block_id);
+            return Some(BlockRange {
```
omegablitz, line 181: any reason not to request the full range here?

monad-state/src/blocksync.rs
```diff
@@ -325,17 +139,15 @@ where
         match wrapped.command {
             BlockSyncCommand::SendRequest { to, request } => {
                 vec![Command::RouterCommand(RouterCommand::Publish {
-                    target: RouterTarget::PointToPoint(to),
+                    target: RouterTarget::TcpPointToPoint(to),
```
omegablitz: I think requests could stay over raptorcast because they're very small

monad-blocksync/src/blocksync.rs
```diff
@@ -0,0 +1,1731 @@
...
+use tracing::trace;
+
+use crate::messages::message::{
+    BlockSyncHeadersResponse, BlockSyncPayloadResponse, BlockSyncRequestMessage,
+    BlockSyncResponseMessage,
+};
+
+// TODO configurable
+// determines the max number of parallel payload requests self can make
+pub const BLOCKSYNC_MAX_PAYLOAD_REQUESTS: usize = 10;
```
omegablitz: make this private now that it's not used externally

monad-blocksync/src/blocksync.rs
```diff
@@ -0,0 +1,1731 @@
...
+
+        // The id of the last header must be block_range.last_block_id
+        if block_range.last_block_id != block_headers.last().unwrap().get_id() {
+            return false;
+        }
+
+        // verify that the headers form a chain by verifying the QCs point to
+        // their parent block ids
+        for (parent_block_header, block_header) in
+            block_headers.iter().zip(block_headers[1..].iter())
```
omegablitz: I know I suggested this, but I'm actually not sure what happens if block_headers.len() == 1

can switch it to the following if that breaks:
block_headers.iter().zip(block_headers.iter().skip(1))

monad-ledger/src/lib.rs
```diff
@@ -219,22 +225,51 @@ where
...
+        }
 
-        match maybe_response {
-            Ok(full_block) => BlockSyncResponseMessage::BlockFound(full_block),
-            Err(_) => BlockSyncResponseMessage::NotAvailable(block_id),
+        BlockSyncHeadersResponse::Found((block_range, headers.into()))
+    }
+
+    fn ledger_fetch_payload(&self, payload_id: PayloadId) -> BlockSyncPayloadResponse {
+        // TODO store payload_id to block_id map to avoid searching the entire cache
```
omegablitz: Can you address this now?

I think it should be simple if you split
`block_cache: HashMap<BlockId, MonadBlock<SCT>>`
into:
`block_header_cache: HashMap<BlockId, Block<SCT>>`
`block_payload_cache: HashMap<PayloadId, Payload>`
`block_cache_index: BTreeMap<Round, (BlockId, PayloadId)>`

monad-blocksync/src/blocksync.rs
```diff
@@ -0,0 +1,1731 @@
...
+
+                if payload_id == payload.get_id() {
+                    trace!(
+                        ?sender,
+                        ?payload_id,
+                        "blocksync: received valid payload response"
+                    );
+                    // valid payload, remove entry and update existing requests
+                    entry.remove();
+                    self.block_sync.self_payload_requests_in_flight -= 1;
```
omegablitz: it would be useful to track self_payload_requests_in_flight as a metric

monad-blocksync/src/blocksync.rs
bamrith: made sure that doesn't break. if length of `v` is 1, `v[1..]` returns `[]` but `v[2..]` panics. changed it to use skip anyway. it "feels safer"

monad-ledger/src/lib.rs
bamrith: done

monad-state/src/statesync.rs
bamrith: good point, i don't think so. changed.

monad-blocksync/src/blocksync.rs
bamrith: adding that in a PR after this ASAP.

PR #1176, opened 2024-10-17
Rate limit for eth_call and eth_estimateGas

eerkaijun:
Setting default rate limit for eth_call and eth_estimateGas to prevent RPC from crashing when being spammed. Default currently set as 1000 concurrent requests which is around 3k RPS when spammed under much higher load. 

Relevant issue: https://github.com/monad-crypto/monad-internal/issues/617

### DIFF
--- a/monad-rpc/src/cli.rs
+++ b/monad-rpc/src/cli.rs
@@ -45,9 +45,15 @@ pub struct Cli {
     #[arg(long, default_value_t = 25_000_000)]
     pub max_response_size: u32,
 
+    /// Otel endpoint to collect metrics data
     #[arg(long)]
     pub otel_endpoint: Option<String>,
 
+    /// Allow pre EIP-155 transactions
     #[arg(long, default_value_t = false)]
     pub allow_unprotected_txs: bool,
+
+    /// Set the max concurrent requests for eth_call and eth_estimateGas
+    #[arg(long, default_value_t = 1000)]
+    pub eth_call_max_concurrent_requests: u32,
 }
--- a/monad-rpc/src/main.rs
+++ b/monad-rpc/src/main.rs
@@ -1,63 +1,62 @@
-use std::path::PathBuf;
+use std::{path::PathBuf, sync::Arc};
 
-use account_handlers::{
-    monad_eth_getBalance, monad_eth_getCode, monad_eth_getProof, monad_eth_getStorageAt,
-    monad_eth_getTransactionCount, monad_eth_syncing,
-};
 use actix::prelude::*;
 use actix_http::body::BoxBody;
 use actix_web::{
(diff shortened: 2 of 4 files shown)


monad-rpc/src/cli.rs
```diff
@@ -45,6 +45,15 @@ pub struct Cli {
...
     #[arg(long)]
     pub otel_endpoint: Option<String>,
+
+    /// Enable or disable eth_call and eth_estimateGas rate limiter
+    #[arg(long, default_value_t = true)]
+    pub rate_limiter_enabled: bool,
+
+    /// Set the rate limit value of concurrent requests
+    #[arg(long, default_value_t = 1000)]
+    pub rate_limit_value: u32,
```
ariqchowdhury: I think calling this rate-limit is inaccurate as we're not doing anything based on rate, we're limiting the max number of concurrent requests. So I would change the names to be consistent with that. 
Also, these should be named to mention eth_call/estimateGas so its clear from the name that only those requests are limited.

eerkaijun: updated

monad-rpc/src/main.rs
```diff
@@ -255,6 +254,17 @@ async fn rpc_select(
...
+            // acquire the permit if rate limiting is enabled
+            let _permit = if let Some(rate_limiter) = &app_state.rate_limiter {
+                Some(
+                    rate_limiter
+                        .try_acquire()
+                        .map_err(|_| JsonRpcError::internal_error("eth_call rate limit".into()))?,
+                )
+            } else {
+                None
+            };
```
bamrith: can probably remove the else if permit is never used
```
if let Some(rate_limiter) = &app_state.rate_limiter {
    let _permit = rate_limiter
        .try_acquire()
        .map_err(|_| JsonRpcError::internal_error("eth_call rate limit".into()))?;
}
```

eerkaijun: updated

monad-rpc/src/cli.rs
```diff
@@ -45,9 +45,19 @@ pub struct Cli {
...
     #[arg(long)]
     pub otel_endpoint: Option<String>,
 
+    /// Allow pre EIP-155 transactions
     #[arg(long, default_value_t = false)]
     pub allow_unprotected_txs: bool,
+
+    /// Enable eth_call and eth_estimateGas concurrent requests limiter
+    #[arg(long, default_value_t = true)]
+    pub eth_call_concurrent_requests_limiter_enabled: bool,
```
philipglazman: This flag might be confusing if `eth_call_concurrent_requests_limit_value` already has a default value. Any reason to not always set the limit to the default value ?

PR #1177, opened 2024-10-17
Add support for configurable upload pacing

omegablitz:
This is particularly important for very stake-uneven networks, as a few nodes may bear the majority of the network's upload burden.

### DIFF
--- a/monad-state/src/blocksync.rs
+++ b/monad-state/src/blocksync.rs
@@ -303,7 +303,7 @@ where
         };
         cmds.into_iter()
             .map(|command| WrappedBlockSyncCommand {
-                request_timeout: *self.delta * 3,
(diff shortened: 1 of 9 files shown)


monad-state/src/blocksync.rs
```diff
@@ -345,7 +345,7 @@ where
             }
             BlockSyncCommand::SendResponse { to, response } => {
                 vec![Command::RouterCommand(RouterCommand::Publish {
-                    target: RouterTarget::PointToPoint(to),
+                    target: RouterTarget::TcpPointToPoint(to),
```
michael-yxchen, line 348: Do we want to bump blocksync request timeout if switching to tcp?

omegablitz: Good catch, fixed

PR #1181, opened 2024-10-18
Add message delay to generic executor

michael-yxchen:
ExecutorDelay adds delay to events polled from the executor. monad-node uses this executor to add latencies to router messages and pace block production. This replaces mock-gossip and removes monad-quic and monad-gossip dependencies from monad-node

### DIFF
--- a/monad-executor-delay/src/driver.rs
+++ b/monad-executor-delay/src/driver.rs
@@ -24,7 +24,7 @@ pub struct ScheduledEvent<E> {
 
 impl<E> PartialOrd for ScheduledEvent<E> {
     fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
-        Some(other.deliver.cmp(&self.deliver))
+        Some(self.cmp(other))
     }
 }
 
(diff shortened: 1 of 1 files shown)


monad-executor-delay/src/driver.rs
```diff
@@ -0,0 +1,176 @@
...
+}
+
+pub struct ScheduledEvent<E> {
+    deliver: Duration,
+    trigger: Trigger<E>,
+}
+
+impl<E> PartialOrd for ScheduledEvent<E> {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(other.deliver.cmp(&self.deliver))
```
andr-dev: I think it's common practice to call Self's Ord impl (`other.cmp(&self)`) so that the ord "logic" stays in one place.

monad-executor-delay/src/driver.rs
```diff
@@ -0,0 +1,176 @@
...
+        self.now = self.now.max(UNIX_EPOCH.elapsed().unwrap());
+        self.now
+    }
+
+    fn peek_tick(&self) -> Option<Duration> {
+        if !self.events.is_empty() {
+            Some(self.events.peek().unwrap().deliver)
+        } else {
+            None
+        }
```
andr-dev:
```rust
self.events.peek().map(|event| event.deliver)
```

omegablitz:
No longer necessary after #1234

PR #1183, opened 2024-10-18
Virtual mempool for rpc

philipglazman:
Branched off https://github.com/monad-crypto/monad-bft/pull/1170. Adds a "virtual mempool" which acts a staging area and filter for incoming transactions. Transactions enter a subpool depending if they are valid + ready for block inclusion, valid but base fee too low, or valid but a nonce gap exists. Transactions are either promoted to the pending_pool and sent to a validator or discarded. Block updates decide promotion. There is a simple eviction policy that bounds the size of the pools.


monad-rpc/src/vpool.rs
```diff
@@ -0,0 +1,817 @@
...
+}
+
+impl From<TransactionSigned> for TxEntry {
+    fn from(tx: TransactionSigned) -> Self {
+        Self {
+            transaction: tx.clone(),
+            sender: tx.recover_signer().unwrap_or_default(),
+            id: tx.hash(),
+            nonce: tx.nonce(),
+        }
```
andr-dev: I think recover_signer(), hash(), and nonce() all take &TransactionSigned so as long as we get those values first we should be able to set `transaction: tx` without the clone.

monad-rpc/src/vpool.rs
```diff
@@ -0,0 +1,817 @@
...
+    }
+
+    async fn add(&self, txn: TxEntry) {
+        if let Some(mut tree) = self.pool.get(&txn.sender) {
+            let mut sub_tree = BTreeMap::new();
+            sub_tree.insert(txn.nonce, txn.clone());
+            tree.append(&mut sub_tree);
+        } else {
+            let mut tree = BTreeMap::new();
+            tree.insert(txn.nonce, txn.clone());
```
andr-dev: We can move
```rust
let mut tree = BTreeMap::new();
tree.insert(txn.nonce, txn.clone());
```
outside of the if else to reduce code dupe.

monad-rpc/src/vpool.rs
```diff
@@ -0,0 +1,817 @@
...
+            match self.pool.entry(txn.sender) {
+                scc::hash_map::Entry::Occupied(mut entry) => {
+                    entry.insert(tree);
```
andr-dev: Do we always want to overwrite an existing tx entry? What if we get two transactions with the same nonce but one of them has a higher tip?

monad-rpc/src/vpool.rs
```diff
@@ -0,0 +1,817 @@
...
+                }
+            }
+        };
+
+        if self.evict.read().await.len() > self.capacity {
+            if let Some(evicted) = self.evict.write().await.pop_back() {
+                self.pool.remove(&evicted);
+            }
+        }
+    }
```
andr-dev: I think will cause a pretty mild TOCTTOU issue. If we only want to check this in the Entry::Vacant branch, we should move it there and hold a write() lock and share it for the push_front, len, and pop_back calls.

monad-rpc/src/vpool.rs
```diff
@@ -0,0 +1,817 @@
...
+
+    fn clear_included(&self, txs: Vec<(Address, u64)>) {
+        for (sender, nonce) in txs {
+            match self.pool.get(&sender) {
+                Some(mut entry) => {
+                    entry.get_mut().remove(&nonce);
+                    if entry.len() == 0 {
+                        drop(entry);
+                        self.pool.remove(&sender);
```
andr-dev: The entry API has a `entry.remove()` method which will remove that entry from the HashMap without having to "find" it again by calling self.pool.remove(...).

monad-rpc/src/vpool.rs
```diff
@@ -0,0 +1,817 @@
...
+
+    // Returns a list of transaction entries that are ready to be promoted because a nonce gap was resolved.
+    fn filter_by_nonce_gap(&self, state: &HashIndex<Address, u64>) -> Vec<TxEntry> {
+        let mut to_promote = Vec::new();
+        for (sender, nonce) in state.iter(&Guard::new()) {
+            if let Some(mut entry) = self.pool.get(sender) {
+                let mut removed = 0;
+                let mut nonce = nonce.clone();
+                for (min_nonce, tx) in entry.iter() {
+                    if *min_nonce > nonce && min_nonce - nonce == 1 {
```
andr-dev: `min_nonce.checked_sub(nonce) == Some(1)`

monad-rpc/src/vpool.rs
```diff
@@ -0,0 +1,817 @@
...
+                        removed += 1;
+                        nonce += 1;
+                    } else {
+                        break;
+                    }
+                }
+
+                if removed == entry.len() {
+                    drop(entry);
+                    self.pool.remove(sender);
```
andr-dev: Same as above, `entry.remove();`

monad-rpc/src/vpool.rs
```diff
@@ -0,0 +1,817 @@
...
+
+impl Ord for BaseFeeKey {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        if self.0 > other.0 {
+            std::cmp::Ordering::Less
+        } else if self.0 < other.0 {
+            std::cmp::Ordering::Greater
+        } else {
+            std::cmp::Ordering::Equal
+        }
```
andr-dev:
```rust
self.0.cmp(other.0).reverse()
```

monad-rpc/src/vpool.rs
```diff
@@ -0,0 +1,817 @@
...
+    }
+}
+
+struct ChainCache {
+    accounts: HashIndex<Address, u64>,
+    base_fee: RwLock<u128>,
+    evict: RwLock<LinkedList<(Address, u64)>>,
+    capacity: usize,
+}
```
andr-dev: This is a data structure I imagine we'd want to update / read from atomically. What's the rationale behind using different synchronization mechanisms for the first three fields?

monad-rpc/src/vpool.rs
```diff
@@ -0,0 +1,817 @@
...
+
+        if base_fee < *self.chain_cache.base_fee.read().await {
+            return TxPoolType::BaseFee;
+        }
+
+        if self
+            .chain_cache
+            .accounts
+            .get(&sender)
+            .map(|entry| nonce - entry.get() > 1)
```
andr-dev: We should use `nonce.checked_sub(entry.get())` to avoid underflow.

monad-rpc/src/vpool.rs
```diff
@@ -0,0 +1,817 @@
...
+            .unwrap_or(false)
+        {
+            return TxPoolType::Queue;
+        }
+
+        let queued = self.queued_pool.by_addr(&sender);
+        if queued.len() > 0 && queued.first().expect("expect at least one element").nonce < nonce {
```
andr-dev:
```rust
if queued.first().map(|tx| tx.nonce < nonce).unwrap_or_default() {
```

monad-rpc/src/vpool.rs
```diff
@@ -0,0 +1,817 @@
...
+            if last_pending_nonce + 1 == nonce {
+                return TxPoolType::Pending;
+            } else if last_pending_nonce >= nonce {
+                return TxPoolType::Discard;
+            } else {
+                return TxPoolType::Queue;
+            }
+        }
+
+        TxPoolType::Pending
```
andr-dev:
```rust
let Some(last_pending_nonce) = pending.last() else {
    return TxPoolType::Pending;
}

if last_pending_nonce + 1 == nonce {
   TxPoolType::Pending
} else if last_pending_nonce >= nonce {
    TxPoolType::Discard;
} else {
    TxPoolType::Queue
}
```

monad-rpc/src/vpool.rs
```diff
@@ -0,0 +1,817 @@
...
+                to_promote.push(entry.clone());
+                self.pool.remove(tx_base_fee);
+            } else {
+                break;
+            }
+        }
+        to_promote
+    }
+
+    fn add(&self, base_fee: u128, txn: TxEntry) {
```
andr-dev: If we can determine `base_fee` from `txn`, we should compute that in this function rather than rely on the caller to get it correct.

monad-rpc/src/vpool.rs
```diff
@@ -0,0 +1,817 @@
...
+                let txs = block
+                    .body
+                    .iter()
+                    .filter_map(|txn| {
+                        txn.recover_signer_unchecked()
+                            .map(|sender| (sender, txn.nonce()))
+                    })
+                    .collect::<Vec<_>>();
+
+                self.pending_pool.clear_included(txs);
```
andr-dev: If we get a block update, the txs could be in any of the three pools. I believe that means we should check for and remove them from queued_pool and base_fee_pool as well. I think queued_pool.filter_by_nonce_gap inherently does this, but I don't think we have a similar mechanism for base_fee_pool.

monad-rpc/src/vpool.rs
```diff
@@ -0,0 +1,817 @@
...
+
+                // Check for available promotions
+                let promote_queued = self.queued_pool.filter_by_nonce_gap(&senders);
+                let promote_basefee = self.base_fee_pool.promote(next_base_fee);
+
+                // Add promoted transactions to the pending pool
+                for promoted in promote_basefee
+                    .into_iter()
+                    .chain(promote_queued)
+                    .into_iter()
```
andr-dev: We can remove this extra `into_iter()`

monad-rpc/src/vpool.rs
```diff
@@ -0,0 +1,817 @@
...
+                {
+                    self.pending_pool.add(promoted.clone()).await;
+                    self.publisher.send(promoted.transaction.clone());
```
andr-dev: No need to clone `promoted.transaction`. Additionally, do we want to check the send result here?

monad-rpc/src/vpool.rs
```diff
@@ -0,0 +1,817 @@
...
+        })
+        .await
+    }
+
+    /// Returns pending + queued transactions for an address
+    pub fn pool_by_address(&self, address: &Address) -> (Vec<TxEntry>, Vec<TxEntry>) {
+        let pending = self.pending_pool.by_addr(address);
+        let queued = self.queued_pool.by_addr(address);
+        (pending, queued)
+    }
```
andr-dev: This function is never used, do we have plans for it?

monad-rpc/src/vpool.rs
```diff
@@ -0,0 +1,817 @@
...
+                        drop(entry);
+                        self.pool.remove(&sender);
+                    }
+                }
+                None => {}
+            }
+        }
+    }
+
+    fn by_addr(&self, address: &Address) -> Vec<TxEntry> {
```
andr-dev: Whenever `by_addr` is used, we only use the `nonce` field. Can we instead return `Vec<&TxEntry>`?

monad-rpc/src/vpool.rs
```diff
@@ -0,0 +1,817 @@
...
+                        ..Default::default()
+                    },
+                    ..Default::default()
+                },
+                1_000,
+            )
+            .await;
+
+        assert_eq!(tx_pool.queued_pool.pool.len(), 0);
+    }
```
andr-dev: We should move these to benches/ and make them criterion benchmarks.

monad-rpc/src/vpool.rs
philipglazman: It's expected that `txpool_contentFrom` rpc call will use this function to retrieve all transactions associated with a sender address

monad-rpc/src/vpool.rs
philipglazman: We do want to replace transactions that have a higher fee since that is a common workflow in evm chains. Made this more clear and added check for overwriting an existing tx entry.

monad-rpc/src/vpool.rs
andr-dev: Got it, in that case can we make the return `Vec<&TxEntry>` as suggested below?

monad-rpc/Cargo.toml
```diff
@@ -45,6 +45,7 @@ schemars = "0.8.21"
 opentelemetry = { workspace = true, features = ["metrics"] }
 opentelemetry-otlp = "0.17.0"
 opentelemetry_sdk = { workspace = true, features = ["rt-tokio"] }
+scc = "2.2.0"
```
ariqchowdhury: workspace cargo

monad-rpc/src/main.rs
```diff
@@ -671,6 +678,20 @@ async fn main() -> std::io::Result<()> {
         }
     });
 
+    let tx_pool = Arc::new(vpool::VirtualPool::new(
+        ipc_sender.clone(),
+        args.vpool_capacity,
+    ));
+
+    // We need to spawn a task to handle changes to the base fee, and block updates
+    let tx_pool2 = tx_pool.clone();
```
ariqchowdhury, line 648: what's the second txpool for?

monad-rpc/src/vpool.rs
```diff
@@ -0,0 +1,920 @@
...
+}
+
+#[rpc(method = "txpool_inspect")]
+#[allow(non_snake_case)]
+pub async fn monad_txpool_inspect() -> JsonRpcResult<String> {
+    Err(JsonRpcError::method_not_supported())
+}
+
+#[derive(Debug, Serialize, schemars::JsonSchema)]
+pub struct TxPoolStatus {
```
ariqchowdhury, line 43: do we need the base_fee_pool size here too?

monad-rpc/src/vpool.rs
```diff
@@ -0,0 +1,920 @@
...
+    })
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Default)]
+pub struct TxEntry {
+    transaction: TransactionSigned,
+    sender: Address,
+    id: TxHash,
+    nonce: u64,
+}
```
ariqchowdhury: nit: can the order be id, sender, nonce, transaction

monad-rpc/src/vpool.rs
andr-dev: Resolved offline, can't return ref because of concurrent datastructure.

monad-rpc/src/vpool.rs
andr-dev: Resolved offline, can't return ref because of concurrent data structure design.

monad-rpc/src/vpool.rs
andr-dev: As an aside, why is `TxEntry` a separate struct? It only implements `From<TransactionSigned>` and seems to be semantically equivalent to `reth_primitives::transaction::TransactionSignedEcRecovered` (TransactionSigned with a guaranteed recovered signer).

philipglazman: That's a better type! missed it. I'll update

monad-rpc/src/main.rs
philipglazman: removed, no reason to have it anymore

monad-rpc/src/vpool.rs
philipglazman: added basefeepool len to queue field

monad-rpc/src/vpool.rs
philipglazman: That's a good point, I'll update this so there is a consistent lock pattern

philipglazman: updated

monad-rpc/src/vpool.rs
philipglazman: I added an explicit check for the queued pool. For base fee pool, the transaction should not be included in the block update since it would be invalid to include the base fee in the block

monad-rpc/src/vpool.rs
```diff
@@ -0,0 +1,938 @@
...
+    }
+
+    async fn decide_pool(&self, txn: &TransactionSignedEcRecovered) -> TxPoolType {
+        let sender = txn.signer();
+        let nonce = txn.nonce();
+        let base_fee = txn.transaction.max_fee_per_gas();
+
+        if base_fee < self.chain_cache.get_base_fee().await {
+            return TxPoolType::BaseFee;
+        }
```
eerkaijun: if a transaction that has maxFeePerGas lower than block base fee, AND also having a nonce gap, it will gets added to the base fee pool. so if the block base fee eventually decreases, the transaction will be promoted to the pending pool (but it will still have a nonce gap since it seems like it didn't check the nonce anymore during base fee promotion). just wanted to check my understanding, so transactions with nonce gap can still be sent to the block proposers, is that the desired behaviour?

monad-rpc/src/vpool.rs
```diff
@@ -0,0 +1,938 @@
...
+            return TxPoolType::Pending;
+        };
+
+        let last_pending_nonce = entry.nonce();
+
+        if last_pending_nonce + 1 == nonce {
+            TxPoolType::Pending
+        } else if last_pending_nonce == nonce {
+            if let Some(entry) = self.pending_pool.get(&txn.signer(), &txn.nonce()) {
+                // Replace a pending transaction if the fee is at least 10% higher than the current fee
```
eerkaijun, line 329: just wondering why 10% and not any transaction that is higher than current fee

philipglazman:
~~Ran txgenerator (few-to-many) with 10k tps, got a 16% slowdown with the current implementation. Tokio's rwlock around chaincache causes an 80% slowdown so I'll need to adjust this.~~

Fixed an issue with rwlock. There's no longer a slowdown with both ordered and unordered transactions

monad-rpc/src/vpool.rs
philipglazman: I used geth as a reference. Their reasoning is to require a somewhat significant increase in the fee in order to limit spam which I found convincing enough. I'll include a comment explaining this

monad-rpc/src/vpool.rs
philipglazman: That sounds good, but need to make monad-rpc a `lib` since thats a requirement for criterion.

monad-rpc/src/vpool.rs
philipglazman: The filtering is done at best effort right now and can definitely be optimized for this edge case - you're understanding is correct. Since your review, I removed the base fee pool since dynamic eip1559 block base fees are not expected to be part of the initial implementation. If a transaction does is below the static block base fee, it will be rejected immediately.

PR #1184, opened 2024-10-21
Peer discovery boilerplate

rgarc:
This PR includes #1124 

This PR adds boilerplate for peer discovery without any observable side-effects to production nodes.
- Basic data types needed for validators to identify each other.
- `DiscoveryEvent` and `DiscoveryCommand`
- A no-op discovery executor `NopDiscovery` was added to allow instantiating a node without any side-effects.

The commit tagged (optional) can be removed without breaking build
- (optional) mockable discovery for inclusion in mock swarm relations

Future work adds a real discovery executor that fleshes out event/commands to implement what will run in production. #1179


monad-mock-swarm/tests/discovery.rs
```diff
@@ -0,0 +1,6 @@
+mod common;
+
+#[test]
+fn discovery() {
+    // TODO(rene): need reasonable testing strategy here
```
rgarc, line 5: Peer discovery will ultimately be implemented only as an executor. That being the case, is testing via mock-swarm recommended?

michael-yxchen:
Discussed offline, we're considering to have RaptorCast be generic over PeerDiscovery. 
Pros:
- Consolidate duplicate peer list/epoch validators in PeerDiscovery and RaptorCast. Peer list views are always consistent
- Minimal impact to monad-state. No new PeerDiscovery event and command types. All changes are contained in RaptorCast

Cons:
- RaptorCast need to "intercept" and process peer discovery messages. Wire format change: `Signature || ProtoMonadMessage` -> `Signature || ProtoMonadNetworkMessage(ProtoMonadMessage | ProtoPeerDiscoveryMessage)`. Need to understand the upgradability impact there. If we want a network message version, it's better to add now

PR #1186, opened 2024-10-21
Fix eth_call test errors on submodule bump

tongzhi1998:


michael-yxchen:
Do you know how it got through our CI checks?

omegablitz:
https://github.com/monad-crypto/monad-bft/blob/006f28489e676d76830d209452efbbdd23e60563/monad-cxx/CMakeLists.txt#L20-L31

tongzhi1998:
> Do you know how it got through our CI checks?

I think we don't run eth_call tests in CI? It requires cfg(triedb) being set (compile the execution submodule), which I think is something we try to avoid.

tongzhi1998:
> Approving to fix master, but ideally we can get this to fail CI. @tongzhi1998 or @eerkaijun do you know what the monad_rpc_test target is for? I see that it's currently only used in the RPC docker image.

Similar answer to Michael's question. It is because we try to avoid compiling the execution submodule in CI (which I don't exactly know why, I think @rgarc has a better idea. Probably because it's too slow). It is only when we set `ETH_CALL_TARGET=monad_rpc_test` that we would compile the execution submodule to run tests. In the CI, we set it to `mock_eth_call` to avoid compiling the submodule.

tongzhi1998:
> submodule

Correct me if I am wrong. Since `statesync` runs on CI and requires compiling execution submodule anyways, we should probably change the logic for `eth_call` to force it to compile the execution submodule too. 
And then we can get rid of mock_eth_call, mock_db, etc. stuff

omegablitz:
CI doesn't compile the execution submodule right now, it just uses headers from the execution submodule.

omegablitz:
Regardless, feel free to merge this PR now to unbreak master - we can figure out the CI stuff separately.

tongzhi1998:
> Regardless, feel free to merge this PR now to unbreak master - we can figure out the CI stuff separately.

Sounds good. Thanks!

PR #1190, opened 2024-10-22
Router level peer disc

rgarc:
Instead of implementing discovery as a router as in #1179  we will make the raptorcast router generic over a discovery executor and own it so that the raptorcast router can interact with a discovery implementation synchronously instead of having to haphazardly communicate asynchronously through events that loop back through `MonadState`.


rgarc:
closed in favor of #1210

PR #1202, opened 2024-10-23
Read block header from triedb

eerkaijun:
This PR switch `eth_getBlockByNumber` to read from triedb instead of ledger files.

### DIFF
--- a/monad-rpc/src/block_handlers.rs
+++ b/monad-rpc/src/block_handlers.rs
@@ -1,12 +1,16 @@
-use alloy_primitives::aliases::{U256, U64};
+use alloy_primitives::{
+    aliases::{U256, U64},
+    FixedBytes,
+};
+use alloy_rlp::Decodable;
 use monad_rpc_docs::rpc;
-use reth_primitives::Block as EthBlock;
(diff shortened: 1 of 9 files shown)


monad-rpc/src/block_handlers.rs
```diff
@@ -295,31 +317,46 @@ pub struct MonadEthGetBlockReceiptsResult(Vec<MonadTransactionReceipt>);
...
+        }
+        TriedbResult::Null => {
+            debug!("No block header found");
+            return Ok(None);
+        }
+        _ => {
+            return Err(JsonRpcError::internal_error(
+                "error reading block header from db".into(),
+            ))
+        }
```
ariqchowdhury, line 349: this block of code is duplicated a few times in this PR, can it be refactored

eerkaijun: makes sense, refactoring it in the final PR for all triedb functions. also the TODOs and FIXMEs are already addressed in the next PR.

eerkaijun: refactored in commit [254c8e3](https://github.com/monad-crypto/monad-bft/pull/1216/commits/254c8e38579c09281572bab7da5b48ca96415e54)

PR #1203, opened 2024-10-24
Transactions traversal in triedb

eerkaijun:
Traverse the transaction triedb to get all the transactions in a particular block

### DIFF
--- a/monad-rpc/src/triedb.rs
+++ b/monad-rpc/src/triedb.rs
@@ -11,9 +11,13 @@ use futures::channel::oneshot;
 use monad_triedb::TriedbHandle;
 use monad_triedb_utils::{
     decode::{rlp_decode_account, rlp_decode_storage_slot},
-    key::{create_addr_key, create_code_key, create_receipt_key, create_storage_at_key},
+    key::{
+        create_addr_key, create_code_key, create_receipt_key, create_storage_at_key,
+        create_transaction_key,
+    },
 };
-use tracing::error;
+use reth_primitives::TransactionSigned;
+use tracing::{error, warn};
 
 const MAX_CONCURRENT_TRIEDB_REQUESTS: usize = 10_000;
 
@@ -35,12 +39,23 @@ impl Default for BlockTags {
 
 enum TriedbRequest {
     BlockNumberRequest(BlockNumberRequest),
+    TransactionsRequest(TransactionsRequest),
     AsyncRequest(AsyncRequest),
 }
 
 struct BlockNumberRequest {
     // a sender to send the block number back to the request handler
     request_sender: oneshot::Sender<u64>,
+}
+
+struct TransactionsRequest {
+    // a sender for the polling thread to send the result back to the request handler
+    request_sender: oneshot::Sender<Option<Vec<Vec<u8>>>>,
+    // triedb_key and key_len_nibbles are used to read items from triedb
+    triedb_key: Vec<u8>,
+    key_len_nibbles: u8,
+    // block number
+    block_tag: BlockTags,
 }
 
 // struct that is sent from the request handler to the polling thread
@@ -67,6 +82,7 @@ pub enum TriedbResult {
     Code(Vec<u8>),
     Receipt(Vec<u8>),
     BlockNum(u64),
+    BlockTransactions(Vec<TransactionSigned>),
 }
 
 fn polling_thread(triedb_path: PathBuf, receiver: mpsc::Receiver<TriedbRequest>) {
@@ -84,6 +100,21 @@ fn polling_thread(triedb_path: PathBuf, receiver: mpsc::Receiver<TriedbRequest>)
                     TriedbRequest::BlockNumberRequest(block_num_request) => {
                         let block_num = triedb_handle.latest_block();
                         let _ = block_num_request.request_sender.send(block_num);
+                    }
+                    TriedbRequest::TransactionsRequest(transaction_request) => {
+                        // Parse block tag
+                        let block_num = match transaction_request.block_tag {
+                            BlockTags::Number(q) => q,
+                            BlockTags::Latest => triedb_handle.latest_block(),
+                        };
+                        let rlp_encoded_transactions = triedb_handle.get_transactions(
+                            &transaction_request.triedb_key,
+                            transaction_request.key_len_nibbles,
+                            block_num,
+                        );
+                        let _ = transaction_request
+                            .request_sender
+                            .send(rlp_encoded_transactions);
                     }
                     TriedbRequest::AsyncRequest(async_request) => {
                         // Process the request directly in this thread
@@ -139,6 +170,10 @@ pub trait Triedb {
         code_hash: EthCodeHash,
         block_tag: BlockTags,
     ) -> impl std::future::Future<Output = TriedbResult> + Send;
+    fn get_transactions(
+        &self,
+        block_num: u64,
+    ) -> impl std::future::Future<Output = TriedbResult> + Send;
 }
 
 pub trait TriedbPath {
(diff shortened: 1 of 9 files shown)


monad-triedb/src/lib.rs
```diff
@@ -188,6 +229,47 @@ impl TriedbHandle {
...
     }
 
+    pub fn get_transactions(
+        &self,
+        key: &[u8],
+        key_len_nibbles: u8,
+        block_id: u64,
+    ) -> Option<Vec<Vec<u8>>> {
+        assert!(key_len_nibbles < u8::MAX - 1); // make sure doesn't overflow
+        assert!((key_len_nibbles as usize + 1) / 2 <= key.len());
```
philipglazman: Instead of `assert` we might want to log error and return none

eerkaijun: great catch, updated

monad-triedb/src/lib.rs
```diff
@@ -58,6 +66,39 @@ pub unsafe extern "C" fn read_async_callback(
...
+/// # Safety
+/// This is used as a callback when traversing the transaction trie
+pub unsafe extern "C" fn transaction_callback(
+    context: *mut std::ffi::c_void,
+    key_ptr: *const u8,
+    key_len: usize,
+    value_ptr: *const u8,
+    value_len: usize,
+) {
+    let transactions = unsafe { &*(context as *mut Transactions) };
```
bamrith: The context passed to `bindings::triedb_traverse_transactions` is `Arc<Transactions>`. This should be `Arc::from_raw`. But since this is dropped at the end, make sure to `increment_strong_counter` before conversion.

monad-rpc/src/triedb.rs
```diff
@@ -393,4 +428,51 @@ impl Triedb for TriedbEnv {
...
+    async fn get_transactions(&self, block_num: u64) -> TriedbResult {
+        // create a one shot channel to retrieve the triedb result from the polling thread
+        let (request_sender, request_receiver) = oneshot::channel();
+
+        // txn_index set to None to indiciate return all transactions
+        let (triedb_key, key_len_nibbles) = create_transaction_key(None);
+
+        if let Err(e) = self
+            .mpsc_sender
+            .clone()
```
bamrith: is clone required ?

monad-rpc/src/triedb.rs
```diff
@@ -85,6 +101,21 @@ fn polling_thread(triedb_path: PathBuf, receiver: mpsc::Receiver<TriedbRequest>)
                         let block_num = triedb_handle.latest_block();
                         let _ = block_num_request.request_sender.send(block_num);
                     }
+                    TriedbRequest::TransactionsRequest(transaction_request) => {
```
bamrith, line 104: Does doing this in the polling thread affect the performance of async requests ?

monad-rpc/src/triedb.rs
eerkaijun: removed

monad-rpc/src/triedb.rs
eerkaijun: transactions traversal should be fast, there's a open issue to switch that to use async API as well: https://github.com/monad-crypto/monad-internal/issues/708 but for now, I think this should be fine rather than maintaining a separate thread pool just for transaction traversal

monad-triedb/src/lib.rs
eerkaijun: updated

monad-triedb/src/lib.rs
```diff
@@ -58,6 +66,42 @@ pub unsafe extern "C" fn read_async_callback(
...
+    };
+
+    if let Ok(mut data) = transactions.data.lock() {
+        data.insert(tx_index, value);
+    } else {
+        warn!("Failed to acquire lock");
+    };
+
+    // prevent Arc<Transactions> from dropping
+    let _ = Arc::into_raw(transactions);
```
bamrith: nit: If the ref count in Arc is never used, you can just use `Box<Transactions>` instead.

monad-triedb/triedb-driver/src/triedb.cpp
```diff
@@ -150,6 +153,78 @@ void triedb_async_read(
     state->initiate();
 }
 
+void triedb_traverse_transactions(
```
Chen-Yifan: this should just be a generic `triedb_traverse()` api, where `key` and `key_len_nibbles` are the prefix we set to start the traversal, `callback` prototype is also sufficient to be a generic callback. 
So can simply rename it to `triedb_traverse()` and all relevant variables and comments below. 
Once that's done, we can call it for traversing receipts or storages under an account and so on.

monad-triedb/src/lib.rs
eerkaijun: updated

monad-triedb/triedb-driver/src/triedb.cpp
eerkaijun: updated to be generic

monad-triedb/triedb-driver/src/triedb.cpp
```diff
@@ -150,6 +153,88 @@ void triedb_async_read(
     state->initiate();
 }
 
+void triedb_traverse(
+    triedb *db, bytes key, uint8_t key_len_nibbles, uint64_t block_id,
+    void *context, CallBack callback)
+{
+    auto nibble = monad::mpt::NibblesView{0, key_len_nibbles, key};
```
Chen-Yifan: rename to `prefix`

monad-triedb/triedb-driver/include/triedb.h
```diff
@@ -28,6 +28,13 @@ int triedb_read_data(
 void triedb_async_read(
     triedb *, bytes key, uint8_t key_len_nibbles, uint64_t block_id,
     void (*completed)(bytes value, int length, void *user), void *user);
+// traverse the transaction trie
+typedef void (*CallBack)(
```
Chen-Yifan: follow snake case, maybe rename to `callback_func`. Please also update the comment.

monad-triedb/triedb-driver/include/triedb.h
```diff
@@ -28,6 +28,13 @@ int triedb_read_data(
...
+    void *context, bytes path, size_t path_len_nibbles, bytes node_value,
+    size_t node_value_size);
```
Chen-Yifan:
```suggestion
    void *context, bytes path, size_t path_len, bytes value,
    size_t value_len);
```
Suggested some renaming on the parameters. Currently there's a mismatch here and your `transaction_callback()` implementation. There you use `key_len` as number of bytes. With this change you should also change the caller side in your traverse machine.

monad-triedb/triedb-driver/src/triedb.cpp
```diff
@@ -150,6 +153,88 @@ void triedb_async_read(
...
+                return true;
+            }
+            path_ = monad::mpt::concat(
+                monad::mpt::NibblesView{path_},
+                branch,
+                node.path_nibble_view());
+
+            if (node.has_value()) { // node is a leaf
+                size_t path_size = path_.nibble_size();
+                unsigned char *path_bytes = new uint8_t[path_size];
```
Chen-Yifan:
```suggestion
                assert((path_.nibble_size() & 1) == 0); // assert even nibble size
                size_t path_bytes = path_.nibble_size() / 2;
                unsigned char *path_data = new uint8_t[path_bytes];
```
Suggested changes accordant to the callback renaming above

eerkaijun: in this case, how would you set byte data in `path_data`? `path_.get( ... )` returns a nibble right?

Chen-Yifan: nothing below requires a change except for renaming `path_bytes` -> `path_data`, `path_size` -> `path_bytes`

eerkaijun: ok

monad-triedb-utils/src/key.rs
```diff
@@ -93,10 +118,55 @@ pub fn create_receipt_key(txn_index: u64) -> (Vec<u8>, u8) {
...
+    let key: Vec<_> = key_nibbles
+        .chunks(2)
+        .map(|chunk| (chunk[0] << 4) | chunk[1])
+        .collect();
+
+    (key, num_nibbles)
+}
+
+// if no txn_index is provided, return the key to traverse the entire table
+pub fn create_transaction_key(txn_index: Option<u64>) -> (Vec<u8>, u8) {
```
Chen-Yifan, line 142: This file can be refactored to be less redundant code, transaction and receipt keys are essentially the same except for the prefix nibble, you can change it to take a separate parameter. Same for other functions, it will be more redundant once you add more stuff. For example, tx hash and code keys are also the same except for the prefix nibble. We need a better solution than just add a separate function for each table. but won't hold up this PR for that. Can do it in a separate PR.

PR #1207, opened 2024-10-25
Add staked peer discovery in router

rgarc:
This PR adds a breaking change to the wire format of messages between
nodes. Previously, nodes exchanged messages of the form

    TCP: Signature || impl Message
    UDP: Signature || RaptorCoded(impl Message)

After this PR, messages will be of the form

    TCP: Signature || impl Message | DiscoveryMessage
    UDP: Signature || RaptorCoded(impl Message | DiscoveryMessage)

The key design design decision that motivated this change was that no
new events should have to be created and passed through the consensus
state machine for peer discovery. However, the router de/serializes
messages with the following bound

    M: Message<_> + Deserializable<Bytes>,
    OM: Serializable<Bytes> + Into<M> + Clone,

and in order for a type to implement the `Message` trait, it has to be
able to emit an associated event, which we wanted to avoid altogether.
To fix this, we introduce two new types.

```rust
    enum InboundRouterMessage<M, PT: PubKey> {
        Application(M),
        Discovery(DiscoveryMessage<PT>),
    }

    enum OutboundRouterMessage<OM, PT: PubKey> {
        Application(&OM),
        DiscoveryMessage(DiscoveryMessage<PT>),
    }
```

along with the following proto definition

```proto
    message ProtoRouterMessage {
      oneof message {
        // keep compatible with ProtoMonadMessage
        bytes app_message = 7;
        ProtoDiscoveryMessage discovery_message = 8;
      }
    }
```

The `bytes app_message = 7` field will be de/serialized to
`MonadMessage` in the same way as the old router.

### DIFF
--- a/monad-raptorcast/src/lib.rs
+++ b/monad-raptorcast/src/lib.rs
@@ -14,7 +14,10 @@ use monad_crypto::certificate_signature::{
     CertificateKeyPair, CertificateSignaturePubKey, CertificateSignatureRecoverable,
 };
 use monad_dataplane::event_loop::{BroadcastMsg, Dataplane, UnicastMsg};
-use monad_discovery::{message::InboundRouterMessage, Discovery};
+use monad_discovery::{
+    message::{InboundRouterMessage, OutboundRouterMessage},
+    Discovery,
+};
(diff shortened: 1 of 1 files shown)


rgarc:
separating the server changes first for backwards compatibility

PR #1208, opened 2024-10-25
Rotate WAL

bamrith:


monad-wal/src/wal.rs
```diff
@@ -25,24 +29,35 @@ impl<M> WALoggerConfig<M>
...
+        new_file_path.push(self.start_index.to_string());
+
+        let curr_file_handle =
+            AppendOnlyFile::read_only(new_file_path.into()).expect("failed to read file");
         let mut logger = WALogger {
             _marker: PhantomData,
-            file_handle: file,
+            file_path: self.file_path,
+
+            curr_file_index: 0,
```
omegablitz, line 48: should this be start_index?

omegablitz: nvm, curr_file_index only used on the write side

monad-wal/src/wal.rs
```diff
@@ -119,22 +118,161 @@ where
...
-        self.file_handle.write_all(&buf)?;
-        self.file_handle.write_all(&msg_buf)?;
+        let msg_len = (EVENT_HEADER_LEN + msg_buf.len()) as u64;
+
+        // check file length before appending message
+        let next_offset = self.curr_file_offset + msg_len;
+        if next_offset > MAX_FILE_SIZE as u64 {
+            // open new file
+            // set length of current file and sync
+            self.curr_file_handle.set_len(self.curr_file_offset)?;
```
omegablitz, line 125: is set_len here necessary?

monad-wal/examples/wal2json.rs
```diff
@@ -14,6 +14,9 @@ use serde::Serialize;
 struct Args {
     #[arg(long)]
     wal_path: PathBuf,
+
+    #[arg(long)]
+    wal_index: u64,
```
omegablitz: remove, will make usage more annoying

monad-wal/src/wal.rs
bamrith: just in case we write to an existing WAL file that's larger

PR #1209, opened 2024-10-25
Change router message format for peer discovery

rgarc:
This PR makes a backwards-compatible change to the format of messages
between nodes. Previously, nodes exchanged messages of the form

    TCP: Signature || impl Message
    UDP: Signature || RaptorCoded(impl Message)

After this PR, messages will be of the form

    TCP: Signature || impl Message | DiscoveryMessage
    UDP: Signature || RaptorCoded(impl Message | DiscoveryMessage)

The key design design decision that motivated this change was that no
new events should have to be created and passed through the consensus
state machine for peer discovery. However, the router de/serializes
messages with the following bound

    M: Message<_> + Deserializable<Bytes>,
    OM: Serializable<Bytes> + Into<M> + Clone,

and in order for a type to implement the `Message` trait, it has to be
able to emit an associated event, which we wanted to avoid altogether.
To fix this, we introduce two new types.

```rust
    enum InboundRouterMessage<M> {
        Application(M),
        Discovery(DiscoveryMessage),
    }

    enum OutboundRouterMessage<OM> {
        Application(&OM),
        Discovery(DiscoveryMessage),
    }
```

When receiving a message, the server will deserialize as an
`InboundRouterMessage`. When sending a message, the client will
serialize it as an `OutboundRouterMessage`. We also introduce a new
proto message that will wrap the old `ProtoMonadMessage`

```proto
    message ProtoRouterMessage {
      oneof message {
        // keep compatible with ProtoMonadMessage
        bytes app_message = 7;
        ProtoDiscoveryMessage discovery_message = 8;
      }
    }
```

The `bytes app_message = 7` field will be de/serialized to
`MonadMessage` in the same way as the old router. For
backwards-compatible deployment to current nodes, we deploy the changes
to the server side first. The server will be able to transparently
handle the old message format. Once that is complete, we can deploy the
client changes.

### DIFF
--- a/monad-discovery/src/lib.rs
+++ b/monad-discovery/src/lib.rs
@@ -1,0 +1,154 @@
+pub mod message;
+
+mod nop_discovery;
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
+
+use bytes::Bytes;
+use monad_crypto::certificate_signature::PubKey;
+use monad_proto::{
+    error::ProtoError,
+    proto::discovery::{
+        proto_ip_address::Address, ProtoIPv4, ProtoIPv6, ProtoIpAddress, ProtoMonadNameRecord,
+        ProtoNetworkEndpoint,
+    },
+};
+use monad_types::NodeId;
+pub use nop_discovery::NopDiscovery;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+pub struct NetworkEndpoint {
+    pub socket_addr: SocketAddr,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct BootstrapPeer<PT: PubKey> {
+    pub node_id: NodeId<PT>,
+    pub endpoint: NetworkEndpoint,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+pub struct MonadNameRecord<PT: PubKey> {
+    pub endpoint: NetworkEndpoint,
+    #[serde(bound = "PT: PubKey")]
+    pub node_id: NodeId<PT>,
+    pub seq_num: u64,
+}
+
+const IPV4_ADDRESS_LEN: usize = 4;
+const IPV6_ADDRESS_LEN: usize = 16;
+
+impl TryFrom<ProtoNetworkEndpoint> for NetworkEndpoint {
+    type Error = ProtoError;
+
+    fn try_from(value: ProtoNetworkEndpoint) -> Result<Self, Self::Error> {
+        let proto_ip = value
+            .ip
+            .ok_or(ProtoError::MissingRequiredField(
+                "ProtoNetworkEndpoint.ip".to_owned(),
+            ))?
+            .address
+            .ok_or(ProtoError::MissingRequiredField(
+                "ProtoIpAddress.address".to_owned(),
+            ))?;
+
+        let ip = match proto_ip {
+            Address::Ipv4(ProtoIPv4 { address }) => {
+                if address.len() != IPV4_ADDRESS_LEN {
+                    return Err(ProtoError::DeserializeError(format!(
+                        "invalid IPV4 address length: {}",
+                        address.len()
+                    )));
+                }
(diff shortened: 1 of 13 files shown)


monad-node/src/config/network.rs
```diff
@@ -2,9 +2,12 @@ use std::net::Ipv4Addr;
 
 use serde::Deserialize;
 
+use crate::config::bootstrap::LocalNameRecord;
+
 #[derive(Debug, Clone, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub struct NodeNetworkConfig {
+    pub local_name_record: LocalNameRecord,
```
rgarc: This is a breaking change for validator operators. They need to add a new section in their `node.toml` that specifies their node's local name record. If desired, this change can be removed from this PR and merged in with the client changes.

rgarc: Loading the local name record was moved to #1210

rgarc:
Tested on stressnet by having all nodes on https://github.com/monad-crypto/monad-bft/commit/adeca409164510b8d91f4578605b7a359e7a4abb and then incrementally upgrading nodes to https://github.com/monad-crypto/monad-bft/commit/11c58f32e0c8e585219e3c5b5a54314d33f32b2c making sure there were no deserialization errors in the logs and that BPS remained high.

monad-discovery/src/message.rs
```diff
@@ -0,0 +1,205 @@
...
+impl<PT: PubKey> Serializable<Bytes> for DiscoveryMessage<PT> {
+    fn serialize(&self) -> Bytes {
+        let msg: ProtoDiscoveryMessage = self.into();
+
+        let mut buf = BytesMut::new();
+        msg.encode(&mut buf)
+            .expect("message serialization shouldn't fail");
+        buf.into()
+    }
+}
```
michael-yxchen: I think we want to implement Serializable on OutboundRouterMessage. That implementation will call app_message.serialize() and discovery_message.into(). 

May not be relevant for this PR cuz it's server side things only

monad-discovery/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+            ))?
+            .address
+            .ok_or(ProtoError::MissingRequiredField(
+                "ProtoIpAddress.address".to_owned(),
+            ))?;
+
+        let ip = match proto_ip {
+            Address::Ipv4(ipv4) => IpAddr::V4(Ipv4Addr::from(ipv4.address)),
+            Address::Ipv6(ProtoIPv6 { address }) => {
+                if address.len() != 16 {
```
michael-yxchen: nit: defined constant for ipv6 address length

monad-discovery/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+        })
+    }
+}
+impl From<&NetworkEndpoint> for ProtoNetworkEndpoint {
+    fn from(value: &NetworkEndpoint) -> Self {
+        match value.socket_addr {
+            SocketAddr::V4(ipv4) => Self {
+                ip: Some(ProtoIpAddress {
+                    address: Some(Address::Ipv4(ProtoIPv4 {
+                        address: ipv4.ip().to_bits(),
```
michael-yxchen: `to_bits()` uses platform native byte order. `octets()` is portable

monad-discovery/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+        })
+    }
+}
+impl<PT: PubKey> From<&MonadNameRecord<PT>> for ProtoMonadNameRecord {
+    fn from(value: &MonadNameRecord<PT>) -> Self {
+        let MonadNameRecord {
+            endpoint,
+            node_id,
+            seq_num,
+        } = value.clone();
```
michael-yxchen: we're taking references to endpoint and node_id immediately after. Is a clone() needed here?

monad-discovery/src/lib.rs
rgarc: switched both IPv4/IPv6 to use bytes proto type

monad-discovery/src/message.rs
rgarc: right. this is dead code. removed

PR #1210, opened 2024-10-25
Service peer discovery requests

rgarc:
builds on #1209 

# Summary

This PR implements the servicing and initiating of peer discovery
requests. Change summary follows

- Load `LocalNameRecord` from `node.toml` by adding a
  `[network.local_name_record]` section. After this PR, validators will
  have to specify their domain name like so

    [network.local_name_record]
    address = "devnet3.devcore4.com:8000"
    seq_num = 2

- Implements a simple peer discovery request/response protocol making
  sure that name records are signed and signatures are checked

- Update RaptorCast client to serialize outbound messages as an
  `OutboundRouterMessage` which was added https://github.com/monad-crypto/monad-bft/pull/1209

- Add metric for number of peers learned through discovery to allow
  showing in grafana

The following is repeated from this [Staked Peer Discovery](https://www.notion.so/monad-labs/staked-peer-discovery-v2-1201413dee6e80e6a30be7cb7e2217ae) notion page.

# Bootstrap process example
In this example, B, C, and D are already up and know each other’s IPs and A would like to join. When initializing MonadState, we emit a BootstrapPeers command which kickstarts the discovery process.

```mermaid
sequenceDiagram
autonumber
    participant A as A
    %%ip: IpAddrseq_num: 0
    box transparent Already up and know eachother's IP addresses
    participant B as B
    %%note over B: discovery_peers(C, 0)(D, 0)
    participant C as C
    %%note over C: discovery_peers(C, 0)(D, 0)
    participant D as D
    end

    
    note over A: bootstrap_peers = [B]
    note over A: Command::BootstrapPeers
   
    
    
    A ->>B: DiscoveryRequest [ A  ]
    B ->B: 
    note over B,B: ADD A to peer list
    B ->>A: DiscoveryResponse [ B, C, D ]
    A -> A: 
    note over A,A: ADD [ B, C, D ] to peer list
    par asynchronously
    A -)B: DiscoveryRequest [ A ]
    note over B: ADD A to peer list<br/>(duplicate)
    A -)C: DiscoveryRequest [ A ]
    note over C: ADD A to peer list
    A -)D: DiscoveryRequest [ A ]
    note over D: ADD A to peer list
    end
    %%par asynchronously
    B ->>A: DiscoveryResponse [ B, C, D ]
    note over A: IGNORE [ B, C, D ]<br/>IP/seq_num already known
    C ->>A: DiscoveryResponse [ B, C, D ]
    note over A: IGNORE [ B, C, D ]<br/>IP/seq_num already known
    D ->>A: DiscoveryResponse [ B, C, D ]
    %%end
    note over A: IGNORE [ B, C, D ]<br/>IP/seq_num already known
    %%

```

1. For every peer in the bootstrap list, A sends a `DiscoveryRequest`  by tagging along with it its own `local_name_record`.
2. B adds A to its peer list 
3. B sends A its peer list `[B, C, D]` 
4. A adds  `[B, C, D]`  to its peer list
5. (asynchronously) sends B a `DiscoveryRequest` .  B treats the duplicate peer the same way as a new peer because it’s idempotent and handles the edge case outlined in step 8.
6. (asynchronously) sends C a `DiscoveryRequest`
7. (asynchronously) sends D a `DiscoveryRequest`
8. B sends A its peer list `[B, C, D]` . Although this is technically redundant, this handles the edge case where B has already received A’s IP/seq_num. For example, A has already done discovery with B, but has since crashed and restarted.
9. C sends A its peer list `[B, C, D]`
10. D sends A its peer list `[B, C, D]`

### Notes on this example

- Nodes always respond with their full peer list if they receive a `DiscoveryRequest` from a node they’ve never seen before, or the `DiscoveryRequest` is a duplicate (the IP/NodeId/seq_num already exist in the mapping)
- They ignore the request if it’s invalid. Either the seq_num is ≤ the current seq_num or the seq_num is the same as the latest known seq_num but the the IP doesn’t match.
- Nodes receiving `DiscoveryResponse` completely ignore duplicates to allow the protocol to terminate.

# Known Issues

- There’s no guarantee that  requests/responses arrive so there needs to be some bounded retry mechanism or a heartbeat that issues a BootstrapPeers command at a periodic, set interval
# Testing
The scenario in the diagram above was tested on stressnet.

### DIFF
--- a/monad-discovery/src/lib.rs
+++ b/monad-discovery/src/lib.rs
@@ -1,20 +1,13 @@
+pub mod convert;
 pub mod message;
 
 mod nop_discovery;
 
-use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
+use std::net::{IpAddr, SocketAddr};
 
-use bytes::Bytes;
 use monad_crypto::{
     certificate_signature::{CertificateSignaturePubKey, CertificateSignatureRecoverable, PubKey},
-    hasher::{Hashable, Hasher as _},
-};
-use monad_proto::{
-    error::ProtoError,
-    proto::discovery::{
-        proto_ip_address::Address, ProtoIPv4, ProtoIPv6, ProtoIpAddress, ProtoMonadNameRecord,
-        ProtoNetworkEndpoint, ProtoSignedMonadNameRecord,
-    },
+    hasher::{Hashable, Hasher},
 };
 use monad_types::NodeId;
 pub use nop_discovery::NopDiscovery;
@@ -27,8 +20,11 @@ pub struct NetworkEndpoint {
 }
 
 impl Hashable for NetworkEndpoint {
-    fn hash(&self, state: &mut impl monad_crypto::hasher::Hasher) {
-        state.update(self.socket_addr.ip().to_string().as_bytes());
+    fn hash(&self, state: &mut impl Hasher) {
+        match self.socket_addr.ip() {
+            IpAddr::V4(ip) => state.update(ip.octets()),
+            IpAddr::V6(ip) => state.update(ip.octets()),
+        }
         state.update(self.socket_addr.port().as_bytes());
     }
 }
@@ -48,7 +44,7 @@ pub struct MonadNameRecord<PT: PubKey> {
 }
 
 impl<PT: PubKey> Hashable for MonadNameRecord<PT> {
-    fn hash(&self, state: &mut impl monad_crypto::hasher::Hasher) {
+    fn hash(&self, state: &mut impl Hasher) {
         self.endpoint.hash(state);
         self.node_id.hash(state);
         state.update(self.seq_num.as_bytes());
@@ -61,7 +57,7 @@ pub enum DiscoveryError<ST: CertificateSignatureRecoverable> {
         existing_peer: MonadNameRecord<CertificateSignaturePubKey<ST>>,
         requesting_peer: MonadNameRecord<CertificateSignaturePubKey<ST>>,
     },
-    DuplicatePeer(SignedMonadNameRecord<ST>),
+    DuplicatePeer(VerifiedMonadNameRecord<ST>),
     SenderRecovery,
     IncorrectSender,
 }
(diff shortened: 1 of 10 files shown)


monad-raptorcast/src/lib.rs
```diff
@@ -117,6 +152,76 @@ where
...
+        discovery_message: DiscoveryMessage<CertificateSignaturePubKey<ST>>,
+    ) -> Vec<
+        Result<
+            DiscoveryNewPeer<CertificateSignaturePubKey<ST>>,
+            DiscoveryIgnored<CertificateSignaturePubKey<ST>>,
+        >,
+    > {
+        match discovery_message {
+            DiscoveryMessage::DiscoveryRequest(DiscoveryRequest { sender }) => {
+                // TODO(rene): any other message authenticity need to happen here?
```
rgarc: We previously recovered the sender from the signature in https://github.com/monad-crypto/monad-bft/blob/03ea243ee52213fb456381e8a369f9c189b26f76/monad-raptorcast/src/lib.rs#L520-L526 so is there anything else that needs to be done to confirm the authenticity of a peer discovery message?

michael-yxchen:
> If node A crashes, it must bump the sequence number in its local_name_record or else the bootstrap peer will reject its DiscoveryRequest

I feel it's inconvenient having to bump name record sequence number every time you restart. Can the server still service DiscoveryRequest if Request{A}==server.record[A]? They should only need to bump local record sequence number when address changes

rgarc:
> > If node A crashes, it must bump the sequence number in its local_name_record or else the bootstrap peer will reject its DiscoveryRequest
> 
> I feel it's inconvenient having to bump name record sequence number every time you restart. Can the server still service DiscoveryRequest if Request{A}==server.record[A]? They should only need to bump local record sequence number when address changes

I went ahead and implemented this and force pushed. I'm trying to think of how to unit-test this since it's tightly coupled to the router.

monad-raptorcast/src/lib.rs
```diff
@@ -117,6 +145,105 @@ where
...
+    > {
+        match self.discovery_peers.get(&peer.node_id) {
+            None => Ok(peer.clone()),
+            Some(existing_peer) => {
+                if peer.seq_num > existing_peer.seq_num {
+                    Ok(peer.clone())
+                } else if peer == existing_peer {
+                    Err(DiscoveryError::DuplicatePeer(peer.clone()))
+                } else {
+                    Err(DiscoveryError::InvalidPeer {
```
andr-dev, line 175: Let's rename this variant to be more specific so it doesn't get reused for other errors in the future, how about `(Inconsistent/Conflicting)NodeId` or `NodeId(Mismatch/Discrepancy)`?

monad-proto/proto/discovery.proto
```diff
@@ -29,3 +29,8 @@ message ProtoMonadNameRecord {
   monad_proto.basic.ProtoNodeId node_id = 2;
   uint64 seq_num = 3;
 }
+
+message ProtoSignedMonadNameRecord {
+  ProtoMonadNameRecord name_record = 1;
+  bytes signature = 2;
```
michael-yxchen: Can reuse monad_proto.signing.ProtoSignature here

monad-node/src/main.rs
```diff
@@ -155,6 +157,33 @@ async fn run(
...
+                .bootstrap
+                .peers
+                .iter()
+                .map(|peer| BootstrapPeer {
+                    node_id: build_node_id(peer),
+                    endpoint: NetworkEndpoint {
+                        socket_addr: resolve_domain(&peer.address),
+                    },
+                })
+                .collect::<Vec<_>>(),
```
michael-yxchen: this feels repetitive as we're passing `bootstrap.peers` to raptorcast as well. Is there a reason to keep them separate?

monad-executor-glue/src/lib.rs
```diff
@@ -124,7 +125,10 @@ pub enum GetValidatorSet<SCT: SignatureCollection> {
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub enum GetMetrics {
     Request,
-    Response(Metrics),
+    Response {
+        state_metrics: Metrics,
+        executor_metrics: BTreeMap<String, u64>,
+    },
 }
```
michael-yxchen: Did we intend to leave these protobuf conversions for ControlPanelEvent empty? https://github.com/monad-crypto/monad-bft/blob/56e55b8bae6ae357f6b803601383febec8e9bbff/monad-executor-glue/src/convert/event.rs#L396-L410

I think for read only events, it's fine to throw in a placeholder type there. We should serialize write events

monad-mock-swarm/src/mock.rs
```diff
@@ -325,7 +325,8 @@ impl<S: SwarmRelation> Executor for MockExecutor<S> {
                 RouterCommand::GetFullNodes => {
                     // TODO
                 }
-                RouterCommand::UpdateFullNodes(vec) => {
+                RouterCommand::UpdateFullNodes(vec) => {}
```
michael-yxchen: nit: follow same format

monad-discovery/src/message.rs
michael-yxchen, line 1: We should organize all protobuf conversions under `monad-discovery/convert/`

monad-discovery/src/lib.rs
```diff
@@ -1,26 +1,38 @@
...
 
 mod nop_discovery;
+
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
 
 use bytes::Bytes;
-use monad_crypto::certificate_signature::PubKey;
+use monad_crypto::{
+    certificate_signature::{CertificateSignaturePubKey, CertificateSignatureRecoverable, PubKey},
+    hasher::{Hashable, Hasher as _},
```
michael-yxchen: Remove `Hasher as _`

monad-discovery/src/lib.rs
```diff
@@ -1,26 +1,38 @@
...
+use zerocopy::AsBytes;
 
 #[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
 pub struct NetworkEndpoint {
     pub socket_addr: SocketAddr,
 }
 
+impl Hashable for NetworkEndpoint {
+    fn hash(&self, state: &mut impl monad_crypto::hasher::Hasher) {
+        state.update(self.socket_addr.ip().to_string().as_bytes());
```
michael-yxchen: I'm biased toward using `octects()` on ipv4/ipv6 variants

monad-discovery/src/lib.rs
```diff
@@ -35,6 +47,75 @@ pub struct MonadNameRecord<PT: PubKey> {
...
+
+    pub fn record(&self) -> &MonadNameRecord<CertificateSignaturePubKey<ST>> {
+        &self.monad_name_record
+    }
+
+    pub fn signature(&self) -> ST {
+        self.signature
+    }
+
+    pub fn verify(self) -> Result<SignedMonadNameRecord<ST>, DiscoveryError<ST>> {
```
michael-yxchen: I find the interface a little cyclic - `verify()` consumes `self` and returns a `Result<Self>`. Outputting a `VerifiedMonadNameRecord` is more intuitive I think

monad-discovery/src/lib.rs
```diff
@@ -146,9 +227,84 @@ impl<PT: PubKey> From<&MonadNameRecord<PT>> for ProtoMonadNameRecord {
...
+            monad_name_record: value
+                .name_record
+                .ok_or(ProtoError::MissingRequiredField(
+                    "ProtoSignedMonadNameRecord.name_record".to_owned(),
+                ))?
+                .try_into()?,
+            signature: ST::deserialize(value.signature.as_bytes()).map_err(|_| {
+                // TODO(rene): use more descriptive error
+                ProtoError::DeserializeError("failed to deserialize signature".to_string())
+            })?,
```
michael-yxchen: Note: use `proto_to_certificate_signature/certificate_signature_to_proto` after switching to `ProtoSignature`

monad-discovery/src/lib.rs
```diff
@@ -146,9 +227,84 @@ impl<PT: PubKey> From<&MonadNameRecord<PT>> for ProtoMonadNameRecord {
...
+
+        let signed_name_record =
+            SignedMonadNameRecord::<SignatureType>::new(name_record.clone(), &k1);
+        assert!(signed_name_record.verify().is_ok());
+
+        let hash = monad_crypto::hasher::HasherType::hash_object(&name_record);
+        let signature = SecpSignature::sign(hash.as_ref(), &k2);
+
+        let malicious_signed_record = SignedMonadNameRecord::with_signature(name_record, signature);
+        assert!(malicious_signed_record.verify().is_err());
```
michael-yxchen: nit: match on err type

monad-raptorcast/examples/service.rs
```diff
@@ -109,13 +110,25 @@ fn service(
...
                     known_addresses,
                     redundancy: 2,
                     local_addr: server_address.to_string(),
                     up_bandwidth_mbps: 1_000,
+                    local_name_record: MonadNameRecord {
+                        endpoint: NetworkEndpoint {
+                            socket_addr: SocketAddr::V4(SocketAddrV4::new(
+                                Ipv4Addr::new(0, 0, 0, 0),
+                                0,
+                            )),
```
michael-yxchen: should we populate with `server_address`?

monad-state/src/lib.rs
```diff
@@ -1113,7 +1114,10 @@ where
                 }
                 ControlPanelEvent::GetMetricsEvent => {
                     vec![Command::ControlPanelCommand(ControlPanelCommand::Read(
-                        ReadCommand::GetMetrics(GetMetrics::Response(self.metrics)),
+                        ReadCommand::GetMetrics(GetMetrics::Response {
+                            state_metrics: self.metrics,
+                            executor_metrics: Default::default(),
+                        }),
```
michael-yxchen: It feels a little weird that a partial response and a full response share the same type. Maybe we can introduce another variant with just state metrics and have ParentExecutor build a real response?

monad-raptorcast/tests/malformed_inputs.rs
```diff
@@ -282,13 +283,25 @@ pub fn set_up_test(
...
                 known_addresses,
                 redundancy: 2,
                 local_addr: rx_addr,
                 up_bandwidth_mbps: 1_000,
+                local_name_record: MonadNameRecord {
+                    endpoint: NetworkEndpoint {
+                        socket_addr: SocketAddr::V4(SocketAddrV4::new(
+                            Ipv4Addr::new(0, 0, 0, 0),
+                            0,
+                        )),
```
michael-yxchen: tx_addr?

monad-raptorcast/src/lib.rs
```diff
@@ -138,6 +156,107 @@ where
...
+        let verified_record = incoming_signed_record.clone().verify()?;
+        let peer = verified_record.record();
+        match self.discovery_peers.get(&peer.node_id) {
+            None => Ok(verified_record),
+            Some(signed_record) => {
+                let existing_peer = signed_record.record();
+                if peer.seq_num > existing_peer.seq_num {
+                    Ok(verified_record)
+                } else if peer == existing_peer {
+                    Err(DiscoveryError::DuplicatePeer(verified_record))
```
michael-yxchen: Is there a use case where we treat DuplicatePeer differently from a valid peer? Basically I think it's cleaner to return `Ok(verified_record)` if we're treating them the same way - the error handling code below merges the two code paths

rgarc: They can't be treated the same way. For a bootstrap node, you always want to service responses, so you treat DuplicatePeer as a valid peer. For the client that wants to bootstrap, you want to ignore duplicate peers. If we return `Ok(verified_record)` here, the protocol wouldn't terminate.

michael-yxchen: thanks that makes sense. i misread the branches in handle_discovery

monad-discovery/src/lib.rs
rgarc: I agree. Adding a new type `VerifiedMonadRecord` with a private constructor, and will only allow creating one by  using the `impl TryFrom<SignedMonadNameRecord> for VerifiedMonadNameRecord`

monad-state/src/lib.rs
rgarc: Did you mean something like this?
https://github.com/monad-crypto/monad-bft/pull/1210/commits/ba80f102cff29458ba14056bf470fbe0ed6e0fc9

monad-executor-glue/src/lib.rs
rgarc: I'm not sure what you mean here. All these events are serialized.

michael-yxchen: Oh I think I got https://github.com/monad-crypto/monad-bft/blob/7744f0acaee2a63d2878173722aa6f584ba87c2d/monad-executor-glue/src/lib.rs#L556 and https://github.com/monad-crypto/monad-bft/blob/7744f0acaee2a63d2878173722aa6f584ba87c2d/monad-executor-glue/src/lib.rs#L125-L128 mixed together. Sorry...

PR #1212, opened 2024-10-25
Avoid duplicate payload requests

bamrith:


monad-blocksync/src/blocksync.rs
```diff
@@ -442,35 +449,50 @@ where
...
-                        BlockSyncRequestMessage::Headers(block_range),
-                    ));
+                    // filter blocks that are already in the blocktree
+                    if let Some(blocktree) = self.maybe_blocktree {
+                        // don't request payloads for committed blocks or blocks that are already
+                        // in the blocktree.
+                        block_headers = block_headers
+                            .into_iter()
+                            .filter_map(|block| {
+                                if block.round <= blocktree.root().round
```
omegablitz: move blocktree lookups to try_initate_payload_requests so this is unified for peer payload requests

omegablitz:
Should improve valnet sync performance dramatically

monad-blocksync/src/blocksync.rs
```diff
@@ -367,9 +367,39 @@ where
...
+                // remove request and clone payload to existing headers
+                self.block_sync.self_payload_requests.remove(&payload_id);
+
+                for (_, (_, payload_requests)) in
+                    self.block_sync.self_completed_headers_requests.iter_mut()
+                {
+                    for (block, maybe_payload) in payload_requests {
+                        if block.payload_id == payload_id && maybe_payload.is_none() {
+                            // clone incase there are multiple requests that require the same payload
+                            *maybe_payload = Some(payload.clone());
```
omegablitz, line 419: Can self_payload_requests_in_flight go out of sync here?

omegablitz: If so, can just remove all of the in_flight counters/metrics - it's not worth the code complication. Can compute it on the fly when checking against BLOCKSYNC_MAX_PAYLOAD_REQUESTS.

omegablitz: We also may want to bump `self_payload_response_successful` here.

bamrith: This is done even before starting a ledger fetch. `self_payload_requests_in_flight` is incremented only when we initiate a ledger fetch. So, I don't think this should affect it.

Bumping the `self_payload_response_successful`

PR #1214, opened 2024-10-26
Replace reserve balance validation with max txn fee validation

mikety:
This PR changes the txn validation. The code path that was validating reserve balance is changed to compute the low bound of an account balance. The account balance is computed as the "account balance K blocks ago"  - (max gas * max_gas_fee + tx value) for all transactions in recent K blocks.
the max cost of transactions in insert, create and validate proposal is validated  to be not greater than the computed balance.

### DIFF
--- a/monad-eth-block-policy/src/lib.rs
+++ b/monad-eth-block-policy/src/lib.rs
@@ -13,6 +13,7 @@ use monad_eth_tx::{EthSignedTransaction, EthTransaction, EthTxHash};
 use monad_eth_types::{Balance, EthAddress, Nonce};
 use monad_state_backend::{StateBackend, StateBackendError};
 use monad_types::{BlockId, Epoch, NodeId, Round, SeqNum, GENESIS_SEQ_NUM};
+use reth_primitives::U256;
 use sorted_vector_map::SortedVectorMap;
 use tracing::trace;
 
@@ -24,6 +25,27 @@ pub enum ReserveBalanceCheck {
     Insert,
     Propose,
     Validate,
+}
+
+pub fn checked_sum(arg1: U256, arg2: U256) -> U256 {
+    let max_u128 = U256::from(u128::MAX);
+    match arg1.checked_add(arg2) {
+        Some(val) => {
+            if val > max_u128 {
+                max_u128
+            } else {
+                val
+            }
+        }
+        None => max_u128,
+    }
+}
+
+fn checked_from(arg: U256) -> u128 {
+    match u128::try_from(arg) {
+        Ok(val) => val,
+        Err(_err) => u128::MAX,
+    }
 }
 
 fn compute_intrinsic_gas(tx: &EthSignedTransaction) -> u64 {
@@ -60,10 +82,22 @@ fn compute_intrinsic_gas(tx: &EthSignedTransaction) -> u64 {
     intrinsic_gas
 }
 
-pub fn compute_txn_carriage_cost(txn: &EthSignedTransaction) -> u128 {
-    let max_fee_per_gas = txn.max_fee_per_gas();
-    let intrinsic_gas = compute_intrinsic_gas(txn);
-    (intrinsic_gas as u128) * max_fee_per_gas
+#[allow(clippy::unnecessary_fallible_conversions)]
+
+pub fn compute_txn_max_value(txn: &EthSignedTransaction) -> U256 {
+    let txn_value = U256::try_from(txn.value()).unwrap();
+    let gas_cost = U256::from(txn.gas_limit() as u128 * txn.max_fee_per_gas());
+
+    checked_sum(txn_value, gas_cost)
+}
+
+pub fn compute_txn_max_value_to_u128(txn: &EthSignedTransaction) -> u128 {
+    let txn_value = compute_txn_max_value(txn);
+    trace!(
+        "compute_txn_max_value gas_cost + txn_value: {:?} ",
+        txn_value
+    );
+    checked_from(txn_value)
 }
 
 // allow for more fine grain debugging if needed
@@ -118,7 +152,7 @@ pub struct EthValidatedBlock<SCT: SignatureCollection> {
     pub orig_payload: Payload,
     pub validated_txns: Vec<EthTransaction>,
     pub nonces: BTreeMap<EthAddress, Nonce>,
-    pub carriage_costs: BTreeMap<EthAddress, Balance>,
+    pub txn_fees: BTreeMap<EthAddress, U256>,
 }
 
 impl<SCT: SignatureCollection> EthValidatedBlock<SCT> {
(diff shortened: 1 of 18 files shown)


monad-eth-block-policy/src/lib.rs
```diff
@@ -60,11 +62,26 @@ fn compute_intrinsic_gas(tx: &EthSignedTransaction) -> u64 {
...
+
+pub fn compute_txn_max_value(txn: &EthSignedTransaction) -> u128 {
+    let txn_value = U256::try_from(txn.value()).unwrap();
+    let mut gas: u128 = 0;
+
+    if txn.kind().is_create() {
+        gas += txn.gas_limit() as u128;
+    } else {
+        gas += compute_intrinsic_gas(txn) as u128;
+    }
```
eerkaijun: I think this is incorrect. Even if the transaction is not a smart contract creation transaction, the gas calculation should still use the gas limit specified by the user. (e.g smart contract call transaction has gas cost larger than intrinsic gas)

This would probably just be `gas = txn.gas_limit()`

mikety: Updated.

monad-eth-block-policy/src/lib.rs
```diff
@@ -62,23 +62,9 @@ fn compute_intrinsic_gas(tx: &EthSignedTransaction) -> u64 {
...
 pub fn compute_txn_max_value(txn: &EthSignedTransaction) -> u128 {
     let txn_value = U256::try_from(txn.value()).unwrap();
-    let mut gas: u128 = 0;
-
-    if txn.kind().is_create() {
-        gas += txn.gas_limit() as u128;
-    } else {
-        gas += compute_intrinsic_gas(txn) as u128;
-    }
+    let  gas = txn.gas_limit() as u128;
```
mikety: Its possible to narrow the scope by separating transactions to create and run a contract and simple transfers.
The pro is that simple transfers use exact intrinsic gas and therefore users will not be overallocating.

monad-eth-block-policy/src/lib.rs
```diff
@@ -60,10 +62,11 @@ fn compute_intrinsic_gas(tx: &EthSignedTransaction) -> u64 {
     intrinsic_gas
 }
 
-pub fn compute_txn_carriage_cost(txn: &EthSignedTransaction) -> u128 {
-    let max_fee_per_gas = txn.max_fee_per_gas();
-    let intrinsic_gas = compute_intrinsic_gas(txn);
-    (intrinsic_gas as u128) * max_fee_per_gas
+pub fn compute_txn_max_value(txn: &EthSignedTransaction) -> u128 {
+    let txn_value = U256::try_from(txn.value()).unwrap();
```
eerkaijun, line 88: can we remove the unwrap here? shouldn't be able to crash the node if transaction value is malformed. 

maybe `unwrap_or_default`

mikety: Good point. Will update. It brings up a question where is the right place to validate if the transaction or its values are malformed. I think it should be done in consensus before static_check.

monad-eth-block-policy/src/lib.rs
```diff
@@ -60,10 +61,17 @@ fn compute_intrinsic_gas(tx: &EthSignedTransaction) -> u64 {
...
+#[allow(clippy::unnecessary_fallible_conversions)]
+pub fn compute_txn_max_value(txn: &EthSignedTransaction) -> u128 {
+    let mut txn_value = U256::try_from(txn.value()).unwrap();
+    let gas_cost = U256::from(txn.gas_limit() as u128 * txn.max_fee_per_gas());
+
+    txn_value = txn_value.add(gas_cost);
+    match u128::try_from(txn_value) {
+        Ok(val) => val,
+        // In the unlikely scenario that txn value is larger than u128 use MAX as a default.
+        Err(_err) => u128::MAX,
```
mikety: added [internal issue ](https://github.com/monad-crypto/monad-internal/issues/755 )to track the fix.

monad-eth-block-policy/src/lib.rs
```diff
@@ -313,25 +321,25 @@ impl CommittedBlkBuffer {
...
         let mut next_validate = base_seq_num + SeqNum(1);
 
         // start iteration from base_seq_num (non inclusive)
-        for (&cache_seq_num, (_nonce, block_carriage_costs)) in self.blocks.range(next_validate..) {
+        for (&cache_seq_num, (_nonce, block_txn_fees)) in self.blocks.range(next_validate..) {
             assert_eq!(next_validate, cache_seq_num);
-            if let Some(account_carriage_cost) = block_carriage_costs.get(eth_address) {
-                carriage_cost += account_carriage_cost;
+            if let Some(account_txn_fee) = block_txn_fees.get(eth_address) {
+                txn_fee += account_txn_fee;
```
mikety: TODO: [use checked for overflow](https://github.com/monad-crypto/monad-internal/issues/755 ) addition.

monad-eth-block-policy/src/lib.rs
mikety: U256::try_from(txn.value()) cant fail - it returns INFALLIBLE.

eerkaijun: for clarity, can it be `U256::from(txn.value())`?

monad-eth-block-policy/src/lib.rs
```diff
@@ -492,49 +517,47 @@ impl EthBlockPolicy {
...
+                // Apply Txn Fees for the txns from committed blocks
+                let CommittedTxnFeeResult {
+                    txn_fee: txn_fee_committed,
                     next_validate,
-                } = self
-                    .committed_cache
-                    .compute_carriage_cost(base_seq_num, address);
+                } = self.committed_cache.compute_txn_fee(base_seq_num, address);
+
+                let txn_fee_committed_u128 = checked_from(txn_fee_committed);
```
eerkaijun, line 533: what's the reasoning for needing to change it back to u128? can we do all calculations and comparisons in U256?

monad-eth-block-policy/src/lib.rs
mikety: it will give a type mismatch error:
```
error[E0271]: type mismatch resolving `<Uint<256, 4> as TryFrom<TxValue>>::Error == ToUintError<Uint<256, 4>>`
   --> monad-eth-block-policy/src/lib.rs:87:32
    |
87  |     let txn_value = U256::from(txn.value());
    |                     ---------- ^^^^^^^^^^^ expected `ToUintError<Uint<256, 4>>`, found `Infallible`
    |                     |
    |                     required by a bound introduced by this call
    |
    = note: expected enum `ToUintError<Uint<256, 4>>`
               found enum `Infallible`
```

monad-eth-block-policy/src/lib.rs
mikety: In all places txn_fees are converted to `u128` to  be comparable with the account balance. The conversion ensures that it will not exceed `u128::MAX` value. The other way around is possible - i.e. checked compare, checked subtraction but it will be more code.

mikety:
I think we should consider making account balance to be U256. However I see it as a separate work, to not obscure changes in the account balance validation.

PR #1215, opened 2024-10-27
Get transaction location in block using transaction hash

eerkaijun:
Recover endpoints that query by transaction hash: https://github.com/monad-crypto/monad-internal/issues/721 

Dependent on https://github.com/monad-crypto/monad/pull/973

### DIFF
--- a/monad-rpc/src/eth_txn_handlers.rs
+++ b/monad-rpc/src/eth_txn_handlers.rs
@@ -449,27 +449,27 @@ pub struct MonadEthGetTransactionReceiptParams {
     tx_hash: EthHash,
 }
 
-#[rpc(method = "eth_getTransactionReceipt", ignore = "file_ledger_reader")]
+#[rpc(method = "eth_getTransactionReceipt")]
 #[allow(non_snake_case)]
 /// Returns the receipt of a transaction by transaction hash.
 pub async fn monad_eth_getTransactionReceipt<T: Triedb>(
-    file_ledger_reader: &FileBlockReader,
     triedb_env: &T,
-    p: MonadEthGetTransactionReceiptParams,
+    params: MonadEthGetTransactionReceiptParams,
 ) -> JsonRpcResult<Option<MonadTransactionReceipt>> {
-    trace!("monad_eth_getTransactionReceipt: {p:?}");
+    trace!("monad_eth_getTransactionReceipt: {params:?}");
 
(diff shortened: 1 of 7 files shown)


monad-rpc/src/eth_txn_handlers.rs
```diff
@@ -540,35 +543,61 @@ pub struct MonadEthGetTransactionByHashParams {
...
+            return Err(JsonRpcError::internal_error(
+                "error reading tx hash from db".into(),
+            ))
+        }
     };
 
-    let Some(block) = file_ledger_reader.get_block_by_hash(txn_value.block_hash) else {
-        return Ok(None);
+    let block_hash: FixedBytes<32>;
+    let header = match triedb_env.get_block_header(block_num).await {
```
andr-dev, line 567: Stylistic choice, might be easier to read `let (block_hash, header) = match ... { ... };`

monad-rpc/src/triedb.rs
```diff
@@ -85,6 +86,8 @@ pub enum TriedbResult {
     Transaction(TransactionSigned),
     BlockTransactions(Vec<TransactionSigned>),
     BlockHeader(Vec<u8>),
+    // (block_num, tx_index)
+    TransactionLocation(u64, u64),
```
andr-dev, line 90: I think here is a good time to be more explicit about what the type represents.
```rust
TransactionLocation {
    block_num: u64,
    tx_index: u64,
}
```

monad-rpc/src/triedb.rs
```diff
@@ -580,4 +587,54 @@ impl Triedb for TriedbEnv {
...
+    async fn get_transaction_location_by_hash(&self, tx_hash: EthTxHash) -> TriedbResult {
+        // create a one shot channel to retrieve the triedb result from the polling thread
+        let (request_sender, request_receiver) = oneshot::channel();
+
+        let (triedb_key, key_len_nibbles) = create_transaction_hash_key(&tx_hash);
+        let completed_counter = Arc::new(AtomicUsize::new(0));
+
+        if let Err(e) = self
+            .mpsc_sender
+            .clone()
```
andr-dev, line 595: I don't think we need to `clone()` the `mpsc_sender` to call `try_send`. Looking back it seems like the other funcs in this file use this `.clone().try_send(...)` pattern. Let's try to remove it :)

PR #1216, opened 2024-10-28
Query block number by block hash

eerkaijun:
Should be the last PR to recover the RPC endpoints: https://github.com/monad-crypto/monad-internal/issues/721

This is dependent on execution to add support for querying { blockHash => blockNum }, which is still in progress currently. 

Also cleaned up various unused block utils functions. This PR is breaking as the RPC cli arguments are changed. With also update monad-infra accordingly

### DIFF
--- a/monad-rpc/src/block_handlers.rs
+++ b/monad-rpc/src/block_handlers.rs
@@ -2,18 +2,17 @@ use alloy_primitives::{
     aliases::{U256, U64},
     FixedBytes,
 };
-use alloy_rlp::Decodable;
 use monad_rpc_docs::rpc;
-use reth_primitives::{keccak256, Header as RlpHeader, ReceiptWithBloom, TransactionSigned};
+use reth_primitives::{Header as RlpHeader, ReceiptWithBloom, TransactionSigned};
 use reth_rpc_types::{Block, BlockTransactions, Header, Transaction, TransactionReceipt};
 use serde::{Deserialize, Serialize};
-use tracing::{debug, error, trace};
+use tracing::trace;
 
 use crate::{
     eth_json_types::{BlockTags, EthHash, MonadBlock, MonadTransactionReceipt, Quantity},
     eth_txn_handlers::{parse_tx_content, parse_tx_receipt},
     jsonrpc::{JsonRpcError, JsonRpcResult},
-    triedb::{get_block_num_from_tag, Triedb, TriedbResult},
+    triedb::{get_block_num_from_tag, Triedb},
 };
 
 fn parse_block_content(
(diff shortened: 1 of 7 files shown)


monad-rpc/src/eth_txn_handlers.rs
```diff
@@ -304,13 +304,16 @@ pub async fn monad_eth_getLogs<T: Triedb>(
...
-                let block = file_ledger_reader
-                    .clone()
-                    .get_block_by_hash(block_hash.into())
-                    .ok_or(JsonRpcError::internal_error(
-                        "error reading block data".into(),
-                    ))?;
-                (block.number, block.number)
+                let block_num = match triedb_env.get_block_number_by_hash(block_hash.0).await {
+                    TriedbResult::BlockNum(block_num) => block_num,
+                    TriedbResult::Null => return Ok(MonadEthGetLogsResult(vec![])),
```
eerkaijun: @philipglazman just wanted to double check with you - if the block hash is not found, `eth_getLogs` should return an empty array right?

philipglazman: That is correct. If block hash is not found, we should return an empty array

monad-rpc/src/block_handlers.rs
```diff
@@ -124,24 +122,66 @@ pub struct MonadEthGetBlock {
...
-    Ok(retval.map(|block| MonadEthGetBlock {
+    let transactions = match triedb_env.get_transactions(block_num).await {
+        TriedbResult::BlockTransactions(transactions) => transactions,
+        TriedbResult::Null => vec![],
+        _ => {
+            return Err(JsonRpcError::internal_error(
+                "error reading transactions from db".into(),
+            ))
+        }
+    };
```
ariqchowdhury: similar comment from another review, maybe this can be put in a helper that returns Result<Vec<Txn>, Err>
Maybe the root problem is the triedb-env being used directly here. It would be nice if there was a higher layer for db access that uses the generic Triedb so that the rpc logic here doesn't have to deal directly with TriedbResult

monad-rpc/src/block_handlers.rs
```diff
@@ -266,13 +318,13 @@ pub async fn block_receipts<T: Triedb>(
 ) -> Result<Vec<TransactionReceipt>, JsonRpcError> {
     let block_num: u64 = block_header.number;
     let transaction_count = transactions.len();
-    let mut block_receipts: Vec<(ReceiptDetails, usize)> = vec![];
```
tongzhi1998, line 271: Let's change this function to use the universal traverse function if possible. Instead of looping for receipts

monad-rpc/src/block_handlers.rs
eerkaijun: refactored in commit [254c8e3](https://github.com/monad-crypto/monad-bft/pull/1216/commits/254c8e38579c09281572bab7da5b48ca96415e54)

monad-rpc/src/block_handlers.rs
eerkaijun: addressing this separately https://github.com/monad-crypto/monad-internal/issues/738

PR #1217, opened 2024-10-28
Transaction Generator rewrite with erc20 and native transfer support

JoeHowarth:
Also:
- metrics: sent tps, committed tps, rpc requests per second, total non-zero accounts tracked
- many configurable load generation types
- no more priv_keys file, accounts loaded from seeds
- stable - no crashes or nonce death spirals


monad-eth-testutil/examples/txgen/shared/erc20.rs
```diff
@@ -0,0 +1,169 @@
...
...90610833565b90915550506040518181526000906001600160a01b0384169060008051602061087083398151915290602001610627565b602081526000825180602084015260005b818110156106d257602081860181015160408684010152016106b5565b506000604082850101526040601f19601f83011684010191505092915050565b80356001600160a01b038116811461070957600080fd5b919050565b6000806040838503121561072157600080fd5b61072a836106f2565b946020939093013593505050565b60008060006060848603121561074d57600080fd5b610756846106f2565b9250610764602085016106f2565b929592945050506040919091013590565b60006020828403121561078757600080fd5b610790826106f2565b9392505050565b600080604083850312156107aa57600080fd5b6107b3836106f2565b91506107c1602084016106f2565b90509250929050565b6000602082840312156107dc57600080fd5b5035919050565b600181811c908216806107f757607f821691505b60208210810361081757634e487b7160e01b600052602260045260246000fd5b50919050565b634e487b7160e01b600052601160045260246000fd5b818103818111156103a8576103a861081d565b808201808211156103a8576103a861081d565b634e487b7160e01b600052603260045260246000fdfeddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3efa2646970667358221220e2f72b4db55ca5299e0cb70a489b5138ab28fa3dbf29dcaa0d4d7a37a044979264736f6c634300081a0033";
```
andr-dev: We probably want to use the `include_str` macro here and put this bytecode in a separate file.

monad-eth-testutil/examples/txgen/shared/json_rpc.rs
```diff
@@ -0,0 +1,138 @@
...
+
+        batch.send().await?;
+
+        let mut output = Vec::with_capacity(num_addrs);
+        for fut in futs {
+            output.push(fut.await.map(|(a, n)| (a, n.to())).map_err(Into::into));
+        }
+
+        let elapsed_ms = now.elapsed().as_millis();
+        trace!(elapsed_ms, "batch_get_transaction_count");
```
andr-dev: We have a struct (`monad_types::DropTimer`) that does this for you :)
```rust
let _timer = DropTimer::start(Duration::ZERO, |elapsed| {
    trace!(?elapsed, "batch_get_transaction_count");
});
```

monad-eth-testutil/Cargo.toml
```diff
@@ -21,16 +21,23 @@ reth-primitives = { workspace = true }
...
 reth-primitives = { workspace = true }
 reth-rpc-types = { workspace = true }
-ruint = "1.11.1"
-serde = { workspace = true }
+serde.workspace = true
 serde_json = { workspace = true }
 thiserror = { workspace = true }
 tokio = { workspace = true, features = ["macros", "rt", "rt-multi-thread"] }
+tracing = { workspace = true, features = ["log", "std"] }
+tracing-subscriber = { workspace = true, features = ["env-filter", "std"] }
```
andr-dev: Why are we using specific versions of these dependencies instead of the ones in the workspace toml?

docker/txgen/Dockerfile
```diff
@@ -0,0 +1,88 @@
...
+  libboost-json1.83-dev \
+  libboost-stacktrace1.83-dev \
+  libboost1.83-dev
+
+# install cargo
+ARG CARGO_ROOT="/root/.cargo"
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | bash -s -- -y
+ENV PATH="${CARGO_ROOT}/bin:$PATH"
+
+WORKDIR /usr/src/monad-bft
```
andr-dev, line 72: I don't think we need all these libs and packages to build and run txgen, where these copied from another Dockerfile?

monad-eth-testutil/examples/txgen/shared/private_key.rs
```diff
@@ -0,0 +1,41 @@
...
+
+        let address = Address::from_slice(&hash[12..]);
+
+        (address, Self { priv_key: pk })
+    }
+
+    pub fn sign_transaction(&self, transaction: &Transaction) -> Signature {
+        sign_message(self.priv_key, transaction.signature_hash()).expect("signature works")
+    }
+}
```
andr-dev: This looks like a copy of `monad-eth-testutil/examples/erc20-sender/account.rs`.

monad-eth-testutil/examples/erc20-sender/main.rs
```diff
@@ -0,0 +1,346 @@
...
+        //     .collect::<std::result::Result<Vec<_>, _>>()?;
+        // let data = logs["data"].as_str().unwrap();
+        // let data = Bytes::from_hex(data)?;
+        // let log = ERC20::ERC20Events::decode_raw_log(&topics, &data, false)?;
+        // match log {
+        //     ERC20::ERC20Events::Transfer(transfer) => {
+        //         println!("Transfer Event: {}", &transfer.value)
+        //     }
+        //     ERC20::ERC20Events::Approval(approval) => todo!(),
+        // }
```
andr-dev: Are we going to remove these comments?

monad-eth-testutil/examples/txgen/generators/mod.rs
```diff
@@ -0,0 +1,206 @@
...
+    tx
+}
+
+pub fn erc20_mint(from: &mut SimpleAccount, amt: U256, erc20: &ERC20) -> TransactionSigned {
+    let tx = erc20.construct_mint(&from.key, from.nonce);
+
+    // update from
+    from.nonce += 1;
+    from.native_bal -= U256::from(400_000 * 1_000); // todo: fixme
+    from.erc20_bal += amt; // todo: fixme
```
andr-dev: Will this be fixed by the time the PR gets merged? If not, a more descriptive todo message would help in understanding whats missing.

monad-eth-testutil/examples/txgen/main.rs
```diff
@@ -0,0 +1,187 @@
...
+        (config.tx_per_sender() * config.senders()) as f64 / config.tps as f64;
+    if time_to_send_txs_from_all_senders < config.refresh_delay_secs {
+        warn!(
+            time_to_send_txs_from_all_senders,
+            refresh_delay = config.refresh_delay_secs,
+            "Not enough senders for given tps to prevent stall during refresh"
+        );
+    }
+
+    tokio::spawn(run::run(client, config)).await?
```
andr-dev: Why are we spawning another thread? I think we can just `run::run(client, config).await` here.

monad-eth-testutil/examples/txgen/run.rs
```diff
@@ -0,0 +1,103 @@
...
+    };
+
+    // spawn all workers
+
+    tokio::spawn(rpc_sender.run());
+    tokio::spawn(recipient_tracker.run());
+    tokio::spawn(metrics.run());
+    tokio::spawn(committed_tx_watcher.run());
+    tokio::spawn(gen.run());
+    tokio::spawn(refresher.run()).await?;
```
andr-dev: We're spawning a bunch of tokio tasks but throwing away the JoinHandle. Depending on implementation and build settings, it's possible that one of these threads terminates or errors for some reason but the rest of the txgen continues running as if nothing happened. Let's code defensively and select on these handles.

Additionally, we probably want to make this function only return an error. I believe the txgen should never terminate with Ok(()) and produce a non-zero status code since the only time we want it to stop is if we ctrl-c or otherwise kill it.

monad-eth-testutil/examples/txgen/shared/json_rpc.rs
```diff
@@ -0,0 +1,138 @@
+use alloy_rpc_client::ReqwestClient;
+use eyre::Result;
+use reth_primitives::{Address, Bytes, TransactionSigned, U256, U64};
+use tracing::{trace, warn};
+
+use super::erc20::ERC20;
+
+pub trait JsonRpc {
```
andr-dev: `JsonRpc` is a generic protocol, let's call this `EthJsonRpc` or similar.

monad-eth-testutil/examples/txgen/workers/gen_harness.rs
```diff
@@ -0,0 +1,200 @@
...
+    //                 accts: group.iter().cloned().map(SimpleAccount::from).collect(),
+    //                 txs: Vec::new(),
+    //             })
+    //             .await
+    //             .unwrap();
+    //     }
+
+    //     info!("Seeding complete");
+    //     self.root = root_slice.pop().unwrap();
+    // }
```
andr-dev: comment ditto

monad-eth-testutil/examples/txgen/workers/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+        let idx = self.next_idx();
+        self.buf[idx].clone()
+    }
+}
+
+pub struct AsyncSeededKeyPool {
+    pub rx: mpsc::Receiver<(Address, PrivateKey)>,
+    pub buf: Vec<(Address, PrivateKey)>,
+    pub cursor: usize,
+}
```
andr-dev: Why do we need an Async version of a `SeededKeyPool`? This struct also doesn't seem to be used anywhere other than in a test.

monad-eth-testutil/Cargo.toml
JoeHowarth: alloy-sol-macro = "0.6"
alloy-sol-types = "0.6"
dashmap = "6.1.0"
eyre = "0.6.12"

aren't in the workspace already, I put them there for the next commit but not sure whether they should be or not

docker/txgen/Dockerfile
JoeHowarth: The build fails without them : (
see:
https://github.com/monad-crypto/monad-bft/actions/runs/11577702705

monad-eth-testutil/examples/txgen/shared/private_key.rs
JoeHowarth: removed erc20-sender, it was just a simplified example to get erc20 working with the crates / types in the mono-repo. No longer needed so nuked

monad-eth-testutil/examples/txgen/workers/mod.rs
JoeHowarth: yeah forgot to remove

monad-eth-testutil/examples/txgen/run.rs
JoeHowarth: For this application it is preferable to try to continue running unless it absolutely cannot. I added a 'helper' vs. 'critical' task wrapper and we now log when a task terminates. If it's a critical task it aborts the whole application.

We prefer to carry on if possible because these are long-running jobs where the chain is being stressed to breaking (and often will break) but we expect the gen to continue functioning whenever possible. Basically I don't want to scrap an 8hr run if the committed tx watcher fails

docker/txgen/Dockerfile
andr-dev: Yep you'll probably need `cmake`, I was more talking about all the libraries and packages being installed that are probably used in monad-crypto/monad but probably aren't needed for txgen.
```
iproute2 pkg-config gnupg software-properties-common wget cgroup-tools
libboost-*
libabsl-dev
libarchive-dev
libbenchmark-dev
libbrotli-dev
etc.
```
I'm also fairly certain you can just `apt-get install cmake` as we just need _some_ cmake available.

monad-eth-testutil/Cargo.toml
andr-dev: Resolved offline.

docker/txgen/Dockerfile
JoeHowarth: it's not quite that easy - spent about an hour getting it to work before resulting to this. It's still the fastest docker build job so it doesn't really matter since the builder layer isn't included in the final layer anyway

PR #1222, opened 2024-10-30
Improve blocktree payload lookup perf

omegablitz:
Improves the performance of BlockTree::get_payload to O(1).


omegablitz:
This is to improve the performance of #1212

PR #1228, opened 2024-10-31
Archive setup

tongzhi1998:
Monad Archive Setup
To run an archive instance:

`CXX=/usr/bin/g++-13 CC=/usr/bin/gcc-13 ASMFLAGS=-march=haswell CFLAGS="-march=haswell" CXXFLAGS="-march=haswell" TRIEDB_TARGET=triedb_driver cargo run --package monad-archive -- --s3-bucket <> --triedb-path <>`

Benchmark Stats:
Archiving 100 blocks from 12M: 8s
Archiving 500 blocks from 12M: 43s

Testing branch: `https://github.com/monad-crypto/monad-bft/tree/tzhi/archive_interface_testing`

### DIFF
--- a/monad-archive/src/archive_interface.rs
+++ b/monad-archive/src/archive_interface.rs
@@ -1,4 +1,3 @@
-use bytes::Bytes;
 use reth_primitives::{Block, ReceiptWithBloom, TransactionSigned};
 
 use crate::{errors::ArchiveError, triedb::BlockHeader};
@@ -38,7 +37,7 @@ pub trait ArchiveWriterInterface {
     // TODO
     async fn archive_traces(
         &self,
-        traces: Bytes,
+        traces: Vec<Vec<u8>>,
         block_num: u64,
         tx_hashes: Vec<[u8; 32]>,
     ) -> Result<(), ArchiveError>;
--- a/monad-archive/src/main.rs
+++ b/monad-archive/src/main.rs
@@ -28,23 +28,23 @@ async fn main() -> Result<(), ArchiveError> {
 
     let args = Cli::parse();
 
+    let concurrency = args.max_concurrency_level;
+
     // This will spin off a polling thread
     let triedb = TriedbEnv::new(&args.triedb_path.unwrap());
 
     // Construct an s3 instance
-    let s3_archive = S3Archive::new(args.s3_bucket, args.s3_region).await?;
+    let s3_archive = S3Archive::new(args.s3_bucket, args.s3_region, concurrency).await?;
     let s3_archive_writer = S3ArchiveWriter::new(s3_archive).await?;
 
     let mut latest_processed_block = (s3_archive_writer.get_latest().await).unwrap_or_default();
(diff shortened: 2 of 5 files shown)


monad-archive/src/archive_interface.rs
```diff
@@ -0,0 +1,101 @@
...
+
+    // eth_getBlockByNumber
+    // eth_getRawBlock
+    // eth_getRawHeader
+    async fn get_block_by_number(&self, block_num: u64) -> Result<Block, ArchiveError>;
+
+    //eth_getBlockTransactionCountByHash
+    async fn get_block_transaction_count_by_hash(
+        &self,
+        block_hash: &[u8],
```
eerkaijun: would recommend setting this as &[u8; 32] for stricter typings

monad-archive/src/archive_interface.rs
```diff
@@ -0,0 +1,101 @@
...
+    // eth_getTransactionByBlockNumberAndIndex
+    async fn get_transaction_by_block_number_and_index(
+        &self,
+        block_number: u64,
+        tx_index: u64,
+    ) -> Result<Transaction, ArchiveError>;
+
+    // eth_getTransactionByHash
+    // eth_getRawTransaction
+    async fn get_transaction_by_hash(&self, tx_hash: &[u8]) -> Result<Transaction, ArchiveError>;
```
eerkaijun: same as above

eerkaijun:
are there any rough performance intuition on uploading data and reading data from archive? e.g. if the load of the chain is 1k TPS, can we also read 1k data from the archive?

tongzhi1998:
> reading



> are there any rough performance intuition on uploading data and reading data from archive? e.g. if the load of the chain is 1k TPS, can we also read 1k data from the archive?

I haven't load-tested read performance yet. Write performance for ethereum block is in the description above. I'll need to play around with tx generator later to test performance for bigger blocks.

monad-archive/src/main.rs
```diff
@@ -0,0 +1,114 @@
...
+    let s3_archive_writer = S3ArchiveWriter::new(s3_archive).await?;
+
+    let mut latest_processed_block = (s3_archive_writer.get_latest().await).unwrap_or_default();
+
+    info!("Latest processed block is : {}", latest_processed_block);
+
+    let concurrency = args.max_concurrency_level;
+
+    // Check for new blocks every 100 ms
+    // Issue requests to triedb, poll data and push to relevant tables
```
philipglazman, line 45: I added a stream type called blockwatcher that can be useful here. Maybe it needs its own crate? https://github.com/monad-crypto/monad-bft/blob/master/monad-rpc/src/block_watcher.rs

tongzhi1998: Yeah, I would just replace the triedb part with this I guess? I was copying triedb.rs into the monad-archive crate for similar reason. I think we should discuss about how to consolidate the code.

PR #1231, opened 2024-11-01
Implement periodic pings to calculate network latency

maxkozlovsky:
Periodically ping other validators and calculate average latency over a number of most recent pings.

The resulting latency will be used in calculating of timestamp adjustments.

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -146,7 +146,7 @@ where
     /// Policy for validating incoming proposals
     pub block_validator: &'a BVT,
     /// Track local timestamp and validate proposal timestamps
-    pub block_timestamp: &'a BlockTimestamp,
+    pub block_timestamp: &'a mut BlockTimestamp<SCT::NodeIdPubKey>,
 
     /// Destination address for proposal payments
     pub beneficiary: &'a EthAddress,
@@ -444,12 +444,15 @@ where
             }
         };
 
-        if let Some(ts_delta) = self
-            .block_timestamp
-            .valid_block_timestamp(block.get_qc().get_timestamp(), block.get_timestamp())
-        {
+        if let Ok(Some(ts_delta)) = self.block_timestamp.valid_block_timestamp(
+            block.get_qc().get_timestamp(),
+            block.get_timestamp(),
+            block.get_round(),
+            &author,
+        ) {
             // only update timestamp if the block advanced us our round
-            if block.get_round() > original_round {
+            if block.get_qc().get_round() == original_round {
+                info!(?ts_delta, "update timestamp");
                 cmds.push(ConsensusCommand::TimestampUpdate(ts_delta));
             }
         }
@@ -714,6 +717,9 @@ where
             message: ProtocolMessage::Vote(vote_msg),
         }
         .sign(self.keypair);
+        if next_leader != *self.nodeid {
+            self.block_timestamp.vote_sent(round);
+        }
         let send_cmd = ConsensusCommand::Publish {
             target: RouterTarget::PointToPoint(next_leader),
             message: msg,
@@ -1095,8 +1101,10 @@ where
             .valid_block_timestamp(
                 validated_block.get_qc().get_timestamp(),
                 validated_block.get_timestamp(),
+                validated_block.get_round(),
+                &validated_block.get_author(),
             )
-            .is_none()
+            .is_err()
         {
             self.metrics.consensus_events.failed_ts_validation += 1;
             warn!(prev_block_ts = ?validated_block.get_qc().get_timestamp(),
@@ -1592,7 +1600,7 @@ mod test {
         validators::create_keys_w_validators,
     };
     use monad_types::{
-        BlockId, Epoch, NodeId, Round, RouterTarget, SeqNum, Stake, GENESIS_SEQ_NUM,
+        BlockId, Epoch, NodeId, PingSequence, Round, RouterTarget, SeqNum, Stake, GENESIS_SEQ_NUM,
     };
     use monad_validator::{
         epoch_manager::EpochManager,
@@ -1646,7 +1654,7 @@ mod test {
         block_validator: BVT,
         block_policy: BPT,
         state_backend: SBT,
-        block_timestamp: BlockTimestamp,
+        block_timestamp: BlockTimestamp<SCT::NodeIdPubKey>,
         beneficiary: EthAddress,
         nodeid: NodeId<CertificateSignaturePubKey<ST>>,
         consensus_config: ConsensusConfig,
@@ -2327,6 +2335,10 @@ mod test {
             MockTxPool::default(),
             || NopStateRoot,
         );
+        for ctx in &mut ctx {
+            ctx.wrapped_state().block_timestamp.update_time(1000);
+        }
+
         let mut wrapped_state = ctx[0].wrapped_state();
 
         // our initial starting logic has consensus in round 1 so the first proposal does not
@@ -2337,6 +2349,42 @@ mod test {
 
         let p2 = env.next_proposal_empty();
         let (author, _, verified_message) = p2.destructure();
+
+        let val_set = wrapped_state
+            .val_epoch_map
+            .get_val_set(
+                &wrapped_state
+                    .epoch_manager
+                    .get_epoch(verified_message.block.round)
+                    .unwrap(),
+            )
+            .unwrap();
+        let val_data = val_set
+            .get_members()
+            .iter()
+            .map(
+                |(node_id, stake)| crate::ValidatorData::<SignatureCollectionType> {
+                    node_id: *node_id,
+                    stake: *stake,
+                    cert_pubkey: node_id.pubkey(),
+                },
+            )
+            .collect::<Vec<_>>();
+
+        wrapped_state
+            .block_timestamp
+            .update_validators(&val_data, wrapped_state.nodeid);
+
+        for _ in 0..30 {
+            wrapped_state.block_timestamp.tick();
+        }
+        std::thread::sleep(Duration::from_millis(1));
+        for val in val_set.get_members().keys() {
+            wrapped_state
+                .block_timestamp
+                .pong_received(*val, PingSequence(1));
+        }
+        wrapped_state.block_timestamp.vote_sent(Round(1));
         let cmds = wrapped_state.handle_proposal_message(author, verified_message.clone());
         assert!(find_timestamp_update_cmd(&cmds).is_some());
 
@@ -2351,7 +2399,10 @@ mod test {
         let p7 = env.next_proposal_empty();
         let (author, _, verified_message) = p7.destructure();
         let cmds = wrapped_state.handle_proposal_message(author, verified_message);
-        assert!(find_timestamp_update_cmd(&cmds).is_some());
+        assert!(
+            find_timestamp_update_cmd(&cmds).is_none(),
+            "no timestamp adjustment because did not vote previous round"
+        );
     }
 
     #[test]
@@ -2890,7 +2941,7 @@ mod test {
 
         // the proposal still gets processed: the node enters a new round, and
         // issues a request for the block it skipped over
-        assert_eq!(cmds.len(), 6);
+        assert_eq!(cmds.len(), 5);
         assert!(matches!(cmds[0], ConsensusCommand::StartExecution));
         assert!(matches!(cmds[1], ConsensusCommand::CheckpointSave(_)));
         assert!(matches!(cmds[2], ConsensusCommand::EnterRound(_, _)));
@@ -2899,7 +2950,6 @@ mod test {
             ConsensusCommand::Schedule { duration: _ }
         ));
         assert!(matches!(cmds[4], ConsensusCommand::RequestSync { .. }));
-        assert!(matches!(cmds[5], ConsensusCommand::TimestampUpdate(_)));
         assert_eq!(
             wrapped_state.metrics.consensus_events.rx_execution_lagging,
             1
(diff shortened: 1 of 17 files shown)


monad-debugger/src/graphql.rs
```diff
@@ -242,6 +242,9 @@ impl<'s> From<&'s MonadEventType> for GraphQLMonadEvent<'s> {
                 Self::TimestampEvent(GraphQLTimestampEvent(*event))
             }
             MonadEvent::StateSyncEvent(event) => Self::StateSyncEvent(GraphQLStateSyncEvent(event)),
+            &MonadEvent::PingRequestEvent(_)
+            | &MonadEvent::PingResponseEvent(_)
+            | &MonadEvent::PingTickEvent => todo!(),
```
michael-yxchen: nit
```suggestion
            MonadEvent::PingRequestEvent(_)
            | MonadEvent::PingResponseEvent(_)
            | MonadEvent::PingTickEvent => todo!(),
```

monad-proto/proto/event.proto
```diff
@@ -181,6 +181,20 @@ message ProtoStateSyncEvent {
...
+}
+
+message PingResponseEvent {
+  monad_proto.basic.ProtoNodeId sender = 1;
+  uint64 sequence = 2;
+}
+
+message PingTickEvent {
+
+}
```
michael-yxchen, line 262: nit: prefix all protobuf types with `Proto`. It's then clear in conversion code which type is protobuf generated

monad-state/src/lib.rs
```diff
@@ -1136,6 +1169,39 @@ where
...
+            MonadEvent::PingTickEvent => {
+                tracing::info!("ping tick");
+                let mut cmds = self
+                    .block_timestamp
+                    .tick()
+                    .into_iter()
+                    .map(|(node, sequence)| {
+                        tracing::info!(?node, ?sequence, "sending ping request");
+                        Command::RouterCommand(RouterCommand::Publish {
+                            target: RouterTarget::TcpPointToPoint(node),
```
michael-yxchen, line 1244: I wonder if UDP is more appropriate here. TcpPointToPoint initiates a new connection with the server, adding an extra round trip tiem to both ping and pong packet

monad-executor-glue/src/lib.rs
```diff
@@ -497,7 +498,7 @@ pub enum StateSyncEvent<SCT: SignatureCollection> {
     /// Execution done syncing
     DoneSync(SeqNum),
 
-    // Statesync-requested block
+    // St 212atesync-requested block
```
michael-yxchen: nit

monad-state/src/lib.rs
```diff
@@ -896,6 +922,13 @@ where
             }
 
             MonadEvent::ValidatorEvent(validator_event) => {
+                match &validator_event {
+                    ValidatorEvent::UpdateValidators((validators, epoch)) => {
+                        self.block_timestamp
+                            .update_validators(&validators.0, &self.nodeid);
+                    }
+                }
+
```
michael-yxchen, line 933: UpdateValidators event locks validator set for a future epoch. The validator set doesn't become active until `EnterRound(epoch,..)` I think you'd want to insert validators to block_timestamp here and call `update_validators` on EnterRound(..). You can add something like `LoopBackCommand(MonadEvent::TimestampUpdateValidators)` here https://github.com/monad-crypto/monad-bft/blob/d2d2c0a1a081e4ea9bb23f71a9bbc260667e28fe/monad-state/src/consensus.rs#L337-L340

monad-state/src/lib.rs
```diff
@@ -1136,6 +1169,39 @@ where
...
+                    .map(|(node, sequence)| {
+                        tracing::info!(?node, ?sequence, "sending ping request");
+                        Command::RouterCommand(RouterCommand::Publish {
+                            target: RouterTarget::TcpPointToPoint(node),
+                            message: VerifiedMonadMessage::PingRequest(sequence),
+                        })
+                    })
+                    .collect::<Vec<_>>();
+                cmds.push(Command::TimerCommand(TimerCommand::Schedule {
+                    duration: std::time::Duration::from_secs(1),
```
michael-yxchen, line 1250: let's move this constant together with PING_PERIOD

monad-consensus-state/src/timestamp.rs
```diff
@@ -49,71 +224,255 @@ impl BlockTimestamp {
...
-                if !self.valid_bounds(curr_block_ts) {
-                    None
+            return Err(Error::Invalid);
+        }
+        if !self.valid_bounds(curr_block_ts) {
+            return Err(Error::OutOfBounds);
+        }
+        // return the delta between expected block time and actual block time for adjustment
+        match &self.sent_vote {
+            Some(vote) if vote.round + Round(1) == round => {
```
michael-yxchen, line 241: To be more accurate, I think we should only adjust timestamp if `block.qc.round == vote.round`, excluding the TC->proposal case

monad-consensus-state/src/timestamp.rs
```diff
@@ -49,71 +224,255 @@ impl BlockTimestamp {
...
+    }
+
+    pub fn pong_received(&mut self, node_id: NodeId<P>, sequence: PingSequence) {
+        self.ping_state.pong_received(node_id, sequence);
+    }
+
+    pub fn vote_sent(&mut self, round: Round) {
+        info!("vote sent for round {:?}", round);
+        match &self.sent_vote {
+            Some(vote) if vote.round >= round => {
```
michael-yxchen: I think we can assert that `round > vote.round` per pacemaker invariants

monad-consensus-state/src/timestamp.rs
```diff
@@ -49,71 +224,255 @@ impl BlockTimestamp {
...
+
+        let my_key = NopKeyPair::from_bytes(&mut [0; 32]).unwrap().pubkey();
+        let my_node = NodeId::new(my_key);
+
+        let k_1 = NopKeyPair::from_bytes(&mut [1; 32]).unwrap().pubkey();
+        let node_1 = NodeId::new(k_1);
+
+        let k_2 = NopKeyPair::from_bytes(&mut [2; 32]).unwrap().pubkey();
+        let node_2 = NodeId::new(k_2);
+
```
michael-yxchen: nit: `monad_testutil::signing::create_keys` may be helpful here

monad-state/src/lib.rs
maxkozlovsky: Yes, meant to be UDP, my mistake.

monad-proto/proto/event.proto
maxkozlovsky: The  types are scoped within proto mod, and they can be imported as 'use monad_proto::PingRequestEvent as ProtoPingRequestEvent' to disambiguate. I think we do not need extra prefix here.

monad-executor-glue/src/lib.rs
maxkozlovsky: Accidental change, removing

monad-state/src/lib.rs
maxkozlovsky: Yes makes sense.

monad-consensus-state/src/timestamp.rs
maxkozlovsky: Ok nice

monad-state/src/lib.rs
maxkozlovsky: Is it worth it to add extra event to be more precise here? We are calculating average over large period of time, following rounds exactly is not likely to make difference.

monad-consensus-state/src/timestamp.rs
maxkozlovsky: Ok will change to use the function.

monad-consensus-state/src/timestamp.rs
maxkozlovsky: Ok this makes it simpler.

monad-consensus-state/src/timestamp.rs
maxkozlovsky: updated check in the caller

monad-proto/proto/event.proto
michael-yxchen: sure let's leave it for now. Message/event serialization will change soon I think

monad-state/src/lib.rs
michael-yxchen: You're right that the precise round doesn't matter that much. Epoch/active validator set matters when some validators exit: say validator `V` is leaving in epoch `e+2`, the current implementation removes timestamp tracking of a validator as soon as `e+2` validator set is scheduled, by the end of epoch `e`. Then in epoch `e+1` (which last on the order of hours or days), it doesn't track latency measurements for `V`. This is the reason I suggested TimeStamper should be made aware of current active validator set via `EnterRound(...)`

monad-consensus-state/src/lib.rs
```diff
@@ -444,12 +444,15 @@ where
...
-            .block_timestamp
-            .valid_block_timestamp(block.get_qc().get_timestamp(), block.get_timestamp())
-        {
+        if let Ok(Some(ts_delta)) = self.block_timestamp.valid_block_timestamp(
+            block.get_qc().get_timestamp(),
+            block.get_timestamp(),
+            block.get_round(),
+            &author,
+        ) {
             // only update timestamp if the block advanced us our round
```
michael-yxchen, line 453: nit: update comment

monad-consensus-state/src/lib.rs
```diff
@@ -2337,6 +2349,42 @@ mod test {
...
+            .collect::<Vec<_>>();
+
+        wrapped_state
+            .block_timestamp
+            .update_validators(&val_data, wrapped_state.nodeid);
+
+        for _ in 0..30 {
+            wrapped_state.block_timestamp.tick();
+        }
+        std::thread::sleep(Duration::from_millis(1));
```
michael-yxchen, line 2381: is it possible to remove sleep from this unit test?

monad-state/src/lib.rs
```diff
@@ -947,7 +981,7 @@ where
                 }
                 StateSyncEvent::Outbound(to, message) => {
                     vec![Command::RouterCommand(RouterCommand::Publish {
-                        target: RouterTarget::TcpPointToPoint(to),
+                        target: RouterTarget::PointToPoint(to),
```
michael-yxchen, line 984: unrelated change?

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,201 @@
...
+    signature_collection::SignatureCollection,
+    validator_data::ValidatorData,
+};
+use monad_crypto::certificate_signature::PubKey;
+use monad_types::{NodeId, PingSequence, Round};
+use tracing::info;
+
+const MAX_LATENCY_SAMPLES: usize = 100;
+
+const PING_PERIOD: Duration = Duration::from_secs(30);
```
andr-dev, line 18: Let's change this constant to be `PING_PERIOD_SECS: u64` and then instantiate the `Duration` when we want a duration rather than calling `to_secs()` on a `Duration` which is technically imprecise.

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,201 @@
...
+    fn update_validators<SCT>(&mut self, validators: &Vec<ValidatorData<SCT>>, my_node: &NodeId<P>)
+    where
+        SCT: SignatureCollection<NodeIdPubKey = P>,
+    {
+        let removed_nodes = self
+            .validators
+            .keys()
+            .filter(|node_id| !validators.iter().any(|v| v.node_id == **node_id))
+            .cloned()
+            .collect::<Vec<_>>();
```
andr-dev, line 128: We don't need to collect this into a `Vec<_>`, let's just use the iterator.

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,201 @@
...
+    pub fn avg_latency(&self) -> Option<Duration> {
+        self.avg_latency
+    }
+}
+
 #[derive(Debug)]
-pub struct BlockTimestamp {
+struct PingState<P: PubKey> {
+    validators: HashMap<NodeId<P>, ValidatorPingState>, // ping state per validator
+    schedule: Vec<Vec<NodeId<P>>>, // schedule of validators to ping, length is period
```
andr-dev, line 92: If we change `PING_PERIOD` to `PING_PERIOD_SECS` as mentioned above, we can make this
```rust
schedule: [Vec<NodeId<P>>; PING_PERIOD_SECS]
```
That way, it's clear that the length of schedule never changes.

monad-executor-glue/src/convert/event.rs
```diff
@@ -7,15 +7,38 @@ use monad_consensus_types::{
 use monad_crypto::certificate_signature::{CertificateSignatureRecoverable, PubKey};
 use monad_proto::{
     error::ProtoError,
-    proto::{blocksync::ProtoBlockSyncSelfRequest, event::*},
+    proto::{
+        blocksync::ProtoBlockSyncSelfRequest,
+        event::{self, *},
```
andr-dev, line 12: This looks funny, if we're already importing * why import event with self? I think we should generally avoid star imports, let's just specify each type here (which should be Proto prefixed).

monad-proto/proto/event.proto
andr-dev: Unresolving this one as there isn't a good reason to change the status quo here. Just because you can change the naming when importing doesn't mean we should arbitrarily change naming conventions. Additionally, you didn't change the naming conventions to disambiguate which was confusing until I scrolled down and saw this conversation. This is also inconsistent with the `ProtoPingRequest/ProtoPingResponse` down below.

monad-state/src/lib.rs
```diff
@@ -1186,6 +1220,39 @@ where
...
+                vec![]
+            }
+            MonadEvent::PingTickEvent => {
+                tracing::info!("ping tick");
+                let mut cmds = self
+                    .block_timestamp
+                    .tick()
+                    .into_iter()
+                    .map(|(node, sequence)| {
+                        tracing::info!(?node, ?sequence, "sending ping request");
```
andr-dev, line 1242: We should probably move this into `debug` as it'll unnecessarily pollute the info log as node count increases.

PR #1233, opened 2024-11-02
Live to sync threshold

omegablitz:
Previously, the StateSync -> Live and Live -> StateSync thresholds were set to the same thing. This meant that premature Live -> StateSync transitions could happen without giving time to blocksync.


monad-consensus-state/src/lib.rs
```diff
@@ -1794,7 +1805,9 @@ mod test {
                     proposal_txn_limit: 5000,
                     proposal_gas_limit: 8_000_000,
                     delta: Duration::from_secs(1),
-                    state_sync_threshold: SeqNum(100),
+                    statesync_to_live_threshold: SeqNum(600),
+                    live_to_statesync_threshold: SeqNum(900),
+                    start_execution_threshold: SeqNum(300),
```
ariqchowdhury, line 1813: these numbers seem long? 900 blocks is 15min.

omegablitz: It is long for "normal" Live -> Sync transition, to revisit later

PR #1234, opened 2024-11-02
Pace voting with vote timer

ariqchowdhury:


monad-consensus-state/src/lib.rs
```diff
@@ -1066,34 +1127,29 @@ where
...
-                version: self.version.into(),
-                message: ProtocolMessage::Vote(vote_msg),
+            match self.consensus.scheduled_vote {
+                Some(OutgoingVoteStatus::TimerFired) => {
+                    // timer already fired for this round so send vote immediately
+                    let vote_cmd = self.create_vote_cmd(round, v);
+                    cmds.extend(vote_cmd);
+                }
+                Some(OutgoingVoteStatus::VoteReady(r))
+                    if r.vote_info.round == v.vote_info.round =>
```
michael-yxchen: nit: we should panic if `r.vote_info.round >= v.vote_info.round`

monad-consensus-state/src/lib.rs
```diff
@@ -1066,34 +1127,29 @@ where
...
-            let next_leader = self.election.get_leader(next_round, next_validator_set);
-            let msg = ConsensusMessage {
-                version: self.version.into(),
-                message: ProtocolMessage::Vote(vote_msg),
+            match self.consensus.scheduled_vote {
+                Some(OutgoingVoteStatus::TimerFired) => {
+                    // timer already fired for this round so send vote immediately
+                    let vote_cmd = self.create_vote_cmd(round, v);
+                    cmds.extend(vote_cmd);
+                }
```
michael-yxchen: Should we reset `self.consensus.scheduled_vote` to None? We'll send all votes immediately if this is in TimerFired state

monad-consensus-state/src/lib.rs
```diff
@@ -659,6 +670,56 @@ where
...
+            target: RouterTarget::PointToPoint(next_leader),
+            message: msg,
+        };
+        debug!(?round, ?vote, ?next_leader, "created vote");
+        self.metrics.consensus_events.created_vote += 1;
+
+        // start the vote-timer for the next round
+        let vote_timer_cmd = ConsensusCommand::ScheduleVote {
+            // FIXME config this
+            duration: Duration::from_millis(5),
```
michael-yxchen: Use value from ConsensusConfig

monad-consensus-state/src/lib.rs
```diff
@@ -233,6 +243,7 @@ where
         });
         ConsensusState {
```
michael-yxchen, line 244: It'd be helpful to check that `config.vote_pace < pacemaker.get_round_timer() - some_delta` to avoid misconfiguration

omegablitz: It's actually kinda annoying to assert these on the creation of ConsensusConfig (we don't use a ctor), which is why I didn't do the same for the statesync thresholds.

I guess we could assert these conditions on ConsensusState init, which accomplishes the same thing but looks uglier

michael-yxchen: I'm inclined to do this one on ConsensusState init, because it depends on pacemaker timeout value. We don't need to do it now though

monad-consensus-state/src/lib.rs
```diff
@@ -659,6 +670,56 @@ where
         cmds
     }
 
+    pub fn handle_vote_timer(&mut self, round: Round) -> Vec<ConsensusCommand<ST, SCT>> {
```
omegablitz, line 674: nit: `#[must_use]`

monad-consensus-state/src/lib.rs
```diff
@@ -659,6 +670,56 @@ where
...
+            self.consensus.scheduled_vote = Some(OutgoingVoteStatus::TimerFired);
+            return vec![];
+        };
+
+        assert_eq!(round, v.vote_info.round);
+
+        self.create_vote_cmd(round, v)
+    }
+
+    pub fn create_vote_cmd(&mut self, round: Round, vote: Vote) -> Vec<ConsensusCommand<ST, SCT>> {
```
omegablitz: nit: `#[must_use]`

My preference below:

Given that this anyway is a mutator function (increments created_vote), rename to `send_vote_and_reset_timer`, and additionally set `self.consensus.scheduled_vote = None` in this function.

Then, remove the `self.consensus.take()` call from handle_vote_timer, and the `self.consensus.scheduled_vote = None` from the `Some(OutgoingVoteStatus::VoteReady(_)) | None` handler below. This also fixes michael's todo below.

This would make it clearer to me that whenever the ConsensusCommand::ScheduleVote command is emitted, scheduled_vote is always cleared.

monad-consensus-state/src/lib.rs
```diff
@@ -1066,34 +1127,29 @@ where
...
+                    // timer already fired for this round so send vote immediately
+                    let vote_cmd = self.create_vote_cmd(round, v);
+                    cmds.extend(vote_cmd);
+                }
+                Some(OutgoingVoteStatus::VoteReady(r))
+                    if r.vote_info.round == v.vote_info.round =>
+                {
+                    panic!("trying to schedule another vote in same round");
+                }
+                Some(OutgoingVoteStatus::VoteReady(_)) | None => {
```
omegablitz, line 1147: `OutgoingVoteStatus::VoteReady(_)` should only happen here if either we're receiving proposals late, or if the majority's pacing isn't in-line with our pacing - is this correct?

Maybe we should bump a metric or log a warning in this case.

PR #1246, opened 2024-11-06
Adjust pacemaker round timer to account for vote delay and local compute

ariqchowdhury:


monad-node/src/main.rs
```diff
@@ -343,7 +343,7 @@ async fn run(
         consensus_config: ConsensusConfig {
             proposal_txn_limit: node_state.node_config.consensus.block_txn_limit,
             proposal_gas_limit: node_state.node_config.consensus.block_gas_limit,
-            delta: Duration::from_millis(node_state.node_config.network.max_rtt_ms / 2),
+            delta: Duration::from_millis(150),
```
omegablitz: Any reason to change this from `max_rtt_ms / 2`?

omegablitz:
image

monad-node/src/main.rs
ariqchowdhury: no, that didn't need to change

PR #1247, opened 2024-11-07
Bump concurrent triedb reads limit and make it configurable

eerkaijun:
Bumping up the default backpressure limit for reading from triedb. 1k TPS indexed by a blockscout instance hits the original limit quite frequently, tested and we are able to support higher limits. Making it a configurable flag too so rpc providers can adjust as needed

### DIFF
--- a/monad-rpc/src/cli.rs
+++ b/monad-rpc/src/cli.rs
@@ -48,4 +48,8 @@ pub struct Cli {
     /// Set the max concurrent requests for eth_call and eth_estimateGas
     #[arg(long, default_value_t = 1000)]
     pub eth_call_max_concurrent_requests: u32,
+
+    /// Set the max concurrent requests for triedb reads
+    #[arg(long, default_value_t = 20_000)]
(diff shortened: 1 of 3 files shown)


omegablitz:
Why does 1k TPS indexed by a blockscout hit the original limit? Did we confirm that this is because blockscout is sending very large _bursts_ of requests? The old 10,000 limit already seems fairly large.

If the underlying issue is that the blockscout requests are sent at a higher _rate_ than what the triedb reads can keep up with, then this will not fix the underlying problem, and will cause requests to get more and more laggy.

eerkaijun:
There are large bursts of requests which causes some being dropped. The triedb reads can keep up without issue at 1k TPS.

eerkaijun:
Gonna run a few more experiments to see at which rate does the triedb reads fall behind, then we can set a better default value for max concurrent requests

eerkaijun:
just fyi tested this on 5k TPS indexing load and RPC still keeping up with the triedb reads

PR #1250, opened 2024-11-09
Remove bincode from monad-debug-node

rgarc:
https://github.com/monad-crypto/monad-internal/issues/371

bincode isn't stable across builds so we switch to something schemaless to prevent against silent type conversion failures.

This commit also removes the obsolete control panel binary


monad-state/src/lib.rs
```diff
@@ -1093,21 +1093,27 @@ where
                         .unwrap()
                         .map
                         .clone();
```
andr-dev: Unrelated to this change but can we try removing this `clone()` and use `into_iter()` below?

PR #1257, opened 2024-11-13
Archive setup

tongzhi1998:
`monad-archive` uploads blocks, receipts and traces to s3 blob storage

The latest block number is polled from triedb, compared to the last block
number uploaded and the blocks in between are uploaded in parallel at
configurable concurrency level.

Multiple archiver processes are intended to be run on multiple nodes
simulataneously, each uploading to their own bucket for redundancy.

This version does not expose metrics, check data integrity nor index txhashes,
all will be added in follow up prs.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -277,3 +277,50 @@ jobs:
           file: docker/txgen/Dockerfile
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+    
+  publish-monad-archive:
+    runs-on: ubuntu-24.04-32
+
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Generate a token
+        id: generate_token
+        uses: tibdex/github-app-token@v1
+        with:
+          app_id: ${{ secrets.REPO_READONLY_GITHUB_APP_ID }}
+          private_key: ${{ secrets.REPO_READONLY_GITHUB_APP_KEY }}
+          permissions: >-
+            {"contents": "read"}
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+          token: ${{ steps.generate_token.outputs.token }}
+
+      - name: Log in to the Peach10 Container registry
+        uses: docker/login-action@v3.1.0
+        with:
+          registry: peach10.devcore4.com
+          username: ${{ secrets.PRIVATE_REGISTRY_USERNAME }}
+          password: ${{ secrets.PRIVATE_REGISTRY_PASSWORD }}
+
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@v5.5.1
+        with:
+          images: |
+            peach10.devcore4.com/monad-crypto/monad-archiver
+          tags: |
+            type=ref,event=branch
+            type=sha,prefix=
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v5.3.0
+        with:
+          context: .
+          push: true
+          file: docker/archive/Dockerfile
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,7 @@
 [workspace]
 
 members = [
+    "monad-archive",
     "monad-async-state-verify",
     "monad-block-persist",
     "monad-blocksync",
@@ -90,6 +91,8 @@ async-graphql = "7.0"
 auto_impl = "1.1.0"
 autocxx = "0.26"
 autocxx-build = "0.26"
+aws-config = "1.5.9"
+aws-sdk-s3 = "1.58.0"
 bincode = "1.3"
 bindgen = "0.69"
 bitvec = "1.0"
@@ -159,6 +162,7 @@ thiserror = "1.0"
 tiny-bip39 = "1.0"
 tiny-keccak = "2"
 tokio = "1.39"
+tokio-retry = "0.3.0"
 tokio-util = "0.7"
 toml = "0.7"
 tracing = "0.1"
--- a/monad-archive/src/cli.rs
+++ b/monad-archive/src/cli.rs
@@ -1,0 +1,28 @@
+use std::path::PathBuf;
+
+use clap::Parser;
+
+#[derive(Debug, Parser)]
+#[command(name = "monad-archive", about, long_about = None)]
+pub struct Cli {
+    #[arg(long)]
+    pub triedb_path: PathBuf,
+
+    #[arg(long)]
+    pub s3_bucket: String,
+
+    #[arg(long)]
+    pub s3_region: Option<String>,
+
+    /// Set the max response size in bytes (the same as monad-rpc)
+    #[arg(long, default_value_t = 25_000_000)]
+    pub max_response_size: u32,
+
+    /// Set the concurrent blocks
+    #[arg(long, default_value_t = 10)]
+    pub max_concurrent_blocks: usize,
+
+    /// Set max block processed per iter
+    #[arg(long, default_value_t = 100)]
+    pub max_blocks_per_iteration: u64,
+}
--- a/monad-archive/src/errors.rs
+++ b/monad-archive/src/errors.rs
@@ -1,0 +1,1 @@
+pub type ArchiveError = eyre::ErrReport;
--- a/monad-archive/src/main.rs
+++ b/monad-archive/src/main.rs
@@ -1,0 +1,162 @@
+use std::{sync::Arc, time::Instant};
+
+use clap::Parser;
+use futures::future::join_all;
+use monad_archive::{
+    archive_interface::ArchiveWriterInterface,
+    cli::Cli,
+    errors::ArchiveError,
+    s3_archive::{S3Archive, S3ArchiveWriter},
+    triedb::{Triedb, TriedbEnv},
+};
+use tokio::{
+    sync::Semaphore,
+    time::{sleep, Duration},
+    try_join,
+};
+use tracing::{debug, error, info, warn, Level};
+
+#[tokio::main]
+async fn main() -> Result<(), ArchiveError> {
+    tracing_subscriber::fmt().with_max_level(Level::INFO).init();
+
+    let args = Cli::parse();
+
+    let max_concurrent_blocks = args.max_concurrent_blocks;
+    let concurrent_block_semaphore = Arc::new(Semaphore::new(max_concurrent_blocks));
+
+    // This will spin off a polling thread
+    let triedb = TriedbEnv::new(&args.triedb_path);
+
+    // Construct an s3 instance
+    let s3_archive = S3Archive::new(args.s3_bucket, args.s3_region).await?;
+    let s3_archive_writer = S3ArchiveWriter::new(s3_archive).await?;
+
+    // Check for new blocks every 100 ms
+    // Issue requests to triedb, poll data and push to relevant tables
+    loop {
+        sleep(Duration::from_millis(100)).await;
+
+        let trie_db_block_number = match triedb.get_latest_block().await {
+            Ok(number) => number,
+            Err(e) => return Err(e),
+        };
+
+        let latest_processed_block = (s3_archive_writer.get_latest().await).unwrap_or_default();
+
+        if trie_db_block_number <= latest_processed_block {
+            debug!(
+                "Nothing to process. S3 archive progress: {}, triedb progress: {}",
+                latest_processed_block, trie_db_block_number
+            );
+            continue;
+        }
+
+        let start_block_number = if latest_processed_block == 0 {
+            0
+        } else {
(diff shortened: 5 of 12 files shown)


monad-archive/src/main.rs
```diff
@@ -0,0 +1,137 @@
...
+) -> Result<(), ArchiveError> {
+    /*  Store Blocks */
+    let block_header = match triedb.get_block_header(current_block).await? {
+        Some(header) => header,
+        None => return Ok(()),
+    };
+    let transactions = triedb.get_transactions(current_block).await?;
+
+    let f_block =
+        s3_archive.archive_block(block_header.clone(), transactions.clone(), current_block);
```
bamrith: can remove clones

monad-archive/src/main.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    /* Store Receipts */
+    let receipts: Vec<ReceiptWithBloom> = triedb.get_receipts(current_block).await?;
+    let mut tx_hashes: Vec<[u8; 32]> = vec![];
+    for transaction in transactions.clone() {
+        tx_hashes.push(transaction.hash.into());
+    }
+
+    let f_receipt = s3_archive.archive_receipts(receipts.clone(), current_block);
```
bamrith: same here

monad-archive/src/main.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    /* Store Traces */
+    let traces: Vec<Vec<u8>> = triedb.get_call_frames(current_block).await?;
+
+    let f_trace = s3_archive.archive_traces(traces.clone(), current_block);
```
bamrith: same here

monad-archive/src/main.rs
```diff
@@ -0,0 +1,137 @@
...
+    };
+    let transactions = triedb.get_transactions(current_block).await?;
+
+    let f_block =
+        s3_archive.archive_block(block_header.clone(), transactions.clone(), current_block);
+
+    /* Store Receipts */
+    let receipts: Vec<ReceiptWithBloom> = triedb.get_receipts(current_block).await?;
+    let mut tx_hashes: Vec<[u8; 32]> = vec![];
+    for transaction in transactions.clone() {
```
bamrith: can remove clone here too. can do this before `archive_block` and use `&transactions` instead. clones just the tx hashes

bamrith: wait, `tx_hashes` isn't used anymore ? can just remove this if so

tongzhi1998: Yeah that all make sense. Thank you! 
Forgot the remove the clones after deleting the hash uploads

monad-archive/src/cli.rs
```diff
@@ -0,0 +1,24 @@
+use std::path::PathBuf;
+
+use clap::Parser;
+
+#[derive(Debug, Parser)]
+#[command(name = "monad-archive", about, long_about = None)]
+pub struct Cli {
+    #[arg(long)]
+    pub triedb_path: Option<PathBuf>,
```
ariqchowdhury: we unwrap on this in main so make it required instead

monad-archive/src/s3_archive.rs
```diff
@@ -0,0 +1,257 @@
...
+            .await;
+        let client = Client::new(&config);
+        Ok(S3Archive {
+            client,
+            bucket,
+            block_table_prefix: "block".to_string(),
+            block_hash_table_prefix: "block_hash".to_string(),
+            receipts_table_prefix: "receipts".to_string(),
+            traces_table_prefix: "traces".to_string(),
+            latest_table_key: "latest".to_string(),
```
ariqchowdhury: should these be configurable from cli?

monad-archive/src/triedb.rs
```diff
@@ -0,0 +1,400 @@
...
+
+        if let Err(e) =
+            self.mpsc_sender
+                .try_send(TriedbRequest::BlockNumberRequest(BlockNumberRequest {
+                    request_sender,
+                }))
+        {
+            error!("Polling thread channel full: {e}");
+            return Err(ArchiveError::custom_error(
+                "could not get latest block from triedb".to_string(),
```
ariqchowdhury: error message here should be more accurate. If send over the channel fails, it implies the channel is full and that the triedb polling thread is unable to service requests fast enough.

monad-archive/src/triedb.rs
```diff
@@ -0,0 +1,400 @@
...
+            ));
+        }
+
+        match request_receiver.await {
+            Ok(block_num) => Ok(block_num),
+            Err(e) => {
+                error!("Error when receiving response from polling thread: {e}");
+                Err(ArchiveError::custom_error(
+                    "could not get latest block from triedb".to_string(),
```
ariqchowdhury: receiver error implies the channel was closed from the other side

monad-archive/src/main.rs
```diff
@@ -0,0 +1,132 @@
...
+            "Processing blocks from {} to {}, start time = {:?}",
+            latest_processed_block + 1,
+            block_number,
+            start
+        );
+
+        let mut start_block_number = latest_processed_block;
+        if latest_processed_block != 0 {
+            start_block_number += 1;
+        }
```
ariqchowdhury:
```
let start_block_number = if latest_processed_block == 0 {
   0
} else {
   latest_processed_block + 1
}
```

monad-archive/src/main.rs
```diff
@@ -0,0 +1,132 @@
...
+
+        let join_handles = (start_block_number..block_number + 1).map(|current_block: u64| {
```
ariqchowdhury: `(start_block_number..=block_number)`

monad-archive/src/main.rs
```diff
@@ -0,0 +1,132 @@
...
+    let f_trace = s3_archive.archive_traces(traces, current_block);
+
+    match try_join!(f_block, f_receipt, f_trace) {
+        Ok(_) => {
+            info!("Successfully archived block {}", current_block);
+        }
+        Err(e) => {
+            error!("Error archiving block {}: {:?}", current_block, e);
+        }
+    }
```
ariqchowdhury, line 159: how do we want to handle errors here? if writing the block fails, are the other keys still supposed to be written?

Cargo.toml
```diff
@@ -170,3 +172,5 @@ url = "2.5.0"
 wasm-bindgen = "0.2"
 zerocopy = "0.6"
 zeroize = "1.3"
+aws-config = "1.5.9"
+aws-sdk-s3 = "1.58.0"
```
bamrith: sort

monad-archive/src/s3_archive.rs
tongzhi1998: I think the names right now are kind of straightforward, but I can make them configurable (maybe with default value)?

monad-archive/src/main.rs
tongzhi1998: modified! Thanks.

monad-archive/src/main.rs
```diff
@@ -0,0 +1,152 @@
...
+        let join_handles = (start_block_number..=block_number).map(|current_block: u64| {
+            let triedb = triedb.clone();
+            let s3_archive_writer = s3_archive_writer.clone();
+
+            let semaphore = concurrent_block_semaphore.clone();
+            tokio::spawn(async move {
+                let permit = semaphore
+                    .acquire()
+                    .await
+                    .expect("Got permit to execute a new block");
```
eerkaijun: what's the reason for needing to limit the number of blocks that can be processed here? Edit: ok nvm i read through the slack messages

eerkaijun: also probably best to remove the `.expect(...)` here to prevent crashing the program, instead log an error instead

monad-archive/src/main.rs
```diff
@@ -0,0 +1,152 @@
...
+};
+
+mod archive_interface;
+mod cli;
+mod errors;
+mod s3_archive;
+mod triedb;
+
+#[tokio::main]
+async fn main() -> Result<(), ArchiveError> {
```
eerkaijun, line 20: this should probably be a long running process, so I think we shouldn't terminate the program if there is an ArchiveError.

monad-archive/src/main.rs
```diff
@@ -0,0 +1,152 @@
...
+    info!("Latest processed block is : {}", latest_processed_block);
+
+    // Check for new blocks every 10 ms
+    // Issue requests to triedb, poll data and push to relevant tables
+    loop {
+        sleep(Duration::from_millis(10)).await;
+
+        let block_number = match triedb.get_latest_block().await {
+            Ok(number) => number,
+            Err(e) => return Err(e),
```
eerkaijun, line 42: related to the comment above. in general I think when there's an Err returned from TriedbEnv, we should perhaps just log the error and move on instead of returning here. 

for future enhancements (perhaps in a subsequent PR), we can add retry logic where necessary (e.g. if polling thread channel error is returned, we can retry again after some interval)

tongzhi1998: Thanks Kai for all the comments! I discussed with @ariqchowdhury and @bamrith last week, and we think it makes sense to halt the program when encountering an error. The reason being that let's say a block is corrupted, it's hard for up to re-upload in the future (we have a 'latest' file that tracks the progress, if it's updated even if the block is bad, then it's hard to roll back and reupload). 
I agree that we should have some sort of retry logic.

monad-archive/src/main.rs
bamrith: i think the `.expect` might be reasonable. the only time acquire throws an error is if the semaphore is closed, which should never happen here

monad-archive/src/main.rs
```diff
@@ -0,0 +1,151 @@
...
+            latest_processed_block + 1
+        };
+
+        let start = Instant::now();
+        info!(
+            "Processing blocks from {} to {}, start time = {:?}",
+            start_block_number, block_number, start
+        );
+
+        let join_handles = (start_block_number..=block_number).map(|current_block: u64| {
```
JoeHowarth: For the case that an archiver is trying to catch up after a significant gap, let's break this range into something more manageable. The indexer uses the following logic:
https://github.com/monad-crypto/monad-bft/pull/1270/files#diff-23ed0b1b6c34ef0a8429115015683b25d6dde5eb80ff9ff01e142cc45d6cd70fR75-R76

monad-archive/src/errors.rs
```diff
@@ -0,0 +1,18 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct ArchiveError {
```
JoeHowarth: Consider replacing this error with eyre::Report which captures backtraces and line numbers and is more ergonomic to use.

ex. before
```
   let data = resp.body.collect().await.map_err(|e| {
            error!("Unable to collect response data: {:?}", e);
            ArchiveError::custom_error(format!("Unable to collect response data: {:?}", e))
        })?;
```

after
```
   let data = resp.body.collect().await
                  .wrap_err("Unable to collect response data")?;
```

monad-archive/src/s3_archive.rs
```diff
@@ -0,0 +1,257 @@
...
+use crate::{archive_interface::ArchiveWriterInterface, errors::ArchiveError, triedb::BlockHeader};
+
+const BLOCK_PADDING_WIDTH: usize = 12;
+
+#[derive(Clone)]
+pub struct S3Archive {
+    pub client: Client,
+    pub bucket: String,
+
+    pub latest_table_key: String,
```
JoeHowarth: nit: these can be &'static str or associated constants

monad-archive/src/main.rs
```diff
@@ -0,0 +1,151 @@
...
+            let triedb = triedb.clone();
+            let s3_archive_writer = s3_archive_writer.clone();
+
+            let semaphore = concurrent_block_semaphore.clone();
+            tokio::spawn(async move {
+                let permit = semaphore
+                    .acquire()
+                    .await
+                    .expect("Got permit to execute a new block");
+                let _ = handle_block(&triedb, current_block, &s3_archive_writer).await;
```
JoeHowarth: This drops errors in handle blocks. These must be passed on or else we will assume the block was archived when it was not

tongzhi1998: Nice catch! Thanks!

michael-yxchen:
~~i can approve if merging unblocks testing, but im a little skeptical about error handling. if upload fails, archive will exit. Is that the behavior we want?~~

nvm saw the conversation above

.github/workflows/publish.yaml
```diff
@@ -277,3 +277,50 @@ jobs:
...
+          registry: peach10.devcore4.com
+          username: ${{ secrets.PRIVATE_REGISTRY_USERNAME }}
+          password: ${{ secrets.PRIVATE_REGISTRY_PASSWORD }}
+
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@v5.5.1
+        with:
+          images: |
+            peach10.devcore4.com/monad-crypto/monad-tx-gen
```
michael-yxchen: image name

PR #1259, opened 2024-11-14
Use OS time directly for block timestamp

maxkozlovsky:
Using in-memory copy of local time can result in high variance of deviation from reference clock depending on the time it takes to process the current event. This could result in time adjustments behaving incorrectly.

Instead of in-memory copy, use os time directly for block timestamps. Add a Clock trait to allow the tests to modify local time transparently.

This should improve accuracy of time adjustments.


michael-yxchen:
Discussed offline -  we plan to explore an alternative option to add elapsed duration to `local_time`. It's less intrusive than moving `timestamp_adjuster` from executor into monad state

PR #1261, opened 2024-11-15
add metrics singleton

mikety:


monad-triedb-cache/src/lib.rs
```diff
@@ -58,7 +59,36 @@ where
                 .collect(),
         };
 
+        let unique_addresses_cnt = addresses.iter().unique().count();
```
michael-yxchen, line 62: nit: calculate `unique_addresses.count()` before cache_misses moves it

monad-triedb-cache/src/lib.rs
```diff
@@ -58,7 +59,36 @@ where
...
+                } else {
+                    metrics.backendcache_events.cache_hits = cache_hit_cnt.unwrap();
+                    metrics.backendcache_events.cache_misses = cache_misses_cnt.unwrap();
+                    trace!(
+                        "Updated metrics cache_hits: {:?} , cache_misses: {:?}",
+                        cache_hit_cnt.unwrap(),
+                        cache_misses_cnt.unwrap()
+                    );
+                }
+            }
```
michael-yxchen: I think we can un-nest this from the if clause. That cleans up the similar metrics update in the else case

monad-triedb-utils/Cargo.toml
```diff
@@ -11,6 +11,7 @@ monad-eth-types = { path = "../monad-eth-types" }
 monad-state-backend = { path = "../monad-state-backend" }
 monad-triedb = { path = "../monad-triedb" }
 monad-types = { path = "../monad-types" }
+monad-metrics = { path = "../monad-metrics" }
```
michael-yxchen: nit: sort

monad-node/src/main.rs
```diff
@@ -386,7 +390,12 @@ async fn run(
                 let otel_meter = maybe_otel_meter.as_ref().expect("otel_endpoint must have been set");
                 let state_metrics = state.metrics();
                 let executor_metrics = executor.metrics();
-                send_metrics(otel_meter, &mut gauge_cache, state_metrics, executor_metrics);
+                if let Ok(mut global_metrics)  = METRICS.try_lock() {
+                    send_metrics(otel_meter, &mut gauge_cache, state_metrics, global_metrics.metrics(), executor_metrics);
+                }
+                else {
+                    send_metrics(otel_meter, &mut gauge_cache, state_metrics, &GlobalMetrics::default(), executor_metrics);
+                }
```
michael-yxchen: would sending default metrics reset all counters on grafana to zero? If so, we can change the interface to accept `Option<&GlobalMetrics>` to minimize noise on the frontend

monad-metrics/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+
+use lazy_static::lazy_static;
+use monad_consensus_types::metrics::GlobalMetrics;
+
+pub struct StaticMetrics {
+    metrics: GlobalMetrics,
+}
+
+lazy_static! {
+    pub static ref METRICS: Mutex<StaticMetrics> = Mutex::new(StaticMetrics::new());
```
michael-yxchen: nit: a RwLock makes the interface cleaner, i.e. we will obtain an immutable reference on metrics tick

monad-triedb-cache/src/lib.rs
mikety: the `unique_addresses` is an iterator, calling `count()` will consume it.

monad-node/src/main.rs
mikety: makes sense, have not checked, but since grafane uses `rate()` it should possible to mess the numberswith `0`

monad-metrics/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "monad-metrics"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+monad-consensus-types = { path = "../monad-consensus-types" }
+lazy_static = "1.5.0"
```
michael-yxchen: nit: move version number to workspace Cargo.toml and use {workspace=true} here

monad-triedb-cache/Cargo.toml
```diff
@@ -10,5 +10,6 @@ bench = false
 monad-state-backend = { path = "../monad-state-backend" }
 monad-eth-types = { path = "../monad-eth-types" }
 monad-types = { path = "../monad-types" }
+monad-metrics = { path = "../monad-metrics" }
```
michael-yxchen, line 12: nit: sort it with other monad crates

monad-triedb-cache/src/lib.rs
```diff
@@ -79,6 +81,41 @@ where
...
         }
 
+        if let Ok(mut global_metrics) = METRICS.try_write() {
+            let metrics = global_metrics.metrics();
+
+            let cur_cache_misses_cnt = if cache_misses.is_empty() {
+                0
+            } else {
+                cache_misses.len()
+            };
```
michael-yxchen:
```suggestion
            let cur_cache_misses_cnt = cache_misses.len();
```
does this work?

mikety: updated

michael-yxchen:
@omegablitz brought up that mock-swarm tests will aggregate metrics from different monad states together in the global metrics counter. Is there some way to have one "global" metrics per monad-state?

monad-consensus-types/src/metrics.rs
```diff
@@ -56,6 +56,62 @@ macro_rules! metrics {
...
+    ) => {
+        #[derive(Clone, Copy, Debug, Default, Serialize, Deserialize)]
+        pub struct $class {
+            $(
+                pub $name: u64
+            ),*
+        }
+    };
+}
+
```
andr-dev, line 114: It doesn't make sense here to copy the original macro just to change a struct names. Let's reduce code dupe and modify the original macro.

monad-metrics/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+use lazy_static::lazy_static;
+use monad_consensus_types::metrics::GlobalMetrics;
+
+pub struct StaticMetrics {
+    metrics: GlobalMetrics,
+}
+
+lazy_static! {
+    pub static ref METRICS: RwLock<StaticMetrics> = RwLock::new(StaticMetrics::new());
+}
```
andr-dev, line 12: All of our metrics are currently u64s so we can easily make this lock free by using AtomicU64s.

monad-node/src/main.rs
```diff
@@ -386,7 +390,12 @@ async fn run(
                 let otel_meter = maybe_otel_meter.as_ref().expect("otel_endpoint must have been set");
                 let state_metrics = state.metrics();
                 let executor_metrics = executor.metrics();
-                send_metrics(otel_meter, &mut gauge_cache, state_metrics, executor_metrics);
+                if let Ok(global_metrics)  = METRICS.try_read() {
+                    send_metrics(otel_meter, &mut gauge_cache, state_metrics, Some(global_metrics.read_metrics()), executor_metrics);
+                }
+                else {
+                    send_metrics(otel_meter, &mut gauge_cache, state_metrics, None, executor_metrics);
+                }
```
andr-dev, line 398:
```rust
send_metrics(otel_meter, &mut gauge_cache, state_metrics, METRICS.try_read().ok(), executor_metrics);
```

PR #1262, opened 2024-11-18
Debug endpoints

eerkaijun:


monad-rpc/src/debug.rs
```diff
@@ -84,7 +101,23 @@ pub async fn monad_debug_getRawTransaction<T: Triedb>(
...
-    Err(JsonRpcError::method_not_supported())
+    let Some(TransactionLocation {
+        tx_index,
+        block_num,
+    }) = triedb_env
+        .get_transaction_location_by_hash(params.tx_hash.0)
+        .await?
+    else {
+        return Ok("0x0".to_string());
+    };
```
michael-yxchen: is there a golden standard implementation to compare against? I wonder if we should return an error code here

eerkaijun: great point - double checked with geth, we should indeed return an error. updated

PR #1264, opened 2024-11-19
Add blocksync tests

bamrith:


monad-blocksync/src/blocksync.rs
```diff
@@ -1347,8 +1352,103 @@ mod test {
...
+
+        let requester = BlockSyncSelfRequester::Consensus;
+        let block_range = BlockRange {
+            last_block_id: BlockId(Hash([0x00_u8; 32])),
+            root_seq_num: SeqNum(1),
+        };
+
+        // requesting a block range should initiate a headers request
+        let cmds = context.handle_self_request(requester, block_range);
+        assert_eq!(cmds.len(), 1);
```
ariqchowdhury: should check that its the expected command, not just any command in case the function ever changes and adds new commands. Write a helper function to find it in the cmds list like we do for vote cmds

monad-blocksync/src/blocksync.rs
```diff
@@ -1347,8 +1352,103 @@ mod test {
...
+
+        let request_cmds = find_send_request_commands(&cmds);
+        assert_eq!(request_cmds.len(), 1);
+        let expected_request = BlockSyncRequestMessage::Headers(block_range);
+        match request_cmds[0] {
+            BlockSyncCommand::SendRequest { request, .. } => {
+                assert_eq!(*request, expected_request);
+            }
+            _ => unreachable!(),
+        }
```
ariqchowdhury: putting these in helper functions can help with the test's readability

monad-blocksync/src/blocksync.rs
bamrith: there are other basic tests that check the exact command and values within them. i didn't want to repeat the same but just sanity check the number of cmds here.

monad-blocksync/src/blocksync.rs
bamrith: will refactor

PR #1266, opened 2024-11-19
Handle vpool edge case

philipglazman:
Chain cache is updated at each new block, but we might detect a nonce gap between blocks updates. Check forwarded transactions first before adding to the queued pool.

### DIFF
--- a/monad-rpc/src/vpool.rs
+++ b/monad-rpc/src/vpool.rs
@@ -307,6 +307,20 @@ impl VirtualPool {
             .map(|cur_nonce| nonce.checked_sub(cur_nonce).map(|v| v > 1).unwrap_or(false))
             .unwrap_or(false)
         {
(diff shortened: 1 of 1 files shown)


monad-rpc/src/vpool.rs
```diff
@@ -307,6 +307,22 @@ impl VirtualPool {
...
+            };
+
+            if entry
+                .nonce()
+                .checked_add(1)
+                .map(|v| v == nonce)
+                .unwrap_or(false)
+            {
+                return TxPoolType::Pending;
+            }
```
michael-yxchen: nit: feel free to ignore
```suggestion
            if let Some(entry) = pending.last() {
                if entry
                .nonce()
                .checked_add(1)
                .map(|v| v == nonce)
                .unwrap_or(false)
                {
                    return TxPoolType::Pending;
                }
            }

            
```

PR #1268, opened 2024-11-20
Remove max_reserve_balance in config

eerkaijun:


eerkaijun:
the node.toml is already updated

PR #1274, opened 2024-11-25
Simple checker

tongzhi1998:


monad-archive/src/bin/monad-archive-checker/main.rs
```diff
@@ -0,0 +1,275 @@
...
+            sleep(Duration::from_secs(10)).await;
+            continue;
+        }
+
+        let mut block_hash_maps = HashMap::new();
+        let mut receipts_hash_maps = HashMap::new();
+        let mut traces_hash_maps = HashMap::new();
+
+        // gathering data from each bucket
+        for s3_archive_reader in &s3_archive_readers {
```
JoeHowarth: This is overly complicated. Make 
```
fn asnyc fn get_block_data(readers: &[impl ArchiveReader], block_num: u64) 
   -> Vec<Option<(BucketStr, Block, Receipt, Trace)>>;
```

monad-archive/src/bin/monad-archive-checker/main.rs
```diff
@@ -0,0 +1,275 @@
...
+                &bucket_receipts,
+                &block_number,
+            );
+            let traces_result =
+                pairwise_check(&check_kinds[2], &s3_buckets, &bucket_traces, &block_number);
+
+            if !block_result && !receipts_result && !traces_result {
+                info!("Block {} is consistant across buckets", &block_number);
+            } else {
+                error!("Block {} is inconsistant across buckets", &block_number);
```
JoeHowarth: We should have a data model for faults, consider https://github.com/monad-crypto/monad-bft/blob/b0a08cb5ba0c7aca3c2bb38f6a9f4ebe222cc717/monad-archive/src/bin/monad-index-checker/main.rs#L100-L140

PR #1278, opened 2024-11-26
Refactor S3 bucket checker to use FaultWriter

tongzhi1998:


monad-archive/src/bin/monad-archive-checker/main.rs
```diff
@@ -195,52 +199,115 @@ async fn get_block_data(s3_archive_readers: &[S3Archive], block_number: u64) ->
...
-                    block_number, blocks_data[i].bucket, blocks_data[j].bucket
+                    block_number, bucket1, bucket2
                 );
-                has_error = true;
+                faults.push(Fault::S3InconsistentBlock {
+                    bucket1: bucket1.clone(),
+                    bucket2: bucket2.clone(),
+                });
             } else {
                 info!(
```
JoeHowarth: Likely downgrade this to debug or trace after testing. This will get extremely noisy on info

monad-archive/src/bin/monad-archive-checker/main.rs
```diff
@@ -195,52 +199,115 @@ async fn get_block_data(s3_archive_readers: &[S3Archive], block_number: u64) ->
...
         }
     }
 
-    return has_error;
+    let res = faults.is_empty();
+
+    if !faults.is_empty() {
+        let block_check_result = BlockCheckResult::new(block_number, faults);
+        if let Err(e) = fault_writer.write_fault(block_check_result).await {
+            error!("Failed to write faults for block {}: {:?}", block_number, e);
```
JoeHowarth: Consider also logging the faults BlockCheckResult obj or Vec<Faults> for debugging

monad-archive/src/fault.rs
```diff
@@ -45,6 +45,29 @@ pub enum Fault {
         txhash: String,
         wrong_block_num: u64,
     },
+
+    S3MissingBlock {
```
JoeHowarth, line 53: MissingBlock and S3MissingBlock should be combined. Likely just delete MissingBlock

monad-archive/src/bin/monad-archive-checker/main.rs
```diff
@@ -195,52 +199,115 @@ async fn get_block_data(s3_archive_readers: &[S3Archive], block_number: u64) ->
...
                     "Traces {} is same between bucket '{}' and bucket '{}'",
-                    block_number, blocks_data[i].bucket, blocks_data[j].bucket
+                    block_number, bucket1, bucket2
                 );
             }
         }
     }
 
-    return has_error;
+    let res = faults.is_empty();
```
JoeHowarth: Consider returning the number of faults instead of bool and logging that in the main loop. Not critical but a small ux improvement

monad-archive/src/bin/monad-index-checker/main.rs
```diff
@@ -165,14 +165,16 @@ async fn handle_blocks(
         for fault in &block_check.faults {
             match fault {
                 Fault::ErrorChecking { .. } => metrics.counter("faults_error_checking", 1),
-                Fault::MissingBlock => metrics.counter("faults_missing_blocks", 1),
```
JoeHowarth, line 168: When I said we should remove this variant, I meant we should combine it with S3MissingBlock since it's the same thing. This checker will do very similar things to the archive checker (pulls receipts, block and traces for each block_num, compares contents to dynamodb) so many of the faults will be present in both. 

The index checker needs to be reworked to account for the data format changes, so I'll take care of this, but wanted to point it out.

PR #1283, opened 2024-11-27
Metrics and handle evictions for virtual mempool

philipglazman:
- Adds otel metrics to virtual mempool (commit from https://github.com/monad-crypto/monad-bft/pull/1260)
- Handles entry evictions and fixes a deadlock issue during eviction that was discovered during stressnet testing

Draft until flexnet test PR is up

### DIFF
--- a/monad-rpc/src/main.rs
+++ b/monad-rpc/src/main.rs
@@ -650,17 +650,18 @@ async fn main() -> std::io::Result<()> {
         ))
     });
 
-    let tx_pool = Arc::new(vpool::VirtualPool::new(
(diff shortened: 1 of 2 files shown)


philipglazman:
flexnet tests: https://github.com/monad-crypto/monad-integration/pull/128

omegablitz:
- one big lock
- eviction more granular - evict individual txs instead of all txs for given account (don't evict all entries for a given sender)
- counter for number of *tx* evictions
- remove time-based eviction, only based on capacity + some eviction policy

PR #1285, opened 2024-11-29
Rebase monad-dataplane onto an io_uring-based async runtime

buytenh:
This commit pulls out the bottom layer of monad-dataplane, that is, the
epoll and network socket interfacing logic, replacing it with an async
runtime, and it rewrites the top layer of monad-dataplane from its current
explicit state machine form to Rust async form, letting the Rust compiler
perform the state machine generation instead.

The advantages of these changes include:

- A massive simplification of the TCP message handling code.  The new
  async version of the TCP-delivered message transmission and reception
  code is a fraction of the size of the original code, and making
  changes to and adding features to the async version of the code is
  much less error-prone.

- Converting the UDP handling code to async makes the pacing sleep in
  the transmission/rebroadcast path non-blocking, so that we can continue
  to process inbound UDP frames and handle TCP traffic while we are in
  pacing sleep, and the UDP transmission path can now actually yield to
  the UDP reception path or to the TCP handling code, instead of having
  to handle all messages queued for transmit to completion before any
  other task can do any work.

- The dataplane thread now uses io_uring to perform I/O, which allows
  submitting multiple I/O operations per kernel entry/exit, and allows
  the use of true asynchronous I/O.

- The UDP reception code now responds to backpressure from the upper
  layer, which means that if RaptorCast is slow to accept UDP packets
  from dataplane, the dataplane thread will temporarily stop reading
  UDP packets from the kernel until the backlog clears, which should
  reduce the amount of cache thrashing when we are under heavy RX load,
  and in case of RX overload, it will push the locus of dropping packets
  out of userspace and back into the kernel, reducing the amount of CPU
  consumed if congestion-induced packet dropping needs to happen.

For the async runtime, we chose monoio, because it is a single-threaded
runtime, it natively supports io_uring, it allows receiving data into
uninitialized buffers, and it uses a buffer ownership model that seems
to be the correct one for use with io_uring -- that is, ownership of
receive/transmit buffers is transferred to the runtime when I/O is
started, and the runtime returns ownership of these buffers to the user
when I/O is complete, which avoids the potential issue involving a Future
that has I/O in flight being synchronously dropped causing the buffer
for the I/O to be freed before the I/O is complete, and it avoids the
runtime having to use bounce buffering for all data submitted for I/O,
which is something that e.g. glommio seems to have to do.

Some testing with monad-testground shows a ~2.7% improvement in round
times with 10 nodes, from 2036 ms for 100 rounds of consensus to 1981 ms,
with a t-test showing a significant difference at a 99.5% confidence
level.  This is presumably because of pacing sleep now being non-blocking,
and the advantages that that brings.

Note that while the async UDP module added by this commit supports
UDP transmit fragmentation offload (UDP_FRAGMENT), it does not (yet)
support UDP_GRO, as the use of UDP_GRO appears to lead to slightly worse
(~1%) monad-testground round times, and this needs further investigation
before we merge this feature.

Since it was very easy to add, the async TCP module added by this commit
additionally supports receiver-side persistent connections.  Once this
will have been rolled out everywhere, we will be able to add transmit
pipelining to the TCP sender.

### DIFF
--- a/monad-dataplane/tests/tests.rs
+++ b/monad-dataplane/tests/tests.rs
@@ -145,3 +145,90 @@ fn tcp_rapid() {
         assert_eq!(received_payload, payload);
     }
 }
+
+const MINIMUM_SEGMENT_SIZE: u16 = 256;
+
+#[test]
+#[timeout(30000)]
+fn broadcast_all_strides() {
+    once_setup();
+
+    let rx_addr = "127.0.0.1:9008".parse().unwrap();
+    let tx_addr = "127.0.0.1:9009".parse().unwrap();
+
+    let mut rx = Dataplane::new(&rx_addr, UP_BANDWIDTH_MBPS);
+    let mut tx = Dataplane::new(&tx_addr, UP_BANDWIDTH_MBPS);
+
+    // Allow Dataplane threads to set themselves up.
+    sleep(Duration::from_millis(10));
+
+    let total_length: usize = 100000;
+
+    let payload: Vec<u8> = (0..total_length)
+        .map(|_| rand::thread_rng().gen_range(0..255))
+        .collect();
+
+    for stride in MINIMUM_SEGMENT_SIZE..=DEFAULT_SEGMENT_SIZE {
+        tx.udp_write_broadcast(BroadcastMsg {
+            targets: vec![rx_addr],
+            payload: payload.clone().into(),
+            stride,
+        });
+
+        let stride = stride.into();
+
+        let num_msgs = total_length.div_ceil(stride);
+
+        for i in 0..num_msgs {
+            let msg: RecvMsg = executor::block_on(rx.udp_read());
+
+            assert_eq!(msg.src_addr, tx_addr);
+            assert_eq!(
+                &msg.payload[..],
+                &payload[i * stride..((i + 1) * stride).min(payload.len())]
+            );
+        }
+    }
+}
+
+#[test]
+#[timeout(30000)]
+fn unicast_all_strides() {
+    once_setup();
+
+    let rx_addr = "127.0.0.1:9010".parse().unwrap();
+    let tx_addr = "127.0.0.1:9011".parse().unwrap();
+
+    let mut rx = Dataplane::new(&rx_addr, UP_BANDWIDTH_MBPS);
+    let mut tx = Dataplane::new(&tx_addr, UP_BANDWIDTH_MBPS);
+
+    // Allow Dataplane threads to set themselves up.
+    sleep(Duration::from_millis(10));
+
+    let total_length: usize = 100000;
+
+    let payload: Vec<u8> = (0..total_length)
+        .map(|_| rand::thread_rng().gen_range(0..255))
+        .collect();
+
+    for stride in MINIMUM_SEGMENT_SIZE..=DEFAULT_SEGMENT_SIZE {
+        tx.udp_write_unicast(UnicastMsg {
+            msgs: vec![(rx_addr, payload.clone().into())],
+            stride,
+        });
+
+        let stride = stride.into();
+
+        let num_msgs = total_length.div_ceil(stride);
+
+        for i in 0..num_msgs {
+            let msg: RecvMsg = executor::block_on(rx.udp_read());
+
+            assert_eq!(msg.src_addr, tx_addr);
+            assert_eq!(
+                &msg.payload[..],
+                &payload[i * stride..((i + 1) * stride).min(payload.len())]
+            );
+        }
+    }
+}


monad-dataplane/src/lib.rs
```diff
@@ -1,35 +1,163 @@
...
+    pub msgs: Vec<(SocketAddr, Bytes)>,
+    pub stride: u16,
+}
+
+#[derive(Clone)]
+pub struct RecvMsg {
+    pub src_addr: SocketAddr,
+    pub payload: Bytes,
+    pub stride: u16,
+}
```
andr-dev, line 39: Can we move these messages into a separate file like `message.rs`?

monad-dataplane/src/lib.rs
```diff
@@ -1,35 +1,163 @@
...
+
+impl Dataplane {
+    /// 1_000 = 1 Gbps, 10_000 = 10 Gbps
+    pub fn new(local_addr: &SocketAddr, up_bandwidth_mbps: u64) -> Self {
+        let local_addr = *local_addr;
+
+        let (tcp_ingress_tx, tcp_ingress_rx) = mpsc::channel(1024);
+        let (tcp_egress_tx, tcp_egress_rx) = mpsc::channel(1024);
+        let (udp_ingress_tx, udp_ingress_rx) = mpsc::channel(12_800);
+        let (udp_egress_tx, udp_egress_rx) = mpsc::channel(12_800);
```
andr-dev: Let's move these magic numbers into some nicely named constants up above.

monad-dataplane/src/lib.rs
```diff
@@ -1,35 +1,163 @@
...
+
+        thread::spawn(move || {
+            RuntimeBuilder::<IoUringDriver>::new()
+                .enable_timer()
+                .build()
+                .expect("Failed building the Runtime")
```
andr-dev, line 60: https://doc.rust-lang.org/std/result/enum.Result.html#recommended-message-style

monad-dataplane/src/lib.rs
```diff
@@ -1,35 +1,163 @@
...
+                .block_on(async move {
+                    tcp::spawn_tasks(local_addr, tcp_ingress_tx, tcp_egress_rx);
+
+                    udp::spawn_tasks(local_addr, udp_ingress_tx, udp_egress_rx, up_bandwidth_mbps);
+
+                    loop {
+                        time::sleep(Duration::from_secs(3600)).await;
```
andr-dev: There's `Sleep::far_future()` so we don't have a magic number in there (far_future is 30 years).

monad-dataplane/src/lib.rs
```diff
@@ -1,35 +1,163 @@
...
+            tcp_msgs_dropped: 0,
+            udp_msgs_dropped: 0,
+        }
+    }
+
+    pub async fn tcp_read(&mut self) -> (SocketAddr, Bytes) {
+        match self.tcp_ingress_rx.recv().await {
+            None => panic!("tcp_ingress_rx channel closed"),
+            Some(msg) => msg,
+        }
```
andr-dev:
```rust
self.tcp_ingress_rx.recv().await.expect("tcp_ingress_rx never closed")
```

monad-dataplane/src/lib.rs
```diff
@@ -1,35 +1,163 @@
...
+            }
+            Err(TrySendError::Closed(_)) => panic!("tcp_egress_tx channel closed"),
+        }
+    }
+
+    pub async fn udp_read(&mut self) -> RecvMsg {
+        match self.udp_ingress_rx.recv().await {
+            None => panic!("udp_ingress_rx channel closed"),
+            Some(msg) => msg,
+        }
```
andr-dev:
```rust
self.udp_ingress_rx.recv().await.expect("udp_ingress_rx never closed")
```

monad-dataplane/src/lib.rs
```diff
@@ -1,35 +1,163 @@
...
+                    self.udp_msgs_dropped += num_msgs_dropped;
+
+                    warn!(
+                        num_msgs_dropped,
+                        total_udp_msgs_dropped = self.udp_msgs_dropped,
+                        msg_length = msg.payload.len(),
+                        "udp_egress_tx channel full, dropping message"
+                    );
+
+                    return;
```
andr-dev, line 136: Why do we want to stop broadcasting if one peer's channel is full? Maybe I'm lacking some context here, in which case a comment would be helpful.

monad-dataplane/src/tcp/rx.rs
```diff
@@ -0,0 +1,274 @@
...
+    while let Some(message) = read_message(conn_id, addr, message_id, &mut tcp_stream).await {
+        if let Err(err) = tcp_ingress_tx.send((addr, message)).await {
+            warn!(
+                conn_id,
+                ?addr,
+                message_id,
+                ?err,
+                "error queueing up received TCP message",
+            );
+            return;
```
andr-dev: Should we break here instead of return? That way, it'll call the `rx_state.drop_connection(...)` which feels like it would cleanly disconnect the peer. Again, might be lacking context in which case a comment would be good.

monad-dataplane/src/tcp/rx.rs
```diff
@@ -0,0 +1,274 @@
...
+    conn_id: u64,
+    addr: SocketAddr,
+    message_id: u64,
+    tcp_stream: &mut TcpStream,
+) -> Option<Bytes> {
+    let start_time = Instant::now();
+
+    let header_bytes = BytesMut::with_capacity(std::mem::size_of::<TcpMsgHdr>());
+
+    let header = match timeout(TCP_HEADER_TIMEOUT, tcp_stream.read_exact(header_bytes)).await {
```
andr-dev, line 147: Here, we probably want to make sure that if the structure of header changes in the future that we don't forget to check those values when reading a message. In places like these, I like to do the little trick of
```rust
let TcpMsgHdr {
  magic,
  version,
} = match ... {
  ...
};
```
If someone adds a field to TcpMsgHdr, they're forced to add the field here as well so it's clear they should probably do something with it :)

monad-dataplane/src/tcp/rx.rs
```diff
@@ -0,0 +1,274 @@
...
+
+    let message_length: usize = header.length.get() as usize;
+
+    if message_length > TCP_MESSAGE_LENGTH_LIMIT {
+        debug!(
+            conn_id,
+            ?addr,
+            message_id,
+            ?header,
+            "received oversized message on TCP connection"
```
andr-dev: Super nit: we haven't received an oversized message but rather we've gotten an indicator that the peer is about to send us an oversized message. Maybe something along the lines of `"received header with oversized message len on TCP connection"`.

monad-dataplane/src/tcp/tx.rs
```diff
@@ -0,0 +1,219 @@
...
+                    conn_id,
+                    ?addr,
+                    ?header,
+                    ?err,
+                    "error writing message on TCP connection"
+                );
+                return;
+            }
+
+            conn_cork(raw_fd, false);
```
andr-dev, line 154: Why do we set `TCP_CORK` before sending every message just to disable it right after? This seems like an unnecessary number of syscalls.

monad-dataplane/src/tcp/tx.rs
```diff
@@ -0,0 +1,219 @@
...
+
+            let num_unacked_bytes = num_unacked_bytes(raw_fd);
```
andr-dev, line 156: `num_unacked_bytes` makes a syscall which we should avoid, along with the rest of the code below, if we're not in debug level. Let's prefix this with
```rust
if !enabled!(Level::DEBUG) {
  return;
}
```

monad-dataplane/src/udp.rs
```diff
@@ -0,0 +1,177 @@
...
+    local_addr: SocketAddr,
+    udp_ingress_tx: mpsc::Sender<RecvMsg>,
+    udp_egress_rx: mpsc::Receiver<(SocketAddr, Bytes, u16)>,
+    up_bandwidth_mbps: u64,
+) {
+    // Bind the UDP socket and clone it for use in different tasks.
+    let udp_socket_rx = UdpSocket::bind(local_addr).unwrap();
+    let raw_fd = udp_socket_rx.as_raw_fd();
+    let udp_socket_tx =
+        UdpSocket::from_std(unsafe { std::net::UdpSocket::from_raw_fd(raw_fd) }).unwrap();
```
andr-dev, line 41: We can get rid of the unsafe here by using monoio's Split/Splitable trait.
```rust
UdpSocket::bind(local_addr).unwrap().into_split()
```
This also enforces that rx doesn't send and tx doesn't receive.

monad-dataplane/src/udp.rs
```diff
@@ -0,0 +1,177 @@
...
+
+    spawn(rx(udp_socket_rx, udp_ingress_tx));
+    spawn(tx(udp_socket_tx, udp_egress_rx, up_bandwidth_mbps));
+}
+
+async fn rx(udp_socket_rx: UdpSocket, udp_ingress_tx: mpsc::Sender<RecvMsg>) {
+    loop {
+        let buf = BytesMut::with_capacity(ETHERNET_SEGMENT_SIZE.into());
+
+        if let (Ok((len, src_addr)), buf) = udp_socket_rx.recv_from(buf).await {
```
andr-dev: If `recv_from` produces an Err(...), won't this will keep looping endlessly?

monad-dataplane/src/udp.rs
```diff
@@ -0,0 +1,177 @@
...
+    let mut messages_to_send: VecDeque<(SocketAddr, Bytes, u16)> = VecDeque::new();
+
+    loop {
+        while messages_to_send.is_empty() || !udp_egress_rx.is_empty() {
+            let Some((addr, payload, stride)) = udp_egress_rx.recv().await else {
+                return;
+            };
+
+            messages_to_send.push_back((addr, payload, stride));
+        }
```
andr-dev, line 115:
```rust
if messages_to_send.is_empty() {
  let Some((addr, payload, stride)) = udp_egress_rx.recv().await else {
    return;
  };
  messages_to_send.push_back((addr, payload, stride));
  
  while let Ok((addr, payload, stride)) = udp_egress_rx.try_recv() {
    messages_to_send.push_back((addr, payload, stride));
  }
}
```
This avoids the async await point if `udp_egress_rx` has multiple values ready.

monad-dataplane/src/udp.rs
```diff
@@ -0,0 +1,177 @@
...
+        let chunk_len = chunk.len();
+
+        let now = Instant::now();
+
+        if next_transmit > now {
+            time::sleep(next_transmit - now).await;
+        } else {
+            let late = now - next_transmit;
+
+            if late > Duration::from_millis(100) {
```
andr-dev: Let's make this a const as well.

monad-dataplane/src/tcp/tx.rs
```diff
@@ -0,0 +1,219 @@
...
+use monoio::{io::AsyncWriteRentExt, net::TcpStream, spawn, time::timeout};
+use tokio::sync::mpsc;
+use tracing::{debug, trace, warn};
+use zerocopy::AsBytes;
+
+use super::{TcpMsgHdr, TCP_MESSAGE_TIMEOUT};
+
+#[derive(Clone)]
+struct TxState {
+    inner: Rc<RefCell<TxStateInner>>,
```
andr-dev, line 21: I would add a comment here that it's safe to use an `Rc<RefCell<T>>` since monoio is thread-per-core so no two tasks with a TxState can be running in parallel. Took me a few minutes to realize this as I was contemplating whether the functions below might panic or not.

monad-dataplane/src/tcp/tx.rs
buytenh: `stream.write_all()` can take many seconds, or even minutes, to complete for a large enough message (note that TCP-delivered messages can currently be up to 1 GiB in size), and one `stream.write_all()` call can translate into thousands of individual io_uring operations.  We need to have `TCP_CORK` set that entire time, to make sure that the write being split on non-MTU boundaries (due to socket buffer limits) can't cause a partial frame to be emitted, and we need to clear `TCP_CORK` afterwards to make sure that pending data in the socket transmit buffer gets sent out even if the tail end of it doesn't completely fill a TCP segment -- so I don't think we can avoid doing it this way.

monad-dataplane/src/lib.rs
buytenh: The udp_egress_tx channel is not a per-peer channel -- there's only one of these per Dataplane.  And if it's full right now, it's very likely to still be full a few nanoseconds later.  Does that make sense, or do you still feel that a clarifying comment is in order?

monad-dataplane/src/tcp/rx.rs
buytenh: If/when this happens, the main thread has already exited, which means that the Dataplane thread is probably also about to be terminated, in which case the number-of-ingress-TCP-connections-per-peer accounting that is done by `rx_state.drop_connection(...)` doesn't matter all that much anymore -- but I'll change this to a break (and the same for the return in `udp.rs:rx()` because mid-function returns are a bit icky.

monad-dataplane/src/tcp/rx.rs
buytenh: Sounds good, will make that change!

monad-dataplane/src/lib.rs
buytenh: I figured it made sense to put this in `lib.rs` since that file now defines the entire interface between Dataplane and its users, so you can easily understand the interface by reading that file, and the file isn't very big as it is -- but I don't feel overwhelmingly strongly about this.

monad-dataplane/src/lib.rs
buytenh: `Sleep::far_future()` appears to be `pub(crate)`:

```
62  |                         time::Sleep::far_future().await;
    |                                      ^^^^^^^^^^ private associated function
```

And so is `Instant::far_future()` which `Sleep::far_future()` uses internally:

```
62 | ...time::Instant::far_future()).await;
   |                   ^^^^^^^^^^ private associated function
```

monad-dataplane/src/tcp/tx.rs
buytenh: Sounds good, will fix!  I'll also see if I can get rid of `let connect_time = Instant::now();` above if `DEBUG` is not enabled.

monad-dataplane/src/tcp/tx.rs
buytenh: If monoio were a work-stealing multi-threaded runtime like tokio, then I think that we would not be able to pass an `Rc<>` to another task in the first place, as `Rc<>` is not `Send`, i.e. this would fail to compile.  And passing an `Arc<RefCell<>>` would make little sense.  In the case that one of our Dataplane tasks does inadvertently keep the `RefCell<>` borrowed across an await point, we do want that to panic.

monad-dataplane/src/udp.rs
buytenh: The proposed version avoids pulling from `udp_egress_rx` if there are messages in `messages_to_send`, while the intent was to keep pulling from `udp_egress_rx` even if there are still messages in `messages_to_send`, as we round-robin over the elements of `messages_to_send` while transmitting, which makes the transmission process more fair.

In the (near) future, we will also add message prioritization, and we will e.g. prioritize sending RaptorCast chunks to validators over sending RaptorCast chunks to full nodes, and if we are on the tail end of round `N` and we are still distributing the proposal among our locally-attached full nodes, but then round `N + 1` starts and a new proposal is ready for transmission, we want transmission of the round `N + 1` proposal to peer validators to take precedence over transmitting the round `N` proposal to our full nodes.  To achieve that, we will also need to keep pulling from `udp_egress_rx` even if we still have messages to send.

I can understand the desire to avoid unnecessary task switching, but I don't get why avoiding await points per se is desirable.  Is the point of that to reduce the size of the `Future`, or are there other reasons to want to avoid that?

monad-dataplane/src/lib.rs
omegablitz: https://docs.rs/futures/latest/futures/future/fn.pending.html

monad-dataplane/src/lib.rs
buytenh: Fixed locally!

monad-dataplane/src/lib.rs
buytenh: Fixed locally!

monad-dataplane/src/lib.rs
buytenh: Can you suggest how this should be reworded?

I have read the linked fragment, and it suggests focusing on why I would think that the result of this method should be Ok, but I don't know all the ways in which this (rather opaque) method can fail or why I would think that none of those potential failure modes should apply to my invocation of this method -- all I know is that if this operation fails, Dataplane initialization cannot continue.

monad-dataplane/src/lib.rs
buytenh: I don't disagree with this, and I'll make that change, but this is (almost) exactly the same as what the existing code does:

```
-fn make_producer_consumer<T>(size: usize) -> (WakeableProducer<T>, WakeableConsumer<T>) {
-    let (ing_send, ing_recv) = RingBuffer::<T>::new(size);
[...]
-
-impl Dataplane {
-    /// 1_000 = 1 Gbps, 10_000 = 10 Gbps
-    pub fn new(local_addr: &SocketAddr, up_bandwidth_mbps: u64, default_mtu: u16) -> Self {
-        let (udp_ing_producer, udp_ing_consumer) = make_producer_consumer(12_800);
-        let (tcp_ing_producer, tcp_ing_consumer) = make_producer_consumer(32);
-
-        let (egr_bcast_send, egr_bcast_recv) = RingBuffer::<BroadcastMsg>::new(2048);
-        let (egr_ucast_send, egr_ucast_recv) = RingBuffer::<UnicastMsg>::new(1024);
-        let (egr_tcp_send, egr_tcp_recv) = RingBuffer::<(SocketAddr, Bytes)>::new(1024);
```

And this pattern is found in a number of other places in the monad-bft codebase as well.

If we want to hold new code to a (much) higher standard than our existing code, then I feel that we should also make an effort to bring our existing code up to the same standard, because if our existing code fails to meet our coding standards, then something is wrong either with our code or our coding standards.

monad-dataplane/src/lib.rs
buytenh: > https://docs.rs/futures/latest/futures/future/fn.pending.html

I'll use that, thanks!

monad-dataplane/src/tcp/rx.rs
buytenh: I want to keep `header`, as I feed `?header` into the various trace events in this function, so I'll instead insert this after the match block:

```
-    if header.magic.get() != HEADER_MAGIC {
+    let TcpMsgHdr {
+        magic: header_magic,
+        version: header_version,
+        length: header_length,
+    } = header;
+
+    if header_magic.get() != HEADER_MAGIC {
```

And switch the various references to `header.foo` over to `header_foo`.

monad-dataplane/src/udp.rs
buytenh: I tried this:

```
@@ -2,12 +2,12 @@ use std::{
     collections::VecDeque,
     io::{Error, ErrorKind},
     net::SocketAddr,
-    os::fd::{AsRawFd, FromRawFd},
+    os::fd::AsRawFd,
     time::{Duration, Instant},
 };
 
 use bytes::{Bytes, BytesMut};
-use monoio::{net::udp::UdpSocket, spawn, time};
+use monoio::{io::Splitable, net::udp::UdpSocket, spawn, time};
 use tokio::sync::mpsc;
 use tracing::warn;
 
@@ -34,13 +34,11 @@ pub fn spawn_tasks(
     udp_egress_rx: mpsc::Receiver<(SocketAddr, Bytes, u16)>,
     up_bandwidth_mbps: u64,
 ) {
-    // Bind the UDP socket and clone it for use in different tasks.
-    let udp_socket_rx = UdpSocket::bind(local_addr).unwrap();
-    let raw_fd = udp_socket_rx.as_raw_fd();
-    let udp_socket_tx =
-        UdpSocket::from_std(unsafe { std::net::UdpSocket::from_raw_fd(raw_fd) }).unwrap();
+    let (udp_socket_rx, udp_socket_tx) = UdpSocket::bind(local_addr).unwrap().into_split();
 
     {
+        let raw_fd = udp_socket_rx.as_raw_fd();
+
         const MTU_DISCOVER: libc::c_int = libc::IP_PMTUDISC_OMIT;
 
         if unsafe {
```

But it seems that `UdpSocket` doesn't get an implementation of `Splitable`:

```
error[E0599]: the method `into_split` exists for struct `UdpSocket`, but its trait bounds were not satisfied
  --> monad-dataplane/src/udp.rs:37:79
   |
37 |     let (udp_socket_rx, udp_socket_tx) = UdpSocket::bind(local_addr).unwrap().into_split();
   |                                                                               ^^^^^^^^^^ method cannot be called on `UdpSocket` due to unsatisfied trait bounds
   |
  ::: /home/buytenh/.cargo/registry/src/index.crates.io-6f17d22bba15001f/monoio-0.2.4/src/net/udp.rs:28:1
   |
28 | pub struct UdpSocket {
   | -------------------- doesn't satisfy `monoio::net::udp::UdpSocket: AsyncWriteRent` or `monoio::net::udp::UdpSocket: Splitable`
   |
   = note: the following trait bounds were not satisfied:
           `monoio::net::udp::UdpSocket: AsyncWriteRent`
           which is required by `monoio::net::udp::UdpSocket: Splitable`
           `&monoio::net::udp::UdpSocket: monoio::io::Split`
           which is required by `&monoio::net::udp::UdpSocket: Splitable`
           `&monoio::net::udp::UdpSocket: AsyncWriteRent`
           which is required by `&monoio::net::udp::UdpSocket: Splitable`
           `&mut monoio::net::udp::UdpSocket: monoio::io::Split`
           which is required by `&mut monoio::net::udp::UdpSocket: Splitable`
           `&mut monoio::net::udp::UdpSocket: AsyncWriteRent`
           which is required by `&mut monoio::net::udp::UdpSocket: Splitable`
```

It does work for `TcpStream` as `TcpStream` implements `AsyncWriteRent`, and the `TcpStream` documentation also mentions types for owned read and write halves, where there is no such thing for `UdpSocket`:  https://docs.rs/monoio/latest/monoio/net/tcp/index.html

monad-dataplane/src/udp.rs
buytenh: Yes.  The expected errors here are transient (e.g. `-EINTR`), but I'll change this into a `match udp_socket_rx.recv_from(buf).await {` and I'll `warn!` if we get an `Err(..)`, so that we won't loop silently in case of an unexpected non-transient error, and we'll be able to determine from event tracing what happened.

monad-dataplane/src/udp.rs
buytenh: Done, with the same comment here as above, given that this code was lifted verbatim from dataplane v1:

```
-                } else {
-                    let late = now - self.next_transmit;
-
-                    if late > std::time::Duration::from_millis(100) {
-                        self.next_transmit = now;
-                    }
-                }
```

monad-dataplane/src/udp.rs
andr-dev: Interesting, yeah UdpSocket implements Split but not Splitable because of that lacking AsyncWriteRent. There's probably some nuance here we're missing.

monad-dataplane/src/tcp/tx.rs
andr-dev: Got it, setting `TCP_CORK` made sense, I was more asking why we disable it each time. Can we add a comment on the `conn_cork(raw_fd, false)` that we need to disable `TCP_CORK` to make sure pending data gets sent out?

monad-dataplane/src/udp.rs
andr-dev: At each await point, the task yields back to the async runtime which introduces an additional cost to each message. Additionally, if we just want to read all messages that are already available, then try_recv() is the method to call. If we want to wait until a message is available, then recv().await is the method to call. Thus, calling is_empty() followed by recv().await is contradictory in intention.

As a simple fix to the fairness you mentioned, we can just move the while outside the if.
```rust
if messages_to_send.is_empty() {
  let Some((addr, payload, stride)) = udp_egress_rx.recv().await else {
    return;
  };
  messages_to_send.push_back((addr, payload, stride));
}

while let Ok((addr, payload, stride)) = udp_egress_rx.try_recv() {
  messages_to_send.push_back((addr, payload, stride));
}
```
This reads as:
1. If we don't have any messages to send, then we want to asynchronously wait until one is available.
2. While there are messages in the channel readily available, add them to messages_to_send.

The only thing to watch out for here is that if udp_egress_rx has a ton of messages, we'll continue to block the thread while reading them in.

monad-dataplane/src/lib.rs
andr-dev: Something along the lines of "System should support IoUringDriver runtime". This specifies why we expect this RuntimeBuilder to succeed while not specifying exactly why the call failed.

monad-dataplane/src/lib.rs
andr-dev: Yep that makes sense, if you can drop in that comment in another PR that would be nice :)

monad-dataplane/src/tcp/tx.rs
buytenh: Actually, I just tested that removing the explicit uncorking does not appear to have any ill effects, as we close the connection immediately afterwards, and that also effectively uncorks the connection.

Once dataplane v2 will have been rolled out everywhere, we want to merge https://github.com/category-labs/monad-bft/tree/buytenh/dataplane-async-tx-pipeline to enable TCP transmit pipelining, and as part of that change, we cork new connections only once, when they are established, and then never uncork them.  Since not explicitly uncorking constitutes a (small) performance improvement, we can perhaps wait with resolving this particular issue for now, and wait for this TCP transmit pipelining change to be merged, which will address this issue.

monad-dataplane/src/tcp/tx.rs
buytenh: The commit at https://github.com/category-labs/monad-bft/commits/buytenh/dataplane-reduce-instant-now-calls/ also gets rid of the `Instant::now()` call, as well as another `Instant::now()` call, if `DEBUG` is not enabled.  I'll PR that change sometime soon.

PR #1288, opened 2024-12-03
Set MTU discovery mode to ignore spoofed ICMP frag_needed

eclark0426:
See discussions on this in https://github.com/monad-crypto/monad-internal/issues/742

This resolves the vulnerability where an arbitrary user on the network can spoof ICMP fragmentation needed packets between every pair of nodes to cause EINVAL errors. However, I think there is still some testing that would be worthwhile to determine what would happen to a node that legitimately has a lower MTU. But, this testing is difficult to do with our currently available testing setups.

### DIFF
--- a/monad-dataplane/src/network.rs
+++ b/monad-dataplane/src/network.rs
@@ -124,6 +124,23 @@ impl<'a> NetworkSocket<'a> {
             panic!("set GRO failed with: {}", std::io::Error::last_os_error());
         }
 
+        let r = unsafe {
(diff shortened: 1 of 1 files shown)


buytenh:
Is `IP_PMTUDISC_OMIT` the only of the 6 values that works?

Does it still set DF on outbound packets but ignore all inbound icmp frag-neededs with this option?  Or does it clear DF?

eclark0426:
`OMIT` and `INTERFACE` resolve the issue. I opted for `OMIT` since it will respect the interface's MTU and allow packets to be fragmented if they exceed the interface MTU. It will clear the DF flag

PR #1289, opened 2024-12-03
Staging archiver

eerkaijun:
@JoeHowarth @tongzhi1998 opening a draft PR so it's easier for me to leave some comments/questions. feel free to close this off if another branch is the canonical one


monad-archive/src/archive_reader.rs
```diff
@@ -0,0 +1,63 @@
...
+        table: String,
+        region: Option<String>,
+        concurrency: usize,
+        metrics: &Metrics,
+    ) -> ArchiveReader {
+        let sdk_config = get_aws_config(region).await;
+        ArchiveReader {
+            s3: S3Archive::new(S3Bucket::new(bucket, &sdk_config, metrics.clone())),
+            dynamodb: DynamoDBArchive::new(table, &sdk_config, concurrency, metrics.clone()),
+        }
```
eerkaijun, line 35: just to double confirm - will s3 and dynamodb always run in the same region?

monad-archive/src/archive_reader.rs
```diff
@@ -0,0 +1,63 @@
...
+    dynamodb: DynamoDBArchive,
+}
+
+impl ArchiveReader {
+    pub async fn new(
+        bucket: String,
+        table: String,
+        region: Option<String>,
+        concurrency: usize,
+        metrics: &Metrics,
```
eerkaijun, line 29: can this be made `Option<Metrics>` instead of `Metrics`?

monad-archive/src/dynamodb.rs
```diff
@@ -0,0 +1,335 @@
...
+                })
+            })
+            .collect::<Vec<Option<TxIndexedData>>>();
+        Ok(output)
+    }
+
+    pub async fn get_txdata(&self, key: impl Into<String>) -> Result<Option<TxIndexedData>> {
+        self.batch_get_txdata(&[key.into()])
+            .await
+            .map(|mut v| v.remove(0))
```
eerkaijun: `v.remove(0)` crashes the program if `key` (the tx hash) does not exist in archive, we should error handle here

monad-archive/src/dynamodb.rs
```diff
@@ -0,0 +1,335 @@
...
+    ) -> Result<()>;
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Default, RlpEncodable, RlpDecodable)]
+pub struct TxIndexedData {
+    pub block_num: u64,
+    pub tx: TransactionSigned,
+    pub trace: Vec<u8>,
+    pub receipt: ReceiptWithBloom,
+}
```
eerkaijun: for RPC we will need to know the `tx_index` of the transaction too (i.e. the position of the transaction in the block). is there a way we can store that too

monad-archive/src/archive_reader.rs
```diff
@@ -0,0 +1,63 @@
...
+        self.s3.get_block_by_number(block_num).await
+    }
+
+    pub async fn get_block_receipts(&self, block_number: u64) -> Result<Vec<ReceiptWithBloom>> {
+        self.s3.get_block_receipts(block_number).await
+    }
+
+    pub async fn get_block_traces(&self, block_number: u64) -> Result<Vec<Vec<u8>>> {
+        self.s3.get_block_traces(block_number).await
+    }
```
eerkaijun, line 71: what would be the best way to get block traces by hash? would you recommend first calling `get_block_by_hash` in archive then `get_block_traces`?

tongzhi1998: Yes, I think that's the best way.

monad-archive/src/archive_reader.rs
JoeHowarth: yes

PR #1293, opened 2024-12-04
Pin Rust version

eclark0426:
The Rust toolchain was not actually being pinned in the Github action, like intended, which started causing lint issues after the stable version was bumped last week. This pins the nightly version that was intended by the action, but using a toolchain file instead


eclark0426:
If we'd prefer to only pin the version for the Github action, I can instead just set the default version as a command step in the action. However, I figured using a toolchain file would make it consistent across all github actions and developers.

omegablitz:
Should we also pin the stable version?

eclark0426:
I don't think there's a way to pin both - we'd need to choose either a nightly version or a stable version

PR #1296, opened 2024-12-04
Fixes for Virtual Mempool

philipglazman:
Replaces https://github.com/monad-crypto/monad-bft/pull/1283

Includes:
- Metrics 
- Removes time-based evictions and evicts based on transaction entries, not the entire sender's entry
- uses recovered_signers_unchecked for perf
- uses single lock for chaincache instead of locks per field


philipglazman:
@omegablitz let me know what branch includes your fixes

monad-eth-txpool/src/test_utils.rs
```diff
@@ -0,0 +1,205 @@
+use std::collections::{BTreeMap, VecDeque};
```
andr-dev: Why are we moving this file into monad-eth-txpool? This is a test harness and should live in monad-eth-txpool/tests.

monad-rpc/Cargo.toml
```diff
@@ -51,3 +52,11 @@ alloy-rpc-client = { workspace = true }
...
 test-case = { workspace = true }
 tempfile = { workspace = true }
+monad-eth-txpool = { path = "../monad-eth-txpool"}
+monad-consensus-types = { path = "../monad-consensus-types" }
+monad-state-backend = { path = "../monad-state-backend" }
+monad-eth-types = { path = "../monad-eth-types"}
+monad-eth-testutil = { path = "../monad-eth-testutil" }
+itertools = { workspace = true }
+monad-testutil = { path = "../monad-testutil" }
+rand = { workspace = true }
```
andr-dev: These monad-* dependencies should be at the top with the others. They should also be alphabetically sorted.

monad-eth-txpool/src/test_utils.rs
philipglazman: I'd like to use some types from the test harness in vpool. I can't import the types if there are in `/tests`. Any suggested workarounds ?

andr-dev: Ah in that case, we'd probably want to move those types to monad-eth-testutil and keep the more txpool specific things in the monad-eth-txpool tests directory.

philipglazman: That would cause a cyclical dependency since monath-eth-testutil will need to import monad-eth-txpool

monad-rpc/src/main.rs
```diff
@@ -531,7 +531,7 @@ impl MonadRpcResources {
         max_response_size: u32,
         allow_unprotected_txs: bool,
         rate_limiter: Arc<Semaphore>,
-        tx_pool: Arc<vpool::VirtualPool>,
+        tx_pool: Arc<tokio::sync::Mutex<vpool::VirtualPool>>,
```
eerkaijun: nit: there are a few places where Mutex is used, can `use tokio::sync::Mutex` at the beginning of the file instead?

monad-rpc/src/vpool.rs
```diff
@@ -567,36 +845,43 @@ mod tests {
...
     }
 
     // Test behavior with fee replacement.
     // `tx_pool.add_transaction` should replace a pending transaction with a higher fee.
     #[tokio::test]
     async fn txpool_fee_replace() {
         let (ipc_sender, ipc_receiver) = flume::bounded::<TransactionSigned>(100);
-        let tx_pool = Arc::new(VirtualPool::new(ipc_sender.clone(), 100));
+        let mut tx_pool = VirtualPool::new(ipc_sender.clone(), 100, None, None);
+
```
eerkaijun: nit: for consistency, can we rename `tx_pool` to `vpool`

monad-rpc/src/vpool.rs
```diff
@@ -654,16 +943,19 @@ mod tests {
     #[tokio::test]
     async fn txpool_stress() {
         let (ipc_sender, ipc_receiver) = flume::bounded::<TransactionSigned>(10_000);
-        let tx_pool = Arc::new(VirtualPool::new(ipc_sender.clone(), 20_000));
+        let mut tx_pool = VirtualPool::new(ipc_sender.clone(), 20_000, None, None);
```
eerkaijun: same as above

monad-rpc/src/vpool.rs
```diff
@@ -638,10 +927,10 @@ mod tests {
             let res = ipc_receiver.recv_async().await.unwrap();
             match i {
                 0 => {
-                    assert_eq!(res, base.clone().into_signed());
+                    assert_eq!(res, base.clone());
                 }
                 1 => {
-                    assert_eq!(res, replace.clone().into_signed());
+                    assert_eq!(res, replace.clone());
```
eerkaijun: this branch is unreachable

eerkaijun: feels like it's clearer with
```
let res = ipc_receiver.recv_async().await.unwrap()
assert_eq!(res, base.clone());

let res = ipc_receiver.recv_async().await.unwrap()
assert_eq!(res, replace.clone());
```

rather than using a for loop

monad-rpc/src/vpool.rs
```diff
@@ -779,21 +1101,621 @@ mod tests {
...
+                .unwrap()
+                .0,
+            &0
+        );
+    }
+
+    #[tokio::test]
+    async fn test_nonce_gap_resolution() {
+        let (ipc_sender, _ipc_receiver) = flume::bounded::<TransactionSigned>(10_000);
+        let mut tx_pool = VirtualPool::new(ipc_sender.clone(), 2, None, None);
```
eerkaijun: `vpool`

monad-rpc/src/vpool.rs
```diff
@@ -779,21 +1101,621 @@ mod tests {
...
+                .inner
+                .write()
+                .await
+                .accounts
+                .insert(account.1, 0)
+                .expect("insert accounts into chain cache");
+        }
+    }
+
+    fn make_txpool_state_backend(nonce: u64) -> Arc<std::sync::Mutex<InMemoryStateInner>> {
```
eerkaijun: feels like it would be better to move all the txpool<>vpool integration tests into `monad-eth-txpool/tests`, so that we don't couple the rpc crates with consensus types (since rpc should be more of a standalone crate)

monad-rpc/src/vpool.rs
```diff
@@ -182,71 +237,196 @@ impl SubPool {
...
+            self.map.insert(addr, node.clone());
+            self.push_back_node(node).await;
+        }
+    }
+
+    async fn evict_front(&mut self) -> Option<Address> {
+        let front_arc = self.head.take()?;
+        let addr = front_arc.lock().await.addr;
+        {
+            let mut front = front_arc.lock().await;
```
eerkaijun: instead of locking twice, i would suggest
```
let addr;
let next_arc = {
    let mut front = front_arc.lock().await;
    addr = front.addr;
    front.next.take()
};

if let Some(next_arc) = next_arc {
     .....
```

monad-eth-txpool/src/test_utils.rs
eerkaijun: my thoughts: https://github.com/category-labs/monad-bft/pull/1296#discussion_r1897902091

monad-rpc/src/vpool.rs
```diff
@@ -65,25 +74,29 @@ pub struct VirtualPool {
...
     // Publish transactions to validator
     publisher: flume::Sender<TransactionSigned>,
+    metrics: Option<Arc<VirtualPoolMetrics>>,
 }
 
 struct SubPool {
     // Mapping of sender to a transaction tree ordered by nonce
     pool: HashMap<Address, TreeIndex<u64, TransactionSignedEcRecovered>>,
-    evict: RwLock<VecDeque<Address>>,
+    evict: EvictionPolicy,
```
eerkaijun: also would prefer if we can rename `evict` to something else - both the function to remove transactions and the eviction policy is named as `evict` now, which can be a bit confusing

philipglazman:
> just to confirm my understanding, the eviction policy evicts address (and their corresponding transactions) that is least recently active?

The eviction policy will remove the least recently active address's transaction by max nonce. This is similar to txpool's eviction policy.

monad-rpc/src/vpool.rs
philipglazman: This approach makes sense. The two downside I see is that the test code is farther from the source and monad-rpc isn't a structured like a library at this point. We can additionally make vpool its own crate (in a future refactor)

monad-eth-txpool/src/test_utils.rs
philipglazman: Moved vpool into a separate crate and moved integration tests to monad-eth-txpool. This will cleanup the dependencies.

PR #1298, opened 2024-12-05
RPC conformance

eerkaijun:
Adding more systematic test cases in flexnet for RPC, covering each endpoint. This PR fixes some issues uncovered

Related:
https://github.com/monad-crypto/monad-integration/pull/130
https://github.com/monad-crypto/monad-integration/pull/132
https://github.com/monad-crypto/monad-integration/pull/134

### DIFF
--- a/monad-rpc/src/block_handlers.rs
+++ b/monad-rpc/src/block_handlers.rs
@@ -10,7 +10,9 @@ use serde::{Deserialize, Serialize};
 use tracing::trace;
 
 use crate::{
-    eth_json_types::{BlockTags, EthHash, MonadBlock, MonadTransactionReceipt, Quantity},
+    eth_json_types::{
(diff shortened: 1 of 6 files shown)


ariqchowdhury:
Can you describe what issues were fixed here?

eerkaijun:
It's for RPC conformance:
1. `eth_getBlockReceipts` should be able to take in block hash in the query as well
2. `eth_feeHistory`  input `reward_percentiles` should be an optional field. (Other validations that we missed out are sorted percentiles and out of range percentiles) 
3. `eth_getBlockTransactionCountByNumber` should return null instead of 0 if the block does not exist
4. `debug_traceTransaction` should return null instead of error if transaction does not exist
5. `debug_traceBlockByNumber` should return error instead of empty array if block does not exist
6. `eth_call` and `eth_estimateGas` should still work even if gas price is set to zero

PR #1299, opened 2024-12-05
WIP: Refactor metrics to separate crate

andr-dev:


monad-metrics/src/macros.rs
```diff
@@ -0,0 +1,261 @@
...
+                    )*
+                )*
+
+                #[derive(Clone, Debug, Default)]
+                pub struct StaticMetricsPolicy;
+
+                impl crate::MetricsPolicy for StaticMetricsPolicy {
+                    type Counter = crate::StaticCounter;
+
+                    fn build(self) -> crate::Metrics<Self>
```
mikety: `build()` is  the only way to access static metrics when there is no reference to the monad_state (e.g. monad_triedb_state get_account_statuses etc). As its frequently used  I suggest to implement `get_metric_component` method to return counters of an individual component.

monad-metrics/src/counter.rs
```diff
@@ -0,0 +1,85 @@
+use std::sync::{
+    atomic::{AtomicU64, Ordering},
+    Arc,
+};
+
+pub trait Counter {
```
mikety: It needs unit tests for Counter implementations.
- Multithreaded add and sub and read tests
- cover the overflow cases to confirm it does not panic
- cover saturating sub cases make sure its expected

monad-metrics/src/counter.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+impl StaticCounter {
+    pub fn from_static(value: &'static AtomicU64) -> Self {
+        Self { value }
+    }
+}
+
+impl Counter for StaticCounter {
+    fn add(&self, val: u64) {
```
mikety: Its possible to have dependent counters that are used to compute an aggregate later in the - e.g. cache hit ratio in grafana uses both `cache_hits` and `cache_misses`. Wrapping only one counter on the overflow over will produce the incorrect result. It may be worth to consider resetting all connected counters instead. This is fine to implement later if at all.

monad-metrics/src/counter.rs
```diff
@@ -0,0 +1,85 @@
...
+        self.value.fetch_add(val, Ordering::Relaxed);
+    }
+
+    fn sub_saturating(&self, val: u64) {
+        self.value
+            .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |value| {
+                Some(value.saturating_sub(val))
+            })
+            .expect("atomics never panic");
```
mikety: nit: I think more accurate is that u64::saturating_sub cant overflow.

monad-metrics/src/lib.rs
```diff
@@ -149,5 +107,33 @@ metrics!(
             peer_payload_request_successful,
             peer_payload_request_failed
         ]
-    )
+    ),
 );
+
+#[derive(Clone, Debug, Default)]
+pub struct MockMetricsPolicy;
```
mikety, line 23: nit: this policy is fully functional and thread safe, so how about `ThreadSafeMetricsPolicy`?

monad-metrics/src/macros.rs
```diff
@@ -0,0 +1,261 @@
+macro_rules! count {
```
mikety: metrics need unit tests. 
verify snapshot() returned values matches the expected after `inc()` / `add()`, `sub_saturations()`
verify multiple calls for `build` for Static and Local policies match the expectations
can use macro to generate  uint tests for metrics.

monad-metrics/src/counter.rs
andr-dev: Ah sorry, fetch_update could panic if the lambda returns a None, which it can't since we explicitly construct a Some(...). This does indicate that my expect message is bad, I'll update that to make this clearer.

PR #1301, opened 2024-12-09
Fix Master

philipglazman:


eclark0426:
Related PR: https://github.com/monad-crypto/monad-bft/pull/1293

My PR pins the nightly version, which also gets rid of the current problems on master

eclark0426:
Closing this and integrating it into https://github.com/monad-crypto/monad-bft/pull/1293 instead

PR #1304, opened 2024-12-09
Mixed MTU Values

eclark0426:
This allows nodes with different values for MONAD_MTU to communicate with each other, changing MTUs as needed during message transmission. This allows for a rolling update of MONAD_MTU values without causing issues in the communication. Additionally, the value has now been moved to `node.toml` to allow for Flexnet to run nodes with differing values without the need to recompile the binary with different values.

Corresponding monad-integration PR: https://github.com/monad-crypto/monad-integration/pull/136

### DIFF
--- a/monad-dataplane/src/event_loop.rs
+++ b/monad-dataplane/src/event_loop.rs
@@ -62,6 +62,7 @@ const TCP_MESSAGE_TIMEOUT: Duration = Duration::from_secs(600);
 pub struct BroadcastMsg {
     pub targets: Vec<SocketAddr>,
     pub payload: Bytes,
+    pub stride: usize,
 }
 
 /// Send a list of unicast payloads
@@ -199,7 +200,7 @@ fn make_producer_consumer<T>(size: usize) -> (WakeableProducer<T>, WakeableConsu
 
 impl Dataplane {
     /// 1_000 = 1 Gbps, 10_000 = 10 Gbps
-    pub fn new(local_addr: &SocketAddr, up_bandwidth_mbps: u64) -> Self {
+    pub fn new(local_addr: &SocketAddr, up_bandwidth_mbps: u64, default_mtu: u16) -> Self {
         let (udp_ing_producer, udp_ing_consumer) = make_producer_consumer(12_800);
         let (tcp_ing_producer, tcp_ing_consumer) = make_producer_consumer(32);
 
@@ -217,6 +218,13 @@ impl Dataplane {
         let tcp_outgoing_connection_timers: [TimerFd; TCP_OUTGOING_MAX_CONNECTIONS] =
             core::array::from_fn(|_| TimerFd::new().unwrap());
 
+        // 64 is Linux kernel limit on max number of segments
+        assert!(
+            (crate::network::max_iovec_len(default_mtu as usize)
+                / crate::network::gso_size(default_mtu as usize))
+                <= 64
+        );
+
         DataplaneEventLoop {
             efd,
             epoll: EpollFd::<NUM_EPOLL_EVENTS>::new().unwrap(),
@@ -227,7 +235,7 @@ impl Dataplane {
             egress_bcast_receiver: egr_bcast_recv,
             egress_ucast_receiver: egr_ucast_recv,
             egress_tcp_receiver: egr_tcp_recv,
-            udp_socket: NetworkSocket::new(local_addr, up_bandwidth_mbps),
+            udp_socket: NetworkSocket::new(local_addr, up_bandwidth_mbps, default_mtu),
 
             tcp_listening_socket,
             tcp_incoming_connections: [const { None }; TCP_INCOMING_MAX_CONNECTIONS],
@@ -325,7 +333,10 @@ impl DataplaneEventLoop {
                 .expect("set nonblocking");
 
             self.epoll
-                .register(UDP_RX_EVENT, RawFd(self.udp_socket.socket.as_raw_fd()))
+                .register(
+                    UDP_RX_EVENT,
+                    RawFd(self.udp_socket.default_socket.as_raw_fd()),
+                )
                 .expect("epoll register of udp must succeed");
             self.epoll
                 .register(UDP_TX_EVENT, self.efd.get_fd())
@@ -489,14 +500,14 @@ impl DataplaneEventLoop {
 
     fn handle_broadcast(&mut self) {
         loop {
-            let (to, payload) = match self.egress_bcast_receiver.pop() {
+            let (to, payload, stride) = match self.egress_bcast_receiver.pop() {
                 Err(PopError::Empty) => {
                     break;
                 }
-                Ok(msg) => (msg.targets, msg.payload),
+                Ok(msg) => (msg.targets, msg.payload, msg.stride),
             };
 
-            self.udp_socket.broadcast_buffer(to, payload);
+            self.udp_socket.broadcast_buffer(to, payload, stride);
         }
     }
 
--- a/monad-dataplane/src/network.rs
+++ b/monad-dataplane/src/network.rs
@@ -12,10 +12,9 @@ pub const MAX_UDP_PKT: usize = 65535;
 const MAX_IPV4_HDR: usize = 20;
 const MAX_UDP_HDR: usize = 8;
 
-// TODO: When running in docker with vpnkit, we seem to occasionally get ICMP frag-neededs
-// for 20 bytes less than the expected MTU -- investigate what's causing this.
-const MONAD_MTU: usize = 1480;
-pub const MONAD_GSO_SIZE: usize = MONAD_MTU - MAX_IPV4_HDR - MAX_UDP_HDR;
+pub const fn gso_size(mtu: usize) -> usize {
+    mtu - MAX_IPV4_HDR - MAX_UDP_HDR
+}
 
 const BUF_SIZE: usize = MAX_UDP_PKT;
 const NUM_RX_MSGHDR: usize = 128;
@@ -25,14 +24,13 @@ const CMSG_LEN: usize = 88;
 // message length is limited by the max limit of the underlying protocol
 //FIXME: This is expected size MAX_UDP_PKT - MAX_IPV4_HDR - MAX_UDP_HDR, but the actual measured
 //number where the packet isn't being fragmented is 65493. investigate
-const MAX_IOVEC_LEN: usize = 65493 / MONAD_GSO_SIZE * MONAD_GSO_SIZE;
+pub const fn max_iovec_len(mtu: usize) -> usize {
+    let gso = gso_size(mtu);
+    65493 / gso * gso
+}
 
 const LINUX_SENDMMSG_VLEN_MAX: usize = 1024;
 const NUM_IOVECS: usize = 1024;
-
-// 64 is linux kernel limit on max number of segments
-#[allow(clippy::assertions_on_constants)]
-const _: () = assert!((MAX_IOVEC_LEN / MONAD_GSO_SIZE) <= 64);
 
 // num msgs in sendmmsg is limited to 1024 in the kernel
 #[allow(clippy::assertions_on_constants)]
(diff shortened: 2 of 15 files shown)


buytenh:
I'll take a look at this.

monad-dataplane/src/event_loop.rs
```diff
@@ -199,7 +200,7 @@ fn make_producer_consumer<T>(size: usize) -> (WakeableProducer<T>, WakeableConsu
 
 impl Dataplane {
     /// 1_000 = 1 Gbps, 10_000 = 10 Gbps
-    pub fn new(local_addr: &str, up_bandwidth_mbps: u64) -> Self {
+    pub fn new(local_addr: &str, up_bandwidth_mbps: u64, mtu: u16) -> Self {
```
buytenh: Rename this to default_mtu or preferred_mtu or something like that?

monad-dataplane/src/network.rs
```diff
@@ -44,8 +49,83 @@ pub struct AlignedCmsg(pub MaybeUninit<[u8; CMSG_LEN]>);
...
+        }
+    }
+
+    fn get_tx_socket(&mut self, mtu: i32) -> &std::net::UdpSocket {
+        if !self.sockets.contains_key(&mtu) {
+            let socket = std::net::UdpSocket::bind("0.0.0.0:0").unwrap();
+            Self::set_sock_opts(&socket, mtu);
+            self.sockets.insert(mtu, socket);
+        }
+        self.sockets.get(&mtu).unwrap()
```
buytenh: Can use `HashMap::entry(key).or_insert_with()` here?

monad-dataplane/src/network.rs
```diff
@@ -44,8 +49,83 @@ pub struct AlignedCmsg(pub MaybeUninit<[u8; CMSG_LEN]>);
...
+impl TxSockets {
+    pub fn new() -> Self {
+        Self {
+            sockets: std::collections::HashMap::new(),
+        }
+    }
+
+    fn get_tx_socket(&mut self, mtu: i32) -> &std::net::UdpSocket {
+        if !self.sockets.contains_key(&mtu) {
+            let socket = std::net::UdpSocket::bind("0.0.0.0:0").unwrap();
```
buytenh: Perhaps this should use the IpAddr part of the configured bind address, in case monad-node was bound to a specific interface?

monad-dataplane/src/network.rs
```diff
@@ -311,16 +352,16 @@ impl<'a> NetworkSocket<'a> {
...
     //
     // TODO use more descriptive enum for return value rather than Option
-    pub fn broadcast_buffer(&mut self, to: Vec<SocketAddr>, data: Bytes) -> Option<()> {
+    pub fn broadcast_buffer(&mut self, to: Vec<SocketAddr>, data: Bytes, stride: usize) -> Option<()> {
         if to.is_empty() {
             return None;
         }
 
-        let num_chunks = data.len().div_ceil(MAX_IOVEC_LEN);
+        let num_chunks = data.len().div_ceil(stride);
```
buytenh: This doesn't seem right -- doesn't this cause `sendmmsg()` to only ever be called with a single packet worth of data?

monad-dataplane/src/network.rs
```diff
@@ -44,8 +49,83 @@ pub struct AlignedCmsg(pub MaybeUninit<[u8; CMSG_LEN]>);
...
+}
+
+impl TxSockets {
+    pub fn new() -> Self {
+        Self {
+            sockets: std::collections::HashMap::new(),
+        }
+    }
+
+    fn get_tx_socket(&mut self, mtu: i32) -> &std::net::UdpSocket {
```
buytenh, line 58: What happens if I (from a remote host) transmit data to one of these bound sockets?

eclark0426: I just send a flood of UDP packets to one of them, and it didn't have any immediately obvious negative effects. It looks like there's a way to set the receive buffer size, so I'll see if I can set it to zero, to make it more explicitly transmit-only and always drop incoming packets

monad-dataplane/src/network.rs
eclark0426: Updated to use `max_iovec_len()`

eclark0426:
> If I send a single (very) large RaptorCast symbol from a remote host, as a fragmented UDP packet with many fragments, to be rebroadcast by this node, won't it cause a panic in the rebroadcast path, as with UDP_SEGMENT we cannot transmit packets larger than the interface MTU?
> 
> Is there some sort of protection against me sending a very tiny (but valid) RaptorCast symbol into the rx path -- that is, will it get dropped somewhere, or wil raptorcast/dataplane now rebroadcast packets of any size?
> 
> If I can send a fragmented UDP packet into raptorcast from a remote host, does that mean I can send packets of all sizes 1..65000-something bytes and thereby cause 65000-something sockets to be created, which could exhaust the number of available file descriptors? (Actually, we would probably run out of UDP port numbers before that, but either way. it would cause a panic, I think?)

Need to test some of these things still, but I think your other comments have all been updated

buytenh:
There has been some ongoing discussion about this in DM with Ellis, but in particular, there may also be a need to manually segment a packet bundle on egress, for example if the stride size for that bundle is larger than what the outgoing interface's MTU can handle, which can happen if we receive a fragmented Raptor symbol from another validator that we have to rebroadcast.  In this case, I don't think we can use `UDP_SEGMENT`, as I don't think `UDP_SEGMENT` lets you segment a packet bundle to a size where each individual segment then also has to be IP-fragmented.

eclark0426:
After the discussion this morning, I think that this PR should be reverted to 5a3bb, at which point it will address https://github.com/category-labs/category-internal/issues/1051. The manual segmentation can be separated out and revisited later if we decide it's worth the effort (https://github.com/category-labs/category-internal/issues/1052)

eclark0426:
This has been rebased onto master, and I believe all of the comments should be addressed. I verified that nodes are still able to communicate with each other with different values using this Flexnet topology:
```

{
    "regions": [
        {
            "name": "us",
            "intraregion_latency": {
                "ms": 10
            },
            "nodes": [
                {
                    "name": "node0",
                    "up_Mbps": 100,
                    "down_Mbps": 100,
                    "execution": true,
                    "rpc": true
                },
                {
                    "name": "node1",
                    "up_Mbps": 100,
                    "down_Mbps": 100,
                    "execution": true,
                    "rpc": true,
                    "mtu": 1240
                },
                {
                    "name": "node2",
                    "up_Mbps": 100,
                    "down_Mbps": 100,
                    "execution": true,
                    "rpc": true,
                    "mtu": 1280
                },
                {
                    "name": "node3",
                    "up_Mbps": 100,
                    "down_Mbps": 100,
                    "execution": true,
                    "rpc": true,
                    "mtu": 1320
                }
            ]
        }
    ],
    "latencies": []
}
```
That topology should be analogous to the scenario we were talking about Monday, when we want to switch from MONAD_GSO_SIZE 1480 to 1500 at a specific epoch n:
- Initially
  - Everyone on MONAD_GSO_SIZE 1480 and interface MTU 1480
  - Analogous to the "typical" Flexnet tests we always run
- During transition
  - Everyone on MONAD_GSO_SIZE 1480, mixed interface MTUs of 1480 and 1500
  - Analogous to the nodes in this topology that have interface MTU 1480 but lower GSO sizes
- After epoch n passes
  - Everyone swaps over to interface size 1500 and MONAD_GSO_SIZE 1500
  - Analogous to the "typical" Flexnet tests we always run, just with a different MTU/GSO_SIZE

I similarly confirmed under Proxmox that having two nodes with an MTU of 1500 and a node with MTU 1480, all with MONAD_GSO_SIZE set to 1480, the nodes are able to communicate and send proposals to each other.

Seems like, given the discussions from Monday, the issue of manual segmentation and reassembly is unneeded because it will only apply in niche cases. That has been turned into a [separate issue](https://github.com/category-labs/category-internal/issues/1052).

monad-dataplane/src/network.rs
```diff
@@ -44,16 +42,113 @@ pub struct AlignedCmsg(pub MaybeUninit<[u8; CMSG_LEN]>);
...
+            );
+        }
+    }
+}
+
+impl Default for TxSockets {
+    fn default() -> Self {
+        Self::new("0.0.0.0:0".parse().unwrap())
+    }
+}
```
buytenh: I can't see a case where using 0.0.0.0:0 by default is ever a good idea.  Do we need a Default impl for TxSockets, or can we remove this?

monad-dataplane/src/network.rs
```diff
@@ -346,7 +405,9 @@ impl NetworkSocket<'_> {
                 self.send_ctrl.msgs[k].msg_hdr.msg_name =
                     self.send_ctrl.name[k].as_ptr() as *const _ as *mut _;
                 self.send_ctrl.msgs[k].msg_hdr.msg_namelen =
-                    unsafe { self.send_ctrl.name[i].assume_init_ref().len() };
+                    unsafe { self.send_ctrl.name[k].assume_init_ref().len() };
```
buytenh, line 402: (Shouldn't this be `[j]`?)

eclark0426: No, I don't think so - the name length should be the length of the address we're sending to, which we just set in `send_ctrl.name[k]` on line 398. Ultimately since we're currently only using IPv4 I don't think it makes a difference (I tested with k and j, and both work fine), but I think k is the correct index to use

monad-dataplane/src/network.rs
eclark0426: Removed

eclark0426:
> it can allow that attacker to then spam data to those sockets (as UDP sockets have both a transmit and a receive half), which will eat up (locked kernel) memory forever

Receive buffer size is [set to 0](https://github.com/category-labs/monad-bft/pull/1304/files#diff-c48fee3e2fa036b1ab1f912e7723ddc4f87b05a8d62a235a25856d5d57291c5aR115-R131) for transmit sockets, so there won't be any memory used up by not taking any data out of the sockets.

>  it allows a remote. attacker to cause us to create many different sockets, which eats up file descriptors, as well as local port numbers

At one point I had added an LRU to help mitigate that - must have been in one of the later commits that got reverted. I can pull that back in if you think it would be helpful?

monad-dataplane/src/network.rs
buytenh: @eclark0426 pointed out that the receive buffer size is set to the minimum value now, which should take care of this.

monad-node/src/config/network.rs
```diff
@@ -10,4 +10,11 @@ pub struct NodeNetworkConfig {
...
     pub max_rtt_ms: u64,
     pub max_mbps: u16,
+
+    #[serde(default = "default_mtu")]
+    pub mtu: u16,
+}
+
+fn default_mtu() -> u16 {
+    1480
 }
```
buytenh, line 22: Can you add this comment (or some variation of it) back here, so that it's a bit clearer why we don't just use 1500?

```
// TODO: When running in docker with vpnkit, we seem to occasionally get ICMP frag-neededs
// for 20 bytes less than the expected MTU -- investigate what's causing this.
```

eclark0426: Re-added

monad-raptorcast/src/lib.rs
```diff
@@ -406,25 +412,34 @@ where
...
                 let dataplane = RefCell::new(&mut this.dataplane);
+                let local_gso_size = gso_size(this.mtu as usize);
+                let stride = message.stride;
                 this.udp_state.handle_message(
                     &mut this.epoch_validators,
-                    |targets, payload| {
-                        // this is the callback used for rebroadcasting
-                        let target_addrs = targets
+                    |targets, payload, bcast_stride| {
+                        let mut complete_message = BytesMut::with_capacity(payload.len());
```
omegablitz: dumb question: how is complete_message different from payload?

eclark0426: Not a dumb question - doesn't seem there's any difference at all, and everything seems to work the same by using payload directly. Likely in one of the now-squashed commits there was some intermediate processing that turned out unneeded

eclark0426: Updated it now to just use payload directly

PR #1305, opened 2024-12-09
Add startup time gap in devnet integration script

mikety:
The change fixes the intermittent Jenkins monad-bft/integration-test
The root cause is the `monad_execution` starts while `monad_mpt` is not fully initialized


michael-yxchen:
docker compose dependency is a more robust way of enforcing this
e.g. the following is configured on devnet hosts
```
bft:
  depends_on:
    mpt:
      condition: service_completed_successfully
      required: false
```

mikety:
Thanks @michael-yxchen , done.

michael-yxchen:
PR changes looks good. Will approve for merging after eth header stuff is settled

mikety:
Fixed by another PR

PR #1306, opened 2024-12-10
BlockSync peers

michael-yxchen:
- Adds functionality to update blocksync peers
- Blocksync peers default to validators in bootstrap peers (known ip)

### Test plan
1. added unit tests for `BlockSyncWrapper::handle_update_peers`
2. debug-node functionalities were manually verified
3. created issue for test coverage on debug-node toolchain: https://github.com/monad-crypto/monad-internal/issues/959

### DIFF
--- a/monad-blocksync/src/blocksync.rs
+++ b/monad-blocksync/src/blocksync.rs
@@ -115,6 +115,8 @@ where
 
     self_request_mode: BlockSyncSelfRequester,
 
+    override_peers: Vec<NodeId<CertificateSignaturePubKey<ST>>>,
+
     rng: ChaCha8Rng,
 }
 
@@ -134,13 +136,13 @@ where
     payload_cache: HashMap<ConsensusBlockBodyId, ConsensusBlockBody<EPT>>,
 }
 
-impl<ST, SCT, EPT> Default for BlockSync<ST, SCT, EPT>
+impl<ST, SCT, EPT> BlockSync<ST, SCT, EPT>
 where
     ST: CertificateSignatureRecoverable,
     SCT: SignatureCollection<NodeIdPubKey = CertificateSignaturePubKey<ST>>,
     EPT: ExecutionProtocol,
 {
-    fn default() -> Self {
+    pub fn new(override_peers: Vec<NodeId<CertificateSignaturePubKey<ST>>>) -> Self {
         Self {
             headers_requests: Default::default(),
             payload_requests: Default::default(),
@@ -149,17 +151,11 @@ where
             self_payload_requests_in_flight: 0,
             self_completed_headers_requests: Default::default(),
             self_request_mode: BlockSyncSelfRequester::StateSync,
+            override_peers,
             rng: ChaCha8Rng::seed_from_u64(123456),
         }
     }
-}
 
-impl<ST, SCT, EPT> BlockSync<ST, SCT, EPT>
-where
-    ST: CertificateSignatureRecoverable,
-    SCT: SignatureCollection<NodeIdPubKey = CertificateSignaturePubKey<ST>>,
-    EPT: ExecutionProtocol,
-{
     fn clear_self_requests(&mut self) {
         self.self_headers_requests.clear();
         self.self_payload_requests.clear();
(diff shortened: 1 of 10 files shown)


ariqchowdhury:
resolves category-labs/category-internal#1064

michael-yxchen:
Need to add another blocksync_peers section in node.toml. Full nodes need `[[bootstrap.peers]]` populated to forward txns

monad-blocksync/src/blocksync.rs
```diff
@@ -321,22 +333,59 @@ where
         cmds
     }
 
+    #[must_use]
+    pub fn handle_update_peers(
```
ariqchowdhury: impl reads more like `handle_set_peers` as its just replacing the old list with the new list. 
do we actually want to replace the list or update with the union of the lists.

monad-blocksync/src/blocksync.rs
```diff
@@ -321,22 +333,59 @@ where
...
+        &mut self,
+        new_peers: Vec<NodeId<CertificateSignaturePubKey<ST>>>,
+    ) -> Vec<BlockSyncCommand<SCT>> {
+        // Blocksync peers cannot be empty or `Self::pick_peer` panics
+        if new_peers.is_empty()
+            || (new_peers.len() == 1 && new_peers.first().unwrap() == self.nodeid)
+        {
+            return vec![BlockSyncCommand::UpdatePeersResult(
+                UpdatePeersResult::Failed("Cannot empty out blocksync peers".to_owned()),
```
ariqchowdhury: doesn't seem like there are many fail cases. Should we just have a specific enum for the fail instead of generic fail with string

monad-blocksync/src/blocksync.rs
```diff
@@ -321,22 +333,59 @@ where
...
+                    // non-validator weight is arbitrarily set as 1
+                    // FIXME: revisit after staking
+                    Some((p, validator_stake.get(p).unwrap_or(&Stake(1))))
+                }
+            })
             .collect_vec();
-        assert!(!members.is_empty(), "no nodes to blocksync from");
-        *members
+        assert!(!weighted_peers.is_empty(), "no nodes to blocksync from");
+        *weighted_peers
```
ariqchowdhury: for blocksync peers, do we want the choice of peers to still be stake weighted? I think you could argue that setting a blocksync peer list of 2 nodes should just sync from them equally because you are listing 2 trusted nodes that you want to blocksync from. 
The stake-weighted choosing makes more sense to me when you are just choosing from the entire validator set because your trust in those nodes is only as much as they are staked.

monad-blocksync/src/blocksync.rs
ariqchowdhury: to elaborate, if this function is handling updates via the cli, its more user friendly if they can just add/remove vs having to re-enter the entire list. Like, if they already have a list of 100 nodes and they want to modify 1

michael-yxchen: removed this method from PR. It'll make more sense once we implement dynamic reload config

monad-blocksync/src/blocksync.rs
michael-yxchen: yeah makes sense to uniformly select

monad-blocksync/src/blocksync.rs
```diff
@@ -109,8 +110,11 @@ struct SelfCompletedHeader<SCT: SignatureCollection> {
     payload_cache: HashMap<PayloadId, Payload>,
 }
 
-impl<ST: CertificateSignatureRecoverable, SCT: SignatureCollection> Default for BlockSync<ST, SCT> {
-    fn default() -> Self {
+impl<ST: CertificateSignatureRecoverable, SCT: SignatureCollection> BlockSync<ST, SCT>
```
omegablitz: super nit: SignatureCollection bound defined twice
maybe move all of them into the `where` clause

michael-yxchen:
Duplicating known peers in both router and blocksync is messy. I implemented it as, when `override_peers` is empty, we select from current validators. Validators discover all other validators. Full nodes should set their known peers as override

PR #1307, opened 2024-12-10
Add trace_block rpc method

philipglazman:
Adds `trace_block` rpc method. `traceAddress` response is not implemented yet. I need to build a bigger callgraph in flexnet contracts to test against anvil.

### DIFF
--- a/monad-rpc/src/trace_handlers.rs
+++ b/monad-rpc/src/trace_handlers.rs
@@ -1,4 +1,4 @@
-use std::{cell::RefCell, rc::Rc};
+use std::{cell::RefCell, rc::Rc, str::FromStr};
 
 use alloy_primitives::{
     aliases::{U256, U64, U8},
@@ -7,6 +7,10 @@ use alloy_primitives::{
 use alloy_rlp::Decodable;
 use monad_rpc_docs::rpc;
 use monad_triedb_utils::triedb_env::Triedb;
+use reth_rpc_types::trace::parity::{
+    Action, CallAction, CallOutput, CallType, CreateAction, CreateOutput,
+    LocalizedTransactionTrace, TraceOutput,
+};
 use serde::{Deserialize, Serialize};
 use tracing::{debug, error, trace};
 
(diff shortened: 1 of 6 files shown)


monad-rpc/src/trace_handlers.rs
```diff
@@ -230,6 +220,171 @@ pub async fn monad_debug_traceBlockByHash<T: Triedb>(
...
+    #[serde(rename_all = "camelCase")]
+    Call {
+        gas_used: Quantity,
+        output: UnformattedData,
+    },
+}
+
+#[derive(Serialize, schemars::JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct MonadTraceBlockResult {
```
eerkaijun: not essential but is it possible to reuse reth types here for the return type? seem to be easier to catch discrepancies by reusing the same type. I saw that they use `LocalizedTransactionTrace`

monad-rpc/src/trace_handlers.rs
```diff
@@ -174,32 +173,24 @@ enum CallKind {
     StaticCall,
 }
 
-#[derive(Deserialize, Debug, schemars::JsonSchema)]
-pub struct MonadDebugTraceBlockByHashParams {
-    block_hash: EthHash,
-    #[serde(default)]
-    tracer: TracerObject,
+#[derive(Debug, Clone, schemars::JsonSchema)]
+pub struct ParityCallKind(CallKind);
```
eerkaijun: is it required to wrap this in ParityCallKind?

philipglazman: It is since the actions `CALL`,`CREATE` are serialized as lowercase for `trace_transaction` and `trace_block` but are uppercase for the other trace methods

monad-rpc/src/trace_handlers.rs
philipglazman: Done!

philipglazman:
replaced by https://github.com/category-labs/monad-bft/pull/1716

PR #1309, opened 2024-12-11
Upgrade and deduplicate alloy dependencies

omegablitz:


omegablitz:
Checked with `cargo tree -d | grep -v "*"`

PR #1310, opened 2024-12-11
[monad-event-ring] add a Rust version of the event client library

kjcamann:
monad execution is gaining an event notification system. It can communicate with external processes via a shared memory communication mechanism, described in the documentation in the `monad` repository. This will initially be used by the RPC server to support WebSocket realtime events.

This commit adds support for Rust clients. It reuses part of the C/C++ client library via FFI, but the performance sensitive part (event_reader.rs) is a native Rust reimplementation.

An example program, eventwatch, is used to demonstrate how the client library works.


monad-exec-event/src/event_client.rs
```diff
@@ -0,0 +1,233 @@
...
+        })
+    }
+}
+
+impl Clone for ImportedEventRing<'_, '_> {
+    fn clone(&self) -> Self {
+        let r = unsafe {
+            monad_event_imported_ring_acquire(self.imported_ring)
+        };
+        assert!(!r.is_null());
```
kjcamann: This is a good question. Please grill me on places where I `assert` or otherwise do things that can crash the program (`expect`, `unwrap`, etc.)

I added this comment to the new branch:

```
// Assert that `r` is not null. null can only be returned in the
// underlying C API if the caller tries to acquire (i.e., increase
// the reference count) on a ring that was already garbage collected.
// We do not expect that to be possible in Rust, since the
// ImportedEventRing acts as an Arc-like type, and properly balances
// acquire/release calls using its Clone and Drop trait implementations.
```

monad-exec-event/src/event_client.rs
```diff
@@ -0,0 +1,255 @@
...
+}
+
+unsafe impl Send for EventProc<'_> {}
+unsafe impl Sync for EventProc<'_> {}
+
+/// Smart pointer to the underlying C imported ring object; it is itself
+/// an Arc-like object. It introduces a lifetime for the shared memory
+/// segments it imports, which is potentially shorter than 'meta
+pub struct ImportedEventRing<'meta, 'ring> {
+    imported_ring: *mut monad_event_imported_ring,
```
philipglazman: I think `imported_ring` should be wrapped in an [AtomicPtr](https://doc.rust-lang.org/stable/std/sync/atomic/struct.AtomicPtr.html) like `std::sync::atomic::AtomicPtr<monad_event_imported_ring>` so that ImportEventRing is [send](https://doc.rust-lang.org/stable/std/marker/trait.Send.html) safe. Otherwise, a client can't spawn a task in tokio to poll `FinalizedBlockUpdateBuilder`.

kjcamann:
This rewritten by Andre a long time ago, closing

PR #1314, opened 2024-12-13
Hoist triedb-driver out of consensus into execution.

ned14:
As we wished to add the testing of exactly what consensus uses to execution's unit tests, this PR along with its corresponding PR for execution https://github.com/category-labs/monad/pull/1045 hoists the triedb-driver out of consensus into execution. There the execution cmake generates Rust FFI bindings which are then consumed by consensus. As within cmake we prefix targets with `monad`, the driver is renamed to
`monad-triedb-driver`.

I fixed a few API problems:

1. Triedb takes a list of filesystem paths, not a single filesystem path. It never at any point took a single filesystem path. `monad-triedb-driver` now takes a `Vec<PathBuf>`. I tried to get `&[Path]` which ought to be the Rust equivalent to C++'s `span<path_view>` to work, but it defeated me and I admit I gave up. I propagted this change through the codebase up to the point where the path is retrieved from config or CLI. I didn't change that bit as it would likely break things. In any case, if in the future multi-device triedb is needed, it's now much easier to implement.

2. The old driver didn't really do much error handling. The new driver returns a `MonadCResult<T>` which is a proper Rust Result which wraps the original C or C++ result losing no fidelity. Much of the code which used the old driver seems to assume that failure is benign e.g. equals not finding a key etc rather than anything catastrophic. In order to minimise disruption to existing code, I have retained that interpretation, even though it is naively optimistic at best and plain wrong otherwise. Somebody better at Rust probably ought to go implement proper error handling later (note that `MonadCResult<T>` can be fuzzy compared for equivalence e.g. "item not found"). For now, nothing should have changed in logic, we do throw away causes of error but we were doing so anyway until now.

3. The old driver took `int` for lengths. C++ is the same as Rust, we use `usize` (`size_t`).

4. cmake can do `rustc` and `bindgen` just fine, but it can't do `cargo` or anything which involves a crate. Therefore, I had to change the API to eliminate any external crates. The `monad-triedb` Rust crate's `TriedbHandle` now 'inherits' from the execution `TriedbHandle` and it supplies the crate derived items. This introduces some dynamic dispatch and an unfortunate double-`Box<>`, but I personally find Rust does not shine in this area in general. It won't let you do indirection collapse and variably sized erased types like C++ happily does (or, if it does, nowhere on Google says how).

I fixed some toolchain problems:

1. Rust's cmake crate doesn't use cmake correctly at all :( so I have worked around the problem using a cmake toolchain file to correctly set the cmake configuration (setting cmake flags on the command line is just plain wrong). This toolchain file sets a Haswell arch, turns on line number only debug info for release builds, sets C and C++ standards to 23, sets position independent code, and tells execution's cmake to produce shared libraries.

(I would urge consensus to stop using shared libraries where possible. They are a security risk compared to static libraries. Also, execution does zero testing of shared libraries, we are 100% static library based with no shared libraries. I'd recommend the same for consensus binaries)

This means you no longer need to set the environment variables `ASMFLAGS`, `CFLAGS` or `CXXFLAGS`.

The next obvious step is to hoist the statesync driver out of consensus into execution, and apply similar API and error propagation improvements. I think let's review and merge this first before considering that.


0cjs:
> I would urge consensus to stop using shared libraries where possible. They are a security risk compared to static libraries.

As well as this, avoiding shared libs makes packaging easier, so generally :+1: for that.

PR #1330, opened 2024-12-23
mike/publish to ecr

mhumeSF:
- replaced repeated job definitions with a matrix strategy
- additionally publish to aws ecr

### DIFF
--- a/monad-triedb-cache/src/lib.rs
+++ b/monad-triedb-cache/src/lib.rs
@@ -92,10 +92,10 @@ where
         if cache.len() > self.execution_delay.0 as usize * 2 {
             let (evicted, _) = cache.pop_first().expect("nonempty");
             if evicted == block {
-                let (latest, _) = cache.last_key_value().expect("nonempty");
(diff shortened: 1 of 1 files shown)


mhumeSF:
@omegablitz @ariqchowdhury @michael-yxchen If any one of you could 👍 🙏

mhumeSF:
@omegablitz @ariqchowdhury @michael-yxchen  This ok to merge?

michael-yxchen:
we need to update repo url on all hosts after publishing to ecr. There's a (small) risk that it might break existing deployment workflows

Unless external validators will pull from the same ecr, I'd prefer to merge when we get a break from frequent deployments and testing

mhumeSF:
All this update does is additionally publish to aws ecr – so no requirement to update repo url everywhere. Only in the case where we would want to pull from ECR which I believe was stressnet instances in Lithuania which had some slow pulling behavior.

mhumeSF:
@michael-yxchen @ariqchowdhury We ok to merge this? Again this only additionally publishes to ECR and will not affect any current deployments.

PR #1340, opened 2024-12-23
Get raw traces from RPC

eerkaijun:
@JoeHowarth wanted a way to get the raw traces from RPC for archiver use

I'll add a corresponding test for this in flexnet too


JoeHowarth:
Making rpc an archive block data source turned out to be a dead end, closing this

PR #1343, opened 2024-12-26
Consolidate crates and constants

eerkaijun:
was writing some code and feels like some crates can be cleaned up:
- `monad-eth-tx` and are just transaction types that can be absorbed into `monad-eth-types`
- also define hardcoded constant (e.g. base fee) in `monad-eth-types`, this will make swapping out these hardcoded values much easier: https://github.com/category-labs/category-internal/issues/1003
- `monad-eth-testutil` has too many consensus dependencies like `monad-eth-block-policy` etc, which often causes cyclical dependencies. these test utils should really just be standalone helper function. did a bit of refactor there


eerkaijun:
lots of changes are already done in the release-testnet branch, closing this

PR #1346, opened 2024-12-29
Make RPC compile post eth-header changes

omegablitz:


monad-triedb-utils/src/triedb_env.rs
```diff
@@ -160,11 +161,11 @@ pub trait Triedb {
         &self,
         txn_index: u64,
         block_num: u64,
-    ) -> impl std::future::Future<Output = Result<Option<TransactionSigned>, String>> + Send;
+    ) -> impl std::future::Future<Output = Result<Option<TxEnvelope>, String>> + Send;
```
eerkaijun, line 163: why do we need to switch this to TxEnvelope, is the transaction format stored in triedb changed?

PR #1347, opened 2024-12-29
Proposal size check

eerkaijun:
https://github.com/category-labs/category-internal/issues/796
https://github.com/category-labs/category-internal/issues/1016

i feel like we only need to enforce proposal size check when creating proposal as it's for making sure proposal can be broadcasted in time, and no need to validate that in eth-block-validator, but let me know if people think otherwise

### DIFF
--- a/monad-consensus-types/src/payload.rs
+++ b/monad-consensus-types/src/payload.rs
@@ -46,8 +46,10 @@ impl AsRef<[u8]> for Gas {
     }
 }
 
-pub const BASE_FEE_PER_GAS: u64 = 1_000;
-pub const PROPOSAL_GAS_LIMIT: u64 = 150_000_000;
+pub const BASE_FEE_PER_GAS: u64 = 50_000_000_000;
+pub const PROPOSAL_GAS_LIMIT: u64 = 300_000_000;
+/// Max proposal size in bytes (average transactions ~400 bytes)
+pub const PROPOSAL_SIZE_LIMIT: u64 = 4_000_000;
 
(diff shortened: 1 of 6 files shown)


monad-consensus-types/src/payload.rs
```diff
@@ -48,6 +48,8 @@ impl AsRef<[u8]> for Gas {
 
 pub const BASE_FEE_PER_GAS: u64 = 1_000;
 pub const PROPOSAL_GAS_LIMIT: u64 = 150_000_000;
+/// Max proposal size in bytes (average transactions ~400 bytes)
+pub const PROPOSAL_SIZE_LIMIT: u64 = 5_000_000;
```
eerkaijun: we can switch this value

monad-eth-txpool/src/tracked/mod.rs
```diff
@@ -232,6 +233,14 @@ where
                 return TrackedTxHeapDrainAction::Skip;
             }
 
+            if total_size
+                .checked_add(tx.size().into())
+                .expect("total size does not overflow")
```
eerkaijun: there's a bunch of expect in this file, we should remove them

eerkaijun: removed some, we should review the remaining ones

monad-consensus-types/src/payload.rs
omegablitz: let's set to 4MB, so 4_000_000

eerkaijun:
ok sounds good, added

PR #1348, opened 2024-12-29
Remove seq_num, timestamp and version from VoteInfo

ariqchowdhury:


monad-consensus-types/src/block.rs
```diff
@@ -232,10 +233,29 @@ where
...
         &self,
-        _: &Self::ValidatedBlock,
-        _: Vec<&Self::ValidatedBlock>,
+        block: &Self::ValidatedBlock,
+        extending_blocks: Vec<&Self::ValidatedBlock>,
         _: &InMemoryState,
     ) -> Result<(), BlockPolicyError> {
+        if block.get_seq_num()
+            != extending_blocks
+                .last()
```
omegablitz: At first glance, I don't see how this works for the first block (seq_num=1), because I'd think that extending_blocks would be empty.

I'm confused as to how the mock-swarm tests pass though...

monad-consensus-types/src/block.rs
```diff
@@ -232,10 +233,29 @@ where
...
+                .expect("must be at least 1 extending block")
+                .get_seq_num()
+                + SeqNum(1)
+        {
+            return Err(BlockPolicyError::BlockNotCoherent);
+        }
+
+        if block.get_timestamp()
+            <= extending_blocks
+                .last()
```
omegablitz: Same as above

monad-eth-block-validator/src/lib.rs
```diff
@@ -192,7 +187,7 @@ where
         if gas_limit != &PROPOSAL_GAS_LIMIT {
             return Err(BlockValidationError::HeaderError);
         }
-        if *timestamp != header.timestamp / 1_000 {
+        if (*timestamp as u128) != header.timestamp_ns / 1_000_000_000 {
```
omegablitz: nit: calling `timestamp.into()` or `u128::from(timestamp)` is "safer" in the sense that if the type of `timestamp` changes, the compiler will complain if it's not losslessy convertable to a u128

monad-testutil/src/proposal.rs
```diff
@@ -112,7 +114,7 @@ where
             })
             .expect("key not in valset");
 
-        let seq_num = qc.get_seq_num() + SeqNum(1);
+        let seq_num = self.last_seq_num + SeqNum(1);
```
omegablitz, line 117: Need to update self.last_seq_num here

monad-eth-block-policy/src/lib.rs
```diff
@@ -638,6 +638,26 @@ where
             .unwrap();
         assert_eq!(first_block.get_seq_num(), self.last_commit + SeqNum(1));
 
+        if block.get_seq_num()
+            != extending_blocks
+                .last()
```
omegablitz: Is this expect safe at genesis?

monad-eth-block-policy/src/lib.rs
```diff
@@ -638,6 +638,26 @@ where
...
+                .get_seq_num()
+                + SeqNum(1)
+        {
+            return Err(BlockPolicyError::BlockNotCoherent);
+        }
+
+        if block.get_timestamp()
+            <= extending_blocks
+                .last()
+                .expect("must be at least 1 extending block")
```
omegablitz: Safe at genesis?

monad-consensus-state/src/lib.rs
```diff
@@ -1339,6 +1337,7 @@ where
     fn request_blocks_if_missing_ancestor(
         &mut self,
         qc: &QuorumCertificate<SCT>,
+        seq_num: Option<SeqNum>,
```
omegablitz: I don't think it makes sense to pass in proposed_seq_num here. We want to blocksync starting from the block_id that high_qc is over.

I think it's simplest to keep the args of this function the same as before, and look up the sequence number of qc.get_block_id() in the blocktree. If it exists, then the full range can be requested. If it doesn't exist, then can request 1 block starting at qc.get_block_id().

monad-consensus-state/src/lib.rs
```diff
@@ -1252,6 +1243,13 @@ where
             .pending_block_tree
             .get_blocks_on_path_from_root(&parent_bid)
             .expect("there should be a path to root");
+        let extending_block = pending_blocktree_blocks
+            .last()
+            .expect("must be at least one block");
```
omegablitz: Not sure this expect will work on genesis

monad-consensus-state/src/lib.rs
```diff
@@ -1059,19 +1070,21 @@ where
     }
 
     #[must_use]
-    fn maybe_statesync(&mut self) -> Vec<ConsensusCommand<ST, SCT, EPT>> {
+    fn maybe_statesync(
+        &mut self,
+        latest_seen_seq_num: SeqNum,
```
omegablitz: It looks like this is called with proposal_seq_num, which I think is incorrect - a proposer could send a proposal with an arbitrarily high sequence number.

We still want to check high_qc_seq_num, which we can do by by checking if high_qc.block_id exists in the blocktree. If it exists, then we check against that sequence number. If it doesn't exist, then we don't do anything.

monad-consensus-state/src/lib.rs
```diff
@@ -972,9 +982,10 @@ where
     fn process_certificate_qc(
         &mut self,
         qc: &QuorumCertificate<SCT>,
+        proposal_seq_num: Option<SeqNum>,
```
omegablitz: I don't think it makes sense to pass the proposal seqnum here, everything should be based on qc.block_id's sequence number

monad-consensus-state/src/lib.rs
omegablitz: Actually, it's probably easiest to reason about if we make this take no arguments, and have this function always work off of high_qc, just like maybe_statesync(). This is because of the observation that we always want to blocksync off high_qc. So we either blocksync 1 block (if we don't have the high_qc.block_id block in our blocktree) from high_qc.block_id, or `high_qc.block_id.seq_num - root` blocks.

We can then call this function in the same places we call maybe_statesync(), namely:
1. when high_qc gets bumped (process_qc)
2. when a new block is added to the blocktree (try_add_and_commit_blocktree)


We can also move:
```rust
if self
    .consensus
    .block_sync_requests
    .contains_key(&qc.get_block_id())
{
    return Vec::new();
}
```

above:
```rust
let Some(qc) = self.consensus.pending_block_tree.get_missing_ancestor(qc) else {
    return Vec::new();
};
```

to make this not slow.

ariqchowdhury: yeah, definitely easier to reason about if its just based on high_qc.

monad-mock-swarm/tests/order.rs
```diff
@@ -218,7 +216,7 @@ fn all_messages_delayed(direction: TransformerReplayOrder) {
         // ledger is in flight and not delivered to all peers yet
         // +2 is the case where the proposal to commit the last block in
         // ledger is received by all peers
-        TransformerReplayOrder::Reverse => (longest_ledger_before + 1, longest_ledger_before + 2),
+        TransformerReplayOrder::Reverse => (0, longest_ledger_before + 2),
```
omegablitz: Trying to gain intution as to why this changed. Does it make 2 blocksync requests now; one for fetching max_blocks=1, and the second for fetching the rest of the blocks?

monad-consensus-state/src/lib.rs
```diff
@@ -1336,49 +1340,60 @@ where
...
+            .pending_block_tree
+            .get_missing_ancestor(high_qc)
+        else {
+            return Vec::new();
+        };
 
-                request_range
-            };
+        let range = if let Some(high_qc_seq_num) =
+            self.consensus.pending_block_tree.get_seq_num_of_qc(high_qc)
```
omegablitz, line 1364: This is correct, but we'll always blocksync from high_qc, even if we have a chain of ancestors above high_qc.

One consequence of this is that if we have a missing block in our blocktree, we will send a new blocksync request for every new block we receive that bumps high_qc, even if we already have have an outstanding blocksync request that would fill the gap.


We could fix this by changing get_seq_num_of_qc to return the sequence number and blockid of the oldest ancestor.

It's unclear if this is worth fixing now though, the current solution might perform well in practice.

monad-mock-swarm/tests/order.rs
omegablitz: It's because it always blocksyncs off high_qc

monad-consensus-state/src/lib.rs
omegablitz: Addressed by #1363

PR #1349, opened 2024-12-29
Remove ledger_commit_info and flatten QC structure

ariqchowdhury:
resolves category-labs/category-internal#1009, resolves category-labs/category-internal#1011, category-labs/category-internal#1012

### DIFF
--- a/monad-state/src/statesync.rs
+++ b/monad-state/src/statesync.rs
@@ -128,7 +128,14 @@ where
         }
 
         if block_header.qc.is_commitable() {
-            let committed_seq_num = block_header.seq_num - self.commit_distance;
+            let latest_seq_num =
+                if let Some(b) = self.full_blocks.get(&block_header.get_parent_id()) {
+                    b.get_seq_num()
+                } else {
+                    return None;
+                };
+
+            let committed_seq_num = latest_seq_num - self.commit_distance;
             if committed_seq_num > root_seq_num + self.resync_threshold {
                 let target_blockid = block_header.qc.info.parent_id;
                 if let Some(target_block) = self.block_headers.get(&target_blockid) {


monad-state/src/statesync.rs
```diff
@@ -127,10 +127,10 @@ where
             self.proposal_buffer.pop_front();
         }
 
-        if commit_condition(block_header.round, block_header.qc.info) {
+        if block_header.qc.is_commitable() {
             let committed_seq_num = block_header.seq_num - self.commit_distance;
```
omegablitz: committed_seq_num should be block_header.qc.get_parent_seq_num()

block_header.seq_num isn't trusted

monad-proto/proto/ledger.proto
omegablitz, line 1: I guess this file can be deleted, doesn't really matter though

ariqchowdhury: yeah, all the proto stuff will be removed at some point

monad-state/src/statesync.rs
ariqchowdhury: are you saying the original code had an issue with using block_header.seq_num?

omegablitz: Yes the original code is messed up also

monad-state/src/statesync.rs
```diff
@@ -128,7 +128,14 @@ where
...
         if block_header.qc.is_commitable() {
-            let committed_seq_num = block_header.seq_num - self.commit_distance;
+            let latest_seq_num =
+                if let Some(b) = self.full_blocks.get(&block_header.get_parent_id()) {
+                    b.get_seq_num()
+                } else {
+                    return None;
+                };
+
+            let committed_seq_num = latest_seq_num - self.commit_distance;
```
omegablitz, line 138: I think now committed_seq_num is always latest_seq_num - SeqNum(1), but there's no meaningful difference, just that the resync_threshold will be off by one.

PR #1352, opened 2024-12-29
wait for execution results to validate a proposal

mikety:
Issue:
Solution:
Tracking issue:

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -9,7 +9,7 @@ use monad_consensus::{
         consensus_message::{ConsensusMessage, ProtocolMessage},
         message::{ProposalMessage, TimeoutMessage, VoteMessage},
     },
-    pacemaker::Pacemaker,
+    pacemaker::{Pacemaker, PacemakerCommand},
     validation::safety::Safety,
     vote_state::VoteState,
 };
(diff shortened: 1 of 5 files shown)


monad-consensus-state/src/lib.rs
```diff
@@ -414,10 +427,28 @@ where
...
         let state_root_action = self.state_root_hash_validation(&p);
 
-        if matches!(state_root_action, StateRootAction::Reject) {
-            return cmds;
+        match state_root_action {
+            StateRootAction::Reject => {
+                trace!("StateRoot Reject {:?}", p.block.get_seq_num().0);
+                return cmds;
+            }
+            StateRootAction::NotReady => {
```
omegablitz, line 435: `self.state_root_hash_validation` tries to look up the state root of `p.block.get_seq_num().max(self.config.execution_delay) - self.config.execution_delay`.

So a malicious proposer can send a proposal with an arbitrarily high seq_num and cause the round timer to be turned off.

Conceptually, I think we should still add the block to the blocktree even if we don't vote on it.

Maybe could move try_vote and try_propose to try_update_coherency, and then get rid of `pending_proposals`? Perhaps. call try_vote on all newly coherent blocks that contain high_qc?

mikety: It should be a way to validate the seq_num of a proposal, to make sure its not malicious. Currently SeqNum::add and SeqNum::sub panics on overflow.

omegablitz: There _shouldn't_ be anywhere where we call SeqNum::sub or SeqNum::add that can overflow.

mikety:
This is reopened as PR #1367

PR #1354, opened 2024-12-30
statesync protocol improvements

maxkozlovsky:
Implement flow control
Add error handling
Server session aborts
Limit memory use on client and server


maxkozlovsky:
was split and merged

PR #1362, opened 2024-12-31
Remove parent_round and parent_id from Vote

omegablitz:
Makes is_committable less error-prone, and removes parent_round/parent_id from QC.

The diff inside `try_commit` looks a lot bigger than it actually is, because I flattened out everything after the commit rule check.


omegablitz:
Opted to not remove qc.parent_round, qc.parent_id

PR #1364, opened 2024-12-31
Remove reth primitives from source files

eerkaijun:
Remove reth primitives from all source files. Mainly conversion of TransactionSigned -> alloy_consensus::TxEnvelope

part of the effort for https://github.com/category-labs/category-internal/issues/1025

mock swarm test pass


monad-eth-block-validator/src/lib.rs
```diff
@@ -93,8 +87,10 @@ where
         }
 
         // recovering the signers verifies that these are valid signatures
-        let signers = recover_signers(transactions, transactions.len())
-            .ok_or(BlockValidationError::TxnError)?;
+        let signers: Vec<Address> = transactions
+            .iter()
+            .map(|tx| tx.recover_signer().map_err(|_| BlockValidationError::TxnError))
```
omegablitz, line 93: Is there a batch sender recovery api?

eerkaijun: they don't have one. switched to parallel recovery

omegablitz: ref: https://github.com/paradigmxyz/reth/blob/fb6902880c42eaf6e1ff80a6ce04050a5244d514/crates/primitives-traits/src/transaction/recover.rs#L19-L25

Cargo.toml
omegablitz, line 1: remove reth-primitives from cargo.toml

eerkaijun: there's still reth primitives used in tests which are not removed yet (haven't got to those since we need to fix most tests anyways).

reth primitives are removed from source files

omegablitz:
Fixed by #1398

PR #1367, opened 2024-12-31
Wait for delayed state root

mikety:
Issue:
Solution:
Tracking issue:

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::{BTreeMap, HashMap},
+    collections::{BTreeMap, HashMap, LinkedList},
     ops::Deref,
     time::Duration,
 };
@@ -10,7 +10,7 @@ use monad_consensus::{
         consensus_message::{ConsensusMessage, ProtocolMessage},
         message::{ProposalMessage, TimeoutMessage, VoteMessage},
     },
-    pacemaker::Pacemaker,
+    pacemaker::{Pacemaker, PacemakerCommand},
     validation::safety::Safety,
     vote_state::VoteState,
 };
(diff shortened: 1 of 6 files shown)


monad-consensus-state/src/lib.rs
```diff
@@ -405,10 +418,28 @@ where
...
         let state_root_action = self.state_root_hash_validation(&p);
 
-        if matches!(state_root_action, StateRootAction::Reject) {
-            return cmds;
+        match state_root_action {
+            StateRootAction::Reject => {
+                trace!("StateRoot Reject {:?}", p.block_header.seq_num.0);
+                return cmds;
+            }
+            StateRootAction::NotReady => {
```
mikety, line 471: Preserved  @omegablitz comment:
self.state_root_hash_validation tries to look up the state root of p.block.get_seq_num().max(self.config.execution_delay) - self.config.execution_delay.

So a malicious proposer can send a proposal with an arbitrarily high seq_num and cause the round timer to be turned off.

Conceptually, I think we should still add the block to the blocktree even if we don't vote on it.

Maybe could move try_vote and try_propose to try_update_coherency, and then get rid of pending_proposals? Perhaps. call try_vote on all newly coherent blocks that contain high_qc?

mikety: the `seq_num `validation should happen[ early ](https://github.com/category-labs/monad-bft/blob/848172e84f80d6bc3c5544e66831cde4463eb5df/monad-state/src/consensus.rs#L318)to avoid the possibility of other issues.

michael-yxchen: we removed seq_num from QC, so seq_num verification requires parent block in the block tree and the check happens in [check_coherency](https://github.com/category-labs/monad-bft/blob/release-testnet/monad-eth-block-policy/src/lib.rs#L650) now

mikety:
This PR is replaced by [PR 1483 ](https://github.com/category-labs/monad-bft/pull/1483/commits/537753641b5c5aee4f91c54f17c0209eec43a086)

PR #1370, opened 2025-01-02
Fix single node startup with docker

eerkaijun:


eerkaijun:
commit is included in this pr: https://github.com/category-labs/monad-bft/pull/1373

PR #1373, opened 2025-01-03
Archive integration to release-testnet

eerkaijun:
E2e tested with RPC, on stressnet data

The RPC archive commit is already merged in master, but for some reason it's reverted on the release testnet branch so have to include it again


monad-archive/src/storage/cloud_proxy.rs
```diff
@@ -0,0 +1,145 @@
...
+
+        assert_eq!(result.get(&tx_hashes[0]).unwrap().tx.tx_hash(), &tx_hashes[0]);
+    }
+
+    #[tokio::test]
+    pub async fn test_cloud_proxy_reader_jank() {
+        let reader = CloudProxyReader::new(
+            "eoc53hGz4C4ZEFbfhB9YH3nd8iCQwqAz52plpZ7H",
+            url::Url::parse("https://9df09fanz1.execute-api.us-east-2.amazonaws.com/prod").unwrap(),
+            "stressnet-devnet2-4".to_owned(),
```
eerkaijun: @JoeHowarth I think we should remove this from the file

monad-rpc/src/main.rs
```diff
@@ -643,6 +664,30 @@ async fn main() -> std::io::Result<()> {
...
+                s3_bucket,
+                Some(region),
+                &archive_url,
+                &archive_api_key,
+                5,
+            )
+            .await
+            {
+                Ok(reader) => Some(reader),
+                Err(_) => None,
```
JoeHowarth: Log error?

monad-archive/src/storage/cloud_proxy.rs
eerkaijun: removed

monad-rpc/src/main.rs
eerkaijun: added

PR #1379, opened 2025-01-06
Archive bft blocks

JoeHowarth:
First pass at archiving bft blocks for data retention purposes. Tested with stressnet-gra004-4 bucket

This compares the files found in `bft-block-path` (and `bft-block-header-path` if not the same) to those already archived and uploads those not found.

Indexing these files will be left for a follow-up pr.

https://github.com/category-labs/category-internal/issues/1048
https://github.com/category-labs/category-internal/issues/1047

### DIFF
--- a/monad-archive/src/bin/monad-archiver/cli.rs
+++ b/monad-archive/src/bin/monad-archiver/cli.rs
@@ -6,9 +6,13 @@ use monad_archive::{ArchiveArgs, BlockDataReaderArgs};
 #[derive(Debug, Parser)]
 #[command(name = "monad-archive", about, long_about = None)]
 pub struct Cli {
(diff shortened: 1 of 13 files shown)


monad-archive/src/bin/monad-archiver/cli.rs
```diff
@@ -22,6 +26,27 @@ pub struct Cli {
     #[arg(long)]
     pub start_block: Option<u64>,
 
+    /// Path to folder containing bft block bodies
+    /// If set, archiver will upload these files to blob store provided in archive_sink
+    #[arg(long)]
+    pub bft_block_path: Option<PathBuf>,
```
ariqchowdhury: nit: `bft_block_body_path`

PR #1391, opened 2025-01-07
Do not terminate current requests when statesync target updates

maxkozlovsky:
If statesync target is updated while statesync is already running, finish the current iteration before switching to new target.

The previous approach of terminating requests can cause a problem if a time to complete statesync request is greater than time to update target. In this case the amount of work keeps increasing with new target and statesync never completes.

### DIFF
--- a/monad-statesync/src/ffi.rs
+++ b/monad-statesync/src/ffi.rs
@@ -192,27 +192,31 @@ impl<PT: PubKey> StateSync<PT> {
                 Some(statesync_send_request),
             );
             let mut current_target = None;
+            let mut next_target = None;
 
             while let Ok(response) = response_rx.recv() {
                 match response {
                     SyncResponse::UpdateTarget(target) => {
-                        tracing::debug!(
-                            ?current_target,
-                            new_target =? target,
-                            "updating statesync target"
-                        );
-                        let mut buf = Vec::new();
-                        target.encode(&mut buf);
-                        unsafe {
-                            bindings::monad_statesync_client_handle_target(
-                                // handle_target can be called on an active or inactive SyncCtx
-                                sync_ctx.get_or_create_ctx(),
-                                buf.as_ptr(),
-                                buf.len() as u64,
-                            )
-                        };
-                        progress.lock().unwrap().update_target(&target);
-                        current_target = Some(target);
+                        if current_target.is_none() {
+                            tracing::debug!(
+                                new_target =? target,
+                                "updating statesync target"
+                            );
+                            let mut buf = Vec::new();
+                            target.encode(&mut buf);
+                            unsafe {
+                                bindings::monad_statesync_client_handle_target(
+                                    // handle_target can be called on an active or inactive SyncCtx
+                                    sync_ctx.get_or_create_ctx(),
+                                    buf.as_ptr(),
+                                    buf.len() as u64,
+                                )
+                            };
+                            progress.lock().unwrap().update_target(&target);
+                            current_target = Some(target);
+                        } else {
+                            next_target.replace(target);
+                        }
                     }
                     SyncResponse::Response((from, response)) => {
                         assert!(current_target.is_some());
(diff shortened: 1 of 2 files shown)


monad-statesync/src/ffi.rs
```diff
@@ -263,13 +269,32 @@ impl<PT: PubKey> StateSync<PT> {
...
+                            bindings::monad_statesync_client_handle_target(
+                                sync_ctx.ctx,
+                                bindings::monad_sync_target {
+                                    n: target.n.0,
+                                    state_root: target.state_root,
+                                },
+                            )
+                        };
+                        progress.lock().unwrap().update_target(target)
+                    }
```
andr-dev:
```rust
if let Some(target) = next_target.take() {
  // second branch
} else {
  // first branch
}
```

maxkozlovsky: ok

monad-statesync/src/ffi.rs
```diff
@@ -277,14 +281,31 @@ impl<PT: PubKey> StateSync<PT> {
...
+                            )
+                        };
+                        progress.lock().unwrap().update_target(&current_target)
+                    } else {
+                        tracing::debug!(?target, "done statesync");
+                        progress.lock().unwrap().update_reached_target(&target);
+                        request_tx
+                            .send(SyncRequest::DoneSync(target))
+                            .expect("request_rx dropped mid DoneSync");
+                    }
```
andr-dev:
```rust
let Some(current_target) = &current_target else {
    tracing::debug!(?target, "done statesync");
    progress.lock().unwrap().update_reached_target(&target);
    request_tx.send(SyncRequest::DoneSync(target)).expect("request_rx dropped mid DoneSync");
    continue;
}

tracing::debug!(
    ?target,
    ?current_target,
    "statesync reached target"
);
...
```

maxkozlovsky: It is not clear if adding `continue`, which is flow control interruption, is an improvement over if/else statement. I'll leave it as is.

monad-statesync/src/ffi.rs
```diff
@@ -308,7 +329,6 @@ impl<PT: PubKey> StateSync<PT> {
             assert!(old_target.number < target.number);
         }
         self.current_target = Some(target.clone());
-        self.outbound_requests.clear();
```
andr-dev, line 321: Why are we no longer clearing outbound requests here?

maxkozlovsky: This is the point of PR, clearing requests restarts statesync discarding already received updates. Instead we finish the transfer for the current target and then sync to new one so forward progress is maintained

maxkozlovsky:
> Just to confirm, have we tested this in stressnet?

tested in flexnet simulating the condition when transfer takes longer than getting to new target, with the current code this causes statesync to run forever.

monad-statesync/src/ffi.rs
andr-dev: Ah got it, I think I got mixed up with the naming.

monad-statesync/src/ffi.rs
```diff
@@ -192,27 +192,31 @@ impl<PT: PubKey> StateSync<PT> {
...
+                            target.encode(&mut buf);
+                            unsafe {
+                                bindings::monad_statesync_client_handle_target(
+                                    // handle_target can be called on an active or inactive SyncCtx
+                                    sync_ctx.get_or_create_ctx(),
+                                    buf.as_ptr(),
+                                    buf.len() as u64,
+                                )
+                            };
+                            progress.lock().unwrap().update_target(&target);
```
omegablitz, line 215: Should this be done unconditionally for a more accurate estimate?

monad-statesync/src/ffi.rs
```diff
@@ -283,14 +287,31 @@ impl<PT: PubKey> StateSync<PT> {
...
+                        let mut buf = Vec::new();
+                        current_target.encode(&mut buf);
+                        unsafe {
+                            bindings::monad_statesync_client_handle_target(
+                                sync_ctx.get_or_create_ctx(),
+                                buf.as_ptr(),
+                                buf.len() as u64,
+                            )
+                        };
+                        progress.lock().unwrap().update_target(current_target)
```
omegablitz, line 307: This can be removed if the progress target is updated unconditionally above.

monad-statesync/src/ffi.rs
```diff
@@ -318,7 +339,6 @@ impl<PT: PubKey> StateSync<PT> {
             assert!(old_target.number < target.number);
         }
         self.current_target = Some(target.clone());
-        self.outbound_requests.clear();
```
omegablitz, line 321: Can delete the OutboundRequests::clear function if it's unlikely to ever be used again

monad-statesync/src/ffi.rs
maxkozlovsky: Does not look like it, the progress works for single target only.

omegablitz: Oh ok nvm then

monad-statesync/src/ffi.rs
maxkozlovsky: deleted

maxkozlovsky:
https://github.com/category-labs/category-internal/issues/1042

PR #1392, opened 2025-01-07
Add hostname to archiver metrics

JoeHowarth:


monad-archive/src/metrics.rs
```diff
@@ -64,6 +65,20 @@ impl Metrics {
...
+            warn!(
+                ?e,
+                default,
+                "Failed to read hostname when initializing metrics. Falling back to default"
+            );
+            default.to_owned()
+        }
+    }
+}
+
```
andr-dev: We should use the hostname crate for this. Additionally, if the archiver runs inside docker, it'll produce the hostname of the docker container which I believe is a random 12 character hex string.

PR #1397, opened 2025-01-08
eth_call chain config

michael-yxchen:
Consensus side changes for https://github.com/category-labs/monad/commit/65bea95b123d176d69081e130fb5a73bcb02a3e8

### DIFF
--- a/monad-rpc/src/gas_handlers.rs
+++ b/monad-rpc/src/gas_handlers.rs
@@ -85,6 +85,11 @@ pub async fn monad_eth_estimateGas<T: Triedb + TriedbPath>(
     }
 
     let sender = params.tx.from.unwrap_or_default();
+    let tx_chain_id = params
+        .tx
+        .chain_id
+        .expect("chain id must be populated")
+        .to::<u64>();
     let mut txn: reth_primitives::transaction::Transaction = params.tx.try_into()?;
 
     if matches!(txn.kind(), TxKind::Call(_)) && txn.input().is_empty() {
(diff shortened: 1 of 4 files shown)


monad-rpc/src/gas_handlers.rs
```diff
@@ -85,13 +85,19 @@ pub async fn monad_eth_estimateGas<T: Triedb + TriedbPath>(
     }
 
     let sender = params.tx.from.unwrap_or_default();
+    let tx_chain_id = params
+        .tx
+        .chain_id
+        .expect("chain id must be populated")
+        .to::<u64>();
```
eerkaijun, line 92: this is probably breakable, for legacy transactions chain ID is not mandatory. I think we should pass in app_state.chain_id into eth_call and doesn't matter what chain id that the user specify in eth_call and eth_estimateGas

michael-yxchen: https://github.com/category-labs/monad-bft/pull/1397/files/9a14d5972dd6daed1d0703015f1ea1a3932be15e#diff-5b5e509032b26a8074a92c39f08222d861d929aea1d29a0f940aec6746eb5db2R83-R85

tx.chain_id is set to app_state.chain_id if None is provided

michael-yxchen: if tx.chain_id != app_state.chain_id, should we early return an error?

eerkaijun: okay my bad, missed that.

andr-dev: In that case, we should change the expect message to something like "chain id populated by RPC" so if this expect somehow gets triggered, it's clear we've had a regression.

omegablitz:
cherry-picked this onto #1458

PR #1398, opened 2025-01-08
Update alloy version, remove reth_primitives

omegablitz:


monad-rpc/src/vpool.rs
```diff
@@ -431,369 +429,9 @@ impl VirtualPool {
...
-    ) {
+    ) -> (Vec<Recovered<TxEnvelope>>, Vec<Recovered<TxEnvelope>>) {
         let pending = self.pending_pool.by_addr(address);
         let queued = self.queued_pool.by_addr(address);
         (pending, queued)
     }
 }
-
-#[cfg(test)]
-mod tests {
```
eerkaijun, line 445: why do we delete all these tests?

omegablitz: Vpool will need to be changed to work well with the new txpool. That's why it's currently disabled on release-testnet.

PR #1399, opened 2025-01-08
Fix log index in rpc receipts

philipglazman:
Return the correct log index for a receipt which is its index in the list of all receipts in the block.

~~For reviewer, should I target `release-testnet` or `master` ?~~

Fixes https://github.com/category-labs/category-internal/issues/1071

### DIFF
--- a/monad-rpc/src/eth_txn_handlers.rs
+++ b/monad-rpc/src/eth_txn_handlers.rs
@@ -66,6 +66,7 @@ pub fn parse_tx_receipt(
     receipt: ReceiptEnvelope,
     block_num: u64,
     tx_index: u64,
+    num_prev_logs: usize,
 ) -> Result<TransactionReceipt, JsonRpcError> {
     let Ok(address) = tx.recover_signer() else {
         error!("transaction sender should exist");
@@ -93,7 +94,7 @@ pub fn parse_tx_receipt(
             block_timestamp,
             transaction_hash: Some(*tx.tx_hash()),
             transaction_index: Some(tx_index),
-            log_index: Some(log_index as u64),
+            log_index: Some((num_prev_logs + log_index) as u64),
             removed: Default::default(),
         })
         .collect();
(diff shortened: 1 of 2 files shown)


ariqchowdhury:
goal is for next release to be off of `master` so let's target there

monad-rpc/src/eth_txn_handlers.rs
```diff
@@ -627,6 +642,20 @@ async fn get_receipt_from_triedb<T: Triedb>(
...
             };
 
+            // Get the the number of previous logs in the block
+            let num_prev_logs = if tx_index > 0 {
+                let receipts = triedb_env
+                    .get_receipts(block_num)
+                    .await
+                    .map_err(JsonRpcError::internal_error)?;
+                receipts
+                    .iter()
```
eerkaijun, line 649: seems like this is also what reth does but feels like it could be expensive in Monad since our blocks are larger. not sure there is a better solution for now though (unless we can store the log index as an extra field in triedb)

monad-rpc/src/eth_txn_handlers.rs
```diff
@@ -430,6 +431,18 @@ pub async fn monad_eth_getTransactionReceipt<T: Triedb>(
...
     if let Some(archive_reader) = archive_reader {
         if let Ok(Some(tx_data)) = archive_reader.get_txdata(params.tx_hash.to_string()).await {
+            let num_prev_logs = if let Ok(bloom_receipts) = archive_reader
+                .get_block_receipts(tx_data.header_subset.block_number)
+                .await
+            {
+                bloom_receipts
+                    .iter()
+                    .take(tx_data.header_subset.tx_index as usize)
+                    .fold(0, |acc, receipt| acc + receipt.receipt.logs.len())
```
eerkaijun: we can consider storing log_index in the HeaderSubset struct during the archiving process. @JoeHowarth thoughts?

ariqchowdhury:
can we add flexnet test for this?

philipglazman:
> can we add flexnet test for this?

Yes, one exists here https://github.com/category-labs/monad-integration/pull/148

ariqchowdhury:
> > can we add flexnet test for this?
> 
> Yes, one exists here [category-labs/monad-integration#148](https://github.com/category-labs/monad-integration/pull/148)

that PR should merge into main right, not release-testnet?

monad-rpc/src/eth_txn_handlers.rs
ariqchowdhury: can open an issue to revisit this and check with triedb if this is possible and if its worth doing.

philipglazman: https://github.com/category-labs/category-internal/issues/1126

monad-rpc/src/eth_txn_handlers.rs
```diff
@@ -418,15 +419,45 @@ pub async fn monad_eth_getTransactionReceipt<T: Triedb>(
...
+                        };
+
+                        let prev_receipts = prev_receipts
+                            .iter()
+                            .take(tx_data.header_subset.tx_index as usize);
+
+                        (
+                            prev_receipts.fold(0, |acc, receipt| acc + receipt.logs().len()),
+                            tx_data.receipt.cumulative_gas_used()
+                                - prev_receipt.cumulative_gas_used(),
```
eerkaijun: is there any reason not to use `tx_data.header_subset.gas_used` directly? we already store that in the archive data

philipglazman: I potentially misunderstood the `header_subset` type. I thought gas_used referered to the cumulative gas used field in the blockheader. Looks like `gas_used` field in `header_subset` is the actual gas used for the transaction entry.

philipglazman: Fixed. I'll use `tx_data.header_subset.gas_used `

PR #1401, opened 2025-01-09
Avoid rebroadcasting invalid symbols

bamrith:
Unit tests added in two separate commits https://github.com/category-labs/monad-bft/commit/9025e6aa9bb2672e9aa52cef6a129d22b255f208 and https://github.com/category-labs/monad-bft/commit/2d94a40866469035a2a19ce9085a0354b4fe01d6 which are tested on this branch.

### DIFF
--- a/monad-raptor/src/r10/nonsystematic/decoder/managed_decoder.rs
+++ b/monad-raptor/src/r10/nonsystematic/decoder/managed_decoder.rs
@@ -1,22 +1,6 @@
-use std::{
-    cmp::Ordering,
-    io::{Error, ErrorKind},
-    iter,
-};
-
-use bitvec::prelude::*;
+use std::{cmp::Ordering, io::Error, iter};
 
 use crate::r10::nonsystematic::decoder::{BufferId, Decoder};
-
-// For a message with K source symbols, we accept up to the first MAX_REDUNDANCY * K
-// encoded symbols.
-//
-// Any received encoded symbol with an ESI equal to or greater than MAX_REDUNDANCY * K
-// will be discarded, as a protection against DoS and algorithmic complexity attacks.
-//
-// We pick 7 because that is the largest value that works for all values of K, as K
-// can be at most 8192, and there can be at most 65521 encoding symbol IDs.
-const MAX_REDUNDANCY: usize = 7;
 
 // We switch from doing peeling only to performing inactivation decoding when
 // num_received_encoded_symbols >= (MULTIPLIER * num_source_symbols) >> SHIFT .
@@ -88,27 +72,23 @@ impl BufferSet {
 #[derive(Debug)]
 pub struct ManagedDecoder {
     num_source_symbols: usize,
-    max_encoded_symbols: usize,
-    seen_esis: BitVec<usize, Lsb0>,
     symbol_len: usize,
     decoder: Decoder,
     buffer_set: BufferSet,
 }
 
 impl ManagedDecoder {
-    pub fn new(num_source_symbols: usize, symbol_len: usize) -> Result<ManagedDecoder, Error> {
-        let max_encoded_symbols = MAX_REDUNDANCY * num_source_symbols;
-
-        let seen_esis = bitvec![usize, Lsb0; 0; max_encoded_symbols];
-
-        let decoder = Decoder::with_capacity(num_source_symbols, max_encoded_symbols)?;
+    pub fn new(
+        num_source_symbols: usize,
+        encoded_symbol_capacity: usize,
+        symbol_len: usize,
+    ) -> Result<ManagedDecoder, Error> {
+        let decoder = Decoder::with_capacity(num_source_symbols, encoded_symbol_capacity)?;
 
         let buffer_set = BufferSet::new(decoder.num_temp_buffers_required(), symbol_len);
 
         Ok(ManagedDecoder {
             num_source_symbols,
-            max_encoded_symbols,
-            seen_esis,
             symbol_len,
             decoder,
             buffer_set,
@@ -118,39 +98,14 @@ impl ManagedDecoder {
     // TODO: Explore accepting Bytes as data, making rx_buffers a vector of enums
     // designating either an owned Box<[u8]> or an un-owned Bytes, and converting
     // un-owned to owned buffers whenever they are targeted for XORing.
-    pub fn received_encoded_symbol(
-        &mut self,
-        data: &[u8],
-        encoding_symbol_id: usize,
-    ) -> Result<(), Error> {
+    pub fn received_encoded_symbol(&mut self, data: &[u8], encoding_symbol_id: usize) {
         if self.symbol_len != data.len() {
-            return Err(Error::new(
-                ErrorKind::InvalidData,
-                format!(
-                    "ManagedDecoder with symbol_len = {} given symbol of length {}",
-                    self.symbol_len,
-                    data.len()
-                ),
-            ));
+            panic!(
+                "ManagedDecoder expected symbol of length = {} given symbol of length {}",
+                self.symbol_len,
+                data.len()
+            );
         }
-
-        if encoding_symbol_id >= self.max_encoded_symbols {
-            return Err(Error::new(
-                ErrorKind::InvalidData,
-                format!(
-                    "ManagedDecoder with max_encoded_symbols = {} received ESI {}",
-                    self.max_encoded_symbols, encoding_symbol_id,
-                ),
-            ));
-        }
-
-        if self.seen_esis[encoding_symbol_id] {
-            return Err(Error::new(
-                ErrorKind::InvalidData,
-                format!("ManagedDecoder already saw ESI {}", encoding_symbol_id,),
-            ));
-        }
-        self.seen_esis.set(encoding_symbol_id, true);
 
         let buf: Box<[u8]> = data.into();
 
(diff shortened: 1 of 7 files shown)


buytenh:
I understand that this is still WIP, but two comments so far:

- With your changes, the definition for (and the comments accompanying) MAX_REDUNDANCY can be moved from monad-raptor to monad-raptorcast, as the Raptor decoder now no longer cares about symbol index thresholds at all.

- Regarding your commit that fixes the decoder test, could you do that as follows?

```
        decoder.received_encoded_symbol(&buf, *esi).unwrap();

-       // Make sure that we detect multiple submissions of the same ESI.
+       // We feed some encoded symbols back into the decoder twice to test the
+       // Redundant buffer handling paths.
        if rand::thread_rng().gen_ratio(1, 100) {
-           decoder.received_encoded_symbol(&buf, *esi).unwrap_err();
+           decoder.received_encoded_symbol(&buf, *esi);
        }
```

  This way, we keep exercising the Raptor decoder path that deals with redundant received symbols, but without `.unwrap_err()`ing when that happens, since that case is now no longer an error from the Raptor decoder's point of view.

monad-raptor/src/r10/nonsystematic/decoder/managed_decoder.rs
```diff
@@ -89,26 +73,24 @@ impl BufferSet {
 pub struct ManagedDecoder {
     num_source_symbols: usize,
     max_encoded_symbols: usize,
```
buytenh: I think we can also move `max_encoded_symbols` to the raptorcast state as well, along with `seen_esis`, as the raptor decoder doesn't use it internally anymore, except to use as capacity estimate in the `::new()` call.

monad-raptor/src/r10/nonsystematic/decoder/managed_decoder.rs
```diff
@@ -89,26 +73,24 @@ impl BufferSet {
...
 
 impl ManagedDecoder {
-    pub fn new(num_source_symbols: usize, symbol_len: usize) -> Result<ManagedDecoder, Error> {
-        let max_encoded_symbols = MAX_REDUNDANCY * num_source_symbols;
-
-        let seen_esis = bitvec![usize, Lsb0; 0; max_encoded_symbols];
-
+    pub fn new(
+        num_source_symbols: usize,
+        max_encoded_symbols: usize,
```
buytenh: Since this is no longer a hard limit on anything, let's rename this argument to `encoded_symbol_capacity` or something along those lines?  This should make it clearer that it's OK to pass in a zero value, if the performance delta doesn't matter.

monad-raptor/tests/managed_decoder.rs
```diff
@@ -3,14 +3,17 @@
...
 
 fn test_single_decode(src: Vec<u8>) {
     let encoder: Encoder = Encoder::new(&src, SYMBOL_LEN).unwrap();
 
     let num_source_symbols = encoder.num_source_symbols();
+    let max_encoded_symbols = MAX_REDUNDANCY * num_source_symbols;
 
-    let mut decoder = ManagedDecoder::new(num_source_symbols, SYMBOL_LEN).unwrap();
+    let mut decoder =
+        ManagedDecoder::new(num_source_symbols, max_encoded_symbols, SYMBOL_LEN).unwrap();
```
buytenh: We could pass in a zero here, since this value is now only a recommendation.

monad-raptorcast/src/udp.rs
```diff
@@ -216,6 +189,66 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
                 }
             };
 
+            if decoder_state.decoder.symbol_len() != parsed_message.chunk.len() {
+                // invalid symbol length
+                tracing::debug!(?key, ?message.src_addr, ?encoding_symbol_id, "received invalid symbol len");
```
buytenh: I think this should be a warn!, since this is a (potentially slashable) error, and we should log the previously received symbol length and the current symbol length, along with the message author, application message hash, and whatever other fields are relevant.

monad-raptorcast/src/udp.rs
```diff
@@ -216,6 +189,66 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
...
+                continue;
+            }
+
+            if encoding_symbol_id >= decoder_state.decoder.max_encoded_symbols() {
+                // invalid symbol id
+                tracing::debug!(?key, ?message.src_addr, ?encoding_symbol_id, "received invalid symbol id");
```
buytenh: Can you add some more trace event fields here as well?

monad-raptorcast/src/udp.rs
```diff
@@ -216,6 +189,66 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
...
+                continue;
+            }
+
+            if decoder_state.seen_esis[encoding_symbol_id] {
+                // duplicate symbol
+                tracing::debug!(?key, ?message.src_addr, ?encoding_symbol_id, "received duplicate symbol id");
```
buytenh: And here.

monad-raptorcast/src/udp.rs
```diff
@@ -216,6 +189,66 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
...
+
+                // forward all broadcast packets to full nodes
+                full_node_forward_batch_guard.queue_forward(payload_start_idx, payload_end_idx);
+            } else if self_hash != parsed_message.recipient_hash {
+                tracing::debug!(
+                    ?self_hash,
+                    recipient_hash =? parsed_message.recipient_hash,
+                    "dropping spoofed message"
+                );
+                continue;
```
buytenh: Moving all these checks to after the message cache key creation and cache lookup/insertion seems to introduce a big DoS hazard -- e.g. we can now spam random messages with random signatures, which will resolve to random author public keys that are not in our validator set, and by the time we figure out that these authors are not in our validator set, we will throw the message away, but the message cache trashing will already have happened.

Ideally, we should do as many checks as we can before the cache lookup/insertion, and keep doing the checks that depend on the cached data after the cache lookup/insertion, i.e., do every sanity check as early as possible, as soon as its dependencies are available.  Do you think that's possible here?

bamrith: I've moved the sanity checks back to where it initially was (before the cache lookup), but the rebroadcast and forwarding is now queued only after the invalid symbol checks.
This doesn't look ideal right now, but this will definitely change when decoding is split from rebroadcast/forwarding.

buytenh: Sounds good.

monad-raptor/src/r10/nonsystematic/decoder/managed_decoder.rs
```diff
@@ -118,54 +98,23 @@ impl ManagedDecoder {
...
-        if self.symbol_len != data.len() {
-            return Err(Error::new(
-                ErrorKind::InvalidData,
-                format!(
-                    "ManagedDecoder with symbol_len = {} given symbol of length {}",
-                    self.symbol_len,
-                    data.len()
-                ),
-            ));
-        }
```
buytenh: One final nit -- could you leave this symbol length check in, but make it panic, instead of returning an Err()?

buytenh:
I wrote a test for this, which I pushed here:

   https://github.com/category-labs/monad-bft/tree/buytenh/raptorcast-rebroadcast-test

This test is currently failing on the master branch, as expected.  With your changes from this PR, the test should start passing (but I haven't verified that).

Note that this depends on another pending PR -- https://github.com/category-labs/monad-bft/pull/1419 .

If you could verify that this test passes on your PR, then I think we can merge your PR once it is mergeable, and then merge PR #1419, and then I'll open a separate PR to add this test.

monad-raptorcast/src/udp.rs
```diff
@@ -232,52 +296,41 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
...
-                    )
+                .received_encoded_symbol(&parsed_message.chunk, encoding_symbol_id);
+
+            *decoder_state
+                .recipient_chunks
+                .entry(parsed_message.recipient_hash)
+                .or_insert(0) += 1;
+
+            if decoder_state.decoder.try_decode() {
+                let Some(mut decoded) = decoder_state.decoder.reconstruct_source_data() else {
```
andr-dev, line 323: I'm not sure what try_decode and reconstruct_source_data() do but it seems like an antipattern to call a fn on a struct and then call another conditioned on the first producing true. Can we make the reconstruct_source_data fn return None if try_decode() is false, emit the `tracing::error!(...)` and return None next, or produce the Some(...)?

Another idea if we want to emit error data would be to change reconstruct_source_data's return to Option<Result<_, _>> so this would look something like
```rust
let Some(decoded_result) else {
  continue;
};

match decoded_result {
  Err(e) => {
    tracing::error!(...);
    continue
  }
  Ok(mut decoded) => {
    ...
  }
}
```

bamrith: That makes sense to me. Before I make that change, is there a specific reason why it's written this way @buytenh ?

buytenh: `reconstruct_source_data()` is somewhat expensive, and there are cases (mainly for testing) where we want to check whether or not decoding was successful but without reconstructing the source data, which is why we don't do this unconditionally -- but those uses live mostly out-of-tree, so we can make the in-tree version of the Raptor decoder make more sense for its specific use case.

buytenh:
LGTM!

monad-raptorcast/src/udp.rs
```diff
@@ -204,9 +212,42 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
...
+                                &parsed_message.author,
+                                || targets.view().keys().copied().collect(),
+                            )
+                        }
+
+                        // forward all broadcast packets to full nodes
+                        full_node_forward_batch_guard
+                            .queue_forward(payload_start_idx, payload_end_idx);
+                    }
+                }
```
michael-yxchen: it'd be nice if we don't duplicate the rebroadcasting code, but maybe it's non trivial to do

bamrith: made it a closure right above this. there are borrowing issues if placed elsewhere.

PR #1403, opened 2025-01-09
Proposal size check and testnet config

eerkaijun:
Closes
https://github.com/category-labs/category-internal/issues/796
https://github.com/category-labs/category-internal/issues/998
https://github.com/category-labs/category-internal/issues/1003

### DIFF
--- a/monad-eth-block-validator/src/lib.rs
+++ b/monad-eth-block-validator/src/lib.rs
@@ -2,10 +2,13 @@ use std::collections::BTreeMap;
 
 use alloy_consensus::transaction::Transaction;
 use alloy_primitives::U256;
+use alloy_rlp::Encodable;
 use monad_consensus_types::{
     block::{Block, BlockKind, BlockPolicy, BlockType},
     block_validator::{BlockValidationError, BlockValidator},
-    payload::{Payload, TransactionPayload},
+    payload::{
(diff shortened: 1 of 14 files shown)


monad-eth-block-validator/src/lib.rs
```diff
@@ -95,14 +104,15 @@ impl EthValidator {
...
 
-                if validated_txns.len() > self.tx_limit {
-                    return Err(BlockValidationError::TxnError);
-                }
-
                 let total_gas = validated_txns
                     .iter()
                     .fold(0, |acc, tx| acc + tx.gas_limit());
-                if total_gas > self.block_gas_limit {
+                if total_gas > PROPOSAL_GAS_LIMIT {
```
omegablitz, line 104: I think we should delete block_txn_limit and block_gas_limit from the node.toml configuration, now that both are consts.

eerkaijun: removed block_gas_limit in node.toml, we still use block_txn_limit. personally i feel like these should all be in node.toml rather than hardcoded constants, but maybe we can revisit later

eerkaijun:
already updated flexnet: https://github.com/category-labs/monad-integration/pull/150

eerkaijun:
https://github.com/category-labs/category-internal/issues/1080

PR #1409, opened 2025-01-09
Use uniform distribution to generate validator index

maxkozlovsky:
Use uniform distribution to avoid bias in index generation.

### DIFF
--- a/monad-validator/Cargo.toml
+++ b/monad-validator/Cargo.toml
@@ -15,7 +15,10 @@ monad-types = { workspace = true }
 
 as-any = { workspace = true }
 auto_impl = { workspace = true }
+rand = { workspace = true }
+rand_chacha = { workspace = true }
 tracing = { workspace = true }
+alloy-primitives = { workspace = true, features = ["rand"] }
 
 [dev-dependencies]
 monad-testutil = { workspace = true }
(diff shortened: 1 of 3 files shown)


monad-validator/Cargo.toml
```diff
@@ -15,7 +15,10 @@ monad-types = { workspace = true }
 
 as-any = { workspace = true }
 auto_impl = { workspace = true }
+rand = { workspace = true }
+rand_chacha = { workspace = true }
 tracing = { workspace = true }
+alloy-primitives = { workspace = true }
```
omegablitz: nit: sort

Cargo.toml
```diff
@@ -91,7 +91,7 @@ alloy-consensus = "0.8"
 alloy-eips = "0.8"
 alloy-json-rpc = "0.8"
 alloy-network = "0.8"
-alloy-primitives = "=0.8.18"
+alloy-primitives = { version = "=0.8.18", features = ["rand"] }
```
omegablitz: Should the feature flag go in the `monad-validator/Cargo.toml`? Not sure what the difference is

maxkozlovsky: Ok moved feature enable to local Cargo.toml. 

Cargo will compute union of features across all sub crates according to my reading, so it is better to enable it locally when only one sub crate needs the feature. If all/most sub crates need feature then it can be moved to top level Cargo.toml.

PR #1413, opened 2025-01-10
Remove StateRootValidator

omegablitz:


monad-consensus-state/src/lib.rs
```diff
@@ -1025,8 +1060,8 @@ where
         // set high_qc to N
         // execution will sync to N-2*delay using state_root in consensus_root
 
-        assert!(self.config.live_to_statesync_threshold > self.state_root_validator.get_delay());
-        let max_committed_seq_num = high_qc_seq_num - self.state_root_validator.get_delay();
+        assert!(self.config.live_to_statesync_threshold > self.config.execution_delay);
+        let max_committed_seq_num = high_qc_seq_num - self.config.execution_delay;
```
michael-yxchen, line 1064: nit: the assertion will fail when `execution_delay == SeqNum::MAX` I assume we're not testing this function in unit tests so it's fine

monad-consensus-state/src/lib.rs
```diff
@@ -2988,6 +3019,7 @@ mod test {
     #[test]
     fn test_unavailable_state_root_during_proposal() {
         let num_state = 4;
+        let execution_delay = SeqNum(1);
```
michael-yxchen, line 3022: nit: delay of 1 works. 0 is might be more explicit cuz we'd never have state root for the same proposal

monad-consensus-state/src/lib.rs
omegablitz: This is true - but I don't immediately know a sensible way of fixing it here, so gonna ignore for now

monad-consensus-state/src/lib.rs
omegablitz: I think the actual delay here is irrelevant, because we're not calling `add_execution_result` anywhere in this function. Will leave as is so the test structure matches the `test_state_root_updates` test below, which uses the same execution_delay of SeqNum(1), but calls add_execution_result.

monad-consensus-state/src/lib.rs
```diff
@@ -344,6 +343,26 @@ where
...
+        {
+            std::collections::btree_map::Entry::Vacant(entry) => {
+                entry.insert(execution_result);
+            }
+            std::collections::btree_map::Entry::Occupied(entry) => {
+                assert_eq!(entry.get(), &execution_result);
+            }
+        };
+        Vec::new()
+    }
```
andr-dev, line 364: We should remove the Vec::new() if this never returns a ConsensusCommand, saves us an allocation.

monad-consensus-state/src/lib.rs
```diff
@@ -871,13 +890,29 @@ where
...
+                    .iter()
+                    .map(|b| b.clone().get_full_block())
                     .collect();
                 cmds.push(ConsensusCommand::<ST, SCT>::LedgerCommit(
                     unvalidated_blocks,
                 ));
 
+                let last_committed_block = blocks_to_commit
+                    .last()
+                    .expect("blocks_to_commit is not empty");
```
andr-dev, line 902: If we're only using `last_committed_block.get_seq_num()`, we should move this `let last_committed_block` above the `let unvalidated_blocks` and immediately get the SeqNum so we can still `into_iter()` above since we won't be holding on to a `blocks_to_commit` ref.

monad-consensus-state/src/lib.rs
omegablitz: This will return a non-empty vector once caught up to release-testnet.

Also, I don't think vector implementations typically allocate on default (empty) constructed vectors.

monad-consensus-state/src/lib.rs
```diff
@@ -1467,41 +1499,29 @@ where
...
+                "execution result not ready, unable to verify"
+            );
+            // TODO unify these metrics
+            self.metrics.consensus_events.rx_execution_lagging += 1;
+            self.metrics.consensus_events.rx_missing_state_root += 1;
+            return StateRootAction::Defer;
+        };
+
+        if expected_execution_results != p.block.get_state_root() {
+            debug!("State root hash in proposal conflicts with local value");
```
andr-dev, line 1517: Do we expect a lot of junk (from other validators or malicious parties) where the state root hash is incorrect? If not, it might be worthwhile adding the `expected_execution_results` and `state_root` to the log for easier debugging if something happens.

monad-consensus-state/src/lib.rs
```diff
@@ -4881,24 +4921,23 @@ mod test {
...
                     u128::MAX,
-                    SeqNum(4),
+                    execution_delay,
                     InMemoryBlockState::genesis(std::iter::once((sender_1_address, 0)).collect()),
                 )
             },
             || EthValidator::new(10000, 1337),
             EthTxPool::default_testing(),
-            || NopStateRoot,
+            execution_delay, // execution_delay
```
andr-dev: Stray comment

monad-types/src/lib.rs
```diff
@@ -188,6 +188,12 @@ impl Rem for SeqNum {
     }
 }
 
+impl SeqNum {
+    pub fn saturating_add(self, other: Self) -> Self {
```
andr-dev: const :)

monad-consensus-state/src/lib.rs
omegablitz: I did this to minimize the diff with release-testnet, which doesn't need an owned block at all: https://github.com/category-labs/monad-bft/blob/1a5c9e66336f5f1943408c684e1e0b6e0de0dfe1/monad-consensus-state/src/lib.rs#L888-L913

monad-consensus-state/src/lib.rs
omegablitz: Should pretty much never happen, we can add this later, don't want to wait for CI again haha.

This is copied from the existing code, so is not a regression.

monad-consensus-state/src/lib.rs
omegablitz: Don't want to wait for CI again, will fix later

monad-types/src/lib.rs
omegablitz: Don't want to wait for CI again, will fix later

omegablitz: jk, fixed

monad-consensus-state/src/lib.rs
omegablitz: jk, fixed

PR #1414, opened 2025-01-10
Add configuration validation

maxkozlovsky:
Add configuration validation functionality. Verify statesync threshold sanity.

### DIFF
--- a/monad-node/src/config/mod.rs
+++ b/monad-node/src/config/mod.rs
@@ -4,6 +4,7 @@ use monad_crypto::certificate_signature::CertificateSignaturePubKey;
 use monad_eth_types::{serde::deserialize_eth_address_from_str, EthAddress};
 use monad_secp::SecpSignature;
 use serde::Deserialize;
+use thiserror::Error;
 
 mod bootstrap;
 pub use bootstrap::{NodeBootstrapConfig, NodeBootstrapPeerConfig};
@@ -23,6 +24,15 @@ pub(crate) type SignatureType = SecpSignature;
 pub type SignatureCollectionType =
     BlsSignatureCollection<CertificateSignaturePubKey<SignatureType>>;
 
+#[derive(Error, Debug)]
+pub enum NodeConfigError {
+    #[error(transparent)]
+    ParseError(#[from] toml::de::Error),
+
+    #[error("Node configuration error: {0}")]
+    ValidationError(String),
+}
+
 #[derive(Debug, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub struct NodeConfig {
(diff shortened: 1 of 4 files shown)


monad-node/src/config/mod.rs
```diff
@@ -57,3 +67,20 @@ pub struct NodeConfig {
 }
 
 pub type ForkpointConfig = Checkpoint<SignatureCollectionType>;
+
+impl NodeConfig {
+    pub fn parse(toml_str: &str) -> Result<Self, NodeConfigError> {
+        let config: NodeConfig = toml::from_str(toml_str)?;
+        config.verify()?;
+        Ok(config)
```
andr-dev, line 75: We should just inline the verify function here.

monad-node/src/config/mod.rs
```diff
@@ -40,9 +50,9 @@ pub struct NodeConfig {
     // must be <= ipc_max_queued_batches
     pub ipc_queued_batches_watermark: u8,
 
-    pub statesync_threshold: u16,
-    pub statesync_max_concurrent_requests: u8,
-    pub statesync_request_timeout_ms: u16,
+    pub statesync_threshold: u64,
+    pub statesync_max_concurrent_requests: u64,
+    pub statesync_request_timeout_ms: u64,
```
andr-dev, line 56: Any reason to change these?

monad-node/src/config/mod.rs
maxkozlovsky: Separate function seems better especially if the number of checks increases.

monad-node/src/config/mod.rs
maxkozlovsky: So the types match for statesync_threshold, execution_delay which is u64. and the rest of them for consistency and to match the use (concurrent_requests is usize where it used, ms is u64 in Duration::from_msecs). There is no reason to use smaller types in single use data structure.

ariqchowdhury (approved):

monad-node/src/config/mod.rs
```diff
@@ -57,3 +67,20 @@ pub struct NodeConfig {
...
+
+impl NodeConfig {
+    pub fn parse(toml_str: &str) -> Result<Self, NodeConfigError> {
+        let config: NodeConfig = toml::from_str(toml_str)?;
+        config.verify()?;
+        Ok(config)
+    }
+
+    fn verify(&self) -> Result<(), NodeConfigError> {
+        if self.statesync_threshold <= self.consensus.execution_delay * 2 {
```
0xSqualo, line 79: @maxkozlovsky should there also be an upper bound check here? 

Considering even a very large number to cover edge cases (to catch corruption or accidental misconfiguration).

PR #1416, opened 2025-01-10
Add static validation condition

eerkaijun:
Do we also want to reject nonce that are too low in `eth_sendRawTransaction`?

### DIFF
--- a/monad-eth-block-policy/src/lib.rs
+++ b/monad-eth-block-policy/src/lib.rs
@@ -5,7 +5,7 @@ use alloy_primitives::U256;
 use itertools::Itertools;
 use monad_consensus_types::{
     block::{Block, BlockPolicy, BlockPolicyError, BlockType, FullBlock},
-    payload::{Payload, PayloadId},
+    payload::{Payload, PayloadId, PROPOSAL_GAS_LIMIT},
     quorum_certificate::QuorumCertificate,
     signature_collection::SignatureCollection,
     state_root_hash::StateRootHash,
@@ -107,6 +107,7 @@ pub enum TransactionError {
     MaxPriorityFeeTooHigh,
     InitCodeLimitExceeded,
     GasLimitTooLow,
+    GasLimitTooHigh,
 }
 
 /// Stateless helper function to check validity of an Ethereum transaction
@@ -139,6 +140,10 @@ pub fn static_validate_transaction(
     let intrinsic_gas = compute_intrinsic_gas(tx);
     if tx.gas_limit() < intrinsic_gas {
         return Err(TransactionError::GasLimitTooLow);
+    }
+
+    if tx.gas_limit() > PROPOSAL_GAS_LIMIT {
+        return Err(TransactionError::GasLimitTooHigh);
     }
 
     Ok(())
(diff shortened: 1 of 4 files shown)


monad-eth-block-policy/src/lib.rs
```diff
@@ -982,6 +987,40 @@ mod test {
             result,
             Err(TransactionError::MaxPriorityFeeTooHigh)
         ));
+
+        // transaction with gas limit lower than intrinsic gas
+        let tx_gas_limit_too_low = TxEip1559 {
```
andr-dev, line 992: I just merged a change that introduces a `make_eip1559_tx` helper fn to monad-eth-testutil that we can use here

monad-eth-txpool/tests/txpool.rs
```diff
@@ -243,39 +243,28 @@ fn run_eth_txpool_test<const N: usize>(events: [TxPoolTestEvent<'_>; N]) {
...
-            add_to_blocktree: true,
-        },
-        TxPoolTestEvent::Block(Box::new(|pool| {
-            assert_eq!(pool.num_txs(), 1);
-        })),
-    ]);
+    run_eth_txpool_test([TxPoolTestEvent::InsertTxs {
+        txs: vec![(&tx, false)],
+        expected_pool_size_change: 0,
+    }]);
```
andr-dev: Let's keep the CreateProposal to further assert it didn't somehow get added to the pool.

monad-eth-txpool/tests/txpool.rs
```diff
@@ -412,7 +401,7 @@ fn another_non_trivial_example() {
 #[test]
 #[traced_test]
 fn attacker_tries_to_include_transaction_with_large_gas_limit_to_exit_proposal_creation_early() {
```
andr-dev: Can we assert here that `10 * GAS_LIMIT < PROPOSAL_GAS_LIMIT`?

monad-eth-block-policy/src/lib.rs
eerkaijun: there will be circular dependency

andr-dev: You should be able to add monad-eth-testutil as a dev dependency.
EDIT: Just tested on master and it works.

monad-rpc/src/eth_txn_handlers.rs
```diff
@@ -357,6 +359,21 @@ pub async fn monad_eth_sendRawTransaction(
             let signer = txn.recover_signer().map_err(|_| {
                 JsonRpcError::custom("cannot ec recover sender from transaction".to_string())
             })?;
+
+            // drop transactions with nonce too low
+            let latest_block_num = get_block_num_from_tag(triedb_env, BlockTags::Latest).await?;
```
omegablitz, line 364: Not sure it's worth doing the extra triedb read here for every ingested transaction - the TxPool will check this anyways.

andr-dev: There are some redundant checks like these which are actually good to have in both rpc and txpool because it avoids overloading bft with junk txs. With the txpool thread, this becomes less of an issue but it's still nice to drop bad txs right at the beginning of the pipeline.

omegablitz: Does the current txpool have a bound on the number of state lookups it does? Or is it always one per tx inserted into the txpool.

eerkaijun: the additional check is mostly for conformance reason

andr-dev: @omegablitz It's bounded by the capacity of the txpool tracked map which is set with state lookups in mind. A full txpool should in the worst case take about 250ms and that will decrease with the txpool thread.

PR #1418, opened 2025-01-10
Remove SeqNum and timestamp from QC; change timestamp to ns

michael-yxchen:
Rebase of: https://github.com/category-labs/monad-bft/commit/b6136ed16f01511502ba9072c6724760c4a1ba0e
Closes: https://github.com/category-labs/category-internal/issues/1093


monad-state/examples/forkpoint-reset.rs
omegablitz, line 1: nit: can delete this (but will need to remove forkpoint-reset from the dockerfiles too)

monad-state/src/lib.rs
```diff
@@ -1300,12 +1291,11 @@ mod test {
                     round: Round(4030),
                     parent_id: BlockId(Hash([0x06_u8; 32])),
                     parent_round: Round(4027),
-                    seq_num: SeqNum(2998), // one block before boundary block
-                    timestamp: 1,
                 },
                 ledger_commit_info: CommitResult::NoCommit,
             },
         };
+        let qc_seq_num = SeqNum(2999);
```
omegablitz: just checking that the change from 2998 -> 2999 was intended

monad-mock-swarm/tests/order.rs
```diff
@@ -219,12 +219,8 @@ fn all_messages_delayed(direction: TransformerReplayOrder) {
...
         // when replayed forward, node should populate blocktree in order
         TransformerReplayOrder::Forward => (0, 7),
-        // when replayed in reverse, should request every block in ledger
-        // +1 is the case where the proposal to commit the last block in
-        // ledger is in flight and not delivered to all peers yet
-        // +2 is the case where the proposal to commit the last block in
-        // ledger is received by all peers
-        TransformerReplayOrder::Reverse => (longest_ledger_before + 1, longest_ledger_before + 2),
+        // when replayed in reverse, only one request is done from high_qc
+        TransformerReplayOrder::Reverse => (1, 1),
```
omegablitz, line 219: I just realized that the changes here got dropped - as in, they never made it into release-testnet:
https://github.com/category-labs/monad-bft/pull/1363/files
These changes stop blocksync spam when replaying propsals in order, as is what happens post-statesync. They also make it so that this test does not change.

We'll want these changes, but not sure how hard they'll be to squash into this commit. If it's too annoying to squash, can defer to later - I'll include it in my giga commit.

monad-consensus-state/src/lib.rs
```diff
@@ -2304,55 +2350,55 @@ mod test {
         assert!(cmds.is_empty());
     }
```
omegablitz, line 2277: nit: can #[ignore] it so it looks less ugly

monad-consensus-state/src/lib.rs
```diff
@@ -1239,20 +1267,41 @@ where
...
+                    extending_block.get_seq_num(),
+                    extending_block.get_seq_num() + SeqNum(1),
+                    self.block_timestamp
+                        .get_valid_block_timestamp(extending_block.get_timestamp()),
+                )
+            } else {
+                (
+                    self.consensus.pending_block_tree.root().seq_num,
+                    self.consensus.pending_block_tree.root().seq_num + SeqNum(1),
+                    self.block_timestamp.get_valid_block_timestamp(0),
```
omegablitz: should this be from root? I guess it's omitted because RootInfo doesn't have timestamp right now?

monad-consensus-state/src/lib.rs
```diff
@@ -1090,22 +1116,24 @@ where
...
-        {
-            self.metrics.consensus_events.failed_ts_validation += 1;
-            warn!(prev_block_ts = ?validated_block.get_qc().get_timestamp(),
-                curr_block_ts = ?validated_block.get_timestamp(),
-                local_ts = ?self.block_timestamp.get_current_time(),
-                "Timestamp validation failed"
-            );
-            return cmds;
-        }
+        // TODO: add get_qc_timestamp function to block tree
```
omegablitz: Does this mean that we don't do any timestamp validation right now, besides the monotonically increasing check?

Or is the check also somewhere else

omegablitz: I guess the missing piece for this is having timestamp in RootInfo?

monad-state/src/lib.rs
michael-yxchen: not intended, will revert

monad-mock-swarm/tests/order.rs
michael-yxchen: I'll try to apply those changes on top on this PR

monad-consensus-state/src/lib.rs
michael-yxchen: I'll re-enable it. The TODO was meant to remind myself

monad-mock-swarm/tests/order.rs
michael-yxchen: Deferred to giga commit. Null block and different blocksync range make it confusing

omegablitz: category-labs/category-internal/issues/1094

PR #1425, opened 2025-01-13
Add watermark for txpool time-based eviction

andr-dev:
Closes https://github.com/category-labs/category-internal/issues/1069.


monad-eth-txpool/src/tracked/mod.rs
```diff
@@ -340,6 +345,14 @@ where
     }
 
     pub fn evict_expired_txs(&mut self) {
```
omegablitz, line 347: Is this used only for time-based expiry? I'm not fully understanding the purpose of EVICT_ADDRESSES_WATERMARK.

PR #1426, opened 2025-01-13
Send latency-insensitive messages over TCP instead of UDP

eclark0426:
Forwarded transactions and block sync commands were both being sent over UDP. To keep consistent with UDP only being used for latency-sensitive messages (proposals, votes, etc), this moves those over to TCP instead.

Closes https://github.com/category-labs/category-internal/issues/901

### DIFF
--- a/monad-state/src/blocksync.rs
+++ b/monad-state/src/blocksync.rs
@@ -118,7 +118,7 @@ where
         cmds.into_iter()
             .map(|command| WrappedBlockSyncCommand {
                 // TODO: timeout should be more aggressive for headers request
-                request_timeout: *self.delta * 5,
+                request_timeout: *self.delta * 7,
                 command,
             })
             .collect()
@@ -140,7 +140,7 @@ where
         match wrapped.command {
             BlockSyncCommand::SendRequest { to, request } => {
                 vec![Command::RouterCommand(RouterCommand::Publish {
-                    target: RouterTarget::PointToPoint(to),
+                    target: RouterTarget::TcpPointToPoint(to),
                     message: VerifiedMonadMessage::BlockSyncRequest(request),
                 })]
             }


omegablitz:
Forwarded txs should be considered latency sensitive, as that's latency observable by end users.

omegablitz:
Although to follow up on that, the raptor redundancy for forwarded txs should be much lower than what it is right now.

eclark0426:
> Forwarded txs should be considered latency sensitive, as that's latency observable by end users.

Pinging @buytenh on this, since he was the one in the original issue that suggested forwarded transactions should be sent over TCP instead of UDP

buytenh:
> > Forwarded txs should be considered latency sensitive, as that's latency observable by end users.
> 
> Pinging @buytenh on this, since he was the one in the original issue that suggested forwarded transactions should be sent over TCP instead of UDP

Moving transaction forwarding from UDP to TCP was originally discussed with @michael-yxchen , but @omegablitz pointed out that transaction forwarding is also sensitive to round timing, and in general, he understands this matter better than I do, so we should probably keep forwarding transactions over UDP.

As to lowering the redundancy, that could increase the risk of the message being undecodeable at the receiver if we have too much packet loss, but perhaps we could, e.g. send encoded symbols 0..1.1*num_source_symbols at the highest priority, encoded symbols 1.1*num_source_symbols..1.5*num_source_symbols at a lower priority, and encoded symbols 1.5*num_source_symbols..2.0*num_source_symbols at the lowest priority, or something like that, such that we will still transmit more than enough redundant symbols for decoding to succeed at the receiver, but not at the (scheduling/latency) expense of other messages we may also want to transmit.  This would require adding some sort of transmit traffic priority ordering scheme to dataplane(-v2), but that isn't very difficult to do.

omegablitz:
I was thinking of lowering the redundancy more as a mechanism to reduce the amount of data being firehosed to the proposer.
Every node in the network will be trying to forward transactions to the proposer simultaneously, so was unsure if high redundancy could cause more harm than good.

buytenh:
> Although to follow up on that, the raptor redundancy for forwarded txs should be much lower than what it is right now.

(This is https://github.com/category-labs/category-internal/issues/985 )

eclark0426:
Moved forwarded txns back to UDP.

omegablitz:
I think the blocksync request timeout will need to be increased, assuming that the blocksync response uses a separate TCP connection. I know it did before, but that might have been updated recently.

If that's the case, now it's 3 delta for request, 3 delta for response?
Before it was 1 delta for request, 3 delta for response.

PR #1427, opened 2025-01-14
Use balance override to estimate allowance if exist

eerkaijun:
If balance override is specified during eth_call and eth_estimateGas, we shouldn't try to get the balance from triedb anymore. 

Added a test case in flexnet: https://github.com/category-labs/monad-bft/pull/1427

### DIFF
--- a/monad-rpc/src/call.rs
+++ b/monad-rpc/src/call.rs
@@ -7,6 +7,7 @@ use monad_cxx::StateOverrideSet;
 use monad_rpc_docs::rpc;
 use monad_triedb_utils::triedb_env::{Triedb, TriedbPath};
 use serde::{Deserialize, Serialize};
+use tracing::trace;
 
 use crate::{
     block_handlers::get_block_num_from_tag,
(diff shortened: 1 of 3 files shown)


monad-rpc/src/call.rs
```diff
@@ -272,18 +272,38 @@ pub async fn sender_gas_allowance<T: Triedb>(
...
             return Ok(block.gas_limit);
         }
 
-        let account = triedb_env
-            .get_account(from.into(), block.number)
-            .await
-            .map_err(JsonRpcError::internal_error)?;
+        let balance = match state_overrides.get(&sender) {
+            Some(override_state) => {
+                if override_state.balance.is_some() {
```
andr-dev: Rust match guards to the rescue
```rust
let balance = match state_overrides.get(&sender) {
  Some(override_state) if override_state.balance.is_some() => {
    override_state.balance.unwrap_or_default()
  }
  _ => {
    ...
  }
}
```
or even better
```rust
let balance = if let Some(balance) = state_overrides.get(&sender).map(|override_state| override_state.balance.unwrap_or_default()) {
  balance
} else {
  ...
}
```

maxkozlovsky: let balance = state_overrides.get(&sender).map(|s| s.balance).unwrap_or_else(|| ...);

PR #1428, opened 2025-01-14
 Fix txpool bug for legacy txs without chain_id

andr-dev:


monad-eth-txpool/src/lib.rs
```diff
@@ -90,14 +90,12 @@ where
         block_policy: &EthBlockPolicy,
         account_balance: &Balance,
     ) -> Result<(), TxPoolInsertionError> {
-        let Some(tx_chain_id) = tx.chain_id() else {
-            return Err(TxPoolInsertionError::NotWellFormed);
+        if let Some(tx_chain_id) = tx.chain_id() {
+            if tx_chain_id != block_policy.get_chain_id() {
+                return Err(TxPoolInsertionError::NotWellFormed);
+            }
         };
```
eerkaijun: when we do `static_validate_transaction` in `ValidEthTransaction::validate`, this is already covered - I don't think we need to check here once more

andr-dev: That's why it wasn't in one of my other branches! Good catch.

PR #1430, opened 2025-01-14
Validate OpenRPC document and fix doc issues

philipglazman:
Cleans up openrpc.json generation. Adds a binary to generate openrpc docs using `cargo run --bin monad-rpc-docs`.

### DIFF
--- a/monad-rpc/src/main.rs
+++ b/monad-rpc/src/main.rs
@@ -7,6 +7,8 @@ use actix_web::{
 };
 use clap::Parser;
 use eth_json_types::serialize_result;
+use fee::FixedFee;
+use meta::{monad_net_version, monad_web3_client_version};
 use monad_archive::archive_reader::ArchiveReader;
 use monad_eth_types::BASE_FEE_PER_GAS;
 use monad_ethcall::EthCallExecutor;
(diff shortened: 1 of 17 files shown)


philipglazman:
removed `txpool*` rpc methods. Yes state override is more complicated to add and omitted it in this PR.  I want to revisit each method with more docstrings in a follow up PR in order to keep this diff small.

monad-rpc/src/main.rs
```diff
@@ -492,18 +491,7 @@ async fn rpc_select(
...
-            monad_eth_getProof(triedb_env, params)
-                .await
-                .map(serialize_result)?
-        }
-        "eth_sendTransaction" => Err(JsonRpcError::method_not_supported()),
-        "eth_signTransaction" => Err(JsonRpcError::method_not_supported()),
-        "eth_sign" => Err(JsonRpcError::method_not_supported()),
-        "eth_hashrate" => Err(JsonRpcError::method_not_supported()),
-        "net_version" => serialize_result(app_state.chain_id.to_string()),
+        "net_version" => serialize_result(monad_net_version(app_state.chain_id)),
```
eerkaijun: moand_net_version now returns Result<String> rather than String, I think this will break rpc conformance

monad-rpc/src/main.rs
```diff
@@ -535,7 +523,7 @@ async fn rpc_select(
                 .await
                 .map(serialize_result)?
         }
-        "web3_clientVersion" => serialize_result(WEB3_RPC_CLIENT_VERSION),
+        "web3_clientVersion" => serialize_result(monad_web3_client_version()),
```
eerkaijun: same as above

monad-rpc/src/main.rs
philipglazman: nice catch! thanks

PR #1431, opened 2025-01-14
Add node IDs to dataplane logging

eclark0426:
Mostly to make testground logging make more sense: https://github.com/category-labs/category-internal/issues/916

### DIFF
--- a/monad-testground/src/main.rs
+++ b/monad-testground/src/main.rs
@@ -326,6 +326,7 @@ async fn run<ST, SCT>(
     <SCT as SignatureCollection>::SignatureType: Unpin,
 {
     let state_backend = InMemoryStateInner::genesis(u128::MAX, SeqNum(4));
+    let nodeid = config.executor_config.nodeid;
     let (mut state, init_commands) = make_monad_state(state_backend.clone(), config.state_config);
     let mut executor = make_monad_executor(index, state_backend, config.executor_config);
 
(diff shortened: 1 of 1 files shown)


eclark0426:
Since I haven't really used testground at all, I'm not sure which logging would be most helpful to give the ID alongside; I just put it into all of the `trace()` logs. @bamrith not sure if you might have any suggestions on where logging of the ID is most helpful, since you've worked with the testground more?

bamrith:
I think it's fine to add it to all traces. I feel like this might be redundant anywhere other than testground because I assume each node will have its own log file.

Edited: Actually, one idea to avoid this can be to add the `NodeId` to the top level span [here](https://github.com/category-labs/monad-bft/blob/master/monad-testground/src/main.rs#L354). That way only testground logs have them

eclark0426:
> Actually, one idea to avoid this can be to add the NodeId to the top level span [here](https://github.com/category-labs/monad-bft/blob/master/monad-testground/src/main.rs#L354). That way only testground logs have them

That... that is a much simpler solution than all the fighting I did with Rust to get the type system to be happy

monad-testground/src/main.rs
```diff
@@ -351,7 +352,7 @@ async fn run<ST, SCT>(
         let ledger_len = executor.ledger().get_blocks().len();
         if ledger_len > last_ledger_len {
             last_ledger_len = ledger_len;
-            ledger_span = tracing::info_span!("ledger_span", last_ledger_len);
+            ledger_span = tracing::info_span!("ledger_span", last_ledger_len, nodeid = ?nodeid);
```
bamrith: nit: `nodeid =` isn't required

PR #1433, opened 2025-01-15
Remove CommitInfo, flatten QC

omegablitz:


omegablitz:
I forgot that I need to update flexnet + forkpoint, so flexnet tests will fail here. This is still reviewable, I will make a subsequent commits updating QcInfo and signature_hash in a bit

monad-consensus-types/src/quorum_certificate.rs
```diff
@@ -129,22 +100,24 @@ impl<SCT: SignatureCollection> QuorumCertificate<SCT> {
...
-                },
-            },
+            info: vote,
             signatures: sigs,
-            signature_hash: sig_hash,
+        }
+    }
+
+    /// Returns a committable block_id, if exists
+    pub fn is_committable(&self) -> Option<BlockId> {
```
michael-yxchen: nit: I'd expect `is_committable` to return a boolean. Maybe `get_committable_parent_id`?

monad-blocktree/src/blocktree.rs
```diff
@@ -308,34 +308,49 @@ where
...
+            let qc = block.validated_block.get_qc();
+            if high_commit_qc
+                .as_ref()
+                .is_some_and(|high_commit_qc| high_commit_qc.get_round() >= qc.get_round())
             {
-                high_commit_qc = Some(qc.clone());
+                // we already have observed a higher committable QC
+                //
+                // pretty sure this can never happen because we're BFS from root, so we could
+                // potentially panic here
```
michael-yxchen: nit: I can't immediately come up with a counter example. Your statement is true if highest is in terms of sequence number, which matches BFS layer. Not sure about round

monad-consensus-state/src/lib.rs
```diff
@@ -850,81 +850,82 @@ where
...
+            // the committable block is not (yet) coherent
+            // likely because execution is lagging
+            return cmds;
+        }
 
-            debug!(
-                num_commits = ?blocks_to_commit.len(),
-                "qc triggered commit"
-            );
+        let blocks_to_commit = self.consensus.pending_block_tree.prune(&qc.info.parent_id);
```
michael-yxchen:
```suggestion
        let blocks_to_commit = self.consensus.pending_block_tree.prune(&committable_block_id);
```

monad-consensus-state/src/lib.rs
```diff
@@ -850,81 +850,82 @@ where
...
-                            <= last_committed_block.get_seq_num()
-                    })
-                {
-                    self.consensus.finalized_execution_results.pop_first();
-                }
+        let unvalidated_blocks = blocks_to_commit
+            .iter()
+            .map(|b| b.clone().get_full_block())
+            .collect();
+        cmds.push(ConsensusCommand::LedgerCommit(unvalidated_blocks));
```
michael-yxchen: nit: `unvalidated_block` is probably a legacy name?

monad-state/src/statesync.rs
```diff
@@ -67,40 +67,33 @@ impl<SCT: SignatureCollection> BlockBuffer<SCT> {
...
-                        },
-                        block_header.get_qc().clone(),
-                    ));
-                }
-                break;
-            }
+        let finalized_block_id = proposal_qc.is_committable()?;
+        let finalized_block = self.block_headers.get(&finalized_block_id)?;
+
+        if finalized_block.get_seq_num() <= self.root + self.resync_threshold {
```
michael-yxchen, line 86: nit: don't know if keeping empty block check matters for this commit. It's probably fine

monad-blocktree/src/blocktree.rs
omegablitz: I'll remove the comment to be safe

monad-state/src/statesync.rs
omegablitz: good catch, fixed

PR #1438, opened 2025-01-15
Refactor node config and remove otel tracing

andr-dev:
This change removes the otel tracing context that is no longer used.

Additionally, a minor change was made to the `node.toml` format to avoid duplicating data in the config. Previously, the toml file had something like the following.
```
name = "stressnet_monad_devnet3"
network_name = "stressnet"
```
To remove the network_name duplication, the fields and values have been changed to
```
node_name = "monad_devnet3"
network_name = "stressnet"
```
where the fields are both required (used to be optional). By making this small toml change now, it also allows us to attach these two fields separately in the metrics publisher, removing the need to do some quirky regex matching and parsing in grafana when building our dashboards. Devops is really excited about this change, but I'm leaving it for a later release and starting with just the config change that easily enables it. These changes have been soaking on devnet15 for the past week.

Closes https://github.com/category-labs/category-internal/issues/1129.


ariqchowdhury:
> This change moves the node config to a separate crate
can you clarify why that's necessary?

andr-dev:
> > This change moves the node config to a separate crate
> > can you clarify why that's necessary?

There was a downstream change that I thought would be nicer if the monad-node config was split out but that is no longer necessary. Updated!

PR #1439, opened 2025-01-16
Check sender balance if gas price is provided in eth_call

philipglazman:
Fixes https://github.com/category-labs/category-internal/issues/1119 

Geth (and all other major evm implementations) will check that the sender's balance can cover gas in eth_call if gas price is provided. Otherwise, it will not check the sender balance. 

This returns the correct error code and message when the sender has insufficient balance to cover gas.

### DIFF
--- a/monad-rpc/src/call.rs
+++ b/monad-rpc/src/call.rs
@@ -294,14 +294,13 @@ pub async fn sender_gas_allowance<T: Triedb>(
             }
         };
 
+        if balance == U256::ZERO {
+            return Err(JsonRpcError::insufficient_funds());
+        }
+
         let gas_limit = balance
             .checked_sub(request.value.unwrap_or_default())
-            .ok_or_else(|| {
-                JsonRpcError::eth_call_error(
-                    "insufficient funds for gas * price + value".to_string(),
-                    None,
-                )
-            })?
+            .ok_or_else(JsonRpcError::insufficient_funds)?
             .checked_div(gas_price)
             .ok_or_else(|| JsonRpcError::internal_error("zero gas price".into()))?;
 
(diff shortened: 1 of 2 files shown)


monad-rpc/src/call.rs
```diff
@@ -360,17 +359,31 @@ pub async fn monad_eth_call<T: Triedb + TriedbPath>(
...
+                        header.header.base_fee_per_gas.unwrap_or_default(),
+                    ))?;
+                }
+                _ => {
+                    header.header.base_fee_per_gas = Some(0);
+                }
+            };
 
             if params.transaction.gas.is_none() {
                 let allowance =
```
eerkaijun: do we still need to call `sender_gas_allowance` here in the case where we don't need to check balance (gas price is set)? I think it can just be set as the block gas limit

monad-rpc/src/call.rs
```diff
@@ -360,17 +359,31 @@ pub async fn monad_eth_call<T: Triedb + TriedbPath>(
...
             }
         }
         _ => {
-            params.transaction.fill_gas_prices(U256::from(
-                header.header.base_fee_per_gas.unwrap_or_default(),
-            ))?;
+            // Geth checks that the sender can pay for gas if gas price is populated.
+            // Set the base fee to zero if gas price is not populated.
+            // https://github.com/ethereum/go-ethereum/pull/20783
+            match params.transaction.gas_price_details {
```
eerkaijun: we `match sender` at the top, not sure if we still need that since both zero address and non-zero address should override block base fee. I think we should move `match params.transaction.gas_price_details` to the top match statement instead.

philipglazman: Good point, I'll make that change.

monad-rpc/src/call.rs
philipglazman: `sender_gas_allowance` will return gas limit if gas price is None but we can make it more explicit

PR #1444, opened 2025-01-17
Concurrent logs read from triedb

eerkaijun:
issue: https://github.com/category-labs/category-internal/issues/1130

for a block range of 1000 with ~1k transactions per block, it takes around 15-20s for the request to be fulfilled (was ~3mins before this commit). for a block range of 1000 with almost empty blocks, it takes ~2 seconds. 

I think we can also consider lowering the max block range (currently 1000), since our blocks can be much larger. I would suggest perhaps 200.

### DIFF
--- a/monad-rpc/src/eth_txn_handlers.rs
+++ b/monad-rpc/src/eth_txn_handlers.rs
@@ -1,5 +1,3 @@
-use std::cmp::min;
-
 use alloy_consensus::{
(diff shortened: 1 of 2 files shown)


jhunsaker:
the problem with this code is it queues up many io ops in parallel putting this particular request in front of others whereas a sequential series of requests would allow other io ops to get in the middle

optimization should be done holistically, not just looking at the performance of one operation

furthermore I think this should be fast even sequentially so we need to do more analysis first to fully understand why it's not

PR #1450, opened 2025-01-21
Parameterize chain ID

eerkaijun:


monad-eth-testutil/examples/txgen/run.rs
```diff
@@ -52,10 +52,11 @@ pub async fn run(client: ReqwestClient, config: Config) -> Result<()> {
         refresh_rx,
         rpc_sender,
         &client,
-        U256::from(1e17),
-        U256::from(1e18),
+        U256::from(1e24),
+        U256::from(1e25),
```
eerkaijun: @JoeHowarth this is what is running on stressnet now (due to higher base fee), but i've got no idea how the seeding works. does that sound reasonable?

JoeHowarth: https://github.com/category-labs/monad-bft/blob/fd3f08f7580992cbc15714c9df46a4408c8fdfe9/monad-eth-testutil/examples/txgen/workers/gen_harness.rs#L78-L106

Here is where seeding happens btw

PR #1451, opened 2025-01-21
Add insert/evict/drop metrics to txpool

andr-dev:
As a hacky solution to get txpool metrics while the metrics upgrades are in progress, this changeset pipes through the txpool events metrics to each txpool call so we can get detailed metrics on insertion rate, tx drop reasons, and many more. These metrics have been tested on devnet15 and have been added to a new TxPool grafana dashboard (https://grafana.devcore4.com/d/aea9wehvok9a8f/txpool).

Closes https://github.com/category-labs/category-internal/issues/1127.
Closes https://github.com/category-labs/category-internal/issues/1128.

### DIFF
--- a/monad-eth-txpool/src/lib.rs
+++ b/monad-eth-txpool/src/lib.rs
@@ -8,6 +8,7 @@ use bytes::Bytes;
 use itertools::{Either, Itertools};
 use monad_consensus_types::{
     block::ProposedExecutionInputs,
+    metrics::TxPoolEvents,
     payload::RoundSignature,
     signature_collection::SignatureCollection,
     txpool::{TxPool, TxPoolInsertionError},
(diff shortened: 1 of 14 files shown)


monad-eth-txpool/src/lib.rs
```diff
@@ -179,21 +174,43 @@ where
...
             .zip(raw_txs)
-            .filter_map(|(insertion_result, b)| match insertion_result {
-                Ok(()) => Some(b),
-                Err(_) => None,
+            .filter_map(|(result, b)| {
+                let Some(error) = result.err() else {
+                    return Some(b);
+                };
+
+                match error {
```
eerkaijun, line 206: should we log the tx hash as well? I think that will be helpful in debugging for the near term, perhaps with log level `trace`.

andr-dev: What's the default log level on stressnet/devnet/testnet nodes? @ariqchowdhury and I are just concerned about the volume of logs at high TPS slowing down the node.

omegablitz: trace is fine, won't be logged in prod

monad-eth-txpool/src/tracked/mod.rs
```diff
@@ -122,7 +127,7 @@ where
             return Ok(Vec::new());
         }
 
-        let _timer = DropTimer::start(Duration::ZERO, |elapsed| {
+        DropTimer::start(Duration::ZERO, |elapsed| {
```
omegablitz: If the `let _timer` part is removed, does this still respect the scope of DropTimer?

andr-dev: Aaaah good catch, I think I misunderstood when someone told me to change this in another PR, fixed!

PR #1464, opened 2025-01-24
Optimistic execution

omegablitz:


omegablitz:
Flexnet doesn't pass because of this: https://github.com/category-labs/monad-integration/blob/f455be849299fe7368bfe26a6b77fde82f2de26a/net-lib/monad_flexnet/ledger.py#L48-L128

monad-block-persist/src/lib.rs
```diff
@@ -44,57 +42,52 @@ where
...
+            id_to_prefix(&body_id.0),
+            BLOCKDB_BODY_EXTENSION
+        ));
+        file_path
     }
 }
 
 pub const BLOCKDB_HEADER_EXTENSION: &str = ".header";
-pub const BLOCKDB_PAYLOAD_EXTENSION: &str = ".body";
+pub const BLOCKDB_BODY_EXTENSION: &str = ".body";
```
ariqchowdhury: nit: can these be at the top of the file

monad-block-persist/src/lib.rs
```diff
@@ -44,57 +42,52 @@ where
...
+
+fn id_to_prefix(hash: &Hash) -> String {
```
ariqchowdhury: nit: blockid_to_hexprefix?

monad-bls/src/bls.rs
```diff
@@ -157,8 +157,8 @@ impl BlsPubKey {
             .map_err(BlsError)
     }
 
-    pub fn compress(&self) -> Vec<u8> {
-        self.0.compress().to_vec()
+    pub fn compress(&self) -> [u8; 48] {
```
ariqchowdhury: this is the G1_COMPRESSED_LEN const?

monad-bls/src/bls.rs
```diff
@@ -230,7 +230,7 @@ impl BlsAggregatePubKey {
         Ok(Self::from_pubkey(&pubkey))
     }
 
-    pub fn compress(&self) -> Vec<u8> {
+    pub fn compress(&self) -> [u8; 48] {
```
ariqchowdhury: this is the G1_COMPRESSED_LEN const?

monad-bls/src/bls.rs
michael-yxchen: yeah. We should use the curve independent `PUBKEY_COMPRESSED_LEN` constant

monad-consensus/src/validation/signing.rs
```diff
@@ -240,7 +241,7 @@ where
     }
 }
 
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, RlpEncodable, RlpDecodable)]
```
michael-yxchen, line 245: RlpEncodableWrapper here? just making a note. ofc we won't make changes in this PR

monad-consensus/src/vote_state.rs
```diff
@@ -439,7 +435,7 @@ mod test {
         let v1_invalid = create_vote_message(&certkeys[1], vote_round, false);
         let v2_valid = create_vote_message(&certkeys[2], vote_round, true);
 
-        let vote_idx = HasherType::hash_object(&v0_valid.vote);
+        let vote_idx = v0_valid.vote;
```
michael-yxchen: nit: rename to vote

monad-ledger/src/lib.rs
```diff
@@ -107,144 +141,68 @@ where
...
 
         if self.block_cache_index.len() > self.block_cache_size {
-            let (_, (block_id, payload_id)) = self.block_cache_index.pop_first().expect("nonempty");
-            self.block_header_cache.remove(&block_id);
+            let (evicted_round, (block_id, payload_id)) =
+                self.block_cache_index.pop_first().expect("nonempty");
+            let last_commit_round = self
+                .last_commit
+                .map(|(_, last_commit_round)| last_commit_round);
+            assert!(evicted_round < last_commit_round.unwrap_or(Round(u64::MAX)));
```
michael-yxchen: we had a panic on this line when supermajority is not online for an extended period. maybe remove this assertion?

monad-ledger/src/lib.rs
```diff
@@ -107,144 +141,68 @@ where
...
-                    .as_ref()
-                    .map(|eth_block| (*seqnum, encode_eth_block(eth_block)))
-            })
-            .collect()
+    fn mark_proposed(&mut self, bft_block_id: &[u8; 32]) -> std::io::Result<()> {
+        let mut event: [u8; 33] = [0_u8; 33]; // FIXME
+        event[0] = 0; // FIXME
+        event[1..].copy_from_slice(bft_block_id);
+        self.wal.write_all(&event)?;
+        self.wal.flush()?;
```
michael-yxchen, line 189: I think we want a `wal.sync_all()` instead. `flush` is a nop

monad-secp/src/lib.rs
```diff
@@ -43,7 +32,24 @@ impl certificate_signature::PubKey for PubKey {
...
+
+impl Encodable for PubKey {
+    fn encode(&self, out: &mut dyn alloy_rlp::BufMut) {
+        self.bytes_compressed().encode(out);
+    }
+}
+
+impl Decodable for PubKey {
+    fn decode(buf: &mut &[u8]) -> alloy_rlp::Result<Self> {
+        let b = <[u8; 33]>::decode(buf)?;
```
michael-yxchen: these constants are defined in `secp256k1::constants`

monad-secp/src/secp.rs
```diff
@@ -147,6 +148,23 @@ impl SecpSignature {
...
 
+impl Encodable for SecpSignature {
+    fn encode(&self, out: &mut dyn alloy_rlp::BufMut) {
+        self.serialize().encode(out);
+    }
+}
+
+impl Decodable for SecpSignature {
+    fn decode(buf: &mut &[u8]) -> alloy_rlp::Result<Self> {
+        let raw_bytes = <[u8; 65]>::decode(buf)?;
```
michael-yxchen, line 162: we'll need to define a constant here. It's not provided by the lib

monad-state/src/lib.rs
```diff
@@ -430,6 +413,42 @@ where
...
+                encode_list::<_, dyn Encodable>(&enc, out);
+            }
+            Self::BlockSyncResponse(m) => {
+                let enc: [&dyn Encodable; 3] = [&monad_version, &3u8, &m];
+                encode_list::<_, dyn Encodable>(&enc, out);
+            }
+            Self::ForwardedTx(m) => {
+                let enc: [&dyn Encodable; 3] = [&monad_version, &4u8, &m];
+                encode_list::<_, dyn Encodable>(&enc, out);
+            }
```
michael-yxchen: do we need to wrap raw txn bytes in a struct and give it a message name prefix?

monad-triedb-utils/src/lib.rs
```diff
@@ -73,6 +86,83 @@ impl TriedbReader {
...
+        }
+
+        let (triedb_key, key_len_nibbles) =
+            create_triedb_key(Version::Proposal(*round), KeyInput::BlockHeader);
+        let eth_header_bytes = self.handle.read(&triedb_key, key_len_nibbles, seq_num.0)?;
+
+        let bft_id = self.get_bft_block_id(seq_num, round)?;
+        if block_id != &bft_id {
+            return None;
+        }
```
michael-yxchen, line 110: can we add a comment on why we need to read bft block id again after reading proposed eth header? assuming it's to ensure block isn't overwritten after read block the first time

monad-types/src/lib.rs
```diff
@@ -243,41 +277,47 @@ impl SeqNum {
...
+    Deserialize,
+    RlpEncodableWrapper,
+    RlpDecodableWrapper,
+)]
+pub struct NodeId<P: PubKey>(
     #[serde(serialize_with = "serialize_pubkey::<_, P>")]
     #[serde(deserialize_with = "deserialize_pubkey::<_, P>")]
     #[serde(bound = "P:PubKey")]
     #[serde(rename(serialize = "node_id", deserialize = "node_id"))]
     // Outer struct always flatten this struct, thus renaming to node_id
```
michael-yxchen, line 298: I assume this can be removed now?

monad-consensus/src/validation/signing.rs
omegablitz: added todo

monad-state/src/lib.rs
omegablitz: Perhaps, but I don't want to make a backwards-incompatible change here. I'll leave a comment.

--
Separately, I don't think we need prefixes for any of these. I think we can remove them from everywhere except right before we call `sign()`. Let me know what you think.

monad-types/src/lib.rs
omegablitz: You're probably right, but I'm scared to change it here in case it breaks something else subtly. Leaving a TODO.

monad-ledger/src/lib.rs
omegablitz: don't need stronger guarantees around this particular WAL

PR #1468, opened 2025-01-24
[monad-rpc] add the ExecUpdateBuilder event processor

kjcamann:
Execution events (as provided by the `monad-exec-event` package) are emitted as soon as they are available. In practice this means:

- Most block header metadata is emitted as soon as block execution starts, when the effects of the block on the state are not computed yet

- Transactions are emitted as soon as they are committed to the block, which might happen out-of-order

- Blocks may execute without error but may never be finalized in the consensus protocol (even unvoted blocks may execute)

To simplify the implementation of geth-compatible real-time event processing, this commit introduces a few utilities that polls low-level events and reassembles them into "block update" objects.

Block updates, and the utility objects that produces them, come in two flavors:

1. A `SpeculativeBlockUpdate` reports the executions as soon as they occur, before finality of the block can be established. This enum type contains other variants which announce when finalization of abandonment occurs for the BFT block ID.

2. A `ChainBlockUpdate`, as the name implies, only reports executions which finalize and thus become part of the blockchain. It is a bit easier to work with, for downstream clients who do not want to filter out abandoned blocks.

Each kind of update is produced by a different kind of "block builder" utility.

This commit also adds a command-line binary, `blockwatch`, which prints either a summary line of each block or the full #[derive(Debug)] information for each block. By default it prints speculative updates, but adding the `-f` flag will cause it to only show finalized blocks (others are ignored).

To support `blockwatch`, this commit also adds a `lib.rs` file so that the update builder utilities (which live in `src`) can be re-used outside the main executable crate by secondary tooling binary crates such as `blockwatch`.

This commit also adds a second command-line binary, `event-rtt-check`, which is used to check the correctness of block updates against real data from the blockchain. The testing approach is documented in the module comment at the top of the file.

### DIFF
--- a/monad-rpc/bin/blockwatch.rs
+++ b/monad-rpc/bin/blockwatch.rs
@@ -1,0 +1,220 @@
+//! Utility for printing block update objects (from the exec_update_builder
+//! module) to stdout
+
+use std::{path::PathBuf, process, str::FromStr};
+
+use clap::Parser;
+use monad_event_ring::{
+    event_reader::EventReader,
+    event_ring::*,
+    event_test_util::{open_event_ring_file, OpenedEventRing},
+    exec_event_types_metadata::{EXEC_EVENT_DEFAULT_RING_PATH, EXEC_EVENT_DOMAIN_METADATA},
+};
+use monad_rpc::exec_update_builder::*;
+
+#[derive(clap::ValueEnum, Clone, Default, Debug, serde::Serialize)]
+#[serde(rename_all = "kebab-case")]
+enum DisplayFormat {
+    #[default]
+    Summary,
+    Debug,
+    Json,
+}
+
+#[derive(Parser)]
+#[command(about = "Utility for watching execution block updates")]
+struct Cli {
+    #[arg(
+        short,
+        long,
+        help = "name of the execution event ring shared memory file"
+    )]
+    #[arg(default_value = EXEC_EVENT_DEFAULT_RING_PATH)]
+    event_ring_path: PathBuf,
+
+    #[arg(
+        short,
+        long,
+        default_value = "proposed",
+        help = "consensus state to reach before reporting (proposed, voted, finalized, or verified)"
+    )]
+    filter: String,
+
+    #[arg(
+        short,
+        long,
+        help = "format used to display the updates",
+        default_value = "summary"
+    )]
+    display: DisplayFormat,
+
+    #[arg(short, long, help = "report orphaned consensus states")]
+    orphans: bool,
+
+    #[arg(
+        short,
+        long,
+        help = "transaction input will not be included in block updates",
+        default_value_t = false
+    )]
+    ignore_txn_input: bool,
+}
+
+fn print_failed_block_info(fail_info: Box<FailedBlockInfo>) {
+    let number = fail_info.eth_header.number;
+    let bft_block_id = fail_info.bft_block_id;
+    let error = fail_info.failure_reason;
+    println!("Failure   => {number} [{bft_block_id:#?}] ERROR: {error:#?}");
+}
+
+fn print_executed_block_info(exec_info: Box<ExecutedBlockInfo>) {
+    let number = exec_info.eth_header.number;
+    let bft_block_id = exec_info.bft_block_id;
+    let txn_count = exec_info.txns.len();
+    let gas = exec_info.eth_header.gas_used;
+    let mut log_count: u64 = 0;
+    for ref txn_info_opt in exec_info.txns {
+        let txn_info = txn_info_opt.as_ref().unwrap();
+        log_count += txn_info.receipt.logs.len() as u64;
+    }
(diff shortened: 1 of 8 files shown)


monad-rpc/bin/blockwatch.rs
```diff
@@ -0,0 +1,213 @@
...
+}
+
+fn print_executed_block_info(exec_info: Box<ExecutedBlockInfo>) {
+    let number = exec_info.eth_header.number;
+    let bft_block_id = exec_info.bft_block_id;
+    let txn_count = exec_info.txns.len();
+    let gas = exec_info.eth_header.gas_used;
+    let mut log_count: u64 = 0;
+    let mut bu_count: u64 = 0;
+    let su_count: u64 = 0;
```
philipglazman: we are not using `su_count`

monad-rpc/bin/blockwatch.rs
```diff
@@ -0,0 +1,213 @@
...
+    event_ring_path: PathBuf,
+
+    #[arg(short, long, help = "only display information for finalized blocks")]
+    finalized: bool,
+
+    #[arg(short, long, help = "display all block info instead of a summary")]
+    dump: bool,
+}
+
+fn print_failed_block_info(fail_info: Box<FailedBlockInfo>) {
```
philipglazman, line 63: Rust has a [Display](https://doc.rust-lang.org/std/fmt/trait.Display.html) trait that can simplify all the `print_*` methods.

monad-rpc/src/exec_update_builder.rs
```diff
@@ -0,0 +1,1042 @@
...
+        let mut block_update = update_state.exec_reassembly[block_flow_id].take().unwrap();
+        let r: &exec_event_types::block_result =
+            &unsafe { *(rs.payload_buf.as_ptr() as *const exec_event_types::block_result) };
+        block_update.eth_block_hash = r.hash;
+        block_update.eth_header.parent_hash = r.parent_hash;
+        block_update.eth_header.logs_bloom = Bloom::from(&r.logs_bloom);
+        block_update.eth_header.state_root = r.state_root;
+        block_update.eth_header.transactions_root = r.transactions_root;
+        block_update.eth_header.receipts_root = r.receipts_root;
+        block_update.eth_header.withdrawals_root = Some(r.withdrawals_root);
```
philipglazman: `= None` We set withdrawals_root to `Some(B256::ZERO)`, but can use None instead

monad-rpc/src/exec_update_builder.rs
```diff
@@ -0,0 +1,1042 @@
...
+        gas_limit: event.gas_limit,
+        gas_used: 0,
+        timestamp: event.timestamp,
+        extra_data: Bytes::copy_from_slice(
+            &event.extra_data.as_slice()[..event.extra_data_length as usize],
+        ),
+        mix_hash: event.mix_hash,
+        nonce: B64::from(&event.nonce),
+        base_fee_per_gas: Some(event.base_fee_per_gas.to::<u64>()),
+        withdrawals_root: Some(B256::ZERO),
```
philipglazman: `: None,`

monad-rpc/bin/blockwatch.rs
kjcamann: Yeah I am going to remove this for now

monad-rpc/src/exec_update_builder.rs
```diff
@@ -0,0 +1,1196 @@
...
+        block_update.eth_header.state_root = eth_exec_output.state_root;
+        block_update.eth_header.receipts_root = eth_exec_output.receipts_root;
+        block_update.eth_header.gas_used = eth_exec_output.gas_used;
+
+        // Compute cumulative gas used
+        let mut cumulative_gas_used: u128 = 0;
+        for (tx_num, opt_tx_info) in block_update.txns.iter_mut().enumerate() {
+            let eth_block_hash = block_update.eth_block_hash;
+            if opt_tx_info.is_none() {
+                panic!("transaction {tx_num} in block {eth_block_hash} is corrupt");
```
philipglazman, line 518: Alternatively we can return the error to the caller so that they deal with it

monad-rpc/src/exec_update_builder.rs
```diff
@@ -0,0 +1,1196 @@
...
+        let (block_flow_id, _) = get_flow_ids(&rs.event);
+        if update_state.exec_reassembly[block_flow_id].is_none() {
+            // In this case, we are seeing a BLOCK_END without having seen the
+            // start of it, which happens when we first join in the middle of
+            // execution running, or because of gaps; we just ignore this
+            return;
+        }
+        let block_update = update_state.exec_reassembly[block_flow_id]
+            .as_mut()
+            .unwrap();
```
philipglazman, line 503: Can be written as 

```rust
let Some(mut block_update) = update_state.exec_reassembly[block_flow_id].as_mut() else {
            // In this case, we are seeing a BLOCK_END without having seen the
            // start of it, which happens when we first join in the middle of
            // execution running, or because of gaps; we just ignore this
            return;
        };
```

kjcamann:
This rewritten by Andre a long time ago, closing

PR #1470, opened 2025-01-24
Disable quill in consensus builds

michael-yxchen:


michael-yxchen:
https://github.com/category-labs/monad/pull/1088#issuecomment-2588251123

PR #1471, opened 2025-01-26
Configurable number of RPC worker threads

omegablitz:


philipglazman:
We can use the same cli argument to set the actix-web workers too. Currently it creates as many workers as the number of cores on the host. `.workers(args.rpc_worker_threads)`

ariqchowdhury:
I think you have to set `workers` on the actix `HttpServer` as well

omegablitz:
Yep, forgot to mark as draft

PR #1473, opened 2025-01-27
Mock indexer

eerkaijun:
A mock indexer to collect some rpc performance data in stressnet. Currently, it calls `eth_getBlockByNumber`, `eth_getBlockReceipt`, `debug_traceBlockByNumber`, `eth_getBalances` for all addresses that are touched in the block. 

Credits to @michael-yxchen for writing most of these. Putting this into monad-bft for easier maintenance.


ariqchowdhury:
I don't think we should call this monad-indexer, its not meant to be example code to build an indexer. The purpose of this is to test rpc and we will later want to add other calls just to increase the load, maybe call it rpc_request_gen

ariqchowdhury:
just noting here; the expectation when running this in stressnet is that if this process falls too far behind, it will exit and alert monitoring will watch the container and raise alarm

PR #1475, opened 2025-01-27
Check transaction for base fee

philipglazman:


eerkaijun:
addressed in https://github.com/category-labs/monad-bft/pull/1530

PR #1477, opened 2025-01-28
Triedb schema update

eerkaijun:
sender address and log index is added to the transaction table and receipt table in triedb respectively. updating archive and rpc to match

### DIFF
--- a/monad-archive/src/tx_indexer.rs
+++ b/monad-archive/src/tx_indexer.rs
@@ -1,8 +1,11 @@
-use alloy_rlp::Encodable;
-use eyre::{bail, Result};
-use monad_triedb_utils::triedb_env::ReceiptWithLogIndex;
+use alloy_consensus::{ReceiptEnvelope, TxEnvelope};
+use alloy_primitives::BlockHash;
+use alloy_rlp::{Decodable, Encodable, RlpDecodable, RlpEncodable};
+use eyre::bail;
+use monad_triedb_utils::triedb_env::{ReceiptWithLogIndex, TxEnvelopeWithSender};
+use serde::{Deserialize, Serialize};
 
-use crate::prelude::*;
+use crate::prelude::{BlockDataReader, *};
 
 #[derive(Clone)]
 pub struct TxIndexArchiver<Store = IndexStoreErased> {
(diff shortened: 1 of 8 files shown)


eerkaijun:
four archiver changes: 
1. for blocks stored in s3, instead of storing `Vec<TxEnvelope>`, it now stores `Vec<TxEnvelopeWithSender>` as the block body.
2. for receipts stored in s3, instead of storing `Vec<ReceiptEnvelope>`, it now stores `Vec<ReceiptWithLogIndex>`
3. for TxIndexedData stored in dynamodb, instead of storing `TxEnvelope` and `ReceiptEnvelope`, it now stores `TxEnvelopeWithSender` and `ReceiptWithLogIndex`
4. added block timestamp to `HeaderSubset` field

these are the new types: https://github.com/category-labs/monad-bft/blob/fc28599/monad-triedb-utils/src/triedb_env.rs#L85

cc @JoeHowarth

JoeHowarth:
This pr looks good but needs to implement fallback decoding to fill in the sender + log index when reading from the index. This needs to be implemented for each of the IndexReader impls currently (aka call the same helper for all impls). For this pr, let's just fill in the sender + log index and we can put a sentinel for the block timestamp since the current code is not easy to call s3 from the index reader.
This will change in the following pr/issue:
I'm going to finish up https://github.com/category-labs/category-internal/issues/1151 tonight which will change how index data is stored and fetched. I will rebase on your branch and while I'm doing it, will correctly fetch the block timestamp from s3 / impl block_data_reader

eerkaijun:
dependent on https://github.com/category-labs/monad/pull/1126

copied over some performance data from the above PR
_great speedup for RPC. for a block with 3k transactions:
eth_getBlockByNumber response time decreases from ~550ms to ~80ms
eth_getBlockReceipts response time decreases from ~630ms to ~120ms
eth_getLogs for one block decreases from ~600ms to ~100ms

each signature recovery usually takes around ~140us so that checks out_

closes https://github.com/category-labs/category-internal/issues/1166

do not merge until the execution submodule is merged and joe's archiver change is up

omegablitz:
> each signature recovery usually takes around ~140us so that checks out

sender recovery should be ~40us I think

eerkaijun:
> > each signature recovery usually takes around ~140us so that checks out
> 
> sender recovery should be ~40us I think

`tx.recover_signer()` this helper function by alloy takes around ~140-160us when being run in my peach server

eerkaijun:
tested with joe's archive reader change, able to read old schema from archive correctly. subsequent test plan is
- try upgrading one archiver on stressnet node after the corresponding execution is merged
- test out querying the archive (including querying block data before the upgrade and after the upgrade)

monad-archive/src/tx_indexer.rs
```diff
@@ -78,3 +81,393 @@ impl<Store: IndexStore> TxIndexArchiver<Store> {
...
+    pub trace: Vec<u8>,
+    pub receipt: ReceiptEnvelope,
+    pub header_subset: HeaderSubsetV0,
+}
+
+#[derive(
+    Debug, Clone, PartialEq, Eq, Default, Serialize, Deserialize, RlpEncodable, RlpDecodable,
+)]
+#[rlp(trailing)]
+pub struct HeaderSubsetV0 {
```
eerkaijun, line 98: I guess when we come to mainnet (or when we fully migrate archive data) this can be removed

JoeHowarth: Eventually old schema versions can be removed as long as all old data has been migrated from all replicas. However I wouldn't be surprised that if there is a schema change a few years from now, the old archive data may not all be migrated since it's huge data volume and likely spread across many organizations and replicas - long winded way of saying maintaining decoding for all previous versions is not a bad thing.

monad-triedb-utils/Cargo.toml
```diff
@@ -21,6 +21,7 @@ alloy-primitives = { workspace = true }
 alloy-rlp = { workspace = true }
 futures = { workspace = true }
 hex = { workspace = true }
+serde = { workspace = true, features = ["derive", "rc"] }
```
omegablitz: nit: this needs rc feature?

monad-triedb-utils/src/triedb_env.rs
```diff
@@ -151,20 +165,20 @@ pub trait Triedb {
...
         &self,
         block_num: u64,
-    ) -> impl std::future::Future<Output = Result<Vec<ReceiptEnvelope>, String>> + Send + Sync;
+    ) -> impl std::future::Future<Output = Result<Vec<ReceiptWithLogIndex>, String>> + Send + Sync;
     fn get_transaction(
         &self,
         txn_index: u64,
         block_num: u64,
-    ) -> impl std::future::Future<Output = Result<Option<TxEnvelope>, String>> + Send;
+    ) -> impl std::future::Future<Output = Result<Option<TxEnvelopeWithSender>, String>> + Send;
```
omegablitz, line 177: Did we consider making these return either a sum type of either TxEnvelope or TxEnvelopeWithSender?

I think it makes more sense for the caller to handle the sender recovery. Otherwise it'll be hard to move that out of the caller's tokio task.

omegablitz: Nvm, as @eerkaijun mentioned, I meant this for the archiver and not for the Triedb trait.

monad-triedb-utils/Cargo.toml
eerkaijun: great catch, removed

PR #1483, opened 2025-01-30
Wait for execution results in proposal

mikety:
Only [Delayed execution result](https://github.com/category-labs/monad-bft/commit/537753641b5c5aee4f91c54f17c0209eec43a086)  commit needs a review.

### Description:
Solution:
### Tracking issue:

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -1,4 +1,9 @@
-use std::{collections::BTreeMap, fmt::Debug, marker::PhantomData, time::Duration};
+use std::{
+    collections::{BTreeMap, LinkedList},
+    fmt::Debug,
+    marker::PhantomData,
+    time::Duration,
+};
 
 use monad_blocktree::blocktree::BlockTree;
 use monad_chain_config::{
@@ -77,6 +82,7 @@ where
 
     finalized_execution_results: BTreeMap<SeqNum, EPT::FinalizedHeader>,
     proposed_execution_results: BTreeMap<Round, ProposedExecutionResult<EPT>>,
+    pending_proposals: BTreeMap<SeqNum, LinkedList<BlockId>>,
 }
 
 impl<ST, SCT, EPT, BPT, SBT, CCT, CRT> PartialEq
@@ -165,6 +171,7 @@ where
 
     /// Destination address for proposal payments
     pub beneficiary: &'a [u8; 20],
+
     /// This nodes public NodeId; what other nodes see in the validator set
     pub nodeid: &'a NodeId<SCT::NodeIdPubKey>,
     /// Parameters for consensus algorithm behaviour
@@ -223,7 +230,9 @@ pub enum StateRootAction {
     /// received a majority quorum on the state root, or the state root is
     /// out-of-range. It's ok to insert to the block tree and observe if a QC
     /// forms on the block. But we shouldn't vote on the block
-    Defer,
+    MissingBlocks,
+    /// Execution result not ready
+    MissingER,
 }
 
 impl<ST, SCT, EPT, BPT, SBT, CCT, CRT> ConsensusState<ST, SCT, EPT, BPT, SBT, CCT, CRT>
@@ -291,6 +300,7 @@ where
 
             finalized_execution_results: Default::default(),
             proposed_execution_results: Default::default(),
+            pending_proposals: Default::default(),
         }
     }
 
@@ -356,8 +366,11 @@ where
         &mut self,
         execution_result: ExecutionResult<EPT>,
     ) -> Vec<ConsensusCommand<ST, SCT, EPT, BPT, SBT>> {
-        match execution_result {
+        let mut cmds = Vec::new();
+        let mut ancestor_block_id = self.consensus.pending_block_tree.root().block_id;
+        let expected_proposal_seqnum: u64 = match execution_result {
             ExecutionResult::Finalized(seq_num, execution_result) => {
+                trace!("Received finalized StateRoot for {:?}", seq_num.0);
                 match self.consensus.finalized_execution_results.entry(seq_num) {
                     std::collections::btree_map::Entry::Vacant(entry) => {
                         entry.insert(execution_result);
@@ -366,9 +379,18 @@ where
                         assert_eq!(entry.get(), &execution_result);
                     }
                 };
-                Vec::new()
+                seq_num.0.saturating_add(self.config.execution_delay.0)
             }
             ExecutionResult::Proposed(execution_result) => {
+                let seq_num = execution_result.seq_num;
+                let block_id = execution_result.block_id;
+
+                trace!(
+                    "Received proposed ExecutionResult for seq_num {:?}, round {:?}, pending_block_tree root round {:?}",
+                    seq_num.0,
+                    execution_result.round,
+                    self.consensus.pending_block_tree.root().round
+                );
                 if execution_result.round <= self.consensus.pending_block_tree.root().round {
                     // not necessary, but here for clarity
                     // try_update_coherency would drop it anyways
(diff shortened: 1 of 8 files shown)


monad-consensus-state/src/lib.rs
```diff
@@ -340,8 +350,12 @@ where
         &mut self,
         execution_result: ExecutionResult<EPT>,
     ) -> Vec<ConsensusCommand<ST, SCT, EPT>> {
+        let mut cmds = Vec::new();
+        let mut expected_proposal_seqnum_opt: Option<u64> = None;
+        let mut round: Option<Round> = None;
         match execution_result {
             ExecutionResult::Finalized(seq_num, execution_result) => {
+                trace!("Received finilized StateRoot for {:?}", seq_num.0);
```
michael-yxchen: nit: typo

monad-consensus-state/src/lib.rs
```diff
@@ -350,23 +364,81 @@ where
                         assert_eq!(entry.get(), &execution_result);
                     }
                 };
-                Vec::new()
+                expected_proposal_seqnum_opt =
+                    seq_num.0.checked_add(self.config.execution_delay.0).take();
```
michael-yxchen: It's ok to panic when seq_num addition overflows. Our custom `impl Add for SeqNum` does that

I think `expected_proposal_seqnum_opt` doesn't need to be an Option<_>. We can assign with `seq_num + self.config.execution_delay` instead

monad-consensus-state/src/lib.rs
```diff
@@ -350,23 +364,81 @@ where
...
             }
             ExecutionResult::Proposed(execution_result) => {
-                if execution_result.round <= self.consensus.pending_block_tree.root().round {
+                let seq_num = execution_result.seq_num;
+                let block_id = execution_result.block_id;
+                round = Some(execution_result.round);
+
```
michael-yxchen: I think round.unwrap() makes the code less readable. We can keep the original code and add `maybe_proposal_round = Some(execution_result.round)` in this block

monad-consensus-state/src/lib.rs
```diff
@@ -350,23 +364,81 @@ where
...
-                self.try_update_coherency(block_id)
+                    .insert(round.unwrap(), execution_result);
+
+                cmds.extend(self.try_update_coherency(block_id));
+
+                expected_proposal_seqnum_opt =
+                    seq_num.0.checked_add(self.config.execution_delay.0).take();
+            }
+        }
+        // check if this result is for a proposal N that is waiting on N-K
```
michael-yxchen: looking at this comment and var name, I think you intend to set `expected_proposal_seqnum_opt` to `Some(_)` only for a proposed execution result? We can nest everything in that match block

monad-consensus-state/src/lib.rs
```diff
@@ -350,23 +364,81 @@ where
...
+        if let Some(expected_proposal_seqnum) = expected_proposal_seqnum_opt {
+            if let Some(l) = self
+                .consensus
+                .pending_proposals
+                .get(&SeqNum(expected_proposal_seqnum))
+            {
+                // if there are multiple pick the proposal with the highest round.
+                if let Some(proposal) = l.iter().max_by_key(|p| p.1.block_header.round) {
+                    if round.is_none() || proposal.1.block_header.round == round.unwrap() {
```
michael-yxchen: I don't think this is correct. `round` is `proposed_execution_result.round` but `proposal.1.block_header.round` is the round number of the pending proposal. They are different

The algorithm should be something like, when we receive `proposed_execution_result`, we check if we have any blocks pending that satisfy
1. `block.seq_num == expected_proposal_seq_num` and
1. are children of `proposed_execution_result.block_id` 

Your code covers the first check with `pending_proposals` lookup. It doesn't check children. A proper check will involve the block tree so it might make sense to add helper functions to blocktree, or keep `pending_proposals` in blocktree entirely

monad-consensus-state/src/lib.rs
```diff
@@ -401,8 +472,45 @@ where
...
+                        &p.block_header.seq_num.0,
+                        p.block_header.round.0
+                    );
+                    val.push_back((author, p));
+                } else {
+                    trace!("StateRoot NotReady proposal seq_num {:?} , round {:?} inserted to the list", p.block_header.seq_num.0, p.block_header.round.0);
+                    self.consensus
+                        .pending_proposals
+                        .insert(p.block_header.seq_num, LinkedList::from([(author, p)]));
+                }
```
michael-yxchen: entry api can simplify it here

```
let pending_list = self.consensus.pending_proposals.entry(&p.block_header.seq_num).or_default();
pending_list.push_back((author, p));
```

monad-consensus-state/src/lib.rs
```diff
@@ -1415,16 +1534,33 @@ where
...
-            self.metrics.consensus_events.rx_execution_lagging += 1;
-            return StateRootAction::Defer;
+            let next_seq_num = if let Some(extending_block) = pending_blocktree_blocks.last() {
+                extending_block.get_seq_num() + SeqNum(1)
+            } else {
+                self.consensus.pending_block_tree.root().seq_num + SeqNum(1)
+            };
+
+            if next_seq_num == p.block_header.seq_num
+                && p.block_header.round == self.consensus.pacemaker.get_current_round() + Round(1)
```
michael-yxchen: it's duplicating some logic with `handle_proposal_message`. we should try to dedup

monad-consensus-state/src/lib.rs
```diff
@@ -2956,11 +3092,92 @@ mod test {
 
         let cmds = wrapped_state.handle_proposal_message(author, p);
 
-        assert_eq!(wrapped_state.consensus.get_current_round(), Round(6));
+        assert_eq!(wrapped_state.consensus.get_current_round(), Round(5));
```
michael-yxchen: I don't think this assertion should be changed. Round is bumped whenever there's a valid QC/TC. Stopping timer behavior shouldn't affect the round number

monad-consensus-types/src/metrics.rs
```diff
@@ -96,6 +96,7 @@ metrics!(
             process_qc,
             creating_proposal,
             rx_execution_lagging,
+            rx_execution_lagging_pending,
```
michael-yxchen: we could use a more explicit name here. Maybe include stopped_timer or something?

monad-updaters/src/state_root_hash.rs
```diff
@@ -222,6 +231,14 @@ where
         let this = self.deref_mut();
 
         if !this.enable_updates {
+            if this.waker.is_none() {
+                this.waker = Some(cx.waker().clone());
+            }
+
+            if this.ready() {
+                this.waker.take().unwrap().wake();
+            }
```
michael-yxchen, line 240: Calling wake when we're returning pending is a bit wasteful. Correctness should be fine

monad-mock-swarm/tests/delay_state_root.rs
```diff
@@ -0,0 +1,357 @@
...
+                );
+                return false;
+            }
+        }
+
+        true
+    }
+
+    #[test]
+    fn delay_state_root_block() {
```
michael-yxchen, line 232: can you help me understand why txns are not included in the first 4 blocks but is in the 5th block? I feel I'm missing something here

monad-mock-swarm/tests/delay_state_root.rs
```diff
@@ -0,0 +1,357 @@
...
+        {}
+
+        for nonce in 10..20 {
+            let eth_txn =
+                make_legacy_tx(sender_1_key, BASE_FEE_PER_GAS.into(), GAS_LIMIT, nonce, 10);
+
+            swarm.send_transaction(node_1_id, alloy_rlp::encode(&eth_txn).into());
+
+            expected_txns.push(eth_txn);
+        }
```
michael-yxchen, line 340: I think txns doesn't matter that much here. We should assert that the fast node can't enter new rounds and create proposals while the slow node is catching up

monad-consensus-state/src/lib.rs
mikety: There are many unit tests and mock-swarm tests set `execution_delay` = `SeqNum::MAX`.  
Changed the `saturating_add` to avoid panics

monad-consensus-state/src/lib.rs
mikety: fixed

monad-consensus-state/src/lib.rs
mikety: Is it possible scenario  that pending execution result matching the proposal is  skipped so the system only sees the  finalized? Can it happen with the blocksync?

michael-yxchen: Checked with @omegablitz. looks like there's no guarantee on the ordering, i.e. we might see finalized before proposed. Ignore the comment

monad-consensus-state/src/lib.rs
mikety: Updated - the proposal is added to pending before `try_vote`, after round is bumped.

monad-consensus-types/src/metrics.rs
mikety: also decrement when timer is reset to have an invariant.

monad-mock-swarm/tests/delay_state_root.rs
mikety: execution_delay is 5 blocks.

monad-mock-swarm/tests/delay_state_root.rs
mikety: Ok, Ill add a few more tests.

monad-consensus-state/src/lib.rs
mikety: since the pending proposal blocks will be in a blocktree its ok to skip this check.

monad-consensus-state/src/lib.rs
mikety: Updated: the proposals are now stored as ValidatedBlocks and `try_vote` is called directly, instead of calling `handle_proposal_message`, this makes the calls more local. The proposal blocks are verified to be in the pending blocktree, so the check 2 is not needed.

monad-updaters/src/state_root_hash.rs
mikety: WIll update to wake only on disable to enable updates

PR #1484, opened 2025-01-30
Gas params for eth_estimateGas

eerkaijun:
similar to https://github.com/category-labs/monad-bft/pull/1439 but we also need to do it for eth_estimateGas

test case in https://github.com/category-labs/monad-integration/pull/186

partly addresses: https://github.com/category-labs/category-external/issues/23

### DIFF
--- a/monad-rpc/src/gas_handlers.rs
+++ b/monad-rpc/src/gas_handlers.rs
@@ -14,7 +14,7 @@ use tracing::trace;
 
 use crate::{
     block_handlers::get_block_num_from_tag,
-    call::{sender_gas_allowance, CallRequest},
+    call::{fill_gas_params, CallRequest},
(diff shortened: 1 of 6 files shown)


monad-rpc/src/gas_handlers.rs
```diff
@@ -64,20 +64,37 @@ pub async fn monad_eth_estimateGas<T: Triedb + TriedbPath>(
...
-            &header.header,
-            &params.tx,
-            &params.state_override_set,
-        )
-        .await?;
-        params.tx.gas = Some(U256::from(allowance));
+    // Geth checks that the sender can pay for gas if gas price is populated.
+    // Set the base fee to zero if gas price is not populated.
+    // https://github.com/ethereum/go-ethereum/pull/20783
+    match params.tx.gas_price_details {
```
philipglazman: We can make this a function since its being used in two places now

PR #1488, opened 2025-01-31
Refactor txpool into an executor

andr-dev:


monad-consensus-types/src/block.rs
```diff
@@ -434,6 +435,19 @@ impl MockableFinalizedHeader for MockExecutionFinalizedHeader {
     }
 }
 
+#[derive(Debug)]
+pub enum OptimisticPolicyCommit<ST, SCT, EPT, BPT, SBT>
```
ariqchowdhury, line 445: can you add a comment describing why this exists in addition to the OptimisticCommit?
Probably could use a better name, I don't have a good suggestion yet

monad-eth-txpool/src/lib.rs
```diff
@@ -1,317 +1,196 @@
...
+
+    fn exec(&mut self, commands: Vec<Self::Command>) {
+        for command in commands {
+            match command {
+                TxPoolCommand::BlockCommit(committed_blocks) => {
+                    for committed_block in committed_blocks {
+                        BlockPolicy::<ST, SCT, EthExecutionProtocol, SBT>::update_committed_block(
+                            &mut self.block_policy,
+                            &committed_block,
+                        );
```
michael-yxchen: nit: `self.block_policy.update_committed_block`

monad-eth-txpool/src/lib.rs
```diff
@@ -1,317 +1,196 @@
...
+                        }
                     }
                 }
+                TxPoolCommand::Reset {
+                    last_delay_committed_blocks,
+                } => {
+                    BlockPolicy::<ST, SCT, EthExecutionProtocol, SBT>::reset(
+                        &mut self.block_policy,
+                        last_delay_committed_blocks.iter().collect(),
+                    );
```
michael-yxchen: nit: `self.block_policy.reset`

monad-updaters/src/parent.rs
```diff
@@ -91,36 +99,42 @@ where
...
         let this = self.deref_mut();
 
         futures::future::select_all(vec![
             this.timer.next().boxed_local(),
             this.control_panel.next().boxed_local(),
             this.ledger.next().boxed_local(),
+            this.txpool.next().boxed_local(),
             this.state_root_hash.next().boxed_local(),
             this.timestamp.next().boxed_local(),
             this.loopback.next().boxed_local(),
```
michael-yxchen, line 140: we should revisit event priority at some point. One thing that comes to mind is that loopback is used for sending statesync and blocksync requests. Maybe that should be prioritized?

monad-state/src/lib.rs
```diff
@@ -1191,20 +1204,23 @@ where
             })
             .take(delay.0 as usize)
             .collect();
-        // reset block_policy
+
+        delay_validated_blocks_from_root.reverse();
```
michael-yxchen: we should follow semantic naming for this variable

monad-consensus-types/src/block.rs
andr-dev: Agreed, unhappy with the name, if anyone has a better name please feel free to share :)

monad-eth-txpool/src/lib.rs
andr-dev: For some reason, I couldn't get it to resolve one of the generic types so the compiler isn't happy with this :(

monad-state/src/consensus.rs
```diff
@@ -246,14 +246,7 @@ where
...
         let ConsensusMode::Live(consensus) = self.consensus else {
-            match event {
-                MempoolEvent::Proposal { .. } | MempoolEvent::ForwardTxs(_) => {
-                    unreachable!("handle_mempool_event when not live")
-                }
-                MempoolEvent::UserTxns(_) | MempoolEvent::ForwardedTxns { .. } => {
-                    return Vec::default()
-                }
-            }
+            return Vec::default();
```
andr-dev: If consensus requested a proposal and suddenly needs to statesync, there's no need to crash the node. Same goes for if the mempool decides to try and forward txs during a statesync.

PR #1490, opened 2025-01-31
Move sender recovery to blocking threadpool

JoeHowarth:


omegablitz:
There's no upper bound on the number of threads performing sender recovery now, but it's still a big improvement. We should move this to a rayon pool or something eventually.

eerkaijun:
any concerns with no upper bound on number of threads?

PR #1492, opened 2025-01-31
Fix gas estimate issue that returns block gas limit

eerkaijun:
fixes https://github.com/category-labs/category-external/issues/28 

bug introduced during alloy upgrade 
https://github.com/category-labs/monad-bft/blob/v0.7.0/monad-rpc/src/gas_handlers.rs#L147


ariqchowdhury:
can we add unit tests for this? There's none for this function

eerkaijun:
we will be adding more flexnet tests

PR #1493, opened 2025-02-01
Adds support for eip1898

philipglazman:
Support [eip1898](https://eips.ethereum.org/EIPS/eip-1898)

Users can use blocktag or hash for the following methods:
- eth_getBalance
- eth_getStorageAt
- eth_getTransactionCount
- eth_getCode
- eth_call
- eth_getProof

There is an optional `requireCanonical` field when blockHash is provided, but we choose to silently ignore it since it we only expose the canonical chain. 

fixes https://github.com/category-labs/category-internal/issues/1305

### DIFF
--- a/monad-rpc/src/account_handlers.rs
+++ b/monad-rpc/src/account_handlers.rs
@@ -161,8 +161,7 @@ mod tests {
     use crate::eth_json_types::{BlockTags, Quantity};
 
     #[test]
-    fn eip_1898_eth_get_storage_at() {
-        // Test cases from eip-1898: https://eips.ethereum.org/EIPS/eip-1898
+    fn params_without_eip_1898() {
         let res: MonadEthGetStorageAtParams = serde_json::from_str(
             r#"["0xDeaDbeefdEAdbeefdEadbEEFdeadbeEFdEaDbeeF", "0x0", "latest"]"#,
         )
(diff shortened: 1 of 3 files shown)


monad-rpc/src/eth_json_types.rs
```diff
@@ -414,21 +414,27 @@ impl<'de> Deserialize<'de> for BlockTags {
...
         D: serde::Deserializer<'de>,
     {
         let buf = String::deserialize(deserializer)?;
         BlockTags::from_str(&buf)
-            .map(BlockReference::BlockTags)
-            .or_else(|_| EthHash::from_str(&buf).map(BlockReference::EthHash))
-            .map_err(|e| serde::de::Error::custom(format!("BlockReference parse failed: {e:?}")))
+            .map(BlockTagOrHash::BlockTags)
+            .or_else(|_| EthHash::from_str(&buf).map(BlockTagOrHash::Hash))
+            .map_err(|e| serde::de::Error::custom(format!("BlockTagOrHash parse failed: {e:?}")))
```
eerkaijun: seems like we also need to support the syntax
```
params: [.... , {"blockNumber": "0x0"}]
```

in addition to 
```
params: [.... , "0x0"]
```

philipglazman: Added tests from the eip. Yes, we need to support that syntax. Updated! @eerkaijun

monad-rpc/src/jsonrpc.rs
```diff
@@ -205,6 +205,14 @@ impl JsonRpcError {
         }
     }
 
+    pub fn resource_not_found() -> Self {
+        Self {
+            code: -32001,
+            message: "Resource not found".into(),
+            data: None,
```
ariqchowdhury: is this the same error that geth returns?

monad-rpc/src/jsonrpc.rs
```diff
@@ -205,6 +205,14 @@ impl JsonRpcError {
         }
     }
 
+    pub fn resource_not_found() -> Self {
+        Self {
+            code: -32001,
+            message: "Resource not found".into(),
```
eerkaijun: can this error message be something like "block requested not found" (feels more descriptive)? though the EIP states "resource not found", seems like none of the RPC providers actually use that

```
curl https://eth-mainnet.g.alchemy.com/v2/<API-KEY> -X POST --data '{"jsonrpc":"2.0", "method":"eth_getBalance", "params":["0xDeaDbeefdEAdbeefdEadbEEFdeadbeEFdEaDbeeF", {"blockHash": "0xd4e5674
0f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb9fa3"}],"id":1}'
{"jsonrpc":"2.0","id":1,"error":{"code":-32000,"message":"header for hash not found"}}

 curl https://eth-mainnet.g.alchemy.com/v2/<API-KEY> -X POST --data '{"jsonrpc":"2.0", "method":"eth_getBalance", "params":["0xDeaDbeefdEAdbeefdEadbEEFdeadbeEFdEaDbeeF", {"blockNumber": "0x12300000000"}],"id":1}'
{"jsonrpc":"2.0","id":1,"error":{"code":-32602,"message":"Unknown block number"}}
```

monad-rpc/src/jsonrpc.rs
```diff
@@ -205,6 +205,14 @@ impl JsonRpcError {
         }
     }
 
+    pub fn resource_not_found() -> Self {
+        Self {
+            code: -32001,
```
eerkaijun: match error code?

monad-rpc/src/account_handlers.rs
```diff
@@ -154,3 +154,51 @@ pub async fn monad_eth_getProof<T: Triedb>(
...
     Err(JsonRpcError::method_not_supported())
 }
+
+#[cfg(test)]
+mod tests {
+    use super::MonadEthGetStorageAtParams;
+    use crate::eth_json_types::{BlockTags, Quantity};
+
+    #[test]
+    fn eip_1898_eth_get_storage_at() {
```
ariqchowdhury: can we separate each case out into an indiv unit test?

monad-rpc/src/jsonrpc.rs
philipglazman: Yea, it seems that none of the clients actually follow the eip 🤦

monad-rpc/src/jsonrpc.rs
philipglazman: According to the eip it is, updated to match geth

monad-rpc/src/jsonrpc.rs
philipglazman: updated to match geth. The eip provides a suggested error but looks like no one implements it 🤦

monad-rpc/src/account_handlers.rs
philipglazman: done

monad-rpc/src/jsonrpc.rs
```diff
@@ -205,6 +205,14 @@ impl JsonRpcError {
         }
     }
 
+    pub fn block_not_found() -> Self {
```
mikety, line 208: Just a note that in the[ spec](https://eips.ethereum.org/EIPS/eip-1898) `If the block is not found, the callee SHOULD raise a JSON-RPC error (the recommended error code is -32001: Resource not found).`

mikety: please ignore  as its discussed in the prior reviews

PR #1494, opened 2025-02-02
Archive Index stores references to block-level data instead of inline

JoeHowarth:
Implements: https://github.com/category-labs/category-internal/issues/1151

Note: for simplicity and reliability, this version fetched the full blob from s3, not just the byte offset range. This will be added later as an optimization.

Flexnet test updated to include txgen (needed for decoding / encoding). 
erc20 deployment tx will use the reference format for it's indexed data and the erc20 txs will use inline, so both formats are tested. Reading + decoding is covered by the index-checker 
https://github.com/category-labs/monad-integration/pull/190

For integration tests that uses both references and inline index data, see: 
https://github.com/category-labs/monad-bft/pull/1494/files#diff-1fafb44aae5b71494772ed7a1ca8ba3daf2dec89f944fc1ba81d1176b5c8e52fR645


JoeHowarth:
> lgtm, a few clarification questions to confirm my understanding:
> 
> * during indexing, it tries to store into InlineV1, and if unsuccessful try storing ReferenceV0?
> * during reading from index, similarly it tries to read as InlineV1, then InlineV0, then ReferenceV0?
### Encoding 
It encodes InlineV1 to see the data length. If the data is small enough to fit inline, then it uploads that. If not, then it uploads the reference.

### Decoding 
https://github.com/category-labs/monad-bft/pull/1494/files#diff-61c92949a05da9049d22ed6f50ea517838f6f7dd136bf485481b0a75ae3d6d35L144-L173

For reading from index, it gets the data in the index kv and decodes it using sentinel bytes (same as triedb schema pr). Then there are several conversion methods that take each repr and return the requested subset of data in the current in-memory format. Decoding is only attempted once (unless the sentinel + version bytes are not present)

PR #1498, opened 2025-02-04
Implement triedb-driver side of async traverse.

ned14:
- Adds a new function `triedb_async_traverse()`.

- Extends `callback_func` with a new parameter indicating that traversal has finished (and so any state can be destroyed).


omegablitz:
Folded your changes in here: #1502

ned14:
Closed as folded above

PR #1502, opened 2025-02-05
Triedb Async Traverse

omegablitz:
- Adds a new function `triedb_async_traverse()`.

- Extends `callback_func` with a new parameter indicating that traversal has finished (and so any state can be destroyed).

### DIFF
--- a/monad-triedb-utils/src/triedb_env.rs
+++ b/monad-triedb-utils/src/triedb_env.rs
@@ -1,4 +1,5 @@
 use std::{
+    fmt::Debug,
     path::{Path, PathBuf},
     sync::{
         atomic::{AtomicUsize, Ordering::SeqCst},
@@ -12,7 +13,7 @@ use alloy_consensus::{Header, ReceiptEnvelope, TxEnvelope};
 use alloy_primitives::{keccak256, Address, FixedBytes};
 use alloy_rlp::{Decodable, RlpDecodable, RlpEncodable};
 use futures::channel::oneshot;
-use monad_triedb::TriedbHandle;
+use monad_triedb::{TraverseEntry, TriedbHandle};
 use serde::{Deserialize, Serialize};
 use tracing::{error, warn};
 
@@ -33,11 +34,11 @@ pub type EthBlockHash = [u8; 32];
 enum TriedbRequest {
     SyncRequest(SyncRequest),
     AsyncRequest(AsyncRequest),
+    AsyncTraverseRequest(AsyncTraverseRequest),
 }
 
 enum SyncRequest {
     BlockNumberRequest(BlockNumberRequest),
-    TraverseRequest(TraverseRequest),
 }
 
 struct BlockNumberRequest {
@@ -45,9 +46,9 @@ struct BlockNumberRequest {
     request_sender: oneshot::Sender<u64>,
 }
 
-struct TraverseRequest {
+struct AsyncTraverseRequest {
     // a sender for the polling thread to send the result back to the request handler
-    request_sender: oneshot::Sender<Option<Vec<Vec<u8>>>>,
+    request_sender: oneshot::Sender<Option<Vec<TraverseEntry>>>,
     // triedb_key and key_len_nibbles are used to read items from triedb
     triedb_key: Vec<u8>,
     key_len_nibbles: u8,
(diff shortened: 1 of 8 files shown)


monad-triedb-utils/src/triedb_env.rs
```diff
@@ -537,8 +532,10 @@ impl Triedb for TriedbEnv {
                 Some(rlp_receipts) => {
                     let receipts = rlp_receipts
                         .iter()
-                        .filter_map(|rlp_receipt| {
-                            ReceiptWithLogIndex::decode(&mut rlp_receipt.as_slice())
+                        .enumerate()
```
ariqchowdhury: why enumerate if ignoring the receipt_idx?

omegablitz:
before:
image
after:
image

monad-triedb-utils/src/triedb_env.rs
```diff
@@ -644,8 +639,10 @@ impl Triedb for TriedbEnv {
                 Some(rlp_transactions) => {
                     let signed_transactions = rlp_transactions
                         .iter()
-                        .filter_map(|rlp_transaction| {
-                            TxEnvelopeWithSender::decode(&mut rlp_transaction.as_slice())
+                        .enumerate()
```
ariqchowdhury: same enumerate comment

monad-triedb-utils/src/triedb_env.rs
omegablitz: We should be checking that receipt_idx matches the rlp-decoded rlp_receipt.key, because the receipt keys should be consecutive in triedb.

I didn't do that final part, and left it as a TODO. I was too lazy to figure out what to do with that error, and we didn't do the validation before anyways.

monad-triedb-utils/src/triedb_env.rs
omegablitz: We should be checking that the idx matches the rlp-decoded rlp_transaction.key, because the transaction keys should be consecutive in triedb.

I didn't do that final part, and left it as a TODO. I was too lazy to figure out what to do with that error, and we didn't do the validation before anyways.

monad-triedb-utils/src/triedb_env.rs
eerkaijun: do you mean that the receipts might not be sorted? previously in `traverse_triedb` we already made sure they are sorted

omegablitz: You're right, they were sorted before. Are these always consecutive starting from 0, or can there be gaps?

eerkaijun: it should always be consecutive starting from 0. the response returned by rpc should be sorted

omegablitz: @eerkaijun could you take another look? I made it assert on garbage in DB (invalid rlp, non-consecutive starting from 0), and re-added sorting

eerkaijun: just to be extra safe, @Chen-Yifan can you confirm that the index will always start from 0 with no gaps?

monad-triedb-utils/src/triedb_env.rs
```diff
@@ -623,40 +625,48 @@ impl Triedb for TriedbEnv {
...
+                let mut transactions = rlp_transactions
+                    .into_iter()
+                    .map(|TraverseEntry { key, value }| {
+                        let idx = usize::decode(&mut key.as_slice())?;
+                        let transaction = TxEnvelopeWithSender::decode(&mut value.as_slice())?;
+
+                        Ok((idx, transaction))
+                    })
+                    .collect::<Result<Vec<_>, alloy_rlp::Error>>()
+                    .expect("tx rlp decode failed");
```
eerkaijun: feels a bit worried for the .expect here, though it should always decode correctly, but feels like it's better to `error!(...)` log it rather than crashing if for whatever reason an unsupported transaction is in the db

eerkaijun:
my only comment above, but otherwise looks good to me

monad-triedb-utils/src/triedb_env.rs
```diff
@@ -516,40 +513,54 @@ impl Triedb for TriedbEnv {
...
+                    return Err(format!(
+                        "receipts missing from db, indices={:?}",
+                        receipts.iter().map(|(idx, _)| idx).collect::<Vec<_>>()
+                    ));
                 }
-                None => Ok(vec![]),
-            },
+                Ok(receipts.into_iter().map(|(_, receipt)| receipt).collect())
+            }
+            Ok(None) => {
```
eerkaijun, line 560: just confirming that an empty block will return Some(vec![]) instead of None right?

omegablitz: Not sure, asking someone from execution

PR #1507, opened 2025-02-06
RLP decoding that works with legacy transactions

eerkaijun:
came across an issue where alloy rlp decoding library doesn't work with legacy transactions after our triedb schema change. the reason is that in alloy legacy transactions are encoded as a list, while typed transactions are encoded as a string with prefix. after our triedb schema change, we wrap all transactions with a string (rlp::encode_string2 below), so we need to also have custom encoding/decoding to take into account the string for legacy transactions
```
return rlp::encode_list2(
            rlp::encode_string2(encoded_tx), rlp::encode_address(sender));
```

### DIFF
--- a/monad-triedb-utils/src/triedb_env.rs
+++ b/monad-triedb-utils/src/triedb_env.rs
@@ -11,7 +11,7 @@ use std::{
 
 use alloy_consensus::{Header, ReceiptEnvelope, TxEnvelope};
 use alloy_primitives::{keccak256, Address, FixedBytes};
-use alloy_rlp::{Decodable, RlpDecodable, RlpEncodable};
+use alloy_rlp::{encode_list, BytesMut, Decodable, Encodable};
 use futures::channel::oneshot;
 use monad_triedb::{TraverseEntry, TriedbHandle};
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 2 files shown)


ariqchowdhury:
Is the writing side supposed to wrap everything with a string or should it consider legacy vs non-legacy txns?

eerkaijun:
you mean in triedb side? yes, triedb wraps both legacy and non-legacy txns in a string

eerkaijun:
Closes https://github.com/category-labs/category-internal/issues/1270

michael-yxchen:
are the archive index stored with the old encoding? do we need to re-run the archive index job from genesis?

eerkaijun:
no we don't need to rerun archive index job from genesis, the archive reader supports reading both old and new archive data

monad-triedb-utils/src/triedb_env.rs
```diff
@@ -86,19 +87,109 @@ pub struct TransactionLocation {
...
+        .encode(&mut encoded_tx);
+        self.tx.encode_2718(&mut encoded_tx);
+        self.sender.encode(&mut encoded_tx);
+
+        alloy_rlp::Header {
+            list: true,
+            payload_length: encoded_tx.len(),
+        }
+        .encode(out);
+        out.put_slice(&encoded_tx);
```
michael-yxchen:
```
let mut encoded_tx: BytesMut = BytesMut::new();
self.tx.encode(&mut encoded_tx);
let enc: [&dyn Encodable; 2] = [&encoded_tx, &self.sender];
encode_list::<_, dyn Encodable>(&enc, out);
```

nit: maybe more copies, but I feel is much easier to read

eerkaijun: sure, updated

PR #1509, opened 2025-02-06
Archive Index stores references to block-level data instead of inline

JoeHowarth:
CI seems stuck on https://github.com/category-labs/monad-bft/pull/1494 so making a new branch/pr to see if that fixes it


JoeHowarth:
implements https://github.com/category-labs/category-internal/issues/1151

PR #1511, opened 2025-02-07
Decode empty receipt list

eerkaijun:
came across a bug in archive reader where we return an error instead of an empty array when there are no receipts in the block. 

should be packaged into the v0.7.2 release

### DIFF
--- a/monad-archive/src/block_data_archive.rs
+++ b/monad-archive/src/block_data_archive.rs
@@ -2,7 +2,7 @@ use core::str;
 
 use alloy_consensus::{Block as AlloyBlock, BlockBody, Header, ReceiptEnvelope, TxEnvelope};
 use alloy_primitives::BlockHash;
-use alloy_rlp::{Decodable, Encodable};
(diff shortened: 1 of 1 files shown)


monad-archive/src/block_data_archive.rs
```diff
@@ -13,6 +13,7 @@ use crate::{prelude::*, rlp_offset_scanner::get_all_tx_offsets, storage::BlockDa
 pub type Block = AlloyBlock<TxEnvelopeWithSender, Header>;
 
 const BLOCK_PADDING_WIDTH: usize = 12;
+const EMPTY_RLP_LIST: u8 = 0xc0;
```
ariqchowdhury: should we use the const from the `alloy_rlp` lib as we're using that for all the rlp stuff?
https://github.com/alloy-rs/rlp/blob/main/crates/rlp/src/lib.rs#L42-L43

eerkaijun: sure, makes sense to use theirs. updated

PR #1512, opened 2025-02-07
Poll triedb more aggressively

omegablitz:


ariqchowdhury:
is this actually better?

omegablitz:
Not sure, this just matches what's done in RWdb here: https://github.com/category-labs/monad/blob/main/libs/db/src/monad/mpt/db.cpp#L627-L641

cc @Chen-Yifan

Chen-Yifan:
Did some benchmark for async reads, async traverse, mixed async read and traverse. 
TLDR: 
- no results show any benefit from queuing up requests before polling. 
- throttle traverse requests to 30 req/sec gives decent performance. 40 req/sec is the max our system can handle. 
- We don't get any benefits from doing traverses concurrently, latency or throughput wise. The throughput of doing one traverse at a time is the same as concurrent traverse because individual traverses are already maxing out io concurrency. 
- observe very different read latency on different ratio of mixed read and traverse, higher read latency when there are more traverses.

### Async Traverse
#### Test setup:
Writer thread upsert to version local table (similar to transaction/receipt tables) of 3000 keys every 0.5 second. 
Reader thread runs this loop, `num_traverses_before_poll` is a user-configurable parameter:
Ran tests for 60 seconds
```
while !stop_signal:
      for i in range(num_traverse_before_poll):
            triedb_async_traverse(random_version, prefix);
      triedb_poll(nonblocking, max);
```
#### Results:
One traverse at a time: `average traverse duration 26.667 ms`

With no rate limiting, `num_traverse_before_poll = 1`, this is the max throughput our system can handle for traverse request only.
```
Traverse is 45.0356 ops/sec. Average traverse duration 1055.97 ms
```
No rate limiting, `num_traverse_before_poll = 10`: similar throughput but bad latency
```
traverse is 43.651 ops/sec. Average traverse duration 46608 ms
```
Adding rate limiting to the call `triedb_async_traverse()`.
```
--traverse-per-second 50 --num-traverse-before-poll 1
Traverse is 37.0959 ops/sec. Average traverse duration 43.3114 ms
```
```
--traverse-per-second 50 --num-traverse-before-poll 10
Traverse is 39.703 ops/sec. Average traverse duration 372.479 ms
```
Takeaways: 
- lower latency in traverse when we poll every time we issue an async traverse request. 
- lower request latency when we rate limit traverse.

### Async Read
Similar test setup as async traverse, random read from version local table of 3000 keys.
```
--num-reads-before-poll 1
 Read is 32001.1 ops/sec. Average read latency 34.3257 us

--num-reads-before-poll 10
 Read is 32004.9 ops/sec. Average read latency 64.6962 us

--num-reads-before-poll 100
 Read is 31989 ops/sec. Average read latency 443.782 us
```

Reading from table of 1M keys (similar to getAccount), similar test setup as async traverse.
```
--num-reads-before-poll 1
 Read is 101522 ops/sec. Average read latency 700.377 us

--num-reads-before-poll 10
 Read is 101091 ops/sec. Average read latency 773.679 us

--num-reads-before-poll 100
 Read is 103277 ops/sec. Average read latency 1163.28 us
```

In summary, above results show no significant benefits but larger read latency from submitting more requests before poll.

---
### Mixed Read and Traverse
Test setup: 
- upsert every 0.5 second, each version has a table of 3000 keys. reader thread performs random read and traverse from a randomly selected version.
- one poll after each async request as previous tests demonstrated better performance, and also poll whenever free cycles in the loop.
- rate limit to 40 traverse / sec (Also observed lower latency when we limit request rate than no rate limiting)

```
--read-traverse-ratio 1
 Traverse is 39.4378 ops/sec. Average traverse duration 51.1456 ms
 Read is 39.5049 ops/sec. Average read latency 16793 us

--read-traverse-ratio 10
 Traverse is 39.3073 ops/sec. Average traverse duration 51.6525 ms
 Read is 393.275 ops/sec. Average read latency 9662.07 us

--read-traverse-ratio 50
 Traverse is 38.1476 ops/sec. Average traverse duration 45.9387 ms
 Read is 1907.28 ops/sec. Average read latency 5059.41 us

--read-traverse-ratio 100
 Traverse is 36.6033 ops/sec. Average traverse duration 41.2652 ms
 Read is 3660.2 ops/sec. Average read latency 3280.24 us

--read-traverse-ratio 200
 Traverse is 33.7641 ops/sec. Average traverse duration 33.7705 ms
 Read is 6752.79 ops/sec. Average read latency 1745.94 us
```
#### Takeaways
1. see long read latency when read traverse ratio is low
2. see quite robust traverse throughput against read-traverse-ratio. It remains stable around 38-40 ops/sec when read-traverse-ratio <= 100, and gets eaten up a little when that ratio goes up to 200. 
3. throttling traverse requests give better performance than without, throttle it to 30 ops/sec is a safe rate to start with.

PR #1513, opened 2025-02-08
RPC websocket support

philipglazman:
Adds `eth_subscribe` and `eth_unsubscribe` for topics newHeads and logs. 

At a high level, we use an actorless actix-web implementation to keep a long-running websocket server to poll for execution event block updates and push messages to sessions. 

Users can expect to receive notifications for new _finalized_ blocks and logs (w/ or w/o filters). The filtering is done server-side for log events.  

Flexnet tests (wip): https://github.com/category-labs/monad-integration/pull/204

### DIFF
--- a/monad-rpc/src/cli.rs
+++ b/monad-rpc/src/cli.rs
@@ -33,6 +33,14 @@ pub struct Cli {
     #[arg(long, default_value_t = 8081)]
     pub ws_port: u16,
 
+    /// Set the max number of concurrent connections for the WebSocket server
+    #[arg(long, default_value_t = 20_000)]
+    pub ws_max_connections: usize,
+
+    /// Set the max number of concurrent subscriptions per connection for the WebSocket server
+    #[arg(long, default_value_t = 1000)]
+    pub ws_max_subscriptions_per_connection: usize,
+
     /// Set the max number of requests in a batch request
     #[arg(long, default_value_t = 5000)]
     pub batch_request_limit: u16,
--- a/monad-rpc/src/main.rs
+++ b/monad-rpc/src/main.rs
@@ -915,7 +915,12 @@ async fn main() -> std::io::Result<()> {
         let (ws_tx, ws_rx) = flume::bounded::<PollResult>(10000);
         let (ws_tx_cmd, ws_rx_cmd) = flume::bounded::<WebSocketServerCommand>(1000);
 
-        let ws_server = WebSocketServer::new(ws_rx_cmd, ws_rx, 10_000);
+        let ws_server = WebSocketServer::new(
+            ws_rx_cmd,
+            ws_rx,
+            args.ws_max_connections,
+            args.ws_max_subscriptions_per_connection,
+        );
         tokio::spawn(async move {
             ws_server.run().await;
         });
--- a/monad-rpc/src/websocket.rs
+++ b/monad-rpc/src/websocket.rs
@@ -22,7 +22,7 @@ use crate::{
         SubscriptionKind,
     },
     jsonrpc::{JsonRpcError, Request, RequestWrapper},
-    websocket_server::{SessionId, SubscriptionId, WebSocketServerCommand},
+    websocket_server::{SessionId, SubscriptionId, WebSocketServerCommand, WebSocketServerError},
 };
 
 const HEARTBEAT_INTERVAL: Duration = Duration::from_secs(5);
@@ -46,7 +46,7 @@ impl WebSocketServerHandle {
     async fn connect(
         &self,
         conn_tx: mpsc::UnboundedSender<WebSocketSessionCommand>,
-    ) -> Option<SessionId> {
+    ) -> Result<SessionId, WebSocketServerError> {
         let (res_tx, res_rx) = oneshot::channel();
         if let Err(err) = self
             .cmd_tx
@@ -54,7 +54,7 @@ impl WebSocketServerHandle {
         {
             warn!("WebSocketServerHandle connect send error: {err:?}");
         }
-        res_rx.await.unwrap_or(None)
+        res_rx.await.unwrap_or(Err(WebSocketServerError::InternalError))
     }
 
     async fn add_subscription(
@@ -62,7 +62,7 @@ impl WebSocketServerHandle {
         conn_id: SessionId,
         kind: SubscriptionKind,
         filter: Option<Filter>,
-    ) -> Option<SubscriptionId> {
+    ) -> Result<SubscriptionId, WebSocketServerError> {
         let (tx, rx) = tokio::sync::oneshot::channel();
         if let Err(err) = self.cmd_tx.send(WebSocketServerCommand::AddSubscription {
             conn_id,
@@ -72,7 +72,7 @@ impl WebSocketServerHandle {
         }) {
             warn!("WebSocketServerHandle add_subscription send error: {err:?}");
         }
-        rx.await.unwrap_or(None)
+        rx.await.unwrap_or(Err(WebSocketServerError::InternalError))
     }
 
     async fn remove_subscription(&self, conn_id: SessionId, id: SubscriptionId) -> bool {
@@ -123,8 +123,8 @@ async fn handler(
 
     // Add the session to the server
     let conn_id = match server.connect(conn_tx).await {
-        Some(id) => id,
-        None => {
+        Ok(id) => id,
+        Err(_) => {
             if let Err(err) = session
                 .close(Some(CloseReason {
                     code: ws::CloseCode::Error,
@@ -170,6 +170,7 @@ async fn handler(
                         last_heartbeat = Instant::now();
                     }
                     Some(Ok(AggregatedMessage::Text(body))) => {
+                        last_heartbeat = Instant::now();
                         let request = to_request::<Request>(&body);
                         match request {
                             Ok(req) => {
@@ -185,6 +186,7 @@ async fn handler(
                         };
                     }
                     Some(Ok(AggregatedMessage::Binary(body))) => {
+                        last_heartbeat = Instant::now();
                         let request = to_request::<Request>(&body);
                         match request {
                             Ok(req) => {
@@ -203,7 +205,8 @@ async fn handler(
                         server.remove_session(conn_id).await;
                         break reason;
                     }
-                    Some(Err(_err)) => {
+                    Some(Err(err)) => {
+                        error!("error in websocket stream: {:?}", err);
                         server.remove_session(conn_id).await;
                         break None;
                     }
@@ -234,7 +237,11 @@ async fn handler(
                         }
                     }
                     Some(WebSocketSessionCommand::Disconnect {}) => {
-                        break None;
+                        server.remove_session(conn_id).await;
+                        break Some(CloseReason {
+                            code: ws::CloseCode::Away,
+                            description: Some("server disconnected stream".to_string()),
+                        });
                     }
                     None => {
                         break None
@@ -257,7 +264,7 @@ async fn handle_request(
 ) {
     match request.method.as_str() {
         "eth_subscribe" => {
-            let req: EthSubscribeRequest = match serde_json::from_value(request.params.clone()) {
+            let req: EthSubscribeRequest = match serde_json::from_value(request.params) {
                 Ok(params) => params,
                 Err(_) => {
                     if let Err(err) = ctx
@@ -296,7 +303,7 @@ async fn handle_request(
 
             let sub_id = server.add_subscription(*conn_id, req.kind, filter).await;
             match sub_id {
-                Some(id) => {
+                Ok(id) => {
                     if let Err(err) = ctx
                         .text(to_response(&crate::jsonrpc::Response::from_result(
                             request.id,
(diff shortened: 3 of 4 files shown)


monad-rpc/src/websocket_server.rs
```diff
@@ -0,0 +1,510 @@
...
+                };
+
+                let filtered_logs: Vec<alloy_primitives::Log> =
+                    receipt.logs().into_iter().cloned().collect();
+
+                let rpc_logs = filtered_logs
+                    .iter()
+                    .map(|log| {
+                        let rpc_log = alloy_rpc_types::Log {
+                            inner: log.clone(),
```
mikety: There is 3 clone() calls on a hot path for the same data: line 307, 315, and 321.

monad-rpc/src/websocket.rs
```diff
@@ -1,149 +1,472 @@
...
-                actor.server.do_send(Disconnect {});
-
-                ctx.stop();
-                return;
-            }
+impl WebSocketServerHandle {
+    async fn connect(
+        &self,
+        conn_tx: mpsc::UnboundedSender<WebSocketSessionCommand>,
+    ) -> Option<SessionId> {
```
mikety: nit: returning Result<> will allow to return a specific error. (there are at least two: "cant accept connections" and "max # of sessions _num_ reached")

monad-rpc/src/websocket.rs
```diff
@@ -1,149 +1,472 @@
...
+        res_rx.await.unwrap_or(None)
+    }
 
-            ctx.ping(b"");
+    async fn add_subscription(
+        &self,
+        conn_id: SessionId,
+        kind: SubscriptionKind,
+        filter: Option<Filter>,
+    ) -> Option<SubscriptionId> {
```
mikety: ditto, nit: returning Result<> will allow to return a specific error

monad-rpc/src/websocket_server.rs
```diff
@@ -0,0 +1,510 @@
...
+        EthSubscribeResult, FixedData, SpeculativeLog, SpeculativeNewHead, SubscriptionKind,
+        SubscriptionResult,
+    },
+    exec_update_builder::{
+        BlockConsensusState, BlockPollResult, BlockUpdate, EventStreamError, ExecutedBlockInfo,
+    },
+    websocket::WebSocketSessionCommand,
+};
+
+const MAX_SUBSCRIPTIONS: usize = 50;
```
mikety: nit: should this be configurable? afaik geth does not have a hard limit of subscriptions.

monad-rpc/src/websocket_server.rs
```diff
@@ -0,0 +1,510 @@
...
+
+        // Publish the block and logs to all subscriptions
+        self.sessions.iter().for_each(|(_, session)| {
+            session.subscriptions.iter().for_each(|(id, info)| {
+                let subscription_msg = PublishMessage {
+                    id: *id,
+                    kind: info.kind.clone(),
+                    block_header: block.header.clone(),
+                    logs: logs.clone(),
+                    filter: info.filter.clone(),
```
mikety: The implemented logic, to  clone the entire message and then check if it passes the subscription filter  in` publish_subscription` can be made more efficient by making all checks before creating a message.

monad-rpc/src/websocket_server.rs
```diff
@@ -0,0 +1,510 @@
...
+                    commit_state: commit_state.clone(),
+                    block_id: block_update.bft_block_id,
+                };
+
+                self.publish_subscription(subscription_msg, info.conn_id);
```
mikety: the processing of block update can be done on a session level since the session holds the conn_tx the mpsc sender.

monad-rpc/src/websocket_server.rs
```diff
@@ -0,0 +1,510 @@
...
+            }
+        };
+
+        match msg.kind {
+            SubscriptionKind::NewHeads
+                if matches!(msg.commit_state, BlockConsensusState::Finalized) =>
+            {
+                let body = EthSubscribeResult {
+                    subscription: msg.id.0,
+                    result: SubscriptionResult::NewHeads(msg.block_header),
```
mikety, line 484: Here and below unless its Logs or MonadLogs that process filters messages are same for all sessions/subscriptions and differ only by subscription_id and channel they sent to the subscriber. It leaves a room for improvement by not replicating computations for all subscriptions for all sessions.

monad-rpc/src/websocket_server.rs
```diff
@@ -0,0 +1,510 @@
...
+        &mut self,
+        conn_id: SessionId,
+        conn_tx: mpsc::UnboundedSender<WebSocketSessionCommand>,
+    ) -> bool {
+        if self.sessions.len() >= self.max_sessions {
+            warn!("max sessions reached, not adding new session");
+            return false;
+        };
+
+        self.sessions.insert(
```
mikety: As is it may quietly update a sessions with the same Id which can lead to many issues. Its better to confirm the result is `None`.

monad-rpc/src/websocket.rs
```diff
@@ -1,149 +1,472 @@
...
+                            if let Ok(body) = serde_json::to_value(&msg) {
+                                let update = crate::jsonrpc::Notification::new(
+                                    "eth_subscription".to_string(),
+                                    body,
+                                );
+                                let _ = session.text(to_response(&update)).await;
+                            };
+                        }
+                    }
+                    Some(WebSocketSessionCommand::Disconnect {}) => {
```
mikety, line 239: Should be implemented to make `disconnect_all` work

monad-rpc/src/websocket_server.rs
philipglazman: Since we will have many sessions for every block, I wanted to "process" the block update once instead of for every session.

monad-rpc/src/cli.rs
```diff
@@ -21,6 +21,14 @@ pub struct Cli {
     #[arg(long, default_value_t = 8080)]
     pub rpc_port: u16,
 
+    /// Enable the WebSocket server
+    #[arg(long, default_value_t = false)]
+    pub ws_enabled: bool,
+
+    /// Set the port number for the WebSocket server
+    #[arg(long, default_value_t = 8081)]
```
eclark0426, line 33: Is having websockets on a different port typical for RPC servers? In my experience with (non-blockchain and non-JSON RPC) servers that serve both HTTP requests and WebSockets, they both use the same port

monad-rpc/src/main.rs
```diff
@@ -889,41 +892,163 @@ async fn main() -> std::io::Result<()> {
...
+    F: futures::Future<Output = Result<T, E>>,
+    E: std::fmt::Display,
+{
+    let duration = std::time::Duration::from_secs(2);
+    let mut retries = 1;
+
+    loop {
+        match attempt().await {
+            Ok(t) => return Ok(t),
+            Err(e) if retries <= 3 => {
```
eclark0426, line 1063: nit we might want to make 3 retries a constant in case we also add retries somewhere else or decide to change the number

monad-rpc/src/websocket.rs
```diff
@@ -1,149 +1,472 @@
...
+            msg = msg_stream.next() => {
+                match msg {
+                    Some(Ok(AggregatedMessage::Ping(bytes))) => {
+                        last_heartbeat = Instant::now();
+                        let _ = session.pong(&bytes).await;
+                    }
+                    Some(Ok(AggregatedMessage::Pong(_))) => {
+                        last_heartbeat = Instant::now();
+                    }
+                    Some(Ok(AggregatedMessage::Text(body))) => {
```
eclark0426, line 172: Should this also update `last_heartbeat`?

monad-rpc/src/websocket.rs
```diff
@@ -1,149 +1,472 @@
...
+                                handle_request(&mut session, &server, &conn_id, req).await;
+                            }
+                            Err(e) => {
+                                let _ = session
+                                    .text(to_response(&crate::jsonrpc::Response::from_error(e)))
+                                    .await;
+                            }
+                        };
+                    }
+                    Some(Ok(AggregatedMessage::Binary(body))) => {
```
eclark0426, line 188: ditto

monad-rpc/src/cli.rs
philipglazman: Yes, it is typical for blockchain nodes to server http requests and websockets using a different port. Similar to [Geth](https://geth.ethereum.org/docs/interacting-with-geth/rpc#websockets-server)

monad-rpc/src/websocket.rs
```diff
@@ -1,149 +1,554 @@
...
+
+    let mut msg_stream = pin!(msg_stream);
+    let mut server_cmds = pin!(conn_rx);
+
+    let close_reason = loop {
+        tokio::select! {
+            _ = interval.tick() => {
+                if Instant::now().duration_since(last_heartbeat) > CLIENT_TIMEOUT {
+                    server.remove_session(conn_id).await;
+                    break None;
```
dshulyak, line 154: maybe just personal preference, but extracting run loop into a separate function with returns would be easier to understand control flow

monad-rpc/src/websocket.rs
```diff
@@ -1,149 +1,554 @@
...
-                    WebsocketMessage::Nop => {}
                 }
             }
-            Err(e) => {
-                debug!("StreamHandler::handle error {:?}", e);
+            cmd = server_cmds.recv() => {
+                match cmd {
+                    Some(WebSocketSessionCommand::PublishMessage { messages }) => {
+                        for msg in messages {
+                            if let Ok(body) = serde_json::to_value(&msg) {
```
dshulyak: there is a lot of copying, it first allocates in to_value and then immediately after that allocates vec in to_response which is only then is copied into ws buffer. there should be a way to serialize into ws buffer directly

monad-rpc/src/websocket.rs
```diff
@@ -1,149 +1,554 @@
...
+                    if update_count == 100 {
+                        break;
+                    }
+                }
+                _ => break,
+            }
+        }
+
+        async move {
+            // sleep for 1 sec
```
dshulyak: i didn't try to run it, but it looks like it will sleep for 100s ? what would it take to implement a better synchronization?

monad-rpc/src/websocket.rs
```diff
@@ -1,149 +1,554 @@
...
+                                    .binary(to_response(&crate::jsonrpc::Response::from_error(e)))
+                                    .await;
+                            }
+                        };
+                    }
+                    Some(Ok(AggregatedMessage::Close(reason))) => {
+                        server.remove_session(conn_id).await;
+                        break reason;
+                    }
+                    Some(Err(_err)) => {
```
dshulyak: maybe debug log it

monad-rpc/src/websocket.rs
```diff
@@ -1,149 +1,554 @@
...
+                                );
+                                let _ = session.text(to_response(&update)).await;
+                            }
+                        }
+                    }
+                    Some(WebSocketSessionCommand::Disconnect {}) => {
+                        break Some(CloseReason {
+                            code: ws::CloseCode::Away,
+                            description: Some("server disconnected stream".to_string()),
+                        });
```
dshulyak, line 244: i noticed that server.remove_session can be extracted to be called after run loop exits, but this is the only case when it is not called. it is expected that server already dropped the session?

monad-rpc/src/websocket.rs
```diff
@@ -1,149 +1,554 @@
...
+
+async fn handle_request(
+    ctx: &mut actix_ws::Session,
+    server: &WebSocketServerHandle,
+    conn_id: &SessionId,
+    request: Request,
+) {
+    match request.method.as_str() {
+        "eth_subscribe" => {
+            let req: EthSubscribeRequest = match serde_json::from_value(request.params.clone()) {
```
dshulyak: does it need a clone here?

monad-rpc/src/websocket.rs
```diff
@@ -1,149 +1,554 @@
...
+                                "cannot subscribe to topic".to_string(),
+                            )),
+                            request.id,
+                        )))
+                        .await;
+                }
             }
         }
+        "eth_unsubscribe" => {
+            let params: EthUnsubscribeRequest = match serde_json::from_value(request.params.clone())
```
dshulyak: does it need a clone?

monad-rpc/src/websocket.rs
philipglazman: This sleep isn't necessary. removing it

monad-rpc/src/websocket.rs
philipglazman: Good point, it should be called

monad-rpc/src/websocket_server.rs
philipglazman: Yep good callout. added cli args to make # of connections and # of subscriptions per connection configurable.

monad-rpc/src/websocket.rs
philipglazman: updated

philipglazman:
updated pr: https://github.com/category-labs/monad-bft/pull/1802

PR #1517, opened 2025-02-10
Refactor ipc executor into txpool

andr-dev:


monad-eth-txpool/src/executor/mod.rs
```diff
@@ -183,10 +175,37 @@ where
...
+            let num_valid_txs = valid_encoded_txs.len();
+
+            metrics.insert_mempool_txs += num_valid_txs as u64;
+
+            return Poll::Ready(Some(MonadEvent::MempoolEvent(MempoolEvent::ForwardTxs(
+                valid_encoded_txs,
+            ))));
+        }
+
+        events.poll_recv(cx).map(|event| {
```
michael-yxchen: This is fine given we never send txns to validators on prod net. Otherwise I think proposal should have higher priority than ipc txns

monad-mock-swarm/src/mock.rs
```diff
@@ -215,8 +214,15 @@ impl<S: SwarmRelation> MockExecutor<S> {
...
+    pub fn send_transaction(
+        &mut self,
+        node_id: ID<CertificateSignaturePubKey<S::SignatureType>>,
+        tx: Bytes,
+    ) {
+        self.txpool.exec(vec![TxPoolCommand::InsertForwardedTxs {
+            sender: *node_id.get_peer_id(),
+            txs: vec![tx],
+        }]);
     }
```
michael-yxchen: I think this is different from previous behavior. Forwarded txs are not sent to future leaders

monad-eth-txpool/src/executor/mod.rs
andr-dev: Aaah I fixed this in my next PR, moving to here thanks!

omegablitz:
don't need to fix this now, but if it's a quick fix we can fold it in: https://github.com/category-labs/monad-bft/blob/9dede2f8c973d983cc77b3147521174a629b8e68/monad-state/src/lib.rs#L817

monad-executor-glue/src/lib.rs
```diff
@@ -290,10 +290,9 @@ where
...
 
-    // TODO(andr-dev): This variant should be removed once IPC is refactored into the txpool
-    InsertTxs {
+    InsertForwardedTxs {
+        sender: NodeId<SCT::NodeIdPubKey>,
         txs: Vec<Bytes>,
-        owned: bool,
     },
 
     // Emitted after blocksync is completed
```
omegablitz: statesync?

PR #1518, opened 2025-02-10
web3_clientVersion returns rpc version

philipglazman:
`web3_clientVersion` returns the rpc client version (e.g. "Monad/v0.7.1-sha")
fixes https://github.com/category-labs/category-internal/issues/1240

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -12,8 +12,8 @@
 /docker/flexnet/logs
 /docker/flexnet/save_logs
 
-# unfortunately some submodules require .git folder to exist...
-/.git
+# unfortunately some submodules and rpc's build script require .git folder to exist...
+!/.git
 
 **/target
 


ariqchowdhury:
do we need to consider processes being updated independently? The way its implemented assumes that the bft and rpc version are coming from the same build but someone could update rpc independently. The rpc API from eth matches their architecture (everything is in the same process). 
This is at least an upgrade over the current behaviour of being unimplemented.

philipglazman:
I think the primary purpose of this endpoint is for app to enable/disable features depending on rpc client support (e.g. new method call or deprecated method call). I can't find a specific example of any app actually doing this. I think it has less to do with bft or execution process. My assumption is that the operator updates all processes together.

philipglazman:
response for web3_clientVersion assuming built at git tag v9.9.9

```json
{
  "jsonrpc": "2.0",
  "result": "Monad/v9.9.9",
  "id": 1
}
```

PR #1525, opened 2025-02-11
Fix Raptorcast dissector crash on secp256k1_context_create()

fabiotdf:
When dissecting Raptorcast messages, wireshark would crash, saying: 

```
[libsecp256k1] illegal argument: secp256k1_ecmult_context_is_built(&ctx->ecmult_ctx) 
```
This was found when running:
```
RUST_BACKTRACE=1 RUST_LOG=trace cargo run -p monad-raptorcast --release --example service -- \
        --addresses             \
        127.0.0.1:8000          \
        127.0.0.2:8000          \
        127.0.0.3:8000          \
        127.0.0.4:8000          \
        127.0.0.5:8000
```

### DIFF
--- a/monad-raptorcast/wireshark/raptorcast.c
+++ b/monad-raptorcast/wireshark/raptorcast.c
@@ -128,7 +128,7 @@ static enum signature_status recover_author(uint8_t *author, size_t authorlen, c
 	static struct signature_cache_entry cache_entries[256];
 
 	if (ctx == NULL) {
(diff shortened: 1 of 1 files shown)


fabiotdf:
Thanks @buytenh! If you don't mind, I will add that info to the commit message

buytenh:
> Thanks @buytenh! If you don't mind, I will add that info to the commit message

Go for it

PR #1526, opened 2025-02-11
RPC uses node.toml for chain_id and metrics service name

philipglazman:
removes `chain-id` and `metrics-service-name` from cli in favor of passing node-toml as `node-config`
related: https://github.com/category-labs/category-internal/issues/1188


monad-rpc/src/cli.rs
```diff
@@ -5,70 +5,66 @@ use clap::Parser;
...
-    /// Set the port number for RPC to listen
-    #[arg(long, default_value_t = 8080)]
-    pub rpc_port: u16,
+    pub archive_api_key: Option<String>,
 
-    /// Set the chain ID
+    /// Set the archive URL to read archive data from
     #[arg(long)]
-    pub chain_id: u64,
+    pub archive_url: Option<String>,
```
eerkaijun: nit: would prefer the archive cli arguments (e.g. archive-url, region, s3-bucket etc) to be grouped together at the end rather than being alphabetically sorted

philipglazman: Fine with me, will group them together

PR #1527, opened 2025-02-11
Chain config

michael-yxchen:
Flexnet PR to test block pacing changeover: https://github.com/category-labs/monad-integration/pull/210


monad-consensus-state/src/lib.rs
```diff
@@ -755,7 +774,12 @@ where
 
         // start the vote-timer for the next round
         let vote_timer_cmd = ConsensusCommand::ScheduleVote {
-            duration: self.config.vote_pace,
+            duration: self
+                .config
+                .chain_config
+                .get_chain_revision(round + Round(1))
+                .chain_params()
+                .vote_pace,
```
michael-yxchen, line 818: @ariqchowdhury can you check if im off by one here?

ariqchowdhury: yeah, I think that's correct. We want the vote-pace to take effect in round 2, so in round 1 we schedule the vote for round 2 and use that different time. If the round2 proposal gets sent out instantly, everyone will still wait until round2 timer expires before voting.

monad-rpc/src/eth_txn_handlers.rs
```diff
@@ -396,7 +398,7 @@ pub async fn monad_eth_sendRawTransaction<T: Triedb>(
     match TxEnvelope::decode(&mut &params.hex_tx.0[..]) {
         Ok(txn) => {
             // drop transactions that will fail consensus static validation
-            if let Err(err) = static_validate_transaction(&txn, chain_id) {
+            if let Err(err) = static_validate_transaction(&txn, chain_id, PROPOSAL_GAS_LIMIT) {
```
michael-yxchen: Behavior change here, as rpc is unaware of ChainConfig/current round. I'm conservatively setting it to the lower/new proposal_gas_limit

monad-chain-config/src/revision.rs
```diff
@@ -33,6 +33,7 @@ impl ChainRevision for MockChainRevision {
 pub struct ChainParams {
     pub tx_limit: usize,
     pub proposal_gas_limit: u64,
+    // Max proposal size in bytes (average transactions ~400 bytes)
     pub proposal_size_limit: u64,
```
ariqchowdhury: nit: could rename this proposal_byte_limit for clarity

monad-rpc/src/eth_txn_handlers.rs
ariqchowdhury: Can we have an option on ChainConfig to return "latest" so that we can still get these constants from the same place as bft. And then we're getting chain-id from the same place too

monad-rpc/Cargo.toml
```diff
@@ -50,6 +50,7 @@ schemars = "0.8.21"
 opentelemetry = { workspace = true, features = ["metrics"] }
 opentelemetry-otlp = { workspace = true }
 opentelemetry_sdk = { workspace = true, features = ["rt-tokio"] }
+monad-chain-config = { version = "0.1.0", path = "../monad-chain-config" }
```
omegablitz: nit: fmt

PR #1529, opened 2025-02-12
Expose voted in RPC

omegablitz:


omegablitz:
To wait for execution changes to merge before this gets merged

monad-rpc/src/eth_json_types.rs
```diff
@@ -379,9 +379,11 @@ impl<'de, const N: usize> Deserialize<'de> for FixedData<N> {
 #[derive(Clone, Debug, Default, PartialEq, Eq, schemars::JsonSchema)]
 #[serde(untagged)]
 pub enum BlockTags {
-    Number(Quantity),
+    Number(Quantity), // voted or finalized
     #[default]
-    Latest,
+    Latest, // voted
+    Safe,             // voted
+    Finalized,        // finalized
```
eerkaijun, line 386: will need to update the `FromStr` below

```
  fn from_str(s: &str) -> Result<Self, Self::Err> {
        match s {
            "earliest" => Ok(Self::Latest),
            "latest" => Ok(Self::Latest),
            "safe" => Ok(Self::Safe),
            "finalized" => Ok(Self::Finalized),
            "pending" => Ok(Self::Latest),
            _ => decode_quantity(s).map(|q| Self::Number(Quantity(q))),
        }
    }
```

i guess earliest is not correct but we can fix that later

monad-rpc/src/block_handlers.rs
```diff
@@ -15,17 +18,12 @@ use crate::{
...
+pub fn get_block_key_from_tag<T: Triedb>(triedb_env: &T, tag: BlockTags) -> BlockKey {
     match tag {
-        BlockTags::Number(n) => Ok(n.0),
-        BlockTags::Latest => triedb_env
-            .get_latest_block()
-            .await
-            .map_err(JsonRpcError::internal_error),
+        BlockTags::Number(n) => triedb_env.get_block_key(SeqNum(n.0)),
+        BlockTags::Latest => triedb_env.get_latest_voted_block_key(),
+        BlockTags::Safe => triedb_env.get_latest_voted_block_key(),
```
andr-dev, line 25: Super nit but we should group these variants to make it clear they make the same triedb_env call.

monad-triedb-utils/src/triedb_env.rs
```diff
@@ -186,7 +197,49 @@ fn polling_thread(
...
+                let mut meta = meta.lock().expect("triedb poller mutex poisoned");
+                meta.latest_finalized = latest_finalized;
+                meta.latest_voted = latest_voted;
+                while meta
+                    .voted_proposals
+                    .first_key_value()
+                    .is_some_and(|(seq_num, _)| seq_num <= &latest_finalized.0)
+                {
+                    meta.voted_proposals.pop_first();
+                }
```
andr-dev, line 234: Might be a good place to use
https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.split_off

monad-rpc/src/block_handlers.rs
omegablitz: Lazy but I'll do this eventually

PR #1530, opened 2025-02-12
Add base fee checking to rpc

ariqchowdhury:


ariqchowdhury:
resolves category-labs/category-internal#1208

monad-rpc/src/eth_txn_handlers.rs
```diff
@@ -417,6 +436,12 @@ pub async fn monad_eth_sendRawTransaction<T: Triedb>(
                 ));
             }
 
+            if !base_fee_validation(txn.max_fee_per_gas(), base_fee_per_gas) {
+                return Err(JsonRpcError::custom(
+                    "MaxFeePerGas too low to be include in upcoming blocks".to_string(),
```
andr-dev: I would use `maxFeePerGas` here as that's how the eth people define it.
https://ethereum.org/en/developers/docs/gas/#maxfee

PR #1533, opened 2025-02-12
Emit tx events from txpool to rpc

andr-dev:


monad-eth-txpool-executor/src/lib.rs
```diff
@@ -128,22 +148,48 @@ where
...
+                        .filter_map(|raw_tx| {
+                            let Ok(tx) = TxEnvelope::decode(&mut raw_tx.as_ref()) else {
+                                num_invalid_bytes += 1;
+                                return None;
+                            };
+
+                            let Ok(signer) = tx.recover_signer() else {
+                                num_invalid_signer += 1;
+                                return None;
+                            };
```
ariqchowdhury, line 190: do we need to do more validation for forwarded txns, like rpc does? https://github.com/category-labs/monad-bft/blob/master/monad-rpc/src/eth_txn_handlers.rs#L418

monad-eth-txpool-executor/src/lib.rs
```diff
@@ -128,22 +148,48 @@ where
...
+                    self.pool
+                        .insert_txs(
+                            &mut context,
+                            &self.block_policy,
+                            &self.state_backend,
+                            txs,
+                            false,
+                            |_| {},
+                        )
+                        .expect("insert suceeds");
```
ariqchowdhury: should insert_txs be returning a Result type if it can never fail?

monad-eth-txpool/src/pool/mod.rs
```diff
@@ -82,149 +75,85 @@ where
...
-
-        self.update_aggregate_metrics(metrics);
-
-        results
+        Ok(())
     }
 
     pub fn create_proposal(
         &mut self,
+        context: &mut EthTxPoolContext<'_>,
```
ariqchowdhury: nit: txpool_context

andr-dev:
Closes https://github.com/category-labs/category-internal/issues/1207.

PR #1535, opened 2025-02-12
eth_getLogs should use latest block if to_block's number is not yet committed

philipglazman:
fixes https://github.com/category-labs/category-internal/issues/1302

### DIFF
--- a/monad-rpc/src/eth_txn_handlers.rs
+++ b/monad-rpc/src/eth_txn_handlers.rs
@@ -192,7 +192,11 @@ pub async fn monad_eth_getLogs<T: Triedb>(
 
             let from_block = get_block_key_from_tag(triedb_env, from_block_tag);
             let to_block = get_block_key_from_tag(triedb_env, to_block_tag);
-            (from_block.seq_num().0, to_block.seq_num().0)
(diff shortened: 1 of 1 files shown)


monad-rpc/src/eth_txn_handlers.rs
```diff
@@ -187,11 +187,15 @@ pub async fn monad_eth_getLogs<T: Triedb>(
...
                     JsonRpcError::internal_error("could not get starting block range".to_string())
                 })?;
-            let to_block = get_block_num_from_tag(triedb_env, to_block_tag)
-                .await
-                .map_err(|_| {
-                    JsonRpcError::internal_error("could not get ending block range".to_string())
-                })?;
+
+            let to_block = match get_block_num_from_tag(triedb_env, to_block_tag).await {
+                Ok(block) => Ok(block),
```
eerkaijun: get_block_num_from_tag doesn't check if the provided block is larger than latest block, so it will still return ok rather than err here

I think the correct behavior should be to get both to_block_tag and BlockTags::Latest and use whichever is smaller

philipglazman: I just rebased and the new version of `get_block_key_from_tag` always returns a block number. I just Updated it to use min(to_block, latest_block)

PR #1536, opened 2025-02-12
Map small-message BuildTarget::Raptorcast(..) to ::Broadcast(..)

buytenh:
Please see the individual commit messages for descriptions of each of the individual changes.

### DIFF
--- a/monad-raptorcast/src/lib.rs
+++ b/monad-raptorcast/src/lib.rs
@@ -270,14 +270,18 @@ where
                             let full_nodes_view = self.full_nodes.view();
 
                             let build_target = match &target {
-                                RouterTarget::Broadcast(_) => BuildTarget::Broadcast(
-                                    epoch_validators_without_self,
-                                    full_nodes_view,
-                                ),
-                                RouterTarget::Raptorcast(_) => BuildTarget::Raptorcast(
-                                    epoch_validators_without_self,
-                                    full_nodes_view,
-                                ),
+                                RouterTarget::Raptorcast(_) if app_message.len() > 19520 => {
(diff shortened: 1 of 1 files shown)


monad-raptorcast/src/lib.rs
```diff
@@ -270,15 +270,16 @@ where
...
-                                    full_nodes_view,
-                                ),
-                                RouterTarget::Raptorcast(_) => BuildTarget::Raptorcast(
+                                RouterTarget::Raptorcast(_) if app_message.len() > 19520 => {
+                                    BuildTarget::Raptorcast(
+                                        epoch_validators_without_self,
+                                        full_nodes_view,
+                                    )
+                                }
+                                _ => BuildTarget::Broadcast(
```
omegablitz: nit: should we assert that the target is Broadcast or Raptorcast? just as a safety

omegablitz:
As a later TODO, we need to add flexnet tests that create bigger blocks. The large block raptorcast path won't get hit in flexnet after this change.

Stressnet will still cover this though.

monad-raptorcast/src/lib.rs
buytenh: Makes sense -- it now does as you suggested: https://github.com/category-labs/monad-bft/pull/1536/commits/94d31a6f62cde60949b0d8c9c95eef4ab7f16709

buytenh:
> As a later TODO, we need to add flexnet tests that create bigger blocks. The large block raptorcast path won't get hit in flexnet after this change.

Created https://github.com/category-labs/category-internal/issues/1303 to track this

PR #1547, opened 2025-02-13
Reset batch timer in txpool ipc server

andr-dev:


andr-dev:
Included in https://github.com/category-labs/monad-bft/pull/1552.

PR #1552, opened 2025-02-13
Add tx batch sending from RPC to txpool

andr-dev:


monad-rpc/src/txpool/mod.rs
```diff
@@ -132,7 +132,7 @@ impl Future for SocketWatcher {
     }
 }
 
-impl<'a> Sink<&'a TxEnvelope> for EthTxPoolBridge {
+impl Sink<Vec<TxEnvelope>> for EthTxPoolBridge {
```
omegablitz: I think you can keep this as a `Sink<TxEnvelope>`.

You just need to use [SinkExt::send_all](https://docs.rs/futures/latest/futures/prelude/sink/trait.SinkExt.html#method.send_all) on the ipc_receiver.drain() iterator instead of [SinkExt::send](https://docs.rs/futures/latest/futures/prelude/sink/trait.SinkExt.html#method.send) so that it doesn't flush after each element. Framed<_> already has [some write buffering](https://docs.rs/tokio-util/latest/src/tokio_util/codec/framed_impl.rs.html#60) (8KB).

I think it's more intuitive that way - it's weirder to deal with a Stream<Vec<Tx>> on the rx side.

PR #1554, opened 2025-02-13
Separate forward tx events into chunks

andr-dev:


omegablitz:
What's the benefit of splitting these into chunks?

PR #1563, opened 2025-02-16
Hit archive if block is statesynced

eerkaijun:
addresses https://github.com/category-labs/category-internal/issues/1315


monad-archive/src/kvstore/triedb_reader.rs
```diff
@@ -46,7 +46,7 @@ impl BlockDataReader for TriedbReader {
             .db
             .get_transactions(BlockKey::Finalized(FinalizedBlockKey(SeqNum(block_num))))
             .await
-            .map_err(|e| eyre!("{e}"))?;
+            .map_err(|e| eyre!("Header exists but not transactions, block {block_num} might be statesynced: {e}"))?;
```
JoeHowarth: nit: for errors please use {e:?} to display the wrapped error stack

PR #1564, opened 2025-02-16
Disable author-not-sender check

omegablitz:
The message validation logic is stronger than it needs to be. We currently are asserting that a consensus message sender matches the author of the message. However, nodes might want to forward messages authored by one node to other nodes.

Additionally, the existing message validation logic does not actually prevent a sender from spoofing chunks authored and sent by other nodes.

This commit disables this check.

### DIFF
--- a/monad-consensus/src/validation/signing.rs
+++ b/monad-consensus/src/validation/signing.rs
@@ -159,7 +159,7 @@ where
         self,
         epoch_manager: &EpochManager,
         val_epoch_map: &ValidatorsEpochMapping<VTF, SCT>,
-        sender: &SCT::NodeIdPubKey,
+        _sender: &SCT::NodeIdPubKey,
     ) -> Result<Verified<ST, Unvalidated<ConsensusMessage<ST, SCT, EPT>>>, Error>
     where
         VTF: ValidatorSetTypeFactory<ValidatorSetType = VT>,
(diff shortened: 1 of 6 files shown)


monad-consensus/src/validation/signing.rs
```diff
@@ -552,22 +547,13 @@ where
...
-/// staked validator
+/// Verify that the author is a staked validator
 fn verify_author<ST: CertificateSignatureRecoverable>(
     validators: &BTreeMap<NodeId<CertificateSignaturePubKey<ST>>, Stake>,
-    sender: &CertificateSignaturePubKey<ST>,
     msg: &Hash,
     sig: &ST,
 ) -> Result<CertificateSignaturePubKey<ST>, Error> {
-    let pubkey = get_pubkey(msg.as_ref(), sig)?.valid_pubkey(validators)?;
-    sig.verify(msg.as_ref(), &pubkey)
```
omegablitz, line 588: Note that I removed the ST::verify here. This is redundant with `get_pubkey`.

monad-consensus/src/validation/signing.rs
```diff
@@ -164,7 +164,7 @@ where
         self,
         epoch_manager: &EpochManager,
         val_epoch_map: &ValidatorsEpochMapping<VTF, SCT>,
-        sender: &SCT::NodeIdPubKey,
+        _sender: &SCT::NodeIdPubKey,
```
andr-dev, line 162: Why not remove this arg?

PR #1565, opened 2025-02-17
Configurable request payload

eerkaijun:
Closes https://github.com/category-labs/category-internal/issues/1316

### DIFF
--- a/monad-rpc/src/cli.rs
+++ b/monad-rpc/src/cli.rs
@@ -29,6 +29,10 @@ pub struct Cli {
     #[arg(long, default_value_t = 5000)]
     pub batch_request_limit: u16,
 
+    /// Set the max request size in bytes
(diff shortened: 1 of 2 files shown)


monad-rpc/src/cli.rs
```diff
@@ -29,6 +29,10 @@ pub struct Cli {
     #[arg(long, default_value_t = 5000)]
     pub batch_request_limit: u16,
 
+    /// Set the max request size in bytes
+    #[arg(long, default_value_t = 10_000)]
+    pub max_request_size: u32,
```
omegablitz: Nit: usize, remove casts below. Otherwise this is still hard capped to 4GB.

ariqchowdhury: this can be usize here

eerkaijun: updated

PR #1567, opened 2025-02-18
Add RPC block cache

omegablitz:
Adds a block cache in front of triedb. For now, only caches transactions. Also, the cache is not hydrated on startup.

### DIFF
--- a/monad-triedb-utils/src/triedb_env.rs
+++ b/monad-triedb-utils/src/triedb_env.rs
@@ -13,7 +13,7 @@ use std::{
 use alloy_consensus::{Header, ReceiptEnvelope, TxEnvelope};
 use alloy_primitives::{keccak256, Address, FixedBytes};
 use alloy_rlp::{encode_list, BytesMut, Decodable, Encodable};
-use futures::channel::oneshot;
+use futures::{channel::oneshot, FutureExt};
 use monad_triedb::{TraverseEntry, TriedbHandle};
 use monad_types::{Round, SeqNum};
 use serde::{Deserialize, Serialize};
@@ -168,6 +168,9 @@ impl Decodable for TxEnvelopeWithSender {
 const MAX_QUEUE_BEFORE_POLL: usize = 100;
 const MAX_POLL_COMPLETIONS: usize = usize::MAX;
 const META_POLL_INTERVAL: Duration = Duration::from_millis(5);
+const MAX_FINALIZED_BLOCK_CACHE_LEN: usize = 200;
+// Unfinalized chains should be very short
+const MAX_VOTED_BLOCK_CACHE_LEN: usize = 3;
 
 fn get_latest_voted_block_key(triedb_handle: &TriedbHandle) -> Option<ProposedBlockKey> {
     let latest_voted_round_before = Round(triedb_handle.latest_voted_round()?);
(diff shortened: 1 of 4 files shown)


monad-triedb-utils/src/triedb_env.rs
```diff
@@ -168,6 +168,8 @@ impl Decodable for TxEnvelopeWithSender {
 const MAX_QUEUE_BEFORE_POLL: usize = 100;
 const MAX_POLL_COMPLETIONS: usize = usize::MAX;
 const META_POLL_INTERVAL: Duration = Duration::from_millis(5);
+const MAX_FINALIZED_CACHE_LEN: usize = 200;
```
ariqchowdhury: nit: MAX_FINALIZED_BLOCK_CACHE_LEN, MAX_VOTED_BLOCK_CACHE_LEN

monad-triedb-utils/src/triedb_env.rs
```diff
@@ -220,8 +222,25 @@ fn polling_thread(
             last_meta_updated = Instant::now();
 
             if last_finalized != latest_finalized || last_voted != latest_voted {
-                last_finalized = latest_finalized;
-                last_voted = latest_voted;
+                if last_finalized != latest_finalized {
```
ariqchowdhury: nit: can we make some bools for these conditions so that we have a descriptive name of the condition?
ie. let descriptive_label = last_finalized != latest_finalized. etc

monad-triedb-utils/src/triedb_env.rs
```diff
@@ -168,6 +168,8 @@ impl Decodable for TxEnvelopeWithSender {
 const MAX_QUEUE_BEFORE_POLL: usize = 100;
 const MAX_POLL_COMPLETIONS: usize = usize::MAX;
 const META_POLL_INTERVAL: Duration = Duration::from_millis(5);
+const MAX_FINALIZED_CACHE_LEN: usize = 200;
+const MAX_VOTED_CACHE_LEN: usize = 3;
```
ariqchowdhury: this is 3 because we're not expecting a large branch of non-consecutive rounds right? A comment would be useful here

monad-triedb-utils/src/triedb_env.rs
```diff
@@ -313,9 +332,64 @@ fn polling_thread(
...
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+fn populate_cache(handle: &TriedbHandle, meta: Arc<Mutex<TriedbEnvMeta>>, block_key: BlockKey) {
+    let tx_receiver = {
+        let (tx_sender, tx_receiver) = oneshot::channel();
+
+        // txn_index set to None to indiciate return all transactions
+        let (triedb_key, key_len_nibbles) =
+            create_triedb_key(block_key.into(), KeyInput::TxIndex(None));
```
ariqchowdhury, line 352: nit: if returning all transactions is always controlled by hardcoding the KeyInput, I would make a helper function that wraps that

omegablitz: I'll fix this later when I get around to cleaning up this file. I want to refactor how these triedb_keys are constructed.

PR #1570, opened 2025-02-19
Ledger tail

omegablitz:
This is very primitive right now, and will only really be useful for chains that are mostly linear. This is the common case though, so will work fine for the purposes of testnet observability.


monad-node/examples/ledger-tail.rs
```diff
@@ -0,0 +1,180 @@
...
+    while let Some(mut next_block) = block_stream.next().await {
+        let mut block_queue = Vec::new();
+        loop {
+            let next_block_id = next_block.get_id();
+            if visited_blocks.contains(&next_block_id) {
+                break;
+            }
+            block_queue.push(next_block.clone());
+            if let Some(next_block_parent) =
+                read_full_block(&block_persist, &next_block.get_parent_id())
```
michael-yxchen, line 83: super nit: read parent_id from header

michael-yxchen: ignore it

omegablitz: resolved offline

PR #1572, opened 2025-02-19
Txgen: do not fetch erc20 balances unless erc20_balance_of cli arg set

JoeHowarth:
This was causing 408s on testnet and stressnet after v0.8
Likely due to a change in how eth_call is being handled

### DIFF
--- a/monad-eth-testutil/examples/txgen/generators/few_to_many.rs
+++ b/monad-eth-testutil/examples/txgen/generators/few_to_many.rs
@@ -5,7 +5,7 @@ pub struct CreateAccountsGenerator {
     pub recipient_keys: SeededKeyPool,
     pub tx_per_sender: usize,
     pub tx_type: TxType,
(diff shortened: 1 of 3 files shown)


monad-eth-testutil/examples/txgen/workers/refresher.rs
```diff
@@ -84,20 +97,29 @@ pub async fn refresh_batch(
...
     );
 
     let native_bals = native_bals?;
     let nonces = nonces?;
-    let erc20_bals = erc20_bals?;
+
+    let erc20_bals = match erc20_bals {
+        Some(b) => Some(b?),
+        None => None,
+    };
```
andr-dev, line 122: Probably my favourite Option function
https://doc.rust-lang.org/std/option/enum.Option.html#method.transpose
Turns an `Option<Result<T, E>>` into a `Result<Option<T>, E>` :)

PR #1576, opened 2025-02-20
Add eth_call to request generator

eerkaijun:


eerkaijun:
> Think it would be more useful to mimic a real usecase like calling balanceof for erc20 contracts

agree but that is already covered by the txgen so thought will add some different variety instead

monad-eth-testutil/examples/rpc-request-generator/main.rs
```diff
@@ -27,6 +27,14 @@ use tracing_subscriber::{
...
 use url::Url;
 
+// bytecode for eth_call contract deployment simulation
+const CONTRACT_BYTECODE: &[u8] = &[
+    0x60, 0x45, 0x80, 0x60, 0x0e, 0x60, 0x00, 0x39, 0x80, 0x60, 0x00, 0xf3, 0x50, 0xfe, 0x7f, 0xff,
+    0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xfe, 0x03, 0x60, 0x16, 0x00, 0x81, 0x60, 0x20,
+    0x82, 0x37, 0x80, 0x35, 0x82, 0x34, 0xf5, 0x80, 0x15, 0x56, 0x03, 0x95, 0x78, 0x18, 0x2f, 0x5b,
+    0x80, 0x82, 0x52, 0x50, 0x50, 0x60, 0x14, 0x60, 0x0c, 0xf3,
+];
+
```
michael-yxchen: what does this contract do? A comment would help

monad-eth-testutil/examples/rpc-request-generator/main.rs
```diff
@@ -608,36 +659,44 @@ impl RpcRequestGenerator {
...
+            // account balances and eth call
             let start = Instant::now();
-            let uniq_addrs = receipts
+            let addrs = receipts
                 .into_iter()
-                .flat_map(|receipt| [Some(receipt.from), receipt.to])
-                .flatten()
-                .unique()
+                .map(|receipt| receipt.from)
                 .collect::<Vec<_>>();
```
michael-yxchen, line 675: what's the reason for changing to only query from address states and removing uniq?

eerkaijun: wanted to have larger number of requests. transactions by the txgen has relatively few unique addresses in a block

monad-eth-testutil/examples/rpc-request-generator/main.rs
eerkaijun: just a test contract we use in flexnet, will add comment

PR #1580, opened 2025-02-20
Compare websocket blocks with rpc response

philipglazman:


monad-eth-testutil/examples/rpc-request-generator/main.rs
```diff
@@ -79,6 +81,11 @@ pub struct Config {
     pub requests_per_block: u64,
     #[arg(long, global = true, default_value_t = 15)]
     pub max_distance_from_tip: u64,
+    #[arg(long, global = true, default_value = "ws://localhost:8081")]
+    pub ws_url: Url,
+    /// Enable rpc-request-generator to listen to new blocks from the websocket stream.
+    #[arg(long, global = true, default_value_t = false)]
+    pub ws_enabled: bool,
```
eerkaijun: can `ws_url` just be Option<Url> so we don't need a separate `ws_enabled` cli argument?

monad-eth-testutil/examples/rpc-request-generator/main.rs
```diff
@@ -190,7 +273,10 @@ impl RpcRequestGenerator {
 
                 Some(index_result) = index_done_receiver.recv() => {
                     let block_number = match index_result {
-                        Ok((num, addrs)) => {
+                        Ok((block, num, addrs)) => {
+                            if self.ws_url.is_some() {
+                                rpc_block_sender.send(block).await.expect("rpc block sender failed");
+                            };
```
eerkaijun, line 348: just noting that average duration to finish indexing a block is currently 500ms-1s with all the rpc requests by the generator, so I presume the rpc will always be lagging behind the websocket stream

monad-eth-testutil/examples/rpc-request-generator/main.rs
```diff
@@ -129,8 +136,76 @@ impl TipRefresher {
...
+                    rpc = rpc_block.header.number,
+                    ws = header.number,
+                    "block number mismatch"
+                );
+
+                // If stream is ahead, take items from rpc recv.
+                if header.number > rpc_block.header.number {
+                    let mut diff = header.number - rpc_block.header.number;
+                    while diff > 0 {
+                        _ = recv.recv().await.take().expect("rpc block not found");
```
eerkaijun: with regards to comment above, _I think_ this `expect` statement will be hit quite easily?

monad-eth-testutil/examples/rpc-request-generator/main.rs
philipglazman: Yep, that's fine. Figured its easier to have `ws_url` provide the default url. But one flag is better than two.

monad-eth-testutil/examples/rpc-request-generator/main.rs
philipglazman: WebSocket streams currently only notify about _finalized_ blocks, so it will be slower than polling for rpc.

PR #1581, opened 2025-02-20
More Caching and CLI Args

eclark0426:


monad-triedb-utils/src/triedb_env.rs
```diff
@@ -853,6 +853,10 @@ impl Triedb for TriedbEnv {
         block_key: BlockKey,
         receipt_index: u64,
     ) -> Result<Option<ReceiptWithLogIndex>, String> {
+        if let Some(receipts) = self.get_block_cache(&block_key) {
```
ariqchowdhury: nit: `if let Some(cache)`?

monad-triedb-utils/src/triedb_env.rs
```diff
@@ -948,6 +952,10 @@ impl Triedb for TriedbEnv {
         block_key: BlockKey,
         txn_index: u64,
     ) -> Result<Option<TxEnvelopeWithSender>, String> {
+        if let Some(txs) = self.get_block_cache(&block_key) {
```
ariqchowdhury: nit: `if let Some(cache)`?

monad-triedb-utils/src/triedb_env.rs
```diff
@@ -556,6 +553,8 @@ struct TriedbEnvMeta {
 
     finalized_cache: BTreeMap<FinalizedBlockKey, BlockCache>,
     voted_cache: BTreeMap<ProposedBlockKey, BlockCache>,
+    finalized_cache_len: usize,
```
omegablitz: I think these "len" params should be passed in as arguments to `populate_cache`.

"Meta" is the cache itself, is poorly named right now.

When we clean up TriedbEnv later, we'll have populate_cache be an actual struct member function and so we won't need to pass in all these args to populate_cache manually.

PR #1583, opened 2025-02-21
Use PayloadConfig instead of JsonConfig

eerkaijun:


monad-rpc/src/cli.rs
```diff
@@ -30,7 +30,7 @@ pub struct Cli {
     pub batch_request_limit: u16,
 
     /// Set the max request size in bytes
-    #[arg(long, default_value_t = 10_000)]
+    #[arg(long, default_value_t = 500_000)]
```
ariqchowdhury: With the original code, what was the actual effective request limit? 10_000 was not taking effect, but we should change the new default value to match the old behaviour so that there is no change in behaviour for existing deployments that don't want anything to change.

ariqchowdhury:
can you link the actix github issue you found

monad-rpc/src/cli.rs
eerkaijun: default was 262_144; // 2^18 bytes (~256kB)

eerkaijun:
https://github.com/actix/actix-web/issues/1419

PR #1585, opened 2025-02-23
Archive: Remove unnecessary checkpointing and mongo concurrency fix

JoeHowarth:
- Remove in-stream checkpointing. Causes deadlock when used with mongo client if `--max-concurrent-blocks > aws-concurrency`.  Also increases write costs by 2-5% without benefit since more frequent checkpointing can be achieved by setting lower `--max-blocks-per-iteration`.
- increase polling time for new blocks in indexer and archiver
- increase mongo connection pool size. Deadlocks observed when pool size < `max_concurrent_blocks` with in-stream checkpointing. Likely unnecessary now that in-stream checkpointing is removed, but small observed performance improvement during backfilling.
- Remove `bulk_put` in mongo implementation since insert_many does not overwrite existing keys. Fallback to default impl since no replace_many exists in api

### DIFF
--- a/monad-archive/src/kvstore/mongo.rs
+++ b/monad-archive/src/kvstore/mongo.rs
@@ -9,6 +9,8 @@ use serde::{Deserialize, Serialize};
 
 use crate::prelude::*;
 
+const MAX_CONNECTION_POOL_SIZE: u32 = 50;
+
 #[derive(Clone)]
 pub struct MongoDbStorage {
     collection: Collection<KeyValueDocument>,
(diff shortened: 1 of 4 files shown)


monad-archive/src/kvstore/mongo.rs
```diff
@@ -56,8 +56,10 @@ impl MongoDbStorage {
         collection_name: &str,
         create_with_capped_size: Option<u64>,
     ) -> Result<Self> {
-        let client_options = ClientOptions::parse(connection_string).await?;
+        let mut client_options = ClientOptions::parse(connection_string).await?;
+        client_options.max_pool_size = Some(50);
```
eerkaijun: named constant?

PR #1590, opened 2025-02-25
Reduce code duplication in TriedbEnv

eclark0426:


monad-triedb-utils/src/triedb_env.rs
```diff
@@ -1120,56 +967,18 @@ impl Triedb for TriedbEnv {
...
-                            Err(String::from("error reading from db"))
-                        }),
-                    None => Ok(None),
-                }
-            }
-            Err(e) => {
-                error!("Error awaiting result: {e}");
-                Err(String::from("error reading from db"))
-            }
+            Some(transaction_location) => Ok(Some(TransactionLocation {
```
andr-dev:
```rust
Some((block_num, tx_index)) => Ok(Some(TransactionLocation {
    block_num,
    tx_index,
})),

File: monad-triedb-utils/src/triedb_env.rs [hunk 2]
```diff
@@ -805,109 +777,179 @@ impl Triedb for TriedbEnv {
         }
     }
 
-    #[tracing::instrument(level = "debug")]
-    async fn get_storage_at(
+    fn send_async_range_request(
```
omegablitz, line 799: (to be changed in future)

Just noting that this function isn't actually a generic range request, it's specialized to tries that are keyed by indices. We don't need to change the naming of this for now, it's obvious enough from reading the code. We can either rename this function or pass in the two range keys as arguments to this function.

monad-triedb-utils/src/triedb_env.rs
```diff
@@ -725,18 +693,27 @@ impl Triedb for TriedbEnv {
...
         }
 
-        // await the result using request_receiver
-        match request_receiver.await {
+        Ok((request_receiver, completed_counter))
+    }
+
+    async fn handle_async_result<T, F>(
+        receiver: oneshot::Receiver<Option<Vec<u8>>>,
+        counter: Arc<AtomicUsize>,
```
omegablitz, line 720: (to be changed in future)

I don't think we need to propagate this counter, we had it before just as a sanity check but it's getting annoying now that we need to pass it around. The current changeset maintains parity at least.

monad-triedb-utils/src/triedb_env.rs
```diff
@@ -921,53 +963,17 @@ impl Triedb for TriedbEnv {
...
-                    None => Ok(None),
-                }
-            }
-            Err(e) => {
-                error!("Error awaiting result: {e}");
-                Err(String::from("error reading from db"))
-            }
-        }
+            |data| {
+                let mut rlp_buf = data.as_slice();
```
omegablitz, line 989: (to be changed in future)

All of these `let mut rlp_buf = data.as_slice()` and `T::decode` should be replaced with [alloy_rlp::decode_exact](https://docs.rs/alloy-rlp/latest/alloy_rlp/fn.decode_exact.html) to assert that there aren't any trailing bytes.

eclark0426:
I created an issue referencing the three future improvements @omegablitz mentioned: https://github.com/category-labs/category-internal/issues/1508

PR #1593, opened 2025-02-26
Account for vote pace in timestamp checks

ariqchowdhury:


monad-consensus-state/src/timestamp.rs
```diff
@@ -94,12 +97,12 @@ mod test {
         let mut b = BlockTimestamp::new(10, 1);
         b.update_time(0);
 
-        assert!(b.valid_block_timestamp(1, 1).is_none());
-        assert!(b.valid_block_timestamp(2, 1).is_none());
-        assert!(b.valid_block_timestamp(0, 11).is_none());
+        assert!(b.valid_block_timestamp(1, 1, 0).is_none());
+        assert!(b.valid_block_timestamp(2, 1, 0).is_none());
+        assert!(b.valid_block_timestamp(0, 11, 0).is_none());
```
mikety, line 102: nit: I suggest adding a test to confirm that vote_delay_ns expands the validation window:
 assert!(matches!(b.valid_block_timestamp(0, 11, 20),
 Some(TimestampAdjustment {
                delta: 10,
                direction: TimestampAdjustmentDirection::Forward
            })
        ));

ariqchowdhury: done

PR #1596, opened 2025-02-26
Try from archive if present for EIP1898

eerkaijun:
Closes https://github.com/category-labs/category-internal/issues/1355 and https://github.com/category-labs/category-external/issues/56

### DIFF
--- a/monad-rpc/src/block_handlers.rs
+++ b/monad-rpc/src/block_handlers.rs
@@ -393,36 +393,47 @@ pub async fn monad_eth_getBlockReceipts<T: Triedb>(
 ) -> JsonRpcResult<Option<MonadEthGetBlockReceiptsResult>> {
     trace!("monad_eth_getBlockReceipts: {params:?}");
 
-    let block_key = get_block_key_from_tag_or_hash(triedb_env, params.block).await?;
-
-    if let Some(header) = triedb_env
-        .get_block_header(block_key)
-        .await
-        .map_err(JsonRpcError::internal_error)?
-    {
-        // if block header is present but transactions are not, the block is statesynced
-        if let Ok(transactions) = triedb_env.get_transactions(block_key).await {
-            let receipts = triedb_env
-                .get_receipts(block_key)
-                .await
-                .map_err(JsonRpcError::internal_error)?;
-            let block_receipts = map_block_receipts(
-                transactions,
-                receipts,
-                &header.header,
-                header.hash,
-                MonadTransactionReceipt,
-            )?;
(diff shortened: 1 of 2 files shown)


monad-rpc/src/gas_handlers.rs
```diff
@@ -145,7 +145,7 @@ fn estimate_gas<T: EthCallProvider>(
 pub struct MonadEthEstimateGasParams {
     tx: CallRequest,
     #[serde(default)]
-    block: BlockTags,
+    block: BlockTagOrHash,
```
philipglazman: The eip-1898 spec doesn't include `eth_estimateGas` but likely ok to deviate

monad-rpc/src/block_handlers.rs
```diff
@@ -29,6 +29,7 @@ pub fn get_block_key_from_tag<T: Triedb>(triedb_env: &T, tag: BlockTags) -> Bloc
 
 pub async fn get_block_key_from_tag_or_hash<T: Triedb>(
     triedb_env: &T,
+    archive_reader: &Option<ArchiveReader>,
     block_reference: BlockTagOrHash,
 ) -> JsonRpcResult<BlockKey> {
```
philipglazman: In our rpc methods, we usually call triedb again after calling `get_block_key_from_tag_or_hash`. If we know that triedb does not have the block_num, it would be nice to skip the subsequent triedb calls and only use archive_reader.

monad-rpc/src/gas_handlers.rs
eerkaijun: removed for consistency

monad-rpc/src/block_handlers.rs
eerkaijun: great point, updated

monad-rpc/src/block_handlers.rs
```diff
@@ -393,36 +393,47 @@ pub async fn monad_eth_getBlockReceipts<T: Triedb>(
...
-    {
-        if let Ok(receipts_with_log_index) = archive_reader.get_block_receipts(block_num.0).await {
-            if let Ok(block) = archive_reader.get_block_by_number(block_num.0).await {
+    if let Some(archive_reader) = archive_reader {
+        let block = match params.block {
+            BlockTagOrHash::BlockTags(tag) => match get_block_key_from_tag(triedb_env, tag) {
+                BlockKey::Finalized(FinalizedBlockKey(block_num)) => {
+                    archive_reader.get_block_by_number(block_num.0).await.ok()
+                }
+                _ => None,
```
omegablitz: nit: can you make this more explicit? `BlockKey::Proposed(_) -> None`

PR #1597, opened 2025-02-26
Enable TCP transmit pipelining

buytenh:
Where dataplane v1 would close an incoming TCP connection after having
received a single message on it, dataplane v2 (merged in commit 6854bf1b,
"Rebase monad-dataplane onto an io_uring-based async runtime") allows
receiving multiple messages over the same TCP connection.

Now that dataplane v2 has been rolled out everywhere, we can keep an
outbound TCP connection to a given peer address open as long as there
are messages to be transmitted to that peer address, and we can transmit
all outbound messages to that peer sequentially over that single TCP
connection.  This patch does that.

Also, while we're touching this code anyway, we're making a few changes
to more aggressively fail and drop outbound TCP messages in case of a
transport layer failure:

- Regardless of the size of the outbound message, we now impose a 10
  second timeout on the initial connect() call.  If the TCP three-way
  handshake does not complete within this timeout, we drop the message.
  (The current behavior is to allow the connect() phase to take up the
  entire computed message timeout time, which can be several minutes for
  a large message.)

- If we fail to connect to a peer, for whatever reason, we drop all
  messages that were queued to be transmitted to this peer.  (The
  current behavior is to re-connect for every outbound message, and to
  not drop subsequent messages if a connection fails to be established
  for the current message.)

- Similarly, if we experience a failure to deliver an outbound message
  after the three-way handshake, we will not try to reconnect and
  re-deliver this message, but we will drop this message, along with
  every other message that was queued to be transmitted to this peer.
  (The current behavior is to fail and drop only the current message.)

The reason for the current behavior, which is to try very hard to never
drop any outbound messages, is that building all the response messages
for a statesync request used to be much costlier than it is now, and we
had no way to abort the generation of a statesync response halfway
through in case of a transport error.

However, we now have a statesync response backpressure mechanism, which
allows us to abort building a statesync response chain early if there is
a transport error, and also, building a statesync response message is
much cheaper now.  In addition, some incidents on stressnet and testnet
have demonstrated that there is potential for queueing up excessive
numbers of messages when we have unhealthy peers, which in some cases
can even lead to running the local node out of memory and the subsequent
OOM killing of the monad-node process.  Finally, with the statesync
backpressure changes, we keep a statesync server busy for as long as
there are chunks to be transferred to the requesting node, which means
that if a transfer is happening very slowly, it can potentially tie up
the statesync server for a very long time.  All this necessitates the
new, more aggressive, dropping behavior.

### DIFF
--- a/monad-dataplane/src/tcp/tx.rs
+++ b/monad-dataplane/src/tcp/tx.rs
@@ -1,14 +1,19 @@
 use std::{
     cell::RefCell,
     collections::{btree_map::Entry, BTreeMap, VecDeque},
-    io::Error,
+    io::{Error, ErrorKind},
     net::SocketAddr,
     os::fd::{AsRawFd, RawFd},
     rc::Rc,
-    time::Instant,
+    time::{Duration, Instant},
 };
 
-use monoio::{io::AsyncWriteRentExt, net::TcpStream, spawn, time::timeout};
+use monoio::{
+    io::AsyncWriteRentExt,
+    net::TcpStream,
+    spawn,
+    time::{sleep, timeout},
+};
 use tokio::sync::mpsc;
 use tracing::{debug, enabled, trace, warn, Level};
 use zerocopy::AsBytes;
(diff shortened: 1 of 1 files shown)


monad-dataplane/src/tcp/tx.rs
```diff
@@ -102,139 +90,124 @@ async fn task_peer(tx_state: TxState, addr: SocketAddr) {
...
+                ErrorKind::Other,
+                format!("error connecting to remote host: {}", err),
+            )
+        })?;
 
-                let duration = Instant::now() - connect_time;
+    trace!(conn_id, ?addr, "outbound TCP connection established");
 
-                let duration_ms = duration.as_millis();
+    conn_cork(stream.as_raw_fd(), true);
```
fabiotdf, line 184: Do we need to uncork somewhere?

buytenh: Uncorking effectively happens when we close the connection -- in this version of transmit pipelining, we send all the messages that have been queued up for transmission, and when no more messages have to be sent, we immediately close the connection.

PR #1600, opened 2025-02-27
Archive: Checker v2 and Fault-Fixer

JoeHowarth:
- Stateless Checker: stores faults and known-good block_num -> replica mappings in S3 for persistance
- Rechecks faults to ensure they are fresh and have not been fixed
- Fault-Fixer manual cli that reads faults and good blocks, copies data from good blocks to fix faults.

Also:
- Removes old checker
- Adds delete to KVStore trait


monad-archive/src/bin/monad-archive-checker/checker.rs
```diff
@@ -0,0 +1,1036 @@
...
+    }
+
+    debug!(
+        block_num,
+        group_count = equivalence_groups.len(),
+        "Found equivalence groups for block"
+    );
+
+    // Find the largest equivalence group - the majority consensus
+    if let Some(largest_group) = equivalence_groups.iter().max_by_key(|group| group.len()) {
```
eerkaijun, line 295: probably rare case but what happens in the event where two groups have same length?

JoeHowarth: Hmm, my first thought was to just take the first group, but on more thinking, let's have the program blow up. This is a truly bad (and rare) situation and that's the loudest thing I can think of.

PR #1605, opened 2025-02-28
Call frame schema change

kkuehlz:
dependent on https://github.com/category-labs/monad/pull/1006

### DIFF
--- a/monad-triedb-utils/src/key.rs
+++ b/monad-triedb-utils/src/key.rs
@@ -34,7 +34,7 @@ pub enum KeyInput<'a> {
     BlockHeader,
     TxHash(&'a [u8; 32]),
     BlockHash(&'a [u8; 32]),
-    CallFrame(Option<u64>),
+    CallFrame,
     BftBlock,
 }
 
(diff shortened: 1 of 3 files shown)


monad-triedb-utils/src/key.rs
```diff
@@ -127,20 +127,45 @@ pub fn create_triedb_key(version: Version, key: KeyInput) -> (Vec<u8>, u8) {
...
+        key_nibbles.push(0);
+    }
+
+    let key: Vec<_> = key_nibbles
+        .chunks(2)
+        .map(|chunk| (chunk[0] << 4) | chunk[1])
+        .collect();
+
+    (key, num_nibbles)
+}
```
andr-dev, line 152: We should put the key_nibbles => (Vec<u8>, u8) process as a helper fn as it's currently duplicated below.

monad-triedb-utils/src/triedb_env.rs
```diff
@@ -1367,3 +1395,46 @@ where
...
+            chunks.keys().collect::<Vec<_>>()
+        ));
+    }
+
+    // concatenate chunks in order
+    let mut complete_call_frame = Vec::new();
+    for (_, chunk_data) in chunks {
+        complete_call_frame.extend(chunk_data);
+    }
+    Ok(complete_call_frame)
```
andr-dev:
```rust
Ok(chunks.into_values().flatten().collect())
```

monad-triedb-utils/src/triedb_env.rs
```diff
@@ -34,20 +34,30 @@ pub type EthTxHash = [u8; 32];
...
 
-enum SyncRequest {
-    TraverseRequest(TraverseRequest),
+struct RangeGetRequest {
+    // a sender for the polling thread to send the result back to the request handler
+    request_sender: oneshot::Sender<Option<Vec<TraverseEntry>>>,
+    // triedb_key and key_len_nibbles are used to read items from triedb
+    prefix_key: Arc<Vec<u8>>,
+    prefix_key_len_nibbles: u8,
+    min_triedb_key: Arc<Vec<u8>>,
```
omegablitz: Can you add a comment for both min/max on if they're inclusive/exclusive?

monad-triedb-utils/src/triedb_env.rs
```diff
@@ -34,20 +34,30 @@ pub type EthTxHash = [u8; 32];
...
+    block_key: BlockKey,
 }
 
 struct TraverseRequest {
     // a sender for the polling thread to send the result back to the request handler
     request_sender: oneshot::Sender<Option<Vec<TraverseEntry>>>,
     // triedb_key and key_len_nibbles are used to read items from triedb
-    triedb_key: Vec<u8>,
+    // Use Arc to keep the key alive until the request is processed by the polling thread
+    triedb_key: Arc<Vec<u8>>,
```
omegablitz: I think this is also broken for AsyncRequest (triedb_async_read).

Is there any reason not to make all these async functions (async_read, async_traverse, async_ranged_get) copy the keys?

monad-triedb-utils/src/triedb_env.rs
eerkaijun: nice, updated

monad-triedb-utils/src/triedb_env.rs
eerkaijun: added to AsyncRequest as well

monad-triedb-utils/src/key.rs
eerkaijun: yeah agree it's bad. planning to do a follow up refactor though, right now there's a lot of duplicate code on creating triedb key :(

monad-triedb-utils/src/triedb_env.rs
omegablitz: Any reason not to copy the keys in triedb-driver? Maintaining the lifetime invariants outside the async functions seems error prone.

eerkaijun: good point, @kkuehlz any thoughts?

eerkaijun: we removed the arc -- it's not needed for async and traverse. for range get, kevin made an update to take in Nibbles instead of NibblesView, so we no longer need arc as well

monad-triedb-utils/src/triedb_env.rs
```diff
@@ -293,21 +304,18 @@ fn polling_thread(
...
+                    triedb_handle.range_get_triedb_async(
+                        &range_request.prefix_key,
+                        range_request.prefix_key_len_nibbles,
+                        &range_request.min_triedb_key,
+                        range_request.min_key_len_nibbles,
+                        &range_request.max_triedb_key,
+                        range_request.max_key_len_nibbles,
+                        range_request.block_key.seq_num().0,
+                        range_request.request_sender,
+                        triedb_async_read_concurrency_tracker.clone(),
```
omegablitz, line 317: Is AsyncRangeGet a traversal?

omegablitz: If so, need to swap the concurrency tracker

eerkaijun: it's technically a traversal but I don't think we should use the traverse concurrency tracker. the common case (especially with our current block gas limit) will only be reading chunk from a single leaf node. I tried to create a test case in flexnet which spans multiple chunks but couldn't get it due to gas limit

PR #1606, opened 2025-02-28
Record txpool create_proposal duration in metrics

andr-dev:


monad-eth-txpool-executor/src/lib.rs
```diff
@@ -175,20 +177,30 @@ where
...
-                                high_qc,
-                                timestamp_ns,
-                                round_signature,
-                                delayed_execution_results,
-                                proposed_execution_inputs,
-                                last_round_tc,
-                            })
-                            .expect("events never dropped"),
+                        Ok(proposed_execution_inputs) => {
+                            let elapsed = Instant::now().duration_since(create_proposal_start);
```
omegablitz: nit: can't you do `let elapsed = create_proposal_start.elapsed()`

PR #1614, opened 2025-03-03
Expand txpool test coverage

andr-dev:


monad-eth-txpool/tests/pool.rs
```diff
@@ -645,6 +698,32 @@ fn test_nonce_exists_in_committed_block() {
     );
 }
 
+#[test]
+#[traced_test]
+fn test_unknown_account() {
+    let tx1 = make_legacy_tx(S1, BASE_FEE, GAS_LIMIT, 0, 10);
```
michael-yxchen, line 704: is the transaction reject because of account nonce? I find it a little confusing if it's rejected because of nonce. txns from unknown account would be rejected because of account balance

andr-dev: It's rejected because of account balance. The nonce of 0 is set correctly. This test is just making sure that the account balance logic works even if the account is not in the block policy, which is what we're setting by providing an empty BTreeMap in the `run_custom`.

PR #1619, opened 2025-03-04
Periodically refresh DNS

michael-yxchen:


monad-updaters/src/config_loader.rs
```diff
@@ -63,93 +68,129 @@ where
...
+                    }
+                    Err(TrySendError::Full(_)) => {
+                        warn!("config update response channel full");
+                    }
+                };
+
+            loop {
+                tokio::select! {
+                    _ = interval.tick() => {
+                        if let Some(update) = peers_table.refresh_dns(){
```
omegablitz: refresh_dns should be in a tokio::spawn_blocking, or to_socket_addrs should use https://docs.rs/tokio/latest/tokio/net/fn.lookup_host.html instead (which wraps spawn_blocking)

michael-yxchen: switched to `tokio::lookup_host`

monad-node/src/main.rs
```diff
@@ -122,6 +122,19 @@ async fn run(node_state: NodeState, reload_handle: ReloadHandle) -> Result<(), (
...
+        .peers
+        .iter()
+        .map(|peer| {
+            (
+                NodeId::new(peer.secp256k1_pubkey),
+                resolve_domain(&peer.address),
+            )
+        })
+        .collect();
+
```
buytenh, line 137: This wasn't specifically introduced by your PR, but this looks like monad-node will panic on start-up if our local resolver (or the peer domain's authoritative DNS server) is not responding, or if any of the peer domain names doesn't exist or points to a resource record set that doesn't include IPv4/v6 address records?

michael-yxchen: yeah that's the current behavior

monad-updaters/src/config_loader.rs
```diff
@@ -202,3 +234,118 @@ where
...
+            .collect::<Vec<_>>();
+        let resolved = Self::resolve_domains(peers).await;
+        for (node_id, maybe_addr) in resolved {
+            // retain old record if hostname fails to resolve
+            if let Some(new_addr) = maybe_addr {
+                let (_, maybe_last_resolved) = self
+                    .peers
+                    .get_mut(&node_id)
+                    .expect("peers table is immutable during lookup");
+                if maybe_last_resolved.is_none()
```
omegablitz: nit: can collapse these two conditions with https://doc.rust-lang.org/std/option/enum.Option.html#method.is_none_or

monad-updaters/src/config_loader.rs
```diff
@@ -202,3 +234,118 @@ where
...
+                if old_hostname == &hostname {
+                    new_table_entry.1 = *maybe_resolved;
+                } else {
+                    debug!(?node_id, ?old_hostname, new_hostname=?hostname, "hostname changed");
+                }
+            }
+        }
+        self.peers = new_table;
+        // always update peers if config is reloaded
+        self.pending_update = true;
```
omegablitz: I don't get why this is needed. Why won't the `*maybe_last_resolved = Some(new_addr);` branch get hit above, setting pending_update = true?

I'm just trying to understand if we can get rid of the pending_update member on PeersTable to simplify things.

michael-yxchen: That's the bug I was running into yday. When peers are removed, `*maybe_last_resolved = Some(new_addr);` branch doesn't get hit. No update is sent to MonadState

Agree that `pending_update` member isn't ideal. I'll replace it with an arg to `refresh_dns` instead

PR #1621, opened 2025-03-05
Hit archive if block hash present but not transactions

eerkaijun:
block hash expires slower than other items in triedb (e.g. transactions, receipts, traces, block headers). if block hash is present but other items such as transactions are not, we should hit archive

### DIFF
--- a/monad-rpc/src/trace_handlers.rs
+++ b/monad-rpc/src/trace_handlers.rs
@@ -201,45 +201,20 @@ pub async fn monad_debug_traceBlockByHash<T: Triedb>(
         .map_err(JsonRpcError::internal_error)?
     {
         let block_key = triedb_env.get_block_key(SeqNum(block_num));
-        let mut resp = Vec::new();
-
-        let tx_ids = triedb_env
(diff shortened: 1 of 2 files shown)


monad-rpc/src/trace_handlers.rs
```diff
@@ -203,32 +203,32 @@ pub async fn monad_debug_traceBlockByHash<T: Triedb>(
...
-            .iter()
-            .map(|tx| *tx.tx.tx_hash())
-            .collect::<Vec<_>>();
-        let call_frames = triedb_env
-            .get_call_frames(block_key)
-            .await
-            .map_err(JsonRpcError::internal_error)?;
+        if let Ok(transactions) = triedb_env.get_transactions(block_key).await {
+            let tx_ids = transactions
+                .iter()
```
andr-dev: `into_iter()`

JoeHowarth: iter() fine here actually since we're just borrowing anyway, but you should remove the collect

PR #1628, opened 2025-03-07
Spring clean dockerfiles

eerkaijun:
Clean up a bunch of unused stuffs and some dockerfiles update - just happen to came across them


michael-yxchen:
Can you check if any of the jenkins pipelines need to be updated? e.g. this pipeline calls `net-run.sh`

eerkaijun:
great point, removed this integration test pipeline which is not running anymore https://github.com/category-labs/category-infra/pull/1041

michael-yxchen:
do you still use that docker-compose? I wonder if we can remove flexnet from monad-bft altogether

eerkaijun:
yeah i still use it quite frequently, easiest way for me to spin up a node for debugging and make modification. i think it's worth leaving it as it's a simple one-command setup for starting a single node, some MF folks are also using it

docker/flexnet/nets/run.sh
eclark0426, line 1: A part of me wonders a bit if this should still be called flexnet, or if there might be some other more appropriate name, since at this point it's just a way to run a single node with all of its components. Also, since a single node is the only option here, it might make sense to just move `compose.yaml` and `run.sh` into the `flexnet` directory instead of keeping the extra `nets` directory, unless you're keeping that there to avoid changing the workflow for people

eerkaijun: keeping a `nets` subdirectory so that we can just copy the entire subdirectory rather than copying the individual files

eclark0426: Do we really even need to copy the entire subdirectory any more? It used to have a bunch of scripts that were run in containers, so it made sense to copy it. But at this point, none of those things exist, so you probably don't even need to copy the entirety of the `nets` directory to the output directory

PR #1629, opened 2025-03-07
Add test for rebroadcasting all symbols

bamrith:


monad-raptorcast/tests/raptorcast_instance.rs
```diff
@@ -81,6 +88,12 @@ pub fn different_symbol_sizes() {
             tx_socket
                 .send_to(&messages[0].1[0..usize::from(segment_size)], messages[0].0)
                 .unwrap();
+
+            // Wait for rebroadcasting activity to complete.
+            std::thread::sleep(Duration::from_millis(10));
+
+            // Verify that the rebroadcast target receives the first symbol.
+            let _ = rebroadcast_socket.recv(&mut []).unwrap();
```
buytenh: Can you move this to outside the `for i in 0..=1` loop?
That is, leave the loop as-is, and then verify the reception before where you added the check that there are no more receptions.

monad-raptorcast/tests/raptorcast_instance.rs
```diff
@@ -225,13 +244,12 @@ pub fn zero_sized_packet() {
     std::thread::sleep(Duration::from_millis(100));
 }
 
-// Verify that received encoded symbols are only rebroadcast when they are received for
-// the first time.
+// Verify that all received encoded symbols are rebroadcast exactly once
```
buytenh: Add "valid" to this description?  Otherwise you're not really changing the meaning of the text, while you are changing what the test does.

monad-raptorcast/tests/raptorcast_instance.rs
```diff
@@ -259,13 +277,36 @@ pub fn duplicate_rebroadcast() {
...
+
+    // Verify that the rebroadcast target has nothing more to receive.
+    assert_eq!(
+        rebroadcast_socket.recv(&mut []).unwrap_err().kind(),
+        ErrorKind::WouldBlock
+    );
+
     // Send 5 copies of the first symbol of the first message.
     for _ in 0..5 {
         tx_socket
```
buytenh: Can you introduce a `for i in 0..=1` loop like in the first test, where in the loop body you perform all the `send_to()` calls, and then depending on whether i is 0 or 1 either verify that all chunks were received or verify that nothing more was received?  If you do it that way, you can remove the `for _ in 0..5` logic and the other existing code from this test.

PR #1632, opened 2025-03-08
Improve statesync response encoding performance

omegablitz:
- serialize statesync upserts as str instead of list
- implement alloy_rlp::Encodable::length for Encodable types


omegablitz:
Takes ~70ms to serialize 1M upserts now, down from ~1.3s on master.

I don't see any timeouts on stressnet now while nodes are statesyncing.

PR #1633, opened 2025-03-09
ethcall via new ffi 

ariqchowdhury:


Cargo.lock
omegablitz, line 1: Bumping all these versions intentional?

monad-cxx/monad-execution
omegablitz, line 1: Can you delete monad-cxx/include, monad-cxx/src, monad-cxx/build.rs, monad-cxx/Cargo.toml, monad-cxx/CMakeList.txt, and remove monad-cxx from Cargo.toml?

That shouldn't break any build.rs - we can lift monad-cxx/monad-execution up to the top-level later.

docker/rpc/Dockerfile
```diff
@@ -74,8 +74,8 @@ RUN ASMFLAGS="-march=haswell" CFLAGS="-march=haswell" CXXFLAGS="-march=haswell -
     cp `ls -Lt $(find target/release | grep -e "libkeccak.so") | awk -F/ '!seen[$NF]++'` . && \
     cp `ls -Lt $(find target/release | grep -e "libsecp256k1.so") | awk -F/ '!seen[$NF]++'` . && \
     cp `ls -Lt $(find target/release | grep -e "libsilkpre.so") | awk -F/ '!seen[$NF]++'` . && \
-    cp `ls -Lt $(find target/release | grep -e "lib${TRIEDB_TARGET}.so") | awk -F/ '!seen[$NF]++'` . && \
-    cp `ls -Lt $(find target/release | grep -e "lib${ETH_CALL_TARGET}.so") | awk -F/ '!seen[$NF]++'` .
+    cp `ls -Lt $(find target/release | grep -e "libmonad_rpc.so") | awk -F/ '!seen[$NF]++'` . && \
+    cp `ls -Lt $(find target/release | grep -e "lib${TRIEDB_TARGET}.so") | awk -F/ '!seen[$NF]++'` .
```
omegablitz: Remove ETH_CALL_TARGET arg above

Also, I think you can change `ARG TRIEDB_TARGET=triedb_driver` -> `ENV TRIEDB_TARGET=triedb_driver`, and change this line to: `grep -e "libtriedb_driver.so"`.

Can ctrl-f and delete any ETH_CALL_TARGET, monad_rpc_test, 
(mock targets won't work anymore)

And then remove associated ETH_CALL_TARGET/TRIEDB_TARGET build args in monad-integration

monad-rpc/src/gas_handlers.rs
```diff
@@ -41,22 +45,24 @@ impl GasEstimator {
...
         }
     }
 }
 
 impl EthCallProvider for GasEstimator {
-    async fn eth_call(&self, txn: TxEnvelope) -> CallResult {
+    async fn eth_call(
+        &self,
+        txn: TxEnvelope,
+        eth_call_executor: Option<Arc<Mutex<EthCallExecutor>>>,
```
omegablitz, line 64: Why take option if it unwraps?

monad-rpc/src/cli.rs
```diff
@@ -54,7 +54,7 @@ pub struct Cli {
     pub eth_get_logs_max_block_range: u64,
 
     /// Set the max concurrent requests for eth_call and eth_estimateGas
-    #[arg(long, default_value_t = 20)]
+    #[arg(long, default_value_t = 2000)]
```
omegablitz: Is this a safe default? BlockHashBuffer can't get thrashed?

monad-ethcall/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+    let mut rlp_encoded_tx = vec![];
+    transaction.encode_2718(&mut rlp_encoded_tx);
+
+    let mut rlp_encoded_block_header = vec![];
+    block_header.encode(&mut rlp_encoded_block_header);
+
+    let mut rlp_encoded_sender = vec![];
+    sender.encode(&mut rlp_encoded_sender);
+
+    let override_ctx = unsafe { bindings::monad_state_override_create() };
```
omegablitz, line 143: I think ideally this would be wrapped in a rust struct with an associated Drop impl that calls destroy, but we can do that later.

Seems easy to accidentally forget to call destroy if we add a new err condition.

monad-ethcall/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+                bindings::set_override_code(
+                    override_ctx,
+                    addr.as_ptr(),
+                    addr.len(),
+                    code.as_ptr(),
+                    code.len(),
+                )
+            }
+
+            if let Some(StorageOverride::State(storage_override)) = &obj.storage_override {
```
omegablitz: match for exhaustiveness checking

monad-ethcall/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+    let mut rlp_encoded_block_header = vec![];
+    block_header.encode(&mut rlp_encoded_block_header);
+
+    let mut rlp_encoded_sender = vec![];
+    sender.encode(&mut rlp_encoded_sender);
+
+    let override_ctx = unsafe { bindings::monad_state_override_create() };
+    for (addr, obj) in state_override_set {
+        unsafe {
+            bindings::add_override_address(override_ctx, addr.as_ptr(), addr.len());
```
omegablitz, line 148: Just checking - is it safe if the StateOverride is all None?

monad-ethcall/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+                }
+            }
+        }
+    }
+
+    let chain_id = match chain_id {
+        1 => bindings::monad_eth_call_chain_ETH_MAINNET,
+        20143 => bindings::monad_eth_call_chain_MONAD_DEVNET,
+        10143 => bindings::monad_eth_call_chain_MONAD_TESTNET,
+        _ => {
```
omegablitz, line 218: Missing monad_state_override_destroy

monad-ethcall/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+            rlp_encoded_tx.as_ptr(),
+            rlp_encoded_tx.len(),
+            rlp_encoded_block_header.as_ptr(),
+            rlp_encoded_block_header.len(),
+            rlp_encoded_sender.as_ptr(),
+            rlp_encoded_sender.len(),
+            block_number,
+            block_round.unwrap_or(u64::MAX),
+            override_ctx,
+            Some(eth_call_submit_callback),
```
omegablitz, line 250: Just checking - it is guaranteed that eth_call_submit_callback is called by execution?

monad-ethcall/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+            sender_ctx_ptr as *mut std::ffi::c_void,
+        )
+    };
+
+    let Ok(result) = recv.await else {
```
omegablitz: This branch shouldn't be possible, correct? Maybe assert or log a loud error if it gets here - I think getting here implies a memory leak.

monad-rpc/src/gas_handlers.rs
ariqchowdhury: temp workaround for the gas_handler tests. the mock thing can get none passed in and its impl doesn't use it. There's definitely something to be refactored here but that can be deferred to later.

monad-ethcall/src/lib.rs
ariqchowdhury: pretty sure, but will confirm with a unit test on eth-call execution side

monad-rpc/src/cli.rs
ariqchowdhury: back to 20 after discussion

monad-ethcall/src/lib.rs
ariqchowdhury: confirmed fine

Cargo.lock
ariqchowdhury: nope, reverted

monad-ethcall/src/lib.rs
```diff
@@ -0,0 +1,346 @@
...
+    // https://github.com/ethereum/execution-apis/blob/main/tests/eth_call/call-revert-abi-error.io
+    // if there is an error message to be decoded, output_data will be the following form:
+    // 4 bytes function signature
+    // 32 bytes data offset
+    // 32 bytes error message length (let's call it x)
+    // x bytes error message (padded to multiple of 32 bytes)
+    let message_start_index = 68_usize;
+    if output_data.len() > message_start_index {
+        // we only return the first 256 bytes of the error message
+        let message_length = output_data[message_start_index - 1] as usize;
```
michael-yxchen, line 345: I think you can examine the other 31 bytes and set message_length to 256 if any bit is set. Getting only the last byte may over truncate if error message is over 256 bytes

monad-ethcall/src/lib.rs
```diff
@@ -0,0 +1,346 @@
...
+        let message_end_index = message_start_index + message_length;
+        if output_data.len() >= message_end_index {
+            // extract the message bytes
+            let message_bytes = &output_data[message_start_index..message_end_index];
+
```
michael-yxchen, line 350: `data_offset` is ignored here. Just confirming that's intentional

monad-ethcall/src/lib.rs
ariqchowdhury: this function is just moved from the existing monad-cxx crate. Can fix things up in a later PR

monad-ethcall/src/lib.rs
ariqchowdhury: not sure, can look at separately

monad-ethcall/src/lib.rs
eerkaijun: maybe we can just use https://docs.rs/alloy/latest/alloy/sol_types/fn.decode_revert_reason.html 

i can clean this up in a subsequent PR

PR #1634, opened 2025-03-10
Record txpool create_proposal lookup metrics

andr-dev:


monad-eth-txpool-executor/src/lib.rs
```diff
@@ -40,7 +41,7 @@ where
 {
     ipc: Pin<Box<EthTxPoolIpcServer<ST, SCT, SBT>>>,
     block_policy: EthBlockPolicy<ST, SCT>,
-    state_backend: SBT,
+    state_backend: StateBackendCache<StateBackendMetricsWrapper<SBT>>,
```
omegablitz: I don't love how this has a specialized SBT here - I think it's a little confusing how the state_backend used here doesn't match the one used in consensus-state.

But I haven't thought of a good alternative either...

andr-dev: Yeah I'm also not a fan but there doesn't seem to be another "simple" way of recording the lookups, especially because this means the tests need to change which is sad. Another option is to make a specialized impl for when SBT is StateBackendCache<StateBackendMetricsWrapper<SBT>> but that just introduces code dupe.

PR #1640, opened 2025-03-11
Archive & Rpc: Support fast indexed eth_getLogs

JoeHowarth:
Create index to support eth_getLogs queries in Mongo backed archive
- Uses 32-bit prefixes of address and topics to filter logs
- When matched, returns TxIndexedData which rpc then does exhaustive
  filtering over the reduced set, transforms the data and returns.
- Also includes benchmarking tool found in monad-archive/src/examples
- Only use the index if the query constrains the result set, otherwise
  use bulk block fetching

4/21 Update: 
Benchmarks with indexes not in RAM: https://www.notion.so/Logs-Index-Benchmarks-1bc75b0ba8408047bd32da368d0c6b52
On the machine used for benchmarking (can-009), MongoDB only used ~60gb of RAM to cache indexes and documents, while the total size of all indexes was ~160gb. This caused a slowdown in the above benchmarks. Warming the cache resulted in ~5x improvement for this solution over the baseline according to Misha's testing. Real RPC provider db hosts generally have much larger amounts of RAM (QN reportedly has ~1 TB of RAM), so this should be much faster in practice. Even if this is not true, the above benchmarks show that non-cached indexes are still faster than baseline for operations where the result set is constrained and only somewhat slower for un-filtered (i.e. whole block) queries. The the fn in rpc now only uses the index if the query is constrained (i.e. some logs will be filtered out)

Initial Draft: 

> This pr creates a new collection in mongodb for log indexes. 
> 
> The `LogIndexDocument` is a compressed representation of a log's block number, address, and topics similar to a bloom filter, but that works natively with mongodb indexes. It uses 32bit prefixes of address and topics to reduce space usage while keeping a less then 1% probability of false positives for 1 log per tx, 10k tps and 10k block range assuming 1 topic queried and a uniform distribution of topic hashes. 
> 
> During tx indexing, this representation is created for each log in a block and inserted into mongo if we're running a mongo backend (nothing is done otherwise). 
> 
> The archive library supports querying for known from and to block numbers and a set of addresses, and possibly non-empty sets of topics. This returns a superset of the matching logs. We then use the included tx hash to join with the tx_index table.
> 
> In the rpc server, we use the joined data (TxIndexData) to construct TransactionReceipts, then use the same filtering logic to ensure that no false positives make it into the response. If the log index collection is not able to be used, we fallback on the original non-indexed method.
> 
> To ensure no rpc regressions, this feature is gated by 2 flags:
> - `use_eth_get_logs_index`: Return the results of hitting the index without using the original slow path. Eventually this will be the default.
> - `dry_run_get_logs_index`: Hit the index and also hit the original slow path. Return the original data, but compare the index data against it and log if there is a mismatch. This allows us to run the index in production without worrying about regressions. 
> 
> If both of the above flags are false, the index will not be hit at all. This allows us to revert to original behavior if this code path results in a crash.

Todo: 
- [x] Implement index using mongo
- [x] Integration tests using docker with small volume of testnet data
- [x] Add to index worker
- [x] Integrate query into rpc along with filtering and fallback
- [x] Test deployment on testnet and gather coarse grain benchmark results
- [ ] Rollout with dry run flag enabled and soak
- [ ] Rollout with use eth get logs enabled


monad-rpc/src/cli.rs
```diff
@@ -102,6 +102,12 @@ pub struct Cli {
     #[arg(long)]
     pub mongo_db_name: Option<String>,
 
+    #[arg(long)]
+    pub use_eth_get_logs_index: bool,
+
+    #[arg(long)]
+    pub dry_run_get_logs_index: bool,
```
eerkaijun: Add comments here so it will show up in the cli `--help` option

monad-archive/src/model/logs_index.rs
```diff
@@ -0,0 +1,354 @@
...
+            // Create capped collection if it doesn't exist
+            db.create_collection(COLLECTION_NAME).await?;
+
+            let logs_collection = db.collection::<LogIndexDocument>(COLLECTION_NAME);
+            logs_collection
+                .create_index(
+                    IndexModel::builder()
+                        .keys(
+                            doc! { "block_number": 1, "address": 1, "topic_0": 1, "topic_1": 1, "topic_2": 1, "topic_3": 1 },
+                        )
```
eerkaijun: is this a cap on the number of topics that the index can support?

monad-rpc/src/eth_txn_handlers.rs
```diff
@@ -373,11 +397,95 @@ pub async fn monad_eth_getLogs<T: Triedb>(
...
         receipt_logs.into_iter().map(MonadLog).collect(),
     ))
 }
 
+async fn get_logs_with_index(
+    reader: &Option<ArchiveReader>,
+    from_block: u64,
+    to_block: u64,
+    filter: &Filter,
+) -> monad_archive::prelude::Result<Vec<Log>> {
```
eerkaijun, line 497: do we need to import Result type from monad_archive

monad-archive/src/model/logs_index.rs
```diff
@@ -0,0 +1,354 @@
...
+    pub async fn hydrate_logs(
+        &self,
+        logs_index: Vec<LogIndexDocument>,
+        tx_index: &impl IndexReader,
+    ) -> Result<HashMap<TxHash, TxIndexedData>> {
+        let tx_hashes = logs_index
+            .iter()
+            .map(|l| l.tx_hash)
+            .collect::<Vec<TxHash>>();
+        tx_index.get_tx_indexed_data_bulk(&tx_hashes).await
```
eerkaijun: if a query asks for a 100 block range without any address or topic filter, and assuming there's 500 transactions in a block, this essentially do 50k reads from the tx index table? any concerns around that?

monad-rpc/src/eth_txn_handlers.rs
```diff
@@ -373,11 +397,95 @@ pub async fn monad_eth_getLogs<T: Triedb>(
         })
         .collect();
 
+    let mut eq = true;
+    // if dry_run_get_logs_index is true and index returned successfully
+    // we can compare the index results to the original method to ensure they match
+    if let Some(index_results) = index_results {
+        let mut index_results = HashSet::<Log>::from_iter(index_results);
```
eerkaijun: are we going to be running this in prod? isn't that going to slow down eth_getLogs since now we need to fetch from the mongodb index and also triedb

monad-archive/src/model/logs_index.rs
JoeHowarth: I believe there are only 4 topics per log max right?

monad-rpc/src/eth_txn_handlers.rs
JoeHowarth: It makes it cleaner to compose errors in these helpers. I used qualified imports to avoid cluttering the namespace and keeping them scoped to fns that interact with archive extensively. 

We also never send these errors as a response, they're only ever logged internally.

monad-archive/src/model/logs_index.rs
JoeHowarth: Good point - new revision should address this

I didn't add it because the original code didn't have it, but we could also short-circuit the query once we've hit `max_response_size`

monad-rpc/src/eth_txn_handlers.rs
JoeHowarth: New revision ensures there's no perf cost if `dry-run` is disabled.

When `dry-run` is enabled, it does fetching and comparison in a separate task to minimize resp latency. It does copy the response into an owned HashSet in the main task, but this seems ok since it's explicitly behind the `dry-run` flag.

monad-archive/src/model/logs_index.rs
```diff
@@ -0,0 +1,502 @@
...
+        if !db
+            .list_collection_names()
+            .await?
+            .contains(&COLLECTION_NAME.to_string())
+        {
+            // If we aren't creating the collection and it's not found, error
+            if !create {
+                bail!("Collection not found: {COLLECTION_NAME}");
+            };
+            // Create capped collection if it doesn't exist
```
mikety: nit: the created collection is not capped. I dont think it should be, either update or remove (the code is clear) the comment.

monad-archive/src/model/logs_index.rs
```diff
@@ -0,0 +1,502 @@
...
+            .iter()
+            .zip(&block.body.transactions)
+            .flat_map(|(receipt, tx)| {
+                let tx_hash = *tx.tx.tx_hash();
+                let logs = receipt.receipt.logs();
+
+                logs.iter().map(move |log| {
+                    let mut this = Self {
+                        block_number,
+                        tx_hash,
```
mikety: It should be the first document  since it is a unique _id key

monad-archive/src/model/logs_index.rs
```diff
@@ -0,0 +1,502 @@
...
+    {
+        let stream = self
+            .query_logs_index_for_tx_hash(from, to, addresses, topics)
+            .await?;
+        Ok(self.hydrate_logs(stream, tx_index).await)
+    }
+
+    /// Join tx_hash from log index query with tx_indexed_data
+    /// Returns a stream with join db queries batched into chunks of 500
+    pub async fn hydrate_logs<'a>(
```
mikety: nit: call it join_tx_index ?
Also have you considered to create a view that joins log_index and tx_index by tx_has?

mikety: Is it possible to keep all information in one mongodb collection (i.e. add fields `address` and topics ) and just create an index? If both collections PK is tx_hash it will be more efficient.

monad-archive/src/model/logs_index.rs
eerkaijun: yeah i think you're right, only max of three indexed topics for a log

eerkaijun:
My remaining thoughts: 
1. Feels like when we merge this to master we should remove the `dry_run_get_logs_index` from eth_getLogs as it feels more like test code. I think it would be ideal if we could make this into a unit/flexnet test instead
2. We now fetch each transaction individually (through `TxIndexedData`) rather than getting the entire block. For large block ranges (with many txs in each block) I feel like this would impact performance, even with short circuit. For example, the default response size limit is currently set to 25MB, and if each `TxIndexedData` takes 2000 bytes, that's still a lot of read requests. Feel free to correct me if my calculation is wrong. 

The above points doesn't block us from testing these changes on a live testnet RPC node and collecting some data though, so I'm good with deploying a patch as is for testing this new feature.

JoeHowarth:
> My remaining thoughts:
> 
> 1. Feels like when we merge this to master we should remove the `dry_run_get_logs_index` from eth_getLogs as it feels more like test code. I think it would be ideal if we could make this into a unit/flexnet test instead
> 2. We now fetch each transaction individually (through `TxIndexedData`) rather than getting the entire block. For large block ranges (with many txs in each block) I feel like this would impact performance, even with short circuit. For example, the default response size limit is currently set to 25MB, and if each `TxIndexedData` takes 2000 bytes, that's still a lot of read requests. Feel free to correct me if my calculation is wrong.
> 
> The above points doesn't block us from testing these changes on a live testnet RPC node and collecting some data though, so I'm good with deploying a patch as is for testing this new feature.

1. `dry run` is certainly an insurance feature. I find it simpler to include experimental (but still tested!) features into main rather than juggling multiple patch images / branches. This allows for flexibility with rollout to 3rd parties and simplifies hot-fixes / general upgrades. When the feature has been baked sufficiently, we simply switch the default and deprecate the flag, then completely remove the flag a release later. That said, I'm open to arguments against this approach. 
2. You're right, if we're running queries where we expect most logs in a block to be returned, this approach is slower. Perhaps we can detect this case more simply by only invoking the index if there are topics or address set? If they are not set, use the existing block-based approach. Thoughts?

monad-archive/src/model/logs_index.rs
JoeHowarth: There will be one `LogIndexDocument` per log emitted, but there can be multiple logs per tx. Perhaps there's a way to have multiple nested indexed fields in the tx object, but having a separate table seems simpler as a non-expert. 

What is the advantage of a view vs imperatively performing the join here? Scanning the docs indicate a performance improvement but I could be missing something.

monad-archive/src/model/logs_index.rs
JoeHowarth: Could you clarify what you mean here? 

My understanding was _id would be generated by mongo as a UUID. None of the fields in this document are guaranteed to be unique, so that's why I didn't set any of them as the primary key.

mikety: `_id` must by a unique field in the collection, so mongo will generate it if its not set. However if there is a unique field it can be used as an _id. If non of the fields are uniqu ( i thought that `tx_hash` is unique ) than its fine as is.

monad-archive/src/model/logs_index.rs
mikety: Ok thank you for the clarification. View will use an aggregation pipeline query ( this is a different query subsystem in mongo ) and it can do things like predicate push down based on the query shape. The fastest "join" is when all data is already in the same collection even if its denormalized.

mikety:
Here are some schema alternatives utilizing mongodb backend features

1. cluster `tx_index` collection: it will eliminate extra lookup via PK index - small gain, minimal changes:
2. nest logs_index in the tx_index 
```
{
  _id: 'a760185ba2021bf8b453dc7e9886c2f0784f2ec6c83e20578747af81e1801560',
  value: Binary.createFromBase64('MgH5A4D...Dt0AA==', 0)
  logs: [
     {
         block_number: Long('8373135'),
         address: Binary.createFromBase64('+Bclfw==', 0),
         tx_hash: Binary.createFromBase64('xqickRNsxD9kPed5YdyQ2YU8SOWqDymVPayj2dS2jsw=', 0),
         topic_0: Binary.createFromBase64('3fJSrQ==', 0),
         topic_1: Binary.createFromBase64('AAAAAA==', 0),
         topic_2: Binary.createFromBase64('AAAAAA==', 0),
         topic_3: null
    },
     {
         block_number: Long('8373135'),
         address: Binary.createFromBase64('+Bclfw==', 0),
         tx_hash: Binary.createFromBase64('xqickRNsxD9kPed5YdyQ2YU8SOWqDymVPayj2dS2jsw=', 0),
         topic_0: Binary.createFromBase64('3fJSrQ==', 0),
         topic_1: Binary.createFromBase64('AAAAAA==', 0),
         topic_2: Binary.createFromBase64('AAAAAA==', 0),
         topic_3: null
    }
  ] 
}
```
the index can be created as `db.tx_index.createIndex({logs.address: 1, logs.topic_0:1, logs.topic_1:1, logs.topic_2:1, logs.topic_03:1})`
This can be a significant gain in the lookup performance and also will account for code simplification as no join implementation is needed.

another more compact implementation possible iff block_number and address have 1:1 relation to tx_hash 
```
{
  _id: 'a760185ba2021bf8b453dc7e9886c2f0784f2ec6c83e20578747af81e1801560',
  value: Binary.createFromBase64('MgH5A4D...Dt0AA==', 0)
  block_number: Long('8373135'),
  address: Binary.createFromBase64('+Bclfw==', 0),
  topics: [
    Binary.createFromBase64('3fJSrQ==', 0), Binary.createFromBase64('AAAAAA==', 0)
  ] 
}
```
1. and 2. can be combined.

mikety:
As we discussed in slack the query plan for the queries using the new index is correct. To further improve the performance the size of ram available for mongo instance should be large enough to keep at least the frequently accessed portion of the index in memory which will very significantly improve latency for the range scans that involve multiple blocks.

monad-rpc/src/eth_txn_handlers.rs
```diff
@@ -229,6 +242,36 @@ pub async fn monad_eth_getLogs<T: Triedb>(
         return Err(FilterError::RangeTooLarge.into());
     }
 
+    let latest_block = get_block_key_from_tag(triedb_env, BlockTags::Latest)
+        .seq_num()
+        .0;
+    // TODO: pipe in actual cache size
+    const DEFAULT_CACHE_SIZE: u64 = 200;
```
eerkaijun: address this todo to get from cli?

JoeHowarth:
fixes https://github.com/category-labs/category-internal/issues/1529

PR #1643, opened 2025-03-12
Clean up revert reason decode

eerkaijun:
use alloy revert reason helper function instead

### DIFF
--- a/monad-ethcall/src/lib.rs
+++ b/monad-ethcall/src/lib.rs
@@ -9,6 +9,7 @@ use alloy_consensus::{Header, Transaction as _, TxEnvelope};
 use alloy_eips::eip2718::Encodable2718;
 use alloy_primitives::{Address, Bytes, B256, U256, U64};
 use alloy_rlp::Encodable;
+use alloy_sol_types::decode_revert_reason;
 use bindings::monad_eth_call_result;
 use futures::channel::oneshot::{channel, Sender};
 use serde::{Deserialize, Serialize};
@@ -297,9 +298,12 @@ pub async fn eth_call(
                     };
 
                     let message = String::from("execution reverted");
-                    let error_message = decode_revert_message(&output_data);
+                    let formatted_message = match decode_revert_message(&output_data) {
+                        Some(error_message) => format!("{}: {}", message, error_message),
+                        None => message,
+                    };
                     CallResult::Failure(FailureCallResult {
-                        message: message + ": " + &error_message,
+                        message: formatted_message,
                         data: Some(format!("0x{}", hex::encode(&output_data))),
                     })
                 } else {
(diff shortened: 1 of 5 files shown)


omegablitz:
The "strip_prefix" seem jank, is this better than the existing impl?

eerkaijun:
current implementation doesn't parse some edge cases well, so i think it's better to use the helper function by alloy. however they add `revert: ` or `panic: ` to their error message, which i think we would want to conform to geth instead (which uses `execution reverted: `), therefore stripping these prefixes if exist

omegablitz:
Isn't this alloy implementation dependent? Feels brittle in the case they change their error messages.. but maybe it's fine if we test all the possible error messages. But at that point, don't know if there's much advantage over having our own implementation.

eerkaijun:
we do have unit tests and also flexnet tests for different types of error, so i'm pretty sure we'll catch it if they change the implementation. i think the main motivation is that there are a lot of edge cases and they also do extra validation, so feel that it's better + easier maintainability as well

monad-ethcall/src/lib.rs
```diff
@@ -331,26 +335,43 @@ pub async fn eth_call(
...
-        // we only return the first 256 bytes of the error message
-        let message_length = output_data[message_start_index - 1] as usize;
-        let message_end_index = message_start_index + message_length;
-        if output_data.len() >= message_end_index {
-            // extract the message bytes
-            let message_bytes = &output_data[message_start_index..message_end_index];
-
-            // attempt to decode the message bytes as UTF-8
-            return String::from_utf8(message_bytes.to_vec()).unwrap_or_default();
+    decode_revert_reason(output_data).and_then(|message| {
```
andr-dev, line 340: Here's the definition for `decode_revert_reason`.
```rust
pub fn decode_revert_reason(out: &[u8]) -> Option<String> {
    RevertReason::decode(out).map(|x| x.to_string())
}
```
If we call `RevertReason::decode`, we can then match on the enum rather than doing janky string stripping.

eerkaijun: yes but there are many subtypes, each having different string parsing which imo we shouldn't need to introduce to our codebase

andr-dev: Not sure what you mean by the different string parsing, but after digging through the docs I believe something like the below should work nicely.
```rust
match RevertReason::decode(output_data) {
    RevertReason::RawString(reason) => reason,
    RevertReason::ContractError(ContractError::Custom(custom)) => custom.to_string(),
    RevertReason::ContractError(ContractError::Revert(revert)) => revert.reason().to_owned(),
    RevertReason::ContractError(ContractError::Panic(panic)) => {
      let reason = panic.to_string();
      reason.strip_prefix("panic: ").map(String::to_string).unwrap_or(reason)
    },
}
```
This way, `"revert: "` / `"panic: "` prefixes in the `RawString` or `ContractError::Custom` cases won't be stripped while avoiding the `ContractError::Revert` `"revert: "` strip.

eerkaijun: we would want to strip all panic and revert. also i think the main concern was just future regression, which wouldn't make much difference as long as we are using their library

PR #1644, opened 2025-03-12
Rpc uses txpool ipc socket to decide if bft is syncing

philipglazman:
RPC's `eth_syncing` method uses the status of the ipc client connection to bft to decide whether to return true or false. 

Bft will bind to the txpool socket after TxPoolCommand::Reset which is called after statesync and blocksync are complete. The `EthTxPoolExecutor`'s ipc connection is an enum with a Ready and NotReady variants. When executing TxPoolCommand::Reset when NotReady, bft will bind to the socket and switch to Ready.

Flexnet test to demonstrate behavior: https://github.com/category-labs/monad-integration/pull/234

Fixes https://github.com/category-labs/category-internal/issues/1135

### DIFF
--- a/monad-rpc/src/account_handlers.rs
+++ b/monad-rpc/src/account_handlers.rs
@@ -114,14 +114,10 @@ pub async fn monad_eth_getTransactionCount<T: Triedb>(
 
 #[allow(non_snake_case)]
 /// Returns an object with data about the sync status or false.
-pub async fn monad_eth_syncing(
-    mempool_state: &Arc<EthTxPoolBridgeState>,
-) -> Result<Value, JsonRpcError> {
+pub async fn monad_eth_syncing() -> Result<Value, JsonRpcError> {
     trace!("monad_eth_syncing");
 
-    let is_connected = mempool_state.is_connected().await;
-
-    serialize_result(serde_json::Value::Bool(!is_connected))
+    serialize_result(serde_json::Value::Bool(false))
 }
 
 #[derive(Deserialize, Debug, schemars::JsonSchema)]
--- a/monad-rpc/src/main.rs
+++ b/monad-rpc/src/main.rs
@@ -409,7 +409,7 @@ async fn rpc_select(
         "eth_chainId" => monad_eth_chainId(app_state.chain_id)
             .await
             .map(serialize_result)?,
-        "eth_syncing" => monad_eth_syncing(&app_state.mempool_state).await,
+        "eth_syncing" => monad_eth_syncing().await,
         "eth_estimateGas" => {
             let Some(triedb_env) = &app_state.triedb_reader else {
                 return Err(JsonRpcError::method_not_supported());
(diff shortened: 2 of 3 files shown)


monad-rpc/src/account_handlers.rs
```diff
@@ -114,8 +117,6 @@ pub async fn monad_eth_getTransactionCount<T: Triedb>(
 pub async fn monad_eth_syncing() -> Result<Value, JsonRpcError> {
     trace!("monad_eth_syncing");
 
-    // TODO. TBD where this data actually comes from
-
     serialize_result(serde_json::Value::Bool(false))
```
eerkaijun, line 120: does this need to be updated?

monad-rpc/src/main.rs
```diff
@@ -673,17 +673,21 @@ async fn main() -> std::io::Result<()> {
     );
     let txpool_state = EthTxPoolBridgeState::new();
 
+
+    // Wait for bft to be in a ready state before starting the RPC server.
+    // Bft will bind to the ipc socket after state syncing.
+    let ipc_path = args.ipc_path;
+    let mut txpool_bridge = retry(
+        || async { EthTxPoolBridge::new(&ipc_path, txpool_state.clone()).await },
+        30,
```
eerkaijun, line 682: how do we determine this 30 retries? if statesync is not complete within this timeframe, I presume rpc will not start?

philipglazman: It's arbitrary. I'm not sure what the p90 statesync time is

philipglazman:
The `eth_sync` issue is causing a lot of operator pain so the intention was to re-use some of the existing scaffolding that we have to fix it immediately. 

The feedback is fair. I think there are couple of alternative approaches: make the ipc socket between bft <> rpc more generalized, use a separate communication system entirely to notify statesync completeness, or use the upcoming events system.  I'm leaning towards using the events system since its both lower lift and less error prone.

andr-dev:
Agreed, using the event system seems like the right way forward although I'm not entirely sure how we would get the statesync status since my understanding is that the event system only produces "execution" info. There's also a ton of other RPC mechanisms that would greatly benefit from tracking event system state so we'll likely add it to RPC sooner or later.

I think it's just tough because the events system doesn't seem to have a clear timeline to merge so we could be waiting anywhere from days to weeks to have that available.

andr-dev:
Given the unclear timeline for the event system, if we do need an immediate fix for this, I'll suggest a change that has a much smaller surface area and should be a couple lines to add/remove, so we can rip it out easily after the event system is merged.

We can add a `bool` flag to the txpool executor which indicates whether it has been `reset` before. In the `poll`, we then skip polling the ipc struct if that flag isn't true. Since we're using unix domain sockets for IPC, I believe each connection must be explicitly accepted on the receiving (server aka bft) side so by skipping that `poll` we'll cause a connection timeout on the RPC side. This avoids doing any of the nasty projection stuff and should at most a 20 line change.

philipglazman:
Replaced with https://github.com/category-labs/monad-bft/pull/1681

PR #1647, opened 2025-03-13
Add historical block requests to rpc-request-generator 

eclark0426:
Opening as draft since Kai's PR is still open and this is based on that

### DIFF
--- a/monad-eth-testutil/examples/rpc-request-generator/main.rs
+++ b/monad-eth-testutil/examples/rpc-request-generator/main.rs
@@ -15,6 +15,7 @@ use alloy_rpc_types_trace::geth::{
 use clap::Parser;
 use futures::{stream::FuturesUnordered, StreamExt};
 use itertools::Itertools;
+use rand::seq::IteratorRandom;
 use tokio::{
     task::JoinHandle,
     time::{Duration, Instant},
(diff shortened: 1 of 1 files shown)


monad-eth-testutil/examples/rpc-request-generator/main.rs
```diff
@@ -146,15 +154,65 @@ struct BlockIndexError {
...
+    }
+
+    async fn process_new_block(
+        &mut self,
+        block_number: BlockNumber,
+        index_done_sender: tokio::sync::mpsc::Sender<
+            Result<(BlockNumber, Vec<Address>), BlockIndexError>,
+        >,
+    ) {
+        for _ in 0..self.requests_per_new_block {
```
eerkaijun: why do we need this for loop? I think we can just pass `requests_per_new_block` into `index_block`

monad-eth-testutil/examples/rpc-request-generator/main.rs
```diff
@@ -146,15 +154,65 @@ struct BlockIndexError {
...
+        self.recent_blocks.push_back(block_number);
+        if self.recent_blocks.len() > self.historical_block_window as usize {
+            self.recent_blocks.pop_front();
+        }
+
+        if !self.recent_blocks.is_empty() {
+            let mut rng = rand::thread_rng();
+
+            let historical_blocks = self
+                .recent_blocks
```
eerkaijun: i think it's easier not to keep a VecDeque, we can just randomly sample based on a historical block range specified, for e.g. 20k block from `block_number` up until 10k block from `block_number`. with this we can look back on older history without keeping a large queue (i.e. 200 blocks is still very close to chain tip)

eclark0426: The reason I'd done it this way was mostly for things like running in Flexnet where you aren't necessarily starting with 200 blocks of history. But I guess you can always just make sure your block number is >0

monad-eth-testutil/examples/rpc-request-generator/main.rs
eclark0426: True; removed the for loop and just passed the number of requests directly

PR #1653, opened 2025-03-14
Debian packaging

0cjs:
See commits for details.

### DIFF
--- a/package/build-package
+++ b/package/build-package
@@ -8,6 +8,8 @@ builddir="$PROJDIR"/.build/package
 mkdir -p "$builddir"
 command cd "$builddir"
 
+####################################################################
+
 #   For testing purposes, we are able to build a package for any commit (it
 #   need not even be on `main`). To avoid people mistakenly installing
 #   random code versions, we use a different package name for packages that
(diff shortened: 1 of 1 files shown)


package/Dockerfile
```diff
@@ -0,0 +1,204 @@
...
+ARG TRIEDB_TARGET=triedb_driver
+
+#   Rust and Cxx build.
+#
+#   We try to copy over just the files that we need to build (especially
+#   avoiding test and packaging files in this layer) so that when we're
+#   changing things not in the build (particularly tests), we can re-use
+#   the cached layer with the last build.
+#
+#   We do not yet have the --exclude option (that's in 1.7-labs), and COPY
```
mhumeSF, line 76: Would `.dockerignore` work for this case?

0cjs: No, it can't work because we need the files, just we need _not_ to copy them until later.
(Worked out in more detail in a video chat.)

alexpotato:
Mentioning this more as a "keep in mind", one thing we DON'T want to do is mix up binaries with symbols with "regular" binaries given that releasing with the symbols == going open source.

Not sure if there should be some kind of check here or somewhere to make sure symbols builds are correctly targeted. 

That's why I'm leaving this is a comment.

0cjs:
> Not sure if there should be some kind of check here or somewhere to make sure symbols builds are correctly targeted.

Yes, I have to work this out with the devs. Generally we want our test packages (e.g., `monad-01ab23cd`) to have the symbols because either developers are using them or we're giving them out to people to whom we want to give a specific "non-release" build to debug. But our production packages (`monad`) invariably want the debug symbols stripped. We'll have to see if we need to build a feature that lets us choose between the two for test builds and, if so, what the API and the default should be.

I reckon that production builds will _always_ be stripped by the package build process, but I need to confirm that with the devs, too.

But thanks for reminding me; I've added comments to `package/build-package` about this.

PR #1654, opened 2025-03-14
Txgen: for large numbers of recipients, use non-cached keypool to prevent OOM

JoeHowarth:


monad-eth-testutil/examples/txgen/cli.rs
```diff
@@ -125,6 +125,12 @@ pub struct Config {
 
     #[arg(long, global = true)]
     pub use_static_tps_interval: bool,
+
+    #[arg(long, global = true, default_value_t = 1000)]
+    pub empty_block_duration_ms: u64,
+
+    #[arg(long, global = true)]
+    pub empty_block_frequency_ms: Option<u64>,
```
andr-dev: Why do we need these flags? Txgen is already occasionally producing empty blocks on stressnet.

JoeHowarth: Ariq asked for more control over empty block timing and how many in a row.

andr-dev: Yep that's fine, but if we want more control over empty blocks we should probably figure out why there's currently empty blocks without these flags / changes.

JoeHowarth: Removed the empty block logic to unblock the main change in this pr

PR #1655, opened 2025-03-14
push server messages out every second

maxkozlovsky:
Push server messages out every second even if there is no full batch available. This avoids unnecessary client timeouts if execution is overloaded and cannot produce data fast enough.

### DIFF
--- a/monad-statesync/src/ipc.rs
+++ b/monad-statesync/src/ipc.rs
@@ -131,12 +131,18 @@ struct WipResponse<PT: PubKey> {
     from: NodeId<PT>,
     rx_time: Instant,
     service_start_time: Instant,
+    /// send next message after this time even if we don't have a full batch
+    next_message_schedule: Instant,
     response: StateSyncResponse,
 
     // list of pending completions, max size of MAX_PENDING_RESPONSES
     // set to none if any of the completions have failed
     pending_response_completions: Option<VecDeque<oneshot::Receiver<()>>>,
 }
+
+// Schedule messages to be sent every second even if we don't have a full batch
+// This prevents the client from timing out as long as the server stays up
+const MESSAGE_SCHEDULE_DURATION: Duration = Duration::from_millis(500);
 
 impl<'a, PT: PubKey> StreamState<'a, PT> {
     fn new(
@@ -168,6 +174,7 @@ impl<'a, PT: PubKey> StreamState<'a, PT> {
         enum Event<PT: PubKey> {
             Execution(Result<u8, tokio::io::Error>),
             Request((NodeId<PT>, StateSyncRequest)),
+            Timeout,
         }
         let fut1 = async {
             let event = self.stream.read_u8().await;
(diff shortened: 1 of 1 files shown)


buytenh:
Can we change this to push out server messages twice a second?

The reason I'm asking is because I am preparing a change to dataplane that allows reusing an outbound TCP connection even if no new messages are available for transmission immediately after transmitting the previous message (which is the current behavior), and my timeout for unsuccessfully waiting for a new message to be queued up into dataplane before I close the outbound connection is 1 second, and my timeout should be longer than your timeout, and I can't stretch my 1 second timeout very much, or we run the risk that the peer will trigger a header reception timeout (which is set to 5 seconds) and close the connection just as we were trying to write a message into it.  (See also https://github.com/category-labs/monad-bft/commit/2d5d5284c9e2a96462b4314b2435bc1c8f229b94 for some background of a similar case.)

Relevant issue: https://github.com/category-labs/category-internal/issues/1411

maxkozlovsky:
https://github.com/category-labs/category-internal/issues/1413

maxkozlovsky:
> Can we change this to push out server messages twice a second?
> 
> The reason I'm asking is because I am preparing a change to dataplane that allows reusing an outbound TCP connection even if no new messages are available for transmission immediately after transmitting the previous message (which is the current behavior), and my timeout for unsuccessfully waiting for a new message to be queued up into dataplane before I close the outbound connection is 1 second, and my timeout should be longer than your timeout, and I can't stretch my 1 second timeout very much, or we run the risk that the peer will trigger a header reception timeout (which is set to 5 seconds) and close the connection just as we were trying to write a message into it. (See also [2d5d528](https://github.com/category-labs/monad-bft/commit/2d5d5284c9e2a96462b4314b2435bc1c8f229b94) for some background of a similar case.)
> 
> Relevant issue: [category-labs/category-internal#1411](https://github.com/category-labs/category-internal/issues/1411)

Ok sure

monad-statesync/src/ipc.rs
```diff
@@ -185,25 +200,93 @@ impl<'a, PT: PubKey> StreamState<'a, PT> {
                 self.handle_execution_message(execution_message).await
             }
             Event::Request((from, request)) => self.handle_request(from, request).await,
+            Event::Timeout => self.handle_timeout().await,
```
omegablitz, line 204: Maybe add a WARN log here? We would expect this to not happen often if the traverses are well-distributed and the upsert batch sizes are small enough.

monad-statesync/src/ipc.rs
```diff
@@ -185,25 +200,93 @@ impl<'a, PT: PubKey> StreamState<'a, PT> {
...
+            .is_err()
+        {
+            tracing::warn!("dropping outbound statesync response, consensus state backlogged?");
+        }
+        completion_receiver
+    }
+
+    async fn send_batch(&mut self) {
+        match &mut self.wip_response {
+            None => (),
```
omegablitz: If I understand correctly, this should never be None?

If so, can do something like the following to reduce nesting and make it clear that this is unexpected:

```rust
let Some(wip_response) = &mut self.wip_response else {
    tracing::error!(...);
    return;
};
// ...
```

maxkozlovsky: ok

monad-statesync/src/ipc.rs
maxkozlovsky: added log in handle_timeout()

PR #1658, opened 2025-03-15
fix lint warnings

maxkozlovsky:


monad-chain-config/src/execution_revision.rs
```diff
@@ -1,20 +1,20 @@
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy)]
 pub enum MonadExecutionRevision {
-    V_ZERO,
```
andr-dev, line 3: For these, we should use an allow to silence the linter rather than switch to `VZero` which is weird.

monad-chain-config/src/revision.rs
```diff
@@ -8,15 +8,15 @@ pub trait ChainRevision: Copy + Clone {
 
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy)]
 pub enum MonadChainRevision {
-    V_0_7_0,
```
andr-dev, line 11: Same here, `V_0_7_0` is preferable.

maxkozlovsky:
fixed to ignore warning

PR #1660, opened 2025-03-15
release/release-to-docker-io: Release internal images to docker.io

0cjs:
This should easily allow anybody with access to CL's internal Docker registry and our public docker.io repository (`categoryxyz/`) to release one or more images built from any commit here.
    
There is a reasonable amount of error checking, and, in comments in the script, considerable detail about how our release procedure works. (These comments should generally be preferred to other external documentation, which may be out of date.)

This implements a substantial part of [cint#1417](https://github.com/category-labs/category-internal/issues/1417), and the other parts about building the release are probably not relevant, since I didn't realise when filing that that a Docker workflow is already doing this.

If this is accepted, we need to update the [Script Way](https://www.notion.so/DevOps-Chain-Network-Guides-17675b0ba84080349dcfc72c2e9f29ec?pvs=4#18a75b0ba84080448feafd4f16bbffc7) section of [Pushing Images](https://www.notion.so/DevOps-Chain-Network-Guides-17675b0ba84080349dcfc72c2e9f29ec?pvs=4#18a75b0ba84080c3b027cb3b3b9ecbf8) in our notion.so chain network guide, and should probably just delete the "Manual Method" and "Old way" sections of that. (Presumably we can still retrieve them from the page history if we need to go back and look at them.) We also probably want to update the information about docker.io logins and access tokens throughout that section, but also see [cint#1416](https://github.com/category-labs/category-internal/issues/1416).

### DIFF
--- a/release/release-to-docker-io
+++ b/release/release-to-docker-io
@@ -1,0 +1,137 @@
+#!/usr/bin/env bash
+#
+#   release-to-docker-io: copy Docker images from the CL internal Docker
+#   registry to the docker.io categoryxyz/ public repository
+#
+#   See usage() for command line options. The one you're most likely to use
+#   is `-i IMAGE-NAME` (which can be given multiple times) to release just
+#   one image, rather than the entire default list of $image_names below.
+#
+#   .github/workflows/publish.yml builds and uploads releasable Docker
+#   images to the categoryxyz/ repository in the private
+#   peach10.devcore4.com registry ($internal_repo). These are always tagged
+#   with an abbreviated commit ID (currently 7 chars).
+#
+#   We release to docker.io images tagged *only* with a Git tag known to
+#   exist in our repo. I.e., the the `…:12ab34c` image in $internal_repo is
+#   `…:GIT-TAG` in $public_repo. (The commit to which these Git tags point
+#   _can_ change, but by convention never does.)
+#
+#   This should work on any system where the user is logged in to both
+#   registries; check `~/.docker/config.json` to see if it has entries in
+#   the `auths:` dictionary for these.
+
+set -Eeuo pipefail
+trap 'ec=$?; echo 1>&2 "INTERNAL ERROR: ec=$ec line=$LINENO cmd=$BASH_COMMAND";
+    exit $ec;' ERR
+
+die()  { ec=$1; shift; echo 1>&2 "$(basename "$0"):" "$@"; exit $ec; }
+fail() { echo -e 1>&2 "FAILED: $@"; exit 1; }
+
+setup_docker() {
+    declare -g docker=docker
+    if ! $docker --version >/dev/null; then
+        die 1 "Cannot run '$docker' command. Check path?"
+    elif ! $docker info >/dev/null 2>&1; then
+        docker="sudo docker --config $HOME/.docker/"
+        #   We do not use `sudo -v` here: in common configurations that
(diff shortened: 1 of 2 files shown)


release/release-to-docker-io
```diff
@@ -0,0 +1,137 @@
...
+
+image_list=()
+while [[ $# -gt 0 ]]; do case "$1" in
+    -h)     usage; exit 0;;
+    -i)     shift; image_list+=("$1"); shift;;
+    -*)     die 2 "Unknown option '$1'; use -h for help";;
+    *)      break;
+esac; done
+#   This is the default set of images we release to docker.io; see the
+#   matrix in .github/workflows/public.yml for the set of images that were
```
alexpotato: I think this should be `publish.yml`

release/release-to-docker-io
```diff
@@ -0,0 +1,137 @@
...
+Usage: $(basename "$0") [-h] [-i image]  GIT-TAG"
+  -h        This help.
+  -i NAME   Name of an image to release. May be specified multiple times.
+_____
+}
+
+image_list=()
+while [[ $# -gt 0 ]]; do case "$1" in
+    -h)     usage; exit 0;;
+    -i)     shift; image_list+=("$1"); shift;;
```
alexpotato, line 61: If I understand correctly, if no image list is provided then it uses the full image list from earlier on?

e.g. if the default is "update all of the images", that's good.

alexpotato:
General ops workflow comment:

I think having it here makes sense.

Would also be nice to have this deployed somewhere so you don't have to pull down the full monad-bft repo to push an image.

release/release-to-docker-io
0cjs: Yup, thanks.

release/release-to-docker-io
0cjs: To be more precise, the default is not "update all of the images," it's "update the images in the default list in this file." There are more images built in `publish.yaml` that are not here. I took this list from the script on the notion page; we should definitely tweak it if things have changed since that was last updated.

(It's possible for me to actually grab the default list from `publish.yaml` in the commit to be released, but my current estimate is that it's not worth the effort given that we're probably moving over the next few weeks to a single Docker image containing all programs for those cases where we're not doing bare host installs.)

alexpotato: Agreed that the hardcoded list you have below is fine for now. 

And good pointing out "all" is a big word encompassing more than we need.

```
    monad-bft monad-rpc monad-archive monad-full-node monad-execution)
```

0cjs:
> Would also be nice to have this deployed somewhere so you don't have to pull down the full monad-bft repo to push an image.

Unfortunately, the script needs access to the repo in order to be able to confirm that the requested tag exists and look up the commit ID to which the tag points. I might be able to work around that with calls to GitHub, but that gets a _lot_ more complex when you consider the authentication etc. that's necessary. (Though that bit of complexity will go away once the repo is made public.)

Towards the more general idea, I'm feeling it might end up being more difficult for users in the end to have to know another place to go get the script; "It's in repo X" is kind of something we do for all sorts of things already, anyway. I did consider putting it in `category-infra`, but in the end the devs are (correctly, IMHO) the ones with the final decision on what should be released, and this particular operation seems to me to be one where "let's get devops out of the way" can apply.

PR #1662, opened 2025-03-17
Fix duplicate blocksync requests

bamrith:


monad-blocksync/src/blocksync.rs
```diff
@@ -674,39 +662,36 @@ where
...
-        }
-
         let self_requester = self_request.requester;
         match payload_response {
             BlockSyncBodyResponse::Found(payload) => {
+                if self_request.to != sender {
+                    // unexpected sender, but use the payload since it's valid
+                    self.metrics.blocksync_events.payload_response_unexpected += 1;
+                }
                 assert_eq!(self_requester, self.block_sync.self_request_mode);
```
michael-yxchen: I think we want to increment counter for unexpected response regardless of found or not found. Is there a specific reason to increment only on Found?

monad-blocksync/src/blocksync.rs
```diff
@@ -723,28 +708,31 @@ where
             }
             BlockSyncBodyResponse::NotAvailable(payload_id) => {
                 if sender.is_some() {
+                    // received not availabe from a peer. ignore the response and re-request
```
michael-yxchen: nit: available typo

monad-blocksync/src/blocksync.rs
```diff
@@ -1404,30 +1407,72 @@ mod test {
...
+        let cmds = context
+            .handle_ledger_response(BlockSyncResponseMessage::headers_not_available(block_range));
+        assert_eq!(cmds.len(), 0);
+    }
+
+    #[test]
+    fn timeout_headers_request_to_peer() {
+        // Handle self request should emit a fetch headers command
+        // If not available in ledger, request from peer with a timeout
+        // Re-request when timeout hits
```
michael-yxchen, line 1437: Are there tests asserting timeout expiry triggering another request and schedule timeout?

monad-blocksync/src/blocksync.rs
```diff
@@ -1560,8 +1653,160 @@ mod test {
         }
     }
 
+    #[test]
+    fn avoid_payload_requests_if_in_flight() {
```
michael-yxchen, line 1659: just confirming my understanding, blocksync will emit response to request 1 after receiving the remaining 3 block payloads right?

monad-blocksync/src/blocksync.rs
bamrith: This is the test that asserts that behavior (for headers atleast, the test `timeout_payload_request` asserts the same for payload timeouts). Both tests assert this logic:
- Try to get from ledger
- Not available
- Request from peer + Schedule timeout
- Timeout triggered
- Request from another peer + Schedule timeout

monad-blocksync/src/blocksync.rs
bamrith: yes

monad-blocksync/src/blocksync.rs
bamrith: No, should increment for both. Changed.

monad-blocksync/src/blocksync.rs
```diff
@@ -583,17 +585,34 @@ where
...
+            }
+            BlockSyncHeadersResponse::NotAvailable(block_range) => {
+                debug!(
+                    ?sender,
+                    ?block_range,
+                    "blocksync: headers response not available"
+                );
+                if sender.is_some() {
+                    // received not available from a peer. ignore the response and re-request
+                    // after timeout
```
omegablitz, line 609: Is there a downside to resetting the timer here and re-sending the request to a different peer?
If it's too complicated we don't need to bother.

monad-blocksync/src/blocksync.rs
```diff
@@ -723,28 +710,31 @@ where
             }
             BlockSyncBodyResponse::NotAvailable(payload_id) => {
                 if sender.is_some() {
+                    // received not available from a peer. ignore the response and re-request
+                    // after timeout
```
omegablitz, line 714: Same here - is there a downside to resetting the timer here and re-sending the request to a different peer?

If I understand right, it seems like receiving NotAvailable from a peer doesn't actually cause any action besides bumping the metric.

monad-blocksync/src/blocksync.rs
omegablitz: I guess the only nuance is that you can only re-send the request to a different peer on NotAvailable if the NotAvailable source matches the current pending timeout.

I think that shouldn't be too complicated, but I could be wrong.

bamrith: That was the previous behavior. The reason for this change was to avoid sending extra requests on receiving not available after the timeout is triggered. i.e.
- request is made to a peer
- timeout is triggered, so another request is made
- received not available for the first request (shouldn't make an extra request here)

Sending the request to a different peer each time will solve this, but we also have to store all the peers this request was made to previously, to avoid requesting from the same peer twice. (not sure about what to do if the request is sent to all peers)

Another solution was to add a request ID to each request and handle the `NotAvailable` case only for the latest request ID, but that might be too much effort to change message types.

omegablitz: Got it, that is more complicated than I thought. We can revisit this later.

PR #1666, opened 2025-03-18
respond to statesync requests with incompatible version

maxkozlovsky:
When statesync request with incompatible version is received, respond with minimum and maximum supported version instead of just dropping the message. This allows clients with higher version to be able to downgrade and complete statesync.

### DIFF
--- a/monad-executor-glue/src/lib.rs
+++ b/monad-executor-glue/src/lib.rs
@@ -583,6 +583,7 @@ impl<SCT: SignatureCollection, EPT: ExecutionProtocol> Debug for MempoolEvent<SC
 const STATESYNC_VERSION_V0: StateSyncVersion = StateSyncVersion { major: 1, minor: 0 };
 const STATESYNC_VERSION_V1: StateSyncVersion = StateSyncVersion { major: 1, minor: 1 };
 pub const SELF_STATESYNC_VERSION: StateSyncVersion = STATESYNC_VERSION_V1;
+pub const STATESYNC_VERSION_MIN: StateSyncVersion = STATESYNC_VERSION_V0;
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, RlpEncodable, RlpDecodable)]
 pub struct StateSyncVersion {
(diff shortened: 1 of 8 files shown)


maxkozlovsky:
https://github.com/category-labs/category-internal/issues/1421

monad-executor-glue/src/lib.rs
```diff
@@ -603,15 +604,15 @@ impl StateSyncVersion {
     }
 
     pub fn is_compatible(&self) -> bool {
-        self.major == SELF_STATESYNC_VERSION.major
+        *self >= STATESYNC_VERSION_MIN && *self <= SELF_STATESYNC_VERSION
     }
 
     pub fn is_compatible_with(&self, with: &Self) -> bool {
```
mikety: nit: `is_compatible_with` should use the same compatibility comparison as  `is_compatible` , I think its a transitive relation, i.e. if `a.is_compatible` &&`!b.is_compatible` then `!a.is_compatible_with(b)`

monad-statesync/src/outbound_requests.rs
```diff
@@ -206,6 +206,11 @@ impl<PT: PubKey> OutboundRequests<PT> {
         None
     }
 
+    pub fn handle_bad_version(&mut self, from: NodeId<PT>, bad_version: StateSyncBadVersion) {
+        // Ignore bad version, this version is always supported
```
mikety, line 210: nit: please add a TODO: about how it will be handling the version incompatibility:
i.e.
if V2 > V1 :
sender V2 : receiver V1 -> bad version sender downgrade protocol and continues
sender V1 : receiver V2 -> receiver downgrade protocol and continues

monad-executor-glue/src/lib.rs
maxkozlovsky: removed is_compatible_with, is not called anywhere

monad-statesync/src/outbound_requests.rs
maxkozlovsky: Added comment

monad-executor-glue/src/lib.rs
```diff
@@ -623,6 +620,44 @@ pub struct StateSyncRequest {
...
+                target,
+                from,
+                until,
+                old_target,
+            })
+        } else {
+            // If the version is not compatible, skip the rest of payload we may not understand
+            Ok(Self {
+                version,
+                prefix: 0,
```
omegablitz, line 650: Is the idea that all the members besides `version` be wrapped in a separate format/version enum at some point?

maxkozlovsky: Yes

PR #1667, opened 2025-03-18
Add more logging around RPC request/response size and timing

eclark0426:
In response to https://category-labs.slack.com/archives/C08ANB698F4/p1742187352497099

Unfortunately, I tried for a while to get information on how long it takes to receive the data from the network and transmit the response, as requested in that message, but was not able to find any way to do this, since this is a middleware that's getting called after Actix has received the data, and before Actix transmits the data. But I did add information on request and response sizes, and the time it takes to process the request, at least.

### DIFF
--- a/monad-rpc/src/main.rs
+++ b/monad-rpc/src/main.rs
@@ -52,6 +52,7 @@ use crate::{
         monad_eth_maxPriorityFeePerGas,
     },
     jsonrpc::{JsonRpcError, JsonRpcResultExt, Request, RequestWrapper, Response, ResponseWrapper},
+    timing::TimingMiddleware,
     trace::{
         monad_trace_block, monad_trace_call, monad_trace_callMany, monad_trace_get,
         monad_trace_transaction,
@@ -78,6 +79,7 @@ mod gas_oracle;
 mod hex;
 mod jsonrpc;
 mod metrics;
+mod timing;
 mod trace;
 mod trace_handlers;
 mod txpool;
@@ -889,6 +891,7 @@ async fn main() -> std::io::Result<()> {
             App::new()
                 .wrap(metrics.clone())
                 .wrap(TracingLogger::<MonadJsonRootSpanBuilder>::new())
+                .wrap(TimingMiddleware)
                 .app_data(web::PayloadConfig::default().limit(args.max_request_size))
                 .app_data(web::Data::new(resources.clone()))
                 .service(web::resource("/").route(web::post().to(rpc_handler)))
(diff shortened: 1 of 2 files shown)


monad-rpc/src/main.rs
```diff
@@ -167,6 +169,11 @@ pub async fn rpc_handler(
         _ => debug!(?body, ?response, "rpc_batch_request/response"),
     }
 
+    info!(
+        "{body:?}, response length: {}",
```
ariqchowdhury: response_length_bytes so that it matches the request log

monad-rpc/src/timing.rs
```diff
@@ -0,0 +1,75 @@
...
+            .and_then(|s| s.parse::<usize>().ok())
+            .unwrap_or(0);
+
+        let fut = self.service.call(req);
+
+        Box::pin(async move {
+            let resp = fut.await?;
+            let processing_end = Instant::now();
+
+            info!(
```
ariqchowdhury: I think this should be debug

monad-rpc/src/main.rs
```diff
@@ -167,6 +169,11 @@ pub async fn rpc_handler(
         _ => debug!(?body, ?response, "rpc_batch_request/response"),
     }
 
+    info!(
```
ariqchowdhury: debug

eclark0426:
I've added the client IP as requested, and consolidated everything to a single log; in an effort to simplify things while also avoiding copying the entire response, I decided to forgo printing out the response data. Since @ariqchowdhury was wanting something with a lower-level printing of the time to send data out the interface, we can probably more easily log at that level while still not needing to copy the data.

monad-rpc/src/timing.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    forward_ready!(service);
+
+    fn call(&self, mut req: ServiceRequest) -> Self::Future {
+        let start_time = Instant::now();
+
+        let request_size = req
+            .request()
+            .headers()
+            .get("content-length")
```
philipglazman, line 55: Client can lie about content-length

eclark0426: This feels to me like more of an edge case/malicious behavior than we need to care about here. Everything I was finding about middleware accessing request data seemed to require making a copy of the data, which seems excessive when all we want is the data size, so I grabbed it from the header instead. Since we're mostly just trying to get an idea of the timing of requests, and not directly using it outside of our own debugging, I think that fabricated content-lengths would be an abnormal case and mostly just add to noise in data, rather than having a significant impact

monad-rpc/src/timing.rs
```diff
@@ -0,0 +1,85 @@
...
+            .realip_remote_addr()
+            .unwrap_or("unknown")
+            .to_string();
+        let fut = self.service.call(req);
+
+        Box::pin(async move {
+            let resp = fut.await?;
+            let processing_end = Instant::now();
+
+            debug!(
```
mikety: It should also print when there is no response, or there is an error processing.

mikety:
Having a trace-id added to the request and being printed all the way thru would be very big help for a script or anyone looking at the "life of the request"
But it seems as a pretty large task that can be done separately

PR #1669, opened 2025-03-18
Speedup rpc txpool bridge cleanup process

andr-dev:
Previously, every cleanup interval, the txpool task would iterate through the entire tx status map and remove old entries. With high sustained TPS, the size of this map would grow significantly leading to slow iteration times, measured on stressnet between 20-80ms at 4k TPS. Instead, we can push each hash and the current timestamp onto a queue and periodically check the front of the queue, resulting in 1-5us cleanups at 4k TPS as measured on stressnet.

This cleanup process is currently run on the same tokio task that submits txs to monad-bft and thus contributes to "txpool not responding" errors due to the introduced delay. This change, when tested on stressnet, allowed for higher TPS while producing the same number of "txpool not responding" errors as without this change.

### DIFF
--- a/monad-rpc/src/txpool/state.rs
+++ b/monad-rpc/src/txpool/state.rs
@@ -1,12 +1,12 @@
 use std::{
-    collections::{HashMap, HashSet},
+    collections::{HashMap, HashSet, VecDeque},
     sync::Arc,
     time::Duration,
 };
 
 use alloy_consensus::TxEnvelope;
 use alloy_primitives::{Address, TxHash};
-use dashmap::DashMap;
+use dashmap::{DashMap, Entry};
 use monad_eth_txpool_types::{EthTxPoolEvent, EthTxPoolSnapshot};
 use tokio::time::Instant;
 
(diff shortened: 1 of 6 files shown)


monad-rpc/src/txpool/state.rs
```diff
@@ -6,16 +6,19 @@ use std::{
 
 use alloy_consensus::TxEnvelope;
 use alloy_primitives::{Address, TxHash};
-use dashmap::DashMap;
+use dashmap::{DashMap, Entry};
 use monad_eth_txpool_types::{
     EthTxPoolDropReason, EthTxPoolEvent, EthTxPoolEvictReason, EthTxPoolSnapshot,
 };
+use scc::Queue;
```
ariqchowdhury: isn't everything synchronous right now? do we need this?

andr-dev: The `EthTxPoolBridgeState` struct is currently shared as part of the web server state, but technically only one thread is accessing that queue / writing to the dashmaps. I'm now convinced the right approach here is to separate the bridge state into two, one being the "write" side which is the tokio task that polls the bridge and manages txs and another being the "read" side which gets added to web server state. This means we won't need scc and the queue will just be a `VecDeque`. I'll start working on these changes, feel free to drop any objections/further questions down below. Thanks for helping me think more critically about this :)

andr-dev: Changes are complete and ready in https://github.com/category-labs/monad-bft/pull/1698!

monad-rpc/src/txpool/state.rs
```diff
@@ -93,61 +87,86 @@ impl EthTxPoolBridgeState {
...
+        &self,
+        eviction_queue: &mut EthTxPoolBridgeEvictionQueue,
+        tx: &TxEnvelope,
+    ) {
         let hash = tx.tx_hash();
-        self.status
-            .entry(*hash)
-            .insert(TxStatusEntry::new(TxStatus::Unknown, Instant::now()));
+        self.status.entry(*hash).insert(TxStatus::Unknown);
+        eviction_queue.push_back((Instant::now(), *hash));
```
dshulyak, line 97: what is the desired behavior if tx already exists in the mempool?
i would expect mempool to evict it based on the last seen timestamp

andr-dev:
> just curious, what is the root cause of "txpool not responding" errors? just mempool not fast enough to process some txs in 200ms (20ms x 10) ? seems like doing cleanup in a separate task will make processing latency more predictable

There's a multitude of things going on. The primary issue is that, since the txpool is running sequentially on the main BFT thread, whenever BFT receives a proposal and attempts to verify it, it makes a bunch of triedb lookups which from my observations can take up to 150ms and scale with TPS/unique senders per block. This means the txpool can't make progress in processing transactions and can lead to txpool not responding errors. In fact, anything else that blocks the main thread will cause this, but most other things aren't slow enough to block the txpool for a significant amount of time like handling a proposal. I am actively working on moving the txpool to a separate thread to alleviate this issue.

Although this is the primary cause of the not responding errors, the bridge cleanup process is a secondary cause of lesser extent, as the previous version took a significant amount of time and would compound ontop of the inline BFT txpool issue outlined above. So yes, removing this slow process will make the latency significantly more predictable.

monad-rpc/src/txpool/state.rs
andr-dev: This is actually a great point we should debate here. With the current implementation, we'll set the status from whatever it is now back to `Unknown` and wait for the txpool to give us an updated status response.

Now that you mentioned this and made me think more, I think we shouldn't do that because if the tx is committed the txpool will return an error for the second tx because the nonce will be too low which is undesirable.

dshulyak: i was referring to eviction policy. i think code in this change will evict based on the first seen timestamp. e.g if tx was added several times to the pool there will be earlier timestamp on the front of the queue and then one somewhere later. and if earlier timestamp crosses TX_EVICT_DURATION_SECONDS we will delete data from map regardless of the latter timestamp

it seems first seen policy will make it harder for rpc client to prevent eviction. with last seen policy rpc client would be able to retry submission approx every TX_EVICT_DURATION_SECONDS and it will guarantee that tx is kept

andr-dev: Agreed, for context I went with the first seen policy simply because it's a cleaner solution and doesn't add too much complexity. That being said, the case where first seen policy is possibly undesirable could be considered "bad" client behavior, so I'm not sure whether it's really worth it to change it. If we do want to, it's definitely possible and would likely have a minimal performance hit.

dshulyak: i don't have strong opinion. i personally would use last seen for the reasons described. i think before this change it was using last seen as well, but hard to tell if it was deliberate choice

PR #1675, opened 2025-03-21
Implement periodic pings to calculate network latency

mikety:
Based on Max PR: https://github.com/category-labs/monad-bft/pull/1231.
It adds network latency measurement with added ping/pong message types between pairs of nodes. 

This is the 1st PR out of the 3 PRs implementing clock adjustment:
the other 2 are:
"Use OS time directly for block timestamp" : https://github.com/category-labs/monad-bft/pull/1777
"Use proposal latency to compute timestamp adjustment" : https://github.com/category-labs/monad-bft/pull/1685

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -160,7 +160,7 @@ where
     /// Policy for validating incoming proposals
     pub block_validator: &'a BVT,
     /// Track local timestamp and validate proposal timestamps
-    pub block_timestamp: &'a BlockTimestamp,
+    pub block_timestamp: &'a mut BlockTimestamp<SCT::NodeIdPubKey>,
 
     /// Destination address for proposal payments
     pub beneficiary: &'a [u8; 20],
@@ -565,14 +565,29 @@ where
         // where timestamp validation is done, in block_policy right now.
         // block_policy doesn't have visibility on whether current round is
         // bumped. Deferring the move
+
         /*
-        if let Some(ts_delta) = self
-            .block_timestamp
-            .valid_block_timestamp(block.get_qc().get_timestamp(), block.get_timestamp())
-        {
-            // only update timestamp if the block advanced us our round
-            if block_round > original_round {
-                cmds.push(ConsensusCommand::TimestampUpdate(ts_delta));
+        if let Some(parent_timestamp) = self
+            .consensus
+            .blocktree()
+            .get_timestamp_of_qc(block.get_qc()) {
+
+            if let Ok(Some(ts_delta)) = self.block_timestamp.valid_block_timestamp(
+                parent_timestamp,
+                block.get_timestamp(),
+                self.config
+                        .chain_config
+                        .get_chain_revision(round)
+                        .chain_params()
+                        .vote_pace
+                        .as_nanos(),
+                &author,
+            ) {
+                // only update timestamp if the block advanced us our round
+                if block_round > original_round {
+                    info!(?ts_delta, "update timestamp");
+                    cmds.push(ConsensusCommand::TimestampUpdate(ts_delta));
+                }
             }
         }
         */
@@ -863,6 +878,9 @@ where
             message: ProtocolMessage::Vote(vote_msg),
         }
         .sign(self.keypair);
+        if next_leader != *self.nodeid {
+            self.block_timestamp.vote_sent(round);
+        }
         let send_cmd = ConsensusCommand::Publish {
             target: RouterTarget::PointToPoint(next_leader),
             message: msg,
@@ -1212,8 +1230,9 @@ where
                     .chain_params()
                     .vote_pace
                     .as_nanos(),
+                validated_block.get_author(),
             )
-            .is_none()
+            .is_err()
         {
             self.metrics.consensus_events.failed_ts_validation += 1;
             warn!(
@@ -1619,6 +1638,7 @@ mod test {
         quorum_certificate::QuorumCertificate,
         signature_collection::{SignatureCollection, SignatureCollectionKeyPairType},
         timeout::Timeout,
+        validator_data,
         voting::{ValidatorMapping, Vote},
     };
     use monad_crypto::{
@@ -1642,7 +1662,7 @@ mod test {
     };
     use monad_types::{
         BlockId, Epoch, ExecutionProtocol, MockableFinalizedHeader, MockableProposedHeader, NodeId,
-        Round, RouterTarget, SeqNum, Stake, GENESIS_SEQ_NUM,
+        PingSequence, Round, RouterTarget, SeqNum, Stake, GENESIS_SEQ_NUM,
     };
     use monad_validator::{
         epoch_manager::EpochManager,
@@ -1655,8 +1675,9 @@ mod test {
     use tracing_test::traced_test;
 
     use crate::{
-        timestamp::BlockTimestamp, ConsensusCommand, ConsensusConfig, ConsensusState,
-        ConsensusStateWrapper, OutgoingVoteStatus,
+        timestamp::{BlockTimestamp, PING_PERIOD_SEC},
+        ConsensusCommand, ConsensusConfig, ConsensusState, ConsensusStateWrapper,
+        OutgoingVoteStatus,
     };
 
     const BASE_FEE: u128 = BASE_FEE_PER_GAS as u128;
@@ -1701,7 +1722,7 @@ mod test {
         block_validator: BVT,
         block_policy: BPT,
         state_backend: SBT,
-        block_timestamp: BlockTimestamp,
+        block_timestamp: BlockTimestamp<SCT::NodeIdPubKey>,
         beneficiary: [u8; 20],
         nodeid: NodeId<CertificateSignaturePubKey<ST>>,
         consensus_config: ConsensusConfig<MockChainConfig, MockChainRevision>,
@@ -2451,6 +2472,10 @@ mod test {
             || EthValidator::new(0),
             execution_delay,
         );
+        for ctx in &mut ctx {
+            ctx.wrapped_state().block_timestamp.update_time(1000);
+        }
+
         let mut wrapped_state = ctx[0].wrapped_state();
 
         // our initial starting logic has consensus in round 1 so the first proposal does not
@@ -2461,6 +2486,50 @@ mod test {
 
         let p2 = env.next_proposal_empty();
         let (author, _, verified_message) = p2.destructure();
+
+        let val_set = wrapped_state
+            .val_epoch_map
+            .get_val_set(
+                &wrapped_state
+                    .epoch_manager
+                    .get_epoch(verified_message.block_header.round)
+                    .unwrap(),
+            )
+            .unwrap();
+        let val_data = val_set
+            .get_members()
+            .iter()
+            .map(
+                |(node_id, stake)| validator_data::ValidatorData::<SignatureCollectionType> {
+                    node_id: *node_id,
+                    stake: *stake,
+                    cert_pubkey: node_id.pubkey(),
+                },
+            )
+            .collect::<Vec<_>>();
+
+        let epoch = wrapped_state
+            .epoch_manager
+            .get_epoch(verified_message.block_header.round)
+            .unwrap();
+
+        let vals = validator_data::ValidatorSetDataWithEpoch::<SignatureCollectionType> {
+            validators: validator_data::ValidatorSetData::<SignatureCollectionType>(val_data),
+            epoch,
+        };
+        wrapped_state
+            .block_timestamp
+            .update_validators(&vals, wrapped_state.nodeid);
+
+        for _ in 0..PING_PERIOD_SEC {
+            wrapped_state.block_timestamp.tick();
+        }
+        for val in val_set.get_members().keys() {
+            wrapped_state
+                .block_timestamp
+                .pong_received(*val, PingSequence(1));
+        }
+        wrapped_state.block_timestamp.vote_sent(Round(1));
         let cmds = wrapped_state.handle_proposal_message(author, verified_message.clone());
         assert!(find_timestamp_update_cmd(&cmds).is_some());
 
@@ -2475,7 +2544,10 @@ mod test {
         let p7 = env.next_proposal_empty();
         let (author, _, verified_message) = p7.destructure();
         let cmds = wrapped_state.handle_proposal_message(author, verified_message);
-        assert!(find_timestamp_update_cmd(&cmds).is_some());
+        assert!(
+            find_timestamp_update_cmd(&cmds).is_none(),
+            "no timestamp adjustment because did not vote previous round"
+        );
     }
 
     #[test]
--- a/monad-consensus-state/src/timestamp.rs
+++ b/monad-consensus-state/src/timestamp.rs
@@ -1,24 +1,253 @@
-use monad_consensus_types::quorum_certificate::{
-    TimestampAdjustment, TimestampAdjustmentDirection,
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    hash::{DefaultHasher, Hash, Hasher},
+    time::{Duration, Instant},
 };
 
+use monad_consensus_types::{
+    quorum_certificate::{TimestampAdjustment, TimestampAdjustmentDirection},
+    signature_collection::SignatureCollection,
+    validator_data::ValidatorSetDataWithEpoch,
+};
+use monad_crypto::certificate_signature::PubKey;
+use monad_types::{Epoch, NodeId, PingSequence, Round};
+use tracing::debug;
+
+const MAX_LATENCY_SAMPLES: usize = 100;
+
+pub const PING_PERIOD_SEC: usize = 30;
+
+pub const PING_TICK_DURATION: Duration = Duration::from_secs(1);
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum Error {
+    Invalid,     // timestamp did not increment compared to previous block
+    OutOfBounds, // timestamp is out of bounds compared to local time
+}
+
+#[derive(Debug, Clone)]
+struct RunningAverage {
+    sum: Duration,
+    max_samples: usize,
+    samples: Vec<Duration>,
+    avg: Option<Duration>,
+    next_index: usize,
+}
+
+impl RunningAverage {
+    pub fn new(max_samples: usize) -> Self {
+        assert!(max_samples > 0);
+        Self {
+            sum: Default::default(),
+            max_samples,
+            samples: Vec::new(),
+            avg: None,
+            next_index: 0,
+        }
+    }
+
+    pub fn add_sample(&mut self, sample: Duration) {
+        if self.samples.len() == self.max_samples {
+            self.sum -= self.samples[self.next_index];
+            self.samples[self.next_index] = sample;
+            self.next_index = (self.next_index + 1) % self.max_samples;
+        } else {
+            self.samples.push(sample);
+        }
+        self.sum += sample;
+        self.avg = Some(self.sum / self.samples.len() as u32);
+    }
+
+    pub fn get_avg(&self) -> Option<Duration> {
+        self.avg
+    }
+}
+
+/// Ping state per validator
+#[derive(Debug, Clone)]
+pub struct ValidatorPingState {
+    sequence: PingSequence,       // sequence number of the last ping sent
+    ping_latency: RunningAverage, // average latency of the last pings
+    last_ping_time: Instant,      // time of the last ping sent
+}
+
+impl Default for ValidatorPingState {
+    fn default() -> Self {
+        let now = Instant::now();
+        Self {
+            sequence: PingSequence(0),
+            ping_latency: RunningAverage::new(MAX_LATENCY_SAMPLES),
+            last_ping_time: now,
+        }
+    }
+}
+
+impl ValidatorPingState {
+    pub fn new() -> Self {
+        Self::default()
+    }
+    // initiate a new ping
+    pub fn start_next(&mut self) -> PingSequence {
+        self.sequence.0 = self.sequence.0.wrapping_add(1);
+        self.last_ping_time = Instant::now();
+        self.sequence
+    }
+
+    pub fn pong_received(&mut self, sequence: PingSequence) -> Option<Duration> {
+        if sequence != self.sequence {
+            return None;
+        }
+        // estimate latency as half the round trip time
+        let elapsed = self.last_ping_time.elapsed() / 2;
+        self.update_latency(elapsed);
+        Some(elapsed)
+    }
+
+    pub fn avg_latency(&self) -> Option<Duration> {
+        self.ping_latency.get_avg()
+    }
+
+    fn update_latency(&mut self, latency: Duration) {
+        self.ping_latency.add_sample(latency);
+    }
+}
+
+#[derive(Debug, Clone)]
+struct PingState<P: PubKey> {
+    current_epoch: Epoch,
+    epoch_validators: BTreeMap<Epoch, Vec<NodeId<P>>>,
+    validators: BTreeMap<NodeId<P>, ValidatorPingState>, // a union of the current and the next epoch validator sets
+    schedule: Box<[Vec<NodeId<P>>; PING_PERIOD_SEC]>,    // schedule of validators to ping
+    tick: usize,                                         // current tick index into schedule
+}
+
+impl<P: PubKey> PingState<P> {
+    pub fn new() -> Self {
+        Self {
+            current_epoch: Epoch(0),
+            epoch_validators: BTreeMap::new(),
+            validators: BTreeMap::new(),
+            schedule: Box::new([const { Vec::new() }; PING_PERIOD_SEC]),
+            tick: 0,
+        }
+    }
+
+    pub fn pong_received(&mut self, node_id: NodeId<P>, sequence: PingSequence) {
+        if let Some(validator_state) = self.validators.get_mut(&node_id) {
+            if let Some(elapsed) = validator_state.pong_received(sequence) {
+                debug!(?node_id, elapsed_ms = ?elapsed.as_millis(), avg_latency_ms = ?validator_state.avg_latency().unwrap_or_default().as_millis(), "ping latency");
+            }
+        }
+    }
+
+    fn update_validators<SCT>(
+        &mut self,
+        val_set: &ValidatorSetDataWithEpoch<SCT>,
+        my_node: &NodeId<P>,
+    ) where
+        SCT: SignatureCollection<NodeIdPubKey = P>,
+    {
+        match self.epoch_validators.entry(val_set.epoch) {
+            std::collections::btree_map::Entry::Vacant(entry) => {
+                entry.insert(
+                    val_set
+                        .validators
+                        .0
+                        .iter()
+                        .filter(|v| v.node_id != *my_node)
+                        .map(|v| v.node_id)
+                        .collect(),
+                );
+            }
+            std::collections::btree_map::Entry::Occupied(entry) => {
+                let val_set: Vec<_> = val_set
+                    .validators
+                    .0
+                    .iter()
+                    .filter(|v| v.node_id != *my_node)
+                    .map(|v| v.node_id)
+                    .collect();
+                assert_eq!(
+                    (*entry.get()).iter().copied().collect::<BTreeSet<_>>(),
+                    val_set.iter().copied().collect::<BTreeSet<_>>(),
+                    "Validators update is not matching existing for the same epoch"
+                )
+            }
+        }
+    }
+
+    fn compute_schedule(&mut self) {
+        let active_validators = self
+            .epoch_validators
+            .iter()
+            .filter(|entry| {
+                *entry.0 == self.current_epoch || *entry.0 == self.current_epoch + Epoch(1)
+            })
+            .flat_map(|entry| entry.1)
+            .cloned()
+            .collect::<BTreeSet<NodeId<P>>>();
+
+        self.validators
+            .retain(|node_id, _| active_validators.contains(node_id));
+
+        for s in self.schedule.iter_mut() {
+            s.clear();
+        }
+
+        // TODO: reimplement shuffling with rand.
+        for validator in active_validators.iter() {
+            self.validators.entry(*validator).or_default();
+            let mut hasher = DefaultHasher::new();
+            validator.hash(&mut hasher);
+            let idx = (hasher.finish() as usize) % self.schedule.len();
+            self.schedule[idx].push(*validator);
+        }
+    }
+
+    // returns list of nodes to send pings to on this tick
+    fn tick(&mut self) -> Vec<(NodeId<P>, PingSequence)> {
+        let mut pings = Vec::new();
+        for node_id in self.schedule[self.tick].iter() {
+            let validator_state = self.validators.get_mut(node_id).unwrap();
+            let sequence = validator_state.start_next();
+            pings.push((*node_id, sequence));
+        }
+        self.tick = (self.tick + 1) % self.schedule.len();
+        pings
+    }
+
+    fn get_latency(&self, node_id: &NodeId<P>) -> Option<Duration> {
+        self.validators.get(node_id).and_then(|v| v.avg_latency())
+    }
+}
+
 #[derive(Debug)]
-pub struct BlockTimestamp {
+struct SentVote {
+    round: Round,
+    timestamp: Instant,
+}
+#[derive(Debug)]
+pub struct BlockTimestamp<P: PubKey> {
     local_time_ns: u128,
 
     max_delta_ns: u128,
 
-    /// TODO: this needs an upper-bound
-    latency_estimate_ns: u128,
+    ping_state: PingState<P>,
+
+    last_sent_vote: Option<SentVote>, // last voted round and timestamp
+
+    default_latency_estimate_ns: u128,
 }
 
-impl BlockTimestamp {
-    pub fn new(max_delta_ns: u128, latency_estimate_ns: u128) -> Self {
-        assert!(latency_estimate_ns > 0);
+impl<P: PubKey> BlockTimestamp<P> {
+    pub fn new(max_delta_ns: u128, default_latency_estimate_ns: u128) -> Self {
+        assert!(default_latency_estimate_ns > 0);
         Self {
             local_time_ns: 0,
             max_delta_ns,
-            latency_estimate_ns,
+            default_latency_estimate_ns,
+            ping_state: PingState::new(),
+            last_sent_vote: None,
         }
     }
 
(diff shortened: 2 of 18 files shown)


monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,258 @@
...
+    signature_collection::SignatureCollection,
+    validator_data::ValidatorData,
+};
+use monad_crypto::certificate_signature::PubKey;
+use monad_types::{NodeId, PingSequence, Round};
+use tracing::info;
+
+const MAX_LATENCY_SAMPLES: usize = 100;
+
+const PING_PERIOD: Duration = Duration::from_secs(30);
```
andr-dev: Let's change this constant to be `PING_PERIOD_SECS: u64` and then instantiate Durations when we want a duration rather than calling `to_secs()` on a `Duration` which is technically imprecise.

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,258 @@
...
+        self.sequence
+    }
+
+    pub fn pong_received(&mut self, sequence: PingSequence) {
+        if sequence != self.sequence {
+            return;
+        }
+        // estimate latency as half the round trip time
+        self.update_latency(self.last_ping_time.elapsed() / 2);
+        info!("ping latency {:?}", self.ping_latency.get_avg());
```
andr-dev: We should move this log to `debug`. Also, can we do something like
```rust
info!(
    latency = self.ping_latency.get_avg(),
    "ping latency"
)
```
This way, it'll be a separate field in the log json so if we ever do any analysis on latencies it's easier to read in.

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,258 @@
...
+    }
+
+    pub fn avg_latency(&self) -> Option<Duration> {
+        self.ping_latency.get_avg()
+    }
+
+    fn update_latency(&mut self, latency: Duration) {
```
andr-dev, line 110: This function is only called in `pong_received`, let's inline it there.

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,258 @@
...
+        self.ping_latency.add_sample(latency);
+    }
+}
+
+#[derive(Debug)]
+struct PingState<P: PubKey> {
+    validators: HashMap<NodeId<P>, ValidatorPingState>, // ping state per validator
+    schedule: Vec<Vec<NodeId<P>>>, // schedule of validators to ping, length is period
```
andr-dev: If we change `PING_PERIOD` to `PING_PERIOD_SECS` as mentioned above, we can make this
```rust
schedule: [Vec<NodeId<P>>; PING_PERIOD_SECS]
```
That way, it's clear that the length of schedule never changes.

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,258 @@
...
+    }
+
+    pub fn pong_received(&mut self, node_id: NodeId<P>, sequence: PingSequence) {
+        if let Some(validator_state) = self.validators.get_mut(&node_id) {
+            validator_state.pong_received(sequence);
+            info!(
+                "node {:?} latency {:?}",
+                node_id,
+                validator_state.avg_latency()
+            );
```
andr-dev: Ditto about log level, do we really want this in the info log?

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,258 @@
...
+    fn update_validators<SCT>(&mut self, validators: &Vec<ValidatorData<SCT>>, my_node: &NodeId<P>)
+    where
+        SCT: SignatureCollection<NodeIdPubKey = P>,
+    {
+        let removed_nodes = self
+            .validators
+            .keys()
+            .filter(|node_id| !validators.iter().any(|v| v.node_id == **node_id))
+            .cloned()
+            .collect::<Vec<_>>();
```
andr-dev: We don't need to collect this into a `Vec<_>`, let's just use the iterator.

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,258 @@
...
+    fn get_latency(&self, node_id: &NodeId<P>) -> Option<Duration> {
+        self.validators.get(node_id).and_then(|v| v.avg_latency())
+    }
+
+    fn next_validator_set_at(&mut self, instant: Instant) -> Vec<NodeId<P>> {
+        if instant <= self.schedule_time {
+            return Vec::new();
+        }
+        let index = (self.schedule_time - self.start_time).as_secs() % self.period as u64;
+        let secs_passed = duration_ceil(instant - self.schedule_time).as_secs();
```
andr-dev: Do we want ceil here? It feels counter-intuitive that, according to this formulation, after 1 nano it's been a second.

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,258 @@
...
+
+#[derive(Debug)]
+pub struct BlockTimestamp<P: PubKey> {
     local_time_ns: u128,
 
     max_delta_ns: u128,
 
+    ping_state: PingState<P>,
+
+    sent_vote: Option<SentVote>, // last voted round and timestamp
```
andr-dev: Calling this `last_sent_vote` makes it clear what the field is and removes the need for the comment.

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +286,301 @@ impl BlockTimestamp {
...
+        if !self.valid_bounds(curr_block_ts, vote_delay_ns) {
+            return Err(Error::OutOfBounds);
+        }
+
+        // adjust for estimated latency
+        let latency = self
+            .ping_state
+            .get_latency(author)
+            .unwrap_or(Duration::from_nanos(
+                self.latency_estimate_ns.try_into().unwrap(),
```
andr-dev: Doesn't seem like a great idea to rely on upstream latency_estimate_ns being within some range, let's return an error here rather than crashing the node if latency_estimate_ns exceeds u64::MAX.

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +286,301 @@ impl BlockTimestamp {
         prev_block_ts: u128,
         curr_block_ts: u128,
         vote_delay_ns: u128,
-    ) -> Option<TimestampAdjustment> {
-        let delta = curr_block_ts.checked_sub(prev_block_ts);
-        match delta {
+        author: &NodeId<P>,
+    ) -> Result<Option<TimestampAdjustment>, Error> {
```
andr-dev, line 285: https://github.com/category-labs/monad-bft/pull/1231/files#r1841184521
Has this been addressed?

monad-executor-glue/src/convert/event.rs
```diff
@@ -13,13 +13,16 @@ use monad_crypto::certificate_signature::{
 };
 use monad_proto::{
     error::ProtoError,
-    proto::{blocksync::ProtoBlockSyncSelfRequest, event::*},
+    proto::{
+        blocksync::ProtoBlockSyncSelfRequest,
+        event::{self, *},
```
andr-dev: This looks funny, if we're already importing * why import event with self? I think we should generally avoid star imports, let's just specify each type here (which should be Proto prefixed).

monad-proto/proto/event.proto
```diff
@@ -259,6 +259,20 @@ message ProtoConfigEvent {
   }
 }
 
+message PingRequestEvent {
```
andr-dev: Let's prefix these with Proto.
https://github.com/category-labs/monad-bft/pull/1231/files#r2031601480

monad-state/src/lib.rs
```diff
@@ -855,6 +898,13 @@ where
             }
 
             MonadEvent::ValidatorEvent(validator_event) => {
+                match &validator_event {
+                    ValidatorEvent::UpdateValidators((validators, epoch)) => {
+                        self.block_timestamp
+                            .update_validators(&validators.0, &self.nodeid);
+                    }
+                }
+
```
andr-dev: https://github.com/category-labs/monad-bft/pull/1231/files#r1847282465

monad-state/src/lib.rs
```diff
@@ -1097,6 +1147,39 @@ where
...
+                vec![]
+            }
+            MonadEvent::PingTickEvent => {
+                tracing::info!("ping tick");
+                let mut cmds = self
+                    .block_timestamp
+                    .tick()
+                    .into_iter()
+                    .map(|(node, sequence)| {
+                        tracing::info!(?node, ?sequence, "sending ping request");
```
andr-dev: We should probably move this into `debug` as it'll unnecessarily pollute the info log as node count increases.

Same goes for the tracing::info above, we should be more selective about what makes it to the `info` log level.

monad-consensus-state/src/timestamp.rs
mikety: I will keep the length of the schedule variable - as a follow up  I intend to make ping_period and max_latency_sample configurable.

monad-consensus-state/src/timestamp.rs
mikety: Its also called in tests

monad-consensus-state/src/timestamp.rs
mikety: This will not work as it  mixes mutable borrow of `self.validators` in line 169 and immutable borrow on line 163.

monad-consensus-state/src/timestamp.rs
mikety: Its a config value - default latency, so if its out of bounds it is a bug.

monad-consensus-state/src/timestamp.rs
mikety: Yes, but its in the another PR: https://github.com/category-labs/monad-bft/pull/1685/files#diff-7f422a2cd73b3262c684d2aff2aaa417c777cbaf8db4cfadd34693088862b92bR569.

monad-executor-glue/src/convert/event.rs
mikety: reverted

monad-consensus-state/src/timestamp.rs
andr-dev: In that case, the more appropriate type is `Box<[Vec<NodeId<P>>]>` which explicitly does not allow for resizing of the array after initialization.

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,248 @@
...
+            tick: 0,
+            period,
+        }
+    }
+
+    pub fn pong_received(&mut self, node_id: NodeId<P>, sequence: PingSequence) {
+        if let Some(val) = self.epoch_validators.get_mut(&self.current_epoch) {
+            if let Some(validator_state) = val.get_mut(&node_id) {
+                validator_state.pong_received(sequence);
+                debug!(?node_id, latency = ?validator_state.avg_latency().unwrap_or_default().as_secs(), "ping latency secs");
```
andr-dev: With `debug`, we can do `latency_secs = ?...` so the message remains `"ping latency"`

monad-consensus-state/src/timestamp.rs
andr-dev: Then let's pull it into this PR :)

monad-executor-glue/src/convert/event.rs
```diff
@@ -1098,3 +1122,51 @@ impl<SCT: SignatureCollection> TryFrom<ProtoConfigEvent> for ConfigEvent<SCT> {
...
+}
+
+impl<SCT: SignatureCollection> TryFrom<ProtoPingRequestEvent> for PingEvent<SCT> {
+    type Error = ProtoError;
+    fn try_from(value: ProtoPingRequestEvent) -> Result<Self, Self::Error> {
+        Ok(Self {
+            sender: match value.sender {
+                Some(sender) => sender.try_into()?,
+                None => Err(ProtoError::MissingRequiredField(
+                    "PingResponseEvent.sender".to_owned(),
```
andr-dev: This should be `PingRequestEvent`.

monad-debugger/src/graphql.rs
```diff
@@ -244,6 +244,10 @@ impl<'s> From<&'s MonadEventType> for GraphQLMonadEvent<'s> {
             }
             MonadEvent::StateSyncEvent(event) => Self::StateSyncEvent(GraphQLStateSyncEvent(event)),
             MonadEvent::ConfigEvent(event) => Self::ConfigEvent(GraphQLConfigEvent(event)),
+            &MonadEvent::PingRequestEvent(_)
+            | &MonadEvent::PingResponseEvent(_)
+            | &MonadEvent::PingTickEvent => todo!(),
+            &MonadEvent::TimestampUpdateValidatorsEvent(_) => todo!(),
```
andr-dev: Why the references here?

monad-executor-glue/src/lib.rs
```diff
@@ -1037,6 +1046,12 @@ where
     StateSyncEvent(StateSyncEvent<ST, SCT, EPT>),
     /// Config updates
     ConfigEvent(ConfigEvent<SCT>),
+    /// Ping events
+    PingRequestEvent(PingEvent<SCT>),
+    PingResponseEvent(PingEvent<SCT>),
+    PingTickEvent,
+    /// Events to update the block validator list
+    TimestampUpdateValidatorsEvent(u64),
```
andr-dev: What does this `u64` mean? Let's change this to a struct variant or make a struct for the event.

monad-consensus-state/src/timestamp.rs
andr-dev: We can also remove the `period` field as it always should be equal to the length of `schedule`, and it's generally best to optimize for single sources of truth.

monad-consensus-state/src/timestamp.rs
mikety: I think it will be clear to have the 3 PRs separated. The PR for using proposal latency relies on another quite big PR to use OS time directly for block timestamp (https://github.com/category-labs/monad-bft/pull/1682) . The current  PR focuses on network p2p latency estimation.

monad-consensus-state/src/timestamp.rs
mikety: Updated to the proposal to use PING_PERIOD_SECS as length when building schedule

andr-dev: I think you might have missed my previous comment about using `Box` over `Vec`.

monad-consensus-state/src/timestamp.rs
andr-dev: Got it, in that case can we call it `default_latency_estimate_ns`? That way, it's clear that this is a config value that shouldn't change after initialization.

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,245 @@
...
+struct RunningAverage {
+    sum: Duration,
+    max_samples: usize,
+    samples: Vec<Duration>,
+    avg: Option<Duration>,
+    next_index: usize,
+}
+
+impl RunningAverage {
+    pub fn new(max_samples: usize) -> Self {
```
andr-dev, line 38: This should never happen unless we do something dumb with the configuration but let's add an assert that `max_samples > 0` for safety.

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,246 @@
...
+    }
+}
+
+impl ValidatorPingState {
+    pub fn new() -> Self {
+        Self::default()
+    }
+    // initiate a new ping
+    pub fn start_next(&mut self) -> PingSequence {
+        self.sequence.0 += 1;
```
michael-yxchen: nit: we should never hit an u64 overflow, but maybe wrapping_add for safety

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,246 @@
...
+
+    fn update_latency(&mut self, latency: Duration) {
+        self.ping_latency.add_sample(latency);
+    }
+}
+
+#[derive(Debug, Clone)]
+struct PingState<P: PubKey> {
+    current_epoch: Epoch,
+    epoch_validators: BTreeMap<Epoch, HashMap<NodeId<P>, ValidatorPingState>>,
```
michael-yxchen: I think we try avoiding `HashMap<NodeId<P>, _>` because DefaultHasher is quite slow. Elsewhere we usually use `BTreeMap`. we should add a benchmark

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,246 @@
...
+    schedule: Box<[Vec<NodeId<P>>; PING_PERIOD_SEC]>, // schedule of validators to ping
```
michael-yxchen: what's the benefit of using a `Box<[_]>`?

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,246 @@
...
+            schedule: Box::new([const { Vec::new() }; PING_PERIOD_SEC]),
+            tick: 0,
+        }
+    }
+
+    pub fn pong_received(&mut self, node_id: NodeId<P>, sequence: PingSequence) {
+        if let Some(val) = self.epoch_validators.get_mut(&self.current_epoch) {
+            if let Some(validator_state) = val.get_mut(&node_id) {
+                validator_state.pong_received(sequence);
+                debug!(?node_id, latency_secs = ?validator_state.avg_latency().unwrap_or_default().as_secs(), "ping latency");
```
michael-yxchen: I think it's valuable to log individual latency measurements, on top of average

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,246 @@
...
+            current_epoch: Epoch(0),
+            epoch_validators: BTreeMap::new(),
+            schedule: Box::new([const { Vec::new() }; PING_PERIOD_SEC]),
+            tick: 0,
+        }
+    }
+
+    pub fn pong_received(&mut self, node_id: NodeId<P>, sequence: PingSequence) {
+        if let Some(val) = self.epoch_validators.get_mut(&self.current_epoch) {
+            if let Some(validator_state) = val.get_mut(&node_id) {
```
michael-yxchen: saw that there's a `get_active_validator_state` function. Should we replace with it here?

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,246 @@
...
+    ) where
+        SCT: SignatureCollection<NodeIdPubKey = P>,
+    {
+        if !self.epoch_validators.contains_key(epoch) {
+            self.epoch_validators.insert(*epoch, HashMap::new());
+        }
+        let res = self
+            .epoch_validators
+            .entry(*epoch)
+            .and_modify(|cur_validators| {
```
michael-yxchen:
```suggestion
        let entry = self.epoch_validators.entry(*epoch).or_default();
        entry...
```

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,246 @@
...
+                // map validators to schedule slots by hashing node id
+                for s in self.schedule.iter_mut() {
+                    s.clear();
+                }
+
+                for node in cur_validators.keys() {
+                    let mut hasher = DefaultHasher::new();
+                    node.hash(&mut hasher);
+                    let idx = (hasher.finish() as usize) % self.schedule.len();
+                    self.schedule[idx].push(*node);
```
michael-yxchen: I'm not understanding this part. Validators are locked in only once, i.e. you lock in validator set for epoch e+2 at the end of e, then it remain unchanged. Also, we'd always lock in the set for a future epoch, then enter that epoch much later. See https://www.notion.so/Epoch-switch-spec-v0-15c75b0ba840810fbdbbe3b9aca4c5f0?pvs=4#15c75b0ba8408154b345ee5aeba291a4

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +274,292 @@ impl BlockTimestamp {
...
+
+        // adjust for estimated latency
+        let latency = self
+            .ping_state
+            .get_latency(author)
+            .unwrap_or(Duration::from_nanos(
+                self.default_latency_estimate_ns.try_into().unwrap(),
+            ));
+
+        let expected_block_ts = self.local_time_ns.saturating_sub(latency.as_nanos());
```
michael-yxchen, line 302: raptorcast is two hops. Proposer->self latency measurement won't be a good estimate of block propagation delay

mikety: agreed, this is addressed in the follow up [PR 1685](https://github.com/category-labs/monad-bft/pull/1685). I converted it to draft now to address some feedback.
This PR is only for network latency estimates.

monad-consensus-state/src/timestamp.rs
mikety: To address @andr-dev  feedback. Im fine with either way, I think the Box is less user friendly but offers some perf improvement for large allocations over Vec, but in this case it will not matter much.

monad-consensus-state/src/timestamp.rs
mikety: At every moment timestamp should keep updating latencies for the current epoch validator set, and for the next epoch validator set, so when new set is active on the epoch change it already has computed  latencies.

monad-consensus-state/src/timestamp.rs
mikety: `get_active_validator_state` returns immutable borrow, here it needs a mutable borrow.

monad-consensus-state/src/timestamp.rs
andr-dev: The benefit of `Box<[_]>` is that the type contractually states "I am a fixed-size array and thus cannot grow/shrink". Vec's should only be used when the size of the dynamic array changes, or as the rust docs put it:
```
A contiguous growable array type, written as Vec<T>, short for ‘vector’.
```
`schedule` should not be growable, and thus should not be a Vec. Performance benefit is likely negligible.

monad-consensus-state/src/timestamp.rs
michael-yxchen: I think the disconnect comes from when `MonadEvent::ValidatorEvent` is generated. `StateRootHashTriedbPoll` yields `ValidatorEvent` for epoch `e+2` when boundary block for `e` is executed. The event is received when the validator set is locked, not when we enter that given epoch

An alternative approach is to generate a loopback command on `PacemakerCommand::EnterRound((Epoch, Round))`. In `MonadState.update()`, we can pass in a `&ValidatorsEpochMapping` to `block_timestamp.enter_round()` It can decide whether it needs to update the current validator set, based on epoch

mikety: When new validators are added on a new epoch they already should have latency already computed.  For that the set of validators from the current and the next epochs should be on ping schedule.

So when `MonadEvent::ValidatorEvent` is generated on a boundary block it will call `update_validators` to update the `epoch_validators` . The updated set is not reflected in the ping schedule.

 When new event `MonadEvent::TimestampEnterRoundEvent` is generated via loopback it will result in calling `compute_schedule` when the epoch is changed.

mikety: @michael-yxchen Ive added a unit test `test_new_epoch` to test how validator set updates will work on epoch change.

andr-dev (dismissed):

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,243 @@
...
+        SCT: SignatureCollection<NodeIdPubKey = P>,
+    {
+        let entry = self.epoch_validators.entry(*val_epoch).or_default();
+        entry.clear();
+        for validator in validators {
+            if validator.node_id == *my_node {
+                continue;
+            }
+            entry.push(NodeId::new(validator.node_id.pubkey()));
+        }
```
michael-yxchen: Receiving two validator updates for the same epoch is unexpected. We should assert new validators matches existing entry, and optionally warn about it. Similar to https://github.com/category-labs/monad-bft/blob/687b1a5018f9cdbd8ce782d00773f9e69b7bbfdd/monad-validator/src/validators_epoch_mapping.rs#L60-L94

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,243 @@
...
+    ) where
+        SCT: SignatureCollection<NodeIdPubKey = P>,
+    {
+        let entry = self.epoch_validators.entry(*val_epoch).or_default();
+        entry.clear();
+        for validator in validators {
+            if validator.node_id == *my_node {
+                continue;
+            }
+            entry.push(NodeId::new(validator.node_id.pubkey()));
```
michael-yxchen:
```suggestion
            entry.push(validator.node_id);
```

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +271,446 @@ impl BlockTimestamp {
...
+
+    pub fn vote_sent(&mut self, round: Round) {
+        self.last_sent_vote = Some(SentVote {
+            round,
+            timestamp: Instant::now(),
+        });
+    }
+
+    pub fn enter_round(&mut self, epoch: &Epoch) {
+        debug!(?epoch, "Enter round");
```
michael-yxchen: round -> epoch

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +271,446 @@ impl BlockTimestamp {
...
+        if self.ping_state.current_epoch != *epoch {
```
michael-yxchen: we can add an assertion `assert!(*epoch >= self.ping_state.current_epoch)`, then change the condition to `*epoch > self.ping_state.current_epoch`

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,243 @@
...
+                *entry.0 == self.current_epoch || *entry.0 == self.current_epoch + Epoch(1)
+            })
+            .flat_map(|entry| entry.1)
+            .cloned()
+            .collect::<BTreeSet<NodeId<P>>>();
+
+        let removed_nodes = self
+            .validators
+            .keys()
+            .filter(|node_id| !active_validators.iter().any(|v| *v == **node_id))
```
michael-yxchen:
```suggestion
            .filter(|node_id| !active_validators.contains(*node_id))
```

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,243 @@
...
+    fn update_latency(&mut self, latency: Duration) {
+        self.ping_latency.add_sample(latency);
+    }
+}
+
+#[derive(Debug, Clone)]
+struct PingState<P: PubKey> {
+    current_epoch: Epoch,
+    epoch_validators: BTreeMap<Epoch, Vec<NodeId<P>>>,
+    validators: BTreeMap<NodeId<P>, ValidatorPingState>,
```
michael-yxchen: can we add a comment to mention that it's the union of current and next epoch validators?

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,243 @@
...
+        let removed_nodes = self
+            .validators
+            .keys()
+            .filter(|node_id| !active_validators.iter().any(|v| *v == **node_id))
+            .cloned()
+            .collect::<Vec<_>>();
+
+        for node_id in removed_nodes {
+            self.validators.remove(&node_id);
+        }
```
michael-yxchen:
```suggestion
        self.validators
            .retain(|node_id, _| active_validators.contains(node_id));
```

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,243 @@
...
+        for s in self.schedule.iter_mut() {
+            s.clear();
+        }
+
+        for validator in active_validators.iter() {
+            self.validators.entry(*validator).or_default();
+            let mut hasher = DefaultHasher::new();
+            validator.hash(&mut hasher);
+            let idx = (hasher.finish() as usize) % self.schedule.len();
+            self.schedule[idx].push(*validator);
```
michael-yxchen, line 203: if your goal is to distribute active validators evenly into the buckets, would it be easier to do a random shuffle of all active validators, store that in a vector, and copy a segment on tick? we can include that in a future PR if it sounds reasonable

monad-consensus-state/src/lib.rs
```diff
@@ -2485,6 +2495,47 @@ mod test {
...
+            .unwrap();
+
+        wrapped_state
+            .block_timestamp
+            .update_validators(&val_data, wrapped_state.nodeid, &epoch);
+
+        for _ in 0..30 {
+            wrapped_state.block_timestamp.tick();
+        }
+        std::thread::sleep(Duration::from_millis(1));
```
michael-yxchen: what's the sleep for?

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,243 @@
...
+            validators: BTreeMap::new(),
+            schedule: Box::new([const { Vec::new() }; PING_PERIOD_SEC]),
+            tick: 0,
+        }
+    }
+
+    pub fn pong_received(&mut self, node_id: NodeId<P>, sequence: PingSequence) {
+        if let Some(validator_state) = self.validators.get_mut(&node_id) {
+            if let Some(elapsed) = validator_state.pong_received(sequence) {
+                debug!(?node_id, elapsed_secs = ?elapsed.as_secs(), avg_latency_secs = ?validator_state.avg_latency().unwrap_or_default().as_secs(), "ping latency");
```
michael-yxchen: `as_secs()` returns the number of whole seconds which is too coarse for latency measurements. `as_millis()` is more appropriate

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +271,446 @@ impl BlockTimestamp {
...
     #[test]
     fn test_block_timestamp_validate() {
-        let mut b = BlockTimestamp::new(10, 1);
-        b.update_time(0);
+        let mut b = BlockTimestamp::<NopPubKey>::new(10, 1);
+        let author = NodeId::new(NopKeyPair::from_bytes(&mut [0; 32]).unwrap().pubkey());
+
+        assert!(b.valid_block_timestamp(1, 1, 0, &author).is_err());
+        assert!(b.valid_block_timestamp(2, 1, 0, &author).is_err());
+        assert!(b.valid_block_timestamp(0, 11, 0, &author).is_err());
```
michael-yxchen: we can be more explicit about the error type here with `assert!(matches!())`

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +271,446 @@ impl BlockTimestamp {
...
+        );
+    }
+
+    use monad_testutil::signing::MockSignatures;
+
+    use super::PING_PERIOD_SEC;
+
+    type SignatureType = NopSignature;
+    type SignatureCollection = MockSignatures<SignatureType>;
+
```
michael-yxchen: nit: move to top of test module

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +271,446 @@ impl BlockTimestamp {
...
+        let validators = vec![
+            ValidatorData::<SignatureCollection> {
+                node_id: nodes[1],
+                stake: Stake(1),
+                cert_pubkey: nodes[1].pubkey(),
+            },
+            ValidatorData {
+                node_id: nodes[2],
+                stake: Stake(1),
+                cert_pubkey: nodes[1].pubkey(),
```
michael-yxchen, line 619:
```suggestion
                cert_pubkey: nodes[2].pubkey(),
```

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +271,446 @@ impl BlockTimestamp {
...
+        ];
+
+        s.update_validators(&validators, &my_node, &Epoch(1));
+
+        assert_eq!(s.epoch_validators.len(), 1);
+        assert_eq!(s.epoch_validators.get(&Epoch(1)).unwrap().len(), 2);
+
+        let validators_1 = vec![validators[0].clone()];
+
+        s.update_validators(&validators_1, &my_node, &Epoch(1));
```
michael-yxchen: a correct client never sends two validator set updates for a given epoch. If somehow two updates were sent, they must be identical

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +271,446 @@ impl BlockTimestamp {
...
+                stake: Stake(1),
+                cert_pubkey: nodes[4].pubkey(),
+            },
+        ];
+
+        let mut pings = Vec::new();
+
+        let expected_e1 = [e1_vals[1].node_id, e1_vals[2].node_id];
+        b.update_validators(&e1_vals, &my_node, &e1);
+        if let Some(vals) = b.ping_state.epoch_validators.get(&e1) {
```
michael-yxchen: for all `if let Some(_)` in this test, we should `.expect` instead

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +271,446 @@ impl BlockTimestamp {
...
+            .validators
+            .keys()
+            .all(|x| expected_e1.contains(x)));
+        assert_eq!(b.ping_state.validators.keys().len(), expected_e1.len());
+        pings.clear();
+
+        for _ in 0..PING_PERIOD_SEC {
+            pings.extend(b.tick());
+        }
+        assert!(pings.iter().all(|x| expected_e1.contains(&x.0)));
```
michael-yxchen: correct me if im wrong, we're more interested in that the ping schedule covers every node in the validator set (`expected_e1`), which is the reverse of what's asserted here

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +271,446 @@ impl BlockTimestamp {
...
+            .keys()
+            .all(|x| expected_e2.contains(x)));
+        assert_eq!(b.ping_state.validators.keys().len(), expected_e2.len());
+        pings.clear();
+
+        for _ in 0..PING_PERIOD_SEC {
+            pings.extend(b.tick());
+        }
+        assert!(pings.iter().all(|x| expected_e2.contains(&x.0)));
+        assert_eq!(pings.len(), expected_e2.len());
```
michael-yxchen: I assume you want to test that BlockTimestamp works as expected after entering the same epoch again. Can we add a comment just to note that?

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +271,446 @@ impl BlockTimestamp {
...
+
+        b.ping_state.current_epoch = Epoch(0);
+        b.ping_state.validators.clear();
+        b.ping_state.epoch_validators.clear();
```
michael-yxchen: This is equivalent to creating a new BlockTimestamp object. We should move everything below into a separate test

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +271,446 @@ impl BlockTimestamp {
...
+            },
+        ];
+
+        let mut pings = Vec::new();
+
+        let expected_e1 = [e1_vals[1].node_id, e1_vals[2].node_id];
+        b.update_validators(&e1_vals, &my_node, &e1);
+        if let Some(vals) = b.ping_state.epoch_validators.get(&e1) {
+            assert!(vals.iter().all(|x| expected_e1.contains(x)));
+        }
```
michael-yxchen: I think a better way is to assert the two vectors are equal, by collecting the node ids into a BTreeSet

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +271,446 @@ impl BlockTimestamp {
...
+        assert!(pings.iter().all(|x| expected_e2_e3.contains(&x.0)));
+        assert_eq!(pings.len(), expected_e2_e3.len());
+    }
+
+    #[test]
+    fn test_ticks() {
+        let mut s = PingState::<NopPubKey>::new();
+
+        let my_key = NopKeyPair::from_bytes(&mut [0; 32]).unwrap().pubkey();
```
michael-yxchen: nit: [get_key](https://github.com/category-labs/monad-bft/blob/219a46c76d21993f75543406a2ad5a555d496d36/monad-testutil/src/signing.rs#L139) is available

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +271,446 @@ impl BlockTimestamp {
...
+
+        s.current_epoch = Epoch(1);
+        s.update_validators(&validators, &my_node, &Epoch(1));
+        s.compute_schedule();
+
+        let mut pings = Vec::new();
+        for _ in 0..PING_PERIOD_SEC {
+            pings.extend(s.tick());
+        }
+
```
michael-yxchen, line 942: we should also assert all validators are pinged

monad-state/src/lib.rs
```diff
@@ -483,6 +497,14 @@ where
                 let enc: Vec<&dyn Encodable> = vec![&monad_version, &5u8, &m];
                 Encodable::length(&enc)
             }
+            Self::PingRequest(m) => {
+                let enc: Vec<&dyn Encodable> = vec![&monad_version, &6u8, &m];
+                Encodable::length(&enc)
+            }
+            Self::PingResponse(m) => {
+                let enc: Vec<&dyn Encodable> = vec![&monad_version, &6u8, &m];
```
michael-yxchen:
```suggestion
                let enc: Vec<&dyn Encodable> = vec![&monad_version, &7u8, &m];
```

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,243 @@
...
+            }
+        }
+    }
+
+    fn update_validators<SCT>(
+        &mut self,
+        validators: &Vec<ValidatorData<SCT>>,
+        my_node: &NodeId<P>,
+        val_epoch: &Epoch,
+    ) where
```
michael-yxchen: I'm not sure if passing by reference really make sense here. NodeId and Epoch implements Copy, and are dereferenced in the function. Validators are copied except for one. We can probably make a Clone in the caller in `MonadState.update()`

monad-state/src/lib.rs
```diff
@@ -1097,6 +1148,43 @@ where
                     ))]
                 }
             },
+            MonadEvent::PingRequestEvent(PingEvent { sender, sequence }) => {
+                tracing::debug!(?sender, ?sequence, "received ping request");
+                vec![Command::RouterCommand(RouterCommand::Publish {
+                    target: RouterTarget::PointToPoint(sender),
+                    message: VerifiedMonadMessage::PingResponse(sequence),
+                })]
```
michael-yxchen: The logic of generating response should be moved to BlockTimestamp

monad-state/src/lib.rs
```diff
@@ -1097,6 +1148,43 @@ where
...
+                            target: RouterTarget::PointToPoint(node),
+                            message: VerifiedMonadMessage::PingRequest(sequence),
+                        })
+                    })
+                    .collect::<Vec<_>>();
+                cmds.push(Command::TimerCommand(TimerCommand::Schedule {
+                    duration: PING_TICK_DURATION,
+                    variant: TimeoutVariant::Ping,
+                    on_timeout: MonadEvent::PingTickEvent,
+                }));
```
michael-yxchen: similar here. We can make a `BlockTimestampEvent` enum, which bundles `PingRequestEvent, PingResponseEvent, PingTickEvent, TimestampEnterRoundEvent` under `MonadEvent`, and call `block_timestamp.update(BlockTimestampEvent) -> Vec<BlockTimestampCommand>`. Statesync isn't a great example to follow. Consensus/blocksync are better ones

monad-types/src/lib.rs
```diff
@@ -588,6 +588,8 @@ impl MonadVersion {
         }
     }
 }
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, RlpEncodable, RlpDecodable)]
```
michael-yxchen: use RlpEncodableWrapper and RlpDecodableWrapper. Saves an extra rlp list header

monad-debugger/src/graphql.rs
```diff
@@ -244,6 +244,10 @@ impl<'s> From<&'s MonadEventType> for GraphQLMonadEvent<'s> {
             }
             MonadEvent::StateSyncEvent(event) => Self::StateSyncEvent(GraphQLStateSyncEvent(event)),
             MonadEvent::ConfigEvent(event) => Self::ConfigEvent(GraphQLConfigEvent(event)),
+            MonadEvent::PingRequestEvent(_)
+            | MonadEvent::PingResponseEvent(_)
+            | MonadEvent::PingTickEvent => todo!(),
+            MonadEvent::TimestampEnterRoundEvent(_) => todo!(),
```
michael-yxchen: will this cause a debugger panic?

monad-consensus-state/src/timestamp.rs
mikety: sounds good, thank you for the suggestion. Added a TODO to reimplement.

monad-consensus-state/src/timestamp.rs
mikety: updated to match` update_validators` change

monad-consensus-state/src/timestamp.rs
mikety: I think it should be ok to pass by reference - as this is transient data. epoch is passed by reference to the `MonadState.update()` as well. Please clarify why passing by value is preferred.

monad-debugger/src/graphql.rs
mikety: updated

monad-state/src/lib.rs
mikety: I have added` BlockTimestampEvent enum` but adding` BlockTimestampCommand`  will require refactoring unrelated parts of code due ti circular dependencies. It needs to implement `From` trait to convert ` BlockTimestampCommand`  into       
```
  Command<
            MonadEvent<ST, SCT, EPT>,
            VerifiedMonadMessage<ST, SCT, EPT>,
            ST,
            SCT,
            EPT,
            BPT,
            SBT,
        >,
```
        
        in the MonadState update, and hence the command enum also needs to be defined in the same module, so overall it seems as a lot larger refactor which I suggest to address in a separate PR.

monad-state/src/lib.rs
mikety: Will address in a follow up PR -see response in the next comment.

monad-consensus-state/src/lib.rs
```diff
@@ -2485,6 +2495,46 @@ mod test {
...
+        let epoch = wrapped_state
+            .epoch_manager
+            .get_epoch(verified_message.block_header.round)
+            .unwrap();
+
+        wrapped_state
+            .block_timestamp
+            .update_validators(&val_data, wrapped_state.nodeid, &epoch);
+
+        for _ in 0..30 {
```
michael-yxchen: nit: use the constant instead of 30

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +278,574 @@ impl BlockTimestamp {
...
+    }
+
+    pub fn vote_sent(&mut self, round: Round) {
+        self.last_sent_vote = Some(SentVote {
+            round,
+            timestamp: Instant::now(),
+        });
+    }
+
+    pub fn enter_round(&mut self, epoch: &Epoch) {
```
michael-yxchen: enter_epoch

monad-consensus-state/src/timestamp.rs
```diff
@@ -1,24 +1,250 @@
...
+                let val_set: Vec<_> = validators
+                    .iter()
+                    .filter(|v| v.node_id != *my_node)
+                    .map(|v| v.node_id)
+                    .collect();
+                assert_eq!(
+                    *entry.get(),
+                    val_set,
+                    "Validators update is not matching existing for the same epoch"
+                )
```
michael-yxchen: the assertion in `validators_epoch_mapping` is comparing two BTreeMaps. Here, though unlikely, validator data with different ordering will trigger the assertion

monad-state/src/lib.rs
```diff
@@ -507,6 +528,10 @@ where
     ForwardedTx(Vec<Bytes>),
     /// State Sync msgs
     StateSyncMessage(StateSyncNetworkMessage),
+
+    PingRequest(PingSequence),
+
+    PingResponse(PingSequence),
```
michael-yxchen, line 538: Missing corresponding decode implementation

monad-types/src/lib.rs
```diff
@@ -588,6 +588,8 @@ impl MonadVersion {
         }
     }
 }
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, RlpEncodableWrapper)]
```
michael-yxchen: missing `RlpDecodableWrapper`

monad-state/src/lib.rs
michael-yxchen: ok, can you add a TODO?

monad-state/src/lib.rs
```diff
@@ -406,6 +409,8 @@ where
     BlockSyncResponse(BlockSyncResponseMessage<ST, SCT, EPT>),
     ForwardedTx(Vec<Bytes>),
     StateSyncMessage(StateSyncNetworkMessage),
+    PingRequest(PingSequence),
+    PingResponse(PingSequence),
```
michael-yxchen: it's your call, i have slight preference toward wrapping PingRequest/Response in another enum, and put that enum here

monad-state/src/lib.rs
```diff
@@ -1097,6 +1151,45 @@ where
...
+                        })
+                        .collect::<Vec<_>>();
+                    cmds.push(Command::TimerCommand(TimerCommand::Schedule {
+                        duration: PING_TICK_DURATION,
+                        variant: TimeoutVariant::Ping,
+                        on_timeout: MonadEvent::BlockTimestampEvent(BlockTimestampEvent::PingTick),
+                    }));
+                    cmds
+                }
+                BlockTimestampEvent::TimestampEnterRound { round, epoch } => {
```
michael-yxchen: BlockTimestampEvent::TimestampEnterRound

If round field is unused, we should remove

monad-executor-glue/src/convert/event.rs
```diff
@@ -1098,3 +1110,88 @@ impl<SCT: SignatureCollection> TryFrom<ProtoConfigEvent> for ConfigEvent<SCT> {
...
+                        ))?
+                        .try_into()?;
+                    let sequence = PingSequence(event.sequence);
+                    BlockTimestampEvent::PingRequest { sender, sequence }
+                }
+                proto_block_timestamp_event::Event::PingResponse(event) => {
+                    let sender = event
+                        .sender
+                        .ok_or(ProtoError::MissingRequiredField(
+                            "BlockTimestampEvent.PingResponce.sender".to_owned(),
```
michael-yxchen: nit: typo

monad-executor-glue/src/lib.rs
```diff
@@ -70,6 +72,7 @@ pub enum TimeoutVariant {
     Pacemaker,
     BlockSync(BlockSyncRequestMessage),
     SendVote,
+    Ping,
```
michael-yxchen: super nit: BlockTimestampPing?

monad-executor-glue/src/lib.rs
```diff
@@ -969,6 +972,12 @@ where
     },
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct PingEvent<SCT: SignatureCollection> {
+    pub sender: NodeId<SCT::NodeIdPubKey>,
+    pub sequence: PingSequence,
+}
```
michael-yxchen: seems unused

monad-consensus-state/Cargo.toml
```diff
@@ -14,6 +14,7 @@ monad-chain-config = { workspace = true }
 monad-consensus = { workspace = true }
 monad-consensus-types = { workspace = true }
 monad-crypto = { workspace = true }
+monad-executor-glue = { workspace = true }
```
michael-yxchen: seems unused

monad-state/src/lib.rs
mikety: Removed, also renamed the event TimestampEnterRound -> TimestampEnterEpoch

monad-state/src/lib.rs
mikety: I will do it in the follow up. Added TODO.

michael-yxchen: just want to mention that if the follow up is merged in a different release than this PR, we need to skip the MonadMessage rlp enum prefix allocated to these two messages

monad-state/src/lib.rs
```diff
@@ -1097,6 +1153,46 @@ where
...
+                        .collect::<Vec<_>>();
+                    cmds.push(Command::TimerCommand(TimerCommand::Schedule {
+                        duration: PING_TICK_DURATION,
+                        variant: TimeoutVariant::BLockTimestmpPing,
+                        on_timeout: MonadEvent::BlockTimestampEvent(BlockTimestampEvent::PingTick),
+                    }));
+                    cmds
+                }
+                BlockTimestampEvent::TimestampEnterEpoch { epoch } => {
+                    self.block_timestamp.enter_epoch(&epoch);
```
michael-yxchen: nit: unnecessary ref; can change function to take value

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +278,574 @@ impl BlockTimestamp {
...
         assert!(matches!(
-            b.valid_block_timestamp(5, 8, 0),
-            Some(TimestampAdjustment {
+            b.valid_block_timestamp(5, 10, 0, &author),
+            Ok(Some(TimestampAdjustment {
                 delta: 1,
                 direction: TimestampAdjustmentDirection::Backward
-            })
+            }))
         ));
```
michael-yxchen, line 448: nit: can we add some comments for these assertions?

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +278,574 @@ impl BlockTimestamp {
...
+            ValidatorData {
+                node_id: nodes[2],
+                stake: Stake(1),
+                cert_pubkey: nodes[2].pubkey(),
+            },
+        ];
+
+        s.update_validators(&validators, &my_node, &Epoch(1));
+        assert_eq!(s.epoch_validators.len(), 1);
+        assert_eq!(s.epoch_validators.get(&Epoch(1)).unwrap().len(), 2);
```
michael-yxchen: nit: we can compare with `validators`

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +278,574 @@ impl BlockTimestamp {
...
+        let validators_1 = vec![validators[0].clone()];
+
+        s.update_validators(&validators_1, &my_node, &Epoch(2));
+        assert_eq!(s.epoch_validators.get(&Epoch(2)).unwrap().len(), 1);
+
+        assert!(!s.validators.contains_key(&nodes[1]));
+
+        s.compute_schedule();
+
+        assert!(s.validators.contains_key(&nodes[1]));
```
michael-yxchen: im confused by these assertions. why are we singling out `nodes[1]`?

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +278,574 @@ impl BlockTimestamp {
...
+
+        let validators_2 = vec![validators[1].clone()];
+        s.update_validators(&validators_2, &my_node, &Epoch(3));
+
+        assert_eq!(s.epoch_validators.get_mut(&Epoch(3)).unwrap().len(), 1);
+        s.compute_schedule();
+        assert!(s.validators.contains_key(&nodes[2]));
```
michael-yxchen: same here, what's special about `nodes[2]`?

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +278,574 @@ impl BlockTimestamp {
...
+        let mut pings: Vec<NodeId<NopPubKey>> = Vec::new();
+
+        let expected_e1 = [e1_vals[1].node_id, e1_vals[2].node_id];
+        b.update_validators(&e1_vals, &my_node, &e1);
+        let vals = b
+            .ping_state
+            .epoch_validators
+            .get(&e1)
+            .expect("get validators");
+        assert!(vals.iter().all(|x| expected_e1.contains(x)));
```
michael-yxchen: we should collect into BTreeSets if we want to assert equality. Current assertion works if `vals` is a subset of `expected_e1`

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +278,574 @@ impl BlockTimestamp {
...
+
+        b.enter_epoch(&e1);
+        assert_eq!(b.ping_state.current_epoch, e1);
+        assert!(b
+            .ping_state
+            .validators
+            .keys()
+            .all(|x| expected_e1.contains(x)));
+        assert_eq!(b.ping_state.validators.keys().len(), expected_e1.len());
+        pings.clear();
```
michael-yxchen: nit: initialize `pings` here

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +278,574 @@ impl BlockTimestamp {
...
+        b.update_validators(&e2_vals, &my_node, &e2);
+        b.enter_epoch(&e2);
+        assert_eq!(b.ping_state.current_epoch, e2);
+        assert!(b
+            .ping_state
+            .validators
+            .keys()
+            .all(|x| expected_e2.contains(x)));
+        assert_eq!(b.ping_state.validators.keys().len(), expected_e2.len());
+        pings.clear();
```
michael-yxchen: alternatively, recreate the pings vec

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +278,574 @@ impl BlockTimestamp {
...
+            .ping_state
+            .epoch_validators
+            .get(&e2)
+            .expect("get validators");
+        assert_eq!(
+            vals.iter().copied().collect::<BTreeSet<_>>(),
+            expected_e2.iter().copied().collect::<BTreeSet<_>>()
+        );
+
+        let expected_e1_e2 = [e1_vals[1].node_id, e2_vals[0].node_id, e2_vals[1].node_id];
```
michael-yxchen: can we use the original nodes array here or add comments? having to reference e1_vals and e2_vals hurts readability

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +278,574 @@ impl BlockTimestamp {
...
+
+        assert_eq!(
+            b.ping_state
+                .validators
+                .keys()
+                .copied()
+                .collect::<BTreeSet<_>>(),
+            expected_e1_e2.iter().copied().collect::<BTreeSet<_>>()
+        );
+        pings.clear();
```
michael-yxchen: same - move ping initialization here

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +278,574 @@ impl BlockTimestamp {
...
+                    .iter()
+                    .map(|x| x.0)
+                    .collect::<Vec<NodeId<NopPubKey>>>(),
+            );
+        }
+
+        assert_eq!(
+            pings.iter().copied().collect::<BTreeSet<_>>(),
+            expected_e2_e3.iter().copied().collect::<BTreeSet<_>>()
+        );
```
michael-yxchen, line 896: we should do all equality checks like this, not subset + cardinality

monad-consensus-state/src/timestamp.rs
```diff
@@ -52,79 +278,574 @@ impl BlockTimestamp {
...
+    }
+
+    #[test]
+    fn test_ticks() {
+        let mut s = PingState::<NopPubKey>::new();
+
+        //let my_key = NopKeyPair::from_bytes(&mut [0; 32]).unwrap().pubkey();
```
michael-yxchen: nit: remove

monad-consensus-state/src/timestamp.rs
mikety: Tests that compute_schedule will add new validators. Added a comment.

monad-consensus-state/src/timestamp.rs
mikety: added comments and updated the test

monad-consensus-state/src/timestamp.rs
mikety: added comments

mikety:
@michael-yxchen  thank you for the thorough and detailed review. I addressed the feedback and rebased on top of the main branch.

PR #1680, opened 2025-03-24
Initial peer discovery setup

eerkaijun:
Standalone new `monad-peer-discovery` crate with basic unit tests. The crate will be integrated into `monad-raptorcast`. Would appreciate any early feedback. Will squash the commits before merging

Reference: https://github.com/category-labs/category-internal/issues/1262 (this PR mainly contains connect and discover operations)

### DIFF
--- a/monad-peer-discovery/src/discovery.rs
+++ b/monad-peer-discovery/src/discovery.rs
@@ -1,5 +1,6 @@
 use std::{
     collections::{BTreeMap, HashMap},
+    net::SocketAddrV4,
     time::Instant,
 };
 
@@ -266,7 +267,7 @@ where
     fn get_sock_addr_by_id(
         &self,
         id: NodeId<CertificateSignaturePubKey<Self::SignatureType>>,
-    ) -> Option<std::net::SocketAddr> {
+    ) -> Option<SocketAddrV4> {
         self.peer_info
             .get(&id)
             .map(|info| info.name_record.name_record.address)
@@ -275,10 +276,7 @@ where
 
 #[cfg(test)]
 mod tests {
-    use std::{
-        net::{SocketAddr, SocketAddrV4},
-        str::FromStr,
-    };
+    use std::{net::SocketAddrV4, str::FromStr};
 
     use alloy_rlp::Encodable;
     use monad_crypto::{
@@ -295,10 +293,10 @@ mod tests {
     type KeyPairType = NopKeyPair;
     type SignatureType = NopSignature;
 
-    fn generate_name_record(keypair: &KeyPairType) -> MonadNameRecord<SignatureType> {
+    fn generate_name_record(keypair: &KeyPairType, seq_num: u64) -> MonadNameRecord<SignatureType> {
         let name_record = NameRecord {
-            address: SocketAddr::V4(SocketAddrV4::from_str("1.1.1.1:8000").unwrap()),
-            seq: 0,
+            address: SocketAddrV4::from_str("1.1.1.1:8000").unwrap(),
+            seq: seq_num,
         };
         let mut encoded = Vec::new();
         name_record.encode(&mut encoded);
@@ -336,10 +334,10 @@ mod tests {
         let peer_info = BTreeMap::from([(peer1_pubkey, PeerInfo {
             last_ping: None,
             last_seen: None,
-            name_record: generate_name_record(peer1),
+            name_record: generate_name_record(peer1, 0),
         })]);
         let mut state = PeerDiscovery {
-            self_record: generate_name_record(peer0),
+            self_record: generate_name_record(peer0, 0),
             peer_info,
             outstanding_lookup_requests: HashMap::new(),
             metrics: HashMap::new(),
@@ -373,6 +371,37 @@ mod tests {
     }
 
     #[test]
+    fn test_drop_pong_with_incorrect_ping_id() {
+        let keys = create_keys::<SignatureType>(2);
+        let peer0 = &keys[0];
+        let peer1 = &keys[1];
+        let peer1_pubkey = NodeId::new(peer1.pubkey());
+
+        let peer_info = BTreeMap::from([(peer1_pubkey, PeerInfo {
+            last_ping: None,
+            last_seen: None,
+            name_record: generate_name_record(peer1, 0),
+        })]);
+        let mut state = PeerDiscovery {
+            self_record: generate_name_record(peer0, 0),
+            peer_info,
+            outstanding_lookup_requests: HashMap::new(),
+            metrics: HashMap::new(),
+            rng: ChaCha8Rng::seed_from_u64(123456),
+        };
+
+        // should not record pong when ping_id doesn't match
+        state.handle_pong(peer1_pubkey, Pong {
+            ping_id: 1,
+            local_record_seq: 0,
+        });
+        let peer_info = state.peer_info.get(&peer1_pubkey);
+        assert!(peer_info.is_some());
+        assert!(peer_info.unwrap().last_ping.is_none());
+        assert!(peer_info.unwrap().last_seen.is_none());
+    }
+
+    #[test]
     fn test_peer_lookup() {
         let keys = create_keys::<SignatureType>(3);
         let peer0 = &keys[0];
@@ -382,7 +411,7 @@ mod tests {
         let peer2_pubkey = NodeId::new(peer2.pubkey());
 
         let mut state: PeerDiscovery<SignatureType> = PeerDiscovery {
-            self_record: generate_name_record(peer0),
+            self_record: generate_name_record(peer0, 0),
             peer_info: BTreeMap::new(),
             outstanding_lookup_requests: HashMap::new(),
             metrics: HashMap::new(),
@@ -395,7 +424,7 @@ mod tests {
         assert!(!state.peer_info.contains_key(&peer2_pubkey));
         let requests = extract_lookup_requests(cmds);
 
-        let record = generate_name_record(peer2);
+        let record = generate_name_record(peer2, 0);
         state.handle_peer_lookup_response(peer1_pubkey, PeerLookupResponse {
             lookup_id: requests[0].lookup_id,
             target: peer2_pubkey,
@@ -405,5 +434,111 @@ mod tests {
         // peer2 should be added to peer info and outstanding requests should be cleared
         assert!(state.peer_info.contains_key(&peer2_pubkey));
         assert_eq!(state.outstanding_lookup_requests.keys().len(), 0);
+    }
+
+    #[test]
+    fn test_update_name_record_sequence_number() {
+        let keys = create_keys::<SignatureType>(3);
+        let peer0 = &keys[0];
+        let peer1 = &keys[1];
+        let peer1_pubkey = NodeId::new(peer1.pubkey());
+
+        let peer_info = BTreeMap::from([(peer1_pubkey, PeerInfo {
+            last_ping: None,
+            last_seen: None,
+            name_record: generate_name_record(peer1, 1),
+        })]);
+        let mut state: PeerDiscovery<SignatureType> = PeerDiscovery {
+            self_record: generate_name_record(peer0, 0),
+            peer_info,
+            outstanding_lookup_requests: HashMap::from([(1, peer1_pubkey), (2, peer1_pubkey)]),
+            metrics: HashMap::new(),
+            rng: ChaCha8Rng::seed_from_u64(123456),
+        };
+
+        // should not update name record if record has lower sequence number
+        let record = generate_name_record(peer1, 0);
+        state.handle_peer_lookup_response(peer1_pubkey, PeerLookupResponse {
+            lookup_id: 1,
+            target: peer1_pubkey,
+            name_records: vec![record],
+        });
+        assert_eq!(
+            state
+                .peer_info
+                .get(&peer1_pubkey)
+                .unwrap()
+                .name_record
+                .seq(),
+            1
+        );
+
+        // should update name record if record has higher sequence number
+        let record = generate_name_record(peer1, 2);
+        state.handle_peer_lookup_response(peer1_pubkey, PeerLookupResponse {
+            lookup_id: 2,
+            target: peer1_pubkey,
+            name_records: vec![record],
+        });
+        assert_eq!(
+            state
+                .peer_info
+                .get(&peer1_pubkey)
+                .unwrap()
+                .name_record
+                .seq(),
+            2
+        );
+    }
+
+    #[test]
+    fn test_drop_invalid_lookup_response() {
+        let keys = create_keys::<SignatureType>(3);
+        let peer0 = &keys[0];
+        let peer1 = &keys[1];
+        let peer1_pubkey = NodeId::new(peer1.pubkey());
+
+        let mut state: PeerDiscovery<SignatureType> = PeerDiscovery {
+            self_record: generate_name_record(peer0, 0),
+            peer_info: BTreeMap::new(),
+            outstanding_lookup_requests: HashMap::new(),
+            metrics: HashMap::new(),
+            rng: ChaCha8Rng::seed_from_u64(123456),
+        };
+
+        // should not record peer lookup response if not in outstanding requests
+        let record = generate_name_record(peer1, 0);
+        state.handle_peer_lookup_response(peer1_pubkey, PeerLookupResponse {
+            lookup_id: 1,
+            target: peer1_pubkey,
+            name_records: vec![record],
+        });
+        assert!(!state.peer_info.contains_key(&peer1_pubkey));
+    }
+
+    #[test]
+    fn test_drop_lookup_response_that_exceeds_max_peers() {
+        let keys = create_keys::<SignatureType>(3);
+        let peer0 = &keys[0];
+        let peer1 = &keys[1];
+        let peer1_pubkey = NodeId::new(peer1.pubkey());
+        let lookup_id = 1;
+
+        let mut state: PeerDiscovery<SignatureType> = PeerDiscovery {
+            self_record: generate_name_record(peer0, 0),
+            peer_info: BTreeMap::new(),
+            outstanding_lookup_requests: HashMap::from([(lookup_id, peer1_pubkey)]),
+            metrics: HashMap::new(),
+            rng: ChaCha8Rng::seed_from_u64(123456),
+        };
+
+        // should not record peer lookup response if number of records exceed max
+        let record = generate_name_record(peer1, 0);
+        state.handle_peer_lookup_response(peer1_pubkey, PeerLookupResponse {
+            lookup_id,
+            target: peer1_pubkey,
+            name_records: vec![record; MAX_PEER_IN_RESPONSE + 1],
+        });
+        assert!(!state.peer_info.contains_key(&peer1_pubkey));
     }
 }
--- a/monad-peer-discovery/src/lib.rs
+++ b/monad-peer-discovery/src/lib.rs
@@ -1,6 +1,6 @@
 use std::{
     collections::HashMap,
-    net::{Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6},
+    net::{Ipv4Addr, SocketAddrV4},
     time::Duration,
 };
 
@@ -44,22 +44,15 @@ pub struct PeerLookupResponse<ST: CertificateSignatureRecoverable> {
 
 #[derive(Debug, Clone, PartialEq)]
 pub struct NameRecord {
-    pub address: SocketAddr,
+    pub address: SocketAddrV4,
     pub seq: u64,
 }
 
 impl Encodable for NameRecord {
     fn encode(&self, out: &mut dyn alloy_rlp::BufMut) {
-        match self.address {
-            SocketAddr::V4(addr) => {
-                let enc: [&dyn Encodable; 4] = [&1u8, &addr.ip().octets(), &addr.port(), &self.seq];
-                encode_list::<_, dyn Encodable>(&enc, out);
-            }
-            SocketAddr::V6(addr) => {
-                let enc: [&dyn Encodable; 4] = [&2u8, &addr.ip().octets(), &addr.port(), &self.seq];
-                encode_list::<_, dyn Encodable>(&enc, out);
-            }
-        }
+        let enc: [&dyn Encodable; 3] =
+            [&self.address.ip().octets(), &self.address.port(), &self.seq];
+        encode_list::<_, dyn Encodable>(&enc, out);
     }
 }
 
@@ -67,29 +60,12 @@ impl Decodable for NameRecord {
     fn decode(buf: &mut &[u8]) -> alloy_rlp::Result<Self> {
         let buf = &mut alloy_rlp::Header::decode_bytes(buf, true)?;
 
-        let addr_type = u8::decode(buf)?;
-        let addr = match addr_type {
-            1 => {
-                let Ok(ip) = <[u8; 4]>::decode(buf) else {
-                    warn!("ip address decode failed: {:?}", buf);
-                    return Err(alloy_rlp::Error::Custom("Invalid IPv4 address"));
-                };
-                let port = u16::decode(buf)?;
-                SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::from(ip), port))
-            }
-            2 => {
-                let Ok(ip) = <[u8; 16]>::decode(buf) else {
-                    warn!("ip address decode failed: {:?}", buf);
-                    return Err(alloy_rlp::Error::Custom("Invalid IPv6 address"));
-                };
-                let port = u16::decode(buf)?;
-                SocketAddr::V6(SocketAddrV6::new(Ipv6Addr::from(ip), port, 0, 0))
-            }
-            _ => {
-                warn!("ip address decode failed: {:?}", buf);
-                return Err(alloy_rlp::Error::Custom("Invalid IP enum"));
-            }
+        let Ok(ip) = <[u8; 4]>::decode(buf) else {
+            warn!("ip address decode failed: {:?}", buf);
+            return Err(alloy_rlp::Error::Custom("Invalid IPv4 address"));
         };
+        let port = u16::decode(buf)?;
+        let addr = SocketAddrV4::new(Ipv4Addr::from(ip), port);
         let seq = u64::decode(buf)?;
 
         Ok(Self { address: addr, seq })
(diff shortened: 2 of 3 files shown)


monad-peer-discovery/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+        }
+    }
+}
+
+pub enum PeerDiscoveryTimerCommand<E, ST: CertificateSignatureRecoverable> {
+    Schedule {
+        node_id: PeerId<ST>, // unique identify request to node
+        duration: Duration,
+        on_timeout: E,
+    },
```
michael-yxchen: I think we need timers for both Ping and PeerLookupRequests. NodeId may not be enough to uniquely identify

monad-peer-discovery/src/discovery.rs
```diff
@@ -0,0 +1,334 @@
...
+#[derive(Debug, Clone)]
+pub struct PeerInfo<ST: CertificateSignatureRecoverable> {
+    pub last_ping: Option<Ping>,
+    pub last_seen: Option<Instant>,
+    pub name_record: MonadNameRecord<ST>,
+}
+
+#[derive(Default)]
+pub struct PeerDiscovery<ST: CertificateSignatureRecoverable> {
+    pub local_record_seq: u64,
```
michael-yxchen: A trait function to update local name record is helpful for testing and updating local name record at runtime

monad-peer-discovery/src/discovery.rs
```diff
@@ -0,0 +1,334 @@
...
+
+        let mut cmds = Vec::new();
+
+        // must have a MonadNameRecord to send a ping
+        if !self.peer_info.contains_key(&to) {
+            return cmds;
+        }
+
+        let ping_msg = Ping {
+            id: random(),
```
michael-yxchen: nit: ThreadRng provides quite high level of security. Does it come with any perf penalties, comparing to pseudorandom rng seeded with something on init?

monad-peer-discovery/src/discovery.rs
```diff
@@ -0,0 +1,334 @@
...
+}
+
+impl<ST> PeerDiscoveryAlgo for PeerDiscovery<ST>
+where
+    ST: CertificateSignatureRecoverable,
+{
+    type SignatureType = ST;
+
+    fn send_ping(&mut self, to: PeerId<ST>) -> Vec<PeerDiscoveryCommand<ST>> {
+        trace!(?to, "sending ping request");
```
michael-yxchen: we need to remember to set `monad_peer_discovery=trace` or bump it to debug during the initial testing phase

monad-peer-discovery/src/discovery.rs
```diff
@@ -0,0 +1,334 @@
...
+
+        let mut cmds = Vec::new();
+
+        // must have a MonadNameRecord to send a ping
+        if !self.peer_info.contains_key(&to) {
+            return cmds;
```
michael-yxchen, line 58: log here. Some invariant is broken I think?

monad-peer-discovery/src/discovery.rs
```diff
@@ -0,0 +1,334 @@
...
+
+        let ping_msg = Ping {
+            id: random(),
+            local_record_seq: self.local_record_seq,
+        };
+
+        // record ping in peer_info
+        self.peer_info
+            .entry(to)
+            .and_modify(|entry| entry.last_ping = Some(ping_msg));
```
michael-yxchen: nit: I'm not sure how useful the entry interface is here. function early returns if `to` doesn't exist. We can do something like
```
let Some(peer_entry) = self.peer_info.get_mut(&to) else {
    return cmds;
}
```
instead of `contains_key`. Then modify `peer_entry` here

monad-peer-discovery/src/discovery.rs
```diff
@@ -0,0 +1,334 @@
...
+
+        let cmds = Vec::new();
+
+        // if ping id matches, update last_seen time
+        self.peer_info.entry(from).and_modify(|info| {
+            if info.last_ping.as_ref().map(|last_ping| last_ping.id) == Some(pong_msg.ping_id) {
+                info.last_ping = None;
+                info.last_seen = Some(Instant::now());
+            }
+        });
```
michael-yxchen: Log if ping id doesn't match. Spam likely

monad-peer-discovery/src/discovery.rs
```diff
@@ -0,0 +1,334 @@
...
+        let cmds = Vec::new();
+
+        // if lookup id is not in outstanding requests, drop the response
+        if self
+            .outstanding_lookup_requests
+            .get(&response.lookup_id)
+            .filter(|&peer| peer == &from)
+            .is_none()
+        {
+            return cmds;
```
michael-yxchen, line 219: add log here

monad-peer-discovery/src/discovery.rs
```diff
@@ -0,0 +1,334 @@
...
+        for (node_id, name_record) in response.name_records {
+            self.peer_info
+                .entry(node_id)
+                .and_modify(|info| info.name_record = name_record.clone())
+                .or_insert_with(|| PeerInfo {
+                    last_ping: None,
+                    last_seen: None,
+                    name_record,
+                });
+        }
```
michael-yxchen: each name record has a signature that we should verify

monad-peer-discovery/src/discovery.rs
eerkaijun: slightly related scenario is:
- Node A send ping to node B
- Node A has not yet receive pong from node B, timer ticks, and Node A sends an updated ping to Node B (last_ping is updated to the more recent ping)
- Node A receives the first pong from node B, ping ID does not match, pong is dropped

currently implementation does not update last_seen. we might want to consider whether this is the desired behavior.

monad-peer-discovery/src/discovery.rs
eerkaijun: good point, agree that we don't need to be cryptographical secure for id generation. perhaps something like `StdRng::seed_from_u64`?

michael-yxchen: what we do in blocksync is seed a `ChaCha8Rng` generator on init, then use that rng as source of random values

monad-peer-discovery/src/discovery.rs
michael-yxchen: I'd imagine these pings are relatively infrequent. Maybe once per minute or something. We should consider this behavior as offline(network heavily congested)/spam

eerkaijun: makes sense, added log

monad-peer-discovery/src/lib.rs
eerkaijun: thinking out loud on validator peer discovery and full node peer discovery

validator peer discovery: my thought is that when a new validator joins and stake, the stateful precompile will emit some sort of event, which existing validators will listen to and send peer lookup requests to try to obtain the ip address of the new validator. if that's the case then it will be driven by staking events rather than a timer event

full node peer discovery: perhaps a timer might make more sense here, so a validator periodically send peer lookup requests to discover new full nodes. alternatively, i think another way which seems more intuitive to me is that full nodes should send ping to validators when they first want to join the network. validators who still have capacity to take full nodes would then send peer lookup requests to try to obtain the ip address of the full nodes. with this we might not need a timer for sending peer lookup requests too

monad-peer-discovery/src/discovery.rs
eerkaijun: to update the node own's name record or its peers' name record?

michael-yxchen: own record

monad-peer-discovery/src/discovery.rs
```diff
@@ -0,0 +1,365 @@
...
+{
+    type SignatureType = ST;
+
+    fn send_ping(&mut self, to: PeerId<ST>) -> Vec<PeerDiscoveryCommand<ST>> {
+        debug!(?to, "sending ping request");
+
+        let mut cmds = Vec::new();
+
+        // must have a MonadNameRecord to send a ping
+        let peer_entry = match self.peer_info.get_mut(&to) {
```
andr-dev: If you do
```rust
let Some(entry) = self.peer_info.get_mut(&to) else {
    ...
}
```
You save one indent level :)

monad-peer-discovery/src/discovery.rs
```diff
@@ -0,0 +1,365 @@
...
+        let cmds = Vec::new();
+
+        // if ping id matches, update last_seen time
+        self.peer_info.entry(from).and_modify(|info| {
+            if info.last_ping.as_ref().map(|last_ping| last_ping.id) == Some(pong_msg.ping_id) {
+                info.last_ping = None;
+                info.last_seen = Some(Instant::now());
+            } else {
+                debug!(?from, "dropping pong because ping id does not match");
+            }
```
andr-dev: Might be nice to have a separate debug message for if we receive a `pong` when there's no `last_ping`.

monad-peer-discovery/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+            }
+        }
+    }
+}
+
+impl Decodable for NameRecord {
+    fn decode(buf: &mut &[u8]) -> alloy_rlp::Result<Self> {
+        let alloy_rlp::Header {
+            list,
+            payload_length: _,
```
andr-dev: We should verify that the `payload_length` doesn't exceed some max value. Additionally, why are we using rlp encoding for this? If we use some other encoding mechanism, we could assert that the payload length is a fixed size right?

monad-peer-discovery/src/lib.rs
michael-yxchen: I could've be more clear with what I meant. When node is made aware of new validators, it should try to obtain its name record. This process requires retry, hence needing a timer

An alternative code path would be
- send periodic pings to validators
- if name record of A is not known (dest of ping not known), send lookup to a random peer, and schedule timer for looking up A again. This timeout will probably be smaller than ping timer

With the above proposal, we don't need to augment the timer. NodeId is enough to identify

michael-yxchen: we're on the same page regarding full node part. Full node should initiate. Validators doesn't have to keep discovering full nodes

monad-peer-discovery/src/discovery.rs
eerkaijun: great suggestion

monad-peer-discovery/src/discovery.rs
```diff
@@ -0,0 +1,387 @@
...
+
+        // if sender of ping does not exist in local record or has higher sequence number
+        // send a PeerLookupRequest to get the node information subsequently
+        if self
+            .peer_info
+            .get(&from)
+            .map(|info| info.name_record.seq() < ping_msg.local_record_seq)
+            .unwrap_or(true)
+        {
+            // TODO: cmd that sends PeerLookupRequest
```
michael-yxchen: what blocks us from sending a PeerLookupRequest? is it that we need the peer address to send the lookup?

michael-yxchen: if peer record has changed (seq num bumped), who are we sending the pong to?

monad-peer-discovery/src/discovery.rs
```diff
@@ -0,0 +1,387 @@
...
+
+        let cmds = Vec::new();
+
+        // if ping id matches, update last_seen time
+        self.peer_info.entry(from).and_modify(|info| {
+            if info.last_ping.as_ref().map(|last_ping| last_ping.id) == Some(pong_msg.ping_id) {
+                info.last_ping = None;
+                info.last_seen = Some(Instant::now());
+            } else {
+                debug!(?from, "dropping pong, ping id not found or does not match");
```
michael-yxchen, line 131: entry not found won't trigger this line, because of the `and_modify`. It might be more readable to use
```rust
if let Some(info) = self.peer_info.get(from) {
  ...
} else {
  ...
}
```

monad-peer-discovery/src/discovery.rs
```diff
@@ -0,0 +1,387 @@
...
+};
+
+#[derive(Debug, Clone)]
+pub struct PeerInfo<ST: CertificateSignatureRecoverable> {
+    pub last_ping: Option<Ping>,
+    pub last_seen: Option<Instant>,
+    pub name_record: MonadNameRecord<ST>,
+}
+
+pub struct PeerDiscovery<ST: CertificateSignatureRecoverable> {
```
michael-yxchen, line 29: we should sign and store our own MonadNameRecord in PeerDiscovery state

monad-peer-discovery/src/discovery.rs
```diff
@@ -0,0 +1,387 @@
...
+            let mut encoded = Vec::new();
+            name_record.name_record.encode(&mut encoded);
+            if let Err(e) = name_record.signature.verify(&encoded, &node_id.pubkey()) {
+                warn!(
+                    ?node_id,
+                    ?e,
+                    "invalid name record signature, dropping request..."
+                );
+                continue;
+            }
```
michael-yxchen: move this to `MonadNameRecord::verify(...)`

monad-peer-discovery/src/discovery.rs
```diff
@@ -0,0 +1,387 @@
...
+        // if lookup id is not in outstanding requests, drop the response
+        if self
+            .outstanding_lookup_requests
+            .get(&response.lookup_id)
+            .filter(|&peer| peer == &from)
+            .is_none()
+        {
+            warn!(
+                ?response,
+                "peer lookup response not in oustanding requests, dropping requests..."
```
```suggestion
                "peer lookup response not in outstanding requests, dropping response..."
```

monad-peer-discovery/src/lib.rs
```diff
@@ -1,2 +1,291 @@
...
+                let Ok(ip) = <[u8; 16]>::decode(buf) else {
+                    warn!("ip address decode failed: {:?}", buf);
+                    return Err(alloy_rlp::Error::Custom("Invalid IPv6 address"));
+                };
+                let port = u16::decode(buf)?;
+                SocketAddr::V6(SocketAddrV6::new(Ipv6Addr::from(ip), port, 0, 0))
+            }
+            _ => {
+                warn!("ip address decode failed: {:?}", buf);
+                return Err(alloy_rlp::Error::Custom("Invalid IP"));
```
michael-yxchen: invalid rlp enum?

monad-peer-discovery/src/lib.rs
```diff
@@ -1,2 +1,291 @@
...
+            payload_length,
+        } = alloy_rlp::Header::decode(buf)?;
+        if !list {
+            return Err(alloy_rlp::Error::UnexpectedString);
+        }
+
+        let remaining_len = buf.len();
+        if remaining_len == 0 || remaining_len < payload_length {
+            return Err(alloy_rlp::Error::InputTooShort);
+        }
```
michael-yxchen:
```suggestion
        let mut payload = alloy_rlp::Header::decode_bytes(buf, true)?;
```

monad-peer-discovery/src/discovery.rs
eerkaijun: already added this to another PR, but let me pull it over here

eerkaijun: the intention is to send the peer lookup request and pong back to the original source IP address (which will probably require the router to keep this IP information somewhere)

monad-peer-discovery/src/lib.rs
eerkaijun: nice, didn't know there's this function - updated

monad-peer-discovery/src/discovery.rs
eerkaijun: added in next pr (router swarm testing)

monad-peer-discovery/src/lib.rs
eerkaijun: good point on RLP encoding, will discuss

monad-peer-discovery/src/discovery.rs
eerkaijun: makes sense, was including it in `peer_info` but agree would be easier to add it to a separate field too

eerkaijun: updated

monad-peer-discovery/src/discovery.rs
eerkaijun: updated

monad-peer-discovery/src/lib.rs
eerkaijun: i don't think we need a separate enum for encoding/decoding related errors, most likely it won't be handled (ie we don't really need to differentiate it, logging is sufficient i think)

monad-peer-discovery/src/discovery.rs

monad-peer-discovery/src/lib.rs
michael-yxchen: sry i meant to put "Invalid rlp enum" in the custom message. Agree that we don't want another error enum

monad-peer-discovery/src/discovery.rs
```diff
@@ -0,0 +1,396 @@
...
+        if self
+            .peer_info
+            .get(&from)
+            .map(|info| info.name_record.seq() < ping_msg.local_record_seq)
+            .unwrap_or(true)
+        {
+            debug!(
+                "Ping sender not found in local record, sending peer lookup request to :{:?}",
+                from
+            );
```
michael-yxchen: nit: 
```suggestion
            debug!(
                target=?from, "Ping sender not found in local record, sending peer lookup request to target",
            );
```

monad-peer-discovery/src/discovery.rs
```diff
@@ -0,0 +1,396 @@
...
+
+        let mut cmds = Vec::new();
+        let peer_lookup_request = PeerLookupRequest::<ST> {
+            lookup_id: self.rng.next_u32(),
+            target,
+        };
+
+        // record request
+        self.outstanding_lookup_requests
+            .insert(peer_lookup_request.lookup_id, to);
```
michael-yxchen, line 158: we should expire entries if we don't hear back in some time

monad-peer-discovery/src/discovery.rs
```diff
@@ -0,0 +1,396 @@
...
+        }
+
+        // update peer info
+        for (node_id, name_record) in response.name_records {
+            // verify signature of name record
+            if let Err(e) = name_record.verify(node_id) {
+                warn!(
+                    ?node_id,
+                    ?e,
+                    "invalid name record signature, dropping request..."
```
michael-yxchen: nit
```suggestion
                    "invalid name record signature, dropping record..."
```

monad-peer-discovery/src/discovery.rs
```diff
@@ -0,0 +1,396 @@
...
+
+            // insert name record
+            self.peer_info
+                .entry(node_id)
+                .and_modify(|info| info.name_record = name_record.clone())
+                .or_insert_with(|| PeerInfo {
+                    last_ping: None,
+                    last_seen: None,
+                    name_record,
+                });
```
michael-yxchen: we should only modify peer record if it comes with a higher seq num

eerkaijun: ok yeah good catch

monad-peer-discovery/src/discovery.rs
eerkaijun: yeap working on pruning related logic in a separate pr. we also need to prune peers that have not responded to pings after some time

monad-peer-discovery/src/discovery.rs
```diff
@@ -0,0 +1,399 @@
...
+
+        let mut cmds = Vec::new();
+
+        // if sender of ping does not exist in local record or has higher sequence number
+        // send a PeerLookupRequest to get the node information subsequently
+        if self
+            .peer_info
+            .get(&from)
+            .map(|info| info.name_record.seq() < ping_msg.local_record_seq)
+            .unwrap_or(true)
```
omegablitz: nit: [Option::is_none_or](https://doc.rust-lang.org/std/option/enum.Option.html#method.is_none_or) is easier to read

monad-peer-discovery/src/discovery.rs
```diff
@@ -0,0 +1,399 @@
...
+        from: NodeId<CertificateSignaturePubKey<ST>>,
+        pong_msg: Pong,
+    ) -> Vec<PeerDiscoveryCommand<ST>> {
+        debug!(?from, ?pong_msg, "handling pong response");
+
+        let cmds = Vec::new();
+
+        // if ping id matches, update last_seen time
+        if let Some(info) = self.peer_info.get_mut(&from) {
+            if info.last_ping.as_ref().map(|last_ping| last_ping.id) == Some(pong_msg.ping_id) {
```
omegablitz: nit: [Option::is_some_and](https://doc.rust-lang.org/std/option/enum.Option.html#method.is_some_and) is easier to read

monad-peer-discovery/src/lib.rs
```diff
@@ -1,2 +1,289 @@
...
+                let port = u16::decode(buf)?;
+                SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::from(ip), port))
+            }
+            2 => {
+                let Ok(ip) = <[u8; 16]>::decode(buf) else {
+                    warn!("ip address decode failed: {:?}", buf);
+                    return Err(alloy_rlp::Error::Custom("Invalid IPv6 address"));
+                };
+                let port = u16::decode(buf)?;
+                SocketAddr::V6(SocketAddrV6::new(Ipv6Addr::from(ip), port, 0, 0))
```
omegablitz: AFAIK we don't support ipv6 - is there a reason to include this? I know that dataplane used to panic if fed an ipv6 address, not sure if that's still the case.

monad-peer-discovery/src/discovery.rs
```diff
@@ -0,0 +1,399 @@
...
+
+use crate::{
+    MonadNameRecord, PeerDiscMetrics, PeerDiscoveryAlgo, PeerDiscoveryCommand,
+    PeerDiscoveryMessage, PeerLookupRequest, PeerLookupResponse, Ping, Pong,
+};
+
+#[derive(Debug, Clone)]
+pub struct PeerInfo<ST: CertificateSignatureRecoverable> {
+    pub last_ping: Option<Ping>,
+    pub last_seen: Option<Instant>,
```
omegablitz, line 25: Is this used for anything? Just will note that `Instant` is effectful, which will make swarm testing harder.

monad-peer-discovery/src/discovery.rs
```diff
@@ -0,0 +1,399 @@
...
+                ?response,
+                "peer lookup response not in outstanding requests, dropping response..."
+            );
+            return cmds;
+        }
+
+        // update peer info
+        for (node_id, name_record) in response.name_records {
+            // verify signature of name record
+            if let Err(e) = name_record.verify(node_id) {
```
omegablitz: Each verify takes ~40us of cpu time. I think the sequence number should be checked first.

monad-peer-discovery/src/discovery.rs
```diff
@@ -0,0 +1,399 @@
...
+    pub last_ping: Option<Ping>,
+    pub last_seen: Option<Instant>,
+    pub name_record: MonadNameRecord<ST>,
+}
+
+pub struct PeerDiscovery<ST: CertificateSignatureRecoverable> {
+    pub self_record: MonadNameRecord<ST>,
+    pub peer_info: BTreeMap<NodeId<CertificateSignaturePubKey<ST>>, PeerInfo<ST>>,
+    // maps lookup id => receiver of the request
+    pub outstanding_lookup_requests: HashMap<u32, NodeId<CertificateSignaturePubKey<ST>>>,
```
omegablitz, line 33: Will this leak if the peer never responds?

monad-peer-discovery/src/discovery.rs
```diff
@@ -0,0 +1,399 @@
...
+        from: NodeId<CertificateSignaturePubKey<ST>>,
+        response: PeerLookupResponse<ST>,
+    ) -> Vec<PeerDiscoveryCommand<ST>> {
+        debug!(?from, ?response, "handling peer lookup response");
+
+        let cmds = Vec::new();
+
+        // if lookup id is not in outstanding requests, drop the response
+        if self
+            .outstanding_lookup_requests
```
omegablitz, line 211: What's the purpose of this check? We aren't verifying that the peer is sending back what we asked for - is that intended?

monad-peer-discovery/src/discovery.rs
```diff
@@ -0,0 +1,399 @@
...
+        {
+            warn!(
+                ?response,
+                "peer lookup response not in outstanding requests, dropping response..."
+            );
+            return cmds;
+        }
+
+        // update peer info
+        for (node_id, name_record) in response.name_records {
```
omegablitz: Do we need to bound the number of signature verifications? Wondering if this is a DOS vector

monad-peer-discovery/src/lib.rs
```diff
@@ -1,2 +1,289 @@
...
+pub struct PeerLookupRequest<ST: CertificateSignatureRecoverable> {
+    lookup_id: u32,
+    target: NodeId<CertificateSignaturePubKey<ST>>,
+}
+
+#[derive(Debug, Clone)]
+pub struct PeerLookupResponse<ST: CertificateSignatureRecoverable> {
+    lookup_id: u32,
+    target: NodeId<CertificateSignaturePubKey<ST>>,
+    name_records: Vec<(NodeId<CertificateSignaturePubKey<ST>>, MonadNameRecord<ST>)>,
```
omegablitz: Doesn't MonadNameRecord already include the ST (and thereby the NodeId)?

monad-peer-discovery/src/discovery.rs
eerkaijun: we will periodically prune nodes that have not been active for some time, `last_seen` will be used for this purpose

monad-peer-discovery/src/lib.rs
eerkaijun: ok i didn't know we don't support ipv6, if that's the case, agree that we should remove this

monad-peer-discovery/src/discovery.rs
eerkaijun: yes, working on the pruning logic in another pr

monad-peer-discovery/src/discovery.rs
eerkaijun: we do not want to process uninitiated response (both for ping and peer lookup request), or requests that are initiated a long time ago and has been pruned, so we want to check if the response corresponds to a request. 

> We aren't verifying that the peer is sending back what we asked for - is that intended?

if the target that we asked for is not available, by design the peer can send back other random name records

monad-peer-discovery/src/discovery.rs
eerkaijun: good point - i think we should limit the number of name records that can be included in a PeerLookupResponse

eerkaijun: added a const of 16 as the max number of records in a peer lookup response. we can decide the optimal value later

monad-peer-discovery/src/lib.rs
eerkaijun: great point - updated

monad-peer-discovery/src/discovery.rs
eerkaijun: removed node_id from PeerLookupResponse, so we need to recover_pubkey first before we can check the sequence number. limited the number of name records a PeerLookupResponse though (16 for now), so I think it's fine

monad-peer-discovery/src/lib.rs
omegablitz: We don't support ipv6, so you can remove this

omegablitz:
Will review again in more detail as subsequent PRs start to go in.

For now, can just remove ipv6 support and merge.

PR #1683, opened 2025-03-24
Warn instead of panic on start up DNS failures

michael-yxchen:
Node updates missing DNS records through periodic DNS refresh

Closes: https://github.com/category-labs/category-internal/issues/1438

### DIFF
--- a/monad-node/src/main.rs
+++ b/monad-node/src/main.rs
@@ -14,7 +14,7 @@ use monad_consensus_state::ConsensusConfig;
 use monad_consensus_types::{metrics::Metrics, signature_collection::SignatureCollection};
 use monad_control_panel::ipc::ControlPanelIpcReceiver;
 use monad_crypto::certificate_signature::{
-    CertificateSignaturePubKey, CertificateSignatureRecoverable,
+    CertificateSignaturePubKey, CertificateSignatureRecoverable, PubKey,
 };
 use monad_eth_block_policy::EthBlockPolicy;
 use monad_eth_block_validator::EthValidator;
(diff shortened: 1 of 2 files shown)


monad-node/src/main.rs
```diff
@@ -123,16 +123,25 @@ async fn run(node_state: NodeState, reload_handle: ReloadHandle) -> Result<(), (
...
+    let mut bootstrap_peers_seen = BTreeSet::new();
+    for peer in node_state.node_config.bootstrap.peers.iter() {
+        if !bootstrap_peers_seen.insert(peer.secp256k1_pubkey) {
+            panic!(
+                "Multiple bootstrap entries for pubkey={}",
+                peer.secp256k1_pubkey
+            );
+        }
+    }
+
```
buytenh: Wrap this inside { } so that bootstrap_peers_seen gets dropped immediately instead of being kept around for the lifetime of the process?

monad-node/src/main.rs
```diff
@@ -508,12 +521,21 @@ where
...
+            return None;
+        }
+    };
+
+    let Some(entry) = resolved.next() else {
+        warn!(?node_id, ?domain, "No DNS record");
+        return None;
+    };
+
+    Some(entry)
```
buytenh: Does it make sense to also map SocketAddrV6s to None here, given that we force our listening address to be a V4 address, and we can therefore not support V6 peers or full nodes?  I.e. only map SocketAddrV4 to Some().

monad-updaters/src/config_loader.rs
```diff
@@ -243,18 +243,22 @@ struct PeersTable<SCT: SignatureCollection> {
...
         let mut peers_table = BTreeMap::new();
         for peer in bootstrap_peers {
             let node_id = NodeId::new(peer.secp256k1_pubkey);
-            let addr = resolved_map
+            let maybe_addr = resolved_map
                 .get(&node_id)
-                .expect("all peers are resolved on startup");
-            peers_table.insert(node_id, (peer.address, Some(*addr)));
+                .expect("all peers are resolved on start up");
+            peers_table.insert(node_id, (peer.address, *maybe_addr));
```
buytenh: nitpick, but the old version was correct: https://www.reddit.com/r/grammar/comments/awe1s6/is_there_difference_between_start_up_startup_and/

monad-updaters/src/config_loader.rs
```diff
@@ -272,11 +276,11 @@ impl<SCT: SignatureCollection> PeersTable<SCT> {
...
                 Ok(Some(addr)) => Some(addr),
                 Ok(None) => {
-                    info!(?node_id, domain=?address, "No DNS record");
+                    warn!(?node_id, domain=?address, "No DNS record");
                     None
                 }
-                Err(e) => {
-                    info!(?node_id, domain=?address, "Failed to resolve: {}", e);
+                Err(err) => {
+                    warn!(?node_id, domain=?address, ?err,"Unable to resolve");
```
buytenh: Missing space between , and "

michael-yxchen:
Fixed all. Thanks for the feedback @buytenh

PR #1685, opened 2025-03-25
Use proposal latency to compute timestamp adjustment

mikety:
Compute estimated timestamp of the current block as  `VoteSent ts +  NetworkLatency + average estimated wait for qc`.

This is the 3rd PR out of the 3 PRs implementing clock adjustment for accurate timestamp estimation. 
the other 2 are:

1st: "Implement periodic pings to calculate network latency" : https://github.com/category-labs/monad-bft/pull/1675
2nd: "Use OS time directly for block timestamp" : https://github.com/category-labs/monad-bft/pull/1777

### DIFF
--- a/monad-blocktimestamp/src/timestamp.rs
+++ b/monad-blocktimestamp/src/timestamp.rs
@@ -4,6 +4,7 @@ use std::{
     time::{Duration, Instant},
 };
 
+// use rust_decimal::{self, Decimal, MathematicalOps, prelude::ToPrimitive};
 use monad_consensus_types::{
     clock::{Clock, TimestampAdjusterConfig},
     quorum_certificate::{TimestampAdjustment, TimestampAdjustmentDirection},
@@ -12,11 +13,13 @@ use monad_consensus_types::{
 };
 use monad_crypto::certificate_signature::PubKey;
 use monad_types::{Epoch, NodeId, PingSequence, Round};
+use sorted_vec::SortedVec;
 use tracing::debug;
 
-use crate::timestamp_adjuster::TimestampAdjuster;
+use crate::{messages::message::PingResponseMessage, timestamp_adjuster::TimestampAdjuster};
 
-const MAX_LATENCY_SAMPLES: usize = 100;
+const MAX_LATENCY_SAMPLES: usize = 11;
+const MAX_PROPOSAL_SAMPLES: usize = 11;
 
 pub const PING_PERIOD_SEC: usize = 30;
 
@@ -28,6 +31,40 @@ pub enum Error {
     OutOfBounds, // timestamp is out of bounds compared to local time
 }
 
+// TODO: make it a _Running_ median
+#[derive(Debug, Clone)]
+struct RunningMedian<T: Ord + Copy> {
+    last_median: Option<T>,
+    period: usize,
+    samples: SortedVec<T>,
+}
+
+// TODO: unify with timestamp_adjuster and compute running median.
+impl<T: Ord + Copy> RunningMedian<T> {
+    pub fn new(period: usize) -> Self {
+        assert!(period % 2 == 1, "median accuracy expects odd period");
+        Self {
+            last_median: None,
+            period,
+            samples: SortedVec::new(),
+        }
+    }
+
+    pub fn add_sample(&mut self, sample: T) {
+        self.samples.insert(sample);
+        if self.samples.len() == self.period {
+            let i = self.samples.len() / 2;
+            self.last_median = Some(self.samples[i]);
+            self.samples.clear();
+        }
+    }
+
+    pub fn get_median(&self) -> Option<T> {
+        self.last_median
+    }
+}
+
+// TODO: move RunningAverage and RunningMedian to the seaparate modules.
 #[derive(Debug, Clone)]
 struct RunningAverage {
     sum: Duration,
@@ -50,6 +87,7 @@ impl RunningAverage {
     }
 
     pub fn add_sample(&mut self, sample: Duration) {
+        // add some simple and inefficient (WIP) cut out fat tails
         if self.samples.len() == self.max_samples {
             self.sum -= self.samples[self.next_index];
             self.samples[self.next_index] = sample;
@@ -69,9 +107,10 @@ impl RunningAverage {
 /// Ping state per validator
 #[derive(Debug, Clone)]
 pub struct ValidatorPingState {
-    sequence: PingSequence,       // sequence number of the last ping sent
-    ping_latency: RunningAverage, // average latency of the last pings
-    last_ping_time: Instant,      // time of the last ping sent
+    sequence: PingSequence,           // sequence number of the last ping sent
+    ping_latency: RunningAverage,     // average latency of the last pings
+    last_ping_time: Instant,          // time of the last ping sent
+    proposal_latency: RunningAverage, // average latency of proposal
 }
 
 impl Default for ValidatorPingState {
@@ -81,6 +120,7 @@ impl Default for ValidatorPingState {
             sequence: PingSequence(0),
             ping_latency: RunningAverage::new(MAX_LATENCY_SAMPLES),
             last_ping_time: now,
+            proposal_latency: RunningAverage::new(MAX_LATENCY_SAMPLES),
         }
     }
 }
@@ -96,8 +136,8 @@ impl ValidatorPingState {
         self.sequence
     }
 
-    pub fn pong_received(&mut self, sequence: PingSequence) -> Option<Duration> {
-        if sequence != self.sequence {
+    pub fn pong_received(&mut self, sequence: u64) -> Option<Duration> {
+        if sequence != self.sequence.0 {
             return None;
         }
         // estimate latency as half the round trip time
@@ -111,7 +151,18 @@ impl ValidatorPingState {
     }
 
     fn update_latency(&mut self, latency: Duration) {
+        debug!(?latency, "add ping latency");
         self.ping_latency.add_sample(latency);
+    }
+    // approximate the time the broadcasted proposal was sent from the time direct ping
+    // was received and the average ping latency
+    fn update_proposal_latency(&mut self, elapsed_since_last_vote: Duration) {
+        self.proposal_latency.add_sample(elapsed_since_last_vote);
+        debug!(?elapsed_since_last_vote, "add proposal latency");
+    }
+
+    pub fn proposal_latency(&self) -> Option<Duration> {
+        self.proposal_latency.get_avg()
     }
 }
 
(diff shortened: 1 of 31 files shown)


monad-blocktimestamp/src/timestamp.rs
```diff
@@ -28,6 +31,40 @@ pub enum Error {
...
+}
+
+// TODO: unify with timestamp_adjuster and compute running median.
+impl<T: Ord + Copy> RunningMedian<T> {
+    pub fn new(period: usize) -> Self {
+        assert!(period % 2 == 1, "median accuracy expects odd period");
+        Self {
+            last_median: None,
+            period,
+            samples: SortedVec::new(),
```
andr-dev, line 49: Does `SortedVec` have a `with_capacity`?

monad-blocktimestamp/src/timestamp.rs
```diff
@@ -58,6 +95,32 @@ impl RunningAverage {
             self.samples.push(sample);
         }
         self.sum += sample;
+        // add some simple and inefficient (WIP) cut out fat tails
```
andr-dev, line 98: I understand the logic behind cutting out fat tails but it seems like a premature optimization of sorts, is this motivated by some behavior we observed in stressnet?

monad-blocktimestamp/src/timestamp.rs
```diff
@@ -69,9 +132,10 @@ impl RunningAverage {
 /// Ping state per validator
 #[derive(Debug, Clone)]
 pub struct ValidatorPingState {
-    sequence: PingSequence,       // sequence number of the last ping sent
-    ping_latency: RunningAverage, // average latency of the last pings
-    last_ping_time: Instant,      // time of the last ping sent
+    sequence: PingSequence,           // sequence number of the last ping sent
+    ping_latency: RunningAverage,     // average latency of the last pings
+    last_ping_time: Instant,          // time of the last ping sent
+    proposal_latency: RunningAverage, // average latency of proposal
```
andr-dev, line 138: We can get rid of the comments using more succinct names for the fields.
```rust
  last_ping_time: Instant,
  last_ping_seq_num: PingSequence,
  avg_ping_latency: RunningAverage,
  avg_proposal_latency: RunningAverage,
```

monad-blocktimestamp/src/timestamp.rs
```diff
@@ -111,8 +176,19 @@ impl ValidatorPingState {
...
     fn update_latency(&mut self, latency: Duration) {
+        debug!(?latency, "add ping latency");
         self.ping_latency.add_sample(latency);
     }
+    // approximate the time the broadcasted proposal was sent from the time direct ping
+    // was received and the average ping latency
+    fn update_proposal_latency(&mut self, elapsed_since_last_vote: Duration) {
+        self.proposal_latency.add_sample(elapsed_since_last_vote);
+        debug!(?elapsed_since_last_vote, "add proposal latency");
+    }
```
andr-dev, line 187: Is this comment describing what the function is doing? It looks like this function is just adding a duration to the average proposal latency.

monad-blocktimestamp/src/timestamp.rs
```diff
@@ -222,23 +301,78 @@ impl<P: PubKey> PingState<P> {
     fn get_latency(&self, node_id: &NodeId<P>) -> Option<Duration> {
         self.validators.get(node_id).and_then(|v| v.avg_latency())
     }
+
+    pub fn update_proposal_latency(&mut self, node_id: &NodeId<P>, latency: Duration) {
+        let validator_state = self.validators.entry(*node_id).or_default();
```
andr-dev, line 306: If `node_id` isn't in `self.validators`, isn't that an indicator that `update_validators` wasn't called or that something else went wrong? Might be better to warn and ignore the latency.

monad-blocktimestamp/src/timestamp.rs
```diff
@@ -222,23 +301,78 @@ impl<P: PubKey> PingState<P> {
...
+    }
 }
 
-#[derive(Debug)]
-struct SentVote {
+#[derive(Debug, Clone)]
+struct WaitState {
+    last_vote_received: Instant,
+    last_vote_round: Round,
+    qc_latency: RunningAverage,
```
andr-dev, line 322: Same as above, `avg_qc_latency` is more descriptive.

monad-blocktimestamp/src/timestamp.rs
```diff
@@ -222,23 +301,78 @@ impl<P: PubKey> PingState<P> {
...
+            qc_latency: RunningAverage::new(max_samples as usize),
+        }
+    }
+
+    pub fn vote_received(&mut self, round: Round) {
+        self.last_vote_received = Instant::now();
+        self.last_vote_round = round;
+    }
+
+    pub fn qc_computed(&mut self, round: Round) {
```
andr-dev, line 340: Super nit: I'm a big fan of the `update_xyz/on_xyz` naming convention or similar because then it makes it very clear that this function should be called every time `xyz` happens, also reads clearer like `on_vote_received` is obviously an "event handler" like function and not like a `is_vote_received` which would tell you if a vote was received for that round but without the prefix both fns would be called `vote_received` so which is which right.

monad-blocktimestamp/src/timestamp.rs
```diff
@@ -222,23 +301,78 @@ impl<P: PubKey> PingState<P> {
...
+        self.qc_latency.get_avg()
+    }
+}
+
+#[derive(Debug, Copy, Clone)]
+struct SentVote<P: PubKey> {
+    node_id: NodeId<P>,
     round: Round,
-    timestamp: Instant,
+    timestamp: Duration,
```
andr-dev, line 359: `timestamp: Duration` is a misnomer, `timestamp` implies a specific point in time while `Duration` implies a length of time.

monad-blocktimestamp/src/timestamp.rs
```diff
@@ -319,41 +464,76 @@ impl<P: PubKey, T: Clock> BlockTimestamp<P, T> {
...
-    ) -> Result<Option<TimestampAdjustment>, Error> {
-        // adjust for estimated latency
-        let latency = self
+    fn compute_clock_adjustment(&self, proposed_block_ts_ns: u128) -> Option<TimestampAdjustment> {
+        // return the delta between expected block time and actual block time for adjustment
+        let vote = self.last_sent_vote.unwrap();
+        let proposal = self.last_received_proposal.unwrap();
+
+        assert!(vote.node_id == proposal.node_id);
+        assert!(vote.round + Round(1) == proposal.round);
```
andr-dev, line 473: Can we do an `if Some(...)` / `if A != B` and warn instead of unwrap/assert here?

monad-blocktimestamp/src/timestamp.rs
```diff
@@ -319,41 +464,76 @@ impl<P: PubKey, T: Clock> BlockTimestamp<P, T> {
...
-            .get_latency(author)
+            .get_latency(&vote.node_id)
             .unwrap_or(Duration::from_nanos(
-                self.default_latency_estimate_ns.try_into().unwrap(),
-            ));
-
-        let now = self.get_current_time();
-        // TODO: add proposal estimated time
-        let expected_block_ts = now.saturating_sub(latency.as_nanos());
+                self.default_latency_estimate_ns as u64,
```
andr-dev, line 481: Random thought: If `BlockTimestamp` is generic on `CCT` and `CRT`, we can use `ChainParams.vote_pace / 2` as the default latency estimate.

monad-blocktimestamp/src/timestamp.rs
```diff
@@ -393,22 +584,82 @@ impl<P: PubKey, T: Clock> BlockTimestamp<P, T> {
...
+        proposed_block_ts_ns: u128,
+        author: &NodeId<P>,
+    ) {
+        debug!(
+            ?proposal_round,
+            ?author,
+            ?proposed_block_ts_ns,
+            "Proposal received",
+        );
+        if let Some(vote) = self.last_sent_vote {
```
andr-dev, line 599: Nit:
```rust
let Some(vote) = self.last_sent_vote else {
  return;
};

...
```

monad-blocktimestamp/src/timestamp.rs
```diff
@@ -393,22 +584,82 @@ impl<P: PubKey, T: Clock> BlockTimestamp<P, T> {
...
         }
     }
+
+    pub fn update_create_proposal_ns(&mut self, _round: Round, elapsed_ns: u128) {
+        self.create_proposal_time_ns.add_sample(elapsed_ns);
+    }
+
+    pub fn vote_received(&mut self, author: &NodeId<P>, round: Round) {
+        match self.avg_wait_for_qc_self.entry(*author) {
+            std::collections::btree_map::Entry::Vacant(entry) => {
```
andr-dev, line 635: Suggestion: A nice way I've found of making this cleaner is to do
```rust
use std::collections::btree_map::{Entry as BTreeMapEntry}

...

match ... {
  BTreeMapEntry::Vacant(entry) => { ... },
  ...
}
```

monad-blocktimestamp/src/timestamp.rs
```diff
@@ -393,22 +584,82 @@ impl<P: PubKey, T: Clock> BlockTimestamp<P, T> {
...
+        self.avg_wait_for_qc_self
+            .iter_mut()
+            .for_each(|(k, v)| v.qc_computed(round));
+    }
+
+    pub fn get_vote_wait(&self, sender: &NodeId<P>) -> Option<Duration> {
+        if let Some(wait_state) = self.avg_wait_for_qc_self.get(sender) {
+            return wait_state.get_avg();
+        }
+        None
```
andr-dev, line 653:
```rust
self.avg_wait_for_qc_self.get(sender)
  .and_then(|wait_state| wait_state.get_avg())
```

monad-blocktimestamp/src/timestamp_adjuster.rs
```diff
@@ -17,33 +20,42 @@ pub struct TimestampAdjuster {
...
+    pub fn new(max_delta_ns: u128, adjustment_period: usize, max_drift: Option<i64>) -> Self {
         assert!(max_delta_ns < i128::MAX as u128);
         assert!(
             adjustment_period % 2 == 1,
             "median accuracy expects odd period"
         );
+        let mut init_adjustment = 0;
+        let mut rng = rand::thread_rng();
+        if let Some(max_drift) = max_drift {
+            init_adjustment = rng.gen_range(max_drift.neg()..max_drift);
```
andr-dev, line 32: Why do we want to randomize the initial adjustment? Likely missing context here, sorry

monad-blocktimestamp/src/timestamp_adjuster.rs
```diff
@@ -17,33 +20,42 @@ pub struct TimestampAdjuster {
...
     }
 
     pub fn add_delta(&mut self, delta: i64) {
-        trace!(delta, "add delta");
+        debug!(delta, "add delta");
         self.deltas.insert(delta);
         if self.deltas.len() == self.adjustment_period {
             let i = self.deltas.len() / 2;
-            self.adjustment += self.deltas[i];
+            let adjustment = self.deltas[i];
```
andr-dev, line 48: We changed this from += to =, is this intentional?

monad-consensus-state/src/lib.rs
```diff
@@ -580,16 +579,21 @@ where
...
-                .is_ok()
-                && block.get_round() > original_round
+                .is_err()
             {
-                self.block_timestamp.proposal_received(
-                    block.get_round(),
-                    block.get_timestamp(),
-                    &author,
-                );
+                warn!("Error");
```
andr-dev, line 584: Let's make this warn a bit more descriptive.

PR #1686, opened 2025-03-25
Integrate expensive eth call

eerkaijun:
Add some expensive eth calls to generator


monad-eth-testutil/examples/rpc-request-generator/bytecode.txt
ariqchowdhury, line 1: can we comment/document what the contract does

monad-eth-testutil/examples/rpc-request-generator/main.rs
```diff
@@ -747,6 +795,84 @@ impl RpcRequestGenerator {
...
+    info!(?nonce, "ending nonce");
+
+    // periodically sends expensive eth call
+    loop {
+        let call_request = TransactionRequest {
+            to: Some(TxKind::Call(contract.addr)),
+            input: Some(
+                Bytes::from_str(
+                    "0xb48f9647000000000000000000000000000000000000000000000000000000000000c350",
+                )
```
ariqchowdhury: where does this come from?

monad-eth-testutil/examples/rpc-request-generator/main.rs
```diff
@@ -747,6 +795,84 @@ impl RpcRequestGenerator {
...
+        .expect("retrieve nonce failed");
+    let nonce = nonce.to();
+    info!(?nonce, "starting nonce");
+
+    let contract = Contract::deploy_tx(&client, nonce, &private_key, chain_id)
+        .await
+        .expect("contract deployment failed");
+    info!(?contract, "start initializing state for expensive eth call");
+
+    // write 50k unique storage slots
```
ariqchowdhury, line 821: its not clear to me what this is doing? its doing 250*250 writes so should be 62500?
And if its writing the value i, it only has to be done once, not on every expensive_eth_call? Or is it a different private key on every call and it later reads something specific to that user?

monad-eth-testutil/examples/rpc-request-generator/bytecode.txt
eerkaijun: the source code of the contract is https://github.com/category-labs/monad-integration/blob/kai/expensive-eth-call/tests/test_contract_interaction/example.sol

I'll get that merge in and reference the commit

monad-eth-testutil/examples/rpc-request-generator/main.rs
eerkaijun: each transaction writes 200 unique storage slots, so we send 250 transaction to build up the state for 50k unique storage slots. it's split up into multiple transactions because it will exceed the block gas limit (so the transactions need to be in different blocks). this setup process is only done once when we first start the generator, subsequence eth call reads will just be calling this contract.

I'll add a comment explaining this to make it clearer

monad-eth-testutil/examples/rpc-request-generator/main.rs
eerkaijun: first four bytes is the function selector and the subsequence 32 bytes is the function argument. 0xc350 corresponds to 50k (reading 50k storage slots). I'll add a comment for this too

eerkaijun: updated it to be clearer

PR #1689, opened 2025-03-27
serialize mongodb tests

mikety:
mongodb containers use the same port for the DB, so tokio tests need to be serialized to avoid race conditions.


Cargo.toml
```diff
@@ -165,6 +165,7 @@ serde = "1.0"
 serde_cbor = "0.11"
 serde_json = "1.0"
 sha2 = "0.10"
+serial_test = "3.2.0"
```
ariqchowdhury: nit: alphabetical order

PR #1691, opened 2025-03-27
Use helper fn to recover txs in txpool tests

andr-dev:


mikety:
LGTM with nit:  https://github.com/category-labs/monad-bft/blob/eb46137ee4eae5006b30b37c78bc041bd81dc567/monad-eth-block-validator/src/lib.rs#L94 
can also use it. if converted to` Result`

andr-dev:
> LGTM with nit:
> 
> https://github.com/category-labs/monad-bft/blob/eb46137ee4eae5006b30b37c78bc041bd81dc567/monad-eth-block-validator/src/lib.rs#L94
> 
> 
> can also use it. if converted to` Result`

The purpose of the function is to only be used in tests, hence why it is in monad-eth-testutil and why it unwraps. Additionally, it is probably best to explicitly recover the tx in the block validator to remove any "layers of abstraction" from validating code.

PR #1692, opened 2025-03-27
Peer discovery swarm

michael-yxchen:
Simulation environment for peer discovery algorithm. It can simulate many rounds of peer discovery message exchange

### DIFF
--- a/monad-peer-disc-swarm/src/builder.rs
+++ b/monad-peer-disc-swarm/src/builder.rs
@@ -1,0 +1,42 @@
+use std::{collections::BTreeMap, time::Duration};
+
+use monad_peer_discovery::algo::PeerDiscoveryBuilder;
+use rand::{Rng, SeedableRng};
+use rand_chacha::ChaCha8Rng;
+
+use crate::{NodeBuilder, Nodes, PeerDiscSwarmRelation};
+
+pub struct PeerDiscSwarmBuilder<S, B>
+where
+    S: PeerDiscSwarmRelation,
+{
+    pub builders: Vec<NodeBuilder<S, B>>,
+    pub seed: u64,
+}
+
+impl<S, B> PeerDiscSwarmBuilder<S, B>
+where
+    S: PeerDiscSwarmRelation,
+    B: PeerDiscoveryBuilder<PeerDiscoveryAlgoType = S::PeerDiscoveryAlgoType>,
+{
+    pub fn build(self) -> Nodes<S> {
+        let _init_span_entered =
+            tracing::trace_span!("init", tick = format!("{:?}", Duration::ZERO)).entered();
+        let mut states = BTreeMap::new();
+        for builder in self.builders {
+            let id = builder.id;
+            let _node_span_entered = tracing::trace_span!("node", id = format!("{}", id)).entered();
+            let node = builder.build();
+            states.insert(id, node);
+        }
+
+        let mut rng = ChaCha8Rng::seed_from_u64(self.seed);
+
+        Nodes {
+            states,
+            tick: Default::default(),
+            rng: ChaCha8Rng::seed_from_u64(rng.r#gen()),
+            current_seed: rng.r#gen(),
+        }
+    }
+}
--- a/monad-peer-disc-swarm/src/driver.rs
+++ b/monad-peer-disc-swarm/src/driver.rs
@@ -1,0 +1,166 @@
+use std::{cmp::Reverse, collections::BTreeMap, time::Duration};
+
+use monad_crypto::certificate_signature::PubKey;
+use monad_executor_glue::RouterCommand;
+use monad_peer_discovery::algo::{
+    PeerDiscoveryAlgo, PeerDiscoveryBuilder, PeerDiscoveryCommand, PeerDiscoveryEvent,
+    PeerDiscoveryMessage, PeerDiscoveryTimerCommand,
+};
+use monad_types::{NodeId, RouterTarget};
+use priority_queue::PriorityQueue;
+
+#[derive(Debug)]
+struct TimerEvent<E> {
+    expire_at: Duration,
+    on_timeout: E,
+}
+
+struct MockDiscTimer<E, P: PubKey> {
+    timers: BTreeMap<NodeId<P>, TimerEvent<E>>,
+    priority_queue: PriorityQueue<NodeId<P>, Reverse<Duration>>,
+
+    tick: Duration,
+}
+
+impl<E, P: PubKey> Default for MockDiscTimer<E, P> {
+    fn default() -> Self {
+        Self {
+            timers: Default::default(),
+            priority_queue: Default::default(),
+            tick: Default::default(),
+        }
+    }
+}
+
+impl<E, P: PubKey> MockDiscTimer<E, P> {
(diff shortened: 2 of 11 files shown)


monad-peer-disc-swarm/src/builder.rs
fabiotdf, line 1: This is just for test, right?

monad-peer-disc-swarm/src/driver.rs
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+
+    // returns the first tick in priority. if timer was reset before step_until
+    // is called, step_until is a no-op
+    fn peek_tick(&self) -> Option<Duration> {
+        self.priority_queue.peek().map(|(_, rev_d)| rev_d.0)
+    }
+
+    fn step_until(&mut self, until: Duration) -> Option<E> {
```
fabiotdf, line 65: Is the underlying value for Duration, the time elapsed since Unix epoch?
In c++ std::chrono, you can't add two time_points but you can subtract them, yielding a duration. I wonder if  a similar representation for time point exists in Rust?

monad-peer-disc-swarm/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+    CertificateSignaturePubKey<<S as PeerDiscSwarmRelation>::SignatureType>;
+
+pub struct Swarm<S>
+where
+    S: PeerDiscSwarmRelation,
+{
+    nodes: BTreeMap<NodeId<SwarmPubKeyType<S>>, Node<S>>,
+}
+
+pub trait PeerDiscSwarmRelation {
```
fabiotdf, line 35: What does "Relation" refer to here? I couldn't understand this

michael-yxchen: "Relation" probably isn't a good word here. The intention was to group all the types associated with a swarm together. Otherwise `Nodes<>/Node<>` will take in a long list of generic types. Feel free to suggest alternatives

monad-peer-disc-swarm/src/driver.rs
michael-yxchen: The `Duration` here is virtual time. It starts from zero and is incremented for the simulation 

rust has `std::time::Instant`. Agree that semantically Instant makes more sense than duration. Let me see if it supports all operations that we need here

monad-peer-disc-swarm/src/builder.rs
michael-yxchen: yeah, the entire crate is test infra

monad-peer-disc-swarm/src/driver.rs
michael-yxchen: `std::time::Instant` is missing a zero instant constructor, which makes it confusing in simulation (starting from a random instant). Our other related test infra (e.g. transformer also operates with `Duration` timestamps) so I'm keeping this

monad-peer-disc-swarm/src/driver.rs
```diff
@@ -0,0 +1,163 @@
...
+
+        self.filter_and_exec(cmds)
+    }
+
+    // selectively execute the timer commands. Returns router commands in the
+    // original order
+    fn filter_and_exec(
+        &mut self,
+        cmds: Vec<PeerDiscoveryCommand<P>>,
+    ) -> Vec<RouterCommand<P, PeerDiscoveryMessage<P>>> {
```
andr-dev, line 136: What filtering is this function doing? Is the `cmds` argument the already filtered commands? If so, we shouldn't call this function `filter_*` as it's misleading.

monad-peer-disc-swarm/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+
+        let mut rng = ChaCha8Rng::seed_from_u64(seed);
+
+        Node {
+            id,
+            peer_disc_driver,
+            executor,
+            pending_inbound_messages: Default::default(),
+            rng: ChaCha8Rng::seed_from_u64(rng.r#gen()),
+            current_seed: rng.r#gen(),
```
andr-dev: A little confused by the rng initialization here. Won't the `rng.r#gen()` call being passed to `seed_from_u64` produce a different value than the one we're setting to `current_seed`. Additionally, why are we seeding the rng to the first rng produces by an rng seeded by our desired `seed` value? This seems like an unnecessary indirection.

If we just want to store some pseudo-random value which we can update periodically, we should call that something along the lines of `last_rng`.
```rust
let mut rng = ChaCha8Rng::seed_from_u64(seed);
let last_rng = rng.r#gen();

Node {
    ...
    rng,
    last_rng,
}
```

monad-peer-discovery/src/mock.rs
```diff
@@ -0,0 +1,162 @@
...
+                        last_ping: None,
+                        alive: false,
+                    })
+                })
+                .collect(),
+            ping_period: self.ping_period,
+            metrics: Default::default(),
+        };
+
+        let mut cmds = Vec::new();
```
andr-dev:
```rust
let mut cmds = peers.into_iter().flat_map(|peer| state.handle_send_ping(peer)).collect();
```

monad-peer-discovery/src/mock.rs
```diff
@@ -0,0 +1,162 @@
...
+        debug!(?target, "handle send ping");
+        let mut cmds = Vec::new();
+
+        if !self.peer_state.contains_key(&target) {
+            return cmds;
+        }
+        cmds.extend(self.reset_timer(target));
+        let peer_state = self.peer_state.get_mut(&target).expect("contains peer");
+
+        let ping_id = peer_state.last_ping.map_or(1, |ping| ping.id + 1);
```
andr-dev: Shouldn't this be `map_or(0, ...)` so the first `ping_id` is 0?

monad-peer-disc-swarm/src/driver.rs
michael-yxchen: the function collects one (RouterCommand) of two variants of PeerDiscoveryCommand, and execute timer command inline. any other name suggestions?

monad-peer-disc-swarm/src/lib.rs
michael-yxchen: agreed. Will update

monad-peer-discovery/src/mock.rs
michael-yxchen: no preference either way. It's just arbitrary. I'll change it to 0

monad-peer-discovery/src/algo.rs
```diff
@@ -0,0 +1,97 @@
...
+        duration: Duration,
+        on_timeout: E,
+    },
+    ScheduleReset {
+        node_id: NodeId<P>,
+    },
+}
+
+pub enum PeerDiscoveryCommand<P: PubKey> {
+    RouterCommand(RouterCommand<P, PeerDiscoveryMessage<P>>),
```
omegablitz: nit: should this be `RouterCommand(PeerDiscoveryMessage<P>)` instead? Is there a use-case for this doing anything besides sending messages?

If this is needed as a RouterCommand anywhere (which I'm not sure of), that translation can happen further up the stack.

michael-yxchen: yeah only for sending messages. will fix

PR #1693, opened 2025-03-27
Support multiple threads and fibers for eth-call

ariqchowdhury:
configurable number of threads and fibers from rpc cli. node lru size also configurable from rpc cli.

bumps execution to get the eth-call changes, but does not include the PriorityPool thread sleep change in this PR.

### DIFF
--- a/monad-rpc/src/cli.rs
+++ b/monad-rpc/src/cli.rs
@@ -54,8 +54,20 @@ pub struct Cli {
     pub eth_get_logs_max_block_range: u64,
 
     /// Set the max concurrent requests for eth_call and eth_estimateGas
-    #[arg(long, default_value_t = 20)]
+    #[arg(long, default_value_t = 1000)]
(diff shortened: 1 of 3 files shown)


monad-rpc/src/cli.rs
```diff
@@ -54,9 +54,18 @@ pub struct Cli {
...
     pub eth_call_max_concurrent_requests: u32,
 
+    #[arg(long, default_value_t = 2)]
+    pub eth_call_executor_threads: u32,
+
+    #[arg(long, default_value_t = 64)]
+    pub eth_call_executor_fibers: u32,
+
+    #[arg(long, default_value_t = 102400)]
+    pub eth_call_executor_node_lru_size: u32,
```
eerkaijun: can we add descriptions to each of these CLI arguments so that it will show up in the `--help` option

omegablitz:
Can you bump the execution submodule again to include the bugfixes

PR #1694, opened 2025-03-28
Archive: dev readme

JoeHowarth:


monad-archive/README.md
```diff
@@ -0,0 +1,109 @@
+# Monad Archive Developer Docs
+
+Resources:
+- Ops references to self-hosted archive (mongo): https://github.com/monad-client-docs/validator-docs/blob/main/docs/full_node/self-hosted-archive/self-hosted-archive.md
+- Ops Runbook: https://www.notion.so/Archive-Ops-Runbook-17c75b0ba84080319873ffaac7fdfff1
+- Archive Writer Dashboard: https://grafana.devcore4.com/d/dec2qtthsf3lsb/archiving3a-writer-overview?orgId=1&from=now-15m&to=now&timezone=browser&var-net=testnet&refresh=5s
```
eerkaijun, line 6: should we exclude private links in repo since this repo will be open sourced eventually

JoeHowarth: I agree it should be removed before open sourcing, but in the meantime since knowledge of this system is still quite silo'd, I think lowering the friction in the case of an outage is a higher priority

PR #1696, opened 2025-03-31
set SO_REUSEADDR by passing option to monoio TcpListener

dshulyak:
closes: https://github.com/category-labs/category-internal/issues/1307

with SO_REUSEADDR on tcp socket application will be able to quickly restart, without waiting
for os to tear down previously opened socket.


dshulyak:
> This bug seems to apply to both dataplane v1 and dataplane v2. The remediation for dataplane v2 will be slightly different than for v1.

@buytenh the issue mentions dataplane v1/v2. but this dataplane seems to be the only one which is used in raptorcast router. am i missing something?

PR #1697, opened 2025-04-01
Log statesync client upsert application time

omegablitz:


zander-xyz:
It would also be helpful to log at the end of statesync
- time overall spent statesyncing
- total time spent applying upserts
- anything else?

omegablitz:
Aashish Welling

one confounder is that the upsert application partially, but not completely, happens in parallel with the actual downloading of data
it’s in parallel until target needs to be reset, after which downloading is halted until the upsert application “catches up”


Alexander Lee

so maybe we need multiple timers then
basically what i’m advocating for is one log at the end of statesync which logs the times of anything relevant
total time spent waiting for a response for each prefix
total time spent downloading
total time spent applying
blocksyncing
etc
and having it on one line would make it super greppable

omegablitz:
Actually, I'll re-enable this for merging and make the statesync timing summary a separate issue

PR #1699, opened 2025-04-01
New execution IPC format

omegablitz:
TODOs:
- [x] update flexnet
- [x] test this works with execution
- [x] test ledger-tail changes
- [x] make associated archiver changes
- [ ] update clear-old-artifacts, reset-workspace.sh (category-infra)

### DIFF
--- a/monad-archive/src/workers/bft_block_archiver.rs
+++ b/monad-archive/src/workers/bft_block_archiver.rs
@@ -1,25 +1,45 @@
 use crate::prelude::*;
 
 const BFT_BLOCK_PREFIX: &str = "bft_block/";
+const BFT_BLOCK_HEADER_EXTENSION: &str = ".header";
+const BFT_BLOCK_BODY_EXTENSION: &str = ".body";
+
+const BFT_BLOCK_HEADER_FILE_PATH: &str = "headers/";
+const BFT_BLOCK_BODY_FILE_PATH: &str = "bodies/";
+
+enum BFTFilePath {
+    Header(PathBuf),
+    Body(PathBuf),
+}
+
+enum BFTFileName {
+    Header(OsString),
+    Body(OsString),
+}
 
 /// Worker that archives BFT consensus block files to durable storage.
 /// Monitors a directory for new block files and uploads them.
 ///
 /// # Arguments
 /// * `store` - Storage backend to write blocks to
-/// * `block_path` - Path to directory containing BFT block files
+/// * `ledger_path` - Path to directory containing BFT block files
 /// * `poll_frequency` - How often to check for new files
 /// * `metrics` - Metrics collection interface
 pub async fn bft_block_archive_worker(
     store: KVStoreErased,
-    block_path: PathBuf,
+    ledger_path: PathBuf,
     poll_frequency: Duration,
     metrics: Metrics,
 ) -> Result<()> {
     info!("Fetching set of already uploaded bft blocks...");
-    let mut uploaded = initialize_uploaded_set(&store)
+    let (mut uploaded_headers, mut uploaded_bodies) = initialize_uploaded_sets(&store)
         .await
         .wrap_err("Failed to fetch set of already uploaded bft blocks")?;
+
+    let mut headers_path = PathBuf::from(&ledger_path);
+    headers_path.push(BFT_BLOCK_HEADER_FILE_PATH);
+    let mut bodies_path = PathBuf::from(&ledger_path);
+    bodies_path.push(BFT_BLOCK_BODY_FILE_PATH);
 
     let mut interval = tokio::time::interval(poll_frequency);
     interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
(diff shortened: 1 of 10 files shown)


monad-archive/src/workers/bft_block_archiver.rs
```diff
@@ -12,61 +27,105 @@ const BFT_BLOCK_PREFIX: &str = "bft_block/";
 /// * `metrics` - Metrics collection interface
```
JoeHowarth, line 27: nit: update comment

monad-archive/src/workers/bft_block_archiver.rs
```diff
@@ -1,6 +1,21 @@
 use crate::prelude::*;
 
 const BFT_BLOCK_PREFIX: &str = "bft_block/";
+const BFT_BLOCK_HEADER_EXTENSION: &str = ".header";
+const BFT_BLOCK_BODY_EXTENSION: &str = ".body";
+
+const BFT_BLOCK_HEADER_PATH: &str = "headers/";
```
JoeHowarth: It's confusing whether this refers to file path or s3 path. Can we go through this file and use file_path instead of path where appropriate?

monad-archive/src/workers/bft_block_archiver.rs
```diff
@@ -1,6 +1,21 @@
 use crate::prelude::*;
 
 const BFT_BLOCK_PREFIX: &str = "bft_block/";
+const BFT_BLOCK_HEADER_EXTENSION: &str = ".header";
```
JoeHowarth, line 4: Existing header and body objects already have .body or .header extension. Did the filenames on disk change to not have these extensions anymore? Otherwise we'll get double extensions

ex:
000005790bdbdc0daf60057a8603af20b2582a29185fd760354f84931742c880.header 00000615c7ff3c184984ffa8f2e5dc0a0cabd3c9aeba2f8e6edeeead3c693045.body

monad-archive/src/workers/bft_block_archiver.rs
```diff
@@ -75,14 +134,37 @@ async fn parallel_upload_iter(
...
 }
 
-async fn upload(store: KVStoreErased, path: PathBuf) -> Option<futures::future::Ready<OsString>> {
-    let fname_os = path.file_name()?.to_os_string();
-    let fname = fname_os.to_str()?;
+async fn upload(
+    store: KVStoreErased,
+    bft_file_path: BFTFilePath,
+) -> Option<futures::future::Ready<BFTFileName>> {
+    let (path, extension, bft_file_name) = match bft_file_path {
```
JoeHowarth: This section feels awkward, but I'm not sure of a better way to write it. Can we bind fname in the first match or does that cause borrow errors since we're moving fname_os too?

monad-archive/src/workers/bft_block_archiver.rs
bamrith: That's right, the filenames don't have these extensions anymore since they are in different directories. I added these extensions to maintain the current format.

monad-archive/src/workers/bft_block_archiver.rs
bamrith: It does cause borrow errors. Agree that it's really awkward, open to any suggestions.

omegablitz:
Should we remove the `--wal-path` and `--wal-checkpoint-freq-secs` archiver args?
We could consider archiving proposed_head/finalized_head as well (but not sure if this is necessary)

finalized_head would probably be useful so that if we wanted to start replaying blocks at an arbitrary point in history, we could look up the nearest finalized_head and start rewinding from there

monad-block-persist/src/lib.rs
```diff
@@ -61,30 +72,59 @@ where
...
+                Err(e) if e.kind() == ErrorKind::AlreadyExists => (),
+                Err(e) => panic!("{}", e),
+            }
+            bodies_path
+        };
+
+        let proposed_head_path = {
+            let mut proposed_head_path = PathBuf::from(&headers_path);
+            proposed_head_path.push(BLOCKDB_PROPOSED_HEAD_PATH);
+            proposed_head_path
```
andr-dev: I think all these pathbuf constructions can be shortened to
```rust
let proposed_head_path = PathBuf::from(&headers_path).join(BLOCKDB_PROPOSED_HEAD_PATH);
```
It's been a while since I've looked over what exactly the`PathBuf` API does so I can't confirm this will produce the same path although my vague memory and the rust documentation seems to suggest so (https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.push).

> Consider using [Path::join](https://doc.rust-lang.org/std/path/struct.Path.html#method.join) if you need a new PathBuf instead of using this function on a cloned PathBuf.

Note: Slight perf hit if the `join(&self) -> PathBuf` doesn't get optimized away but likely negligible if that's even the case.

monad-block-persist/src/lib.rs
```diff
@@ -61,30 +72,59 @@ where
     EPT: ExecutionProtocol,
 {
     pub fn new(ledger_path: PathBuf) -> Self {
+        let headers_path = {
+            let mut headers_path = PathBuf::from(&ledger_path);
+            headers_path.push(BLOCKDB_HEADERS_PATH);
+            match std::fs::create_dir(&headers_path) {
```
andr-dev, line 77: By using `create_dir_all`, you can avoid needing to catch the `ErrorKind::AlreadyExists` branch and just cleanly `.unwrap()`, downside is we might not want to create parent directories if they don't exist. Not sure which is better :shrug:.

PR #1700, opened 2025-04-02
Update rpc to asynchronously wait for txpool events

andr-dev:
During the public testnet frenzy, we made the `eth_sendTransaction/eth_sendRawTransaction` endpoints poll the tx status returned from the txpool. This solution was meant to be temporary as it's far from ideal to have workers continuously polling when we can have a "notify" system. This changeset addresses exactly that by introducing a oneshot channel into the txpool bridge state so that `sendTransaction` calls will
1. Reply as soon as a tx status becomes available
2. Not experience as many "txpool not responding" error messages as we can set a longer timeout (1 second) since there is a minimal performance hit from waiting on a oneshot channel

### DIFF
--- a/monad-rpc/src/eth_txn_handlers.rs
+++ b/monad-rpc/src/eth_txn_handlers.rs
@@ -17,7 +17,7 @@ use monad_triedb_utils::triedb_env::{
 };
 use monad_types::SeqNum;
 use serde::{Deserialize, Serialize};
-use tracing::{debug, trace, warn};
+use tracing::{debug, error, trace, warn};
 
 use crate::{
     block_handlers::{block_receipts, get_block_key_from_tag},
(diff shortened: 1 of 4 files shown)


monad-rpc/src/eth_txn_handlers.rs
```diff
@@ -444,35 +444,39 @@ pub async fn monad_eth_sendRawTransaction<T: Triedb>(
             let hash = *tx.tx_hash();
             debug!(name = "sendRawTransaction", txn_hash = ?hash);
 
-            if let Err(err) = txpool_bridge_client.try_send(tx) {
+            let (tx_status_tx, tx_status_rx) = tokio::sync::oneshot::channel::<TxStatus>();
```
ariqchowdhury: nit: tx_status_send, tx_status_recv

monad-rpc/src/eth_txn_handlers.rs
```diff
@@ -444,35 +444,39 @@ pub async fn monad_eth_sendRawTransaction<T: Triedb>(
...
                     }
-                    Some(TxStatus::Dropped { reason }) => {
-                        Err(JsonRpcError::custom(reason.as_user_string()))
+                    TxStatus::Pending | TxStatus::Tracked | TxStatus::Committed => {
+                        return Ok(hash.to_string())
                     }
-                    Some(TxStatus::Pending | TxStatus::Tracked | TxStatus::Committed) => {
-                        Ok(hash.to_string())
+                    TxStatus::Unknown => {
+                        error!("txpool bridge sent unknown status");
```
ariqchowdhury, line 468: what leads to Unknown being sent from the bridge?

monad-rpc/src/txpool/state.rs
```diff
@@ -15,17 +15,18 @@ use super::TxStatus;
 const TX_EVICT_DURATION_SECONDS: u64 = 15 * 60;
 
 pub(super) type EthTxPoolBridgeEvictionQueue = VecDeque<(Instant, TxHash)>;
+pub(super) type TxStatusSender = tokio::sync::oneshot::Sender<TxStatus>;
 
 #[derive(Clone)]
 pub struct EthTxPoolBridgeStateView {
-    status: Arc<DashMap<TxHash, TxStatus>>,
+    status: Arc<DashMap<TxHash, (TxStatus, Option<TxStatusSender>)>>,
```
ariqchowdhury, line 22: what's the situation where the status sender would be None?

monad-rpc/src/eth_txn_handlers.rs
andr-dev: Nothing, it should never send Unknown, so if it does there's 100% a bug hence the error.

monad-rpc/src/txpool/state.rs
andr-dev: A couple reasons for this:
1. Since `TxStatusSender` is a oneshot channel, it needs to be consumed leaving a None
2. If we get new status updates after the first (like committed), then we don't need to send that over a channel because eth_sendTransaction/eth_sendRawTransaction has already responded
3. If BFT restarts and we get a snapshot

PR #1702, opened 2025-04-03
Full-node Raptorcast pt 1: Secondary instance

fabiotdf:
This is the first part of the full-node raptorcast implementation. It contains just the secondary raptorcast instance with group management logic and some unit tests for it. This does not change the primary instance nor the normal routing of commands (left for a future PR when this is already in master).


monad-discovery/src/message.rs
```diff
@@ -95,9 +96,232 @@ impl<PT: PubKey> From<&DiscoveryMessage<PT>> for ProtoDiscoveryMessage {
     }
 }
 
+//==============================================================================
```
andr-dev: A rule I like to follow is that if you feel the need to add comment separators then you probably want to split things out into more files. We don't use comment separators elsewhere in the codebase, so I would suggest following that convention.

monad-discovery/src/message.rs
```diff
@@ -95,9 +96,232 @@ impl<PT: PubKey> From<&DiscoveryMessage<PT>> for ProtoDiscoveryMessage {
...
+                    "ProtoMonadNameRecord.end_round".to_owned(),
+                ))?
+                .try_into()?,
+        })
+    }
+}
+
+//------------------------------------------------------------------------------
+// ProtoPrepareGroupResponse
+//------------------------------------------------------------------------------
```
andr-dev: Ditto for these comment separators. Let's move the From/TryFrom impls below to a separate moduled called `convert`. You can take a look at `monad-consensus-types` for an example of the convention we're following.

monad-raptorcast/Cargo.toml
```diff
@@ -28,13 +28,15 @@ itertools = { workspace = true }
...
 tracing = { workspace = true }
+iset = { workspace = true }
 
 [dev-dependencies]
 clap = { workspace = true, features = ["derive"] }
 criterion = { workspace = true }
 futures-util = { workspace = true }
 tokio = { workspace = true, features = ["macros", "rt-multi-thread", "sync"] }
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
+monad-testutil = { workspace = true }
```
andr-dev: All `monad-*` dependencies should go at the top as a separate group like they are for the `[dependencies]` group above.

monad-raptorcast/src/config.rs
```diff
@@ -0,0 +1,74 @@
...
+{
+    //pub key: ST::KeyPairType, // note: CertificateKeyPair can't be copied
+
+    // Note1: CertificateKeyPair can't be copied, hence the share.
+    // Note2: make_monad_executor() from testground doesn't let us use
+    // Rc<RefCell<ST::KeyPairType>> for shared_key and rustc complains:
+    // error[E0277]: `Rc<RefCell<<ST as CertificateSignature>::KeyPairType>>` cannot be sent between threads safely
+    // Note3: monad-node also complains when using Rc<RefCell>:
+    //error[E0277]: `Rc<RefCell<KeyPair>>` cannot be sent between threads safely
+    pub shared_key: Arc<Mutex<ST::KeyPairType>>,
```
andr-dev: Where is `shared_key` being used? I did a quick ctrl+f and didn't find anything. I'm also not sure about the context, but based on the name I'm guessing that it's not modified after initialization in which case we should probably just `Arc<ST::KeyPairType>`.

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2281 @@
...
+    max_invite_wait: TimePoint, // how long to wait for a node to respond to an invite before we try some other ones.
+    deadline_round_dist: Round, // group formation deadlined in terms of how close to the group's start_round we are
+    init_empty_round_span: Round, // like round_span but for the case when we need an empty, locked group right now
+}
+
+impl Default for GroupSchedulingConfig {
+    fn default() -> GroupSchedulingConfig {
+        GroupSchedulingConfig {
+            max_group_size: 50,
+            round_span: Round(120), // approx 1 minute
```
andr-dev: This only holds when the `vote_pace` is 500ms. We might want to pass in the chain config and update based on those values, unclear how important this constant is for network-wide "correctness".

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2281 @@
...
+                // while we allow invites for a future group to complete.
+                if self.curr_group.start_round > Round(0){
+                    tracing::warn!(
+                        "No group for RaptorcastSecondary round {:?}", round);
+                }
+            },
+        }
+        // Fallback group for error cases.
+        self.curr_group = GroupAsPublisher::new_empty_locked(
+            round, self.scheduling_cfg.init_empty_round_span);
```
andr-dev: I don't think this is how rustfmt formats things, have you run `cargo +nightly fmt`?

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2281 @@
...
+            full_nodes_candidates: FullNodes::new(Vec::new()),
+            num_invites_sent: 0,
+            num_invites_rejected: 0,
+            start_round,
+            end_round: start_round + span,
+            next_invite_tp: TimePoint::MAX,
+        }
+    }
+
+    pub fn is_locked_in(&self) -> bool
```
andr-dev: I'm not locked in, let's rename this to `is_locked`.

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2281 @@
...
+        let next_invitees = FullNodes::new(
+            self
+            .full_nodes_candidates
+            .list
+            .iter()
+            .skip(self.num_invites_sent)
+            .take(num_missing)
+            .cloned()
+            .collect::<Vec<_>>()
+            .to_vec());
```
andr-dev: If we're collecting a `Vec<_>`, why are we then `to_vec()`?

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2281 @@
...
+    type Item = E;
+
+    // Since we are sending to full-nodes only, and not receiving anything from them,
+    // we don't need to handle any receive here and this is just to satisfy traits
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let this = self.get_mut();
+
+        if this.waker.is_none() {
+            this.waker = Some(cx.waker().clone());
+        }
```
andr-dev, line 1104: Why are we storing a waker here?

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2281 @@
...
+where
+    ST: CertificateSignatureRecoverable,
+    M: Message<NodeIdPubKey = CertificateSignaturePubKey<ST>> + Deserializable<Bytes>,
+    OM: Serializable<Bytes> + Into<M> + Clone,
+{
+    pub fn new(config: &RaptorCastConfig<ST>,
+               channel_from_primary: Receiver<FullNodesGroupMsg<ST>>
+        ) -> Self
+    {
+        tracing::info!("NEW RAPTORCAST-SECONDARY");
```
andr-dev, line 854: Let's not put screaming traces in the logs.

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2281 @@
...
+                                    as CertificateKeyPair>::PubKeyType>
+            = OutboundRouterMessage::FullNodesGroup(group_msg);
+        let udp_msg = BroadcastMsg {
+            stride : monad_dataplane::udp::segment_size_for_mtu(self.mtu),
+            payload : router_msg.serialize(),
+            targets: dest_node_ids
+                .list
+                .iter()
+                .filter_map(|node_id| {
+                    self.known_addresses.get(node_id).map(|addr| (*addr))
```
andr-dev:
```rust
self.known_addresses.get(node_id).cloned()
```

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2281 @@
...
+where
+    ST: CertificateSignatureRecoverable,
+    M: Message<NodeIdPubKey = CertificateSignaturePubKey<ST>> + Deserializable<Bytes>,
+    OM: Serializable<Bytes> + Into<M> + Clone,
+{
+    type Command = RouterCommand<CertificateSignaturePubKey<ST>, OM>;
+
+    fn exec(&mut self, commands: Vec<Self::Command>)
+    {
+        tracing::info!("RaptorCastSecondary EXEC");
```
andr-dev, line 933: I'm not convinced there's a significant value add in logging this message each time exec is called.

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2281 @@
...
+        // Fallback group for error cases.
+        self.curr_group = GroupAsPublisher::new_empty_locked(
+            round, self.scheduling_cfg.init_empty_round_span);
+    }
+
+    // Advances the state machine to the given "time point".
+    // Time point is represented as a Round for now, but can be Duration later.
+    // For now called from UpdateCurrentRound, but later from timer mechanism
+    // May return a message (invite or confirm group) that needs to be sent
+    // to some full-nodes. The set of full-nodes is also returned.
```
andr-dev: This is already implied by the `Option<(FullNodesGroupMsg<ST>, FullNodesST<ST>)>` return type.

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2281 @@
...
+    {
+        self.always_ask_full_nodes = replace_fn;
+        // Remove the nodes from always_ask, otherwise we might send two
+        // invites to the same node.
+        self.peer_disc_full_nodes.list.retain(|node| {
+            !self.always_ask_full_nodes.list.contains(node)
+        });
+    }
+
+    // For Peer Discovery
```
andr-dev: Comment ditto here and for the "For UpdateFullNodes" above. These are unnecessary comments that just add more mental burden when reading the code when you're already using nice descriptive function names.

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2281 @@
...
+    pub fn upsert_peer_disc_full_nodes(
+        &mut self,
+        additional_fn: FullNodes<CertificateSignaturePubKey<ST>>)
+    {
+        for node in additional_fn.list {
+            if !self.peer_disc_full_nodes.list.contains(&node) &&// already have
+               !self.always_ask_full_nodes.list.contains(&node)&& // already ask
+               node != self.validator_node_id // we can't be a candidate
```
andr-dev: This would be much easier to read if the conditions were inverted.

monad-raptorcast/src/config.rs
fabiotdf: Ah thanks! This thing was really bothering me. I was getting a "cannot be sent between threads safely" error and tried a few combinations but the one you suggest is much better.

It's `signing_key` in RaptorCastSecondary and `key` in RaptorCast.

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: Thanks, will keep this in mind for the next PR where I'm integrating the config. This PR only contains the new RaptorCastSecondary that is not active yet + some tests.
Added a comment to the config

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: This is just temporarily for my tests, I will remove before this class becomes active.
It will never appear in prod

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: deleted the comment

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: deleted the comment

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: done

monad-raptorcast/src/lib.rs
michael-yxchen, line 1: we need to list
```rust
pub mod config;
pub mod raptorcast_secondary;
```
to compile them

monad-raptorcast/src/config.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub redundancy: u8,
+}
+
+pub struct RaptorCastConfig<ST>
+where
+    ST: CertificateSignatureRecoverable,
+{
+    // Both primary and secondary instances need to sign raptorcast messages but
+    // CertificateKeyPair can't be copied.
+    pub shared_key: Arc<ST::KeyPairType>,
```
michael-yxchen, line 17: does it have to be Arc<> in config? I think it's possible to have `ST::KeyPairType` in config and `Arc` wrap it in constructor

monad-raptorcast/src/config.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    /// 1_000 = 1 Gbps, 10_000 = 10 Gbps
+    //pub up_bandwidth_mbps: u64,
+    pub mtu: u16,
+
+    /// How should we determine if the process is a full-node or validator?
+    pub is_full_node: bool,
+
+    /// Allows multiple Raptorcast instances in an ParentExecutor to share the same dataplane
+    pub shared_dataplane: Arc<Mutex<Dataplane>>,
```
michael-yxchen: It's a little weird to put `Dataplane` in a config struct. I'd assume a config is something you can deserialize from a file

monad-raptorcast/src/raptorcast_secondary.rs
michael-yxchen, line 1: note on sharing `Dataplane`. The only self mutable part in `udp_write_unicast` is actually incrementing the metric counters. The channel writes take an immutable reference. I think we can modify the metrics counters to be atomic and make the write interface take `&self`. That way we can share `Dataplane` without a lock

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2278 @@
...
+{
+    // Constant data
+    validator_node_id: NodeIdST<ST>,
+    scheduling_cfg: GroupSchedulingConfig,
+
+    // Dynamic data
+    group_schedule: BTreeMap<Round, GroupAsPublisher<ST>>, // start_round -> GroupAsPublisher
+    always_ask_full_nodes: FullNodesST<ST>, // priority ones, coming from config
+    peer_disc_full_nodes: FullNodesST<ST>,  // public ones, via peer discovery
+    rng: StdRng, // random number generator for shuffling full-nodes
```
michael-yxchen: nit: for portability we use `rand_chacha` crate directly. Also `ChaChaRng8` might be sufficient here cuz we don't rely on any security properties of this rng I think

monad-raptorcast/src/config.rs
fabiotdf: How about making it an argument to ::new() and remove it from the config?

michael-yxchen: yeah that works

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2278 @@
...
+
+        // Remove all groups that have ended.
+        while !self.group_schedule.is_empty()
+            && self.group_schedule.first_key_value().unwrap().1
+                .end_round <= round
+        {
+            let group = self.group_schedule.pop_first();
+            tracing::warn!(
+                "Removed stale RaptorcastSecondary group: {:?}", group);
+        }
```
michael-yxchen: might be able to simplify with `self.group_schedule.retain(...)`

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2278 @@
...
+                // when there are gaps in the round sequence?
+                tracing::error!("No group scheduled for RaptorcastSecondary \
+                    round {:?}, next group is {:?}",
+                    round, group);
+            },
+            None => {
+                // Somehow we didn't manage to form a group in time for the
+                // new round. Let's fall back to an empty group for some rounds
+                // while we allow invites for a future group to complete.
+                if self.curr_group.start_round > Round(0){
```
michael-yxchen: i feel this `if` wouldn't be necessary if we initialize appropriately

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2278 @@
...
+        // Fallback group for error cases.
+        self.curr_group = GroupAsPublisher::new_empty_locked(
+            round, self.scheduling_cfg.init_empty_round_span);
+    }
+
+    // Advances the state machine to the given "time point".
+    // Time point is represented as a Round for now, but can be Duration later.
+    // For now called from UpdateCurrentRound, but later from timer mechanism
+    fn step_until(&mut self, time_point: TimePoint)
+        -> Option<(FullNodesGroupMsg<ST>, FullNodesST<ST>)>
```
michael-yxchen: do you have a plan in mind on how (frequently) to call `step_until` function after we switch to wallclock times? We formulate it using timers/timeouts in other places but not sure if it will translate to your use case

monad-raptorcast/src/config.rs
fabiotdf: The constructors for RaptorCastPrimary and RaptorCastSecondary would need to take Arc<ST> as arguments, because those signing keys are not copyable by design.

Having the Arc in the config is more concise because some tests (and potentially monad-node) builds just a RaptorCast for router rather than MultiRouter.

Can I assume we never want to use RaptorCast directly but only via MultiRouter?

michael-yxchen: I think we'll only use MultiRouter in monad-node. Some tests might use only RaptorCast. We can keep it as this and decide when integrating

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2278 @@
...
+            .iter()
+            .skip(self.num_invites_sent)
+            .take(num_missing)
+            .cloned()
+            .collect::<Vec<_>>()
+            .to_vec());
+        let invite_msg = FullNodesGroupMessage::PrepareGroup(prep_grp_data);
+        self.num_invites_sent += next_invitees.list.len();
+        Some((invite_msg, next_invitees))
+    }
```
michael-yxchen: If I read it correctly, we send invite to each full node at most once (never retry). If we have the capacity and time budget, do we want to re-ask those who haven't responded?

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2278 @@
...
+            end_round: start_round + round_span,
+            next_invite_tp: TimePoint::MIN,
+        };
+
+        // Add randomized public nodes.
+        // We don't know how many will refuse, timeout, or ignore the invite,
+        // so we just include all and let a timer periodically pick more nodes.
+        let mut rand_public_nodes = peer_disc_full_nodes.clone();
+        rand_public_nodes.list.shuffle(rng);
+        new_group.full_nodes_candidates.list.extend(rand_public_nodes.list);
```
michael-yxchen: I think we want to exclude duplicated entries, i.e. remove node from `peer_disc_full_nodes` if it is in `always_ask_full_nodes`

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2278 @@
...
+        if self.full_nodes_accepted.list.contains(&candidate) {
+            tracing::warn!("Ignoring duplicate response from FullNode {:?} \
+                who was already accepted into RaptorcastSecondary group {:?}",
+                candidate, self);
+            return;
+        }
+        if accepted {
+            self.full_nodes_accepted.list.push(candidate);
+        } else {
+            self.num_invites_rejected += 1;
```
michael-yxchen, line 483: debug log here. We will have very limited visibility into the client so more logging on the server side is helpful

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2278 @@
...
+
+        for group in self.confirmed_groups.values(curr_round .. curr_round_end)
+        {
+            let mut peer_set = EpochValidators::new();
+            for peer in &group.all_peers {
+                let common_stake = Validator{stake:Stake(1)};
+                peer_set.validators.insert(peer.clone(), common_stake);
+            }
+            self.curr_raptorcast_groups.insert(group.validator_id, peer_set);
+        }
```
michael-yxchen: I wonder if it simplifies the logic if we split confirmed group into two disjoint sets: activated and confirmed (but not activated)
- activated groups start round <= current round < end round
- current round < confirmed group start round

Groups move in one direction: accepted -> confirmed -> activated -> (garbage collected)

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2278 @@
...
+    // Lets Raptorcast's UdpState know which peers to re-broadcast to.
+    // Must match the argument 'get_peers` for UdpState::handle_message().
+    // `get_peers` is the refactored version of old arg `epoch_validators`.
+    pub fn get_curr_raptorcast_group(&self, _: Epoch, author: NodeIdST<ST>
+        ) -> EpochValidators<ST>
+    {
+        self.curr_raptorcast_groups.get(&author)
+            .unwrap_or(&EpochValidators::new())
+            .clone()
+    }
```
michael-yxchen: should we return the group members instead of epoch validators? If raptorcast is enforcing the type to be EpochValidators, I think we should change it there

monad-raptorcast/src/raptorcast_secondary.rs
andr-dev: Haha I was actually thinking about this two days ago because with my new metrics implementation we can immediately make the Dataplane reference immutable with minor code changes. I already made a changeset that introduces Dataplane metrics + makes this immutable, not sure if it'll be ready before this.

monad-raptorcast/src/config.rs
fabiotdf: Thanks, I tried to remove the Arc from config but it was quite tricky for me to adapt e.g. testground to it, as it first creates configs and then instantiate raptorcast. Testground (in my other branch) also allows of both RaptorCast and MulitRouter (=RaptorCast + RaptorCastSecondary) and so I pass the same config as & reference to both primary and secondary instances. Therefore I cannot take the key out of the config as this would leave the config partially moved for when I need to pass it to the secondary instance.

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: Yeah that would be great, this way this PR can be a bit more isolated

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: Done

monad-raptorcast/src/config.rs
fabiotdf: Done

monad-raptorcast/src/lib.rs
fabiotdf: Added this because otherwise the unit tests wouldn't run!

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: Done

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: Thanks!

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: Initially I had but the implementation changed. 
Maybe it makes more sense to keep Round as a time unit for now?

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: We don't retry for the same round, but for other rounds we randomize and the same node can be asked again.

fabiotdf: In the implementation, a node not responding in time is treated the same way as if the node refused to join (i.e. due to bandwidth). I was thinking that if it times out, the node is probably too busy, restarting, network congested, crashed, etc. The node can still get invited for future rounds, if it is in an always_ask node or a random node from peer discovery.

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: `upsert_peer_disc_full_nodes()` and `update_always_ask_full_nodes()` already makes sure they only contain unique entries and the sets don't overlap. I've also added a check to the constructor to make sure the always_ask nodes we get from config are also unique

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: Done

monad-raptorcast/Cargo.toml
fabiotdf: Done

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: What behaviour would you prefer?

monad-raptorcast/src/raptorcast_secondary.rs
michael-yxchen: cool thanks for the clarification

monad-raptorcast/src/raptorcast_secondary.rs
michael-yxchen: makes sense. let's keep your current impl

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2254 @@
...
+
+}
+
+//==============================================================================
+// For when the router is playing the role of: CLIENT
+// That is, we are a full-node receiving group invites from a validator
+//==============================================================================
+
+type Bandwidth = i64;
```
michael-yxchen: nit: any reason to use signed numbers for bandwidth?

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2254 @@
...
+                            "RaptorcastSecondary received self-overlapping \
+                            invite for rounds [{:?}, {:?}) from validator {:?}",
+                            invite_msg.start_round, invite_msg.end_round,
+                            invite_msg.validator_id);
+                        accept = false;
+                    }
+                    // Check that we'll have enough bandwidth during round span
+                    future_bandwidth +=
+                        self.bandwidth_cost(group.all_peers.len());
+                }
```
michael-yxchen: should we include `pending_confirms` in this bandwidth calculation?

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2254 @@
...
+                Role::Publisher(publisher)
+            },
+            super::config::SecondaryRcModeConfig::Client(client_cfg) => {
+                let client = Client::new(node_id, client_cfg);
+                Role::Client(client)
+            },
+            super::config::SecondaryRcModeConfig::None =>
+                panic!("secondary_instance is not set in config during \
+                    instantiation of RaptorCastSecondary"),
+        };
```
michael-yxchen: leaving a note here: we're planning to merge `monad-node` (validator binary) and `monad-full-node` (full node binary), so it's possible for a node to switch between roles at runtime

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2254 @@
...
+
+    fn send_group_msg(
+        &self,
+        group_msg: FullNodesGroupMsg<ST>,
+        dest_node_ids: FullNodesST<ST>,
+    )
+    {
+        let router_msg: OutboundRouterMessage<'_, OM,
+                                <<ST as CertificateSignature>::KeyPairType
+                                    as CertificateKeyPair>::PubKeyType>
```
michael-yxchen: we have type alias for this named `CertificateSignaturePubkey<ST>`

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2254 @@
...
+
+}
+
+impl<ST, M, OM, SE> Executor for RaptorCastSecondary<ST, M, OM, SE>
+where
+    ST: CertificateSignatureRecoverable,
+    M: Message<NodeIdPubKey = CertificateSignaturePubKey<ST>> + Deserializable<Bytes>,
+    OM: Serializable<Bytes> + Into<M> + Clone,
+{
+    type Command = RouterCommand<CertificateSignaturePubKey<ST>, OM>;
```
michael-yxchen, line 930: note for integration: I would consider creating a new RouterCommand type for RC secondary. `Multirouter::exec` splits top level router command into respective router commands

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2254 @@
...
+    ST: CertificateSignatureRecoverable,
+{
+    start_round: Round, // inclusive
+    end_round: Round, // exclusive
+    validator_id: NodeIdST<ST>, // the raptorcast publisher for this group
+    all_peers: Vec<NodeIdST<ST>>, // including self + author
+}
+
+//==============================================================================
+// TODO: Should Role be a trait with member functions instead of an enum?
```
michael-yxchen: enum is good to support changing roles at runtime

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2254 @@
...
+                },
+                Role::Client(client) => {
+                    if let Some((out_msg, dest_validator_node_id)) =
+                        client.on_receive_group_message(msg)
+                    {
+                        this.send_single_msg(out_msg, &dest_validator_node_id);
+                    }
+                },
+            }
+        }).map_or(Poll::Pending, Poll::Ready);
```
michael-yxchen: I think we should handle TryRecvError Empty and Disconnected separately

michael-yxchen:
https://github.com/category-labs/category-internal/issues/1382#issuecomment-2797984931

monad-discovery/src/message.rs
fabiotdf: Removed the separators. If you don't mind, I will do the file splint in the next PR. 
The current PR is a subset of larger PR that exists in a separate branch and I've split them due to the scope.
Currently I'm maintaining the 2 branches in synch but if I start moving
things between files it will get quite tricky to maintain the sync.

monad-discovery/src/message.rs
fabiotdf: Same as above, removed the separators but would like to leave the file split for the next PR

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: I had that originally but then the data structure returned to represent
the raptorcast group was recreated on every message/proposal instead of
just once per round.

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: Done. However I left the changes to UdpState in my other PR.

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: Ups, before some refactoring I had a `self.remaining_bandwidth`, which could become negative during run-time adjustment. 
This is no longer the case so I changed the definition to u64. Thanks!

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: Done

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: Thanks for the heads up! Comments updated

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: I had that originally but there was a lot of duplication in the code. I can try again during integration, perhaps it will come out cleaner this time after all the refactoring already gone into it

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: Makes sense, thanks! Comment updated

monad-raptorcast/src/raptorcast_secondary.rs
andr-dev: In that case, keep it in your stash or in a local branch. We shouldn't add code that's "temporary" into master.

Cargo.toml
```diff
@@ -132,6 +132,7 @@ hdwallet = "0.4.1"
 hex = "0.4"
 indexmap = "2.4.0"
 inotify = "0.11"
+iset = "0.3.1"
```
andr-dev: This dep is out of order.

monad-raptorcast/src/raptorcast_secondary.rs
michael-yxchen: what's being recreated? Trying to understand if you can lookup group in activated

monad-raptorcast/src/raptorcast_secondary.rs
michael-yxchen: I feel a blend of round and timer driven is more reasonable. You'd want to timeout PrepareGroup requests based on time

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: Sorry, don't know why this comment wasn't showing up for me before.
I copied this pattern from the primary instance, which also uses this to wake up consumption of `PeerManagerResponse` from this.pending_events

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: Yes it will be removed before merging this to master, I just want a green light for this PR before removing it, so its easier to maintain my other branch in synch. 
The same applies to the directory layout for the protobuf messages.
Yet a third thing is the commented out `PublishToFullNodes`.

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: Done

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: In the current implementation, `curr_raptorcast_groups`  only gets recreated when the round changes.
With the alternative implementation, a similar structure would be created every time we look up a group.
So splitting it doesn't simplify but makes things a bit more complicated (and I think also slower)

Cargo.toml
fabiotdf: Fixed

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: I tried the atomic + `Arc<Dataplane>` approach but got stuck in this error in RaptorCast::poll_next():
```
error[E0596]: cannot borrow data in an `Arc` as mutable
   --> monad-raptorcast/src/lib.rs:477:45
    |
477 |             let Poll::Ready(message) = pin!(this.dataplane.udp_read()).poll_unpin(cx) else {
    |                                             ^^^^^^^^^^^^^^ cannot borrow as mutable
    |
    = help: trait `DerefMut` is required to modify through a dereference, but it is not implemented for `Arc<Dataplane>`
```
udp_read() is mut because Received::recv() is mut. How to get past this?

michael-yxchen: I see. A solution could be adding a method to dataplane that returns an immutable writer half. Primary owns dataplane. Secondary writes to the writing half

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2296 @@
...
+            if round > self.curr_round + Round(1) {
+                tracing::warn!(
+                    "RaptorCastSecondary detected round gap \
+                    {:?} -> {:?}",
+                    self.curr_round,
+                    round
+                );
+            }
+            self.curr_round = round;
+        }
```
michael-yxchen, line 148: unnecessary scope

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2296 @@
...
+            );
+            let maybe_invites =
+                new_group.advance_invites(time_point, &self.scheduling_cfg, self.validator_node_id);
+            self.group_schedule.insert(new_group.start_round, new_group);
+            return maybe_invites;
+        }
+
+        // No invites to send out at the moment.
+        None
+    }
```
michael-yxchen, line 237: I feel correctness of the function (early return) is implied by `round_span + deadline_round_dist > invite_lookahead`, which is probably something any reasonable validator will use. We can either explicitly verify config in constructor, or remove the early return and always check both branches

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2296 @@
...
+            None => {
+                // We didn't manage to form a group in time for the new round.
+                // Might be due to gap, unexpectedly long delays or Round 0/
+                // We currently have an empty group for some rounds while we
+                // allow invites for a future group to complete.
+            }
+        }
+        // Fallback group for error cases.
+        self.curr_group =
+            GroupAsPublisher::new_empty_locked(round, self.scheduling_cfg.init_empty_round_span);
```
michael-yxchen, line 196: why do we want to lock an empty group? is it possible to start inviting to the next group right away?

monad-raptorcast/src/raptorcast_secondary.rs
```diff
@@ -0,0 +1,2296 @@
...
+where
+    ST: CertificateSignatureRecoverable,
+{
+    full_nodes_accepted: FullNodesST<ST>,
+
+    // Pre-randomized permutation of always_ask_full_nodes[] + peer_disc_full_nodes[]
+    // Stays const once created.
+    full_nodes_candidates: FullNodesST<ST>,
+    num_invites_sent: usize,     // also an index into full_nodes_candidates[]
+    num_invites_rejected: usize, // just for debug info for now
```
michael-yxchen, line 313: a clearer way of organizing this is making a PublisherState enum. `Inviting`, `Pending` (activation), `Active`, so we're not populating fields with meaningless values. we can improved it in a future PR

monad-raptorcast/src/raptorcast_secondary.rs
michael-yxchen: what's preventing us from storing the group once it's confirmed, and retrieve that on a lookup? The only events that can change group status are 1) round 2) group messages. I'm missing why it needs to be recreated every lookup

fabiotdf: Originally `curr_raptorcast_groups` was in the format needed (or close to) by RaptorCast's `UdpState` (an EpochValidator map).
UdpState doesn't know what the current round is, so it cannot lookup into round spans, and hence the user of both `Client` and `UdpState`. At this "glue", this argument would have to be recreated on each invocation into UdpState

However, this glue around `curr_raptorcast_groups` is currently being refactored. Let me get back to you to get your thoughts on the new api

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: I think I now understand this. In your suggestion, `Inviting` would e.g. embed the fields `num_invites_sent` and `num_invites_rejected`, and these wouldn't be visible in the other states. Correct?

I will try squeeze this in after you've had a look at the integration branch I've [linked offline](https://github.com/category-labs/monad-bft/blob/ab7f2b6a1b30d8f75b7b4b6e59bf909f182185fb/monad-raptorcast/src/udp.rs#L117) you in the chat, just to simplify the merging.

In the integration branch above, I think `Active` state in your suggestion would be replaced by the new `util::Group` class, which is directly usable by UdpState send/recv API, shortening `GroupAsPublisher` life time so that it only exists during `Inviting` and `Pending` states

monad-raptorcast/src/raptorcast_secondary.rs
fabiotdf: We do get to start inviting but from `step_until()` rather than `enter_round()`

If we always call both of the methods about, I think we should never reach line 195, but I initialize anyway to something safe just as a defensive measure. I reused the `locked` init rather than creating a new one. This init just makes locally (from `GroupAsPublisher`) sure that it is impossible to add peers to it in the future.

PR #1704, opened 2025-04-04
Move node config from monad-node to config crate

andr-dev:
Previously, `monad-node` served as both a library and a binary which was counter-intuitive. Additionally, this library would export types that should live in the `monad-node-config` crate. This led to an issue with a recent change where RPC would be unable to read the same `node.toml` config used by BFT due to a poor type export. This changeset addresses all of the aforementioned issues.


monad-node/examples/ledger-tail.rs
michael-yxchen, line 1: can we switch to MonadNodeConfig on this line?
```rust
let node_config: NodeConfig<monad_secp::PubKey> = toml::from_str(
```

PR #1707, opened 2025-04-07
Limit amount of memory allocated on the statesync client

maxkozlovsky:
Send completion message after processing server responses. Limit messages outstanding to the client without receiving completion messages. This puts a limit on the amount of memory that can be used on the client.

Timeout the client session if the client fails to acknowledge messages within certain period.

Handle server having mismatched version. If server does not support clients maximum version, downgrade version and retry requests.

### DIFF
--- a/monad-statesync/src/outbound_requests.rs
+++ b/monad-statesync/src/outbound_requests.rs
@@ -5,10 +5,16 @@ use std::{
 };
 
 use monad_crypto::certificate_signature::PubKey;
-use monad_executor_glue::{StateSyncBadVersion, StateSyncRequest, StateSyncResponse};
+use monad_executor_glue::{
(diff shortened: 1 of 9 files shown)


monad-statesync/src/outbound_requests.rs
```diff
@@ -81,58 +93,63 @@ impl<PT: PubKey> InFlightRequest<PT> {
...
+            );
+            return Vec::new();
+        }
+
+        if response.response_index == self.response_index {
+            let curr_index = self.response_index;
+            self.response_index += 1;
+            let mut responses = vec![response];
+
+            // Remove consequitive responses from out-of-order queue
```
omegablitz: nit: consecutive

maxkozlovsky:
https://github.com/category-labs/category-internal/issues/1042

PR #1708, opened 2025-04-08
monad-triedb: get rid of copying every triedb response

dshulyak:
closes: https://github.com/category-labs/category-internal/issues/1494

introduced new type TriedbVec.
this type is initialized in a callbacks from triedb executor and passed back to the callers, specifically rpc.

in rpc it is usually used to decode into a concrete type and dropped after that

- [ ] is it safe to dealloc triedb buffer in a different thread?


monad-triedb/src/lib.rs
```diff
@@ -121,15 +178,8 @@ pub unsafe extern "C" fn traverse_callback(
...
-        let key = std::slice::from_raw_parts(key_ptr, key_len).to_vec();
-        key
-    };
-
-    let value = unsafe {
-        let value = std::slice::from_raw_parts(value_ptr, value_len).to_vec();
-        value
-    };
+    let key = unsafe { TriedbVec::from_response(key_ptr, key_len) };
+    let value = unsafe { TriedbVec::from_response(value_ptr, value_len) };
```
dshulyak, line 182: i had wrong assumption. buffer is freed in cpp code, so we need to copy here

dshulyak:
i will propose another way to address this

PR #1710, opened 2025-04-09
Split validator configs from forkpoint 

omegablitz:
With this commit, the validator sets for epochs are specified separately from the forkpoint. The one gotcha here is that if the new validators.toml file is invalid at the point in time the next epoch rolls around, the node will crash. This is intentional, as unlike the node.toml config, using a stale validator set can be quite bad.

### DIFF
--- a/monad-updaters/src/triedb_state_root_hash.rs
+++ b/monad-updaters/src/triedb_state_root_hash.rs
@@ -2,7 +2,7 @@ use std::{
     collections::HashMap,
     marker::PhantomData,
     ops::DerefMut,
-    path::Path,
+    path::{Path, PathBuf},
     pin::Pin,
     sync::{Arc, Mutex},
     task::{Context, Poll, Waker},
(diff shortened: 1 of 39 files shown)


monad-updaters/src/triedb_state_root_hash.rs
```diff
@@ -177,35 +176,17 @@ where
...
-
-            self.next_val_data = Some(next_validator_data);
+            self.next_val_data = Some(ValidatorSetDataWithEpoch {
+                epoch: locked_epoch,
+                validators: ValidatorsConfig::read_from_path(&self.validators_path)
+                    // I'm hesitant to provide any fallback for this, because
+                    // having the wrong validator set can be catastrophic.
+                    //
+                    // This file should never be manually edited anyways.
+                    .expect("failed to read validators_path")
```
omegablitz, line 186: Make this more verbose, validators.toml file corrupt, print the actual error

omegablitz: "did you manually try and edit this file"

eerkaijun:
can we also update https://github.com/category-labs/monad-bft/blob/master/docker/single-node/nets/compose.yaml#L70 to include the new `--validators-path` argument

PR #1711, opened 2025-04-09
Add RequestId to timing traces

eclark0426:


monad-rpc/src/timing.rs
```diff
@@ -30,6 +31,30 @@ where
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct RequestId {
+    id: String,
```
ariqchowdhury: does this need to be a string? Can't it just be the int?

monad-rpc/src/timing.rs
```diff
@@ -30,6 +31,30 @@ where
...
+}
+
+impl RequestId {
+    pub fn new() -> Self {
+        Self {
+            id: format!("{:x}", rand::random::<u128>()),
```
ariqchowdhury: u64 is should be fine?

monad-rpc/src/timing.rs
eclark0426: Ah yeah probably - originally I was trying to use a UUID

PR #1712, opened 2025-04-09
Call eth_call during gas estimation when no calldata is provided

philipglazman:
Previously we assumed that omission of calldata meant that we can always return 21_000 for gas estimation. To be safe, we should execute eth_call to check that assumption since it can be a call to a payable function without calldata

fixes https://github.com/category-labs/category-internal/issues/1499

### DIFF
--- a/monad-rpc/src/gas_handlers.rs
+++ b/monad-rpc/src/gas_handlers.rs
@@ -8,9 +8,7 @@ use alloy_primitives::{Address, TxKind, U256, U64};
 use alloy_rpc_types::FeeHistory;
 use monad_ethcall::{CallResult, EthCallExecutor, StateOverrideSet};
 use monad_rpc_docs::rpc;
-use monad_triedb_utils::triedb_env::{
-    BlockKey, FinalizedBlockKey, ProposedBlockKey, Triedb, TriedbPath,
-};
+use monad_triedb_utils::triedb_env::{BlockKey, FinalizedBlockKey, ProposedBlockKey, Triedb};
 use monad_types::{Round, SeqNum};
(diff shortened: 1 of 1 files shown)


monad-rpc/src/gas_handlers.rs
```diff
@@ -96,8 +96,19 @@ async fn estimate_gas<T: EthCallProvider>(
...
     if matches!(txn.kind(), TxKind::Call(_)) && txn.input().is_empty() {
-        return Ok(Quantity(21_000));
-    }
+        let mut request = call_request.clone();
+        request.gas = Some(U256::from(21_000));
+        let txn: TxEnvelope = request.try_into()?;
+
+        if matches!(
+            provider
+                .eth_call(txn.clone(), eth_call_executor.clone())
```
eerkaijun: before sending to eth_call, I think we should get_code for `call_request.to` to check if there is any deployed code at the address (similar to geth and reth implementation), with this we do one less eth_call when there's code deployed

monad-rpc/src/gas_handlers.rs
```diff
@@ -96,8 +96,19 @@ async fn estimate_gas<T: EthCallProvider>(
 
     // simple transfer
```
eerkaijun: can we also expand the comment here for clarity? here's geth comment:
```
// If the transaction is a plain value transfer, short circuit estimation and
// directly try 21000. Returning 21000 without any execution is dangerous as
// some tx field combos might bump the price up even for plain transfers (e.g.
// unused access list items). Ever so slightly wasteful, but safer overall.
```

PR #1713, opened 2025-04-10
Archive: chunk large mongo documents to avoid 16mb limit

JoeHowarth:
Includes https://github.com/category-labs/monad-bft/pull/1714

### DIFF
--- a/monad-archive/src/bin/monad-archiver/cli.rs
+++ b/monad-archive/src/bin/monad-archiver/cli.rs
@@ -35,6 +35,13 @@ pub struct Cli {
     #[arg(long)]
     pub stop_block: Option<u64>,
 
+    /// Skip bad blocks
+    /// If set, archiver will skip blocks that fail to archive
+    /// and log an error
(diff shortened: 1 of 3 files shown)


monad-archive/src/kvstore/mongo.rs
```diff
@@ -163,12 +192,40 @@ impl KVStore for MongoDbStorage {
...
+                    .await
+                    .wrap_err_with(|| {
+                        format!("MongoDB put operation failed for chunk {}", chunk_num)
+                    })?;
+            }
+
+            KeyValueDocument {
+                _id: key.as_ref().to_string(),
+                value: None,
+                chunks: Some((data.len() as f32 / CHUNK_SIZE as f32).ceil() as u32),
```
omegablitz: https://doc.rust-lang.org/std/primitive.usize.html#method.div_ceil

JoeHowarth: oh nice, updated

PR #1714, opened 2025-04-10
Add --unsafe-skip-bad-blocks

JoeHowarth:
This allows continueing archiving past known bad
blocks.
This option should not be enabled during normal operation

### DIFF
--- a/monad-archive/src/bin/monad-archiver/cli.rs
+++ b/monad-archive/src/bin/monad-archiver/cli.rs
@@ -35,6 +35,13 @@ pub struct Cli {
     #[arg(long)]
     pub stop_block: Option<u64>,
 
+    /// Skip bad blocks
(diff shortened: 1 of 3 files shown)


monad-archive/src/bin/monad-archiver/cli.rs
```diff
@@ -35,6 +35,13 @@ pub struct Cli {
     #[arg(long)]
     pub stop_block: Option<u64>,
 
+    /// Skip bad blocks
+    /// If set, archiver will skip blocks that fail to archive
+    /// and log an error
+    /// DO NOT ENABLE UNDER NORMAL OPERATION
+    #[arg(long)]
```
omegablitz, line 42: does this need a `default_value_t`?

JoeHowarth: boolean flags default to false

PR #1715, opened 2025-04-10
Add artificial limits to eth_call and eth_estimateGas gas amounts

eclark0426:
monad-integration updated as well to account for new gas limits: https://github.com/category-labs/monad-integration/pull/245

### DIFF
--- a/monad-rpc/src/cli.rs
+++ b/monad-rpc/src/cli.rs
@@ -65,6 +65,14 @@ pub struct Cli {
     #[arg(long, default_value_t = 102400)]
     pub eth_call_executor_node_lru_size: u32,
 
+    /// Set the gas limit for eth_call
+    #[arg(long, default_value_t = 30_000_000)]
+    pub eth_call_gas_limit: u64,
+
+    /// Set the gas limit for eth_estimateGas
+    #[arg(long, default_value_t = 30_000_000)]
+    pub eth_estimate_gas_gas_limit: u64,
+
     /// Set the max concurrent requests for triedb reads
     #[arg(long, default_value_t = 20_000)]
     pub triedb_max_buffered_read_requests: u32,
--- a/monad-rpc/src/gas_handlers.rs
+++ b/monad-rpc/src/gas_handlers.rs
@@ -91,6 +91,8 @@ async fn estimate_gas<T: EthCallProvider>(
     provider: &T,
     eth_call_executor: Option<Arc<Mutex<EthCallExecutor>>>,
     call_request: &mut CallRequest,
+    gas_limit: u64,
+    protocol_gas_limit: u64,
 ) -> Result<Quantity, JsonRpcError> {
     let mut txn: TxEnvelope = call_request.clone().try_into()?;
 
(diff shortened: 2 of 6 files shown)


monad-rpc/src/cli.rs
```diff
@@ -65,6 +65,14 @@ pub struct Cli {
     #[arg(long, default_value_t = 102400)]
     pub eth_call_executor_node_lru_size: u32,
 
+    /// Set the gas limit for eth_call
+    #[arg(long, default_value_t = 30_000_000)]
+    pub eth_call_gas_limit: u64,
+
+    /// Set the gas limit for eth_estimateGas
+    #[arg(long, default_value_t = 30_000_000)]
```
eerkaijun, line 73: i think we should default the gas limit to be block gas limit (150M in current testnet)

monad-rpc/src/gas_handlers.rs
```diff
@@ -91,6 +91,8 @@ async fn estimate_gas<T: EthCallProvider>(
     provider: &T,
     eth_call_executor: Option<Arc<Mutex<EthCallExecutor>>>,
     call_request: &mut CallRequest,
+    gas_limit: u64,
+    protocol_gas_limit: u64,
```
eerkaijun, line 95: why do we need both gas limit and protocol gas limit? I think if we decide to make eth_call gas limit a CLI argument, that should just be the de facto gas limit (i.e. block gas limit would not matter anymore)

eclark0426: The thinking here was that we would want a different response if they've hit the artificial limit as opposed to the protocol limit, since with the artificial limit on eth_call/estimateGas, a transaction might run out of gas when it would run fine when the actual transaction is submitted. It's to cover the case where `aritificialLimit < actualGasConsumed < protocolLimit`

monad-rpc/src/cli.rs
eclark0426: Ariq, Andre and I had been talking about the default value and 30 million is what we had decided seemed appropriate as a default - the thinking was that this way, the default value will reduce the number of lengthy eth_calls for anyone who doesn't explicitly set the option, and anyone who intentionally changes it would know that they are allowing the possibility of eth_calls that hang a while

monad-rpc/src/gas_handlers.rs
```diff
@@ -108,9 +110,15 @@ async fn estimate_gas<T: EthCallProvider>(
             gas_refund,
             ..
         }) => (gas_used, gas_refund),
-        monad_ethcall::CallResult::Failure(error) => {
-            return Err(JsonRpcError::eth_call_error(error.message, error.data))
-        }
+        monad_ethcall::CallResult::Failure(error) => match error.error_code {
+            monad_ethcall::EthCallResult::OutOfGas if gas_limit < protocol_gas_limit => {
```
eerkaijun, line 114: what about the else statement here (ie. `gas_limit` >= `protocol_gas_limit`)? I think we should still return an error here rather than returning an estimated gas value

eclark0426: I think this case should be covered by what gets filled in from the `fill_gas_params` call before the call to estimateGas - that will make sure that the highest gas in the request is the block gas limit: https://github.com/category-labs/monad-bft/blob/37dad879399ad229fd21301aca4889f944a2bb33/monad-rpc/src/call.rs#L356

PR #1716, opened 2025-04-10
Implement trace_block for rpc

philipglazman:
Fixes https://github.com/category-labs/category-external/issues/70

### DIFF
--- a/monad-rpc/src/trace_handlers.rs
+++ b/monad-rpc/src/trace_handlers.rs
@@ -5,6 +5,10 @@ use alloy_primitives::{
     Address, Bytes,
 };
 use alloy_rlp::Decodable;
+use alloy_rpc_types::trace::parity::{
+    Action, CallAction, CallOutput, CallType, CreateAction, CreateOutput, CreationMethod,
+    LocalizedTransactionTrace, SelfdestructAction, TraceOutput, TransactionTrace,
+};
 use monad_archive::prelude::{ArchiveReader, BlockDataReader, IndexReader};
 use monad_rpc_docs::rpc;
 use monad_triedb_utils::triedb_env::{BlockKey, FinalizedBlockKey, Triedb};
@@ -114,7 +118,7 @@ pub struct MonadDebugTraceTransactionParams {
     tracer: TracerObject,
 }
 
-#[derive(Serialize, Debug, schemars::JsonSchema)]
+#[derive(Clone, Serialize, Debug, schemars::JsonSchema)]
 #[serde(rename_all = "camelCase")]
 pub struct MonadCallFrame {
     #[serde(rename = "type")]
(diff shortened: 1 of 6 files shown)


monad-rpc/src/trace_handlers.rs
```diff
@@ -370,6 +406,186 @@ pub async fn monad_debug_traceTransaction<T: Triedb>(
...
+            }),
+        };
+
+        let result = match top_level_frame.result.typ {
+            CallKind::Create | CallKind::Create2 => TraceOutput::Create(CreateOutput {
+                address: to,
+                code: frame.output.0.clone().into(),
+                gas_used: frame.gas_used.0,
+            }),
+            _ => TraceOutput::Call(CallOutput {
```
andr-dev: Can we have the remaining variants explicitly enumerated or are there too many?

andr-dev (dismissed):
Looks reasonable, not too familiar with the trace implementation.

philipglazman:
Can I get another review? I updated the flexnet tests to account for more complex subtraces and found that my implementation was not including children call frames. In the response, it is expected we return a flattened call frame response instead of just the top level one.

eerkaijun (dismissed):

monad-rpc/src/trace_handlers.rs
```diff
@@ -536,6 +764,29 @@ async fn build_call_tree(
...
+fn flatten_call_tree_with_paths(root: MonadCallFrame) -> Vec<(Vec<usize>, MonadCallFrame)> {
+    let mut result = Vec::new();
+    let mut stack = vec![(vec![], root)];
+
+    while let Some((path, mut current)) = stack.pop() {
+        result.push((path.clone(), current.clone()));
+        let children = std::mem::take(&mut current.calls);
+
+        for (i, child) in children.iter().rev().enumerate() {
+            let index = children.len() - 1 - i;
```
andr-dev, line 776: We can flip the order of the enumerate here to avoid creating `index` manually.
```rust
for (index, child) in children.iter().enumerate().rev() {
  ...
  ```

File: monad-rpc/src/trace_handlers.rs [hunk 3]
```diff
@@ -394,6 +427,201 @@ pub async fn monad_debug_traceTransaction<T: Triedb>(
...
+    Ok(res)
+}
+
+fn count_sub_call_frames(root: &MonadCallFrame) -> usize {
+    let mut count = 0;
+    let mut stack = vec![];
+
+    for child in &root.calls {
+        stack.push(Rc::clone(child))
+    }
```
andr-dev, line 611: I think we can directly build stack here.
```rust
let mut stack = root.calls.iter().map(Rc::clone).collect();
```

monad-rpc/src/trace_handlers.rs
```diff
@@ -394,6 +427,201 @@ pub async fn monad_debug_traceTransaction<T: Triedb>(
...
+            if let Ok(call_frames) = archive_reader.get_block_traces(block_num.0).await {
+                let tx_ids = block
+                    .body
+                    .transactions
+                    .iter()
+                    .map(|tx| *tx.tx.tx_hash())
+                    .collect::<Vec<_>>();
+
+                let mut traces = Vec::new();
+                for (call_frame, tx_id) in call_frames.iter().zip(tx_ids.into_iter()) {
```
andr-dev, line 500: We should probably produce an error if `call_frames.len() != tx_ids.len()`. Should never happen, but if it does this method might produce erroneous data.

philipglazman:
Requires https://github.com/category-labs/monad/pull/1300 in execution in order for the call frames to be returned correctly

PR #1719, opened 2025-04-14
Integrate peer discovery into swarm

eerkaijun:


monad-peer-disc-swarm/tests/ping_pong.rs
```diff
@@ -65,11 +97,181 @@ fn test_ping_pong() {
...
+    let new_name_record = MonadNameRecord {
+        name_record: new_name_record,
+        signature,
+    };
+    let cmds = node_a_state
+        .peer_disc_driver
+        .update_name_record(new_name_record);
+    let router_cmds = node_a_state.peer_disc_driver.filter_and_exec(cmds);
+    node_a_state.executor.exec(router_cmds);
+    while nodes.step_until(Duration::from_secs(0)) {}
```
eerkaijun: feels like I'm not using monad-peer-disc-swarm correctly, is there a better way to do this? basically i want to manually invoke some cmds which should be executed by the node

michael-yxchen:
NoserRouterScheduler (or any router scheduler) routes messages by NodeId. That means messages are delivered no matter you have the peer's name record or no. To test peer discovery in a more realistic environment, we should route by SockAddr. Let me know if I can help

monad-peer-disc-swarm/tests/ping_pong.rs
michael-yxchen: i didn't think of this use case when writing swarm

some ideas
- introduce a new `SwarmEvent` type `TestEvent`, handle it in `Node.step_until`, calling `driver.update()` and handle side effects accordingly
- change `peer_disc_driver.update_name_record` to buffer side effects and apply them on next `step_until`. this might not work well with the existing `SwarmEvent` variants, so we might need new events anyways
- remove the state/Node and re-add it with different name record. This might be closer to what will happen in practice

michael-yxchen: https://github.com/category-labs/monad-bft/pull/1730

monad-peer-disc-swarm/tests/ping_pong.rs
```diff
@@ -29,30 +35,52 @@ impl PeerDiscSwarmRelation for PingPongPeerDiscSwarm {
     >;
 }
 
+type KeyPairType = NopKeyPair;
+type PubKeyType = NopPubKey;
 type SignatureType = NopSignature;
```
michael-yxchen: nit: prefer to use `type KeyPairType = <SignatureType as CertificateSignature>::KeyPairType;` for consistency

monad-peer-discovery/src/discovery.rs
```diff
@@ -1,40 +1,101 @@
...
     pub outstanding_lookup_requests: HashMap<u32, NodeId<CertificateSignaturePubKey<ST>>>,
     pub metrics: PeerDiscMetrics,
+    pub ping_period: Duration,
     pub rng: ChaCha8Rng,
 }
 
+pub struct PeerDiscoveryBuilder<ST: CertificateSignatureRecoverable> {
+    pub self_id: NodeId<CertificateSignaturePubKey<ST>>,
+    pub self_record: MonadNameRecord<ST>,
+    pub peer_info: BTreeMap<NodeId<CertificateSignaturePubKey<ST>>, PeerInfo<ST>>,
```
michael-yxchen, line 44: is there a situation where we'd specify `last_ping` and `last_seen`? If not, we can make it a `BTreeMap<NodeId, MonadNameRecord>`, or just `Vec<MonadNameRecord>` as pubkey can be recovered from secp signature

monad-peer-discovery/src/discovery.rs
```diff
@@ -1,40 +1,101 @@
...
+            ping_period: self.ping_period,
+            rng: ChaCha8Rng::seed_from_u64(self.rng_seed),
+        };
+
+        let cmds = self
+            .peer_info
+            .into_iter()
+            .filter(|(node_id, _)| *node_id != self.self_id)
+            .flat_map(|(node_id, _)| state.send_ping(node_id))
+            .collect();
```
michael-yxchen, line 75: to be addressed in a separate PR - we can think about how to make periodic ping less bursty

monad-peer-disc-swarm/tests/ping_pong.rs
```diff
@@ -29,30 +35,52 @@ impl PeerDiscSwarmRelation for PingPongPeerDiscSwarm {
...
 
+fn generate_name_record(keypair: &KeyPairType) -> MonadNameRecord<SignatureType> {
+    let name_record = NameRecord {
+        address: SocketAddrV4::from_str("1.1.1.1:8000").unwrap(),
```
michael-yxchen, line 42: using the same address for nodes is fine for now, because router scheduler routes message by target node id. I'm planning to change it to route by address instead. we will need to allocate unique addresses to nodes then

monad-peer-disc-swarm/tests/ping_pong.rs
```diff
@@ -65,11 +93,193 @@ fn test_ping_pong() {
...
+    let mut nodes = swarm_builder.build();
+    while nodes.step_until(Duration::from_secs(0)) {}
+
+    // NodeA, NodeB and NodeC should now have peer_info of each other
+    for state in nodes.states().values() {
+        let state = state.peer_disc_driver.get_peer_disc_state();
+        for node_id in all_peers.keys() {
+            assert!(state.peer_info.contains_key(node_id));
+        }
     }
```
michael-yxchen: are there any useful metrics to assert on?

eerkaijun: added assertion on number of requests and responses

monad-peer-discovery/src/discovery.rs
eerkaijun: I think for testing purposes we might want to be able to start with some predefined state (e.g. testing of pruning with outdated last_seen etc)

monad-peer-disc-swarm/tests/ping_pong.rs
andr-dev: Can we use `0.0.0.0:8000` instead? Feels like a more reasonable dummy value.

eerkaijun: we won't use hardcoded dummy value anymore in a subsequent pr

PR #1720, opened 2025-04-15
monad-dataplane: set socket buffers to 62500000 by default

dshulyak:
closes: https://github.com/category-labs/category-internal/issues/1505

socket buffer sizes are configurable and default to 62_500_000 as recommended in the setup guide.
if `buffer_size` field set to null explicitly in the network config it won't be modified and operator will be able to configure their own sizes. if requested buffer size can't be set (for example if 62500000 is higher than the max allowed) we will print error in the log.

it also sets this buffer to 1MB in two tests as the default buffer may not be sufficient, and in such case udp socket will silently drop incoming messages leading to a halted tests. the other option is to adjust payload, but that seems error prone

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -173,6 +173,7 @@ serde_json = "1.0"
 serial_test = "3.2.0"
 sha2 = "0.10"
 simple-xml-builder = "1.1"
+socket2 = "0.5.9"
 sorted_vector_map = "0.2.0"
 sorted-vec = "0.8.3"
 syn = { version = "1.0.0", features = ["full"] }
--- a/monad-dataplane/src/lib.rs
+++ b/monad-dataplane/src/lib.rs
@@ -6,8 +6,76 @@ use monoio::{IoUringDriver, RuntimeBuilder};
 use tokio::sync::{mpsc, mpsc::error::TrySendError};
 use tracing::warn;
 
+pub(crate) mod buffer_ext;
 pub mod tcp;
 pub mod udp;
+
+pub struct DataplaneBuilder {
+    local_addr: SocketAddr,
+    /// 1_000 = 1 Gbps, 10_000 = 10 Gbps
+    up_bandwidth_mbps: u64,
+    buffer_size: Option<usize>,
+}
+
+impl DataplaneBuilder {
+    pub fn new(local_addr: &SocketAddr, up_bandwidth_mbps: u64) -> Self {
+        Self {
+            local_addr: *local_addr,
+            up_bandwidth_mbps,
+            buffer_size: None,
+        }
+    }
+
+    /// with_buffer_size sets the buffer size for udp and tcp sockets that are managed by dataplane
+    /// to a requested value.
+    pub fn with_buffer_size(mut self, buffer_size: usize) -> Self {
+        self.buffer_size = Some(buffer_size);
+        self
+    }
+
+    pub fn build(self) -> Dataplane {
+        let DataplaneBuilder {
+            local_addr,
+            up_bandwidth_mbps,
+            buffer_size,
+        } = self;
+
+        let (tcp_ingress_tx, tcp_ingress_rx) = mpsc::channel(TCP_INGRESS_CHANNEL_SIZE);
+        let (tcp_egress_tx, tcp_egress_rx) = mpsc::channel(TCP_EGRESS_CHANNEL_SIZE);
+        let (udp_ingress_tx, udp_ingress_rx) = mpsc::channel(UDP_INGRESS_CHANNEL_SIZE);
+        let (udp_egress_tx, udp_egress_rx) = mpsc::channel(UDP_EGRESS_CHANNEL_SIZE);
+
+        thread::spawn(move || {
+            RuntimeBuilder::<IoUringDriver>::new()
+                .enable_timer()
+                .build()
+                .expect("Failed building the Runtime")
+                .block_on(async move {
+                    tcp::spawn_tasks(local_addr, tcp_ingress_tx, tcp_egress_rx, buffer_size);
+
+                    udp::spawn_tasks(
+                        local_addr,
+                        udp_ingress_tx,
+                        udp_egress_rx,
+                        up_bandwidth_mbps,
+                        buffer_size,
+                    );
+
+                    futures::future::pending::<()>().await;
+                });
+        });
+
+        Dataplane {
+            tcp_ingress_rx,
+            tcp_egress_tx,
+            udp_ingress_rx,
+            udp_egress_tx,
+
+            tcp_msgs_dropped: 0,
+            udp_msgs_dropped: 0,
+        }
+    }
+}
 
 pub struct Dataplane {
     tcp_ingress_rx: mpsc::Receiver<(SocketAddr, Bytes)>,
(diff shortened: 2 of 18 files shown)


omegablitz:
onboarding sysctls we recommend, for [reference](https://github.com/monad-client-docs/validator-docs/blob/6f290c72d3a80e609ce33d43aa539b6fb4d216a0/docs/onboarding/new-validators.md?plain=1#L171-L178):
```
sysctl -w net.core.rmem_default=62500000
sysctl -w net.core.rmem_max=62500000
sysctl -w net.core.wmem_default=62500000
sysctl -w net.core.wmem_max=62500000
sysctl -w net.ipv4.tcp_rmem="4096 62500000 62500000"
sysctl -w net.ipv4.tcp_wmem="4096 62500000 62500000"
```

I do think it makes sense to check the socket buffer sizes explicitly, to catch this sort of misconfiguration.

dshulyak:
makes sense, i was going to ask if they are raised for testnet. 
i don't have access to that page though.

i can extend this change to add both rw sizes to both tcp and udp sockets

Cargo.toml
```diff
@@ -174,6 +174,7 @@ serde_json = "1.0"
 serial_test = "3.2.0"
 sha2 = "0.10"
 simple-xml-builder = "1.1"
+socket2 = "0.5.9"
```
andr-dev, line 176: We are already using `0.5.7`, let's try and avoid updating existing dependencies when adding them from Cargo.lock to Cargo.toml.

monad-dataplane/src/udp.rs
```diff
@@ -230,3 +244,24 @@ fn is_eafnosupport(err: &Error) -> bool {
 
     err.len() >= EAFNOSUPPORT.len() && &err[0..EAFNOSUPPORT.len()] == EAFNOSUPPORT
 }
+
+trait UdpSSocketRecvBufferExt {
```
andr-dev: Is the extra `S` a typo?

monad-dataplane/src/udp.rs
```diff
@@ -230,3 +244,24 @@ fn is_eafnosupport(err: &Error) -> bool {
...
+    fn set_recv_buffer_size(&self, size: usize) -> io::Result<()>;
+    fn recv_buffer_size(&self) -> io::Result<usize>;
+}
+
+impl UdpSSocketRecvBufferExt for UdpSocket {
+    fn set_recv_buffer_size(&self, size: usize) -> io::Result<()> {
+        let sock = unsafe { socket2::Socket::from_raw_fd(self.as_raw_fd()) };
+        let rst = sock.set_recv_buffer_size(size);
+        _ = sock.into_raw_fd(); // defuse drop
```
andr-dev: This looks a bit confusing, probably because it's nuanced, but is the idea here that by turning the socket into a raw file descriptor we won't call `socket2`'s `Drop` which closes the socket?

monad-dataplane/src/lib.rs
```diff
@@ -9,6 +9,35 @@ use tracing::warn;
...
+        self
+    }
+
+    pub fn build(self) -> Dataplane {
+        Dataplane::new(
+            &self.local_addr,
+            self.up_bandwidth_mbps,
+            self.recv_buffer_size,
+        )
+    }
```
andr-dev, line 77: If we're going to introduce a builder, can we refactor the socket creation process so that if any of the buffer size calls fail, it panics here in the `build` rather than after Dataplane is created? Otherwise, this builder isn't really doing anything other than storing some values which then get passed to the Dataplane constructor.

monad-dataplane/src/udp.rs
andr-dev: Also what's the point of this trait? We introduce it and only use it once, seems redundant ...

monad-dataplane/src/lib.rs
dshulyak: it provides api to store optional arguments and passes them to private constructor. so doing nothing except passing was the goal

all constructor panics nowadays in the thread that runs io uring driver (e.g monoio UdpSocket can't be initialized and then sent to that thread). so it is probably a bit too much to refactor here

monad-dataplane/src/udp.rs
dshulyak: it is extension trait, allows to call this method on UdpSocket (which is type from another crate). a bit more typing to declare it but i don't see why is that an issue

monad-dataplane/src/udp.rs
dshulyak: yes

Cargo.toml
dshulyak: by using you mean that it was in cargo.lock? it is a dependency of some modules, but none of the modules were importing it directly, thats why i added

i personally don't see issues with updating minor version, usually projects setup dependabot that allows to merge updates when convenient but i certainly can revert to 0.5.7

monad-dataplane/src/udp.rs
andr-dev: Ah got it, missed that it's an extension.

dshulyak:
@andr-dev @omegablitz i extended this change and set buffers by default to 62_500_000, please take a look when available.

dshulyak:
unable to start node0

```
error: unexpected argument '--validators-path' found

  tip: a similar argument exists: '--triedb-path'

Usage: monad-node <--bls-identity <BLS_IDENTITY>|--secp-identity <SECP_IDENTITY>|--node-config <NODE_CONFIG>|--forkpoint-config <FORKPOINT_CONFIG>|--devnet-chain-config-override <DEVNET_CHAIN_CONFIG_OVERRIDE>|--genesis-path <GENESIS_PATH>|--wal-path <WAL_PATH>|--ledger-path <LEDGER_PATH>|--mempool-ipc-path <MEMPOOL_IPC_PATH>|--triedb-path <TRIEDB_PATH>|--control-panel-ipc-path <CONTROL_PANEL_IPC_PATH>|--statesync-ipc-path <STATESYNC_IPC_PATH>|--statesync-sq-thread-cpu <STATESYNC_SQ_THREAD_CPU>|--otel-endpoint <OTEL_ENDPOINT>|--keystore-password <KEYSTORE_PASSWORD>|--record-metrics-interval-seconds <RECORD_METRICS_INTERVAL_SECONDS>>

For more information, try '--help'.
```

monad-dataplane/src/tcp/tx.rs
```diff
@@ -199,6 +208,20 @@ async fn connect_and_send_messages(
             )
         })?;
 
+    if let Some(buffer_size) = buffer_size {
+        stream.set_send_buffer_size(buffer_size as usize)?;
+        let actual_buffer_size = stream.send_buffer_size()?;
+        if actual_buffer_size != buffer_size as usize {
```
bamrith: Is this check required ? Trying to understand when this can fail even after `set_send_buffer_size` returns `Ok()`. Same for `set_recv_buffer_size` above.

monad-dataplane/src/tcp/rx.rs
```diff
@@ -113,7 +119,26 @@ async fn task_connection(
     addr: SocketAddr,
     mut tcp_stream: TcpStream,
     tcp_ingress_tx: mpsc::Sender<(SocketAddr, Bytes)>,
+    buffer_size: Option<u64>,
 ) {
+    if let Some(buffer_size) = buffer_size {
+        tcp_stream
+            .set_recv_buffer_size(buffer_size as usize)
+            .expect("unable to set receive buffer size");
```
bamrith, line 127: are these panics caught by the worker thread ?

dshulyak: they are not. it was just a common pattern to panic in dataplane setup (here and in udp)

monad-dataplane/src/tcp/tx.rs
dshulyak: buffer size will be limited by the configured maximal buffer size. i opted for error instead of panic not to disrupt ops too much

monad-dataplane/src/tcp/rx.rs
```diff
@@ -113,7 +119,26 @@ async fn task_connection(
...
+        let actual_buffer_size = tcp_stream
+            .recv_buffer_size()
+            .expect("unable to get receive buffer size");
+        if actual_buffer_size != buffer_size as usize {
+            error!(
+                conn_id,
+                ?addr,
+                buffer_size,
+                actual_buffer_size,
+                "unable to set tcp receive buffer size"
```
omegablitz: I'm thinking that we should panic here so this error can't be ignored.

Perhaps can also add a hint to the error message to make sure sysctls are set properly.

monad-dataplane/src/lib.rs
```diff
@@ -6,9 +6,77 @@ use monoio::{IoUringDriver, RuntimeBuilder};
...
 
+pub(crate) mod buffer_ext;
 pub mod tcp;
 pub mod udp;
 
+pub struct DataplaneBuilder {
+    local_addr: SocketAddr,
+    /// 1_000 = 1 Gbps, 10_000 = 10 Gbps
+    up_bandwidth_mbps: u64,
+    buffer_size: Option<u64>,
```
omegablitz: nit: should we make this a usize? it needs to be casted everywhere downstream as is

dshulyak: updated

monad-dataplane/src/tcp/rx.rs
dshulyak: added hints and converted to panic

monad-testground/src/main.rs
```diff
@@ -291,6 +291,7 @@ where
                             local_addr: address.parse().unwrap(),
                             up_bandwidth_mbps: 1_000,
                             mtu: DEFAULT_MTU,
+                            buffer_size: Some(65_000_000),
```
michael-yxchen: nit: is this set different on purpose?

dshulyak: mistake, thanks for noticing

PR #1721, opened 2025-04-15
Implement statically-typed metrics system

andr-dev:
In monad-bft, we need to support two metrics use cases:

1. In production, we want the node process to use a global pool of static metrics which (in an ideal world) can also be optionally disabled completely with zero runtime cost (enabled by this PR, coming soon to monad-node).
2. In mock-swarm, we want to spin up multiple fake "nodes" in the same process but let them store distinct metrics. We leverage this in several tests where we assert that specific metrics are specific values on specific nodes.

Previously, we accomplished this by using the `ExecutorMetrics` abstraction which joins together multiple `HashMap`s to create a single `HashMap` with every metric in it. This has many downsides:

1. Metric names are not standardized and it is possible to accidentally introduce two metrics with the same namespace, which would lead to missing data.
2. If an executor forgets to report an initial value for a metric, the metric will report `no data` making it difficult in prometheus/grafana to tell if there's no data because the node is down or if it's because it was never set.
3. Executors which use tasks / threads / other async runtimes (dataplane) cannot store metrics in a native and easy-to-use way because `HashMap<&'static str, u64>` evidently does not implement `Send` or `Sync`.
4. Joining together multiple `HashMap`s is suboptimal.

This change introduces the `MetricsPolicy` trait which defines how metrics should be initialized for a given node. In production, the `StaticMetricsPolicy` enforces that `Counter`s and `Gauge`s are initialized from static AtomicU64s which can be cache aligned to improve performance. In mock swarm, the `MockMetricsPolicy` initializes `Counter`s and `Gauge`s using `Arc<AtomicU64>`. In all cases, the metric instruments are thread safe and thus can be cheaply shared with tasks / threads / other async runtimes enabling metrics everywhere (including dataplane, next PR!). Additionally, the `Counter` and `Gauge` traits make it so that state / executors do not need to think about what `MetricsPolicy` is being used and enforce semantics like `Counter`s cannot be decremented. This approach also makes it easy to add new instruments down the line like histograms.

To make this process simple for downstream consumers, `monad-metrics-macro` defines a simple to use macro that makes it easy to add/remove existing metrics while easily answering:

1. What metrics namespace is state / this executor in?
2. What counters, gauges, or sub-namespaces does state / this executor have?

Here's an example:
```rust
metrics_bft! {
    Ledger {
        #[counter]
        num_commits,

        #[counter]
        num_tx_commits,

        #[gauge]
        block_num
    }
}
```

This macro creates a struct with a field for each metric name, making metric storage statically-typed. This will help enable exciting new metrics-related features like automated alert config generation since we can do static analysis on them (aka add more codegen to the metrics macro).

This metrics system has been running smoothly on testnet for multiple weeks, and updated dashboards for these metrics can be found [here](https://grafana.devcore4.com/dashboards/f/eehx7co5jmcjke/metricsv2) including the highly-anticipated [dataplane dashboard](https://grafana.devcore4.com/d/eeibfi7jttssga/bft-dataplane) (split into a separate PR).


dshulyak:
can you share a bit more about what were the requirements for rolling out your own library for this? e.g reference to prior discussion would be also enough for me

i am mainly curious why using otel directly (or something on top of it) is a worse approach?

dshulyak:
thanks for the explanation.

i guess i don't follow why metrics were defined as HashMap<&'static str, u64> in the first place.
i think problems that you described do not exist if you just use otel types directly. ofcourse if you need to make some metrics disabled at startup thats a separate thing, but it is not a blocker either.

do you know what was the reason not to use otel types directly? for example if there will be a need to track latencies of some operations in bft, will you want to add your own histogram type for that?

UPD so for example archiver codebase is doing exactly that, with a small twist

andr-dev:
> i guess i don't follow why metrics were defined as HashMap<&'static str, u64> in the first place. i think problems that you described do not exist if you just use otel types directly. ofcourse if you need to make some metrics disabled at startup thats a separate thing, but it is not a blocker either.

They were originally made HashMap<&'static str, u64> because of the first problem I described above, where we need to be able to spin up multiple "nodes" within the same process, so using something in static memory wouldn't work. Additionally, we didn't want to commit to a metrics library prematurely and lock ourselves in, possibly introducing a massive refactor later down the line. As a quick workaround, we did something very primitive which has worked fairly well for the last little while. I've listed the benefits of moving away from this model above. Lastly, using otel metrics does not solve the other problems of standardized naming.

> do you know what was the reason not to use otel types directly? for example if there will be a need to track latencies of some operations in bft, will you want to add your own histogram type for that?

We're using the otel types but only in monad-bft which is where the external metrics instrumentation is instantiated. This way, state and executors don't need to depend on the exact instrumentation. As to why I didn't make an otelified Counter/Gauge struct in the `monad_metrics::instruments` module, the otel provided `Counter` and `Gauge` structs are very strict on how metric values are updated and wouldn't support something like `inc` or `dec_saturating` for `Gauge` without doing a read + write which can lead to problems when metrics are shared across the thread boundary.

monad-node/src/main.rs
```diff
@@ -543,27 +553,70 @@ fn resolve_domain_v4<P: PubKey>(node_id: &NodeId<P>, domain: &String) -> Option<
...
+        otel_counter.record(counter.read(), &[]);
+    };
+
+    let mut on_gauge = |name, gauge: &MP::Gauge| {
+        let otel_gauge = gauge_cache
+            .entry(name)
+            .or_insert_with(|| meter.u64_gauge(name).try_init().unwrap());
+
+        otel_gauge.record(gauge.read(), &[]);
+    };
```
dshulyak: there is an exportable metrics pattern (u64_exportable_gauge and so on) in otel crate. it allows to register a callback that will be called when the otel exporter runs. i think integration-wise it is much better than doing it manually.

one other thing is that metrics are usually defined with more context (description, units, labels). and specifically for histograms relevant buckets. one example is in monad consensus there are some errors counter, usually it would be exported as a single metric with multiple labels, so that it is very easy to track it and get more details if needed. basically what i want to point out is that there will be a need to map custom counters to exported metrics in a more flexible

monad-metrics/src/lib.rs
```diff
@@ -0,0 +1,35 @@
...
+
+mod instruments;
+
+pub trait MetricsPolicy: std::fmt::Debug + Unpin + Send + Sync + 'static {
+    type Counter: Counter;
+    type Gauge: Gauge;
+}
+
+#[derive(Debug)]
+pub struct StaticMetricsPolicy;
```
dshulyak: i think policy is totally unnecessary. overhead of Arc for metrics is negligible. I would suggest to profile that if you think that it will have non-negligible performance difference. the downside is that all codebase gets polluted with policy trait

one other option is to just have Arc defined for whole metrics struct rather than each metric individually, such as. so you will still have each atomic close to each other in memory, but again it is totally unnecessary

```rs
struct Metrics {
     counter_1: Counter<u64>,
     gauge_1: Gauge<u64>
}

let metrics = Arc<Metrics::default()>
```

and then just clone it wherever you want

andr-dev: I think the main rationale for adding the MetricsPolicy trait is to prevent subcrates from being tied to the specific metrics implementation. The alternative to not using MetricsPolicy is to add opentelemetry as a dependency to every crate which means building a single crate now requires compiling a large tree of effectively "unused" dependencies. As such, from the code pollution standpoint it seems like dependency pollution in this case is significantly worse than an additional trait.

Additionally, if we ever decide to switch from opentelemetry to something else down the line, it will require a massive refactor of metrics initialization + metrics usage, vs with the metrics policy nothing would change except `monad-node`.

dshulyak: what i suggested is to leave only LocalCounter/LocalGauge which are just wrappers around Arc AtomicU64 . libraries import those types and use them without any traits. there is no dependency on the opentelemtry types anywhere in this scheme

then you have a question, how to export those metrics. current approach iterates over all metrics and exports them in callbacks, so only binary that exports them have a dependency on opentelemetry. nothing changes here as well

so where is the problem with refactoring or dependency chains?

---

that said, i don't think that this is a solid approach for exporting metrics. just as an exercise try to define histogram with 100 buckets and export it using current approach. will you just copy whole thing everytime in the callback? what if there are 100 such histograms? 

i would expect monad-metrics crate to provide implementation for exporting to opentelemtry in a more native way, and that will require dependency on the opentelemtry. but it doesn't mean that every crate will have to be changed when exporter implementation changes

andr-dev:
After further thought, I agree that the MetricsPolicy trait isn't necessary. I originally added it so nodes could completely disable metrics but that doesn't seem like a significant value add for the amount of complexity it introduces. Pushed a change that removes it completely.

dshulyak:
i added several examples here https://github.com/category-labs/monad-bft/blob/7e35096ffea36e487d8cc0626e7f8d83db389d9a/monad-node/examples/metrics.rs to show how i would approach defining metrics so that we are not limited in using common metrics features and in general export them in a more "native" way

monad-consensus-types/src/metrics.rs
```diff
@@ -1,147 +1,90 @@
...
+                invalid_signature,
+                author_not_sender,
+                invalid_tc_round,
+                insufficient_stake,
+                invalid_seq_num,
+                val_data_unavailable,
+                invalid_vote_message,
+                invalid_version,
+                invalid_epoch
+            },
```
dshulyak: i would prefer to atleast somewhat invert this structure, so that we can have more flexibility.
for example you hardcoded "type" as a label key. that seems unfortunate limitation, it should be possible to specify it here, maybe:

```
validations_errors: group_counter{
    label: type,
    members: [],
}
```

later this can be extended to include help, as in

```
validations_errors: group_counter{
    label: type,
    help: "...",
    members: [],
}
```

---

one other alternative that seems superior both from UX and flexibility perspective is to use normal rust syntax with derive, rough example:

```
#[derive(MonadMetrics)]
struct Metrics {
     #[label = type, help = "", members = [

     ]]
     valiidation_errors: GroupCounter,
     
     #[namespace]
     consensus_events: ConsensusEvents,
}

#[derive(MonadMetrics)]
struct ConsensusEvents {
     local_timeout: Counter,
     handle_proposal: Counter
}

```

andr-dev: Agreed, wasn't too happy about attaching "type" to every label in monad-node and we can change this easily. I'll work something out and update!

For the "help", I was envisioning using standard "rust doc" syntax to keep things "familiar".
```rust
counters: [
  #[doc = "This is a description of the counter"]
  my_counter
]
```
Then, this would get attached to the otel description in monad-node. This will be included in a subsequent PR, likely right after the histogram PR since that one makes parsing attributes simple.

---

The main issue I saw with using a derive macro is that it's difficult to "namespace" things nicely because rust syntax doesn't allow you to build multi-level structures. For example, dataplane has an rx and tx side which, if you want to separate namespace wise, would be separate structs in the derive context.
```rust
#[derive(MonadMetrics)]
#[monad_metrics(namespace = "monad_bft.dataplane.tx")]
struct TxDataplaneMetrics {
  my_counter0: Counter,
  my_counter1: Counter,
  ...
}

#[derive(MonadMetrics)]
#[monad_metrics(namespace = "monad_bft.dataplane")]
struct DataplaneMetrics {
  tx: TxDataplaneMetrics,
  ...
}
```
Because the `DataplaneMetrics` struct macro isn't run in the same context as the dataplane one, it's difficult to enforce that the namespace for `TxDataplaneMetrics` is `monad_bft.dataplane.tx` making it more difficult to enforce that the parent namespace doesn't have any conflicts. The current syntax avoids this issue by having the entire metric tree in the same "context". The struct naming is also automatically built from the path to the root which is a nice to have but isn't a strict requirement.

That being said, having a more "rust" like syntax for familiarity reasons is a good north star and I have some ideas for something "in the middle" that might work nicely.

andr-dev: What do we think of the following syntax?
```rust
#[doc = "Invalid consensus message that is subsequently ignored"]
#[label(
    name = "error",
    variants = [
      invalid_author,
      not_well_formed_sig,
    ]
)]
validation_errors,
```

dshulyak: with this syntax it is easier for me to understand how it resolves, so i like it.
maybe also put a field type (counter, gauge, histogram) into parameters

```
#[
  counter( 
    name = "error", 
    variants = [ 
       invalid_author, 
       not_well_formed_sig, ] 
   )
] 
validation_errors,
```

andr-dev: I originally went for the `counters: [ ... ]` syntax so you wouldn't have to repeat counter multiple times, but I see how having the tag in the attributes makes it easier to read so happy to make the change. As a small modification, I think "name" is confusing here because the metric is actually called "validation_errors". How about the following:
```rust
#[doc = "Other metric description"]
#[counter/gauge]
other_metric,

#[doc = "Invalid consensus message that is subsequently ignored"]
#[counter/gauge(
    label = "error",
    variants = [
      invalid_author,
      not_well_formed_sig,
    ]
)]
validation_errors,

#[doc = "Elapsed description"]
#[histogram(buckets = [1_000, 25_000, 100_000, ...])]
elapsed
```
The document attribute and histogram support will be split out into separate PRs to keep this one as short as possible, so this PR will include the `#[counter/gauge]` and it's labeled version. Does this look good?

dshulyak: looks good

andr-dev:
Macro syntax has been updated, slightly more verbose but hopefully easier to follow.

JoeHowarth:
I don't see any glaring bugs, but the system feels overly complex. 
Pain points I see:
- declarative macro makes it unclear what code is actually generated and what that code does
- Separation of otel reporting plumbing from metrics crates means system is spread across 2 code bases and harder to reuse across other services (rpc, archive, txgen, etc.). Consider rolling the reporting pipeline w/ observers into a shared setup module
- Dynamic attributes not supported and non-trivial to extend without knowledge of declarative macros and ramp up time reading the plumbing of monad-metrics-macros. This is used in archive metrics reporting 

It's unclear why we need new types on top of base otel Counter, Gauge, Histogram, etc. 
The only usage where we need to subtract or decrement appears to be in `block_sync.rs`:
```rust
                self.metrics
                    .self_payload_requests_in_flight
                    .sub_saturating(1);
```


Have you considered using something like:

```rust
trait MetricStr: EnumIter {
  fn as_static_str() -> &'static str
}

#[derive(Clone)]
pub struct Metrics<Counters, Gauges> {
     counters: HashMap<Counters, Counter<u64>>,
     gauges: HashMap<Gauges, Gauge<u64>>,
}

impl Metrics<Counters: MetricStr, Gauges: MetricStr> {
   pub fn new() -> Self {
       Self { 
           counters: HashMap::from_iter(Counters::iter()
                        .map(|c| (c.as_static_str(), Counter::new())), 
           gauges: ... 
       }
   }
   pub fn gauge(&mut self, name: Gauges) -> &mut Gauge<u64> { ... }
   pub fn counter(...) -> &mut Counter<u64> {..}
   pub fn inc_counter(&mut self, name: Counters) {..}
   ... etc, other helper methods as needed
}
```

Then for 

```rust
metrics_bft! {
    Ledger {
        #[counter]
        num_commits,

        #[counter]
        num_tx_commits,

        #[gauge]
        block_num
    }
}
```

Becomes 

```rust
#[derive(EnumIter)] // possibly have a simple derive macro to impl MetricStr
pub(crate) enum LedgerCounters {
  NumCommits,
  NumTxCommits,
}

impl MetricStr for LedgerMetricNames {
   fn as_static_str(&self) -> ... {
       match self {
            NumCommits => "num_commits" // or whatever 
       }
   }
}

#[derive(EnumIter, MetricStr)] // possibly have a simple derive macro to impl MetricStr
pub(crate) enum LedgerGauges {
   BlockNum
}

pub(crate) type LedgerMetrics = Metrics<LedgerCounters, LedgerGauges>;

// Usage:
fn foo(metrics: LedgerMetrics) {
    ...
    metrics.counter(LedgerCounters::NumCommits).add(1, &[]);
    // or with helper:
    metrics.inc_counter(LedgerCounters::NumCommits));
}
```

This type of approach is vastly simpler and easier to understand / reason about, and to my knowledge can do everything that the approach implemented in this library can do. 

tl;dr:
- Why do we *need* this complexity? 
- Why are otel types separated from monad-metrics? 
- Why is reporting pipeline boiler plate not co-located with metric library?

PR #1727, opened 2025-04-17
update otel to latest (0.29) versions together with related deps

dshulyak:
in this version exporters do not rely on the tokio runtime to send data, 
they were refactored to send it in blocking way on a dedicated thread.

change can be reviewed, it is all about changes in the interfaces, but i will
merge it after next release


dshulyak:
thanks, i assumed that merge queue will squash and use pr message a commit instead

PR #1733, opened 2025-04-22
Prune nodes periodically

eerkaijun:
addresses https://github.com/category-labs/category-internal/issues/1514

### DIFF
--- a/monad-peer-disc-swarm/tests/ping_pong.rs
+++ b/monad-peer-disc-swarm/tests/ping_pong.rs
@@ -60,7 +60,7 @@ fn test_ping_pong() {
         .map(|k| {
             (NodeId::new(k.pubkey()), PeerInfo {
                 last_ping: None,
-                last_seen: None,
+                unresponsive_pings: 0,
                 name_record: generate_name_record(k),
             })
         })
@@ -69,18 +69,29 @@ fn test_ping_pong() {
         builders: keys
             .iter()
             .enumerate()
-            .map(|(i, key)| NodeBuilder {
-                id: NodeId::new(key.pubkey()),
-                algo_builder: PeerDiscoveryBuilder {
-                    self_id: NodeId::new(key.pubkey()),
-                    self_record: generate_name_record(key),
-                    peer_info: all_peers.clone(),
-                    ping_period: Duration::from_secs(1),
-                    rng_seed: 123456,
-                },
-                router_scheduler: NoSerRouterConfig::new(all_peers.keys().cloned().collect())
-                    .build(),
-                seed: i.try_into().unwrap(),
+            .map(|(i, key)| {
+                let self_id = NodeId::new(key.pubkey());
+                let peer_info = all_peers
+                    .clone()
+                    .into_iter()
+                    .filter(|(id, _)| id != &self_id)
+                    .collect::<BTreeMap<_, _>>();
+                NodeBuilder {
+                    id: NodeId::new(key.pubkey()),
+                    algo_builder: PeerDiscoveryBuilder {
+                        self_id,
+                        self_record: generate_name_record(key),
+                        peer_info,
+                        ping_period: Duration::from_secs(2),
+                        prune_period: Duration::from_secs(4),
+                        request_timeout: Duration::from_secs(1),
+                        prune_threshold: 3,
+                        rng_seed: 123456,
+                    },
+                    router_scheduler: NoSerRouterConfig::new(all_peers.keys().cloned().collect())
+                        .build(),
+                    seed: i.try_into().unwrap(),
+                }
             })
             .collect(),
         seed: 7,
@@ -90,13 +101,13 @@ fn test_ping_pong() {
 
     while nodes.step_until(Duration::from_secs(10)) {}
 
-    // first ping is sent out at t=0. we expect >=10 at t=10
+    // first ping is sent out at t=0. we expect 6 at t=10 (inclusive of t=0 and t=10)
     for state in nodes.states().values() {
         let metrics = state.peer_disc_driver.get_peer_disc_state().metrics();
-        assert!(metrics["send_ping"] >= 10);
-        assert!(metrics["send_pong"] >= 10);
-        assert!(metrics["recv_ping"] >= 10);
-        assert!(metrics["recv_pong"] >= 10);
+        assert!(metrics["send_ping"] == 6);
+        assert!(metrics["send_pong"] == 6);
+        assert!(metrics["recv_ping"] == 6);
+        assert!(metrics["recv_pong"] == 6);
     }
 }
 
@@ -110,16 +121,16 @@ fn test_new_node_joining() {
     let (bootstrap_keys, third_key) = (&keys[0..2], &keys[2]);
 
     // initialize peer info of the three nodes
-    // NodeA name record: NodeA, NodeB
-    // NodeB name record: NodeA, NodeB
-    // NodeC name record: NodeA, NodeB, NodeC
+    // NodeA name record: NodeB
+    // NodeB name record: NodeA,
+    // NodeC name record: NodeA, NodeB
     // (we can assume that NodeA and NodeB are the bootstrap nodes in a network)
     let bootstrap_peers: BTreeMap<NodeId<PubKeyType>, PeerInfo<SignatureType>> = bootstrap_keys
         .iter()
         .map(|k| {
             (NodeId::new(k.pubkey()), PeerInfo {
                 last_ping: None,
-                last_seen: None,
+                unresponsive_pings: 0,
                 name_record: generate_name_record(k),
             })
         })
@@ -129,7 +140,7 @@ fn test_new_node_joining() {
         .map(|k| {
             (NodeId::new(k.pubkey()), PeerInfo {
                 last_ping: None,
-                last_seen: None,
+                unresponsive_pings: 0,
                 name_record: generate_name_record(k),
             })
         })
@@ -138,22 +149,34 @@ fn test_new_node_joining() {
         builders: keys
             .iter()
             .enumerate()
-            .map(|(i, key)| NodeBuilder {
-                id: NodeId::new(key.pubkey()),
-                algo_builder: PeerDiscoveryBuilder {
-                    self_id: NodeId::new(key.pubkey()),
-                    self_record: generate_name_record(key),
-                    peer_info: if key.pubkey() == third_key.pubkey() {
-                        all_peers.clone()
-                    } else {
-                        bootstrap_peers.clone()
+            .map(|(i, key)| {
+                let self_id = NodeId::new(key.pubkey());
+                let base_peers = if key.pubkey() == third_key.pubkey() {
+                    &all_peers
+                } else {
+                    &bootstrap_peers
+                };
+                let peer_info = base_peers
+                    .clone()
+                    .into_iter()
+                    .filter(|(id, _)| id != &self_id)
+                    .collect::<BTreeMap<_, _>>();
+                NodeBuilder {
+                    id: NodeId::new(key.pubkey()),
+                    algo_builder: PeerDiscoveryBuilder {
+                        self_id,
+                        self_record: generate_name_record(key),
+                        peer_info,
+                        ping_period: Duration::from_secs(2),
+                        prune_period: Duration::from_secs(4),
+                        request_timeout: Duration::from_secs(1),
+                        prune_threshold: 3,
+                        rng_seed: 123456,
                     },
-                    ping_period: Duration::from_secs(1),
-                    rng_seed: 123456,
-                },
-                router_scheduler: NoSerRouterConfig::new(all_peers.keys().cloned().collect())
-                    .build(),
-                seed: i.try_into().unwrap(),
+                    router_scheduler: NoSerRouterConfig::new(all_peers.keys().cloned().collect())
+                        .build(),
+                    seed: i.try_into().unwrap(),
+                }
             })
             .collect(),
         seed: 7,
@@ -165,6 +188,9 @@ fn test_new_node_joining() {
     for (node_id, state) in nodes.states() {
         let state = state.peer_disc_driver.get_peer_disc_state();
         for node_id in all_peers.keys() {
+            if node_id == &state.self_id {
+                continue;
+            }
             assert!(state.peer_info.contains_key(node_id));
         }
 
@@ -174,6 +200,8 @@ fn test_new_node_joining() {
             assert!(metrics["recv_lookup_request"] == 2);
             assert!(metrics["send_lookup_response"] == 2);
         }
+
+        assert!(state.outstanding_lookup_requests.is_empty());
     }
 }
 
@@ -190,7 +218,7 @@ fn test_update_name_record() {
         .map(|k| {
             (NodeId::new(k.pubkey()), PeerInfo {
                 last_ping: None,
-                last_seen: None,
+                unresponsive_pings: 0,
                 name_record: generate_name_record(k),
             })
         })
@@ -199,18 +227,29 @@ fn test_update_name_record() {
         builders: keys
             .iter()
             .enumerate()
-            .map(|(i, key)| NodeBuilder {
-                id: NodeId::new(key.pubkey()),
-                algo_builder: PeerDiscoveryBuilder {
-                    self_id: NodeId::new(key.pubkey()),
-                    self_record: generate_name_record(key),
-                    peer_info: all_peers.clone(),
-                    ping_period: Duration::from_secs(1),
-                    rng_seed: 123456,
-                },
-                router_scheduler: NoSerRouterConfig::new(all_peers.keys().cloned().collect())
-                    .build(),
-                seed: i.try_into().unwrap(),
+            .map(|(i, key)| {
+                let self_id = NodeId::new(key.pubkey());
+                let peer_info = all_peers
+                    .clone()
+                    .into_iter()
+                    .filter(|(id, _)| id != &self_id)
+                    .collect::<BTreeMap<_, _>>();
+                NodeBuilder {
+                    id: NodeId::new(key.pubkey()),
+                    algo_builder: PeerDiscoveryBuilder {
+                        self_id,
+                        self_record: generate_name_record(key),
+                        peer_info,
+                        ping_period: Duration::from_secs(2),
+                        prune_period: Duration::from_secs(4),
+                        request_timeout: Duration::from_secs(1),
+                        prune_threshold: 3,
+                        rng_seed: 123456,
+                    },
+                    router_scheduler: NoSerRouterConfig::new(all_peers.keys().cloned().collect())
+                        .build(),
+                    seed: i.try_into().unwrap(),
+                }
             })
             .collect(),
         seed: 7,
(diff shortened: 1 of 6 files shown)


monad-peer-discovery/src/discovery.rs
```diff
@@ -27,14 +28,20 @@ pub struct PeerInfo<ST: CertificateSignatureRecoverable> {
     pub name_record: MonadNameRecord<ST>,
 }
 
+#[derive(Debug, Clone, Copy)]
+pub struct LookupInfo<ST: CertificateSignatureRecoverable> {
+    pub timestamp: Instant,
```
michael-yxchen: Instants might not work well with swarm testing because we're using virtual times, passed in as Duration

For unresponsive peers, we can probably prune by # of unresponded pings. Outstanding lookups, maybe we can schedule a timer for them as well?

monad-peer-discovery/src/discovery.rs
```diff
@@ -27,14 +28,20 @@ pub struct PeerInfo<ST: CertificateSignatureRecoverable> {
...
 pub struct PeerDiscovery<ST: CertificateSignatureRecoverable> {
     pub self_id: NodeId<CertificateSignaturePubKey<ST>>,
     pub self_record: MonadNameRecord<ST>,
     pub peer_info: BTreeMap<NodeId<CertificateSignaturePubKey<ST>>, PeerInfo<ST>>,
-    // maps lookup id => receiver of the request
-    pub outstanding_lookup_requests: HashMap<u32, NodeId<CertificateSignaturePubKey<ST>>>,
+    pub outstanding_lookup_requests: HashMap<u32, LookupInfo<ST>>,
     pub metrics: PeerDiscMetrics,
     pub ping_period: Duration,
+    pub retain_period: Duration,
```
michael-yxchen: `retain_period` needs to be a multiple (3x or more) of `ping_period`. I think pruning by number of unresponded pings is probably cleaner

monad-peer-discovery/src/discovery.rs
```diff
@@ -332,7 +357,29 @@ where
         cmds
     }
 
-    // TODO: record metrics when handling event
+    fn prune(&mut self) -> Vec<PeerDiscoveryCommand<ST>> {
```
michael-yxchen, line 496: something to think about - do we want to treat validators and dedicated/prioritized/public full nodes separately?

monad-peer-discovery/src/discovery.rs
michael-yxchen, line 1: we should add unit tests for `prune`

monad-peer-disc-swarm/tests/ping_pong.rs
```diff
@@ -283,6 +318,61 @@ fn test_update_name_record() {
...
+
+    let mut nodes = swarm_builder.build();
+
+    // retain period is set to 3 seconds (i.e. prune every 3 seconds)
+    while nodes.step_until(Duration::from_secs(6)) {}
+
+    for state in nodes.states().values() {
+        let metrics = state.peer_disc_driver.get_peer_disc_state().metrics();
+        assert!(metrics["prune"] == 2);
+    }
```
michael-yxchen: Limitations of using `Instant`, we can't test that offline nodes are actually getting pruned

monad-peer-discovery/src/discovery.rs
michael-yxchen: we'll need to retry lookups, so timers are helpful there

eerkaijun: added

monad-peer-discovery/src/discovery.rs
eerkaijun: added

monad-peer-disc-swarm/tests/ping_pong.rs
eerkaijun: removed the use of Instant and switched to Duration

monad-peer-discovery/src/discovery.rs

monad-peer-discovery/src/discovery.rs
eerkaijun: good point, I think validators should not be pruned as easily as full nodes. right now we don't differentiate between the type of nodes in the peer discovery state yet

monad-peer-discovery/src/discovery.rs
```diff
@@ -125,10 +180,14 @@ where
         };
 
         // record ping in peer_info
+        if peer_entry.last_ping.is_some() {
+            // previous ping has not been fulfilled
+            peer_entry.unresponsive_pings += 1;
```
michael-yxchen: we should increment this counter on timeout, not on first send. Imagine we dump peer discovery state for debug purpose. `unresponsive_ping = 1` for live peer can confuse people

monad-peer-discovery/src/discovery.rs
```diff
@@ -212,18 +271,40 @@ where
...
+            None => {
+                let lookup_id = self.rng.next_u32();
+                self.outstanding_lookup_requests
+                    .insert(lookup_id, LookupInfo {
+                        num_retries: 0,
+                        receiver: to,
+                    });
+                lookup_id
+            }
         };
```
michael-yxchen: we should generate a new lookup_id when previous one times out. `num_retries` should accumulate

monad-peer-discovery/src/lib.rs
michael-yxchen, line 1: I have different opinions on how timeouts should be handled

we should add `handle_ping_timeout(target)` and `handle_peer_lookup_timeout(lookup_id)` to `PeerDiscoveryAlgo`. Existing impl merges timeout path with normal code path. It can cause confusion, e.g. `unresponsive_ping` is 1 for a new peer

those `handle_X_timeout` methods can be very simple - bookkeeping, prune nodes/switch servicer if above threshold, then call into normal send message code paths

we can eliminate `prune` event. All the pruning is done on timeouts, instead of perdiocally

monad-peer-discovery/src/discovery.rs
```diff
@@ -212,18 +271,40 @@ where
...
 
-        // record request
-        self.outstanding_lookup_requests
-            .insert(peer_lookup_request.lookup_id, to);
+        let peer_lookup_request = PeerLookupRequest::<ST> { lookup_id, target };
+
+        // schedule for peer lookup retry
+        cmds.extend(self.reset_lookup_timer(to, target, lookup_id));
```
michael-yxchen: if there's a pending peer_lookup to the same servicer/to, sending another peer lookup cancels previous timer. This is undesirable behavior

monad-peer-discovery/src/discovery.rs
```diff
@@ -212,18 +271,40 @@ where
...
+                } else {
+                    warn!("lookup id not found in outstanding lookup requests");
+                    return cmds;
+                }
+                lookup_id
+            }
+            // new lookup request
+            None => {
+                let lookup_id = self.rng.next_u32();
+                self.outstanding_lookup_requests
```
michael-yxchen: we should generate lookup_ids such that they don't collide

monad-peer-discovery/src/discovery.rs
eerkaijun: this counter is only incremented on timeout - when there's a previous ping that we have not receive a pong for. `unresponsive_pings` is set as 0 on first send

monad-peer-discovery/src/lib.rs
eerkaijun: `unresponsive_ping` is 0 for a new peer. Right now the logic is to have periodic pings and increment `unresponsive_ping` if previous ping has not been fulfilled when we are ready send the next ping (so the ping period and ping timeout is the same value). Not opposed to separating out ping timeouts but that means we'll use two separate timers for sending pings and recording ping timeouts.

Agree on peer lookup timeout, since it's not a periodic process and probably much cleaner to separate initiation and retries. 

I think we should still keep the `prune` event though. Right now our peer discovery naively pruned nodes based on number of consecutive unresponsive pings so it's straightforward, but for e.g. in libp2p peers are ranked based on criterias like duration of connection etc and pruned periodically, so I think in general we should still have pruning logic done periodically rather than on timeouts.

monad-peer-discovery/src/discovery.rs
eerkaijun: actually what's the motivation for generating a new lookup_id during retries? I feel like using the same id is cleaner, since I think the only purpose of the lookup id is to make sure we only entertain PeerLookupResponse where we initiate the request (to prevent DOS), so with the same request I think it doesn't matter if the lookup_id is the same

monad-peer-discovery/src/discovery.rs
michael-yxchen: got it. make sense

monad-peer-discovery/src/discovery.rs
eerkaijun: very low probability for lookup id collision in `next_u32`, i think a random id is better than for e.g. an monotically increasing value because otherwise it will be easy for byzantine nodes to guess and send PeerLookupResponse with matching lookup_id

monad-peer-discovery/src/discovery.rs
michael-yxchen: on a high level, lookup_id is the unique identifier for a request/response pair. If client times out a request, I think it should consider that request dead and stop accepting response with that id

monad-peer-discovery/src/discovery.rs
michael-yxchen: agree that probability of collision is low. it's better to prevent it altogether when we can. We don't need to use monotonically increasing values. Can generate next value if previous id is taken

monad-peer-discovery/src/lib.rs
michael-yxchen: cool. Agree that we can keep `prune` event

> we'll use two separate timers for sending pings and recording ping timeouts.

logically those are different - send period is much larger than timeout. Minor implementation detail: for pong coming after timeout, we can still accept but mark it late

monad-peer-discovery/src/discovery.rs
michael-yxchen: https://github.com/category-labs/category-internal/issues/1546

monad-peer-discovery/src/discovery.rs
eerkaijun: ok sure updated

monad-peer-discovery/src/lib.rs
eerkaijun: separated ping timeout and peer lookup timeout

monad-peer-discovery/src/discovery.rs
```diff
@@ -15,26 +15,36 @@ use tracing::{debug, warn};
...
     pub self_id: NodeId<CertificateSignaturePubKey<ST>>,
     pub self_record: MonadNameRecord<ST>,
     pub peer_info: BTreeMap<NodeId<CertificateSignaturePubKey<ST>>, PeerInfo<ST>>,
-    // maps lookup id => receiver of the request
-    pub outstanding_lookup_requests: HashMap<u32, NodeId<CertificateSignaturePubKey<ST>>>,
+    pub outstanding_lookup_requests: HashMap<u32, LookupInfo<ST>>,
     pub metrics: PeerDiscMetrics,
     pub ping_period: Duration,
+    pub prune_period: Duration,
+    pub timeout_period: Duration,
```
michael-yxchen: nit: rename to message/request_timeout

monad-peer-discovery/src/discovery.rs
```diff
@@ -64,36 +77,84 @@ impl<ST: CertificateSignatureRecoverable> PeerDiscoveryAlgoBuilder for PeerDisco
...
+                duration: self.timeout_period,
+                on_timeout: PeerDiscoveryEvent::PingTimeout { to: peer, ping_id },
+            }),
+            PeerDiscoveryCommand::TimerCommand(PeerDiscoveryTimerCommand::Schedule {
+                node_id: peer,
+                timer_kind: TimerKind::SendPing,
                 duration: self.ping_period,
                 on_timeout: PeerDiscoveryEvent::SendPing { to: peer },
             }),
         ]
```
michael-yxchen: we can either do `ScheduleReset` and `Schedule` for all timers, or rely on timer implementation to auto-reset

monad-peer-discovery/src/discovery.rs
```diff
@@ -64,36 +77,84 @@ impl<ST: CertificateSignatureRecoverable> PeerDiscoveryAlgoBuilder for PeerDisco
...
+        ping_id: u32,
     ) -> Vec<PeerDiscoveryCommand<ST>> {
         vec![
             PeerDiscoveryCommand::TimerCommand(PeerDiscoveryTimerCommand::ScheduleReset {
                 node_id: peer,
+                timer_kind: TimerKind::SendPing,
             }),
             PeerDiscoveryCommand::TimerCommand(PeerDiscoveryTimerCommand::Schedule {
                 node_id: peer,
+                timer_kind: TimerKind::PingTimeout { ping_id },
```
michael-yxchen: are we expected to schedule two timers to the same target with different `ping_id`s?

monad-peer-discovery/src/discovery.rs
```diff
@@ -332,7 +427,86 @@ where
...
+                lookup_info.num_retries + 1
+            }
+            None => {
+                debug!("lookup id not found in outstanding lookup requests");
+                return cmds;
+            }
+        };
+
+        // generate new unique lookup id
+        self.outstanding_lookup_requests.remove(&lookup_id);
```
michael-yxchen: this can be simplified with something like
```
let Some(lookup_info) = self.outstanding_lookup_requests.get_mut(&lookup_id) else {
    // not found
    return cmds;
}
...
```

monad-peer-discovery/src/discovery.rs
```diff
@@ -332,7 +427,86 @@ where
...
+        while self
+            .outstanding_lookup_requests
+            .contains_key(&new_lookup_id)
+        {
+            new_lookup_id = self.rng.next_u32();
+        }
+        self.outstanding_lookup_requests
+            .insert(new_lookup_id, LookupInfo {
+                num_retries,
+                receiver: target,
```
michael-yxchen, line 475: note for later, retry with different servers

monad-peer-discovery/src/discovery.rs
```diff
@@ -332,7 +427,86 @@ where
...
+
+    fn prune(&mut self) -> Vec<PeerDiscoveryCommand<ST>> {
+        debug!("pruning unresponsive nodes");
+
+        *self.metrics.entry("prune").or_default() += 1;
+        let mut cmds = Vec::new();
+
+        // prune unresponsive nodes
+        self.peer_info
+            .retain(|_, info| info.unresponsive_pings < self.prune_threshold);
```
michael-yxchen, line 504: should we move this to `handle_ping_timeout`?

monad-peer-discovery/src/discovery.rs
```diff
@@ -417,19 +591,22 @@ mod tests {
             outstanding_lookup_requests: HashMap::new(),
             metrics: HashMap::new(),
             ping_period: Duration::from_secs(60),
+            prune_period: Duration::from_secs(120),
+            timeout_period: Duration::from_secs(30),
```
michael-yxchen: nit: i understand that this is a test, but 30 second timeout is too long

monad-peer-disc-swarm/tests/ping_pong.rs
```diff
@@ -90,13 +101,13 @@ fn test_ping_pong() {
 
     while nodes.step_until(Duration::from_secs(10)) {}
 
-    // first ping is sent out at t=0. we expect >=10 at t=10
+    // first ping is sent out at t=0. we expect 6 at t=10 (inclusive of t=0 amd t=10)
```
michael-yxchen: typo
```suggestion
    // first ping is sent out at t=0. we expect 6 at t=10 (inclusive of t=0 and t=10)
```

monad-peer-discovery/src/discovery.rs
```diff
@@ -64,36 +77,84 @@ impl<ST: CertificateSignatureRecoverable> PeerDiscoveryAlgoBuilder for PeerDisco
...
+    fn schedule_lookup_timer(
+        &self,
+        to: NodeId<CertificateSignaturePubKey<ST>>,
+        target: NodeId<CertificateSignaturePubKey<ST>>,
+        lookup_id: u32,
+    ) -> Vec<PeerDiscoveryCommand<ST>> {
+        vec![PeerDiscoveryCommand::TimerCommand(
+            PeerDiscoveryTimerCommand::Schedule {
+                node_id: to,
+                timer_kind: TimerKind::RetryPeerLookup,
```
michael-yxchen: would it make sense to include lookup_id RetryPeerLookup? so we can send more than lookups to a peer

eerkaijun: good point, originally had that but got lost in rebase, updated

monad-peer-discovery/src/discovery.rs
eerkaijun: as I discussed i think pruning logic of nodes should be done on prune event, as eventually we are likely to have different criterias to prune nodes that require scanning the peer_info map

monad-peer-discovery/src/discovery.rs
eerkaijun: this is for the unlikely event where ping period < timeout period

monad-peer-discovery/src/discovery.rs
eerkaijun: do we need ScheduleReset for ping timeout?

monad-peer-discovery/src/discovery.rs
michael-yxchen: we can reject such cases on init. It doesn't sound too useful

michael-yxchen: let's remove ping_id from TimerKind::PingTimeout

monad-peer-discovery/src/discovery.rs
michael-yxchen: technically you don't because `timeout << period`, but I think always resetting is less error prone

eerkaijun: added

monad-peer-discovery/src/discovery.rs

monad-peer-discovery/src/discovery.rs
```diff
@@ -332,7 +431,82 @@ where
...
+            message: PeerDiscoveryMessage::PeerLookupRequest(peer_lookup_request),
+        });
+
+        *self.metrics.entry("retry_lookup_request").or_default() += 1;
+
+        cmds
+    }
+
+    fn prune(&mut self) -> Vec<PeerDiscoveryCommand<ST>> {
+        debug!("pruning unresponsive nodes");
```
andr-dev: Can we update this comment to make it clear it's coming from peer discovery?

monad-peer-discovery/src/discovery.rs
```diff
@@ -64,36 +77,88 @@ impl<ST: CertificateSignatureRecoverable> PeerDiscoveryAlgoBuilder for PeerDisco
             outstanding_lookup_requests: Default::default(),
             metrics: Default::default(),
             ping_period: self.ping_period,
+            prune_period: self.prune_period,
+            request_timeout: self.request_timeout,
```
michael-yxchen, line 83: nit: let's assert `ping_period > request_timeout` since we removed ping_id from ping timers

PR #1737, opened 2025-04-23
peer-disc-swarm: router message using IP addr

michael-yxchen:


michael-yxchen:
Tests are failing: blocked on https://github.com/category-labs/category-internal/issues/1513

monad-peer-disc-swarm/src/lib.rs
```diff
@@ -394,10 +429,34 @@ where
...
+                // Find the target node with routing_table, which is the ground
+                // truth for name records
+                let Some(target_node) = self.routing_table.route(&peer_addr) else {
+                    debug!(to=?message.to.get_peer_id(), addr=?peer_addr, "routing failed: no node found with addr");
+                    continue;
+                };
+
+                if message.to.get_peer_id() != target_node {
+                    debug!("local peer table outdated");
+                }
```
eerkaijun: what's the scenario where this will happen?

michael-yxchen: `message.to` is the desired logical target. `target_node` is resolved using sender's peer table, then global routing table registry

say A and B swap their IPs, then C will send message meant for A to B, and vice versa, until it's local table is updated

PR #1738, opened 2025-04-23
Concurrent eth_call request updates

eclark0426:
- Replaces the concurrent call CLI arg with using the same value as the number of fibers in the executor
- Adds an opt-in RPC method to check how many slots are available for eth_call requests

### DIFF
--- a/monad-rpc/src/call.rs
+++ b/monad-rpc/src/call.rs
@@ -1,4 +1,12 @@
-use std::{cmp::min, sync::Arc};
+use std::{
+    cmp::min,
+    collections::HashMap,
+    sync::{
+        atomic::{AtomicU64, Ordering},
+        Arc,
+    },
+    time::{Duration, Instant},
+};
 
 use alloy_consensus::{Header, SignableTransaction, TxEip1559, TxEnvelope, TxLegacy};
 use alloy_primitives::{Address, PrimitiveSignature, TxKind, Uint, U256, U64, U8};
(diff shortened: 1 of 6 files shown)


monad-rpc/src/call.rs
```diff
@@ -361,6 +363,13 @@ pub async fn sender_gas_allowance<T: Triedb>(
     }
 }
 
+#[allow(non_snake_case)]
+pub async fn monad_admin_ethCallCapacity(
+    eth_call_executor: Arc<Mutex<EthCallExecutor>>,
+) -> JsonRpcResult<u32> {
+    Ok(eth_call_available_capacity(eth_call_executor).await)
+}
```
eerkaijun: who are we targeting to be the user of this endpoint?

eclark0426: I think it's primarily intended for RPC providers

monad-rpc/src/call.rs
```diff
@@ -18,9 +23,94 @@ use crate::{
...
     trace_handlers::{decode_call_frame, MonadCallFrame, TracerObject},
 };
 
+#[derive(Debug)]
+struct EthCallRequestStats {
+    entry_time: Instant,
+}
+
+#[derive(Debug, Clone, Default)]
+struct CumulativeStats {
```
mikety, line 39: This is a good candidate to be tracked with [metrics](https://github.com/category-labs/monad-bft/blob/master/monad-consensus-types/src/metrics.rs) .

monad-rpc/src/call.rs
```diff
@@ -18,9 +23,94 @@ use crate::{
...
+    total_requests: u64,
+    total_errors: u64,
+    queue_rejections: u64,
+}
+
+#[derive(Debug)]
+pub struct EthCallStatsTracker {
+    active_requests: Arc<Mutex<HashMap<RequestId, EthCallRequestStats>>>,
```
mikety, line 67: Updating `cumulative_stats` and `active_requests` will cause lock contention if there are enough concurrent requests. One way to improve is to use Atomic counters instead for CumulativeStats.

monad-rpc/src/main.rs
```diff
@@ -268,20 +293,42 @@ async fn rpc_select(
...
+                        "eth_call concurrent requests limit".into(),
+                    ));
+                }
+            };
 
             let params = serde_json::from_value(params).invalid_params()?;
-            monad_eth_call(
+
+            app_state
+                .eth_call_stats_tracker
```
mikety: calls to `eth_call_stats_tracker` should be conditioned ( or noop ) to `enable_eth_call_statistics`. There should be no overhead when `enable_eth_call_statistics` is `false`.

eclark0426:
Yes, I can work on those later today after finishing up #1752. For now I'll merge this to make sure it gets in before the release

PR #1739, opened 2025-04-24
monad-rpc: refactor dispatch loop and add histogram to track latency by method

dshulyak:
splitted rpc_select into separate functions with macro-generated enum. primarily so that i can define span with &'static method name in a single place. 

enables latency distribution per method

image

rate per method

image

and allows to determine if any time spent on cpu in actix handlers. wait time here means waiting on a database execution

image

will add that as dashboards once merged

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5183,6 +5183,7 @@ dependencies = [
  "monad-ethcall",
  "monad-node-config",
  "monad-rpc-docs",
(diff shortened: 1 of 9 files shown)


eerkaijun:
can you rebase against main and i can help review this

dshulyak:
@eerkaijun rebased. methods should be in the original order

PR #1740, opened 2025-04-24
Return random peers when target node not in local name records

eerkaijun:
closes https://github.com/category-labs/category-internal/issues/1534

### DIFF
--- a/monad-peer-discovery/src/discovery.rs
+++ b/monad-peer-discovery/src/discovery.rs
@@ -8,7 +8,7 @@ use monad_crypto::certificate_signature::{
     CertificateSignaturePubKey, CertificateSignatureRecoverable,
 };
 use monad_types::NodeId;
-use rand::{RngCore, SeedableRng};
+use rand::{RngCore, SeedableRng, seq::SliceRandom};
 use rand_chacha::ChaCha8Rng;
 use tracing::{debug, warn};
 
@@ -351,7 +351,16 @@ where
         } else {
             match self.peer_info.get(&target) {
                 Some(info) => vec![info.name_record],
-                None => vec![],
+                None => {
+                    // return random subset of peers up to MAX_PEER_IN_RESPONSE
+                    let mut peers: Vec<_> = self.peer_info.iter().collect();
+                    peers.shuffle(&mut self.rng);
+                    peers
+                        .into_iter()
+                        .take(MAX_PEER_IN_RESPONSE)
+                        .map(|(_k, v)| v.name_record)
+                        .collect()
+                }
             }
         };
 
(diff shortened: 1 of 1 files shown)


michael-yxchen:
we should also make client side changes, i.e. how to handle new peers discovered, in this or a separate PR

related: https://github.com/category-labs/category-internal/issues/1539

eerkaijun:
good point, tracking it here https://github.com/category-labs/category-internal/issues/1542

monad-peer-discovery/src/discovery.rs
```diff
@@ -349,7 +349,17 @@ where
...
                 Some(info) => vec![info.name_record],
-                None => vec![],
+                None => {
+                    // return random subset of peers up to MAX_PEER_IN_RESPONSE
+                    let mut node_ids: Vec<_> = self.peer_info.keys().collect();
+                    node_ids.shuffle(&mut self.rng);
+                    node_ids.truncate(MAX_PEER_IN_RESPONSE);
+
+                    node_ids
+                        .iter()
```
michael-yxchen: nit: into_iter

monad-peer-discovery/src/discovery.rs
```diff
@@ -349,7 +349,17 @@ where
...
+                        .filter_map(|id| self.peer_info.get(id).map(|info| info.name_record))
+                        .collect::<Vec<_>>()
```
michael-yxchen: nit: I wonder if this alternative saves some lookups. `peers` is a vec of references so memory impact should be fine

```suggestion
                    let mut peers: Vec<_> = self.peer_info.iter().collect();
                    peers.shuffle(&mut self.rng);
                    peers
                        .into_iter()
                        .take(MAX_PEER_IN_RESPONSE)
                        .map(|(_k, v)| v.name_record)
                        .collect()
```

eerkaijun: yeap i think your version is better, updated

monad-peer-discovery/src/discovery.rs
```diff
@@ -351,7 +351,16 @@ where
         } else {
             match self.peer_info.get(&target) {
                 Some(info) => vec![info.name_record],
-                None => vec![],
+                None => {
+                    // return random subset of peers up to MAX_PEER_IN_RESPONSE
+                    let mut peers: Vec<_> = self.peer_info.iter().collect();
```
dshulyak, line 356: is it a response for a query from network? 
i guess if peer_info gets large this might be a dos concern in a combination with requests spam

not sure if thats a big concern, but it is also possible to pick 16 random numbers and then select them in order from tree. it will avoid copying the whole tree on every request

dshulyak: i noticed that this was considered in a resolved comment, looks like not a problem

PR #1747, opened 2025-04-28
Try deserialize compressed messages

bamrith:


monad-compress/src/zstd.rs
```diff
@@ -0,0 +1,61 @@
...
+            level: DEFAULT_COMPRESSION_LEVEL as i32,
+        }
+    }
+}
+
+impl CompressionAlgo for ZstdCompression {
+    type CompressError = std::io::Error;
+    type DecompressError = std::io::Error;
+
+    fn new(quality: u32, _window_bits: u32, _custom_dictionary: Vec<u8>) -> Self {
```
andr-dev, line 24: Might be good to quickly assert here that quality <= `MAX_COMPRESSION_LEVEL`.

monad-compress/src/zstd.rs
```diff
@@ -0,0 +1,61 @@
+use crate::CompressionAlgo;
+
+pub const MAX_COMPRESSION_LEVEL: u32 = 15;
+pub const DEFAULT_COMPRESSION_LEVEL: u32 = 3;
```
andr-dev: The `zstd` crate defines this as a constant (https://docs.rs/zstd/0.13.2/zstd/constant.DEFAULT_COMPRESSION_LEVEL.html)

monad-compress/src/zstd.rs
andr-dev: Even better, there's a function `https://docs.rs/zstd/0.13.2/zstd/fn.compression_level_range.html` which you can call here to enforce the quality is within the range.

monad-raptorcast/src/lib.rs
```diff
@@ -358,30 +369,44 @@ where
...
-        // if that fails, try to deserialize as an old message instead
-        return match M::deserialize(bytes) {
+    if let Ok(inbound) =
+        InboundRouterMessage::<M, CertificateSignaturePubKey<ST>>::deserialize(bytes)
+    {
+        Ok(inbound)
+    // if that fails, try to deserialize as an old message
+    } else if let Ok(old_message) = M::deserialize(bytes) {
+        Ok(InboundRouterMessage::Application(old_message))
+    // else, decompress, then try to deserialize as an old message
```
omegablitz, line 389: Do we need some sort of versioning in the message itself to indicate that it's compressed? It seems suspicious to me to try to try each deserialization method and succeed on the first one that passes, but I haven't thought about this much. We want to ensure that bytes can't be misinterpreted in a way that the sender didn't intend.

bamrith: Right now, we only have one message type in raptorcast, so I didn't really think about this. It might be useful to start wrapping it with message type and version, especially if we might change the compression algorithm in the future. I'm guessing it will be useful for peer discovery messages too.

marcograss:
using zstd on network input to decompress might have some security pitfalls, especially considering we don't limit the size of the output. 

I could send a small message, like 300kb or something, and get a 10 GB zstd output.

Another variation of the attack is to craft a computationally worst case for zstd and saturate the CPU with a small bandwidth to send those intensive zstd compressed packets.

Did we consider this scenario?

omegablitz:
@marcograss Do you have any references we can read up on for the pathological inputs? Want to get a sense for how easy they are to construct

marcograss:
> @marcograss Do you have any references we can read up on for the pathological inputs? Want to get a sense for how easy they are to construct

for big input: a 10 giga can fit like 300 kb `dd if=/dev/zero bs=1G count=10 | zstd > poc.zst`  `-rw-rw-r-- 1 marco marco 336655 May  6 07:55 poc.zst`

for worst case resource testcase I will have to review the algo to craft one or generate one with a fuzzer or similar approaches

PR #1749, opened 2025-04-28
Fix slow txpool forward ingress queueing

andr-dev:


monad-eth-txpool-executor/src/forward.rs
```diff
@@ -105,7 +109,11 @@ impl EthTxPoolForwardingManagerProjected<'_> {
             ..
         } = self;
 
-        ingress.extend(txs);
+        if let Some(max_extend) = ingress.len().checked_sub(INGRESS_MAX_SIZE) {
```
omegablitz: Is this backwards? if ingress.len() < INGRESS_MAX_SIZE, this always returns None?

If so, I think this just needs to be `ingress.extend(txs.into_iter().take(INGRESS_MAX_SIZE.saturating_sub(ingress.len())))`

monad-eth-txpool-executor/src/forward.rs
```diff
@@ -68,13 +71,14 @@ impl EthTxPoolForwardingManager {
...
-            let mut chunk = ingress.split_off(INGRESS_CHUNK_MAX_SIZE);
-            std::mem::swap(ingress, &mut chunk);
-            Poll::Ready(chunk)
+        if let Some(at) = ingress.len().checked_sub(INGRESS_CHUNK_MAX_SIZE) {
+            if at > 0 {
+                ingress.rotate_left(INGRESS_CHUNK_MAX_SIZE);
+                return Poll::Ready(ingress.split_off(at).into());
+            }
         }
+
```
omegablitz: To me this is harder to read than `Poll::Ready(ingress.drain(..INGRESS_CHUNK_MAX_SIZE))`, because in that case there's no control flow branching and no need to look up what VecDeque::split_off returns. The performance difference will be negligible compared to disk reads.

Don't need to change though, maybe it's just personal preference.

andr-dev: Nah I caved, simple it is :)

monad-eth-txpool-executor/src/forward.rs
```diff
@@ -19,10 +20,11 @@ const EGRESS_MAX_RETRIES: usize = 2;
 
 const INGRESS_CHUNK_MAX_SIZE: usize = 128;
 const INGRESS_CHUNK_INTERVAL_MS: u64 = 8;
+const INGRESS_MAX_SIZE: usize = 64 * 1024;
```
omegablitz, line 23: 128 transactions can be pulled off the queue every 8ms, correct?
This would take `64 * 1024 * (8ms / 128) = 4s` to drain, which feels quite long.

I think we can set this a lot smaller, maybe like 8 * 1024 as an upper bound? It just needs to big enough to tolerate bursts, but we don't want it to be too big or it'll take a long time to drain.

andr-dev: Yeah I was thinking about this, but with the change the drain time is a function of the `CHUNK_SIZE` now so it seemed fairly reasonable to leave the `MAX_SIZE` at touch higher just in case. I'll also be decreasing the interval very soon with the addition of the txpool thread since it'll have its own dedicated compute time. As part of that, I'll revisit all the constants and update them accordingly.

PR #1750, opened 2025-04-28
Fix poll sleep implementation

maxkozlovsky:
The current implementation has a problem that there is no reference kept to the future, so it gets destroyed right after poll returns and sleep never fires.

Add sleep_future field to keep a reference to sleep future.

Make OutboundRequest.poll not async to avoid lifetime issues of sleep future having a reference to parent structure.

### DIFF
--- a/monad-statesync/src/ffi.rs
+++ b/monad-statesync/src/ffi.rs
@@ -1,7 +1,7 @@
 use std::{
     ffi::CString,
     ops::DerefMut,
-    pin::{pin, Pin},
+    pin::Pin,
     sync::{Arc, Mutex},
     task::{Context, Poll},
     time::Duration,
@@ -9,7 +9,7 @@ use std::{
 
 use alloy_consensus::Header;
 use alloy_rlp::Encodable;
-use futures::{FutureExt, Stream};
+use futures::{Future, Stream};
 use monad_crypto::certificate_signature::PubKey;
 use monad_executor_glue::{
     SessionId, StateSyncBadVersion, StateSyncRequest, StateSyncResponse, StateSyncUpsertType,
@@ -17,7 +17,10 @@ use monad_executor_glue::{
 };
 use monad_types::{DropTimer, NodeId, SeqNum};
 
-use crate::{bindings, outbound_requests::OutboundRequests};
+use crate::{
+    bindings,
+    outbound_requests::{OutboundRequests, RequestPollResult},
+};
 
 type StateSyncContext = Box<dyn FnMut(bindings::monad_sync_request)>;
 
@@ -40,6 +43,8 @@ pub(crate) struct StateSync<PT: PubKey> {
     response_tx: std::sync::mpsc::Sender<SyncResponse<PT>>,
 
     progress: Arc<Mutex<Progress>>,
+
+    sleep_future: Option<Pin<Box<tokio::time::Sleep>>>,
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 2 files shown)


monad-statesync/src/ffi.rs
```diff
@@ -430,12 +437,24 @@ impl<PT: PubKey> Stream for StateSync<PT> {
...
 
-        let fut = this.outbound_requests.poll();
-        if let Poll::Ready((peer, request)) = pin!(fut).poll_unpin(cx) {
-            return Poll::Ready(Some(SyncRequest::Request((peer, request))));
+        match this.outbound_requests.poll() {
+            RequestPollResult::Request(peer, request) => {
+                Poll::Ready(Some(SyncRequest::Request((peer, request))))
+            }
+            RequestPollResult::Timer(Some(instant)) => {
+                this.sleep_future = Some(Box::pin(tokio::time::sleep_until(instant.into())));
```
omegablitz: This was not introduced by your change, but can you only recreate sleep_future if [sleep_future.deadline()](https://docs.rs/tokio/latest/tokio/time/struct.Sleep.html#method.deadline) != instant?

IIRC recreating a tokio::Sleep is pretty costly, and every time this gets polled (which will be a lot unfortunately due to how ParentExecutor works), the Sleep will get recreated as is.

monad-statesync/src/ffi.rs
```diff
@@ -430,12 +437,24 @@ impl<PT: PubKey> Stream for StateSync<PT> {
...
+                match sleep_future.as_mut().poll(cx) {
+                    Poll::Ready(_) => {
+                        this.sleep_future = None;
+                        cx.waker().wake_by_ref();
+                        Poll::Pending
+                    }
+                    Poll::Pending => Poll::Pending,
+                }
+            }
+            RequestPollResult::Timer(None) => Poll::Pending,
```
omegablitz: Should we also set this.sleep_future = None here? It's not a correctness issue, but might make things clearer.

monad-statesync/src/ffi.rs
maxkozlovsky: When this branch is taken the sleep will always be non-zero because of how OutboundRequests.poll works. There is no optimization possible.

monad-statesync/src/ffi.rs
maxkozlovsky: ok

monad-statesync/src/ffi.rs
omegablitz: Yes, but if sleep_future is already Some(x) and the timer instant is also Some(x), it will recreate the tokio::Sleep with the same x

maxkozlovsky: ok

monad-statesync/src/ffi.rs
```diff
@@ -430,12 +437,34 @@ impl<PT: PubKey> Stream for StateSync<PT> {
...
+            }
+            RequestPollResult::Timer(Some(instant)) => {
+                if this
+                    .sleep_future
+                    .as_ref()
+                    .map_or(true, |f| f.deadline() != instant.into())
+                {
+                    this.sleep_future = Some(Box::pin(tokio::time::sleep_until(instant.into())));
+                }
+                let sleep_future = this.sleep_future.as_mut().unwrap();
```
andr-dev: We can shorten this (and get rid of the unwrap) with the following:
```rust
this.sleep_future.take_if(|f| f.deadline() != instant.into());
let sleep_future = this.sleep_future.get_or_insert_with(|| Box::pin(tokio::time::sleep_until(instant.into())));
```

maxkozlovsky: changed to use Sleep.reset() so no recreation of sleep is necessary when it is Some()

PR #1751, opened 2025-04-28
Add support for monad mainnet config

tongzhi1998:
Execution PR: https://github.com/category-labs/monad/pull/1282 (merged)

### DIFF
--- a/monad-ethcall/src/lib.rs
+++ b/monad-ethcall/src/lib.rs
@@ -12,6 +12,10 @@ use alloy_rlp::Encodable;
 use alloy_sol_types::decode_revert_reason;
 use bindings::monad_eth_call_result;
 use futures::channel::oneshot::{channel, Sender};
+use monad_chain_config::{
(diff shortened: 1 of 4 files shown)


monad-ethcall/src/lib.rs
```diff
@@ -240,6 +240,7 @@ pub async fn eth_call(
         1 => bindings::monad_chain_config_CHAIN_CONFIG_ETHEREUM_MAINNET,
         20143 => bindings::monad_chain_config_CHAIN_CONFIG_MONAD_DEVNET,
         10143 => bindings::monad_chain_config_CHAIN_CONFIG_MONAD_TESTNET,
+        143 => bindings::monad_chain_config_CHAIN_CONFIG_MONAD_MAINNET,
```
maxkozlovsky: should we use constants defined above here instead of numbers

monad-ethcall/src/lib.rs
```diff
@@ -240,6 +240,7 @@ pub async fn eth_call(
         1 => bindings::monad_chain_config_CHAIN_CONFIG_ETHEREUM_MAINNET,
```
maxkozlovsky: The variable might be better named chain_config instead of chain_id

PR #1753, opened 2025-04-29
Simplify estimate_gas logic

tongzhi1998:


monad-rpc/src/gas_handlers.rs
```diff
@@ -120,32 +118,10 @@ async fn estimate_gas<T: EthCallProvider>(
...
-                    ))
-                }
-            }
-        } else {
-            (gas_used.sub(1), txn.gas_limit())
-        };
+    let mut upper_bound_gas_limit = txn.gas_limit();
+    let mut lower_bound_gas_limit = (gas_used + gas_refund)
+        .saturating_sub(1)
+        .max(MIN_TRANSFER_GAS - 1);
```
eerkaijun, line 124: doesn't seem to be equivalent to the above code - if so, what's the motivation for this change?

tongzhi1998: It seems weird to call eth_call twice before starting to binary search. We can use the result from the first eth_call as lower / upper bound for the binary search.

PR #1755, opened 2025-04-29
Allow manually invoke test event in swarm

eerkaijun:
Allow manually invoking test event in peer disc swarm for easier testing of different scenarios. 

E.g would be useful for https://github.com/category-labs/monad-bft/pull/1740 to test for various lookup scenarios

### DIFF
--- a/monad-peer-disc-swarm/src/lib.rs
+++ b/monad-peer-disc-swarm/src/lib.rs
@@ -58,6 +58,7 @@ pub enum SwarmEvent {
     DriverEvent,
     ExecutorEvent,
     InboundMessage,
+    TestEvent,
 }
 
 pub struct NodeBuilder<S, B>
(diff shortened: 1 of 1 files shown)


michael-yxchen:
If we're invoking test event on a specific peer discovery state, unit test might be more appropriate. We should focus on higher level behavior in swarm tests - nodes joining/leaving, network msg drops etc

eerkaijun:
my thinking is that it would allow easier testing of standalone peer discovery logic in peer-disc-swarm without needing external consensus logic. for example, sending peer lookup request should be triggered when new validators are locked in an upcoming epoch. manually invoking these test events allow us to test for peer discovery related flows without needing to couple with other consensus logic. unit tests means we would need to manually send and handle all requests etc which would be messy

michael-yxchen:
it's possible to "mock" out consensus and just simulate the validator set locked part, through a new executor in `MockPeerDiscExecutor`. I guess with your approach, you can schedule a validator set change at slightly different time points as well. The difference is more complexity in test themselves vs. test infra

for unit test, i meant instantiating the state machine exactly how you want it, since you have access to all the private fields

PR #1756, opened 2025-04-29
Implement Send+Sync State Backend Wrapper

andr-dev:
In an upcoming PR, the txpool will be moved to a separate task which will run on the same tokio runtime as monad-node. As such, to start it as a tokio task, the run loop must be Send + Sync. Since the txpool needs a state backend to lookup account balances and nonces, we need to have a Send+Sync SBT, but the triedb handle is currently not thread-safe. This change introduces a simple wrapper around an SBT which spins up a thread to offload all SBT computation to the thread, enabling any SBT to be Send+Sync. An added side benefit is that with this change, consensus state and the txpool share the same state backend cache.

Currently, the wrapper implements StateBackend which means all operations are still synchronous. As an open discussion point, it might be valuable to introduce an AsyncStateBackend trait to enable awaiting on results from the thread channel.

### DIFF
--- a/monad-state-backend/src/thread.rs
+++ b/monad-state-backend/src/thread.rs
@@ -1,0 +1,236 @@
+use std::sync::{mpsc, Arc};
+
+use alloy_primitives::Address;
+use monad_eth_types::EthAccount;
+use monad_types::{BlockId, Round, SeqNum};
+use tracing::warn;
+
+use crate::{StateBackend, StateBackendError};
+
+// Since the StateBackendThreadClient is synchronous, it will only allow one inflight request per
+// sync context so a value of 16 allows 16 threads to simulatneously make state backend requests.
+const MAX_INFLIGHT_REQUESTS: usize = 16;
+
+enum StateBackendThreadRequest {
+    GetAccountStatuses {
+        block_id: BlockId,
+        seq_num: SeqNum,
+        round: Round,
+        is_finalized: bool,
+        addresses: Vec<Address>,
+        tx: mpsc::SyncSender<Result<Vec<Option<EthAccount>>, StateBackendError>>,
+    },
+    RawReadEarliestFinalizedBlock {
+        tx: mpsc::SyncSender<Option<SeqNum>>,
+    },
+    RawReadLatestFinalizedBlock {
+        tx: mpsc::SyncSender<Option<SeqNum>>,
+    },
+    TotalDbLookups {
+        tx: mpsc::SyncSender<u64>,
+    },
+}
+
+#[derive(Clone)]
+pub struct StateBackendThreadClient {
+    handle: Arc<std::thread::JoinHandle<()>>,
+    request_tx: mpsc::SyncSender<StateBackendThreadRequest>,
+}
+
+impl StateBackendThreadClient {
+    pub fn new<SBT>(state_backend: impl FnOnce() -> SBT + Send + 'static) -> Self
+    where
+        SBT: StateBackend,
+    {
+        let (request_tx, request_rx) = mpsc::sync_channel(MAX_INFLIGHT_REQUESTS);
+
+        let handle = Arc::new(std::thread::spawn(move || {
+            StateBackendThread::new(state_backend, request_rx).run()
+        }));
+
+        Self { handle, request_tx }
+    }
+
+    fn send_and_recv_request<T>(
+        &self,
+        request: impl FnOnce(mpsc::SyncSender<T>) -> StateBackendThreadRequest,
+    ) -> T {
+        if self.handle.is_finished() {
+            panic!("StateBackendThread terminated!");
+        }
+
+        let (tx, rx) = mpsc::sync_channel(0);
+
+        self.request_tx
+            .send(request(tx))
+            .expect("StateBackendThread is alive");
+
+        rx.recv().expect("StateBackendThread sends response")
+    }
+}
+
+impl StateBackend for StateBackendThreadClient {
+    fn get_account_statuses<'a>(
+        &self,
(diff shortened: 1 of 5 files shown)


mikety:
lgtm, as a follow up I suggest to add a few tests, adding a couple of StateBackendThreadClients sending requests to the StateBackend.

monad-state-backend/src/thread.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    fn send_and_recv_request<T>(
+        &self,
+        request: impl FnOnce(mpsc::SyncSender<T>) -> StateBackendThreadRequest,
+    ) -> T {
+        if self.handle.is_finished() {
+            panic!("StateBackendThread terminated!");
+        }
+
+        let (tx, rx) = mpsc::sync_channel(0);
```
dshulyak, line 62: oneshot?

andr-dev: https://doc.rust-lang.org/std/sync/mpsc/fn.sync_channel.html
"This channel has an internal buffer on which messages will be queued. bound specifies the buffer size. When the internal buffer becomes full, future sends will block waiting for the buffer to open up. Note that a buffer size of 0 is valid, in which case this becomes “rendezvous channel” where each [send](https://doc.rust-lang.org/std/sync/mpsc/struct.SyncSender.html#method.send) will not return until a [recv](https://doc.rust-lang.org/std/sync/mpsc/struct.Receiver.html#method.recv) is paired with it."

We're already blocking until the response message comes in, so it's better here to just not have an internal buffer and have the send block as well.

andr-dev:
> implementing async backend would be a better option. you do mention that it is possible, but why not to go for it? the follow up already uses tokio tasks for running pool. this way you will eliminate the risk of blocking tokio runtime accidentally

I gave it a shot and it's unfortunately not as short/nice as I expected it to be since we'll need a new AsyncStateBackend trait and would need to migrate any existing code over (can be restricted to just txpool). I am planning on going for round 2 in the next week or two but it was getting a bit unwieldily in my branch.

EDIT: I just remembered why this was a nightmare. We don't use state backend directly, we use it through the EthBlockPolicy so we'd have to turn that async as well.

> one other thing that would make sense is to allow to build proposal concurrently with accounts lookups, so that spike in disk latency wouldn't block proposal building process

The actual proposal building process is extremely fast (<2ms), the slow part is the account lookups so there's likely little benefit in doing so.

dshulyak:
> The actual proposal building process is extremely fast (<2ms), the slow part is the account lookups so there's likely little benefit in doing so.

why there is likely no benefit of doing so, if account lookups can delay proposal building and account lookup latency depends on disk usage. just put it differently, if proposal can be built concurrently with slow accounts lookup then they can't delay that process, even if proposal building is instant

andr-dev:
> why there is likely no benefit of doing so, if account lookups can delay proposal building and account lookup latency depends on disk usage. just put it differently, if proposal can be built concurrently with slow accounts lookup then they can't delay that process, even if proposal building is instant

During the building of a proposal, you must first do an account lookup before using that account in the proposal. In the current implementation, after doing these account lookups, the txpool then builds the proposal (which takes <2ms).

If I'm understanding this correctly, you are proposing doing the lookups and the proposal building process concurrently, so that we're not stuck waiting for the account lookups. In this case, the lookups would still take the same amount of time and we'd save at most <2ms by not having to build the proposal at the end after we've finished all the lookups. Given the lookups can be >100ms, there's simply not a strong benefit to saving those last <2ms all while introducing significant code complexity.

dshulyak:
i was thinking about lookups in insert_txs. if there are many requests and some of them are slow, can that delay create_proposal?

andr-dev:
> i was thinking about lookups in insert_txs. if there are many requests and some of them are slow, can that delay create_proposal?

At most `max(MAX_BATCH_LEN, INGRESS_CHUNK_MAX_SIZE) = max(128, 128) = 128` txs (and thus 128 addresses) are inserted at a time, so the max delay is the account lookup time for 128 addresses which at 50us each is 6.4ms.

dshulyak:
i think thats an optimistic estimate. it is more relevant to look at upper percentiles (e.g p99) to understand by how much it can delay. i instrumented codebase to capture distribution for processing ipc txs and create proposal requests. i run single network with 100 000 senders and 3000 target tps.

image

and then i also launched rpc generator, you will see a bump in both create_proposal and ipc_insert_txs.
so even with this workload insert txs can delay create_proposal by ~20-30ms. i don't know enough about other timing expectation to understand if it can cause an issue, but that solution mentioned above eliminates this risk

dshulyak:
btw if we look at more degenerate cases with p999, latency can spike up to 100ms for insert_txs, thats seems pretty high. 

image

andr-dev:
@dshulyak To clarify, when I said there's likely little performance benefit, I meant within the context of the txpool. That being said, I do think there's some significant performance benefits for other parts of the system if we can make an `AsyncStateBackend` trait so we can defer CPU time to other tasks while waiting for account lookups. It seems like you're pretty excited about moving in this direction and I won't have time for another week or two to try it again so feel free to take a stab at it :)

dshulyak:
it is not that i am super excited. it is just a common problem that can be addressed proactively, upper percentiles are often ignored then you get a condition that appears once a week, hard to diagnose and has unpleasant downstream effects. 

for instance i don't know what will happen if proposal creation is delayed by 100ms. could it lead to triggering a timeout? that will make leader unhappy. also could be that it just delays progress a bit, so consensus is reached in 600ms instead of 500, thats much better but also can be avoided entirely. anyway i will create an issue to followup

PR #1757, opened 2025-04-29
Peer discovery message RLP encodings

michael-yxchen:


monad-peer-discovery/src/message.rs
```diff
@@ -0,0 +1,189 @@
...
+            }
+            PeerDiscoveryMessage::PeerLookupResponse(response) => {
+                PeerDiscoveryEvent::PeerLookupResponse { from, response }
+            }
+        }
+    }
+}
+
+impl<ST: CertificateSignatureRecoverable> Encodable for PeerDiscoveryMessage<ST> {
+    fn encode(&self, out: &mut dyn alloy_rlp::BufMut) {
```
eerkaijun, line 39: why do we need to rlp encode peer discovery message?

michael-yxchen: pasting from dm: we’re encoding all network messages with rlp

monad-peer-discovery/src/message.rs
```diff
@@ -0,0 +1,194 @@
...
+                &mut payload,
+            )?)),
+            _ => Err(alloy_rlp::Error::Custom(
+                "failed to decode unknown PeerDiscoveryMessage",
+            )),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq)]
```
bamrith: What's the reason to not derive `RlpEncodable` and `RlpDecodable` here ?

michael-yxchen: alloy_rlp crate doesn't implement those two traits on `Option<T>`

bamrith: if it is at the end of a struct, I think you can add `[rlp(trailing)]` ? we've done the same for other structs in the codebase.

michael-yxchen: cool, it has some savings (~1 byte) so I'll switch over

PR #1758, opened 2025-04-30
TxPool Task

andr-dev:
This changeset introduces the TokioTaskUpdater which easily enables running Updaters as a tokio task with very minor changes to the updater. The txpool will be the first to use said wrapper to let the txpool run in parallel to the main tokio task.

### DIFF
--- a/monad-eth-txpool-executor/src/lib.rs
+++ b/monad-eth-txpool-executor/src/lib.rs
@@ -2,7 +2,10 @@ use std::{
     io,
     marker::PhantomData,
     pin::Pin,
-    sync::atomic::{AtomicU64, Ordering},
+    sync::{
+        atomic::{AtomicU64, Ordering},
+        Arc,
+    },
     task::Poll,
     time::Duration,
 };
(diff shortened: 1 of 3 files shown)


dshulyak:
also, there are no relevant tests in this module. is it tested somewhere else?

monad-eth-txpool-executor/src/lib.rs
```diff
@@ -74,36 +76,66 @@ where
...
+        mut command_rx: mpsc::Receiver<
+            Vec<TxPoolCommand<ST, SCT, EthExecutionProtocol, EthBlockPolicy<ST, SCT>, SBT>>,
+        >,
+        event_tx: mpsc::Sender<MonadEvent<ST, SCT, EthExecutionProtocol>>,
+    ) {
+        use futures::StreamExt;
+
+        loop {
+            tokio::select! {
+                commands = command_rx.recv() => {
```
dshulyak, line 154: i think "biased" makes sense here, so that TxPoolCommand::CreateProposal prioritized over polling from ipc/network

andr-dev: I think commands in general should be prioritized here, good idea :)

PR #1759, opened 2025-04-30
TxPool preload account balances

andr-dev:
Currently, txpool proposal creation time is dominated by account lookups. As the number of unique senders in the txpool increases, the time required to lookup these accounts increases linearly. For thousands of accounts, this can delay proposal creation times by >100ms.

This changeset introduces a preload manager which predicts what seqnum/round the node will be a leader and concurrently loads these account balances ahead of time. This lookup adds the account to the StateBackendCache so that the account lookup at proposal time is fast.

### DIFF
--- a/monad-eth-txpool-executor/src/preload.rs
+++ b/monad-eth-txpool-executor/src/preload.rs
@@ -1,0 +1,416 @@
+use std::{
+    collections::{btree_map::Entry, BTreeMap, HashSet},
+    pin::Pin,
+    task::{Context, Poll, Waker},
+    time::Duration,
+};
+
+use alloy_primitives::Address;
+use futures::FutureExt;
+use indexmap::IndexSet;
+use monad_consensus_types::signature_collection::SignatureCollection;
+use monad_crypto::certificate_signature::{
+    CertificateSignaturePubKey, CertificateSignatureRecoverable,
+};
+use monad_eth_block_policy::EthValidatedBlock;
+use monad_types::{Round, SeqNum};
+use tokio::time::Sleep;
+use tracing::{debug, warn};
+
+// Each account lookup takes ~30us so this blocks the thread for at most a touch under 4ms.
+const PRELOAD_CHUNK_MAX_ADDRESSES: usize = 128;
+const PRELOAD_INTERVAL_MS: u64 = 1;
+const COMMITTED_TO_PROPOSAL_SEQNUM_DIFF: SeqNum = SeqNum(2);
+
+pub struct EthTxPoolPreloadEntry {
+    // On insertion, we set preload to true. After that, if the SeqNum associated with the entry
+    // is no longer in the upcoming predicted leader seqnums then preload is set to false.
+    preload: bool,
+    pending: IndexSet<Address>,
+    done: HashSet<Address>,
+}
+
+pub struct EthTxPoolPreloadManager {
+    map: BTreeMap<SeqNum, EthTxPoolPreloadEntry>,
+    timer: Pin<Box<Sleep>>,
+    waker: Option<Waker>,
+}
+
+impl Default for EthTxPoolPreloadManager {
+    fn default() -> Self {
+        Self {
+            map: Default::default(),
+            timer: Box::pin(tokio::time::sleep(Duration::ZERO)),
+            waker: None,
+        }
+    }
+}
+
+impl EthTxPoolPreloadManager {
+    pub fn enter_round(
+        &mut self,
+        round: Round,
+        last_commit_seqnum: SeqNum,
+        upcoming_leader_rounds: Vec<Round>,
+        generate_requests: impl FnOnce() -> Vec<Address>,
+    ) {
+        self.map = self.map.split_off(&(last_commit_seqnum + SeqNum(1)));
+
+        let upcoming_predicted_proposal_seqnums = upcoming_leader_rounds
+            .iter()
+            .filter_map(|leader_round| {
+                let round_diff = leader_round.0.checked_sub(round.0)?;
+
+                Some(last_commit_seqnum + COMMITTED_TO_PROPOSAL_SEQNUM_DIFF + SeqNum(round_diff))
+            })
+            .collect::<HashSet<SeqNum>>();
+
+        self.map.retain(|predicted_proposal_seqnum, entry| {
+            // We recreate the pending sets from the txpool snapshot if we're a leader in the
+            // current round or the next.
+            entry.pending.clear();
+
+            // If the last round advances as a result of a TC, then the round number increments
+            // while the seqnum does not. In this case, some of the previously predicted proposal
+            // seqnums are no longer correct. These entries should be invalidated to avoid
+            // unnecessary preloading.
+            if !upcoming_predicted_proposal_seqnums.contains(predicted_proposal_seqnum) {
+                if entry.done.is_empty() {
+                    return false;
+                }
+
+                entry.preload = false;
+            }
+
+            true
+        });
+
+        let preload_round = upcoming_leader_rounds.contains(&round);
+        let preload_next_round = upcoming_leader_rounds.contains(&(round + Round(1)));
+
+        if !(preload_round || preload_next_round) {
+            return;
+        }
+
+        if preload_round {
+            let round_predicted_proposal_seqnum =
+                last_commit_seqnum + COMMITTED_TO_PROPOSAL_SEQNUM_DIFF;
+
+            self.map
+                .entry(round_predicted_proposal_seqnum)
+                .or_insert(EthTxPoolPreloadEntry {
+                    preload: true,
+                    pending: IndexSet::default(),
+                    done: HashSet::default(),
+                });
+        }
+
+        if preload_next_round {
+            let next_round_predicted_proposal_seqnum =
+                last_commit_seqnum + COMMITTED_TO_PROPOSAL_SEQNUM_DIFF + SeqNum(1);
+
+            self.map
+                .entry(next_round_predicted_proposal_seqnum)
+                .and_modify(|entry| {
+                    // We enable preload for the next round, which remains enabled until we receive a
+                    // CreateProposal event for that seqnum and it gets disabled or the entry is
+                    // removed.
+                    entry.preload = true;
+                })
+                .or_insert(EthTxPoolPreloadEntry {
+                    preload: true,
+                    pending: IndexSet::default(),
+                    done: HashSet::default(),
+                });
+        }
+
+        let requests = generate_requests();
+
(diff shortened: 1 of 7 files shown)


monad-eth-txpool-executor/src/preload.rs
```diff
@@ -0,0 +1,234 @@
...
+use monad_consensus_types::signature_collection::SignatureCollection;
+use monad_crypto::certificate_signature::{
+    CertificateSignaturePubKey, CertificateSignatureRecoverable,
+};
+use monad_eth_block_policy::EthValidatedBlock;
+use monad_types::{DropTimer, Round, SeqNum};
+use tokio::time::Sleep;
+use tracing::{debug, warn};
+
+// Each account lookup takes ~30us so this blocks the thread for at most a touch under 8ms.
```
dshulyak: in my experiments on non-idle system any balance lookup that hits disk takes way more than that. even fio reports single page disk request to have latency ~50us on the waw2 server with nvme, and each balance query will likely have to issue more than 1. 30us might be hitting a cache somewhere

i would add a histogram to track distribution. i suspect that the total time will be much higher. it will also depend on the load as a disk is shared resource. 

this is an example of values that i would expect on non-idle system
image

monad-eth-txpool-executor/src/preload.rs
```diff
@@ -0,0 +1,234 @@
...
+    fn default() -> Self {
+        Self {
+            map: Default::default(),
+            timer: Box::pin(tokio::time::sleep(Duration::ZERO)),
+            waker: None,
+        }
+    }
+}
+
+impl EthTxPoolPreloadManager {
```
dshulyak, line 49: it will be nice to have atleast sanity test to show what it is doing in isolation

monad-eth-txpool-executor/src/preload.rs
```diff
@@ -0,0 +1,234 @@
...
+
+        let preload_round = upcoming_leader_rounds.contains(&round);
+        let preload_next_round = upcoming_leader_rounds.contains(&(round + Round(1)));
+
+        if !(preload_round || preload_next_round) {
+            return;
+        }
+
+        if preload_next_round {
+            let next_round_predicted_proposal_seqnum = last_commit_seqnum + SeqNum(2) + SeqNum(1);
```
dshulyak: what is (2) + (1) ?

andr-dev: The `1` comes from this being for the next round. The `2` is because at round N for last_committed_seqnum M we propose seqnum `M+2` because there's one block in our pending block tree which we're extending thus (2) + (1).

andr-dev:
> i think there are no relevant tests in this module. is it tested just by running a node? it also makes it harder to review channges

Ah thanks for pointing it out, I had written tests in a separate branch to make it easier to identify the core changes when I made a custom image to test on a stressnet node. I've squashed them in and pushed them up!

> i don't want to delay this change (i assume it was planned), but i personally would expect more justifications of the change. i assume it wants to improve proposal creation time, but i don't see any metrics that can provide insight into that. usually you will want atleast to have relevant histogram and observe distribution with the relevant workload

Thanks for asking about how I got insights into how this changeset impacts proposal creation times. The metric you're looking for is `monad_bft_txpool_create_proposal_elapsed_ns` which can be combined with other txpool metrics to monitor proposal creation times with respect to
1. How many unique addresses are present in the subset of nonce-valid transactions in the txpool (`monad_bft_txpool_pool_create_proposal_available_addresses`)
2. How many non-cached backend lookups were performed from the subset above (`monad_bft_txpool_pool_create_proposal_backend_lookups`)

These two views, along with many other visualizations, were added to the [TxPool dashboard](https://grafana.devcore4.com/d/cecyguhbuovlsa/txpool?orgId=1&from=now-1h&to=now&timezone=browser&var-net=stressnet&var-nodes=$__all&refresh=5s) a few weeks ago to show the average P50, P95, and P99 create_proposal durations from the past hour over time.

In addition to the above, this change also added [new metrics](https://github.com/category-labs/monad-bft/pull/1759/files#diff-358c223d4ce61f030830272f0eecefef65cc2cc7eff6f8c8d450ff01eeb656d6R29-R30) for monitoring the preloading process independently using the `monad_bft_txpool_preload_.*` metrics which are also displayed on the dashboard linked above. These metrics give us meaningful visibility into both the preloading module and proposal creation performance.

Over the past 2-3 weeks, I have done substantial stressnet testing with unique senders at high TPS to validate that the preloading module effectively loads accounts into the StateBackendCache. Results from running 5k TPS with unique senders showed a reduction in proposal creation times from >100ms down to <20ms. There is still additional work to do to get that 20ms number down to <2ms which has mostly been identified as the preloading module being unable to preload accounts when the node receives a proposal as consensus state submits a large account balance lookup request to validate the proposal which at that TPS and with that many senders takes >100ms. That being said, as it stands now this change results in significant performance benefits which would directly carry over with no further modification if we improve the behavior detailed above and any other related ones that we may discover.

On the broader note about justification, I'm always happy to elaborate on the reasoning behind changes if anything is unclear as the most effective way to get the full context behind a change is by asking clarifying questions. Thanks again for reviewing and sharing your thoughts.

dshulyak:
thanks for explanation.

i was actually confused about that metric monad_bft_txpool_create_proposal_elapsed_ns . 
it definitely says something that it became lower, but the accuracy of that metric depends on reporting interval.
e.g if metrics are reported every 10s and proposal is built every 500ms, then it will report only every 20th latency.
it is much more reliable to collect into histogram from every proposal created

monad-state/src/consensus.rs
```diff
@@ -230,7 +232,8 @@ where
         consensus_cmds
             .into_iter()
             .map(|cmd| WrappedConsensusCommand {
-                state_root_delay: consensus.config.execution_delay,
+                state_root_delay: self.consensus_config.execution_delay,
```
mikety, line 250: nit: the reason not using `consensus.config` to be consistent with the rest of the code here?

monad-eth-txpool-executor/src/preload.rs
```diff
@@ -0,0 +1,413 @@
...
+        generate_requests: impl FnOnce() -> Vec<Address>,
+    ) {
+        self.map = self.map.split_off(&(last_commit_seqnum + SeqNum(1)));
+
+        let upcoming_predicted_proposal_seqnums = upcoming_leader_rounds
+            .iter()
+            .filter_map(|leader_round| {
+                let round_diff = leader_round.0.checked_sub(round.0)?;
+
+                Some(last_commit_seqnum + SeqNum(2) + SeqNum(round_diff))
```
mikety: nit: Use a constant with descriptive name or add a comment to clarify SeqNum(2).

monad-eth-txpool-executor/src/preload.rs
```diff
@@ -0,0 +1,413 @@
...
+            })
+            .collect::<HashSet<SeqNum>>();
+
+        self.map.retain(|predicted_proposal_seqnum, entry| {
+            // We recreate the pending sets from the txpool snapshot if we're a leader in the
+            // current round or the next.
+            entry.pending.clear();
+
+            // If the last round TCd then the round would advance while the seqnum does not, in
```
mikety: clarify TCd

monad-state/src/consensus.rs
```diff
@@ -494,6 +477,48 @@ where
 
         Ok((author, validated_mesage.into_inner()))
     }
+
+    fn compute_upcoming_leader_round_pairs<
+        const INCLUDE_CURRENT_ROUND: bool,
```
mikety, line 497: is   `INCLUDE_CURRENT_ROUND == true` used or planned to be used?

monad-eth-txpool-executor/src/preload.rs
```diff
@@ -0,0 +1,413 @@
...
+                entry.preload = false;
+            }
+
+            true
+        });
+
+        let preload_round = upcoming_leader_rounds.contains(&round);
+        let preload_next_round = upcoming_leader_rounds.contains(&(round + Round(1)));
+
+        if !(preload_round || preload_next_round) {
```
mikety, line 91: Please clarify when `preload_round`, which is being entered and therefore is being loaded is used.
If the goal is only to preload the next round the code can be made less general and simpler.

monad-eth-txpool-executor/src/preload.rs
mikety: Please add clarifying comments

monad-state/src/consensus.rs
andr-dev: `consensus.config` is just a reference to `self.consensus_config`. We can no longer use `consensus.config` because we need to call `self.get_upcoming_leader_rounds()` so that needs an immutable reference to `self` but there are mutable references to `self` in `consensus`.

monad-state/src/consensus.rs
andr-dev: It is being used https://github.com/category-labs/monad-bft/pull/1759/files#diff-4b7dd7839f4cba2948d2da038c372c68a8c36a85b73a8b59a28dbfe3f8cb71b4R517

monad-eth-txpool-executor/src/preload.rs
andr-dev: `preload_round` and `preload_next_round` are booleans which indicate whether we should preload accounts for the current / next round. If there is a delay between the `EnterRound` command and the `CreateProposal` command for a given round, the `preload_round` flag will allow for some additional batches of preloading.

monad-eth-txpool-executor/src/preload.rs
andr-dev: I added a constant for `SeqNum(2)` as you suggested, so now it reads "last_commit_seqnum + COMMITTED_TO_PROPOSAL_SEQNUM_DIFF + SeqNum(1)" which should help clarify the arithmetic.

Generally, when people suggest adding comments, my gut reaction is that the code needs to be written in a way as to not require comments. If people can't understand what the code is doing without a comment, you should rewrite it so they can without one :)

monad-eth-txpool/src/pool/mod.rs
```diff
@@ -197,6 +199,24 @@ where
...
+        block_policy: &EthBlockPolicy<ST, SCT>,
+        state_backend: &SBT,
+    ) {
+        if !self.tracked.try_promote_pending(
+            event_tracker,
+            block_policy,
+            state_backend,
+            &mut self.pending,
+            0,
+            128,
```
mikety: 1. `128` Should be either converted to a constant or pass as an argument.
2. nit: consider using  promote_pending on line 185

PR #1761, opened 2025-04-30
Handling local_name_record in ping

michael-yxchen:


monad-peer-discovery/src/discovery.rs
```diff
@@ -119,11 +119,15 @@ where
             return cmds;
         };
 
+        let maybe_attach_record = if peer_entry.last_seen.is_none() {
+            Some(self.self_record)
+        } else {
+            None
+        };
```
eerkaijun: `last_seen` is removed in the pruning PR, so we will need another way to determine whether to attach self record. i'm thinking either
1. have a field that records number of successful pings which gets reset during the `prune` event (this might be most similar to other peer discovery impl where peers are ranked based on connection duration)
2. or just always attach self record in `send_ping`, since I think the size of a name record is pretty small anyways

monad-peer-discovery/src/discovery.rs
```diff
@@ -151,19 +155,49 @@ where
...
+                            info.name_record = peer_name_record;
+                        })
+                        .or_insert_with(|| PeerInfo {
+                            last_ping: None,
+                            last_seen: None,
+                            name_record: peer_name_record,
+                        });
+                } else {
+                    debug!("invalid signature in ping.local_name_record");
+                    // TODO: skip rest?
```
eerkaijun: i think we can not send pong since our name record is outdated and we don't have a valid new one

monad-peer-discovery/src/discovery.rs
```diff
@@ -151,19 +155,49 @@ where
...
+                )
+            } else if self.peer_info.get(&from).is_some_and(|local| {
+                peer_name_record.seq() == local.name_record.seq()
+                    && peer_name_record != local.name_record
+            }) {
+                warn!(
+                    ?from,
+                    "peer updated name record without bumping sequence number"
+                )
+            }
```
eerkaijun: similarly I think we can exit early and not send pong?

monad-peer-discovery/src/discovery.rs
michael-yxchen: what about attaching name record if `unresponsive_pings > 0`? slightly inaccurate, but behaves as expected starting second ping

self record is probably around 50 bytes (4 bytes ip, 2 bytes port, 1 byte seq (u64 but small), 33 bytes sig, + overhead). low in absolute overhead, high relatively - ping msg is only a few bytes

monad-peer-discovery/src/discovery.rs
michael-yxchen: changed to return early

monad-peer-discovery/src/discovery.rs
eerkaijun: When a node joins the network, it will send pings to other validators to initiate connections. With this approach, all its first ping will be dropped as the ping message does not have the name record, so it will need to wait at least `ping_period` to be connected. Will that be undesirable?

michael-yxchen: yea that's the inaccurate part. Can you point me to how other protocols do heuristic based pruning?

michael-yxchen: thanks, I’ll set it to always attach for now, and keep the message encoding to allow optional (2 extra bytes in rlp i think). if we later find some other heuristic to decide when not to include, we’ll have the flexibility to do so

PR #1762, opened 2025-04-30
Version router messages

bamrith:


monad-raptorcast/src/lib.rs
```diff
@@ -54,8 +56,8 @@ where
 pub struct RaptorCast<ST, M, OM, SE>
 where
     ST: CertificateSignatureRecoverable,
-    M: Message<NodeIdPubKey = CertificateSignaturePubKey<ST>> + Deserializable<Bytes>,
-    OM: Serializable<Bytes> + Into<M> + Clone,
+    M: Message<NodeIdPubKey = CertificateSignaturePubKey<ST>> + Deserializable<Bytes> + Decodable,
+    OM: Serializable<Bytes> + Encodable + Into<M> + Clone,
```
michael-yxchen, line 63: do we need both `Deserializable` and `Decodable` trait bounds? it looks weird visually

edit: I answered my own question after reading message.rs. It's required now because we're using De/Serializable atm, and RouterMessage requires app message to be De/Encodable. After client switch over, when we retire old message format, we can remove De/Serializable trait bounds

monad-raptorcast/src/message.rs
```diff
@@ -0,0 +1,110 @@
...
+struct NetworkMessageVersion {
+    pub serialize_version: u32,
+    pub compression_version: u32,
+}
+
+impl NetworkMessageVersion {
+    pub fn version() -> Self {
+        Self {
+            serialize_version: 1,
+            compression_version: 1,
```
michael-yxchen: Create constants for these versions

monad-raptorcast/src/message.rs
```diff
@@ -0,0 +1,110 @@
...
+impl<OM: Encodable> OutboundRouterMessage<OM> {
+    pub fn try_serialize(self) -> Result<Bytes, SerializeError> {
+        let mut buf = BytesMut::new();
+
+        let version = NetworkMessageVersion::version();
+        match self {
+            Self::AppMessage(app_message) => {
+                if version.compression_version == 1 {
+                    // encode as uncompressed message
+                    let enc: [&dyn Encodable; 3] = [&version, &1u8, &app_message];
```
michael-yxchen, line 41: what is `1_u8` used for?

monad-raptorcast/src/message.rs
```diff
@@ -0,0 +1,110 @@
...
+impl From<alloy_rlp::Error> for DeserializeError {
+    fn from(err: alloy_rlp::Error) -> Self {
+        DeserializeError(format!("rlp decode error: {:?}", err))
+    }
+}
+
+impl<M: Decodable> InboundRouterMessage<M> {
+    pub fn try_deserialize(data: &Bytes) -> Result<Self, DeserializeError> {
+        let mut buf = data.as_ref();
+        let mut payload = Header::decode_bytes(&mut buf, true).map_err(DeserializeError::from)?;
```
michael-yxchen: nit
```suggestion
        let mut payload = Header::decode_bytes(&mut buf.as_ref(), true).map_err(DeserializeError::from)?;
```

monad-raptorcast/src/message.rs
```diff
@@ -0,0 +1,110 @@
...
+                match version.compression_version {
+                    1 => {
+                        // decode as uncompressed message
+                        let app_message =
+                            M::decode(&mut payload).map_err(DeserializeError::from)?;
+                        Ok(Self::AppMessage(app_message))
+                    }
+                    2 => {
+                        // decompress message
+                        let mut decompressed_app_message = Vec::new();
```
michael-yxchen, line 102: i haven't looked closely at Encodable/Decodable impls. Would BytesMut here reduce copies?

monad-compress/src/zstd.rs
```diff
@@ -0,0 +1,64 @@
...
+    use super::*;
+    #[test]
+    fn test_lossless_compression() {
+        let mut data = Vec::new();
+        File::open("examples/txbatch.rlp")
+            .unwrap()
+            .read_to_end(&mut data)
+            .unwrap();
+
+        let algo = ZstdCompression::new(8, 0, Vec::new());
```
michael-yxchen: nit: should we use `ZstdCompression::default` cuz I saw that's used in raptorcast message compression

monad-raptorcast/src/message.rs
bamrith: To differentiate the message types. I assumed the peer discovery message would have `2_u8`, full node messages `3_u8` etc.

monad-raptorcast/src/message.rs
bamrith: I'll read the impls. `Vec` was used here only because the `CompressionAlgo` trait impl takes in a `&mut Vec<u8>`.

monad-raptorcast/src/message.rs
michael-yxchen: Ah cool, i forgot

monad-raptorcast/src/message.rs
```diff
@@ -0,0 +1,116 @@
...
+pub struct SerializeError(pub String);
+
+impl<OM: Encodable> OutboundRouterMessage<OM> {
+    pub fn try_serialize(self) -> Result<Bytes, SerializeError> {
+        let mut buf = BytesMut::new();
+
+        let version = NetworkMessageVersion::version();
+        match self {
+            Self::AppMessage(app_message) => {
+                if version.compression_version == 1 {
```
michael-yxchen: we should make constants for these as well

monad-raptorcast/src/message.rs
bamrith: All the encoders use the output as `std::io::Write` and use only `write()` and `flush()`. I don't see any copying being done as they encode in internal buffers and only append to the output. Also `BytesMut` doesn't impl `std::io::Write`

PR #1768, opened 2025-05-02
Return an error when already servicing statesync request

maxkozlovsky:
Instead of queueing statesync requests return an error immediately. Queueing does not really work and can only result in timeout on the client and retry on a different server, causing the same request to be run several times across the cluster.

On the client keep a count of outstanding requests per server and don't send additional requests to the same server. If a server is already executing request from a different client, mark the server temporary inactive and retry request on a different server. Reactivate the server after random backoff time has passed.

### DIFF
--- a/monad-executor-glue/src/lib.rs
+++ b/monad-executor-glue/src/lib.rs
@@ -559,7 +559,9 @@ const STATESYNC_VERSION_V0: StateSyncVersion = StateSyncVersion { major: 1, mino
 const STATESYNC_VERSION_V1: StateSyncVersion = StateSyncVersion { major: 1, minor: 1 };
 // Client is required to send completions since this version
 pub const STATESYNC_VERSION_V2: StateSyncVersion = StateSyncVersion { major: 1, minor: 2 };
-pub const SELF_STATESYNC_VERSION: StateSyncVersion = STATESYNC_VERSION_V2;
+// Request includes session id allocated by client, mirrored by server response
+pub const STATESYNC_VERSION_V3: StateSyncVersion = StateSyncVersion { major: 1, minor: 3 };
+pub const SELF_STATESYNC_VERSION: StateSyncVersion = STATESYNC_VERSION_V3;
 pub const STATESYNC_VERSION_MIN: StateSyncVersion = STATESYNC_VERSION_V0;
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, RlpEncodable, RlpDecodable)]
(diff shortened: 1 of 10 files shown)


maxkozlovsky:
https://github.com/category-labs/category-internal/issues/971

monad-executor-glue/src/lib.rs
```diff
@@ -850,15 +923,92 @@ impl Debug for StateSyncResponse {
...
+            }
+            Self::DbError => {
+                let enc: [&dyn Encodable; 1] = [&3u8];
+                encode_list::<_, dyn Encodable>(&enc, out);
+            }
+        }
+    }
+
+    fn length(&self) -> usize {
+        // max enum value is << 127
```
dshulyak: < 127 ?

monad-state/src/lib.rs
```diff
@@ -396,6 +396,57 @@ where
...
+                StateSyncNetworkMessage::Response(_)
+                | StateSyncNetworkMessage::BadVersion(_)
+                | StateSyncNetworkMessage::ResponseError(_) => {
+                    vec![Command::StateSyncCommand(StateSyncCommand::Message((
+                        from, message,
+                    )))]
+                }
+                StateSyncNetworkMessage::Completion(_) => Vec::new(),
+            },
+            ConsensusMode::Live(ref mut consensus) => match message {
```
dshulyak: is it unused? can be (_)

monad-statesync/src/ffi.rs
```diff
@@ -387,6 +391,18 @@ impl<PT: PubKey> StateSync<PT> {
         self.outbound_requests.handle_bad_version(from, bad_version);
     }
 
+    pub fn handle_response_error(&mut self, from: NodeId<PT>, error: StateSyncError) {
+        if !self.state_sync_peers.iter().any(|trusted| trusted == &from) {
```
dshulyak, line 395: i am not familiar with statesync, generally i would expect that sender would track outbound requests for each peer and discard responses that came from unexpected peers. would make more sense then checking if they are trusted or not

monad-statesync/src/outbound_requests.rs
```diff
@@ -241,75 +306,121 @@ impl<PT: PubKey> OutboundRequests<PT> {
...
+    // Update version to the peer's version and insert to inflight requests
+    fn insert_request(
+        &mut self,
+        node: NodeId<PT>,
+        mut request: StateSyncRequest,
+    ) -> StateSyncRequest {
+        self.pending_requests.remove(&request);
+        let peer = self.peers.get_mut(&node).expect("peer not found");
+        request.version = peer.version;
+        request.session_id = SessionId(rand::random());
```
dshulyak, line 379: version and session_id are invalid before they are set here?
i think it is better to wrap the original request to avoid ambiguity

dshulyak (dismissed):

monad-statesync/src/ffi.rs
maxkozlovsky: Yes eventually we can do that, so far we need to keep it for backward compatibility. This pr adds session id to the client request which will allow to match responses to request. Once the older protocol versions which do not have session id are retired this code can be removed.

monad-executor-glue/src/lib.rs
maxkozlovsky: ok

monad-state/src/lib.rs
maxkozlovsky: ok

monad-statesync/src/outbound_requests.rs
maxkozlovsky: Request is initialized with default and then version is adjusted depending on the chosen peer. It is already a function parameter owned locally, I don't think if wrapping is necessary.

PR #1769, opened 2025-05-03
Fix eth call contract creation max code size

eerkaijun:
Closes https://github.com/category-labs/category-internal/issues/1524

### DIFF
--- a/monad-rpc/src/call.rs
+++ b/monad-rpc/src/call.rs
@@ -11,6 +11,7 @@ use std::{
 use alloy_consensus::{Header, SignableTransaction, TxEip1559, TxEnvelope, TxLegacy};
 use alloy_primitives::{Address, PrimitiveSignature, TxKind, Uint, U256, U64, U8};
 use alloy_rpc_types::AccessList;
+use monad_chain_config::execution_revision::MonadExecutionRevision;
 use monad_ethcall::{eth_call, CallResult, EthCallExecutor, StateOverrideSet};
 use monad_rpc_docs::rpc;
 use monad_triedb_utils::triedb_env::{
@@ -316,6 +317,8 @@ impl TryFrom<CallRequest> for TxEnvelope {
                     to: if let Some(to) = call_request.to {
                         TxKind::Call(to)
                     } else {
+                        // EIP-3860
+                        check_contract_creation_size(&call_request)?;
                         TxKind::Create
                     },
                     value: call_request.value.unwrap_or_default(),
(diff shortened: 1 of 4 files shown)


monad-rpc/src/call.rs
```diff
@@ -364,9 +373,9 @@ impl TryFrom<CallRequest> for TxEnvelope {
                     to: if let Some(to) = call_request.to {
                         TxKind::Call(to)
                     } else {
-                        // EIP-170
-                        if let Some(code) = call_request.input.data.as_ref() {
-                            if code.len() > 0x6000 {
+                        // EIP-3860
+                        if let Some(code) = call_request.input.input.as_ref() {
+                            if code.len() > 2 * MAX_CODE_SIZE {
```
dshulyak: this block and one that starts at L318 are identical, maybe can be moved into a helper

monad-rpc/src/call.rs
```diff
@@ -316,6 +318,13 @@ impl TryFrom<CallRequest> for TxEnvelope {
                     to: if let Some(to) = call_request.to {
                         TxKind::Call(to)
                     } else {
+                        // EIP-3860
+                        if let Some(code) = call_request.input.input.as_ref() {
```
philipglazman: The input field can be either `input` or `data` (for backwards compatability). Do we need to check `call_request.input.data` as well?  Potentially we can make a deserialization function for `CallInput` that accepts either `data` or `input`

monad-rpc/src/call.rs
```diff
@@ -364,9 +373,9 @@ impl TryFrom<CallRequest> for TxEnvelope {
                     to: if let Some(to) = call_request.to {
                         TxKind::Call(to)
                     } else {
-                        // EIP-170
-                        if let Some(code) = call_request.input.data.as_ref() {
-                            if code.len() > 0x6000 {
+                        // EIP-3860
+                        if let Some(code) = call_request.input.input.as_ref() {
```
philipglazman: same Q here as above

monad-rpc/src/call.rs
eerkaijun: we already sanitized `data` field to be `input` in https://github.com/category-labs/monad-bft/blob/bd8f0ac7a839b7bb6aae773af99b30f9abaeb775/monad-rpc/src/call.rs#L385-L396

monad-rpc/src/call.rs
```diff
@@ -30,6 +30,8 @@ use crate::{
     trace_handlers::{decode_call_frame, MonadCallFrame, TracerObject},
 };
 
+const MAX_CODE_SIZE: usize = 128 * 1024;
```
ariqchowdhury: are there other places the max-code-size is used in bft?

eerkaijun: now importing from monad-chain-config

PR #1770, opened 2025-05-05
monad-pprof, monad-rpc: jemallocator to use in monad-rpc and pprof integration for heap profiling

dshulyak:
- [x] followup to integrate with monad-node

it allows to expose heap profile over http api. instructions are available in readme.
but also jemmalloc is known to be more efficient than the std allocator, for example here is a latency distribution of getBalance requests. 

image


Cargo.toml
```diff
@@ -197,3 +199,5 @@ vergen-git2 = "1.0.5"
 wasm-bindgen = "0.2"
 zerocopy = "0.6"
 zeroize = "1.3"
+tikv-jemallocator = "0.6"
+jemalloc_pprof = "0.7"
```
dshulyak: need to sort it

dshulyak:
disabled it by default. i think other may run it when they will be reporting memory related bugs

PR #1777, opened 2025-05-07
Use os time directly for block timestamp

mikety:
This is the 2nd PR out of the 3 PRs implementing clock adjustment for accurate timestamp estimation.
the other 2 are:

1st: "Implement periodic pings to calculate network latency" : https://github.com/category-labs/monad-bft/pull/1675
3rd: "Use proposal latency to compute timestamp adjustment" : https://github.com/category-labs/monad-bft/pull/1685

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -1,4 +1,4 @@
-use std::{collections::BTreeMap, fmt::Debug, marker::PhantomData, time::Duration};
+use std::{cmp::max, collections::BTreeMap, fmt::Debug, marker::PhantomData, time::Duration};
 
 use monad_blocktree::blocktree::BlockTree;
 use monad_chain_config::{
@@ -133,7 +133,7 @@ enum OutgoingVoteStatus {
     VoteReady(Vote),
 }
 
-pub struct ConsensusStateWrapper<'a, ST, SCT, EPT, BPT, SBT, VTF, LT, BVT, CCT, CRT, TS>
+pub struct ConsensusStateWrapper<'a, ST, SCT, EPT, BPT, SBT, VTF, LT, BVT, CCT, CRT, CL>
 where
     ST: CertificateSignatureRecoverable,
     SCT: SignatureCollection<NodeIdPubKey = CertificateSignaturePubKey<ST>>,
@@ -145,7 +145,7 @@ where
     BVT: BlockValidator<ST, SCT, EPT, BPT, SBT>,
     CCT: ChainConfig<CRT>,
     CRT: ChainRevision,
-    TS: Clock,
+    CL: Clock,
 {
     pub consensus: &'a mut ConsensusState<ST, SCT, EPT, BPT, SBT, CCT, CRT>,
 
@@ -163,7 +163,7 @@ where
     /// Policy for validating incoming proposals
     pub block_validator: &'a BVT,
     /// Track local timestamp and validate proposal timestamps
-    pub block_timestamp: &'a mut BlockTimestamp<SCT::NodeIdPubKey, TS>,
+    pub block_timestamp: &'a mut BlockTimestamp<SCT::NodeIdPubKey, CL>,
 
     /// Destination address for proposal payments
     pub beneficiary: &'a [u8; 20],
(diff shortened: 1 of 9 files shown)


monad-consensus-state/src/lib.rs
```diff
@@ -131,7 +133,7 @@ enum OutgoingVoteStatus {
     VoteReady(Vote),
 }
 
-pub struct ConsensusStateWrapper<'a, ST, SCT, EPT, BPT, SBT, VTF, LT, BVT, CCT, CRT>
+pub struct ConsensusStateWrapper<'a, ST, SCT, EPT, BPT, SBT, VTF, LT, BVT, CCT, CRT, TS>
```
andr-dev: Let's use `CL` for consistency with below. I think `CL` makes more sense if you want to call the trait `Clock`. If you want to go with a trait name adjacent to "timestamp" then `TS` is more fitting. That being said, from the SystemClock impl, I think `Clock` is a good name :)

monad-consensus-state/src/timestamp.rs
```diff
@@ -227,55 +231,79 @@ struct SentVote {
...
-        if self.local_time_ns <= prev_block_ts {
-            prev_block_ts + 1
+        let now = self.clock.get();
+        if let Some(adjuster) = &self.adjuster {
+            let adjustment = adjuster.get_adjustment();
+            if adjustment >= 0 {
+                now.checked_add(adjustment as u128).unwrap_or(now)
+            } else {
+                now.saturating_sub(adjustment.unsigned_abs() as u128)
+            }
```
andr-dev, line 300: Instead of having TimestampAdjuster expose the current adjustment through a getter, it might be cleaner to expose a function which takes a timestamp and adjusts it so that the internals of the TimestampAdjuster remain "hidden".

monad-consensus-state/src/timestamp.rs
```diff
@@ -227,55 +231,79 @@ struct SentVote {
...
-        let lower_bound = self.local_time_ns.saturating_sub(max_delta_ns);
-        let upper_bound = self.local_time_ns.saturating_add(max_delta_ns);
+        let max_delta_ns = self.max_delta_ns.saturating_add(vote_delay_ns);
+        let lower_bound = now.saturating_sub(self.max_delta_ns);
+        let upper_bound = now.saturating_add(self.max_delta_ns);
 
         lower_bound <= timestamp && timestamp <= upper_bound
     }
 
+    pub fn handle_adjustment(&mut self, delta: TimestampAdjustment) {
```
andr-dev: Where is this function being used? I'm seeing that it has been added in this changeset but a couple of previous usages directly on the `adjuster` have been deleted.

monad-consensus-state/src/timestamp.rs
```diff
@@ -227,55 +231,79 @@ struct SentVote {
...
+        if let Some(adjuster) = &mut self.adjuster {
+            adjuster.handle_adjustment(delta);
+        }
+    }
+
+    pub fn get_valid_block_timestamp(&self, prev_block_ts: u128) -> u128 {
```
andr-dev: Is there some semantic difference between retrieving a block timestamp vs a valid block timestamp? If not, then I would remove the `valid` from this function name. Given that it's likely being called when creating a new block and intuitively uses `prev_block_ts`, maybe it should be called `get_next_block_timestamp`?

monad-consensus-state/src/timestamp.rs
```diff
@@ -743,7 +781,7 @@ mod test {
 
     #[test]
     fn test_new_epoch() {
-        let mut b = BlockTimestamp::<NopPubKey>::new(10, 1);
+        let mut b = BlockTimestamp::<NopPubKey, TestClock>::new(10, 1, AdjusterConfig::Disabled);
```
andr-dev: We should add a unit test with `AdjusterConfig::Enabled`. Hopefully, mock swarm should catch if something goes wrong with the `Enabled` case but it'll be harder to pinpoint through that test environment.

monad-state/src/lib.rs
```diff
@@ -767,10 +771,10 @@ where
     CCT: ChainConfig<CRT>,
     CRT: ChainRevision,
 {
-    pub fn build(
+    pub fn build<CL: Clock>(
```
andr-dev, line 775: It seems reasonable to add `CL` to `MonadStateBuilder`, what do you think?

mikety: I will add a TODO for the separate PR which  will add a few swarm tests as well.

monad-consensus-state/src/timestamp.rs
mikety: There is a reference to it in the monad-consensus-state/src/lib.rc:598.
It will all come together in the follow up PR, this PR focuses on adding the OS time to blockstamp.

monad-consensus-state/src/timestamp.rs
mikety: It will be better to address in the follow-up PR, where the adjuster, proposal response duration, network latency, and proposal creation average latency will be put together.

monad-consensus-state/src/timestamp.rs
mikety: It will be better to address in the follow-up PR, where the adjuster, proposal response duration, network latency, and proposal creation average latency will be put together.

monad-consensus-state/src/timestamp.rs
mikety: @andr-dev , updated.

PR #1778, opened 2025-05-08
Peer discovery driver

eerkaijun:
splitting out the peer discovery driver and router integration so they can be reviewed separately

### DIFF
--- a/monad-peer-discovery/src/driver.rs
+++ b/monad-peer-discovery/src/driver.rs
@@ -1,0 +1,223 @@
+use std::{
+    collections::{HashMap, VecDeque},
+    ops::DerefMut,
+    task::{Context, Poll},
+    time::Duration,
+};
+
+use futures::{Stream, StreamExt};
+use monad_crypto::certificate_signature::{
+    CertificateSignaturePubKey, CertificateSignatureRecoverable,
+};
+use monad_types::NodeId;
+use tokio_util::time::{DelayQueue, delay_queue::Key};
+use tracing::error;
+
+use crate::{
+    PeerDiscoveryAlgo, PeerDiscoveryCommand, PeerDiscoveryEvent, PeerDiscoveryMessage,
+    PeerDiscoveryTimerCommand, TimerKind,
+};
+
+pub enum PeerDiscoveryEmit<ST: CertificateSignatureRecoverable> {
+    // TODO: other output events
+    RouterCommand {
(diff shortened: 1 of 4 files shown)


monad-peer-discovery/src/driver.rs
```diff
@@ -0,0 +1,218 @@
...
+
+    fn schedule_reset(
+        &mut self,
+        node_id: NodeId<CertificateSignaturePubKey<ST>>,
+        timer_kind: TimerKind,
+    ) {
+        if let Some((key, _event)) = self.events.remove(&(node_id, timer_kind)) {
+            // DelayQueue timer panics if key is not found, which indicates a
+            // logic error - inconsistency between timers and events
+            self.timers.remove(&key);
```
mikety, line 71: should warn or assert if None

monad-peer-discovery/src/driver.rs
```diff
@@ -0,0 +1,218 @@
...
+        }
+    }
+}
+
+pub struct PeerDiscoveryDriver<PD>
+where
+    PD: PeerDiscoveryAlgo,
+{
+    pd: PD,
+    timer: PeerDiscTimers<PD::SignatureType>,
```
mikety, line 127: nit: a bit misleading name : timers or schedule is more descriptive

monad-peer-discovery/src/driver.rs
```diff
@@ -0,0 +1,218 @@
...
+        Self {
+            pd: peer_discovery,
+            timer: Default::default(),
+            pending_emits: Default::default(),
+        }
+    }
+
+    pub fn update(&mut self, event: PeerDiscoveryEvent<PD::SignatureType>) {
+        let cmds = match event {
+            PeerDiscoveryEvent::SendPing { to } => self.pd.send_ping(to),
```
mikety, line 142: nit: there are no any metrics or logging, will it be helpful to have metrics?

eerkaijun: metrics are recorded in the corresponding peer discovery functions, i.e. in `send_ping` function

monad-peer-discovery/src/driver.rs
eerkaijun: the DelayQueue will panic if `key` is not present

eerkaijun: added an error log

PR #1779, opened 2025-05-08
Full-node Raptorcast: group map utilities & test

fabiotdf:
This is part of a series of PRs that split a larger PR for Full-Node RaptorCast integration

### DIFF
--- a/monad-raptorcast/Cargo.toml
+++ b/monad-raptorcast/Cargo.toml
@@ -30,6 +30,8 @@ rand = { workspace = true }
 tracing = { workspace = true }
 
 [dev-dependencies]
+monad-testutil = { workspace = true }
+
 clap = { workspace = true, features = ["derive"] }
 criterion = { workspace = true }
 futures-util = { workspace = true }
--- a/monad-raptorcast/src/util.rs
+++ b/monad-raptorcast/src/util.rs
@@ -1,12 +1,13 @@
-use std::collections::BTreeMap;
+use std::{collections::BTreeMap, fmt};
 
+use itertools::Itertools;
 use monad_crypto::{
     certificate_signature::{CertificateSignaturePubKey, CertificateSignatureRecoverable, PubKey},
     hasher::{Hasher, HasherType},
 };
-use monad_types::{NodeId, Stake};
+use monad_types::{Epoch, NodeId, Round, RoundSpan, Stake};
 
-#[derive(Clone)]
+#[derive(Clone, Default)]
 pub struct EpochValidators<ST>
 where
     ST: CertificateSignatureRecoverable,
@@ -102,6 +103,7 @@ pub struct Validator {
     pub stake: Stake,
 }
 
+// Argument for raptorcast send
 #[derive(Debug)]
 pub enum BuildTarget<'a, ST: CertificateSignatureRecoverable> {
     Broadcast(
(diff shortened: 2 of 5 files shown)


monad-raptorcast/src/util.rs
```diff
@@ -37,6 +37,17 @@ where
...
+where
+    ST: CertificateSignatureRecoverable,
+{
+    fn default() -> Self {
+        Self {
+            validators: BTreeMap::new(),
+        }
+    }
+}
+
```
eerkaijun: You can just `#[derive(Default)]` for EpochValidators

monad-raptorcast/src/util.rs
```diff
@@ -143,3 +155,523 @@ impl<const N: usize> std::fmt::Debug for HexBytes<N> {
 
 pub type NodeIdHash = HexBytes<20>;
 pub type AppMessageHash = HexBytes<20>;
+pub type NodeIdST<ST> = NodeId<CertificateSignaturePubKey<ST>>;
```
eerkaijun: We use `NodeId<CertificateSignaturePubKey<ST>>` in other places, so not sure if we want this type alias. If we do eventually, we would want to replace them in other places too.

monad-raptorcast/src/util.rs
```diff
@@ -143,3 +155,523 @@ impl<const N: usize> std::fmt::Debug for HexBytes<N> {
...
+// that contains stake info per validator.
+#[derive(Clone, PartialEq, Eq)]
+pub struct Group<ST>
+where
+    ST: CertificateSignatureRecoverable,
+{
+    // The node_id of the validator publishing to full-nodes.
+    validator_id: Option<NodeIdST<ST>>,
+    round_span: RoundSpan,
+    all_peers: Vec<NodeIdST<ST>>, // Includes self (+ author_id if re-broadcasting to validators).
```
eerkaijun: is there any reason for needing to store self in `all_peers` too rather than being a separate field? seems like subsequently there are a bunch of filtering functions to extract self

monad-raptorcast/src/util.rs
```diff
@@ -143,3 +155,523 @@ impl<const N: usize> std::fmt::Debug for HexBytes<N> {
 
 pub type NodeIdHash = HexBytes<20>;
 pub type AppMessageHash = HexBytes<20>;
+pub type NodeIdST<ST> = NodeId<CertificateSignaturePubKey<ST>>;
+
+// This represents a raptorcast group abstracted over the use cases below:
+// 1) Validator->Validator raptorcast recv & re-broadcast
+// 2) Validator->FullNode raptorcast recv & re-broadcast
```
michael-yxchen, line 151: is 2) a legit case? I thought that's handled by a list of forward peers. We're missing Fullnode->Fullnode raptorcast recv & rebroadcast use case here

monad-raptorcast/src/util.rs
```diff
@@ -143,3 +155,523 @@ impl<const N: usize> std::fmt::Debug for HexBytes<N> {
...
+pub struct Group<ST>
+where
+    ST: CertificateSignatureRecoverable,
+{
+    // The node_id of the validator publishing to full-nodes.
+    validator_id: Option<NodeIdST<ST>>,
+    round_span: RoundSpan,
+    all_peers: Vec<NodeIdST<ST>>, // Includes self (+ author_id if re-broadcasting to validators).
+    self_id_index: usize, // So the iterator can skip our own node_id when iterating for re-broadcast peers
+}
```
michael-yxchen: I'm inclined not to merge rebroadcast group representation for validator->validator and fullnode->fullnode rebroadcast path. Special values for `validator_id` and `round_span` can be more easily enforced if we create different types altogether

my understanding is that `udp.handle_message()` function iterates through the group iterator returned by `iterate_rebroadcast_peers`. we can maintain different group representations, but expose same group iterator interface

monad-raptorcast/src/util.rs
```diff
@@ -143,3 +155,523 @@ impl<const N: usize> std::fmt::Debug for HexBytes<N> {
...
+            round_span: RoundSpan::default(),
+            all_peers,
+            self_id_index,
+        }
+    }
+
+    // For the use case where we re-raptorcast to full-nodes
+    pub fn new_fullnode_group(
+        all_peers: Vec<NodeIdST<ST>>,
+        self_id_opt: Option<&NodeIdST<ST>>,
```
michael-yxchen: when would we set it to None?

monad-raptorcast/src/util.rs
```diff
@@ -143,3 +155,523 @@ impl<const N: usize> std::fmt::Debug for HexBytes<N> {
...
+{
+    // For the use case where we re-raptorcast to validators
+    pub fn new_validator_group(
+        all_peers: Vec<NodeIdST<ST>>, // enough only for re-broadcasting
+        self_id: &NodeIdST<ST>,
+    ) -> Self {
+        let self_id_index = all_peers
+            .iter()
+            .position(|peer| peer == self_id)
+            .unwrap_or(usize::MAX);
```
michael-yxchen: `self_id_index` is determined at construction time. Would it be easier if we remove `self_id` from the list? A normal vector iterator can replace `GroupIterator` for fullnode->fullnode rebroadcast. it might be slightly complicated for validator<>validator rebroadcast though because we need to remove a dynamic author

monad-raptorcast/src/util.rs
```diff
@@ -143,3 +155,523 @@ impl<const N: usize> std::fmt::Debug for HexBytes<N> {
...
+
+    // There are cases where we need to check that the source node is valid
+    // before we get to call iter_skip_self_and_author()
+    pub fn check_author_node_id(&self, author_id: &NodeIdST<ST>) -> bool {
+        if let Some(root_vid) = self.validator_id {
+            // Case for full-node raptorcasting
+            &root_vid == author_id
+        } else {
+            // Case for validator-to-validator
+            self.all_peers.contains(author_id)
```
michael-yxchen: we're gonna call this function pretty often. Let's switch all_peers to a `BTreeeMap` for faster membership check

monad-raptorcast/src/util.rs
fabiotdf: Cool, thanks!

monad-raptorcast/src/util.rs
fabiotdf: It is just constructed like this all the way from when the validator first created the group, and I wanted to avoid all the copying and  rebuilding that was happening to the peer list.
The iterator has to skip author ID as well, so the filtering mechanism is needed anyway.

monad-raptorcast/src/util.rs
fabiotdf: It makes the code much more readable though...

fabiotdf: Removed

eerkaijun: agree, we should probably switch them all out at some point

monad-raptorcast/src/util.rs
```diff
@@ -143,3 +144,526 @@ impl<const N: usize> std::fmt::Debug for HexBytes<N> {
...
+            RoundSpan::new(Round(3), Round(8)),
+        );
+        assert_eq!(group.size_excl_self(), 0);
+        assert!(group.contains_self());
+        assert!(group.is_empty_besides_self());
+        assert_eq!(group.get_validator_id(), &nid(3));
+        assert_eq!(group.get_all_peers(), &vec![nid(1)]);
+        assert_eq!(group.get_round_span(), &RoundSpan::new(Round(3), Round(8)));
+        let empty_iter: Vec<_> = group.empty_iterator().collect();
+        assert_eq!(empty_iter.len(), 0);
```
michael-yxchen: super nit: `is_empty`

monad-raptorcast/src/util.rs
```diff
@@ -143,3 +144,526 @@ impl<const N: usize> std::fmt::Debug for HexBytes<N> {
...
+        assert_eq!(&it, &vec![nid(1), nid(2), nid(3), nid(4)]);
+
+        // "Randomized" iterations
+        let mut permutations_seen = HashSet::<Vec<NodeId<CertificateSignaturePubKey<ST>>>>::new();
+        for seed in 1..10 {
+            let it: Vec<_> = group
+                .iter_skip_self_and_author(&nid(100), seed)
+                .cloned()
+                .collect();
+            permutations_seen.insert(it.clone());
```
michael-yxchen: nit: unnecessary clone

monad-raptorcast/src/util.rs
fabiotdf: The full-nodes do not initiate raptorcast, do they? They only re-raptorcast.
Did I misunderstand it?

Regarding forward peers. My understanding is that from config, we wanted 2 list of full-nodes.

a) "dedicated" full nodes, AKA forward peers. This uses the old mechanism already present in primary raptorcast, where full proposals are forwarded to full-nodes.

b) "prioritized" full nodes. These full nodes are always asked first when a validator is forming a new raptorcast group intended to be used for full-nodes. This is the list of peers that `Group<ST>` represents. Initially, it is a validator that is raptor-casting to the full-nodes. As full-nodes receive this raptorcasting, they also re-raptorcast to other fullnodes, using the same `Group<>`. In my interpretation, We're still in case 2. Should I change the code comment so that case 2 is called "Full-node->Full-node"?

monad-raptorcast/src/util.rs
fabiotdf: How about we sort the peers when the validator creates the group?
Then we can just do a binary search here, which is even more efficient than `BTreeSet`, as its all contiguous in memory, and keeps the random iterator simple and efficient (avoids chunks always being sent to same node)

monad-raptorcast/src/util.rs
fabiotdf: It would be a pessimization though, since we'd be moving large elements around. Specially in the dynamic author case.
This filtering mechanism is lightweight and avoids the large and constant copying/rebuilding of peerlist that was is the old code.
Let me know if you still want this, despite the overhead

monad-raptorcast/src/util.rs

monad-raptorcast/src/util.rs

monad-raptorcast/src/util.rs
fabiotdf: When the validator is creating a full-node group, `self_id_opt` is `None`.
The ID of the validator itself goes into argument `validator_id`

monad-raptorcast/src/util.rs
michael-yxchen: got it, I misunderstood. I thought you were referring to validator servicing "dedicated" full nodes. Keeping it as "Validator->FullNode" is fine

monad-raptorcast/src/util.rs
michael-yxchen: Would it make sense to set both `self_id_opt` and `validator_id` to the validator itself? We don't need to change it now. we can do it during the type refactor

monad-raptorcast/src/util.rs
michael-yxchen: cool, let's keep your impl

monad-raptorcast/src/util.rs
michael-yxchen: sounds good. `SortedVec/SortedSet` might be helpful

monad-raptorcast/src/util.rs
fabiotdf: I think this is probably making `Group<>` ultra-specific on its context use...

I'd need to also create a new `ReBroadcastGroupMap<>` group map type that is a container for the new group type
Then, I'd have to pass both group map types to `udp.handle_message()`
Then, inside handle_message I'd need to duplicate code that is otherwise identical except for the group map type.

```
    pub fn handle_message(
        &mut self,
        group_map: &ReBroadcastGroupMap<ST>, // (msg.epoch, msg.author_id) -> Group. Replaces `epoch_validators`
      // would need group_map_2 here for the new type
        rebroadcast: impl FnMut(Vec<NodeId<CertificateSignaturePubKey<ST>>>, Bytes, u16), // RC chunks to other peers
        forward: impl FnMut(Bytes, u16), // full messages to dedicated full-nodes
        message: RecvMsg,                // chunk from UDP recv
    ) -> Vec<(NodeId<CertificateSignaturePubKey<ST>>, Bytes)> // returns app messages decoded from raptor-10
```

What do you think?

monad-raptorcast/src/util.rs
fabiotdf: Sounds good! Fixed

monad-raptorcast/src/util.rs
fabiotdf: Done

monad-raptorcast/src/util.rs
fabiotdf: Now that the peers are getting sorted anyway (for efficient `author_id` lookup), `self_id` is removed from all_peers

monad-raptorcast/src/util.rs
```diff
@@ -143,3 +145,522 @@ impl<const N: usize> std::fmt::Debug for HexBytes<N> {
...
+        let mut sorted_other_peers = all_peers;
+        if self_id != &validator_id {
+            // Swap self_id in `all_peers` with the last element, then pop.
+            let self_ix = sorted_other_peers
+                .iter()
+                .position(|peer| peer == self_id)
+                .expect(
+                    "Could not find own node id when instantiating a \
+                        Raptorcast group for full-nodes",
+                );
```
michael-yxchen, line 225: reminder to check `all_peers.contains(self_id)` in caller

monad-raptorcast/src/util.rs
```diff
@@ -143,3 +145,522 @@ impl<const N: usize> std::fmt::Debug for HexBytes<N> {
...
+            round_span: RoundSpan::default(),
+            sorted_other_peers,
+        }
+    }
+
+    // For the use case where we re-raptorcast to full-nodes
+    // Validators Raptorcasting to full-nodes should set `self_id` == `validator_id`
+    pub fn new_fullnode_group(
+        all_peers: Vec<NodeId<CertificateSignaturePubKey<ST>>>,
+        self_id: &NodeId<CertificateSignaturePubKey<ST>>,
```
michael-yxchen, line 209: nit: is reference needed?

monad-raptorcast/src/util.rs
```diff
@@ -143,3 +145,522 @@ impl<const N: usize> std::fmt::Debug for HexBytes<N> {
...
+            let self_ix = sorted_other_peers
+                .iter()
+                .position(|peer| peer == self_id)
+                .expect(
+                    "Could not find own node id when instantiating a \
+                        Raptorcast group for full-nodes",
+                );
+            let last_ix = sorted_other_peers.len() - 1;
+            sorted_other_peers.swap(self_ix, last_ix);
+            sorted_other_peers.pop();
```
michael-yxchen: super nit: `sorted_other_peers.swap_remove(self_ix)`

monad-raptorcast/src/util.rs
michael-yxchen: pasting from slack

I don’t think we need an extra `group_map_2` . [Containers for v->v/v->fn](https://github.com/category-labs/monad-bft/pull/1779/files#diff-51f39c3513b760064bef16aee658a1a480fef63221f21b3b22d3e846dabffd5bR379-R383) can be different but we can unify them at the iterator level, i.e. return value of `iterate_rebroadcast_peers` conceptually both are `impl std::iter::Iterator<Item=NodeId>`

monad-raptorcast/src/util.rs
```diff
@@ -143,3 +145,522 @@ impl<const N: usize> std::fmt::Debug for HexBytes<N> {
...
+    pub fn new_fullnode_group(
+        all_peers: Vec<NodeId<CertificateSignaturePubKey<ST>>>,
+        self_id: &NodeId<CertificateSignaturePubKey<ST>>,
+        validator_id: NodeId<CertificateSignaturePubKey<ST>>,
+        round_span: RoundSpan,
+    ) -> Self {
+        // We will call `check_author_node_id()` often, so sorting here will
+        // allow us to use binary search instead of linear search.
+        let mut sorted_other_peers = all_peers;
+        if self_id != &validator_id {
```
michael-yxchen, line 216: why do we need this check

fabiotdf: Because if we are a validator, we won't find our node ID among the peers and that will cause panic

monad-raptorcast/src/util.rs
fabiotdf: But we are already looking for self_id among `sorted_other_peers`, and if that's not found, panic

monad-raptorcast/src/util.rs
fabiotdf: Nice, thanks!

monad-raptorcast/src/util.rs
fabiotdf: My impression was that if you don't need to store the value, then a reference (pointer) is preferable, so we aren't moving a NodeId, which is quite big. Here we just need to find `self_id` in `all_peers` and don't need to store it. In contracts, we do store `validator_id`, so we move that value in

monad-raptorcast/src/util.rs
fabiotdf: How would the new signature for `handle_message` look like then? 

Currently (in my other PR) it's passing a `ReBroadcastGroupMap<ST>` as copied above. 
Note that `handle_message()` needs to look up `epoch` and `author_id` into the group map, you can't look up in advance and pass an iterator, because depending on the epoch and on author_id, the iterator will be iterating over different nodeId sets

monad-raptorcast/src/util.rs
michael-yxchen: we receive `all_peers` from the network. a malicious sender can exclude the node_id from the message and cause our client to panic, if we don't verify. Panic should only happen if there's a logical error, not to assert on unverified data received over the network

monad-raptorcast/src/util.rs
michael-yxchen: that's why I suggested changing the return value of `iterate_rebroadcast_peers`. I haven't thought too carefully but I think you can use the same iterator type for both v->v/v->fn containers as both are backed by a vector

monad-raptorcast/Cargo.toml
```diff
@@ -18,6 +18,7 @@ monad-executor-glue = { workspace = true }
 monad-merkle = { workspace = true }
 monad-raptor = { workspace = true }
 monad-secp = { workspace = true }
+monad-testutil = { workspace = true }
```
eerkaijun: nit: this can be moved into [dev-dependencies]

monad-raptorcast/src/util.rs
```diff
@@ -143,3 +145,520 @@ impl<const N: usize> std::fmt::Debug for HexBytes<N> {
...
+    ST: CertificateSignatureRecoverable,
+{
+    fn default() -> Self {
+        Self {
+            validator_id: None,
+            round_span: RoundSpan::default(),
+            sorted_other_peers: Vec::new(),
+        }
+    }
+}
```
eerkaijun: we can just `#[derive(Default)]` for Group

monad-raptorcast/src/util.rs
```diff
@@ -143,3 +145,520 @@ impl<const N: usize> std::fmt::Debug for HexBytes<N> {
...
+        self_id: &NodeId<CertificateSignaturePubKey<ST>>,
+    ) -> Self {
+        // We will call `check_author_node_id()` often, so sorting here will
+        // allow us to use binary search instead of linear search.
+        let sorted_other_peers: Vec<_> = all_peers
+            .iter()
+            .filter(|peer| *peer != self_id)
+            .sorted()
+            .cloned()
+            .collect();
```
eerkaijun: nit: remove cloning and `sorted_other_peers` type should also be automatically inferred:

```
let sorted_other_peers = all_peers
    .into_iter()
    .filter(|peer| peer != self_id)
    .sorted()
    .collect();
```

monad-raptorcast/src/util.rs
```diff
@@ -143,3 +145,520 @@ impl<const N: usize> std::fmt::Debug for HexBytes<N> {
...
+        self_id: &NodeId<CertificateSignaturePubKey<ST>>,
+        validator_id: NodeId<CertificateSignaturePubKey<ST>>,
+        round_span: RoundSpan,
+    ) -> Self {
+        // We will call `check_author_node_id()` often, so sorting here will
+        // allow us to use binary search instead of linear search.
+        let mut sorted_other_peers = all_peers;
+        if self_id != &validator_id { // The validator won't find its own nodeid among the peers
+            // Swap self_id in `all_peers` with the last element, then pop.
+            let self_ix = sorted_other_peers
```
eerkaijun: naming: is this supposed to be `self_idx`?

eerkaijun:
to squash before merging

monad-raptorcast/src/util.rs
fabiotdf: Ah yes, that makes sense! Thanks for the clarification
Fixed this in the upcoming PR, since otherwise if fixed here, I'd have to allow construction of an invalid `Group<>`.
The client full-node will now ignore a `ConfirmGroup` if self_id isn't listed in the group

monad-raptorcast/Cargo.toml
fabiotdf: Fixed

monad-raptorcast/src/util.rs

monad-raptorcast/src/util.rs

monad-raptorcast/src/util.rs
fabiotdf: I don't see it ATM but let's catch up on this in the upcoming UdpState PR?

monad-raptorcast/src/util.rs
fabiotdf: Fixed, thanks!

PR #1780, opened 2025-05-09
Peer discovery integration

eerkaijun:
Integration of peer discovery into router. Currently initial set of peers are loaded through node.toml (kept as a separate field for testing but should be combined with bootstrap peers eventually). 

Manually tested with flexnet that it works, to come up with more robust flexnet tests


monad-node/src/main.rs
```diff
@@ -525,6 +570,29 @@ where
...
+        ),
+        seq: 0, // TODO: configurable
+    };
+    let mut encoded = Vec::new();
+    name_record.encode(&mut encoded);
+    let signature = ST::sign(&encoded, identity);
+    MonadNameRecord {
+        name_record,
+        signature,
+    }
```
michael-yxchen: use `MonadNameRecord::new`

monad-node/src/main.rs
```diff
@@ -507,13 +513,52 @@ where
     <M as Deserializable<Bytes>>::ReadError: 'static,
     OM: Serializable<Bytes> + Encodable + Clone + Send + Sync + 'static,
 {
+    let peer_info = peer_discovery_config
+        .peers
+        .iter()
+        .filter_map(|peer| {
+            let node_id = NodeId::new(peer.secp256k1_pubkey);
+            let maybe_addr = resolve_domain_v4(&node_id, &peer.address);
```
michael-yxchen: `peer.address` should be an ipv4 address, not hostname. otherwise signature can be invalidated if DNS record is changed

monad-node/src/main.rs
```diff
@@ -507,13 +513,52 @@ where
...
+                    PeerInfo {
+                        last_ping: None,
+                        unresponsive_pings: 0,
+                        name_record: MonadNameRecord {
+                            name_record: NameRecord {
+                                address: addr_v4,
+                                seq: peer.record_seq_num,
+                            },
+                            signature: peer.name_record_sig,
+                        },
```
michael-yxchen: we need to verify the signature, and log if it fails

monad-node/src/main.rs
```diff
@@ -507,13 +513,52 @@ where
...
+                    },
+                ))
+            } else {
+                None
+            }
+        })
+        .collect();
+
+    // TODO: make it configurable in node.toml
```
michael-yxchen, line 563: when we move peer discovery configs to node.toml, I think we should put self name record in config file, with signature, and verify on client startup. Name record might be different from bind address if there's some special router setup

monad-node-config/src/peers.rs
```diff
@@ -0,0 +1,33 @@
...
+#[serde(deny_unknown_fields)]
+pub struct PeerConfig<ST: CertificateSignatureRecoverable> {
+    #[serde(bound = "ST: CertificateSignatureRecoverable")]
+    pub peers: Vec<PeerDiscoveryConfig<ST>>,
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct PeerDiscoveryConfig<ST: CertificateSignatureRecoverable> {
+    pub address: String,
```
michael-yxchen: noted in monad-node main: address should be ipv4 address. we can consider changing the type here

monad-raptorcast/src/lib.rs
```diff
@@ -93,23 +103,31 @@ pub enum RaptorCastEvent<E, P: PubKey> {
...
+        B: PeerDiscoveryAlgoBuilder<PeerDiscoveryAlgoType = PD>,
+    {
         let self_id = NodeId::new(config.key.pubkey());
         let mut builder = DataplaneBuilder::new(&config.local_addr, config.up_bandwidth_mbps);
         if let Some(buffer_size) = config.buffer_size {
             builder = builder.with_buffer_size(buffer_size);
         }
         let dataplane = builder.build();
+
+        let (peer_discovery, init_cmds) = config.peer_discovery_builder.build();
```
michael-yxchen: shall we pass the builder to `PeerDiscoveryDriver::new()` and handle everything there?

monad-peer-discovery/src/mock.rs
```diff
@@ -20,18 +20,12 @@ struct PeerState<ST: CertificateSignatureRecoverable> {
 // Periodically pings the peer. Peers responds with a pong. It's not an actual
 // peer discovery implementation
 pub struct PingPongDiscovery<ST: CertificateSignatureRecoverable> {
-    self_id: NodeId<CertificateSignaturePubKey<ST>>,
-    local_name_record: MonadNameRecord<ST>,
     peer_state: BTreeMap<NodeId<CertificateSignaturePubKey<ST>>, PeerState<ST>>,
-
     ping_period: Duration,
-
```
michael-yxchen, line 28: why is self id and name record removed?

monad-raptorcast/src/lib.rs
```diff
@@ -35,13 +40,15 @@ use util::{BuildTarget, EpochValidators, FullNodes, Validator};
 
 const SIGNATURE_SIZE: usize = 65;
 
-pub struct RaptorCastConfig<ST>
+pub struct RaptorCastConfig<ST, B>
 where
     ST: CertificateSignatureRecoverable,
+    B: PeerDiscoveryAlgoBuilder,
```
michael-yxchen: I think we can remove this trait bound in struct

monad-raptorcast/src/lib.rs
```diff
@@ -136,7 +154,7 @@ where
     ) {
         match self.known_addresses.get(to) {
```
michael-yxchen, line 159: what's the plan to switch addr lookup to peer discovery?

monad-raptorcast/src/lib.rs
```diff
@@ -501,6 +542,39 @@ where
                         RaptorCastEvent::Message(message.event(NodeId::new(from))).into(),
                     ));
                 }
+                InboundRouterMessage::PeerDiscoveryMessage(_) => {
+                    // peer discovery message should come through udp
```
michael-yxchen, line 567: let's add debug/trace log here

monad-testground/src/executor.rs
```diff
@@ -26,17 +30,18 @@ use monad_updaters::{
 use monad_validator::{simple_round_robin::SimpleRoundRobin, validator_set::ValidatorSetFactory};
 use tracing_subscriber::EnvFilter;
 
-pub enum RouterConfig<ST, SCT, EPT>
+pub enum RouterConfig<ST, SCT, EPT, B>
 where
     ST: CertificateSignatureRecoverable,
     SCT: SignatureCollection<NodeIdPubKey = CertificateSignaturePubKey<ST>>,
     EPT: ExecutionProtocol,
+    B: PeerDiscoveryAlgoBuilder,
```
michael-yxchen: nit: i think this can also be removed after we remove it from `RaptorCastConfig`

monad-testground/src/executor.rs
```diff
@@ -53,13 +58,14 @@ where
     },
 }
 
-pub struct ExecutorConfig<ST, SCT, EPT>
+pub struct ExecutorConfig<ST, SCT, EPT, B>
 where
     ST: CertificateSignatureRecoverable,
     SCT: SignatureCollection<NodeIdPubKey = CertificateSignaturePubKey<ST>>,
     EPT: ExecutionProtocol,
+    B: PeerDiscoveryAlgoBuilder,
```
michael-yxchen: same here

monad-testground/src/executor.rs
```diff
@@ -104,6 +110,7 @@ where
                 MonadMessage<ST, SCT, MockExecutionProtocol>,
                 VerifiedMonadMessage<ST, SCT, MockExecutionProtocol>,
                 MonadEvent<ST, SCT, MockExecutionProtocol>,
+                PingPongDiscovery<ST>,
```
michael-yxchen, line 108: we can create a `MockDiscovery` that does absolutely nothing if that's helpful

monad-types/src/lib.rs
```diff
@@ -384,6 +384,35 @@ where
...
+    let hex_str = "0x".to_string() + &hex::encode(signature.serialize());
+    serializer.serialize_str(&hex_str)
+}
+
+pub fn deserialize_certificate_signature<'de, D, ST>(deserializer: D) -> Result<ST, D::Error>
+where
+    ST: CertificateSignatureRecoverable,
+    D: Deserializer<'de>,
+{
+    let buf = <std::string::String as Deserialize>::deserialize(deserializer)?;
```
michael-yxchen: super nit
```suggestion
    let buf = <String as Deserialize>::deserialize(deserializer)?;
```

monad-updaters/src/config_loader.rs
```diff
@@ -63,7 +63,8 @@ where
 
 pub struct ConfigLoader<ST, SCT, EPT>
 where
-    SCT: SignatureCollection,
+    ST: CertificateSignatureRecoverable,
+    SCT: SignatureCollection<NodeIdPubKey = CertificateSignaturePubKey<ST>>,
```
michael-yxchen: do we need this trait bound in struct or can it be moved to implementation?

michael-yxchen:
let's also put a kill switch on sending peer discovery messages so it's disabled in the next release

monad-updaters/src/config_loader.rs
eerkaijun: updated

monad-peer-discovery/src/mock.rs
eerkaijun: i think we should treat PingPongDiscovery as a mock implementation since it won't be used, simplifying it for easier initialization

monad-testground/src/executor.rs
eerkaijun: I think PingPongDiscovery is pretty barebone, should be fine for now

monad-raptorcast/src/lib.rs
eerkaijun: great point, updated

monad-raptorcast/src/lib.rs
eerkaijun: my thought is that we would replace `known_addresses` with `peer_discovery_driver.pd.peer_info` when we enable peer discovery, so during the second phase of rollout. 

or if we're conservative we can enable peer discovery during the second phase, but still keep known_addresses, and only switch it during the third phase

we should be able to test in parallel

monad-testground/src/executor.rs
michael-yxchen: it still sends out msgs, should we disable that?

eerkaijun: no idea how testground works so your call. if so, I think we can modify PingPongDiscovery to just be a mock discovery, don't think we need to maintain both PingPong and mock

eerkaijun:
Updated - corresponding flexnet changes here https://github.com/category-labs/monad-integration/pull/255

monad-testground/src/executor.rs
michael-yxchen: agree that we don't want both pingpong and mock

fine to leave this PR as in. I'll probably reduce PingPong to a Nop. I created it just to test peer discovery swarm. we don't need it anymore

docker/devnet/monad/config/node.toml
```diff
@@ -44,6 +52,7 @@ peers = []
 # secp256k1_pubkey = "..."
 
 [statesync]
+local_record_seq_num = 0
```
michael-yxchen: move to `[peer_discovery]` section

monad-raptorcast/src/lib.rs
```diff
@@ -522,6 +561,47 @@ where
...
+                    debug!(
+                        ?message,
+                        "dropping peer discovery message, should come through udp channel"
+                    );
+                    continue;
+                }
+            }
+        }
+
+        while let Poll::Ready(Some(peer_disc_emit)) = this.peer_discovery_driver.poll_next_unpin(cx)
```
michael-yxchen: discussed offline: disable peer discovery altogether - not processing messages or polling driver

docker/devnet/monad/config/node.toml
eerkaijun: oh lol good catch haha

monad-raptorcast/src/lib.rs
eerkaijun: updated

monad-raptorcast/src/lib.rs
```diff
@@ -469,9 +502,15 @@ where
                                 InboundRouterMessage::AppMessage(app_message) => {
                                     Some(app_message.event(from))
                                 }
+                                InboundRouterMessage::PeerDiscoveryMessage(peer_disc_message) => {
+                                    // handle peer discovery message in driver
+                                    this.peer_discovery_driver
+                                        .update(peer_disc_message.event(from));
+                                    None
```
michael-yxchen: we should guard this block with `PEER_DISCOVERY_ENABLED`, otherwise messages will queue up during upgrade

monad-node-config/src/peers.rs
```diff
@@ -0,0 +1,36 @@
...
+    #[serde(bound = "ST: CertificateSignatureRecoverable")]
+    pub peers: Vec<PeerDiscoveryConfig<ST>>,
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct PeerDiscoveryConfig<ST: CertificateSignatureRecoverable> {
+    pub address: SocketAddrV4,
+
+    pub record_seq_num: u64,
```
fabiotdf, line 25: How is this seqno set? Incremented by one every time the owning peer changes its ip address?

eerkaijun: yes, incremented when the name record is updated. I guess technically it doesn't have to increment by 1, as long as it's larger than previous name record

michael-yxchen: yeah that's the idea - owning peer should increment seq_no and re-sign when they update their ip address

i was thinking that we will ship a toolchain that takes in current config file + new address, and modify (bump seqno and re-sign) config file in place so it's less error prone. Operator changing just local address will hit a self name record verification failure that prevent the client from starting

if you have any other ideas, let us know

monad-node/src/main.rs
```diff
@@ -507,13 +513,72 @@ where
...
+                        last_ping: None,
+                        unresponsive_pings: 0,
+                        name_record: MonadNameRecord {
+                            name_record,
+                            signature: peer.name_record_sig,
+                        },
+                    },
+                )),
+                Err(_) => {
+                    warn!(?node_id, "invalid name record signature on startup");
```
fabiotdf: How about "invalid name record signature in config"?

monad-node/src/main.rs
fabiotdf: Will we have peers behind NAT? In that case the bind address and WAN-facing address will be different too

monad-raptorcast/tests/raptorcast_instance.rs
```diff
@@ -390,6 +392,10 @@ pub fn set_up_test(
                 key: rx_keypair,
                 full_nodes: Default::default(),
                 known_addresses,
```
fabiotdf, line 394: How come `RaptorCast::known_addresses` still maps nodeId to normal SockAddr  when we have peer discovery? Shouldn't it map to `MonadNameRecord` ?

eerkaijun: we haven't remove `known_addresses` yet. that will come in future PRs when we are confident to switch over. currently peer discovery is a standalone process integrated into router but not used yet

PR #1784, opened 2025-05-12
Txgen: prevent unbounded memory growth in recipient tracker

JoeHowarth:


JoeHowarth:
Deployed to txgen vps box

PR #1785, opened 2025-05-13
Full-node Raptorcast: UdpState adaptation

fabiotdf:
This is part 3 of a PR series that splits the original, larger Full-node RaptorCast PR.
This one focus on changes to RaptorCasts UdpState API, that now uses the new `RebroadcastGroupMap`.

This map is needed so a full-node knows which peers (other full-nodes)  to re-raptorcast to, given the `author_id` in the inbound message, as it is able to participate in multiple RC groups simultaneously.

### DIFF
--- a/monad-raptorcast/src/lib.rs
+++ b/monad-raptorcast/src/lib.rs
@@ -31,7 +31,7 @@ use monad_types::{
 pub mod message;
 pub mod udp;
 pub mod util;
-use util::{BuildTarget, EpochValidators, FullNodes, Validator};
+use util::{BuildTarget, EpochValidators, FullNodes, ReBroadcastGroupMap, Validator};
 
 const SIGNATURE_SIZE: usize = 65;
 
@@ -65,7 +65,10 @@ where
     key: ST::KeyPairType,
     redundancy: u8,
 
+    // Raptorcast group with stake information. For the send side (i.e., initiating proposals)
     epoch_validators: BTreeMap<Epoch, EpochValidators<ST>>,
+    rebroadcast_map: ReBroadcastGroupMap<ST>,
+
     // TODO support dynamic updating
     full_nodes: FullNodes<CertificateSignaturePubKey<ST>>,
     known_addresses: HashMap<NodeId<CertificateSignaturePubKey<ST>>, SocketAddr>,
@@ -100,14 +103,17 @@ where
     OM: Serializable<Bytes> + Encodable + Into<M> + Clone,
 {
     pub fn new(config: RaptorCastConfig<ST>) -> Self {
+        assert!(config.redundancy >= 1);
         let self_id = NodeId::new(config.key.pubkey());
         let mut builder = DataplaneBuilder::new(&config.local_addr, config.up_bandwidth_mbps);
         if let Some(buffer_size) = config.buffer_size {
             builder = builder.with_buffer_size(buffer_size);
         }
         let dataplane = builder.build();
+        let is_fullnode = false; // This will come from the config in upcoming PR
         Self {
             epoch_validators: Default::default(),
+            rebroadcast_map: ReBroadcastGroupMap::new(self_id, is_fullnode),
             full_nodes: FullNodes::new(config.full_nodes),
             known_addresses: config.known_addresses,
 
@@ -171,9 +177,10 @@ where
         let self_id = NodeId::new(self.key.pubkey());
         for command in commands {
             match command {
-                RouterCommand::UpdateCurrentRound(epoch, _round) => {
+                RouterCommand::UpdateCurrentRound(epoch, round) => {
                     assert!(epoch >= self.current_epoch);
                     self.current_epoch = epoch;
+                    self.rebroadcast_map.delete_expired_groups(epoch, round);
                     while let Some(entry) = self.epoch_validators.first_entry() {
                         if *entry.key() + Epoch(1) < self.current_epoch {
                             entry.remove();
@@ -186,6 +193,8 @@ where
                     epoch,
                     validator_set,
                 } => {
+                    self.rebroadcast_map
+                        .push_group_validator_set(validator_set.clone(), epoch);
                     if let Some(epoch_validators) = self.epoch_validators.get(&epoch) {
                         assert_eq!(validator_set.len(), epoch_validators.validators.len());
                         assert!(validator_set.into_iter().all(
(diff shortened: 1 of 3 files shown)


monad-raptorcast/src/lib.rs
```diff
@@ -416,12 +425,20 @@ where
                 break;
             };
 
+            // Enter the received raptorcast chunk into the udp_state for reassembly.
+            // If the field "first-hop recipient" in the chunk has our node Id, then
+            // we are responsible for broadcasting this chunk to other validators.
+            // Once we have enough (redundant) raptorcast chunks, recreate the
+            // decoded (AKA parsed, original) message.
+            // Send the decoded message to our dedicated full-nodes once we have the
+            // whole thing.
```
michael-yxchen: minor correction: dedicated full nodes are streamed chunks as we receive them. We re-encode the proposal and send to prioritized/public full nodes

monad-raptorcast/src/udp.rs
```diff
@@ -116,11 +116,12 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
     /// Given a RecvMsg, emits all decoded messages while rebroadcasting as necessary
     pub fn handle_message(
         &mut self,
-        epoch_validators: &mut BTreeMap<Epoch, EpochValidators<ST>>,
-        rebroadcast: impl FnMut(Vec<NodeId<CertificateSignaturePubKey<ST>>>, Bytes, u16),
-        forward: impl FnMut(Bytes, u16),
-        message: RecvMsg,
-    ) -> Vec<(NodeId<CertificateSignaturePubKey<ST>>, Bytes)> {
+        group_map: &ReBroadcastGroupMap<ST>, // (msg.epoch, msg.author_id) -> Group. Replaces `epoch_validators`
```
michael-yxchen: nit: clean up comment

monad-raptorcast/src/udp.rs
```diff
@@ -217,19 +212,18 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
...
                     if self_hash == parsed_message.recipient_hash {
-                        let epoch_validators = epoch_validators
-                            .get_mut(&Epoch(parsed_message.epoch))
-                            .expect("parsed message was validated");
-                        let targets =
-                            epoch_validators.view_without(vec![&parsed_message.author, &self_id]);
+                        let targets = group_map.iterate_rebroadcast_peers(
+                            Epoch(parsed_message.epoch),
+                            &parsed_message.author,
+                        );
```
michael-yxchen: would it make sense to have `iterate_rebroadcast_peers` return an `Option<GroupIterator>` here, and skip `queue_broadcast` on `None`?

monad-raptorcast/src/udp.rs
```diff
@@ -526,28 +520,44 @@ pub fn build_messages_with_length<ST>(
 where
     ST: CertificateSignatureRecoverable,
 {
+    assert!(redundancy >= 1);
```
michael-yxchen: logically this makes sense, but we didn't verify when building `RaptorCast/RaptorCastConfig`. we should move the assertion there instead of checking it when building every message

monad-raptorcast/src/udp.rs
```diff
@@ -639,14 +660,14 @@ where
             assert!(is_broadcast && is_raptor_broadcast);
 
             tracing::trace!(
-                self_id =? key.pubkey(),
+                self_id =? self_id,
```
michael-yxchen: nit
```suggestion
                ?self_id,
```

monad-raptorcast/src/udp.rs
```diff
@@ -683,6 +704,10 @@ where
                 }
             }
 
+            // Build the UDP (complete) messages to be sent to all (if any) full nodes assigned to this validator.
+            // Sending without raptor-casting.
+            // This path is only taken when the current validator is the one broadcasting the proposal.
+            // When it is not, the messages are transmitted to full nodes from somewhere else (Stream for RaptorCast<>?)
```
michael-yxchen: dedicated full nodes get a copy of all raptorcast chunks. When our node is not the leader, chunks are forwarded in the handle_message path. When our node is the leader generating chunks, we're forwarding all chunks here

monad-raptorcast/src/udp.rs
```diff
@@ -695,8 +720,52 @@ where
                 }
             }
         }
+        BuildTarget::FullNodeRaptorCast(group) => {
+            assert!(is_broadcast && is_raptor_broadcast);
+
+            tracing::trace!(
+                self_id =? self_id,
```
michael-yxchen: same here

monad-raptorcast/src/udp.rs
```diff
@@ -695,8 +720,52 @@ where
...
+                unix_ts_ms,
+                app_message_len,
+                redundancy,
+                data_size,
+                num_packets,
+                app_message_hash =? app_message_hash,
```
michael-yxchen: same here

monad-raptorcast/src/lib.rs
fabiotdf: Ah yes, fixed.

monad-raptorcast/src/udp.rs
fabiotdf: Implemented

monad-raptorcast/src/udp.rs
fabiotdf: Moved the assertion to the suggested spot

monad-raptorcast/src/udp.rs
fabiotdf: Done

monad-raptorcast/src/udp.rs
fabiotdf: Done

monad-raptorcast/src/udp.rs
fabiotdf: Sorry, old comment! 
Fixed and replaced with our comment instead. Thanks

monad-raptorcast/src/udp.rs
fabiotdf: done

monad-raptorcast/src/udp.rs
fabiotdf: done

PR #1787, opened 2025-05-13
Support new message type for router messages

eerkaijun:
Add new message type (peer discovery message) to router messages. Adding subsequent message type (e.g. full node message) should be trivial. This change is backwards compatible.


monad-raptorcast/src/lib.rs
```diff
@@ -34,6 +34,7 @@ pub mod util;
 use util::{BuildTarget, EpochValidators, FullNodes, Validator};
 
 const SIGNATURE_SIZE: usize = 65;
+const NEW_ROUTER_MESSAGE_FORMAT: bool = false;
```
eerkaijun, line 37: this constant can be switched when we want to serialize with new message type

michael-yxchen:
I think it’s best to contain `RouterMessage` within router implementation, ie MonadState `Command::RouterCommand` should still use `VerifiedMonadMessage`

we would poll `PeerDiscoveryEmit` out of `PeerDiscoveryDriver` in `RaptorCast::poll_next` Ideally the `PeerDiscoveryMessage` is directly sent out in `poll_next`. Might need to do some reorganizing to reuse code from exec

eerkaijun:
Folding the new router message type into https://github.com/category-labs/monad-bft/pull/1780

PR #1788, opened 2025-05-13
Add message_len to router message

bamrith:


monad-raptorcast/src/message.rs
```diff
@@ -95,16 +98,19 @@ impl<M: Decodable> InboundRouterMessage<M> {
...
                         // decompress message
-                        let mut decompressed_app_message = Vec::new();
+                        let mut decompressed_writer = BoundedWriter::new(decompressed_message_len);
                         ZstdCompression::default()
-                            .decompress(&compressed_app_message, &mut decompressed_app_message)
+                            .decompress(&compressed_app_message, &mut decompressed_writer)
                             .map_err(|err| {
                                 DeserializeError(format!("decompression error: {:?}", err))
                             })?;
+                        let decompressed_app_message: Bytes = decompressed_writer.into();
```
michael-yxchen, line 168: should we verify `decompressed_app_message.len() == decompressed_message_len`?

monad-raptorcast/src/message.rs
```diff
@@ -95,16 +98,19 @@ impl<M: Decodable> InboundRouterMessage<M> {
                         Ok(Self::AppMessage(app_message))
                     }
                     DEFAULT_ZSTD_VERSION => {
+                        let decompressed_message_len =
+                            u32::decode(&mut payload).map_err(DeserializeError::from)?;
```
michael-yxchen, line 157: is it possible to put a bound on this value? it can be rather generous (e.g. I think 4MB is enough - 2MB txn payload + header + serialization overheads), but we'd also need to remember to update the value if we're every increasing the block size limit

if we decide to put a bound here, let's add a check on sending side and log if serialized message is over the limit

bamrith: I thought that would be good as well but wasn't sure how big the statesync and forwarded txns message would be.

bamrith: Discussed on slack. To be addressed later https://github.com/category-labs/category-internal/issues/1605

dshulyak:
i think message length field wasn't really necessary. 

it is the same as reading first MAX_MESSAGE_SIZE from zsdt stream, if stream is shorter you will just read that, if stream is longer you will read atmost MAX_MESSAGE_SIZE and then fail validation. something like the code below (that BoundedWriter probably achieves the same, but i didn't check)

```rs
use zstd::stream::Decoder;
use std::io::{Read, Cursor};

fn streaming_decompress_with_limit(
    compressed_data: &[u8], 
    max_size: usize
) -> Result<Vec<u8>, Box<dyn std::error::Error>> {
    let cursor = Cursor::new(compressed_data);
    let mut decoder = Decoder::new(cursor)?;
    
    let mut result = Vec::new();
    let mut buffer = [0u8; 4096];
    
    loop {
        match decoder.read(&mut buffer) {
            Ok(0) => break,
            Ok(n) => {
                if result.len() + n > max_size {
                    return Err(format!("Output exceeds {} bytes", max_size).into());
                }
                result.extend_from_slice(&buffer[..n]);
            }
            Err(e) => return Err(e.into()),
        }
    }
    
    Ok(result)
}
```

---

one other thing is that decompressing into 1.5MB takes 1ms, so it is somewhat easy to block consensus thread with this workload

shared my other thoughts here https://github.com/category-labs/category-internal/issues/1644

PR #1790, opened 2025-05-15
Fix BLS minor issues

dnkolegov-ar:
This PR fixes minor issues in BLS implementation:
- update `blst` crate
- enable zeroization for secret key wrapper
- add public key validation calls
- add public key validation tests


monad-bls/src/bls.rs
```diff
@@ -151,20 +158,30 @@ impl BlsPubKey {
         self.0.serialize().to_vec()
     }
 
-    pub fn deserialize(message: &[u8]) -> Result<Self, BlsError> {
-        blst_core::PublicKey::deserialize(message)
+    pub fn deserialize(msg: &[u8]) -> Result<Self, BlsError> {
+        if msg.len() != PUBKEY_BYTE_LEN {
+            return Err(BlsError(BLST_BAD_ENCODING));
+        }
```
michael-yxchen, line 164: i believe `blst_core::PublicKey::deserialize` impl checks msg length. is this additional one needed? if so, should we add it to signature deserialize/uncompress too?

dnkolegov-ar: You are right, but `deserialize ` accepts both compressed and uncompressed formats, and your implementation only one of them

michael-yxchen: got it, thanks!

PR #1791, opened 2025-05-16
monad-rpc: perfetto integration

dshulyak:
it requires a bit of a setup. i will describe that later. will give an overview of some features here

breakdown by what is executed on each cpu

image

performance counters, in this example it is context switches. but also without docker it can do callstack sampling with symbolization at specified frequency (eg. what you expect from perf record -F 99)

image

scheduler tracing

image

process stats

image

and application tracing (it uses my fork for now, i will move that to the org later).
note that it shows what is being executed on a thread and not a total time of the span, in case of async work same span will appear multiple times

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2135,6 +2135,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "codespan-reporting"
+version = "0.12.0"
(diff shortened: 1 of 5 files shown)


dshulyak:
this will be superseded with a custom tooling

PR #1792, opened 2025-05-16
Send ping to newly added peers

eerkaijun:
Closes https://github.com/category-labs/category-internal/issues/1542

### DIFF
--- a/monad-peer-disc-swarm/tests/peer_discovery.rs
+++ b/monad-peer-disc-swarm/tests/peer_discovery.rs
@@ -428,13 +428,19 @@ fn test_prune_nodes() {
 fn test_peer_lookup_random_nodes() {
     let keys = create_keys::<SignatureType>(5);
 
-    let subset_keys = &keys[0..2];
+    let subset_keys = &keys[0..1];
     let all_keys = &keys[0..4];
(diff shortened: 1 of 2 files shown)


monad-peer-disc-swarm/tests/peer_discovery.rs
```diff
@@ -519,16 +510,23 @@ fn test_peer_lookup_random_nodes() {
...
     for node_id in all_peers.keys() {
+        if node_id == &node_b_state.self_id {
+            continue;
+        }
         assert!(node_b_state.peer_info.contains_key(node_id));
     }
+
+    // check that NodeB sends pings to new nodes
+    let metrics = node_b_state.metrics();
+    assert_eq!(metrics["send_ping"], 3);
```
michael-yxchen, line 531: we also want to assert that NodeC/D have NodeB in their peer list. To guarantee it's strictly from B's ping msg and not from a peer lookup response from A, we can remove NodeA from C/D's initial peer list. It's a little unrealistic but is more targeted

eerkaijun: cool, updated

PR #1793, opened 2025-05-16
use named threads for easier identification

dshulyak:
easier to identify in tooling. some thread names are shortened to fit in 15 limit, otherwise they will be truncated at that limit.
there are still some threads that are not named at creation, but i think they are in monad code, i will try to adjust them there.
and i didn't touch threads created by tokio or actix, they are already named differently and easy to identify

image

### DIFF
--- a/monad-statesync/src/ffi.rs
+++ b/monad-statesync/src/ffi.rs
@@ -4,6 +4,7 @@ use std::{
     pin::Pin,
     sync::{Arc, Mutex},
     task::{Context, Poll},
+    thread,
     time::Duration,
 };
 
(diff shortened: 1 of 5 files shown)


monad-statesync/src/ffi.rs
```diff
@@ -317,7 +318,7 @@ impl<PT: PubKey> StateSync<PT> {
             }
             // this loop exits when execution is about to start
             assert!(sync_ctx.ctx.is_none());
-        });
+        }).expect("failed to spawn statesync thread");
```
michael-yxchen, line 321: should we also `expect` at other thread spawning spots?

dshulyak: we should, thanks

PR #1794, opened 2025-05-18
Update ubuntu, gcc

omegablitz:


omegablitz:
Depends on category-labs/monad#1324

PR #1795, opened 2025-05-19
Refactor to always ping new peers when name record modified or added

eerkaijun:
Realized I forgot to also make sure that nodes send ping to new nodes in the `handle_ping` code path. Updated here and some slight refactor to make sure whenever we add or modify name records, we should also schedule timer to send pings.

### DIFF
--- a/monad-peer-discovery/src/discovery.rs
+++ b/monad-peer-discovery/src/discovery.rs
@@ -161,6 +161,35 @@ impl<ST: CertificateSignatureRecoverable> PeerDiscovery<ST> {
             },
         )]
     }
+
+    fn insert_new_peers(
+        &mut self,
+        node_id: NodeId<CertificateSignaturePubKey<ST>>,
+        name_record: MonadNameRecord<ST>,
+    ) -> Vec<PeerDiscoveryCommand<ST>> {
+        // insert name record
+        if let Some(info) = self.peer_info.get_mut(&node_id) {
+            if name_record.seq() > info.name_record.seq() {
+                // update name record
+                debug!(?node_id, ?name_record, "name record updated");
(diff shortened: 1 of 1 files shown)


monad-peer-discovery/src/discovery.rs
```diff
@@ -161,6 +161,33 @@ impl<ST: CertificateSignatureRecoverable> PeerDiscovery<ST> {
...
+    fn insert_new_peers(
+        &mut self,
+        node_id: NodeId<CertificateSignaturePubKey<ST>>,
+        name_record: MonadNameRecord<ST>,
+    ) -> Vec<PeerDiscoveryCommand<ST>> {
+        // insert name record
+        if let Some(info) = self.peer_info.get_mut(&node_id) {
+            if name_record.seq() > info.name_record.seq() {
+                // update name record
+                info.name_record = name_record;
```
michael-yxchen: nit: I missed this earlier, can we add a log line here?

monad-peer-discovery/src/discovery.rs
```diff
@@ -161,6 +161,33 @@ impl<ST: CertificateSignatureRecoverable> PeerDiscovery<ST> {
...
+                // exit if seq num is not incremented
+                return vec![];
+            }
+        } else {
+            // peer is not present, insert peer
+            self.peer_info.insert(node_id, PeerInfo {
+                last_ping: None,
+                unresponsive_pings: 0,
+                name_record,
+            });
```
michael-yxchen: also here

monad-peer-discovery/src/discovery.rs
eerkaijun: added

PR #1798, opened 2025-05-19
Full-node Raptorcast: Refactored Group message de/serialization

fabiotdf:
This is part of a series of PR split from the original Full-node RaptorCast integration.
This PR contains just the Group messages adapted to the new, recent serialization method by Kai

### DIFF
--- a/monad-raptorcast/src/lib.rs
+++ b/monad-raptorcast/src/lib.rs
@@ -12,7 +12,7 @@ use std::{
 use alloy_rlp::{Decodable, Encodable};
 use bytes::{Bytes, BytesMut};
 use futures::{channel::oneshot, FutureExt, Stream};
-use message::{DeserializeError, InboundRouterMessage};
+use message::InboundRouterMessage;
 use monad_consensus_types::signature_collection::SignatureCollection;
 use monad_crypto::certificate_signature::{
     CertificateKeyPair, CertificateSignaturePubKey, CertificateSignatureRecoverable, PubKey,
@@ -29,8 +29,11 @@ use monad_types::{
 };
 
 pub mod message;
+pub mod raptorcast_secondary;
+pub mod rlp;
 pub mod udp;
 pub mod util;
+use rlp::DeserializeError;
 use util::{BuildTarget, EpochValidators, FullNodes, ReBroadcastGroupMap, Validator};
 
 const SIGNATURE_SIZE: usize = 65;
@@ -379,9 +382,9 @@ fn try_deserialize_router_message<
     M: Message<NodeIdPubKey = CertificateSignaturePubKey<ST>> + Deserializable<Bytes> + Decodable,
 >(
     bytes: &Bytes,
-) -> Result<InboundRouterMessage<M>, DeserializeError> {
+) -> Result<InboundRouterMessage<M, ST>, DeserializeError> {
     // try to deserialize as a new message first
-    let Ok(inbound) = InboundRouterMessage::<M>::try_deserialize(bytes) else {
+    let Ok(inbound) = InboundRouterMessage::<M, ST>::try_deserialize(bytes) else {
         // if that fails, try to deserialize as an old message instead
         return match M::deserialize(bytes) {
             Ok(old_message) => Ok(InboundRouterMessage::AppMessage(old_message)),
@@ -469,6 +472,12 @@ where
                                 InboundRouterMessage::AppMessage(app_message) => {
                                     Some(app_message.event(from))
                                 }
+                                InboundRouterMessage::FullNodesGroup(full_nodes_group_message) => {
+                                    tracing::warn!(
+                                        "Handling of FullNodesGroup implemented in upcoming PR"
+                                    );
+                                    None
+                                }
                             },
                             Err(err) => {
                                 tracing::warn!(
@@ -521,6 +530,10 @@ where
                     return Poll::Ready(Some(
                         RaptorCastEvent::Message(message.event(NodeId::new(from))).into(),
                     ));
+                }
+                InboundRouterMessage::FullNodesGroup(_group_message) => {
+                    // pass TCP message to MultiRouter
+                    tracing::warn!("FullNodesGroup protocol via TCP not implemented");
                 }
             }
         }
--- a/monad-raptorcast/src/message.rs
+++ b/monad-raptorcast/src/message.rs
@@ -1,62 +1,35 @@
-use alloy_rlp::{encode_list, Decodable, Encodable, Header, RlpDecodable, RlpEncodable};
+use alloy_rlp::{encode_list, Decodable, Encodable, Header};
 use bytes::{Bytes, BytesMut};
-use monad_compress::{zstd::ZstdCompression, CompressionAlgo};
+use monad_crypto::certificate_signature::CertificateSignatureRecoverable;
 
-const SERIALIZE_VERSION: u32 = 1;
-// compression versions
-const UNCOMPRESSED_VERSION: u32 = 1;
-const DEFAULT_ZSTD_VERSION: u32 = 2;
+use super::{
+    raptorcast_secondary::group_message::FullNodesGroupMessage,
+    rlp::{
+        try_deserialize_rlp_message, try_serialize_rlp_message, DeserializeError,
+        NetworkMessageVersion, SerializeError,
+    },
+};
 
-#[derive(RlpEncodable, RlpDecodable)]
-struct NetworkMessageVersion {
-    pub serialize_version: u32,
-    pub compression_version: u32,
+const MESSAGE_TYPE_APP: u8 = 1;
+const MESSAGE_TYPE_GROUP: u8 = 3;
+
+pub enum OutboundRouterMessage<OM, ST: CertificateSignatureRecoverable> {
+    AppMessage(OM),                            // MESSAGE_TYPE_APP
+    FullNodesGroup(FullNodesGroupMessage<ST>), // MESSAGE_TYPE_GROUP
 }
 
-impl NetworkMessageVersion {
-    pub fn version() -> Self {
-        Self {
-            serialize_version: SERIALIZE_VERSION,
-            compression_version: UNCOMPRESSED_VERSION,
-        }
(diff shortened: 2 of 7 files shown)


monad-raptorcast/src/message.rs
```diff
@@ -1,117 +1,62 @@
...
+use super::{
+    raptorcast_secondary::group_message::FullNodesGroupMessage,
+    rlp::{
+        try_deserialize_rlp_message, try_serialize_rlp_message, DeserializeError,
+        NetworkMessageVersion, SerializeError,
+    },
+};
+
+const MESSAGE_TYPE_APP: u8 = 1;
+const MESSAGE_TYPE_GROUP: u8 = 2;
```
michael-yxchen: nit: can you use 3 instead. Peer discovery messages are using 2 in https://github.com/category-labs/monad-bft/pull/1780

monad-raptorcast/src/lib.rs
```diff
@@ -522,6 +530,10 @@ where
                         RaptorCastEvent::Message(message.event(NodeId::new(from))).into(),
                     ));
                 }
+                InboundRouterMessage::FullNodesGroup(_group_message) => {
+                    // pass TCP message to MultiRouter
+                    unimplemented!("FullNodesGroup protocol via TCP not implemented");
```
michael-yxchen: let's nerf that to a warning so client doesn't panic on malformed msg

monad-raptorcast/src/lib.rs
```diff
@@ -27,8 +27,11 @@ use monad_executor_glue::{
 use monad_types::{
     Deserializable, DropTimer, Epoch, ExecutionProtocol, NodeId, RouterTarget, Serializable,
 };
+use rlp::DeserializeError;
```
michael-yxchen: super nit: reorder it after `pub mod rlp;` because rlp is an internal mod not external crate

monad-raptorcast/src/lib.rs
```diff
@@ -469,6 +472,11 @@ where
                                 InboundRouterMessage::AppMessage(app_message) => {
                                     Some(app_message.event(from))
                                 }
+                                InboundRouterMessage::FullNodesGroup(full_nodes_group_message) => {
+                                    unimplemented!(
+                                        "Handling of FullNodesGroup implemented in upcoming PR"
+                                    );
```
michael-yxchen: nit: let's nerf this temporarily as well

monad-raptorcast/src/rlp.rs
```diff
@@ -0,0 +1,94 @@
...
+
+#[derive(Debug)]
+pub struct SerializeError(pub String);
+
+pub fn try_serialize_rlp_message<M: Encodable>(
+    msg_type: u8,
+    rlp_message: M,
+    version: NetworkMessageVersion,
+    out_buf: &mut dyn BufMut,
+) -> Result<(), SerializeError> {
```
michael-yxchen: compression only sees gains in AppMessage/proposals. Group and peer discovery messages should be sent uncompressed - most payloads are random (NodeId/signatures) so compression is wasteful there

we can strip out compression from this helper function

monad-raptorcast/src/message.rs
fabiotdf: Done

monad-raptorcast/src/lib.rs
fabiotdf: Good point! done

monad-raptorcast/src/lib.rs
fabiotdf: Done

monad-raptorcast/src/lib.rs
fabiotdf: Done

monad-raptorcast/src/rlp.rs
fabiotdf: Good point, but may I instead pass a `version` argument that requests no compression? This way the code stays cleaner

monad-raptorcast/src/rlp.rs
```diff
@@ -0,0 +1,94 @@
...
+            serialize_version: SERIALIZE_VERSION,
+            compression_version: UNCOMPRESSED_VERSION,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct SerializeError(pub String);
+
+pub fn try_serialize_rlp_message<M: Encodable>(
```
michael-yxchen: nit: `pub (crate)`

monad-raptorcast/src/rlp.rs
```diff
@@ -0,0 +1,94 @@
...
+#[derive(Debug)]
+pub struct DeserializeError(pub String);
+
+impl From<alloy_rlp::Error> for DeserializeError {
+    fn from(err: alloy_rlp::Error) -> Self {
+        DeserializeError(format!("rlp decode error: {:?}", err))
+    }
+}
+
+pub fn try_deserialize_rlp_message<M: Decodable>(
```
michael-yxchen: also pub crate

monad-raptorcast/src/rlp.rs
fabiotdf: Pushed the change suggested above ^
Please let me know if there still are any objections regarding this

monad-raptorcast/src/message.rs
```diff
@@ -1,117 +1,62 @@
...
-                    // encode as compressed message
-                    let enc: [&dyn Encodable; 3] = [&version, &1u8, &compressed_app_message];
-                    encode_list::<_, dyn Encodable>(&enc, &mut buf);
-                } else {
-                    unreachable!()
-                };
+                try_serialize_rlp_message(MESSAGE_TYPE_APP, app_message, version, &mut buf)?;
+            }
+            Self::FullNodesGroup(generic_grp_message) => {
+                generic_grp_message.try_serialize_inner(version)?;
```
eerkaijun: this calls into `try_serialize_inner`, which then calls into `try_serialize_rlp_message`. the message type passed in are `MESSAGE_TYPE_PREP_REQ`, `MESSAGE_TYPE_PREP_RES` and `MESSAGE_TYPE_CONF_GRP` which are constants conflicting with APP_MESSAGE and peer discovery message etc. or am I thinking it wrongly?

monad-raptorcast/src/message.rs
```diff
@@ -1,117 +1,62 @@
...
-                            .map_err(DeserializeError::from)?;
-                        Ok(Self::AppMessage(app_message))
-                    }
-                    _ => Err(DeserializeError("unknown compression version".into())),
-                }
+            MESSAGE_TYPE_APP => {
+                let decoded_msg: AM = try_deserialize_rlp_message(version, &mut payload)?;
+                Ok(Self::AppMessage(decoded_msg))
+            }
+            MESSAGE_TYPE_GROUP => {
```
eerkaijun, line 57: `MESSAGE_TYPE_GROUP` doesn't seem to get used during serialization

monad-raptorcast/src/message.rs
```diff
@@ -1,117 +1,62 @@
...
+                let decoded_msg =
+                    FullNodesGroupMessage::try_deserialize_inner(version, &mut payload)?;
+                Ok(Self::FullNodesGroup(decoded_msg))
```
michael-yxchen: check `compression_version` is uncompressed

monad-raptorcast/src/message.rs
michael-yxchen: good catch. Cleaner way is to implement `Encodable/Decodable` for `FullNodesGroupMessage`

fabiotdf: How do I pass `version` to Encodable implementation?

michael-yxchen: you don't. This is an example from `MonadMessage`: https://github.com/category-labs/monad-bft/blob/8b2ba58e30b7eee23b3079e7f0aa9af60add35bd/monad-state/src/lib.rs#L427-L461

you attach version and enum discriminant when serializing `RouterMessage`

eerkaijun: perhaps something like https://github.com/category-labs/monad-bft/blob/master/monad-peer-discovery/src/message.rs#L38

the NetworkVersionMessage will just be used in the outer serialization

fabiotdf: Does the recent commit look more correct? No `version` information is passed, though

michael-yxchen: last commit i see is from yesterday at 4p. did you push?

monad-raptorcast/src/message.rs
fabiotdf: Fixed

monad-raptorcast/src/message.rs
fabiotdf: I did, and I only see your 48min old comments now. What I committed doesn't 100% match the example you copied. Let me review

monad-raptorcast/src/rlp.rs
fabiotdf: Done

monad-raptorcast/src/message.rs
fabiotdf: Now ignoring the `version` field, so always uncompressed

monad-raptorcast/src/message.rs
fabiotdf: So we encode a dummy version on every layer but also ignore it when deserializing?

michael-yxchen: yeah we're ignoring it rn because we only have one message version. It will be used when upgrading

fabiotdf: What if sender sends version 2 and receiver, being a build behind, doesn't know 2 exists? It can crash or just interpret messages wrongly.
I added a check here and if the receiver sees a version higher than it knows about, it returns an error

michael-yxchen: you're right, we should add that check

realistically at least 1 release before we upgrade, we will add the version check

fabiotdf:
Closed & reworked on a different PR

PR #1799, opened 2025-05-19
Fix warnings

fabiotdf:
Just addressing a few annoying warnings

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -422,7 +422,7 @@ where
 
         // a valid proposal will advance the pacemaker round so capture the original round before
         // handling the proposal certificate
-        let original_round = self.consensus.pacemaker.get_current_round();
+        let _original_round = self.consensus.pacemaker.get_current_round();
         cmds.extend(self.proposal_certificate_handling(&p));
 
         // author, leader, round checks
--- a/monad-eth-block-validator/src/lib.rs
+++ b/monad-eth-block-validator/src/lib.rs
@@ -167,7 +167,7 @@ where
 
         let ProposedEthHeader {
             ommers_hash,
-            beneficiary,
+            beneficiary: _,
             transactions_root,
             withdrawals_root,
             difficulty,
--- a/monad-state/src/lib.rs
+++ b/monad-state/src/lib.rs
@@ -521,7 +521,7 @@ where
 {
     fn decode(buf: &mut &[u8]) -> alloy_rlp::Result<Self> {
         let mut payload = Header::decode_bytes(buf, true)?;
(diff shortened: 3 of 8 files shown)


monad-consensus-state/src/lib.rs
```diff
@@ -422,7 +422,7 @@ where
 
         // a valid proposal will advance the pacemaker round so capture the original round before
         // handling the proposal certificate
-        let original_round = self.consensus.pacemaker.get_current_round();
+        let _original_round = self.consensus.pacemaker.get_current_round();
```
maxkozlovsky, line 425: If this value is not going used, why not remove the line? The comment as well.

michael-yxchen: there's some code (on timestamp adjustment) commented out referencing `original_round`. I think we can keep it around until we decide what to do about that

monad-eth-block-validator/src/lib.rs
```diff
@@ -183,6 +183,8 @@ where
             parent_beacon_block_root,
         } = &header.execution_inputs;
 
+        let _ = beneficiary;
```
maxkozlovsky: Can be written without separate assignment as 
```
let ProposedEthHeader {
  ...
  beneficiary: _,
  ...
}
```

monad-triedb-utils/src/triedb_env.rs
```diff
@@ -201,8 +201,8 @@ fn polling_thread(
     max_async_read_concurrency: usize,
     receiver_traverse: mpsc::Receiver<TriedbRequest>,
     max_async_traverse_concurrency: usize,
-    max_finalized_block_cache_len: usize,
```
omegablitz, line 204: These args can be removed

monad-state/src/lib.rs
```diff
@@ -944,6 +944,7 @@ where
                     block_range,
                     full_blocks,
                 } => {
+                    let _ = block_range;
```
maxkozlovsky: Same here
```
block_range: _,
```

monad-eth-block-validator/src/lib.rs
fabiotdf: Ah yes, thanks!

monad-state/src/lib.rs
```diff
@@ -944,6 +944,7 @@ where
                     block_range,
```
omegablitz: `block_range: _,`

fabiotdf: Done

monad-state/src/lib.rs
fabiotdf: Done

PR #1801, opened 2025-05-19
Add limit on the size of data in upserts list

maxkozlovsky:
Limit the total message size in addition to number of upserts limit. This puts tighter bound on the message sizes not dependent on upsert sizes in batch.


maxkozlovsky:
https://github.com/category-labs/category-internal/issues/1606

PR #1802, opened 2025-05-19
RPC websocket support 

philipglazman:
Adds eth_subscribe and eth_unsubscribe for logs, monadLogs, newHeads, monadNewHeads, and monadEventStream.

The monadEventStream topic is experimental and was included in order to help Tenderly get started with the event system quickly.  The current thinking is that this topic will be removed once the websocket feature is ready for production.

Commit [1fd8df5](https://github.com/category-labs/monad-bft/pull/1802/commits/1fd8df5088be21189460c01b0410826f92424629) is also subject to change with an on-going rewrite the execution events reader. For reviewers, please only review files changes in the `monad-rpc` directory.

Flexnet tests: https://github.com/category-labs/monad-integration/pull/204

### DIFF
--- a/monad-rpc/src/websocket_server.rs
+++ b/monad-rpc/src/websocket_server.rs
@@ -40,7 +40,10 @@ pub enum Event {
 }
 
 impl WebSocketServer {
-    pub fn new(rx: flume::Receiver<PollResult>, broadcaster: tokio::sync::broadcast::Sender<Event>) -> Self {
+    pub fn new(
+        rx: flume::Receiver<PollResult>,
+        broadcaster: tokio::sync::broadcast::Sender<Event>,
+    ) -> Self {
         Self {
             rx,
             block_builder: BlockBuilder::new(),
@@ -90,27 +93,37 @@ impl WebSocketServer {
                             outcome,
                         } = &event
                         {
-                            if let ConsensusStateResult::Finalization {
-                                finalized_proposal: _,
-                                abandoned_proposals,
-                            } = self.consensus_state_tracker.update_proposal(
+                            match self.consensus_state_tracker.update_proposal(
                                 proposal_meta.block_number,
                                 &proposal_meta.id,
                                 *outcome,
                             ) {
-                                for abandoned in abandoned_proposals {
+                                ConsensusStateResult::Finalization {
+                                    finalized_proposal: _,
+                                    abandoned_proposals,
+                                } => {
+                                    for abandoned in abandoned_proposals {
+                                        self.process_block_update(
+                                            &abandoned.user_data,
+                                            ReferendumOutcome::Abandoned,
+                                        )
+                                        .await;
+                                        self.process_event_stream_item(
+                                            StreamEvent::AbandonedProposal {
+                                                proposal_meta: abandoned.proposal_meta,
+                                            },
+                                        )
+                                        .await;
+                                    }
+                                }
+                                ConsensusStateResult::Verification(verified_proposal) => {
                                     self.process_block_update(
-                                        &abandoned.user_data,
-                                        ReferendumOutcome::Abandoned,
-                                    )
-                                    .await;
-                                    self.process_event_stream_item(
-                                        StreamEvent::AbandonedProposal {
-                                            proposal_meta: abandoned.proposal_meta,
-                                        },
+                                        &verified_proposal.user_data,
+                                        ReferendumOutcome::Advanced(*outcome),
                                     )
                                     .await;
                                 }
+                                _ => {}
                             }
 
                             if let Some(opt_proposal_state) = self
--- a/monad-rpc/src/websocket.rs
+++ b/monad-rpc/src/websocket.rs
@@ -725,7 +725,8 @@ mod tests {
     }
 
     fn create_test_server(rx_exec_events: Receiver<PollResult>) -> actix_test::TestServer {
-        let (websocket_broadcast_tx, websocket_broadcast_rx) = tokio::sync::broadcast::channel::<Event>(10_000);
+        let (websocket_broadcast_tx, websocket_broadcast_rx) =
+            tokio::sync::broadcast::channel::<Event>(10_000);
 
         let ws_server = WebSocketServer::new(rx_exec_events, websocket_broadcast_tx.clone());
         tokio::spawn(async move {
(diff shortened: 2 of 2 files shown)


monad-rpc/src/eth_json_types.rs
```diff
@@ -459,6 +463,114 @@ impl Default for BlockTagOrHash {
...
+    Voted,
+    Finalized,
+    Verified,
+    Abandoned,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct SpeculativeNewHead {
+    #[serde(flatten)]
+    pub header: alloy_rpc_types::eth::Header,
```
eerkaijun, line 536: nit: already imported, can just be `Header`

monad-rpc/src/eth_json_types.rs
```diff
@@ -459,6 +463,114 @@ impl Default for BlockTagOrHash {
...
+    pub block_id: BlockId,
+    pub commit_state: BlockCommitState,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct SpeculativeLog {
+    #[serde(flatten)]
+    pub log: alloy_rpc_types::eth::Log,
```
eerkaijun, line 544: nit: can just be `Log`

monad-rpc/src/websocket_server.rs
```diff
@@ -0,0 +1,313 @@
...
+            total_difficulty: Some(difficulty),
+            size: Some(U256::ZERO),
+        };
+        let mut rpc_txns: Vec<alloy_rpc_types::eth::Transaction> = Vec::new();
+        for txn_info in block_update.transactions.iter() {
+            assert!(txn_info.is_some());
+            if let Some(txn_info) = txn_info {
+                let eff_gas_price = txn_info
+                    .txn_envelope
+                    .effective_gas_price(rpc_header.base_fee_per_gas.clone().take());
```
eerkaijun, line 219: I _think_ this can just be `.effective_gas_price(rpc_header.base_fee_per_gas);`

monad-rpc/src/websocket_server.rs
```diff
@@ -0,0 +1,313 @@
...
+        let rpc_header = alloy_rpc_types::eth::Header {
+            hash: block_hash,
+            inner: block_update.eth_header.clone(),
+            total_difficulty: Some(difficulty),
+            size: Some(U256::ZERO),
+        };
+        let mut rpc_txns: Vec<alloy_rpc_types::eth::Transaction> = Vec::new();
+        for txn_info in block_update.transactions.iter() {
+            assert!(txn_info.is_some());
+            if let Some(txn_info) = txn_info {
```
eerkaijun, line 216: can this be written something like
```
let rpc_txns: Vec<_> = block_update
    .transactions
    .iter()
    .filter_map(|txn_info| {
        ...
    })
    .collect();
```

this should remove the nested if

monad-rpc/src/websocket_server.rs
```diff
@@ -0,0 +1,313 @@
...
+                    transaction_index: Some(txn_info.txn_index),
+                    effective_gas_price: Some(eff_gas_price),
+                    from: txn_info.sender,
+                });
+            }
+        }
+
+        let block = alloy_rpc_types::eth::Block {
+            header: rpc_header,
+            uncles: vec![], // TODO(ken): not currently available, only historical?
```
eerkaijun, line 233: uncles won't exist in monad

monad-exec-events/src/block_builder.rs
```diff
@@ -0,0 +1,893 @@
...
+/// Collection of all event information aggregated during the execution
+/// of a single block
+#[derive(Clone, Debug, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
+pub struct ExecutedBlockInfo {
+    pub consensus_state: ConsensusState,
+    pub proposal_meta: ProposalMetadata,
+    pub chain_id: alloy_primitives::ChainId,
+    pub eth_header: alloy_consensus::Header,
+    pub eth_block_hash: B256,
+    pub transactions: Vec<Option<TransactionInfo>>,
```
eerkaijun, line 68: probably out of scope of this PR but curious why is this `Vec<Option<TransactionInfo>>` instead of `Vec<TransactionInfo>`?

monad-rpc/src/websocket_server.rs
```diff
@@ -0,0 +1,313 @@
...
+        };
+
+        // Create a list of logs
+        let mut log_idx = 0;
+        let logs: Vec<alloy_rpc_types::Log> = block_update
+            .transactions
+            .iter()
+            .filter_map(|txn| {
+                // If the transaction is None, we can skip it.
+                let (receipt, tx_id, tx_index) = match txn {
```
eerkaijun, line 245: under what scenario will transaction be None?

monad-rpc/src/websocket_server.rs
```diff
@@ -0,0 +1,313 @@
...
+                        txn.txn_index,
+                    ),
+                    None => return None,
+                };
+
+                let filtered_logs: Vec<alloy_primitives::Log> =
+                    receipt.logs().to_owned().into_iter().collect();
+
+                let rpc_logs = filtered_logs
+                    .into_iter()
```
eerkaijun, line 258: we can simplify this to
```
let rpc_logs = receipt
    .logs()
    .to_owned()
    .into_iter()
    .....
```

and remove `filtered_logs`

monad-exec-events/src/block_builder.rs
andr-dev: It has to do with the ordering of events coming on the event ring. Essentially, the guarantees are that after seeing a BlockStart, you'll get a stream of tx events and then finally a BlockEnd. Each tx write is scheduled on a separate fiber so it's possible (and highly likely) that you'll see txs "out of order" within a block. By using Vec<Option<...>>, we can check on the final BlockEnd that we actually got an event for every tx. This will be made more clear after the event ring API refactor.

monad-rpc/src/websocket_server.rs
andr-dev: If the tx is None, that means there was a bug in execution / event ring. This will be asserted internally in the new block builder so it doesn't get polluted outside the library.

PR #1805, opened 2025-05-19
Refactor RPC handlers to separate module

andr-dev:


monad-rpc/src/handlers/mod.rs
```diff
@@ -0,0 +1,586 @@
...
+}
+
+#[tracing::instrument(level = "debug", skip(app_state))]
+async fn rpc_select(
+    app_state: &MonadRpcResources,
+    method: &str,
+    params: Value,
+    request_id: RequestId,
+) -> Result<Value, JsonRpcError> {
+    match method {
```
dshulyak, line 166: i had somewhat conflicting change here https://github.com/category-labs/monad-bft/pull/1739/ . not a problem to rebase obviously, but just in case you plan more stuff

andr-dev: Got it, yeah this should be the bulk of it. Thanks for the heads up!

PR #1806, opened 2025-05-20
Actively look for new peers when below threshold

eerkaijun:
Partially https://github.com/category-labs/category-internal/issues/1539

### DIFF
--- a/monad-peer-disc-swarm/tests/peer_discovery.rs
+++ b/monad-peer-disc-swarm/tests/peer_discovery.rs
@@ -104,9 +104,10 @@ fn test_ping_pong() {
                         self_record: generate_name_record(key),
                         peer_info,
                         ping_period: Duration::from_secs(5),
-                        prune_period: Duration::from_secs(30),
+                        refresh_period: Duration::from_secs(30),
                         request_timeout: Duration::from_secs(1),
                         prune_threshold: 3,
+                        min_active_connections: 5,
                         rng_seed: 123456,
                     },
                     router_scheduler: NoSerRouterConfig::new(all_peers.keys().cloned().collect())
@@ -194,9 +195,10 @@ fn test_new_node_joining() {
                         self_record: generate_name_record(key),
                         peer_info,
                         ping_period: Duration::from_secs(2),
-                        prune_period: Duration::from_secs(4),
+                        refresh_period: Duration::from_secs(4),
                         request_timeout: Duration::from_secs(1),
                         prune_threshold: 3,
+                        min_active_connections: 5,
                         rng_seed: 123456,
                     },
                     router_scheduler: NoSerRouterConfig::new(all_peers.keys().cloned().collect())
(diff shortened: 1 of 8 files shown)


monad-peer-discovery/src/discovery.rs
```diff
@@ -403,24 +419,29 @@ where
...
-                        .take(MAX_PEER_IN_RESPONSE)
-                        .map(|(_k, v)| v.name_record)
-                        .collect()
-                }
+                None => vec![],
             }
         };
 
+        // if open discovery, return more nodes other than requested nodes
+        if request.open_discovery {
```
maxkozlovsky, line 430: maybe able to use choose_multiple instead
collect before calling extend should not be necessary

```
    name_records.extend(self.peer_info
        .values()
        .choose_multiple(self.rng, MAX_PEER_IN_RESPONSE - name_records.len())
        .into_iter()
        .map(|peer| peer.name_record))

```

monad-peer-discovery/src/discovery.rs
```diff
@@ -555,18 +579,33 @@ where
...
 
         // prune unresponsive nodes
+        // TODO: prioritization of validator vs full node
         self.peer_info
             .retain(|_, info| info.unresponsive_pings < self.prune_threshold);
 
+        // if fall below the min active peers, choose a few peers and lookup for new peers
+        if self.peer_info.len() < self.min_active_connections {
+            for _ in 0..NUM_LOOKUP_PEERS {
+                let to = match self.peer_info.keys().choose(&mut self.rng) {
```
maxkozlovsky: may be better to use choose_multiple here as well since choose() called multiple times may produce same peer twice

monad-peer-discovery/src/discovery.rs
```diff
@@ -555,18 +579,33 @@ where
...
+                    Some(key) => key,
+                    None => {
+                        error!("Peer info is empty, restart node!");
+                        return vec![];
+                    }
+                };
+                cmds.extend(self.send_peer_lookup_request(*to, self.self_id, true));
```
maxkozlovsky: might improve readability by using enum type for the last argument instead of boolean since it is not possible to understand what 'true' stands for without looking up definition

monad-peer-discovery/src/discovery.rs
eerkaijun: thanks, updated

monad-peer-discovery/src/discovery.rs
eerkaijun: it will always be two option only though, not sure if it's worth putting into a separate enum

monad-peer-discovery/src/discovery.rs
```diff
@@ -363,6 +370,7 @@ where
         &mut self,
         to: NodeId<CertificateSignaturePubKey<ST>>,
         target: NodeId<CertificateSignaturePubKey<ST>>,
+        open_discovery: bool,
```
michael-yxchen, line 373: when would we want to set `open_discovery` false?

monad-peer-disc-swarm/tests/peer_discovery.rs
michael-yxchen, line 1: can we add a TODO for a swarm test on discovering more peers below watermark? don't have to do it now if haven't started, as we're prioritizing stressnet tests

monad-peer-discovery/src/discovery.rs
eerkaijun: i was thinking in the future there would be scenario where we want only targeted node, for example there is one missing validator in the peer list (due to new validator that never reach out or for whatever reason).

michael-yxchen: updated spec

PR #1809, opened 2025-05-20
monad-rpc: temporarily disable otel spans reporting

dshulyak:
fallback solution to unblock release if we won't find a root cause of the problem.

monad-rpc heap consistently grows at high rate when DEBUG log enabled

image

we restarted container with jemallocator pprof enabled and it clearly shows that the leak in otel exporter

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5228,7 +5228,6 @@ dependencies = [
  "toml 0.7.8",
  "tracing",
  "tracing-actix-web",
- "tracing-opentelemetry",
  "tracing-subscriber",
  "vergen-git2",
 ]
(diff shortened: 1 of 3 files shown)


dshulyak:
there are also some huge logs in rpc debug output, some reaching ~3.2MB per line

dshulyak:
with this commit and same workload on swe-004 memory usage looks much more stable. i checked that logs are still available when traces are disabled

image

dshulyak:
so overnight swe004 (fixed version) had mostly stable memory usage at around 300MB. can007 (old version) consistently grows in memory and at the time of writing was ~19GB. both have similar load, mirrored from MF nodes. i also generated some additional requests to swe004

image
image

i think this change should be merged as in current state tracing is not useful anyway, and revise it before enabling again.

PR #1812, opened 2025-05-21
monad-node: use jemallocator with pprof integration

dshulyak:
was running it on swe4 since may 27. after reset node is working correctly. rss is consistently lower compared with a node using 0.9.3 rc by ~100MB, but i don't know how deterministic is that.

image

cpu usage is the same

image


dshulyak:
you are right, it is likely not accurate. i wanted to confirm that there are no regressions in cpu usage. also i think if i calculate rate over longer period (5-10m) it will smooth out workload differences. so assuming if it eventually handles same workload it is good enough to conclude that there is no significant changes, and might be explained by slightly different hw. 

image

feature flag makes it possible to recompile without jemalloc just in case if there will be unnoticed regression

PR #1814, opened 2025-05-21
Prune nodes when above max active connections

eerkaijun:
A few changes: 
- Only prune non validators node
- When servicing peer lookup response, prioritize sending validator nodes
- When above max active connections, prune full nodes

### DIFF
--- a/monad-node/src/main.rs
+++ b/monad-node/src/main.rs
@@ -13,7 +13,9 @@ use futures_util::{FutureExt, StreamExt};
 use monad_chain_config::{revision::ChainRevision, ChainConfig};
 use monad_consensus_state::ConsensusConfig;
 use monad_consensus_types::{
-    metrics::Metrics, signature_collection::SignatureCollection, validator_data::ValidatorsConfig,
+    metrics::Metrics,
+    signature_collection::SignatureCollection,
+    validator_data::{ValidatorSetDataWithEpoch, ValidatorsConfig},
 };
 use monad_control_panel::ipc::ControlPanelIpcReceiver;
 use monad_crypto::certificate_signature::{
@@ -39,7 +41,7 @@ use monad_statesync::StateSync;
 use monad_triedb_cache::StateBackendCache;
 use monad_triedb_utils::TriedbReader;
 use monad_types::{
-    Deserializable, DropTimer, NodeId, Round, SeqNum, Serializable, GENESIS_SEQ_NUM,
+    Deserializable, DropTimer, Epoch, NodeId, Round, SeqNum, Serializable, GENESIS_SEQ_NUM,
 };
 use monad_updaters::{
     checkpoint::FileCheckpoint, config_loader::ConfigLoader, loopback::LoopbackExecutor,
@@ -158,6 +160,7 @@ async fn run(node_state: NodeState, reload_handle: ReloadHandle) -> Result<(), (
         })
         .collect();
 
+    let current_epoch = node_state.forkpoint_config.high_qc.get_epoch();
     let router: BoxUpdater<_, _> = {
         let raptor_router = build_raptorcast_router::<
             SignatureType,
@@ -174,6 +177,8 @@ async fn run(node_state: NodeState, reload_handle: ReloadHandle) -> Result<(), (
                 .filter_map(|(node_id, maybe_addr)| Some((node_id, maybe_addr?)))
                 .collect(),
             &node_state.node_config.fullnode.identities,
+            locked_epoch_validators.clone(),
+            current_epoch,
         )
         .await;
 
@@ -500,6 +505,8 @@ async fn build_raptorcast_router<ST, SCT, M, OM>(
     identity: ST::KeyPairType,
     known_addresses: Vec<(NodeId<SCT::NodeIdPubKey>, SocketAddr)>,
     full_nodes: &[FullNodeIdentityConfig<CertificateSignaturePubKey<ST>>],
+    locked_epoch_validators: Vec<ValidatorSetDataWithEpoch<SCT>>,
+    current_epoch: Epoch,
 ) -> RaptorCast<ST, M, OM, MonadEvent<ST, SCT, ExecutionProtocolType>, PeerDiscovery<ST>>
 where
     ST: CertificateSignatureRecoverable,
@@ -561,15 +568,38 @@ where
         })
         .collect();
 
+    // TODO: make it configurable in node.toml
+    let epoch_validators = locked_epoch_validators
+        .iter()
+        .map(|epoch_validators| {
+            (
+                epoch_validators.epoch,
+                epoch_validators
+                    .validators
+                    .0
+                    .iter()
+                    .map(|validator| validator.node_id)
+                    .collect(),
+            )
+        })
+        .collect();
+    let dedicated_full_nodes = full_nodes
+        .iter()
+        .map(|full_node| NodeId::new(full_node.secp256k1_pubkey))
+        .collect();
     let peer_discovery_builder = PeerDiscoveryBuilder {
         self_id: NodeId::new(identity.pubkey()),
         self_record,
+        current_epoch,
+        epoch_validators,
+        dedicated_full_nodes,
         peer_info,
         ping_period: Duration::from_secs(peer_discovery_config.ping_period),
         refresh_period: Duration::from_secs(peer_discovery_config.refresh_period),
         request_timeout: Duration::from_secs(peer_discovery_config.request_timeout),
         prune_threshold: peer_discovery_config.prune_threshold,
         min_active_connections: peer_discovery_config.min_active_connections,
+        max_active_connections: peer_discovery_config.max_active_connections,
         rng_seed: 123456,
     };
     RaptorCast::new(RaptorCastConfig {
--- a/monad-peer-disc-swarm/tests/peer_discovery.rs
+++ b/monad-peer-disc-swarm/tests/peer_discovery.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::BTreeMap,
+    collections::{BTreeMap, BTreeSet},
     net::{Ipv4Addr, SocketAddrV4},
     str::FromStr,
     time::Duration,
@@ -24,7 +24,7 @@ use monad_peer_discovery::{
 use monad_router_scheduler::{NoSerRouterConfig, NoSerRouterScheduler, RouterSchedulerBuilder};
 use monad_testutil::signing::create_keys;
 use monad_transformer::{GenericTransformer, GenericTransformerPipeline, LatencyTransformer};
-use monad_types::NodeId;
+use monad_types::{Epoch, NodeId};
 use tracing_test::traced_test;
 struct PeerDiscSwarm {}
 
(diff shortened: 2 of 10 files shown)


monad-peer-discovery/src/discovery.rs
```diff
@@ -40,6 +40,8 @@ pub struct LookupInfo<ST: CertificateSignatureRecoverable> {
 pub struct PeerDiscovery<ST: CertificateSignatureRecoverable> {
     pub self_id: NodeId<CertificateSignaturePubKey<ST>>,
     pub self_record: MonadNameRecord<ST>,
+    pub current_epoch: Epoch,
+    pub epoch_validators: BTreeMap<Epoch, BTreeSet<NodeId<CertificateSignaturePubKey<ST>>>>,
```
maxkozlovsky, line 44: Do we need any sets except current and next? May be can have two sets here instead of a map of sets?

Does the set need to be sorted? If not, hash set may be better choice.

monad-peer-discovery/src/discovery.rs
```diff
@@ -581,10 +590,41 @@ where
...
+        self.peer_info.retain(|node_id, info| {
+            info.unresponsive_pings < self.prune_threshold || is_validator(node_id)
+        });
+
+        // if still above max active peers, randomly choose a few full nodes and prune them
+        // TODO: do not prune full nodes in ConfirmGroup
+        if self.peer_info.len() > self.max_active_connections {
+            let num_to_prune = self.peer_info.len() - self.max_active_connections;
+            let full_nodes: Vec<_> = self
+                .peer_info
```
maxkozlovsky, line 630: may skip some collects and empty checks 

```
let nodes_to_prune = self
            .peer_info
            .keys()
            .filter_map(|node| (!is_validator(node)).then_some(node))
            .choose_multiple(&mut self.rng, num_to_prune);
```

monad-peer-discovery/src/discovery.rs
```diff
@@ -605,6 +645,32 @@ where
...
+        debug!(
+            ?epoch,
+            ?validators,
+            "updating validator set in peer discovery"
+        );
+
+        let cmds = Vec::new();
+        self.epoch_validators.insert(epoch, validators);
+
+        cmds
```
maxkozlovsky, line 698: Simply return Vec:new() without temporary? The other implementation below does that

monad-peer-discovery/src/discovery.rs
```diff
@@ -581,10 +590,41 @@ where
...
-        // TODO: prioritization of validator vs full node
-        self.peer_info
-            .retain(|_, info| info.unresponsive_pings < self.prune_threshold);
+        // we currently do not prune validators in current and next epoch
+        self.peer_info.retain(|node_id, info| {
+            info.unresponsive_pings < self.prune_threshold || is_validator(node_id)
+        });
+
+        // if still above max active peers, randomly choose a few full nodes and prune them
+        // TODO: do not prune full nodes in ConfirmGroup
```
maxkozlovsky, line 626: Would it be simpler to avoid adding nodes into peer_info if it is already at max_active_connections?

monad-peer-discovery/src/discovery.rs
```diff
@@ -605,6 +654,32 @@ where
...
     }
 
+    fn update_current_epoch(&mut self, epoch: Epoch) -> Vec<PeerDiscoveryCommand<ST>> {
+        debug!(?epoch, "updating current epoch in peer discovery");
+
+        let cmds = Vec::new();
+        self.current_epoch = epoch;
+
+        cmds
+    }
```
eerkaijun: to clear old epoch as well

monad-peer-disc-swarm/tests/peer_discovery.rs
```diff
@@ -515,6 +523,11 @@ fn test_peer_lookup_random_nodes() {
         target: nonexistent_id,
         open_discovery: true,
     };
+    let add_validators_event = PeerDiscoveryEvent::UpdateValidatorSet {
+        epoch: Epoch(0),
```
michael-yxchen: super nit: epoch starts from 1

monad-peer-discovery/src/discovery.rs
eerkaijun: we will be prioritizing validators, so we should still insert them but prune away full nodes. I think in the future we plan to have better connection heuristics too, e.g. nodes might have reputation depending on how often they are offline etc

monad-peer-discovery/src/discovery.rs
```diff
@@ -77,10 +82,13 @@ impl<ST: CertificateSignatureRecoverable> PeerDiscoveryAlgoBuilder for PeerDisco
         >,
     ) {
         assert!(self.ping_period > self.request_timeout);
+        assert!(self.max_active_connections > self.min_active_connections);
 
         let mut state = PeerDiscovery {
             self_id: self.self_id,
             self_record: self.self_record,
+            current_epoch: Epoch(0),
+            epoch_validators: Default::default(),
```
michael-yxchen: we should initialize these from the builder, similar to `MonadStateBuilder`

monad-peer-discovery/src/discovery.rs
```diff
@@ -576,15 +593,47 @@ where
...
         *self.metrics.entry("refresh").or_default() += 1;
         let mut cmds = Vec::new();
 
+        // check whether a node id is a validator in the current and next epoch
+        let current_epoch_validators = self.epoch_validators.get(&self.current_epoch);
+        let next_epoch_validators = self.epoch_validators.get(&(self.current_epoch + Epoch(1)));
+        let is_validator = |node_id: &NodeId<CertificateSignaturePubKey<ST>>| {
+            current_epoch_validators.is_some_and(|validators| validators.contains(node_id))
+                || next_epoch_validators.is_some_and(|validators| validators.contains(node_id))
+        };
```
michael-yxchen, line 612: should we make this a helper function?

monad-peer-discovery/src/discovery.rs
eerkaijun: thanks, updated

monad-peer-discovery/src/discovery.rs
eerkaijun: right now i believe validators are loaded through the `validators.toml` file, which can specify validator set arbitrarily in the future. @michael-yxchen are these validators being loaded into memory immediately on startup?

monad-peer-discovery/src/discovery.rs
```diff
@@ -576,15 +593,41 @@ where
...
+                .peer_info
+                .keys()
+                .filter_map(|node| (!is_validator(node)).then_some(node))
+                .cloned()
+                .choose_multiple(&mut self.rng, num_to_prune);
+
+            for node_id in nodes_to_prune {
+                debug!(?node_id, "pruning excessive full nodes");
+                self.peer_info.remove(&node_id);
+            }
```
michael-yxchen, line 645: just a note that this may cause high churn in peers. We're unconditionally adding peers sending us a ping message and from peer lookup responses. Adding too many cna trigger a prune which removes older nodes

I think other implementations address this by
  - considering node age when pruning
  - adding new nodes to a queue instead of immediately to peer list

we can address this later

monad-peer-discovery/src/discovery.rs
michael-yxchen: no only locked ones (listed in `forkpoint.toml`) are loaded on startup. The rest are loaded as respective epochs are locked. This pattern is used everywhere else we keep `epoch_validators` though, so I would suggest keeping the pattern for now

Currently, there's a short window where we can have 3 validator sets. That might change soon

monad-peer-discovery/src/discovery.rs
```diff
@@ -1127,6 +1232,63 @@ mod tests {
...
+            }),
+            (peer3_pubkey, PeerInfo {
+                last_ping: None,
+                unresponsive_pings: 0,
+                name_record: generate_name_record(peer3, 0),
+            }),
+        ]);
+        // Peer1 and Peer2 in validator set
+        let mut epoch_validators = BTreeMap::new();
+        epoch_validators.insert(Epoch(0), BTreeSet::from([peer1_pubkey, peer2_pubkey]));
```
michael-yxchen: nit: epoch starts at 1

monad-peer-discovery/src/discovery.rs
```diff
@@ -576,15 +593,41 @@ where
...
+        if self.peer_info.len() > self.max_active_connections {
+            let num_to_prune = self.peer_info.len() - self.max_active_connections;
+            let nodes_to_prune = self
+                .peer_info
+                .keys()
+                .filter_map(|node| (!is_validator(node)).then_some(node))
+                .cloned()
+                .choose_multiple(&mut self.rng, num_to_prune);
+
+            for node_id in nodes_to_prune {
```
michael-yxchen: we can log when `nodes_to_prune` is empty, i.e. we have more validators than `high_watermark`

monad-peer-discovery/src/discovery.rs
eerkaijun: great point, I think we should definitely have better heuristics.

on a separate but slightly relevant note, i think initially we should also recommend people to set `max_active_connections` to be higher than validator set size

michael-yxchen:
in `node.toml` we have a list of dedicated full nodes. We should also make them persistent

monad-peer-discovery/src/discovery.rs
eerkaijun: trying to keep similar pattern. in mock.rs I return Vec::new() since that's consistent for all functions

monad-node/src/main.rs
```diff
@@ -562,15 +566,36 @@ where
         .collect();
 
     // TODO: make it configurable in node.toml
+    let current_epoch = locked_epoch_validators
+        .first()
+        .expect("at least one epoch should be specified")
```
eerkaijun: is this a safe assumption

monad-peer-discovery/src/discovery.rs
eerkaijun: i tried but can't get it to work - self.is_validator borrows immutable self and self.rng needs it to be mutable. if we put self.is_validator as mutable then there's borrow conflict with self.peer_info

eerkaijun:
> in `node.toml` we have a list of dedicated full nodes. We should also make them persistent

added, now keeping track of dedicated full nodes in peer discovery

monad-peer-discovery/src/discovery.rs
michael-yxchen: ok let's leave it for now

monad-node/src/main.rs
michael-yxchen: can you use whatever MonadStateBuilder is using? getting from high_qc i believe

eerkaijun: updated

monad-peer-discovery/src/discovery.rs
```diff
@@ -576,15 +598,52 @@ where
...
+                .peer_info
+                .keys()
+                .filter_map(|node| {
+                    (!is_validator(node) && !is_dedicated_full_node(node)).then_some(node)
+                })
+                .cloned()
+                .choose_multiple(&mut self.rng, num_to_prune);
+
+            if nodes_to_prune.is_empty() {
+                debug!("more validators than max active connections");
```
michael-yxchen: nit: i think this can be bumped to info. since we're pinning dedicated full nodes, should we update the log to say "pinned nodes" instead of validators?

monad-peer-discovery/src/discovery.rs
eerkaijun: https://github.com/category-labs/category-internal/issues/1619

PR #1815, opened 2025-05-21
Full-node Raptorcast: Refactored Group message de/serialization

fabiotdf:
This is part of a series of PR split from the original Full-node RaptorCast integration.
This PR contains just the Group messages, take #2

### DIFF
--- a/monad-raptorcast/src/lib.rs
+++ b/monad-raptorcast/src/lib.rs
@@ -34,6 +34,7 @@ use monad_types::{
 use tracing::{debug, error, warn};
 
 pub mod message;
+pub mod raptorcast_secondary;
 pub mod udp;
 pub mod util;
 use util::{BuildTarget, EpochValidators, FullNodes, ReBroadcastGroupMap, Validator};
@@ -510,6 +511,10 @@ where
                                     }
                                     None
                                 }
+                                InboundRouterMessage::FullNodesGroup(_full_nodes_group_message) => {
+                                    warn!("Handling of FullNodesGroup implemented in upcoming PR");
+                                    None
+                                }
                             },
                             Err(err) => {
                                 warn!(
@@ -569,6 +574,11 @@ where
                         ?message,
                         "dropping peer discovery message, should come through udp channel"
                     );
+                    continue;
+                }
+                InboundRouterMessage::FullNodesGroup(_group_message) => {
+                    // pass TCP message to MultiRouter
+                    warn!("FullNodesGroup protocol via TCP not implemented");
                     continue;
                 }
             }
--- a/monad-raptorcast/src/message.rs
+++ b/monad-raptorcast/src/message.rs
@@ -4,6 +4,8 @@ use monad_compress::{zstd::ZstdCompression, CompressionAlgo};
 use monad_crypto::certificate_signature::CertificateSignatureRecoverable;
 use monad_peer_discovery::PeerDiscoveryMessage;
 
+use super::raptorcast_secondary::group_message::FullNodesGroupMessage;
+
 const SERIALIZE_VERSION: u32 = 1;
 // compression versions
 const UNCOMPRESSED_VERSION: u32 = 1;
@@ -24,9 +26,14 @@ impl NetworkMessageVersion {
     }
 }
 
+const MESSAGE_TYPE_APP: u8 = 1;
+const MESSAGE_TYPE_PEER_DISC: u8 = 2;
+const MESSAGE_TYPE_GROUP: u8 = 3;
+
 pub enum OutboundRouterMessage<OM, ST: CertificateSignatureRecoverable> {
     AppMessage(OM),
     PeerDiscoveryMessage(PeerDiscoveryMessage<ST>),
+    FullNodesGroup(FullNodesGroupMessage<ST>),
 }
 
 #[derive(Debug)]
@@ -41,7 +48,7 @@ impl<OM: Encodable, ST: CertificateSignatureRecoverable> OutboundRouterMessage<O
             Self::AppMessage(app_message) => {
                 if version.compression_version == UNCOMPRESSED_VERSION {
                     // encode as uncompressed message
-                    let enc: [&dyn Encodable; 3] = [&version, &1u8, &app_message];
+                    let enc: [&dyn Encodable; 3] = [&version, &MESSAGE_TYPE_APP, &app_message];
                     encode_list::<_, dyn Encodable>(&enc, &mut buf);
                 } else if version.compression_version == DEFAULT_ZSTD_VERSION {
                     // compress message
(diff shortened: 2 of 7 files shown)


monad-raptorcast/src/lib.rs
```diff
@@ -510,6 +511,12 @@ where
                                     }
                                     None
                                 }
+                                InboundRouterMessage::FullNodesGroup(full_nodes_group_message) => {
+                                    tracing::warn!(
```
eerkaijun: nit: `warn` is already imported at start of file, switch to just `warn!`

monad-raptorcast/src/lib.rs
```diff
@@ -571,6 +578,10 @@ where
                     );
                     continue;
                 }
+                InboundRouterMessage::FullNodesGroup(_group_message) => {
+                    // pass TCP message to MultiRouter
+                    tracing::warn!("FullNodesGroup protocol via TCP not implemented");
```
eerkaijun: same as above

monad-raptorcast/src/message.rs
```diff
@@ -127,6 +136,11 @@ impl<M: Decodable, ST: CertificateSignatureRecoverable> InboundRouterMessage<M,
                     PeerDiscoveryMessage::decode(&mut payload).map_err(DeserializeError::from)?;
                 Ok(Self::PeerDiscoveryMessage(peer_disc_message))
             }
+            3 => {
```
eerkaijun: in the previous PR, this is put into a constant, shall we still do that?

monad-raptorcast/src/raptorcast_secondary/group_message.rs
```diff
@@ -0,0 +1,236 @@
...
+    pub node_id: NodeId<CertificateSignaturePubKey<ST>>,
+    pub accept: bool,
+}
+
+#[derive(Debug, Clone, RlpEncodable, RlpDecodable)]
+#[rlp(trailing)]
+pub struct ConfirmGroup<ST: CertificateSignatureRecoverable> {
+    pub prepare: PrepareGroup<ST>,
+    pub peers: Vec<NodeId<CertificateSignaturePubKey<ST>>>,
+    pub name_records: Option<Vec<MonadNameRecord<ST>>>,
```
eerkaijun: should we switch this to `Vec<MonadNameRecord<ST>>`?

monad-raptorcast/src/raptorcast_secondary/group_message.rs
```diff
@@ -0,0 +1,236 @@
...
+
+    fn length(&self) -> usize {
+        size_of_val(&GROUP_MSG_VERSION)
+            + size_of_val(&MESSAGE_TYPE_PREP_REQ)
+            + match self {
+                Self::PrepareGroup(prep_req) => prep_req.length(),
+                Self::PrepareGroupResponse(prep_res) => prep_res.length(),
+                Self::ConfirmGroup(conf_grp) => conf_grp.length(),
+            }
+    }
```
eerkaijun: is this needed? I think there's a default implementation

monad-raptorcast/src/message.rs
fabiotdf: Dunno, seems the literal number is used everywhere else? I just copied the pattern

monad-raptorcast/src/raptorcast_secondary/group_message.rs
fabiotdf: I'm fine with changing, but Michael's spec was explicit about this being optional. Is that still the case?

monad-raptorcast/src/raptorcast_secondary/group_message.rs
fabiotdf: The default encodes the message to determine the size, so essentially encoded twice.
It is not needed but recommended

monad-raptorcast/src/lib.rs
fabiotdf: Done

monad-raptorcast/src/lib.rs
fabiotdf: Done

monad-raptorcast/src/message.rs
fabiotdf: Added back

monad-raptorcast/src/raptorcast_secondary/group_message.rs
michael-yxchen: yeah we can update to `Vec<..>`. i'll update the spec

monad-raptorcast/src/raptorcast_secondary/group_message.rs
```diff
@@ -0,0 +1,236 @@
...
+        );
+
+        let decoded_enum =
+            FullNodesGroupMessage::<ST>::decode(&mut encoded_bytes.as_slice()).unwrap();
+        match decoded_enum {
+            FullNodesGroupMessage::PrepareGroup(dser_msg) => {
+                assert_eq!(dser_msg.validator_id, org_msg.validator_id);
+                assert_eq!(dser_msg.max_group_size, org_msg.max_group_size);
+                assert_eq!(dser_msg.start_round, org_msg.start_round);
+                assert_eq!(dser_msg.end_round, org_msg.end_round);
```
michael-yxchen: nit: we can derive/impl Eq and just do `assert_eq!(org_enum, decoded_enum)`

monad-raptorcast/src/raptorcast_secondary/group_message.rs
michael-yxchen: yes overriding length impl is recommended. I feel the calculation is not correct though. It's missing rlp list header (1 or more bytes depending on total payload length). Also `size_of_val` is only correct for values in `[0x00, 0x7f]`. Values above 0x7f are encoded with two bytes. Let's use default impl and optimize it later

monad-raptorcast/src/raptorcast_secondary/group_message.rs
```diff
@@ -0,0 +1,236 @@
...
+                Self::ConfirmGroup(conf_grp) => conf_grp.length(),
+            }
+    }
+}
+
+impl<ST: CertificateSignatureRecoverable> Decodable for FullNodesGroupMessage<ST> {
+    fn decode(buf: &mut &[u8]) -> alloy_rlp::Result<Self> {
+        let mut payload = Header::decode_bytes(buf, true)?;
+        let version = u8::decode(&mut payload)?;
+        if version > GROUP_MSG_VERSION {
```
michael-yxchen: should that be `!=`?

monad-raptorcast/src/message.rs
michael-yxchen: either is fine for now

monad-raptorcast/src/raptorcast_secondary/group_message.rs
fabiotdf: Ah yes. Removed

monad-raptorcast/src/raptorcast_secondary/group_message.rs
fabiotdf: The intention was > because the code will know about older versions (and can act accordingly) but not newer.
The version deserialized can still be 0 though! Fixed the check

monad-raptorcast/src/raptorcast_secondary/group_message.rs
fabiotdf: Done

monad-raptorcast/src/raptorcast_secondary/group_message.rs
fabiotdf: Done

PR #1816, opened 2025-05-21
Update vulnerable packages

eclark0426:
Turns out I did a silly, and pretty much everything could be fixed automatically just with `cargo update` 🤦🏻‍♀️
Was thinking at first I needed to manually update all the packages which depending on vulnerable packages....


eerkaijun:
i presume the keystore generation has been tested to work the same after the package change?

eclark0426:
> i presume the keystore generation has been tested to work the same after the package change?

I ran it locally, and from what I could tell, it was the same. But it might be worth verifying that going through the full process of using the keystore generation for a new node still works

eclark0426:
Dismissing old reviews since I added a Rust version change. On rebase+update, one of the dependencies was requiring a newer rustc version.

monad-archive/src/model/block_data_archive.rs
```diff
@@ -700,7 +700,7 @@ mod tests {
                             address: Default::default(),
                             data: LogData::new(
                                 vec![],
-                                repeat(42).take(20).collect::<Vec<u8>>().into(),
+                                repeat_n(42, 20).collect::<Vec<u8>>().into(),
```
dshulyak, line 703: seems unrelated to update

monad-dataplane/src/tcp/tx.rs
```diff
@@ -267,13 +262,10 @@ async fn connect_and_send_messages(
...
-                ErrorKind::Other,
-                format!(
-                    "error writing message {} on TCP connection: {}",
-                    message_id, err
-                ),
-            )
+            Error::other(format!(
+                "error writing message {} on TCP connection: {}",
+                message_id, err
+            ))
```
dshulyak, line 268: thats std::io:Error ? seems unrelated

monad-archive/src/model/block_data_archive.rs
eclark0426: Yes, these changes are because of clippy after the Rust upgrade

PR #1819, opened 2025-05-22
Move peer discovery config to node.toml

eerkaijun:
In preparation for stressnet testing

### DIFF
--- a/docker/devnet/monad/config/node.toml
+++ b/docker/devnet/monad/config/node.toml
@@ -34,7 +34,14 @@ peers = []
 # secp256k1_pubkey = "..."
 
 [peer_discovery]
-local_record_seq_num = 0
+self_address = "0.0.0.0:8000"
+self_record_seq_num = 0
+self_name_record_sig = "7911fbf0df1307723c51658349ed396dd6ee28fde925277bae53fcd77705c4313e003a12fb562e1c20cc616511f7a09f986a3f6af6849cdedff6f5684557b04f01"
+ping_period = 30
+refresh_period = 120
+request_timeout = 5
+prune_threshold = 5
+min_active_connections = 0
 peers = []
 # [[peer_discovery.peers]]
 # address = "x.x.x.x:xxxx"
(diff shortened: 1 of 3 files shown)


docker/devnet/monad/config/node.toml
```diff
@@ -34,7 +34,14 @@ peers = []
 # secp256k1_pubkey = "..."
 
 [peer_discovery]
-local_record_seq_num = 0
+self_address = "0.0.0.0:8000"
```
eclark0426, line 37: Is there ever a case where this wouldn't be identical to `{network.bind_address_host}:{network.bind_address_port}`? Seems like it's just a duplicate of each other, with the potential to get out of sync

michael-yxchen: they would be different if node is behind NAT

eclark0426: Is `self_address` what gets sent out to other peers as the address to connect to? I'm not entirely sure how it actually gets used. If yes, is `0.0.0.0` the best default? I'm not sure if I have a better alternative, but it just seems a bit odd to me to use an invalid IP as a default

michael-yxchen: yeah you're right. `0.0.0.0` is unreasonable to send out as own name record. I don't know what's a reasonable value though

eerkaijun: this node.toml is only used to start a single node btw, in actual network we will use the actual ip

eclark0426: Maybe the default should be `127.0.0.1` or `localhost` then?

eerkaijun: can switch to whatever but i don't think it matter for a single node. this is just to keep it consistent with the bind address

eclark0426: Sure, in the grand scheme it's unimportant since this shouldn't be used outside of a single node. I still think `127.0.0.1` would make more sense though since this isn't a bind address but where the (nonexistent) other nodes would connect to

PR #1821, opened 2025-05-22
Sign name record program

eerkaijun:
A binary for a node to sign its name record (read private key from keystore file and sign on ip)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5081,9 +5081,13 @@ name = "monad-peer-discovery"
 version = "0.1.0"
 dependencies = [
  "alloy-rlp",
+ "clap 4.4.10",
  "futures",
+ "hex",
  "monad-crypto",
  "monad-executor-glue",
+ "monad-keystore",
+ "monad-node-config",
  "monad-secp",
  "monad-testutil",
  "monad-types",
@@ -5091,6 +5095,7 @@ dependencies = [
  "rand_chacha 0.3.1",
  "test-case",
  "tokio-util",
+ "toml 0.7.8",
  "tracing",
 ]
 
--- a/docker/devnet/Dockerfile
+++ b/docker/devnet/Dockerfile
@@ -70,7 +70,7 @@ ARG TRIEDB_TARGET=triedb_driver
 # Builder
 COPY . .
 RUN ASMFLAGS="-march=haswell" CFLAGS="-march=haswell" CXXFLAGS="-march=haswell -DQUILL_ACTIVE_LOG_LEVEL=QUILL_LOG_LEVEL_CRITICAL" \
-    CC=gcc-15 CXX=g++-15 cargo build --release --bin monad-node --bin monad-keystore --bin monad-debug-node --example ledger-tail --example wal2json --example wal-tool --example triedb-debug --example triedb-bench && \
+    CC=gcc-15 CXX=g++-15 cargo build --release --bin monad-node --bin monad-keystore --bin monad-debug-node --example ledger-tail --example wal2json --example wal-tool --example triedb-debug --example triedb-bench --example sign-name-record && \
     mv target/release/monad-node monad-node && \
     mv target/release/monad-keystore keystore && \
     mv target/release/monad-debug-node monad-debug-node && \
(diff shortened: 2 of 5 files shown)


monad-peer-discovery/examples/sign_record.rs
```diff
@@ -0,0 +1,43 @@
...
+    #[arg(long)]
+    port: u16,
+
+    /// Local record sequence number
+    #[arg(long, default_value_t = 0)]
+    local_record_seq_num: u64,
+
+    /// Private key
+    #[arg(long)]
+    private_key: String,
```
dshulyak: i think it is better to load it from file so it doesn't stay in console history and there won't be a need to paste it.
if they key is important i would also expect to store it encrypted, but i don't know what is the purpose of it.
is it just a p2p key or long term validator key that they are using for staking/payouts?

monad-peer-discovery/examples/sign_record.rs
```diff
@@ -0,0 +1,43 @@
...
+
+/// Example command to run the following program
+/// cargo run --example sign_record -- --ip 0.0.0.0 --port 8888 --private-key <...>
+
+#[derive(Debug, Parser)]
+#[command(name = "monad-peer-discovery", about, long_about = None)]
+struct Args {
+    /// IP address in format x.x.x.x
+    #[arg(long)]
+    ip: String,
```
dshulyak: can be parsed directly into Ipv4Addr. also help will be more useful than docstring here

monad-peer-discovery/examples/sign_record.rs
```diff
@@ -0,0 +1,43 @@
...
+
+    /// Port number
+    #[arg(long)]
+    port: u16,
```
dshulyak: value_parser = clap::value_parser!(u16).range(1..=65535),
help = "port number between 1 and 65535"

monad-peer-discovery/examples/sign_record.rs
```diff
@@ -0,0 +1,43 @@
...
+
+    /// Local record sequence number
+    #[arg(long, default_value_t = 0)]
+    local_record_seq_num: u64,
```
dshulyak: help with explanation when would you change it will be useful

monad-peer-discovery/examples/sign_record.rs
```diff
@@ -0,0 +1,43 @@
+use std::net::{Ipv4Addr, SocketAddrV4};
+
+use clap::Parser;
+use monad_peer_discovery::{MonadNameRecord, NameRecord};
+use monad_secp::{KeyPair, SecpSignature};
+
+/// Example command to run the following program
+/// cargo run --example sign_record -- --ip 0.0.0.0 --port 8888 --private-key <...>
```
dshulyak: examples would be nice to put into long_about

monad-peer-discovery/examples/sign_record.rs
eerkaijun: yeap, the plan is to read from keystore.json file (same key as the secp voting key), this is a temporary solution to deploy it to stressnet

monad-peer-discovery/examples/sign_record.rs
dshulyak: realized that it is u16. clap already provides good enough validation
> error: invalid value '1111111' for '--port <PORT>': 1111111 is not in 0..=65535

monad-peer-discovery/examples/sign_record.rs
eerkaijun: I realized that clap will parse the docstring into about

monad-peer-discovery/examples/sign_record.rs
michael-yxchen, line 1: nit: can we change it to `sign-name-record` to match naming convention of other binaries shipped with docker?

eerkaijun:
reading ip and port from node.toml and pasting into it again feels circular to me, I think these should just be command line arguments and we can output the signature in the formatted version to be pasted into node.toml

michael-yxchen:
i made a mistake in the above feedback. i meant reading ip/port from cmdline, and seq_num from `node.toml`

monad-peer-discovery/examples/sign_record.rs
eerkaijun: updated

PR #1824, opened 2025-05-27
Fix: Don't block rpc startup on eth_getLogs index creation

JoeHowarth:
The ensures that:
  - RPCs no longer block for 12-40 minutes during startup when the index doesn't exist
  - The index will still be created in the background for future eth_getLogs queries
  - Writers/indexers continue to block on index creation as they need it before writing
  - The system maintains backward compatibility

  fixes https://github.com/category-labs/category-internal/issues/1626

### DIFF
--- a/monad-archive/src/model/logs_index.rs
+++ b/monad-archive/src/model/logs_index.rs
@@ -5,6 +5,7 @@ use mongodb::{
     bson::{doc, spec::BinarySubtype, Binary, Bson, Document},
     Collection, IndexModel,
 };
+use tracing::info;
 
 use crate::{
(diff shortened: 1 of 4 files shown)


monad-archive/src/model/logs_index.rs
```diff
@@ -40,13 +42,42 @@ impl LogsIndexArchiver {
...
         };
-        collection
-            .create_index(IndexModel::builder().keys(keys).build())
-            .await
-            .inspect_err(|e| {
-                error!("Failed to create index: {e:?}");
-            })
-            .ok()?;
+
+        if is_reader {
```
mikety, line 46: nit: add to metrics the status of the index store / index

PR #1826, opened 2025-05-28
Full-node Raptorcast: config for secondary instance

fabiotdf:
Isolates the config changes needed for the secondary (full-node) raptorcast instance

### DIFF
--- a/monad-node-config/src/fullnode_raptorcast.rs
+++ b/monad-node-config/src/fullnode_raptorcast.rs
@@ -1,0 +1,37 @@
+use monad_crypto::certificate_signature::PubKey;
+use monad_types::Round;
+use serde::{Deserialize, Serialize};
+
+use super::fullnode::FullNodeConfig;
+
+#[derive(Debug, Deserialize, Serialize, Clone)]
+#[serde(deny_unknown_fields)]
+pub enum SecondaryRaptorCastModeConfig {
+    Client,
+    Publisher,
+    // No "None" option, as FullNodeRaptorCastConfig is now wrapped in Option<>
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone)]
+#[serde(deny_unknown_fields)]
+pub struct FullNodeRaptorCastConfig<P: PubKey> {
+    pub mode: SecondaryRaptorCastModeConfig,
+
+    #[serde(bound = "P:PubKey")]
+    pub full_nodes_prioritized: FullNodeConfig<P>,
+    pub raptor10_fullnode_redundancy_factor: u8, // validator -> full-nodes
+
+    // RaptorCastConfigSecondaryPublisher::GroupSchedulingConfig
+    pub max_group_size: usize,
+    pub round_span: Round,
+    pub invite_lookahead: Round,
+    pub max_invite_wait: Round,
+    pub deadline_round_dist: Round,
+    pub init_empty_round_span: Round,
+
+    // RaptorCastConfigSecondaryClient
+    pub bandwidth_cost_per_group_member: u64,
+    pub bandwidth_capacity: u64,
+    pub invite_future_dist_min: Round,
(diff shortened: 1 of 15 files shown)


monad-node-config/src/lib.rs
```diff
@@ -24,9 +24,13 @@ mod consensus;
...
 mod network;
 mod peers;
+
+pub mod fullnode_raptorcast;
+pub use fullnode_raptorcast::FullNodeRaptorCastConfig;
+
 mod sync_peers;
 
-#[derive(Debug, Deserialize)]
+#[derive(Debug, Deserialize, Clone)]
```
eerkaijun, line 33: do we need clone here?

monad-node-config/src/fullnode_raptorcast.rs
```diff
@@ -0,0 +1,37 @@
...
+    pub full_nodes_prioritized: FullNodeConfig<P>,
+    pub raptor10_redundancy_fullnodes: u8, // validator -> full-nodes
+
+    // RaptorCastConfigSecondaryPublisher::GroupSchedulingConfig
+    pub max_group_size: usize,
+    pub round_span: Round,
+    pub invite_lookahead: Round,
+    pub max_invite_wait: Round,
+    pub deadline_round_dist: Round,
+    pub init_empty_round_span: Round,
```
eerkaijun, line 30: should we reuse GroupSchedulingConfig struct if it's the same?

monad-node-config/src/fullnode_raptorcast.rs
```diff
@@ -0,0 +1,37 @@
...
+
+    // RaptorCastConfigSecondaryClient
```
eerkaijun, line 32: same as above

monad-node-config/src/lib.rs
```diff
@@ -60,6 +64,10 @@ pub struct NodeConfig<ST: CertificateSignatureRecoverable> {
     #[serde(bound = "ST: CertificateSignatureRecoverable")]
     pub peer_discovery: PeerConfig<ST>,
 
+    pub raptor10_redundancy_validators: u8, // validator -> validator
+    #[serde(bound = "ST: CertificateSignatureRecoverable")]
+    pub fullnode_raptorcast: FullNodeRaptorCastConfig<CertificateSignaturePubKey<ST>>,
```
eerkaijun: my thought is that `fullnode_raptorcast` would be an optional field, i.e. not all validators will want to participate in full node broadcast, if so they can just leave this empty

monad-node-config/src/lib.rs
fabiotdf: It does make things easier later when building up the new multi-router. Is there a problem with config being clonable?

monad-node-config/src/fullnode_raptorcast.rs
fabiotdf: I had that originally, but not all config structs could be used as-is.
Also, you can't include monad-node-config from monad-raptorcast or vice versa, so it was more trouble than worth it. But if you have a strong opinion on this, I can implement. WDYT?

monad-node-config/src/lib.rs
eerkaijun: no, was just curious

monad-node-config/src/fullnode_raptorcast.rs
eerkaijun: ok let's just leave it as it is then (there shouldn't be any circular dependency right now but maybe that arises in your following prs, we can refactor later)

fabiotdf: Alright

monad-node-config/src/lib.rs
fabiotdf: Fixed

PR #1827, opened 2025-05-29
monad-dataplane: utility to spam node at a configurabl rate with configurable number of connections

dshulyak:
spam binary: can be used to send tcp/udp traffic using dataplane. implements simple udp protocol to split/reconstruct user messages

tc.sh: script to setup up to namespaces with emulated network conditions


dshulyak:
i am actually not sure if anyone will be using this script for testing. if you think that it shouldn't be here, let me know and i will keep separately

dshulyak:
removed script to setup latency/throughput limits, will push it elsewhere if someone will want to use that

README.md
```diff
@@ -25,28 +25,99 @@ To run a Monad consensus client, follow instructions [here](monad-node/README.md
...
+
+
+
+
+
+
+
+A guide for using the `tc.sh` script with the `spam` performance testing tool to simulate network conditions and measure performance.
+
+## Quick Start
```
eerkaijun: shall this be moved to a subfolder readme rather than the repo main readme?

dshulyak: sorry forgot to revert it

PR #1829, opened 2025-05-29
monad-dataplane: address list to track connection status and task to expire banned nodes

dshulyak:
two modules that are used in the followup prs:

addrlist tracks banned / trusted ip connections. banned connections are meant to be dropped on accept, trusted connections will bypass total connections limit

ban_expiry implements a future to unban connections after timeout.

### DIFF
--- a/monad-dataplane/src/addrlist.rs
+++ b/monad-dataplane/src/addrlist.rs
@@ -81,76 +81,126 @@ impl AddrEntry {
     }
 }
 
+#[derive(Debug)]
+struct Entries(HashMap<IpAddr, AddrEntry>);
+
 #[derive(Debug, Clone)]
 pub(crate) struct Addrlist {
-    entries: Arc<Mutex<HashMap<IpAddr, AddrEntry>>>,
+    entries: Arc<Mutex<Entries>>,
+}
+
+impl Entries {
+    fn new() -> Self {
+        Self(HashMap::new())
+    }
+
+    fn add_trusted(&mut self, addr: &IpAddr) {
+        self.0
+            .entry(*addr)
+            .and_modify(|e| e.set_trusted(true))
+            .or_insert_with(AddrEntry::new_trusted);
+    }
+
+    fn remove_trusted(&mut self, addr: &IpAddr) {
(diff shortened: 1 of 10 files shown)


monad-dataplane/src/addrlist.rs
```diff
@@ -0,0 +1,228 @@
...
+            .entry(addr)
+            .and_modify(|e| e.ban_at(timestamp))
+            .or_insert_with(|| AddrEntry::new_banned_at(timestamp));
+    }
+
+    pub(crate) fn unban(&self, addr: &IpAddr) {
+        let mut entries = self.entries.lock().unwrap();
+        if let Some(entry) = entries.get_mut(addr) {
+            entry.unban();
+            if !entry.is_trusted() {
```
mikety, line 128: From the code follow that `trusted` address is always in the map banned or unbanned, is this a "known" address, so there is no need to repopulate it? If so should it have different banning policy?

monad-dataplane/src/ban_expiry.rs
```diff
@@ -0,0 +1,179 @@
...
+            } => {
+                while let Some((addr, timestamp)) = queue.front() {
+                    // we track two timestamps
+                    // 1. when ip address was last banned
+                    // 2. and when it was inserted into the queue for expiry
+                    // we separate them so that ban can be re-newed before original ban has expired
+                    if timestamp.elapsed() >= ban_duration {
+                        addrlist.banned_at(addr).is_some_and(|banned_at| banned_at.elapsed() >= ban_duration)
+                            .then(|| {
+                                debug!("unban {}", addr);
```
mikety: nit: `debug!(?addr, "unban")` is json -  better for parsing logs

monad-dataplane/src/addrlist.rs
dshulyak: trusted is meant to be used for validators. 

i don't think that they require different banning policy in dataplane.

ban should be reserved for clearly misbehaving node (either due to bug or maliciously). 
maybe in the protocols (statesync/blocksync) it will make sense to set different rate limits for requests that are considered misbehaving

PR #1830, opened 2025-05-29
Switch to peer discovery

eerkaijun:
Switch over completely to use peer discovery (remove `known_addresses` in `monad-raptorcast`). Also allow debug-node to manually insert more peers and also get the current peer list. Tested with flexnet, and also confirmed that debug-node works as expected after the change. Tested and upgraded a few stressnet nodes

Main changes in `monad-peer-discovery`, `monad-raptorcast` and `monad-updaters/config-loader.rs`


monad-types/src/lib.rs
```diff
@@ -413,6 +416,21 @@ where
     ST::deserialize(bytes.as_ref()).map_err(<D::Error as serde::de::Error>::custom)
 }
 
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+pub struct PeerEntry<ST: CertificateSignatureRecoverable> {
+    #[serde(serialize_with = "serialize_pubkey::<_, CertificateSignaturePubKey<ST>>")]
+    #[serde(deserialize_with = "deserialize_pubkey::<_, CertificateSignaturePubKey<ST>>")]
+    #[serde(bound = "ST: CertificateSignatureRecoverable")]
+    pub pubkey: CertificateSignaturePubKey<ST>,
+    pub addr: SocketAddr,
```
michael-yxchen: should we make this `SocketAddrV4`?

monad-updaters/src/config_loader.rs
```diff
@@ -155,37 +101,44 @@ where
...
+            .bootstrap
+            .peers
+            .into_iter()
+            .map(|peer| PeerEntry {
+                pubkey: peer.secp256k1_pubkey,
+                addr: SocketAddr::V4(peer.address),
+                signature: peer.name_record_sig,
+                record_seq_num: peer.record_seq_num,
+            })
+            .collect();
```
michael-yxchen: we should verify name records here

edit: i guess verifying in peer discovery state machine is also fine

monad-proto/proto/event.proto
```diff
@@ -151,8 +151,10 @@ message ProtoSockAddr {
 }
 
 message ProtoPeerRecord {
-  monad_proto.basic.ProtoNodeId node_id = 1;
+  monad_proto.basic.ProtoPubkey pubkey = 1;
```
michael-yxchen: why are we changing the type here?

monad-types/src/lib.rs
```diff
@@ -413,6 +416,21 @@ where
     ST::deserialize(bytes.as_ref()).map_err(<D::Error as serde::de::Error>::custom)
 }
 
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+pub struct PeerEntry<ST: CertificateSignatureRecoverable> {
```
michael-yxchen: is it possible to use `MonadNameRecord` here? reduces struct transcribing

monad-updaters/src/config_loader.rs
eerkaijun: yeap verified in the `update_peers` function in peer discovery

monad-types/src/lib.rs
eerkaijun: initially i was also planning to use `MonadNameRecord` but switched to a new struct for two main reasons
- want to have a pubkey field so we don't have to `recover_pubkey`. we don't need to store pubkey in `MonadNameRecord` as it's stored as mappings of NodeId -> MonadNameRecord
- don't want to pollute `MonadNameRecord` with serialization and deserialization

monad-proto/proto/event.proto
eerkaijun: `PeerEntry` now uses pubkey, previously it was a tuple of NodeId and SocketAddr

monad-node-config/src/bootstrap.rs
```diff
@@ -1,21 +1,33 @@
...
+    #[serde(serialize_with = "serialize_pubkey::<_, CertificateSignaturePubKey<ST>>")]
+    #[serde(deserialize_with = "deserialize_pubkey::<_, CertificateSignaturePubKey<ST>>")]
+    #[serde(bound = "ST: CertificateSignatureRecoverable")]
+    pub secp256k1_pubkey: CertificateSignaturePubKey<ST>,
+
+    #[serde(serialize_with = "serialize_certificate_signature::<_, ST>")]
+    #[serde(deserialize_with = "deserialize_certificate_signature::<_, ST>")]
+    #[serde(bound = "ST: CertificateSignatureRecoverable")]
+    pub name_record_sig: ST,
 }
```
michael-yxchen, line 33: what's the reason to augment old bootstrap peers instead of putting name records under peer discovery section then later retire bootstrap?

eerkaijun: bootstrap peers and peer discovery config options feels semantically different, so i think it's better to keep them as separate sections

michael-yxchen: sry i meant that you're changing bootstrap configs here to follow the MonadNameRecord format. I wonder if there's some benefit for keeping unsigned bootstrap peers, in case we want to manually override something. And have the MonadNameRecords under peer discovery, as those are used to initialize peer discovery

eerkaijun: i don't think there are use cases for unsigned bootstrap peers, if manually overriding we should still verify signature

here we can use `debug-node` to update bootstrap peers (with signature) during runtime

monad-types/src/lib.rs
michael-yxchen: ok the reasons above make sense

we should move the struct definition elsewhere. monad-types is reserved for basic types only

monad-updaters/src/config_loader.rs
```diff
@@ -155,37 +107,50 @@ where
...
+                    _ => return None,
+                };
+                Some(PeerEntry {
+                    pubkey: peer.secp256k1_pubkey,
+                    addr,
+                    signature: peer.name_record_sig,
+                    record_seq_num: peer.record_seq_num,
+                })
+            })
+            .collect();
```
michael-yxchen: this will again freeze up the main thread during dns lookup. this can take up to a few seconds. we can offload this to that separate task, but instead of periodically resolving, resolve until a record is available. There's some benefit of moving record verification to that task so main thread won't spend time verifying

monad-peer-discovery/src/mock.rs
```diff
@@ -160,14 +166,27 @@ where
         Vec::new()
     }
 
+    fn update_peers(&mut self, _peers: Vec<PeerEntry<ST>>) -> Vec<PeerDiscoveryCommand<ST>> {
+        debug!("updating peers");
+
+        Vec::new()
+    }
```
michael-yxchen: `NopDiscovery` supports peer resolution. should we implement this function then?

monad-peer-discovery/src/discovery.rs
```diff
@@ -121,7 +121,7 @@ impl<ST: CertificateSignatureRecoverable> PeerDiscoveryAlgoBuilder for PeerDisco
             .flat_map(|(node_id, _)| state.send_ping(node_id))
             .collect::<Vec<_>>();
 
-        cmds.extend(state.reset_refresh_timer());
+        cmds.extend(state.refresh());
```
michael-yxchen, line 125: what's the observation that led to this change?

monad-peer-discovery/src/discovery.rs
```diff
@@ -716,17 +716,59 @@ where
...
+
     fn metrics(&self) -> &PeerDiscMetrics {
         &self.metrics
     }
 
-    fn get_sock_addr_by_id(
-        &self,
-        id: &NodeId<CertificateSignaturePubKey<Self::SignatureType>>,
-    ) -> Option<SocketAddrV4> {
+    fn get_addr_by_id(&self, id: &NodeId<CertificateSignaturePubKey<ST>>) -> Option<SocketAddr> {
```
michael-yxchen: why this return value type change?

monad-peer-disc-swarm/tests/peer_discovery.rs
```diff
@@ -674,8 +675,8 @@ fn test_peer_lookup_retry() {
         if node == &node_b {
             let state = state.peer_disc_driver.get_peer_disc_state();
             let metrics = state.metrics();
-            assert_eq!(metrics["lookup_timeout"], 4);
-            assert_eq!(metrics["drop_lookup_response"], 4);
+            assert_eq!(metrics["lookup_timeout"], 8);
+            assert_eq!(metrics["drop_lookup_response"], 8);
```
michael-yxchen, line 679: why is the expected value doubled?

monad-peer-discovery/src/driver.rs
```diff
@@ -196,6 +198,28 @@ impl<PD: PeerDiscoveryAlgo> PeerDiscoveryDriver<PD> {
 
         emits
     }
+
+    pub fn get_addr(
+        &self,
+        node_id: &NodeId<CertificateSignaturePubKey<PD::SignatureType>>,
+    ) -> Option<SocketAddr> {
+        self.pd.get_addr_by_id(node_id)
```
michael-yxchen: i think peer discovery can operate with `SocketAddrV4` only. we can wrap it with `SocketAddr::V4` here

monad-updaters/src/local_router.rs
```diff
@@ -73,7 +76,7 @@ impl<PT: PubKey> LocalRouterConfig<PT> {
     }
 }
 
-pub struct LocalPeerRouter<M: Message, OM> {
+pub struct LocalPeerRouter<M: Message, ST, OM> {
```
michael-yxchen: nit: can we make `ST` the first generic arg?

monad-raptorcast/src/lib.rs
```diff
@@ -190,10 +185,11 @@ where
     OM: Serializable<Bytes> + Encodable + Into<M> + Clone,
     PD: PeerDiscoveryAlgo<SignatureType = ST>,
 {
-    type Command = RouterCommand<CertificateSignaturePubKey<ST>, OM>;
+    type Command = RouterCommand<ST, OM>;
 
     fn exec(&mut self, commands: Vec<Self::Command>) {
         let self_id = NodeId::new(self.key.pubkey());
+        let known_addresses = self.peer_discovery_driver.get_known_addresses();
```
michael-yxchen: i think we can modify `udp_build` to take in `&impl PeerDiscovery` so we don't make a copy of the addresses every time we exec. let's add a TODO now and fix separately

monad-peer-discovery/src/discovery.rs
eerkaijun: we should refresh immediately on startup so that the node can sends out peer lookup requests. previously it was waiting for the refresh period which causes unnecessary delays

monad-peer-disc-swarm/tests/peer_discovery.rs
eerkaijun: we now refresh at startup so the first refresh occurs at t=0

monad-types/src/lib.rs
eerkaijun: not sure where's a better place, `monad-executor-glue`?

monad-updaters/src/config_loader.rs
eerkaijun: updated

monad-peer-discovery/src/discovery.rs
eerkaijun: my thought is to keep `get_addr_by_id` and `get_known_addresses` to return the same type (so both should return `SocketAddr` or both should return `SocketAddrV4`). for `get_known_addresses` in `discovery.rs`, if we return `SocketAddrV4`, we will need to iterate the map again and wrap it in `SocketAddr::V4` in `driver.rs`. in my opinion i think keeping the interface generic with using `SocketAddr` is cleaner, but not opposed to changing as well

monad-types/src/lib.rs
michael-yxchen: yeah `monad-executor-glue` is fine for now

monad-updaters/src/local_router.rs
```diff
@@ -105,12 +108,13 @@ impl<M: Message, OM> LocalPeerRouter<M, OM> {
     }
 }
 
-impl<M, OM> Executor for LocalPeerRouter<M, OM>
+impl<M, ST, OM> Executor for LocalPeerRouter<ST, M, OM>
```
michael-yxchen: super nit: reorder generics `<M, ST, OM>` in this file

monad-peer-discovery/src/discovery.rs
michael-yxchen: peer discovery works purely with ipv4 addresses right now. let's use `SocketAddrV4` in all `PeerDiscoveryAlgo` interfaces and wrap it in driver

monad-raptorcast/src/lib.rs
```diff
@@ -190,10 +185,12 @@ where
     OM: Serializable<Bytes> + Encodable + Into<M> + Clone,
     PD: PeerDiscoveryAlgo<SignatureType = ST>,
 {
-    type Command = RouterCommand<CertificateSignaturePubKey<ST>, OM>;
+    type Command = RouterCommand<ST, OM>;
 
     fn exec(&mut self, commands: Vec<Self::Command>) {
         let self_id = NodeId::new(self.key.pubkey());
+        // TODO: perhaps pass this directly to udp_build to avoid calling on every exec
+        let known_addresses = self.peer_discovery_driver.get_known_addresses();
```
michael-yxchen: move it under `PublishCommand` variant. I think we rarely, if not never, have multiple publish commands in the same vector

monad-raptorcast/src/lib.rs
```diff
@@ -257,6 +251,9 @@ where
                         warn!(?elapsed, "long time to publish message")
                     });
 
+                    // TODO: perhaps pass this directly to udp_build to avoid calling on every exec
+                    let known_addresses = self.peer_discovery_driver.get_known_addresses();
```
fabiotdf, line 255: When eventually refactoring `udp::build_messages_with_length()` to avoid calling `get_know_address()` on every exec, would it be possible to pass a reference to the generic trait `PeerDiscoveryAlgo` or would we have to pass the concrete `PeerDiscoveryDriver<PD>` ?

eerkaijun: i _think_ it would have to be PeerDiscoveryDriver

PR #1833, opened 2025-05-30
monad-dataplane: tcp connection control channel and rx rate limiter

dshulyak:
ideally every protocol will have a rate limit for indidual type of messages, but that will be too invasive to implement right now.

in this change dataplane rate limits the number of tcp req/s with a burst allowance.
rate limit is applied individually for every connection.

the assumption is that dataplane api users are expected to disconnect/ban a peer that abuses
large/expensive messages, such as sending large block/statesync replies when they were not requested from that specific peer. or just generally sending invalid large blobs. so the worst behavior will be limited to small/valid messages, such as peer discovery requests.

### DIFF
--- a/monad-dataplane/src/tcp/mod.rs
+++ b/monad-dataplane/src/tcp/mod.rs
@@ -26,6 +26,7 @@ use monoio::{
     spawn,
 };
 use tokio::sync::mpsc::{self, UnboundedReceiver, UnboundedSender};
+use tracing::trace;
 use zerocopy::{
     byteorder::little_endian::{U32, U64},
     AsBytes, FromBytes,
@@ -159,17 +160,32 @@ impl TcpControl {
     #[allow(unused)]
     pub(crate) fn disconnect_ip(&self, ip: IpAddr) {
         let map = self.0.lock().unwrap();
-        for (_, tx) in map.range((ip, u16::MIN, u64::MIN)..(ip, u16::MAX, u64::MAX)) {
+        let mut count = 0;
+        for (id, tx) in map.range((ip, u16::MIN, u64::MIN)..(ip, u16::MAX, u64::MAX)) {
             let _ = tx.send(TcpControlMsg::Disconnect);
+            count += 1;
         }
+        trace!(
+            ?ip,
+            connections_disconnected = count,
+            "completed ip disconnect"
+        );
     }
 
     #[allow(unused)]
     pub(crate) fn disconnect_socket(&self, ip: IpAddr, port: u16) {
         let map = self.0.lock().unwrap();
+        let mut count = 0;
         for (_, tx) in map.range((ip, port, u64::MIN)..(ip, port, u64::MAX)) {
             let _ = tx.send(TcpControlMsg::Disconnect);
+            count += 1;
         }
+        trace!(
+            ?ip,
+            port,
+            connections_disconnected = count,
+            "completed socket disconnect"
+        );
     }
 }
 
--- a/monad-dataplane/src/tcp/rx.rs
+++ b/monad-dataplane/src/tcp/rx.rs
@@ -69,27 +69,57 @@ impl RxState {
 
     fn apply_limits(&self, ip: IpAddr) -> Result<ConnectionToken, ()> {
         let status = self.addrlist.status(&ip);
-        trace!(?ip, %status, "applying limits for incoming tcp connection");
         match status {
             Status::Banned => {
-                warn!(?ip, "address is banned, rejecting tcp connection");
+                warn!(?ip, "banned address attempting connection, dropping");
                 Err(())
             }
-            Status::Trusted => Ok(ConnectionToken::Trusted),
+            Status::Trusted => {
+                let inner_ref = self.inner.borrow();
+                trace!(
+                    ?ip,
+                    total_connections = inner_ref.num_connections,
+                    connection_limit = inner_ref.tcp_connections_limit,
+                    "trusted peer connection accepted"
+                );
+                Ok(ConnectionToken::Trusted)
+            }
             Status::Unknown => {
                 let mut inner_ref = self.inner.borrow_mut();
                 if inner_ref.num_connections >= inner_ref.tcp_connections_limit {
+                    debug!(
+                        ?ip,
+                        total_connections = inner_ref.num_connections,
+                        connection_limit = inner_ref.tcp_connections_limit,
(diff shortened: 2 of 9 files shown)


monad-dataplane/src/tcp/rx.rs
```diff
@@ -4,85 +4,133 @@ use std::{
...
-
-        if *count_ref < PER_PEER_CONNECTION_LIMIT {
-            *count_ref += 1;
-
-            Ok(())
-        } else {
-            Err(())
+    fn apply_limits(&self, ip: IpAddr) -> Result<(), ()> {
+        let status = self.addrlist.status(&ip);
+        trace!(?ip, %status, "accepting incoming TCP connection");
```
andr-dev: This trace message feels a bit misleading, mind rewording it?

monad-dataplane/src/tcp/rx.rs
```diff
@@ -4,85 +4,133 @@ use std::{
...
         let mut inner_ref = self.inner.borrow_mut();
+        inner_ref.num_connections -= 1;
 
-        let count_ref = inner_ref.num_connections.get_mut(ip).unwrap();
-
-        if *count_ref > 1 {
-            *count_ref -= 1;
-        } else {
-            inner_ref.num_connections.remove(ip);
+        if let Some(count_ref) = inner_ref.num_connections_per_ip.get_mut(ip) {
```
andr-dev: Can we
```rust
let Some(count_ref) = ... else {
  warn!(...);
  return;
}
```
If this ever happens, there's most likely a bug in the counting somewhere.

monad-dataplane/src/tcp/mod.rs
```diff
@@ -57,3 +62,221 @@ fn message_timeout(len: usize) -> Duration {
...
+        }
+    }
+
+    #[allow(unused)]
+    pub(crate) fn disconnect_socket(&self, ip: IpAddr, port: u16) {
+        let map = self.0.lock().unwrap();
+        for (_, tx) in map.range((ip, port, u64::MIN)..(ip, port, u64::MAX)) {
+            let _ = tx.send(TcpControlMsg::Disconnect);
+        }
+    }
```
andr-dev, line 189: What's the use case for being able individually disconnect a socket vs disconnecting an IP? Is this to allow banning certain types of activity from a given IP?

monad-dataplane/src/tcp/rx.rs
```diff
@@ -108,29 +156,54 @@ pub async fn task(local_addr: SocketAddr, tcp_ingress_tx: mpsc::Sender<(SocketAd
 }
 
 async fn task_connection(
+    tcp_rps: usize,
+    tcp_rps_burst: usize,
```
andr-dev: Can we put these into a struct? I can see a future where we also limit bytes or other stuff and as we keep adding fields it'll become a mess.

andr-dev: Ideally, that struct can provide an associated `into_rate_limiter()` that builds the `RateLimiter` below. If it's in a struct, we can also cleanly use `NonZeroU32` to avoid the `.try_into().unwrap()` right?

monad-dataplane/src/tcp/rx.rs
```diff
@@ -108,29 +156,54 @@ pub async fn task(local_addr: SocketAddr, tcp_ingress_tx: mpsc::Sender<(SocketAd
...
+            .allow_burst((tcp_rps_burst as u32).try_into().unwrap()),
+    );
+    loop {
+        select! {
+            biased;
+            ctl = control_rx.recv() => {
+                if let Some(TcpControlMsg::Disconnect) = ctl {
+                    trace!(conn_id, ?addr, "disconnecting tcp connection");
+                    break;
+                }
```
andr-dev: Let's make this a match so we show all possible variants. Also if `ctl` is `None`, should we throw a WARN/ERROR and disconnect to be safe?

monad-dataplane/src/tcp/rx.rs
```diff
@@ -108,29 +156,54 @@ pub async fn task(local_addr: SocketAddr, tcp_ingress_tx: mpsc::Sender<(SocketAd
...
+            },
+            msg = read_message(conn_id, addr, message_id, &mut tcp_stream) => {
+                if rate_limiter.check().is_err() {
+                    warn!(conn_id, ?addr, "rate limit exceeded");
+                    break;
+                }
+                if let Some(message) = msg {
```
andr-dev: Let's unroll this guy to avoid the indenting.
```rust
let Some(message) = msg else {
  break;
}
```

monad-dataplane/src/tcp/rx.rs
andr-dev: Looking at 1834 this is even more valuable since `DataplaneBuilder` also uses these fields.

monad-dataplane/src/tcp/mod.rs
dshulyak: yes, generally if ip is banned every socket should be disconnected.  and as for choosing ban/disconnect i think it is better to reserve ban for clearly disruptive activity (e.g repeated misbehavior with expensive requests)

monad-dataplane/src/tcp/mod.rs
```diff
@@ -57,3 +63,241 @@ fn message_timeout(len: usize) -> Duration {
...
+pub(crate) enum TcpControlMsg {
+    Disconnect,
+}
+
+pub(crate) type TcpIdentifier = (IpAddr, u16, u64);
+pub(crate) type TcpControlSender = UnboundedSender<TcpControlMsg>;
+pub(crate) type TcpControlReceiver = UnboundedReceiver<TcpControlMsg>;
+
+#[derive(Debug, Clone)]
+pub(crate) struct TcpControl(Arc<Mutex<BTreeMap<TcpIdentifier, TcpControlSender>>>);
```
andr-dev, line 136: Dataplane runs in a monoio instance right? In that case, shouldn't we use async Mutex?

dshulyak: i think we shouldn't, it would be a concern if we would hold mutex for long time (e.g tenth of milliseconds), or across await points.
but in this case it is just small amount of data, and very likely any operation will be in micro-seconds range

there is this doc in tokio https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use

PR #1836, opened 2025-06-02
Support dns resolution on startup

eerkaijun:
nice to have (allow both raw ip and dns) - also supported by ethereum clients like geth and reth

### DIFF
--- a/monad-node-config/src/peers.rs
+++ b/monad-node-config/src/peers.rs
@@ -1,5 +1,3 @@
-use std::net::SocketAddrV4;
-
 use monad_crypto::certificate_signature::{
     CertificateSignaturePubKey, CertificateSignatureRecoverable,
 };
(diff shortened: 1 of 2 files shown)


monad-node/src/main.rs
```diff
@@ -549,8 +549,18 @@ where
...
+    let self_address = match resolve_domain_v4(
+        &NodeId::new(identity.pubkey()),
+        &peer_discovery_config.self_address,
+    ) {
+        Some(SocketAddr::V4(addr)) => addr,
+        _ => panic!(
+            "Unable to resolve self address: {:?}",
+            peer_discovery_config.self_address
+        ),
+    };
```
michael-yxchen: nit: similar below
```suggestion
    let Some(SocketAddr::V4(self_address)) = resolve_domain_v4(
        &NodeId::new(identity.pubkey()),
        &peer_discovery_config.self_address,
    ) else {
        panic!(
            "Unable to resolve self address: {:?}",
            peer_discovery_config.self_address
        );
    };
```

monad-node-config/src/peers.rs
```diff
@@ -34,7 +32,7 @@ pub struct PeerConfig<ST: CertificateSignatureRecoverable> {
 #[derive(Debug, Deserialize, Serialize, Clone)]
 #[serde(deny_unknown_fields)]
 pub struct PeerDiscoveryConfig<ST: CertificateSignatureRecoverable> {
-    pub address: SocketAddrV4,
+    pub address: String,
```
fabiotdf, line 35: I see we use `SocketAddrV4` but we also have `SocketAddr`. Do we only support ipv4?

eerkaijun: yeap, currently only ipv4 is supported

PR #1837, opened 2025-06-02
Fix eth_estimateGas discrepancy

philipglazman:
Ottersec found two discprenancies in our gas handler.

See this [issue](https://github.com/category-labs/category-internal/issues/1387) for details

### DIFF
--- a/monad-rpc/src/handlers/eth/gas.rs
+++ b/monad-rpc/src/handlers/eth/gas.rs
@@ -23,6 +23,9 @@ use crate::{
     jsonrpc::{JsonRpcError, JsonRpcResult},
 };
 
+/// Additional gas added during a CALL.
+const CALL_STIPEND: u64 = 2_300;
+
 trait EthCallProvider {
     async fn eth_call(
         &self,
@@ -132,7 +135,8 @@ async fn estimate_gas<T: EthCallProvider>(
     };
 
     let upper_bound_gas_limit = txn.gas_limit();
-    call_request.gas = Some(U256::from((gas_used + gas_refund) * 64 / 63));
+    // Set gas to used + refund + call stipend and apply the 63/64 rule
+    call_request.gas = Some(U256::from((gas_used + gas_refund + CALL_STIPEND) * 64 / 63));
     txn = call_request.clone().try_into()?;
 
     let (mut lower_bound_gas_limit, mut upper_bound_gas_limit) =
@@ -155,7 +159,7 @@ async fn estimate_gas<T: EthCallProvider>(
                 }
             }
         } else {
-            (gas_used.sub(1), txn.gas_limit())
+            (gas_used.sub(1), upper_bound_gas_limit)
         };
 
     // Binary search for the lowest gas limit.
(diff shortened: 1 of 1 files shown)


monad-rpc/src/handlers/eth/gas.rs
```diff
@@ -132,7 +132,8 @@ async fn estimate_gas<T: EthCallProvider>(
     };
 
     let upper_bound_gas_limit = txn.gas_limit();
-    call_request.gas = Some(U256::from((gas_used + gas_refund) * 64 / 63));
+    // Set gas to used + refund + call stipend and apply the 63/64 rule
+    call_request.gas = Some(U256::from((gas_used + gas_refund + 2300) * 64 / 63));
```
eerkaijun: use named constant for 2300?

philipglazman: done

monad-rpc/src/handlers/eth/gas.rs
```diff
@@ -23,6 +23,9 @@ use crate::{
     jsonrpc::{JsonRpcError, JsonRpcResult},
 };
 
+/// Additional gas added during a CALL.
+const CALL_STIPEND: u64 = 2_300;
```
ariqchowdhury, line 27: why does geth add this amount to ethcall? 
https://github.com/search?q=repo%3Aethereum%2Fgo-ethereum%20callStipend&type=code
geth 
Looks like its added during execution as well, we must be doing it in the vm as well. 
If there's a const in execution that represents this value, we should name this the same and comment where it comes from in our codebase.

philipglazman: In the evm, 2_300 gas is forwarded to a callee contract when the following conditions are met: the caller contract is using the CALL opcode, sending eth, and has not set an explicit gas amount. If we don't have the stipend, we will get `CallResult::Failure` which will end up returning a higher gas price for users.

I demonstrated this by calling `estimateGas` on contract like this in flexnet:

```
contract Test {
    function callWithoutGas(address payable to) external payable returns (bool success) {
        // This sends ETH with implicit 2300 gas
        (success, ) = to.call{value: msg.value}("");
        require(success, "internal call failed"); 
    }
}

contract NeedsGas {
    uint public x;

    receive() external payable {
        x = 1;
    }
}
```
With the additional stipend required for gas forwarding, we will successfully return.


Looking at our execution code, we do add 2300 to the calls like Geth inside evmone, but the const variable is under a different name.

ariqchowdhury: can we change the const name here to match execution?

philipglazman: I actually misread at first glance, evmone doesn't have a const type for the call stipend. It has one for MIN_CALLEE_GAS which is also 2,300 but is semantically different. 

https://github.com/ethereum/evmone/blob/fde8552eb909aa90059b431b2a52c1a2f93726aa/lib/evmone/instructions_calls.cpp#L163C34-L163C41

PR #1838, opened 2025-06-02
Refactor rpc to use chainstate 

philipglazman:
Introduces ChainState type that abstracts triedb and archive readers in order to make handler code more readable

### DIFF
--- a/monad-rpc/src/chainstate/mod.rs
+++ b/monad-rpc/src/chainstate/mod.rs
@@ -1,0 +1,583 @@
+use alloy_consensus::{Header as RlpHeader, Transaction as _};
+use alloy_primitives::{FixedBytes, U256};
+use alloy_rpc_types::{Block, BlockTransactions, Header, Transaction, TransactionReceipt};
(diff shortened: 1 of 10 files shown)


monad-rpc/src/chainstate/mod.rs
```diff
@@ -0,0 +1,585 @@
...
+                        block.header.hash_slow(),
+                        crate::eth_json_types::MonadTransactionReceipt,
+                    )
+                    .map_err(|_| Error::ResourceNotFound)?;
+                    return Ok(block_receipts);
+                }
+            }
+        }
+
+        todo!();
```
ariqchowdhury: should this return an error?

philipglazman: Yes 🤦

philipglazman:
> looks good to me, to make sure flexnet CI passes before merging, since that should cover most rpc conformance aspects

Ran locally and it works! Flexnet CI is having some issues

PR #1840, opened 2025-06-03
Fix outstanding lookup request

eerkaijun:
Found a small issue of wrongly recording receiver in `outstanding_lookup_requests`


monad-peer-discovery/src/discovery.rs
michael-yxchen, line 1: is there a test covering this code change?

eerkaijun: added in the unit test

michael-yxchen: if i understand it correctly, the bug is in `handle_peer_lookup_timeout` path but none of the unit tests seems to call that function and assert on receiver

eerkaijun: added

PR #1842, opened 2025-06-04
Tests for min max watermark

eerkaijun:
swarm tests for peer discovery watermark

### DIFF
--- a/monad-peer-disc-swarm/tests/peer_discovery.rs
+++ b/monad-peer-disc-swarm/tests/peer_discovery.rs
@@ -758,4 +758,182 @@ fn test_ping_timeout() {
     }
 }
 
-// TODO: swarm tests peer discovery watermark
+#[traced_test]
+#[test]
+fn test_min_watermark() {
+    // 4 nodes: NodeA, NodeB, NodeC, NodeD
+    // NodeA does not have any peer in the beginning
+    // NodeB, NodeC, and NodeD has NodeA as their peer in the beginning
+    let keys = create_keys::<SignatureType>(4);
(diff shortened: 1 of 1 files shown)


monad-peer-disc-swarm/tests/peer_discovery.rs
```diff
@@ -758,4 +758,180 @@ fn test_ping_timeout() {
...
+                    outbound_pipeline: vec![],
+                }
+            })
+            .collect(),
+        seed: 7,
+    };
+
+    let mut nodes = swarm_builder.build();
+
+    while nodes.step_until(Duration::from_secs(5)) {}
```
maxkozlovsky: Maybe add a comment where does 5 come from or define a constant for peer refresh timeout and derive this value from that constant.

monad-peer-disc-swarm/tests/peer_discovery.rs
```diff
@@ -758,4 +758,180 @@ fn test_ping_timeout() {
...
+        let metrics = state.metrics();
+
+        // NodeB, NodeC, and NodeD should send lookup request to NodeA and discover each other
+        for peer_id in all_peers.keys() {
+            if peer_id == &state.self_id {
+                continue;
+            }
+            assert!(state.peer_info.contains_key(peer_id));
+
+            if node_id == &node_a {
```
maxkozlovsky: this probably should be outside of inner for loop since it only depends on node_id

eerkaijun: good point, updated, thanks

PR #1844, opened 2025-06-04
Add support for testnet-2

tongzhi1998:


monad-chain-config/src/lib.rs
```diff
@@ -72,6 +73,11 @@ impl MonadChainConfig {
 
             info!("Using override devnet chain config");
             Ok(override_config)
+        } else if chain_id == MONAD_TESTNET2_CHAIN_ID {
+            if devnet_override.is_some() {
+                warn!("Ignoring chain config from file in testnet");
```
michael-yxchen, line 78: nit: testnet-2

PR #1845, opened 2025-06-04
Archive: Rechecker improvements and inspector cli

JoeHowarth:
Squashed commits:
    - Add rechecker-v2 worker that rechecks entire fault chunks from scratch
    - Implements --use-rechecker-v2 flag to enable full chunk revalidation instead of individual fault rechecking, with automatic backup of old fault data.
    - Add inspector mode to archive checker for fault analysis
    - Provides commands to list fault ranges, inspect individual blocks across replicas, and view block validation status with detailed output options.
    - Fix rechecker-v2 test failures and compilation warnings
    Update collect_fault_chunks calls to include new parameters and fix test assumptions
    - Optimize collect_fault_chunks with parallel fetching
    Use buffered stream with concurrency limit of 20 for better performance
    - Fix bug where cleared faults remain in S3
    Add tests to verify and fix store_checking_results to delete fault chunks when replicas no longer have faults
    - Update CLI help to make rechecker v2 the default
    Remove mentions of non-v2 rechecker from examples
    - Cleanup checker changes
    - Add --force-recheck option to rechecker v2 for rechecking all chunks regardless of fault status. This enables comprehensive revalidation of archive data after repairs or migrations.
    - Change rechecker default to run once and exit, add --worker flag for continuous mode. This makes rechecker more suitable for one-off validations and cron jobs.
    - Add MongoDB integration test for archive checker with mixed storage
      backends
    - Move TestMongoContainer helper to test_utils


monad-archive/src/bin/monad-archive-checker/checker.rs
```diff
@@ -349,37 +348,85 @@ fn find_consensus(
             }
         }
     } else {
-        panic!("No equivalence group found. This means we do not know which block has the correct data. Manual intervention required!");
+        // This should not happen with proper consensus, but log critical error if it does
```
mikety, line 351: using `max_by_key` to find largest_group can also work on the 50/50 split. Consider using a tighter model - i.e. mo re than 1 replica stores differently == error ?

Issue #1832, opened 2025-05-30
flaky test workers::bft_block_archiver::tests::test_bft_block_archive_worker


dshulyak:
failed in unrelated changes https://github.com/category-labs/monad-bft/actions/runs/15341685942/job/43168957402?pr=1829

```
 failures:

---- workers::bft_block_archiver::tests::test_bft_block_archive_worker stdout ----

thread 'workers::bft_block_archiver::tests::test_bft_block_archive_worker' panicked at monad-archive/src/workers/bft_block_archiver.rs:199:9:
assertion `left == right` failed
  left: []
 right: ["block1.body", "block1.header", "block2.body", "block2.header"]
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

JoeHowarth:
https://github.com/category-labs/monad-bft/pull/1849

PR #1849, opened 2025-06-05
Archive: Fix flaky bft block upload test

JoeHowarth:


JoeHowarth:
fixes https://github.com/category-labs/monad-bft/issues/1832

Issue #1850, opened 2025-06-06
`well_formed` panic when round is zero


eerkaijun:
Ottersec

It seems that if a [ProtocolMessage::Proposal](https://github.com/category-labs/monad-bft/blob/aa9ead24e60c7c65b35f1582b6150dbc278dce98/monad-consensus/src/messages/consensus_message.rs#L27) is received with block_header.round == 0, the node will panic due to an unchecked subtraction in the well_formed function.

eerkaijun:
@ariqchowdhury sorry i didn't see you self assigning before opening the pr, but feel free to close mine off if there's another preferred pr

PR #1852, opened 2025-06-06
Underflow check for round number

eerkaijun:
closes https://github.com/category-labs/monad-bft/issues/1850 

since round >= 1 for first block, proposal with round 0 can be safely dropped

### DIFF
--- a/monad-consensus/src/validation/message.rs
+++ b/monad-consensus/src/validation/message.rs
@@ -11,6 +11,11 @@ pub fn well_formed<SCT>(
     qc_round: Round,
     tc: &Option<TimeoutCertificate<SCT>>,
 ) -> Result<(), Error> {
+    // do not allow Round(0) to prevent underflow
+    if round == Round(0) {
+        return Err(Error::NotWellFormed);
+    }
+
     let prev_round = round - Round(1);
     let valid_qc_round = qc_round == prev_round;
 


monad-consensus/src/validation/message.rs
```diff
@@ -11,6 +11,11 @@ pub fn well_formed<SCT>(
     qc_round: Round,
     tc: &Option<TimeoutCertificate<SCT>>,
 ) -> Result<(), Error> {
+    // do not allow Round(0) to prevent underflow
+    if round == Round(0) {
```
michael-yxchen, line 15: can we use `monad-types::GENESIS_ROUND`?

eerkaijun: updated

monad-consensus/src/validation/message.rs
```diff
@@ -11,6 +11,11 @@ pub fn well_formed<SCT>(
     qc_round: Round,
     tc: &Option<TimeoutCertificate<SCT>>,
 ) -> Result<(), Error> {
+    // do not allow Round(0) to prevent underflow
+    if round == GENESIS_ROUND {
```
ariqchowdhury: should this be explicitly checking Round(0)? GENESIS_ROUND could change and then this code will still crash.

michael-yxchen: ok right. I think we should reject both Round(0) and anything <= GENESIS_ROUND

michael-yxchen: maybe this isn't the right place to check for `<= GENESIS_ROUND`, feel free to ignore

eerkaijun: switched back to Round(0)

Issue #1853, opened 2025-06-07
raptorcast: node panics when received message is shorter than expected slice length


eerkaijun:
Ottersec:

It seems that in the current parsing code of RaptorCast, it attempts to [slice](https://github.com/category-labs/monad-bft/blob/aa9ead24e60c7c65b35f1582b6150dbc278dce98/monad-raptorcast/src/lib.rs#L544-L546) incoming messages with a fixed length without validating the message's actual length. If a received message is shorter than the expected slice length, the node will panic. Attached is the POC that we made.

```
diff --git a/Cargo.lock b/Cargo.lock
index 8109379..2c159e6 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5000,6 +5000,7 @@ dependencies = [
  "monad-consensus-types",
  "monad-control-panel",
  "monad-crypto",
+ "monad-dataplane",
  "monad-eth-block-policy",
  "monad-eth-block-validator",
  "monad-eth-txpool-executor",
@@ -5031,6 +5032,7 @@ dependencies = [
  "tracing",
  "tracing-subscriber",
  "vergen-git2",
+ "zerocopy 0.6.5",
 ]
 
 [[package]]
diff --git a/monad-node/Cargo.toml b/monad-node/Cargo.toml
index 346f738..a996840 100644
--- a/monad-node/Cargo.toml
+++ b/monad-node/Cargo.toml
@@ -8,6 +8,10 @@ edition = "2021"
 name = "monad-node"
 bench = false
 
+[[bin]]
+name = "monad-node-test"
+path = "src/bin/monad-node-test.rs"
+
 [dependencies]
 monad-bls = { workspace = true }
 monad-chain-config = { workspace = true }
@@ -35,6 +39,8 @@ monad-types = { workspace = true }
 monad-updaters = { workspace = true, features = ["monad-triedb", "tokio"] }
 monad-validator = { workspace = true }
 monad-wal = { workspace = true }
+monad-dataplane = { workspace = true }
+zerocopy = { workspace = true }
 
 alloy-rlp = { workspace = true }
 bytes = { workspace = true }
diff --git a/monad-node/src/bin/monad-node-test.rs b/monad-node/src/bin/monad-node-test.rs
new file mode 100644
index 0000000..e7c0812
--- /dev/null
+++ b/monad-node/src/bin/monad-node-test.rs
@@ -0,0 +1,81 @@
+use std::error::Error;
+use tokio::net::TcpStream;
[...]

PR #1854, opened 2025-06-09
Message slicing check

eerkaijun:
closes https://github.com/category-labs/monad-bft/issues/1853
closes https://github.com/category-labs/monad-bft/issues/1908 
tested with their poc

### DIFF
--- a/monad-raptorcast/src/lib.rs
+++ b/monad-raptorcast/src/lib.rs
@@ -545,6 +545,14 @@ where
 
         while let Poll::Ready((from_addr, message)) = pin!(this.dataplane.tcp_read()).poll_unpin(cx)
         {
+            // check message length to prevent panic during message slicing
+            if message.len() < SIGNATURE_SIZE {
+                warn!(
+                    ?from_addr,
+                    "invalid message, message length less than signature size"
+                );
+                continue;
+            }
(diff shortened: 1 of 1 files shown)


monad-raptorcast/src/lib.rs
```diff
@@ -545,6 +545,14 @@ where
 
         while let Poll::Ready((from_addr, message)) = pin!(this.dataplane.tcp_read()).poll_unpin(cx)
         {
+            // check message length to prevent panic during message slicing
```
dshulyak, line 548: i think it makes sense to add this check in dataplane, to avoid even passing through channel

eerkaijun: initially was thinking that too but feels more error prone to keep the check and slicing in separate places (e.g. future refactor might not have context)

dshulyak: if it hardcoded then yes thats likely, but it can be be also passed in builder

andr-dev: Feels like the right place to make a custom type that encapsulates this logic and enforces it can only be created if this condition is met ... not sure what this is called, correctness through type creation?

eerkaijun: will leave it as it is for now

Issue #1411, opened 2025-01-10
RaptorCast Message broadcast forwarding with old timestamps can spam network


0xSqualo:
On the RaptorCast re-broadcasting of parsed messages (e.g. a validator sends a message to another validator to broadcast to all of the validators in the current epoch), currently RaptorCast only checks that the [signature is from a validator in the current epoch, and the current client processing the message was the expected receiver](https://github.com/category-labs/monad-bft/blob/d65b61b3e6db50d339bb7ed5f32709d5cc3d1fd6/monad-raptorcast/src/udp.rs#L129):

```rust
// pub fn handle_message(...)
// ...
        if parsed_message.broadcast {
                let Some(epoch_validators) = epoch_validators.get_mut(&Epoch(parsed_message.epoch))
                else {
                    tracing::debug!(
                        epoch =? parsed_message.epoch,
                        "don't have epoch validators populated",
                    );
                    continue;
                };
                if !epoch_validators
                    .validators
                    .contains_key(&parsed_message.author)
                {
                    tracing::debug!(
                        author =? parsed_message.author,
                        "not in validator set"
                    );
                    continue;
                }
                if self_hash == parsed_message.recipient_hash {
                    let targets =
                        epoch_validators.view_without(vec![&parsed_message.author, &self_id]);
                    batch_guard.queue_broadcast(
                        payload_start_idx,
                        payload_end_idx,
                        &parsed_message.author,
                        || targets.view().keys().copied().collect(),
                    )
                }

                // forward all broadcast packets to full nodes
                full_node_forward_batch_guard.queue_forward(payload_start_idx, payload_end_idx);
            } else if self_hash != parsed_message.recipient_hash {
                tracing::debug!(
                    ?self_hash,
                    recipient_hash =? parsed_message.recipient_hash,
                    "dropping spoofed message"
                );
                continue;
            }
// ...
```

The message parsing up until this point also doesn't [validate the timestamp or epoch in anyway](https://github.com/category-labs/monad-bft/blob/d65b61b3e6db50d339bb7ed5f32709d5cc3d1fd6/monad-raptorcast/src/udp.rs#L750).

The main issue with this, is messages with an extremely "old" (or maliciously intended) timestamp / epoch are still forward broadcasted to the current epoch's validator set, and then will be further forwarded on by each client, unnecessarily wasting the entire validator set's bandwidth.

[...]

PR #1856, opened 2025-06-10
monad-raptorcast: validate udp message timestamp against configured max_age_ms

dshulyak:
closes: https://github.com/category-labs/monad-bft/issues/1411

max_age_ms defaults to 10s

### DIFF
--- a/monad-node-config/src/network.rs
+++ b/monad-node-config/src/network.rs
@@ -16,6 +16,9 @@ pub struct NodeNetworkConfig {
 
     #[serde(default = "default_mtu")]
     pub mtu: u16,
+
+    #[serde(default = "default_udp_message_max_age_ms")]
+    pub udp_message_max_age_ms: u64,
 }
 
 // When running in docker with vpnkit, the maximum safe MTU is 1480, as per:
@@ -27,4 +30,8 @@ fn default_mtu() -> u16 {
 fn default_buffer_size() -> Option<usize> {
     // recommended value at the time of the commit
     Some(62_500_000)
+}
+
+fn default_udp_message_max_age_ms() -> u64 {
+    10_000 // 10 seconds in milliseconds
 }
--- a/monad-raptorcast/src/lib.rs
+++ b/monad-raptorcast/src/lib.rs
@@ -135,7 +135,7 @@ where
 
             current_epoch: Epoch(0),
 
-            udp_state: udp::UdpState::new(self_id),
+            udp_state: udp::UdpState::new(self_id, config.udp_message_max_age_ms),
             mtu: config.mtu,
 
             dataplane,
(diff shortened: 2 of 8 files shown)


monad-raptorcast/src/lib.rs
```diff
@@ -140,7 +141,10 @@ where
 
             current_epoch: Epoch(0),
 
-            udp_state: udp::UdpState::new(self_id),
+            udp_state: udp::UdpState::new(
+                self_id,
+                config.message_max_age_secs.saturating_mul(1000),
```
andr-dev: Why not just configure in `ms`? I could see a future where we set this significantly lower than 60 seconds, say 800ms.

monad-raptorcast/src/udp.rs
```diff
@@ -1081,6 +1096,25 @@ where
     })
 }
 
+fn ensure_valid_timestamp(unix_ts_ms: u64, max_age_ms: u64) -> Result<(), MessageValidationError> {
+    let current_time_ms = if let Ok(current_time_elapsed) = std::time::UNIX_EPOCH.elapsed() {
+        current_time_elapsed.as_millis() as u64
+    } else {
+        warn!("system time is before unix epoch, ignoring timestamp");
```
andr-dev, line 1174: lol if this ever happens we should probably log at `error` level

monad-raptorcast/src/lib.rs
dshulyak: it might be problamatic to set lower than 1s, due to risks of running into clock synchronization problems. but yeah it can be expressed in ms to avoid conversion

monad-raptorcast/src/udp.rs
dshulyak: doesnt make much difference, i just didn't want to unwrap

monad-node-config/src/network.rs
```diff
@@ -28,3 +31,7 @@ fn default_buffer_size() -> Option<usize> {
     // recommended value at the time of the commit
     Some(62_500_000)
 }
+
+fn default_message_max_age_ms() -> u64 {
+    60_000 // 60 seconds in milliseconds
```
omegablitz: I think this can be set a lot lower; maybe 5s? The round timeout is ~1s.

monad-node-config/src/network.rs
```diff
@@ -16,6 +16,9 @@ pub struct NodeNetworkConfig {
 
     #[serde(default = "default_mtu")]
     pub mtu: u16,
+
+    #[serde(default = "default_message_max_age_ms")]
+    pub message_max_age_ms: u64,
```
omegablitz: This is only for raptorcast'd messages over UDP, correct?

Idk if it's worth incorporating that into the name somehow to be clear this doesn't apply to TCP messages.
Probably doesn't matter too much because most people won't override this.

dshulyak: yes correct, i will update

dshulyak: prefixed with udp

monad-node-config/src/network.rs
dshulyak: i will try 10s

PR #1860, opened 2025-06-10
Add metric to identify validator nodes

andr-dev:


michael-yxchen:
cross-posting discussions offline

the places to "call" `update_metrics` are
1) in `MonadStateBuilder.build`, so it's populated correctly on init
2) when handling proposal, maybe `ConsensusChildState.update` [here](https://github.com/category-labs/monad-bft/blob/31c88925ace51f605e8f6268b3a12cb6079215ff/monad-state/src/consensus.rs#L215-L216) so we don't insert those logic into core consensus

edit: `ConsensusChildState::new()` covers both cases

monad-state/src/consensus.rs
```diff
@@ -534,6 +538,30 @@ where
             .map(|(_, round)| round)
             .collect()
     }
+
+    fn update_metrics(&mut self) {
+        let ConsensusMode::Live(_) = self.consensus else {
+            return;
+        };
```
michael-yxchen: i think we don't need this check. update metrics during syncing is valid

monad-state/src/consensus.rs
```diff
@@ -534,6 +538,30 @@ where
...
+
+        let Some(validator_set) = self
+            .val_epoch_map
+            .get_val_set(&self.consensus.current_epoch())
+        else {
+            todo!("handle non-existent validatorset for current epoch");
+        };
```
michael-yxchen: log instead of panic here, because we should allow node to keep statesyncing even if they don't have the latest validator set

michael-yxchen (dismissed):

PR #1862, opened 2025-06-11
Targeted validator lookup request

eerkaijun:
closes https://github.com/category-labs/category-internal/issues/1643

a node could fall into a scenario where the peer list is larger than the minimum active connections, so the node is no longer actively looking for new peers. in this scenario, we should try to send target peer lookup request to fill up the remaining missing validators

### DIFF
--- a/monad-peer-disc-swarm/tests/peer_discovery.rs
+++ b/monad-peer-disc-swarm/tests/peer_discovery.rs
@@ -23,8 +23,9 @@ use monad_peer_discovery::{
         GAUGE_PEER_DISC_DROP_LOOKUP_RESPONSE, GAUGE_PEER_DISC_DROP_PONG,
         GAUGE_PEER_DISC_LOOKUP_TIMEOUT, GAUGE_PEER_DISC_PING_TIMEOUT,
         GAUGE_PEER_DISC_RECV_LOOKUP_REQUEST, GAUGE_PEER_DISC_RECV_PING, GAUGE_PEER_DISC_RECV_PONG,
-        GAUGE_PEER_DISC_REFRESH, GAUGE_PEER_DISC_SEND_LOOKUP_REQUEST, GAUGE_PEER_DISC_SEND_PING,
-        GAUGE_PEER_DISC_SEND_PONG, PeerDiscovery, PeerDiscoveryBuilder, PeerInfo,
+        GAUGE_PEER_DISC_RECV_TARGETED_LOOKUP_REQUEST, GAUGE_PEER_DISC_REFRESH,
+        GAUGE_PEER_DISC_SEND_LOOKUP_REQUEST, GAUGE_PEER_DISC_SEND_PING, GAUGE_PEER_DISC_SEND_PONG,
+        PeerDiscovery, PeerDiscoveryBuilder, PeerInfo,
     },
 };
 use monad_router_scheduler::{NoSerRouterConfig, NoSerRouterScheduler, RouterSchedulerBuilder};
@@ -585,16 +586,30 @@ fn test_peer_lookup_random_nodes() {
 
 #[traced_test]
 #[test]
-fn test_peer_lookup_retry() {
+fn test_peer_lookup_targeted_nodes() {
     let keys = create_keys::<SignatureType>(3);
-    let node_a_key = &keys[0];
-    let node_a = NodeId::new(node_a_key.pubkey());
-    let node_b_key = &keys[1];
-    let node_b = NodeId::new(node_b_key.pubkey());
-    let node_c_key = &keys[2];
-    let node_c = NodeId::new(node_c_key.pubkey());
-    let (subset_keys, _) = (&keys[0..2], &keys[2]);
-    let all_peers: BTreeMap<NodeId<PubKeyType>, PeerInfo<SignatureType>> = subset_keys
+
+    let subset_keys = &keys[0..2];
+
+    let node_a = NodeId::new(keys[0].pubkey());
+    let node_b = NodeId::new(keys[1].pubkey());
+    let node_c = NodeId::new(keys[2].pubkey());
+
+    // initialize peer info
+    // NodeA name record: NodeB, NodeC
+    // NodeB name record: NodeA
+    // All three nodes are validators, NodeB is missing NodeC name record
+    let all_peers: BTreeMap<NodeId<PubKeyType>, PeerInfo<SignatureType>> = keys
+        .iter()
+        .map(|k| {
+            (NodeId::new(k.pubkey()), PeerInfo {
+                last_ping: None,
+                unresponsive_pings: 0,
+                name_record: generate_name_record(k),
+            })
+        })
+        .collect();
+    let node_b_peers: BTreeMap<NodeId<PubKeyType>, PeerInfo<SignatureType>> = keys[0..2]
         .iter()
         .map(|k| {
             (NodeId::new(k.pubkey()), PeerInfo {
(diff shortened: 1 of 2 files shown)


monad-peer-discovery/src/discovery.rs
```diff
@@ -678,6 +687,35 @@ where
...
+                        .peer_info
+                        .keys()
+                        .cloned()
+                        .choose_multiple(&mut self.rng, NUM_LOOKUP_PEERS);
+
+                    for peer in chosen_peers {
+                        cmds.extend(self.send_peer_lookup_request(peer, validator, false));
+                    }
+                }
+            }
```
michael-yxchen: i think we should prioritize validators lookup, regardless if we've reached min_active_connections or not

eerkaijun: currently we are already prioritizing validators lookup (server are already sending validators in lookup response). this pr is mainly for logic to decide when to do open discovery vs targeted discovery. i think it's better to do open discovery when number of peers are low, so peer list grows fast. so e.g. if there are 300 nodes in the network, initially we will do open discovery which each request results in 16 validators in the response (there could be overlap from different peers but i think that's fine), rather than sending multiple individual targeted requests

michael-yxchen: prioritizing validators in responses is a server choice and not part of the protocol. currently in open discovery, we're always setting self_id as target. when server choose not to honor open discovery bit, e.g. if constrained by resources, we end up worse off than targeted lookups

my point is client should use an unknown validator id as target during open discovery. it doesn't need to send multiple targeted requests

eerkaijun: ok i agree that we should switch self id to an unknown validator id

monad-peer-discovery/src/discovery.rs
```diff
@@ -452,6 +452,10 @@ where
 
         // if open discovery, return more nodes other than requested nodes
         if request.open_discovery {
+            *self
+                .metrics
+                .entry("handle_open_lookup_request")
```
andr-dev: Let's call this "lookup_request_open" and the next one "lookup_request_targeted".

monad-peer-discovery/src/discovery.rs
```diff
@@ -667,17 +676,48 @@ where
...
+                        .keys()
+                        .cloned()
+                        .choose_multiple(&mut self.rng, NUM_LOOKUP_PEERS);
 
-            for peer in chosen_peers {
-                cmds.extend(self.send_peer_lookup_request(peer, self.self_id, true));
+                    for peer in chosen_peers {
+                        cmds.extend(self.send_peer_lookup_request(peer, validator, false));
+                    }
+                }
```
andr-dev: It seems like this entire chunk will only get triggered if we receive a peer lookup request, why not run this periodically?

monad-peer-discovery/src/discovery.rs
```diff
@@ -667,17 +676,48 @@ where
...
+                    .peer_info
+                    .keys()
+                    .cloned()
+                    .choose_multiple(&mut self.rng, NUM_LOOKUP_PEERS);
+                debug!(?chosen_peers, "discover more peers");
+
+                for peer in chosen_peers {
+                    if let Some(validator_id) = missing_validators.iter().choose(&mut self.rng) {
+                        cmds.extend(self.send_peer_lookup_request(peer, *validator_id, true));
+                    }
```
andr-dev: Let's add a comment here explaining why we're using a random `validator_id`. This was unclear to me until I saw the discussion from earlier.

monad-peer-discovery/src/discovery.rs
eerkaijun: this `refresh` function is already run periodically

monad-peer-discovery/src/discovery.rs
```diff
@@ -697,7 +712,21 @@ where
...
+            // if already above the min active peers, send targeted peer lookup request for each missing validator
+            debug!(
+                ?missing_validators,
+                "sending targeted peer lookup for missing validators"
+            );
+
+            for validator in missing_validators {
+                if let Some(chosen_peer) = self.peer_info.keys().choose(&mut self.rng) {
+                    cmds.extend(self.send_peer_lookup_request(*chosen_peer, validator, false));
+                }
```
michael-yxchen: i saw comment around `NUM_LOOKUP_PEERS` says that it's used when below min peers. does it make sense to extend it to the other case here so we gradually discover remaining validators?

monad-peer-disc-swarm/tests/peer_discovery.rs
```diff
@@ -583,6 +584,104 @@ fn test_peer_lookup_random_nodes() {
...
+        let state = state.peer_disc_driver.get_peer_disc_state();
+        let metrics = state.metrics();
+        for node_id in all_peers.keys() {
+            if node_id == &state.self_id {
+                continue;
+            }
+            assert!(state.peer_info.contains_key(node_id));
+        }
+
+        if state.self_id == node_a {
```
michael-yxchen, line 685: is this because node_a has the "smallest" pubkey of all?

eerkaijun: i realize the test is a bit outdated so i updated it now. now it will send deterministically send lookup request to node a because that's the only other node in its peer list

monad-peer-discovery/src/discovery.rs
eerkaijun: sure. i think we should recommend operators to set min active connections to be something close to the number of validators, otherwise lookup will be very slow

michael-yxchen: agreed, can we compile config recommendations somewhere in notion?

monad-peer-discovery/src/discovery.rs
```diff
@@ -690,17 +697,41 @@ where
...
+            // if already above the min active peers, send targeted peer lookup request for missing validators
+            for peer in chosen_peers {
+                if let Some(validator_id) = missing_validators.iter().choose(&mut self.rng) {
+                    debug!(
+                        ?validator_id,
+                        "sending targeted peer lookup for missing validator"
+                    );
+                    cmds.extend(self.send_peer_lookup_request(peer, *validator_id, false));
+                }
             }
```
michael-yxchen: we should send `min(NUM_LOOKUP_PEERS, missing_validators.len(), peer_info.len())` lookup requests. we should make sure that we don't look up the same validator more than once, or send more than one requests to a peer

michael-yxchen: this also implies that we will query offline validators' records every `refresh_period`, maybe that's fine in production

eerkaijun: > query offline validators' records

yeah i guess if the validator has never started up and its name record is not in any of the other nodes' peer list

monad-peer-discovery/src/discovery.rs
```diff
@@ -690,17 +697,40 @@ where
...
+            .into_iter()
+            .flatten()
+            .chain(next_epoch_validators.into_iter().flatten())
+            .filter(|validator| {
+                !self.peer_info.contains_key(validator) && *validator != &self.self_id
+            })
+            .cloned()
+            .collect::<HashSet<_>>()
+            .iter()
+            .copied()
```
```suggestion
            .into_iter()
```

monad-peer-discovery/src/discovery.rs
michael-yxchen: this will be resolved when we introduce connection list to peer discovery

PR #1864, opened 2025-06-11
check execution results as part of coherency

omegablitz:


monad-consensus-state/src/lib.rs
```diff
@@ -4274,11 +4063,10 @@ mod test {
     #[test]
     fn test_blocksync_requests() {
         let num_states = 2;
-        let execution_delay = SeqNum(4);
+        let execution_delay = SeqNum::MAX;
```
andr-dev, line 4068: Why do we need to set this to max now?

monad-triedb-utils/src/lib.rs
```diff
@@ -321,6 +321,56 @@ impl StateBackend for TriedbReader {
...
+                .raw_read_latest_finalized_block()
+                .is_some_and(|latest_finalized| seq_num <= &latest_finalized)
+        {
+            trace!(?seq_num, "triedb read eth header finalized");
+            // check finalized
+
+            // block <= latest
+            let Some(header) = self.get_finalized_eth_header(seq_num) else {
+                // TODO: Use a more descriptive error
+                return Err(StateBackendError::NotAvailableYet);
```
andr-dev, line 342: Should we add a warn/error trace here?

monad-consensus-state/src/lib.rs
omegablitz: We don't need to (it passes without changing), but there's no reason for it not to be SeqNum::MAX.
I changed it because I think it's clearer for only the tests dependent on execution_delay to not use SeqNum::MAX. Idk why it wasn't set to SeqNum::MAX before.

monad-triedb-utils/src/lib.rs
omegablitz: This branch almost certainly won't get hit, the same codepath also exists for TriedbReader::get_account_statuses (this was copied from there), and that's never been hit.

But I can add an error trace anyways, can't hurt

monad-consensus-state/src/lib.rs
andr-dev: Yep that makes sense, I was originally under the impression the change was necessary for the tests to pass now and was trying to backtrack why.

PR #1867, opened 2025-06-12
Add event server to RPC

andr-dev:


monad-rpc/src/event/server.rs
```diff
@@ -0,0 +1,270 @@
...
+                EventNextResult::Ready(event_descriptor) => event_descriptor,
+            };
+
+            let Some(result) = block_builder.process_event_descriptor(&event_descriptor) else {
+                continue;
+            };
+
+            match result {
+                BlockBuilderResult::Rejected => {
+                    unimplemented!();
```
philipglazman, line 72: why should we panic here?

andr-dev: `Rejected` only occurs if the block is invalid and should only occur when, for example, replaying ethereum history. We _could_ reset the block builder like we do in the `PayloadExpired` case but one or both of BFT/execution would crash soon after regardless.

PR #1868, opened 2025-06-12
Implement websocket subscriptions in RPC

philipglazman:


monad-rpc/src/cli.rs
```diff
@@ -150,4 +158,8 @@ pub struct Cli {
         default_value = ""
     )]
     pub pprof: String,
+
+    /// Sets the socket path for the monad execution event server
+    #[arg(long)]
+    pub exec_event_path: PathBuf,
```
ariqchowdhury: double checking, we are always expecting the execution event stuff to be running?

monad-rpc/src/main.rs
```diff
@@ -264,6 +267,31 @@ async fn main() -> std::io::Result<()> {
...
+            HttpServer::new(move || {
+                App::new()
+                    .app_data(web::Data::new(ws_server_handle.clone()))
+                    .service(
+                        web::resource("/").route(web::get().to(websocket::handler::ws_handler)),
+                    )
+            })
+            .bind((args.rpc_addr.clone(), args.ws_port))?
+            .shutdown_timeout(1)
+            .workers(4),
```
ariqchowdhury: so the websocket server will have an additional 4 threads?

monad-rpc/src/cli.rs
kjcamann: It's possible the execution daemon can be running without it.

It's only there if a command line parameter `--exec-event-ring=<filename>` is present on the `cmd/monad` command line.

You should probably tolerate it not being there, in which case

- WebSockets won't work (it used to be optional, i.e., you needed `--ws-enabled`, not sure about now)
- You'd need to get all info from TrieDB

Or alternatively you could just die and say "if you want to run RPC, you need it."

andr-dev: `exec_event_path` is only used if `ws_enabled` is passed so we should make this optional, and I think it's totally reasonable to die and say you need `exec_event_path` for websockets. The alternative would be to support normal RPC requests over websockets without subscriptions which feels wrong.

philipglazman: I set `exec_event_path` to optional but require it when the operator sets `--ws-enabled` since it would not make since to run a websocket server without the event system.

monad-rpc/src/main.rs
philipglazman: I halved it to 2 which should be plenty

monad-rpc/src/main.rs
```diff
@@ -264,6 +267,31 @@ async fn main() -> std::io::Result<()> {
...
+                    .service(
+                        web::resource("/").route(web::get().to(websocket::handler::ws_handler)),
+                    )
+            })
+            .bind((args.rpc_addr.clone(), args.ws_port))?
+            .shutdown_timeout(1)
+            .workers(2),
+        )
+    } else {
+        None
```
andr-dev, line 292: I think we might want to do this the other way around and first check if `exec_event_path` is set. Later when we add the buffer, we'll want to start the thread that populates the buffer regardless of whether ws_enabled is set. Something along the lines of
```rust
let (event_ring, ws_server_handle, events_for_cache) = if let Some(exec_event_path) = args.exec_event_path {
  // do ring and buffer setup
  
  let ws_server_handle = args.ws_enabled.then(|| {
    // do ws setup
  });
  
  (Some(event_ring), ws_server_handle, Some(events_for_cache))
} else {
  if args.ws_enabled {
    panic!("exec-event-path is not set but is required for websockets");
  }
  
  (None, None, None)
}
```

monad-rpc/src/main.rs
ariqchowdhury: thinking about it more, can we make it configurable?

philipglazman: Fair, made it configurable

PR #1870, opened 2025-06-13
Full-node Raptorcast: Integration

fabiotdf:
This activates full-node Raptorcast, integrates it with Peer discovery and primary Raptorcast instance.
Tested on flexnet, with associated PR on `monad-integration`

### DIFF
--- a/monad-node-config/src/fullnode_raptorcast.rs
+++ b/monad-node-config/src/fullnode_raptorcast.rs
@@ -9,7 +9,7 @@ use super::fullnode::FullNodeConfig;
 pub enum SecondaryRaptorCastModeConfig {
     Client,
     Publisher,
-    // No "None" option, as FullNodeRaptorCastConfig is now wrapped in Option<>
+    None, // Disables secondary raptorcast
 }
 
 #[derive(Debug, Deserialize, Serialize, Clone)]
@@ -34,4 +34,5 @@ pub struct FullNodeRaptorCastConfig<P: PubKey> {
     pub bandwidth_capacity: u64,
     pub invite_future_dist_min: Round,
     pub invite_future_dist_max: Round,
+    pub invite_accept_heartbeat_ms: u64,
 }
--- a/monad-node/src/main.rs
+++ b/monad-node/src/main.rs
@@ -1,7 +1,9 @@
 use std::{
     collections::HashMap,
     marker::PhantomData,
-    net::{SocketAddr, SocketAddrV4, ToSocketAddrs},
+    net::{SocketAddr, ToSocketAddrs},
+    process,
+    sync::Arc,
     time::Duration,
 };
 
@@ -20,6 +22,7 @@ use monad_control_panel::ipc::ControlPanelIpcReceiver;
 use monad_crypto::certificate_signature::{
     CertificateKeyPair, CertificateSignaturePubKey, CertificateSignatureRecoverable, PubKey,
 };
+use monad_dataplane::DataplaneBuilder;
 use monad_eth_block_policy::EthBlockPolicy;
 use monad_eth_block_validator::EthValidator;
 use monad_eth_txpool_executor::{EthTxPoolExecutor, EthTxPoolIpcConfig};
@@ -27,7 +30,7 @@ use monad_executor::{Executor, ExecutorMetricsChain};
 use monad_executor_glue::{LogFriendlyMonadEvent, Message, MonadEvent};
 use monad_ledger::MonadBlockFileLedger;
 use monad_node_config::{
-    ExecutionProtocolType, FullNodeIdentityConfig, NodeBootstrapConfig, NodeNetworkConfig,
+    ExecutionProtocolType, FullNodeIdentityConfig, NodeBootstrapConfig, NodeConfig,
     PeerDiscoveryConfig, SignatureCollectionType, SignatureType,
 };
 use monad_peer_discovery::{
@@ -35,7 +38,12 @@ use monad_peer_discovery::{
     MonadNameRecord, NameRecord,
 };
 use monad_pprof::start_pprof_server;
-use monad_raptorcast::{RaptorCast, RaptorCastConfig};
+use monad_raptorcast::config::{
+    GroupSchedulingConfig, RaptorCastConfig, RaptorCastConfigPrimary,
+    RaptorCastConfigSecondaryClient, RaptorCastConfigSecondaryPublisher,
+    SecondaryRaptorCastModeConfig,
+};
+use monad_router_multi::MultiRouter;
 use monad_state::{MonadMessage, MonadStateBuilder, VerifiedMonadMessage};
 use monad_state_backend::StateBackendThreadClient;
 use monad_statesync::StateSync;
@@ -54,7 +62,7 @@ use monad_wal::{wal::WALoggerConfig, PersistenceLoggerBuilder};
 use opentelemetry::metrics::MeterProvider;
 use opentelemetry_otlp::{MetricExporter, WithExportConfig};
 use tokio::signal::unix::{signal, SignalKind};
-use tracing::{error, event, info, warn, Instrument, Level};
+use tracing::{debug, error, event, info, warn, Instrument, Level};
 use tracing_subscriber::{
     fmt::{format::FmtSpan, Layer as FmtLayer},
     layer::SubscriberExt,
@@ -164,7 +172,7 @@ async fn run(node_state: NodeState, reload_handle: ReloadHandle) -> Result<(), (
             MonadMessage<SignatureType, SignatureCollectionType, ExecutionProtocolType>,
             VerifiedMonadMessage<SignatureType, SignatureCollectionType, ExecutionProtocolType>,
         >(
-            node_state.node_config.network.clone(),
+            node_state.node_config.clone(),
             node_state.node_config.peer_discovery,
             node_state.router_identity,
             node_state.node_config.bootstrap.clone(),
@@ -483,14 +491,14 @@ async fn run(node_state: NodeState, reload_handle: ReloadHandle) -> Result<(), (
 }
 
 async fn build_raptorcast_router<ST, SCT, M, OM>(
-    network_config: NodeNetworkConfig,
+    node_config: NodeConfig<ST>,
     peer_discovery_config: PeerDiscoveryConfig<ST>,
     identity: ST::KeyPairType,
     bootstrap_nodes: NodeBootstrapConfig<ST>,
     full_nodes: &[FullNodeIdentityConfig<CertificateSignaturePubKey<ST>>],
     locked_epoch_validators: Vec<ValidatorSetDataWithEpoch<SCT>>,
     current_epoch: Epoch,
-) -> RaptorCast<ST, M, OM, MonadEvent<ST, SCT, ExecutionProtocolType>, PeerDiscovery<ST>>
+) -> MultiRouter<ST, M, OM, MonadEvent<ST, SCT, ExecutionProtocolType>, PeerDiscovery<ST>>
 where
     ST: CertificateSignatureRecoverable,
     SCT: SignatureCollection<NodeIdPubKey = CertificateSignaturePubKey<ST>>,
@@ -511,6 +519,28 @@ where
             peer_discovery_config.self_address
         );
     };
+
+    tracing::debug!(
+        "Monad-node ({}) starting at address: {}, pid: {}",
+        if cfg!(feature = "full-node") {
+            "full-node"
+        } else {
+            "validator"
+        },
+        &self_address,
+        process::id()
+    );
+
+    let network_config = node_config.network;
+
+    let mut dp_builder = DataplaneBuilder::new(
+        &SocketAddr::V4(self_address),
+        network_config.max_mbps.into(),
+    );
+    if let Some(buffer_size) = network_config.buffer_size {
+        dp_builder = dp_builder.with_udp_buffer_size(buffer_size);
+    }
+
     let self_record = NameRecord {
         address: self_address,
         seq: peer_discovery_config.self_record_seq_num,
@@ -598,22 +628,70 @@ where
         max_active_connections: peer_discovery_config.max_active_connections,
         rng_seed: 123456,
     };
-    RaptorCast::new(RaptorCastConfig {
-        key: identity,
-        full_nodes: full_nodes
-            .iter()
-            .map(|full_node| NodeId::new(full_node.secp256k1_pubkey))
-            .collect(),
+
+    let secondary_instance = {
+        if let Some(cfg_2nd) = node_config.fullnode_raptorcast {
+            match cfg_2nd.mode {
+                monad_node_config::fullnode_raptorcast::SecondaryRaptorCastModeConfig::None => {
+                    debug!("Configured with Secondary RaptorCast instance: None");
+                    SecondaryRaptorCastModeConfig::None
+                }
+
+                monad_node_config::fullnode_raptorcast::SecondaryRaptorCastModeConfig::Client => {
+                    debug!("Configured with Secondary RaptorCast instance: Client");
+                    SecondaryRaptorCastModeConfig::Client(RaptorCastConfigSecondaryClient {
+                        bandwidth_cost_per_group_member: cfg_2nd.bandwidth_cost_per_group_member,
+                        bandwidth_capacity: cfg_2nd.bandwidth_capacity,
+                        invite_future_dist_min: cfg_2nd.invite_future_dist_min,
+                        invite_future_dist_max: cfg_2nd.invite_future_dist_max,
+                        invite_accept_heartbeat: Duration::from_millis(cfg_2nd.invite_accept_heartbeat_ms),
+                        raptor10_redundancy: cfg_2nd.raptor10_fullnode_redundancy_factor,
+                    })
+                }
+
+                monad_node_config::fullnode_raptorcast::SecondaryRaptorCastModeConfig::Publisher => {
+                    debug!("Configured with Secondary RaptorCast instance: Publisher");
+                    let full_nodes_prioritized: Vec<NodeId<CertificateSignaturePubKey<ST>>> = cfg_2nd
+                        .full_nodes_prioritized
+                        .identities
+                        .iter()
+                        .map(|id| NodeId::new(id.secp256k1_pubkey))
+                        .collect();
+                    SecondaryRaptorCastModeConfig::Publisher(RaptorCastConfigSecondaryPublisher {
+                        full_nodes_prioritized,
+                        group_scheduling: GroupSchedulingConfig {
+                            max_group_size: cfg_2nd.max_group_size,
+                            round_span: cfg_2nd.round_span,
+                            invite_lookahead: cfg_2nd.invite_lookahead,
+                            max_invite_wait: cfg_2nd.max_invite_wait,
+                            deadline_round_dist: cfg_2nd.deadline_round_dist,
+                            init_empty_round_span: cfg_2nd.init_empty_round_span,
+                        },
+                        raptor10_redundancy: cfg_2nd.raptor10_fullnode_redundancy_factor,
+                    })
+                }
+            }
+        } else {
+            SecondaryRaptorCastModeConfig::None
+        }
+    };
+
+    MultiRouter::new(
+        RaptorCastConfig {
+            shared_key: Arc::new(identity),
+            mtu: network_config.mtu,
+            primary_instance: RaptorCastConfigPrimary {
+                raptor10_redundancy: node_config.raptor10_validator_redundancy_factor,
+                fullnode_dedicated: full_nodes
+                    .iter()
+                    .map(|full_node| NodeId::new(full_node.secp256k1_pubkey))
+                    .collect(),
+            },
+            secondary_instance,
+        },
+        dp_builder,
         peer_discovery_builder,
-        redundancy: 3,
-        local_addr: SocketAddr::V4(SocketAddrV4::new(
-            network_config.bind_address_host,
-            network_config.bind_address_port,
-        )),
-        up_bandwidth_mbps: network_config.max_mbps.into(),
-        mtu: network_config.mtu,
-        buffer_size: network_config.buffer_size,
-    })
+    )
 }
 
 fn resolve_domain_v4<P: PubKey>(node_id: &NodeId<P>, domain: &String) -> Option<SocketAddr> {
--- a/monad-peer-discovery/src/discovery.rs
+++ b/monad-peer-discovery/src/discovery.rs
@@ -16,8 +16,8 @@ use tracing::{debug, info, trace, warn};
 
 use crate::{
     MonadNameRecord, NameRecord, PeerDiscoveryAlgo, PeerDiscoveryAlgoBuilder, PeerDiscoveryCommand,
-    PeerDiscoveryEvent, PeerDiscoveryMessage, PeerDiscoveryTimerCommand, PeerLookupRequest,
-    PeerLookupResponse, Ping, Pong, TimerKind,
+    PeerDiscoveryEvent, PeerDiscoveryMessage, PeerDiscoveryMetricsCommand,
+    PeerDiscoveryTimerCommand, PeerLookupRequest, PeerLookupResponse, Ping, Pong, TimerKind,
 };
 
 /// Maximum number of peers to be included in a PeerLookupResponse
@@ -737,6 +737,11 @@ where
 
         // reset timer to schedule for the next refresh
         cmds.extend(self.reset_refresh_timer());
+
+        // export metrics
+        cmds.push(PeerDiscoveryCommand::MetricsCommand(
+            PeerDiscoveryMetricsCommand(self.metrics.clone()),
+        ));
 
         cmds
     }
--- a/monad-raptorcast/examples/service.rs
+++ b/monad-raptorcast/examples/service.rs
@@ -2,6 +2,7 @@ use std::{
     collections::{BTreeMap, HashMap},
     net::{SocketAddr, SocketAddrV4},
     num::ParseIntError,
+    sync::Arc,
     time::{Duration, Instant},
 };
 
@@ -13,11 +14,9 @@ use monad_crypto::certificate_signature::{
     CertificateKeyPair, CertificateSignature, CertificateSignaturePubKey,
     CertificateSignatureRecoverable, PubKey,
 };
-use monad_dataplane::udp::DEFAULT_MTU;
 use monad_executor::Executor;
 use monad_executor_glue::{Message, RouterCommand};
-use monad_peer_discovery::mock::{NopDiscovery, NopDiscoveryBuilder};
-use monad_raptorcast::{RaptorCast, RaptorCastConfig, RaptorCastEvent};
+use monad_raptorcast::{new_defaulted_raptorcast_for_tests, RaptorCastEvent};
 use monad_secp::SecpSignature;
 use monad_types::{Deserializable, Epoch, NodeId, RouterTarget, Serializable, Stake};
 use tracing_subscriber::fmt::format::FmtSpan;
@@ -46,7 +45,6 @@ pub fn main() {
 
 type SignatureType = SecpSignature;
 type PubKeyType = CertificateSignaturePubKey<SignatureType>;
-type PeerDiscoveryType = NopDiscovery<SignatureType>;
 
 fn service(
     num_rt: usize,
@@ -55,6 +53,7 @@ fn service(
     num_broadcast: u32,
     message_len: usize,
 ) {
+    // Configure peers (5 in the example)
     assert!(message_len >= 4);
     let num_peers = addresses.len() as u32;
     let keys: Vec<_> = (0..num_peers)
@@ -94,7 +93,7 @@ fn service(
         std::sync::mpsc::channel::<(NodeId<PubKeyType>, <MockMessage as Message>::Event)>();
 
     let rts: Vec<_> = std::iter::repeat_with(|| {
-        tokio::runtime::Builder::new_multi_thread()
+        tokio::runtime::Builder::new_multi_thread() // The MultiThread builder spawn a thread
             .enable_all()
             .worker_threads(threads_per_rt)
             .build()
@@ -102,7 +101,7 @@ fn service(
     })
     .take(num_rt)
     .collect();
-
+    // Iterate the 2 runtime cycles
     rts.iter()
         .cycle()
         .zip(keys.into_iter().zip(tx_reader))
@@ -114,33 +113,21 @@ fn service(
             let known_addresses = known_addresses.clone();
 
             rt.spawn(async move {
-                let service_config = RaptorCastConfig {
-                    key,
-                    full_nodes: Default::default(),
-                    peer_discovery_builder: NopDiscoveryBuilder {
-                        known_addresses,
-                        ..Default::default()
-                    },
-                    redundancy: 2,
-                    local_addr: server_address,
-                    up_bandwidth_mbps: 1_000,
-                    mtu: DEFAULT_MTU,
-                    buffer_size: None,
-                };
-
-                let mut service = RaptorCast::<
+                let mut service = new_defaulted_raptorcast_for_tests::<
                     SignatureType,
                     MockMessage,
                     MockMessage,
                     <MockMessage as Message>::Event,
-                    PeerDiscoveryType,
-                >::new(service_config);
+                >(server_address, known_addresses, Arc::new(key));
+
                 service.exec(vec![RouterCommand::AddEpochValidatorSet {
                     epoch: Epoch(0),
-                    validator_set: all_peers.iter().map(|peer| (*peer, Stake(0))).collect(),
+                    validator_set: all_peers.iter().map(|peer| (*peer, Stake(1))).collect(),
                 }]);
+
                 loop {
                     tokio::select! {
+                        // service.next() retrieves the next event from the RaptorCast service.
                         maybe_message = service.next() => {
                             let message = maybe_message.expect("never terminates");
                             rx_writer
@@ -158,18 +145,19 @@ fn service(
         });
 
     let (tx_peer, tx_router) = tx_writer.first_key_value().expect("at least 1 tx");
-
+    // Here: 1 `service` main thread + `9 `tokio-runtime-w` threads
     std::thread::sleep(Duration::from_secs(1));
 
     let start = Instant::now();
     let mut expected_message_ids = HashMap::new();
     for broadcast_id in 0..num_broadcast {
         let message = MockMessage::new(broadcast_id, message_len);
+        let command = RouterCommand::Publish {
+            target: RouterTarget::Raptorcast(Epoch(0)),
+            message,
+        };
         tx_router
-            .send(RouterCommand::Publish {
-                target: RouterTarget::Broadcast(Epoch(0)),
-                message,
-            })
+            .send(command)
             .expect("reader should never be dropped");
         expected_message_ids.insert(message.id, num_peers);
     }
--- a/monad-raptorcast/src/config.rs
+++ b/monad-raptorcast/src/config.rs
@@ -1,11 +1,10 @@
-use std::{collections::HashMap, net::SocketAddr, sync::Arc};
+use std::{sync::Arc, time::Duration};
 
 use monad_crypto::certificate_signature::{
     CertificateSignaturePubKey, CertificateSignatureRecoverable,
 };
 use monad_types::{NodeId, Round};
 
-#[derive(Clone)]
 pub struct RaptorCastConfig<ST>
 where
     ST: CertificateSignatureRecoverable,
@@ -15,9 +14,6 @@ where
     // to both primary and secondary instances, so we don't wan't to partially
     // move the Config when wrapping shared_key into an Arc during the ctor.
     pub shared_key: Arc<ST::KeyPairType>,
-
-    // IP addresses ov the entries in fullnode_dedicated and full_nodes_prioritized
-    pub known_addresses: HashMap<NodeId<CertificateSignaturePubKey<ST>>, SocketAddr>,
 
     // For splitting large app messages (e.g. block proposals) into chunks that
     // fit into UDP datagrams for the raptorcast protocol.
@@ -33,6 +29,20 @@ where
     // Validators and full-nodes who do not want to participate in validator-
     // to-full-node raptor-casting may opt out of this.
     pub secondary_instance: SecondaryRaptorCastModeConfig<ST>,
+}
+
+impl<ST> Clone for RaptorCastConfig<ST>
+where
+    ST: CertificateSignatureRecoverable,
+{
+    fn clone(&self) -> Self {
+        RaptorCastConfig {
+            shared_key: self.shared_key.clone(),
+            mtu: self.mtu,
+            primary_instance: self.primary_instance.clone(),
+            secondary_instance: self.secondary_instance.clone(),
+        }
+    }
 }
 
 /// Configuration for the primary instance of RaptorCast (group of validators)
@@ -82,9 +92,13 @@ pub struct RaptorCastConfigSecondaryClient {
     pub bandwidth_cost_per_group_member: u64,
     pub bandwidth_capacity: u64,
     // When being invited to a raptorcast group, we will only accept the invite
-    // if the group is not too far or too soon in the future.
+    // if the group is not too far or too soon in the future, unless we haven't
+    // seen any proposals in `invite_accept_heartbeat`
     pub invite_future_dist_min: Round,
     pub invite_future_dist_max: Round,
+    pub invite_accept_heartbeat: Duration,
+    /// This applies when sending point-to-point group messages
+    pub raptor10_redundancy: u8,
 }
 
 impl Default for RaptorCastConfigSecondaryClient {
@@ -93,7 +107,9 @@ impl Default for RaptorCastConfigSecondaryClient {
             bandwidth_cost_per_group_member: 1,
             bandwidth_capacity: u64::MAX,
             invite_future_dist_min: Round(1),
-            invite_future_dist_max: Round(36000), // ~10 hours into the future, with default round length of 500ms
+            invite_future_dist_max: Round(600), // ~5 minutes into the future, with current round length of 500ms
+            invite_accept_heartbeat: Duration::from_secs(10),
+            raptor10_redundancy: 2,
         }
     }
 }
--- a/monad-raptorcast/src/lib.rs
+++ b/monad-raptorcast/src/lib.rs
@@ -1,10 +1,13 @@
 use std::{
-    cell::RefCell,
     collections::{BTreeMap, HashMap, VecDeque},
     marker::PhantomData,
-    net::SocketAddr,
+    net::{SocketAddr, SocketAddrV4},
     ops::DerefMut,
     pin::{pin, Pin},
+    sync::{
+        mpsc::{Receiver, Sender},
+        Arc, Mutex,
+    },
     task::{Context, Poll, Waker},
     time::Duration,
 };
@@ -12,13 +15,15 @@ use std::{
 use alloy_rlp::{Decodable, Encodable};
 use bytes::{Bytes, BytesMut};
 use futures::{channel::oneshot, FutureExt, Stream, StreamExt};
+use itertools::Itertools;
 use message::{InboundRouterMessage, OutboundRouterMessage};
 use monad_consensus_types::signature_collection::SignatureCollection;
 use monad_crypto::certificate_signature::{
     CertificateKeyPair, CertificateSignaturePubKey, CertificateSignatureRecoverable,
 };
 use monad_dataplane::{
-    udp::segment_size_for_mtu, BroadcastMsg, Dataplane, DataplaneBuilder, TcpMsg, UnicastMsg,
+    udp::{segment_size_for_mtu, DEFAULT_MTU},
+    BroadcastMsg, Dataplane, DataplaneBuilder, RecvMsg, TcpMsg, UnicastMsg,
 };
 use monad_executor::{Executor, ExecutorMetrics, ExecutorMetricsChain};
 use monad_executor_glue::{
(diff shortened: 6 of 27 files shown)


monad-raptorcast/Cargo.toml
```diff
@@ -20,6 +20,7 @@ monad-peer-discovery = { workspace = true }
 monad-raptor = { workspace = true }
 monad-secp = { workspace = true }
 monad-types = { workspace = true }
+monad-testutil = { workspace = true }
```
eerkaijun: move this to dev dependencies

monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -257,20 +330,46 @@ where
...
                     }
                     Role::Publisher(publisher) => {
+                        tracing::trace!("RaptorCastSecondary UpdateCurrentRound (Publisher): epoch: {:?}, round: {:?}", epoch, round);
                         self.curr_epoch = epoch;
+                        // The publisher needs to be periodically informed about new nodes out there,
+                        // so that it can randomize when creating new groups.
+                        {
+                            let pd = self.peer_discovery_driver.lock().unwrap();
+                            let latest_pd_seqno = pd.get_global_seqno();
+                            if latest_pd_seqno > self.name_records_cache_seqno {
```
eerkaijun: with our current peer discovery implementation, this probably happen very often, so the caches are updated very frequently

monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -356,16 +455,71 @@ where
...
                 Role::Client(client) => {
+                    tracing::trace!("RaptorCastSecondary received group message");
                     // Received group message from validator
+                    if let FullNodesGroupMessage::ConfirmGroup(confirm_msg) = &inbound_grp_msg {
+                        let num_mappings = confirm_msg.name_records.len();
+                        if num_mappings > 0 {
+                            if num_mappings == confirm_msg.peers.len() {
+                                let mut peers: Vec<PeerEntry<ST>> = Vec::new();
+                                for rec in &confirm_msg.name_records {
+                                    if let Ok(node_id) = rec.recover_pubkey() {
```
eerkaijun: is it possible to use `confirm_msg.peers` to get the node id? `recover_pubkey` is more expensive and we do that again inside `update_peers`

monad-raptorcast/Cargo.toml
fabiotdf: Done

monad-raptorcast/src/lib.rs
```diff
@@ -367,61 +485,29 @@ where
...
-        known_addresses,
-    );
-
-    UnicastMsg {
-        msgs: messages,
-        stride: segment_size,
+        ExecutorMetricsChain::default().push(self.metrics.as_ref())
+        // FIXME: retrieving metrics from an object behind a Mutex
+        // probably requires some refactoring somewhere.
+        //.push(self.peer_discovery_driver.metrics())
```
fabiotdf: If someone knows an easy fix for this, please give me a shout. 
If not straightforward, maybe we could create a separate PR for the refactoring?

monad-peer-discovery/src/discovery.rs
```diff
@@ -83,6 +83,9 @@ pub struct PeerDiscovery<ST: CertificateSignatureRecoverable> {
     // maximum number of active connections before pruning
     pub max_active_connections: usize,
     pub rng: ChaCha8Rng,
+    // Incremented by one every time peer_info changes. Helps the user minimize
+    // the frequency of calls to get_name_records()
+    pub global_seqno: u64,
```
michael-yxchen: I'm inclined to not introduce this change, as it will soon be replaced by refactoring `build_messages` to accept a peer discovery driver. let me see if I can write it quickly on top of your branch

monad-node-config/src/fullnode_raptorcast.rs
```diff
@@ -9,7 +9,7 @@ use super::fullnode::FullNodeConfig;
 pub enum SecondaryRaptorCastModeConfig {
     Client,
     Publisher,
-    // No "None" option, as FullNodeRaptorCastConfig is now wrapped in Option<>
+    None, // Simplifies flexnet tests
```
michael-yxchen: does None mean that secondary raptorcast is disabled?

monad-node/src/main.rs
```diff
@@ -483,14 +491,14 @@ async fn run(node_state: NodeState, reload_handle: ReloadHandle) -> Result<(), (
 }
 
 async fn build_raptorcast_router<ST, SCT, M, OM>(
-    network_config: NodeNetworkConfig,
+    node_config: NodeConfig<ST>, // network_config: NodeNetworkConfig,
```
michael-yxchen: nit: cleanup

monad-router-multi/src/lib.rs
```diff
@@ -0,0 +1,187 @@
...
+impl<ST, M, OM, SE, PD> MultiRouter<ST, M, OM, SE, PD>
+where
+    ST: CertificateSignatureRecoverable,
+    M: Message<NodeIdPubKey = CertificateSignaturePubKey<ST>> + Decodable,
+    OM: Encodable + Into<M> + Clone,
+    PD: PeerDiscoveryAlgo<SignatureType = ST>,
+{
+    pub fn new<B>(
+        cfg: &RaptorCastConfig<ST>,
+        shared_dataplane: Arc<Mutex<Dataplane>>,
```
michael-yxchen: cleaner to pass in `DataplaneBuilder` and construct `Dataplane` in `MultiRouter`

monad-node/src/main.rs
```diff
@@ -598,22 +629,70 @@ where
...
         rng_seed: 123456,
     };
-    RaptorCast::new(RaptorCastConfig {
-        key: identity,
-        full_nodes: full_nodes
-            .iter()
-            .map(|full_node| NodeId::new(full_node.secp256k1_pubkey))
-            .collect(),
+
+    let secondary_instance = {
```
michael-yxchen, line 632: we should think about how a node can convert between publisher and client. This happens when they enter/leave the validator set. Problem for later though

monad-node-config/src/fullnode_raptorcast.rs
fabiotdf: Correct. In this case the primary raptorcast instance functions standalone like before

monad-node/src/main.rs
fabiotdf: Does that imply switching between validator and full-node roles?

monad-node/src/main.rs
fabiotdf: I sware I removed this comment 100 times!!! Not sure how it got back here, but removed now :)

monad-peer-discovery/src/discovery.rs
fabiotdf: Removing & testing

monad-node/src/main.rs
michael-yxchen: yeah they can switch between validator & full node roles

monad-node-config/src/fullnode_raptorcast.rs
michael-yxchen: can we update the comment?

monad-raptorcast/src/raptorcast_secondary/mod.rs
fabiotdf: removed caches

monad-raptorcast/src/raptorcast_secondary/mod.rs
fabiotdf: Done

monad-router-multi/src/lib.rs
fabiotdf: Done

monad-raptorcast/src/raptorcast_secondary/client.rs
```diff
@@ -261,13 +263,16 @@ where
...
-                        "RaptorcastSecondary ignoring late confirm, curr_round \
+                        "RaptorCastSecondary ignoring late confirm, curr_round \
                         {:?}, confirm = {:?}",
                         self.curr_round,
                         confirm_msg
                     );
                     return None;
                 }
+
+                let is_receiving_proposals = self.is_receiving_proposals();
```
michael-yxchen, line 274: nit: remove variable

monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -134,13 +147,13 @@ where
...
         build_target: BuildTarget<ST>,
-        app_message: Bytes,
+        outbound_message: Bytes,
         mtu: u16,
         signing_key: &Arc<ST::KeyPairType>,
         redundancy: u8,
         known_addresses: &HashMap<NodeId<CertificateSignaturePubKey<ST>>, SocketAddr>,
     ) -> UnicastMsg {
-        let app_message_len = app_message.len();
+        let app_message_len = outbound_message.len();
```
michael-yxchen: nit
```suggestion
        let outbound_message_len = outbound_message.len();
```

monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -153,10 +166,13 @@ where
             .try_into()
             .expect("unix epoch doesn't fit in u64");
 
+        tracing::trace!("RaptorCastSecondary::udp_build() mtu: {}, app_message_len: {}, redundancy: {}, segment_size: {}",
+            mtu, app_message_len, redundancy, segment_size);
```
michael-yxchen:
```suggestion
        tracing::trace!(?mtu, ?outbound_message_len, ?redundancy, ?segment_size, "RaptorCastSecondary::udp_build());
```

monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -174,7 +190,14 @@ where
         &self,
         group_msg: FullNodesGroupMessage<ST>,
         dest_node: &NodeId<CertificateSignaturePubKey<ST>>,
+        known_addresses: &HashMap<NodeId<CertificateSignaturePubKey<ST>>, SocketAddr>,
     ) {
+        tracing::trace!(
+            "RaptorCastSecondary send_single_msg to {:?}: {:?}",
+            dest_node,
+            group_msg
+        );
```
michael-yxchen:
```suggestion
        tracing::trace!(
            ?dest_node,
            ?group_msg,
            "RaptorCastSecondary send_single_msg"
        );
```

monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -197,35 +220,66 @@ where
         &self,
         group_msg: FullNodesGroupMessage<ST>,
         dest_node_ids: FullNodes<CertificateSignaturePubKey<ST>>,
+        known_addresses: &HashMap<NodeId<CertificateSignaturePubKey<ST>>, SocketAddr>,
     ) {
+        tracing::trace!(
+            "RaptorCastSecondary send_group_msg to {:?}: {:?}",
+            dest_node_ids,
+            group_msg
+        );
```
michael-yxchen: similar here

monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -197,35 +220,66 @@ where
...
+                        "RaptorCastSecondary: No name record found for node_id {:?} when sending out ConfirmGroup message: {:?}",
+                        node_id, group_msg
+                    );
+                }
+            }
+            return FullNodesGroupMessage::ConfirmGroup(filled_confirm_msg);
+        }
         group_msg
-        // TODO: ask peer recovery for ConfirmGroup::name_records
     }
```
michael-yxchen, line 273: looks like it's only used when sending a ConfirmGroup message. Should we move it to `publisher.rs`?

monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -197,35 +220,66 @@ where
...
-    M: Message<NodeIdPubKey = CertificateSignaturePubKey<ST>> + Deserializable<Bytes> + Decodable,
-    OM: Serializable<Bytes> + Into<M> + Clone + Encodable,
+    M: Message<NodeIdPubKey = CertificateSignaturePubKey<ST>> + Decodable,
+    OM: Encodable + Into<M> + Clone,
+    PD: PeerDiscoveryAlgo<SignatureType = ST>,
 {
     type Command = RouterCommand<ST, OM>;
 
     fn exec(&mut self, commands: Vec<Self::Command>) {
+        tracing::trace!("RaptorCastSecondary Exec {} commands", commands.len());
```
michael-yxchen: i think we can remove traces like these

monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -249,24 +303,48 @@ where
...
+                        {
+                            let pd = self.peer_discovery_driver.lock().unwrap();
+                            let known_addresses = pd.get_known_addresses();
+                            let nodes: Vec<_> = known_addresses.keys().copied().collect();
+                            tracing::trace!(
+                                "RaptorCastSecondary updating {} full nodes from PeerDiscovery",
+                                nodes.len()
+                            );
+                            publisher.upsert_peer_disc_full_nodes(FullNodes::new(nodes));
+                        }
```
michael-yxchen: is this still needed now that we removed peers cache?

monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -249,24 +303,48 @@ where
...
+
                         if let Some((group_msg, fn_set)) =
                             publisher.enter_round_and_step_until(round)
                         {
-                            self.send_group_msg(group_msg, fn_set);
+                            let known_addresses = {
+                                let pd = self.peer_discovery_driver.lock().unwrap();
+                                &pd.get_known_addresses()
+                            };
+                            self.send_group_msg(group_msg, fn_set, known_addresses);
```
michael-yxchen: `get_known_addresses` returns by value
```suggestion
                            let known_addresses = self.peer_discovery_driver.lock().unwrap().get_known_addresses();
                            self.send_group_msg(group_msg, fn_set, &known_addresses);
```

monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -249,24 +303,48 @@ where
...
+                        OutboundRouterMessage::<OM, ST>::AppMessage(message).serialize();
+                    let outbound_message_len = outbound_message.len();
 
                     let _timer = DropTimer::start(Duration::from_millis(20), |elapsed| {
-                        tracing::warn!(?elapsed, app_message_len, "long time to publish message")
+                        tracing::warn!(
+                            ?elapsed,
+                            outbound_message_len,
+                            "long time to publish message"
+                        )
```
michael-yxchen: move role and group size check above serialization

monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -275,41 +353,45 @@ where
...
+                    // send to full nodes.
+                    let rc_chunks: UnicastMsg = Self::udp_build(
+                        &epoch,
+                        build_target,
+                        outbound_message,
+                        self.mtu,
+                        &self.signing_key,
+                        self.raptor10_redundancy,
+                        &known_addresses,
+                    );
```
michael-yxchen:
```suggestion
                    let known_addresses = self.peer_discovery_driver.lock().unwrap().get_known_addresses();

                    // Split outbound_message into raptorcast chunks that we can
                    // send to full nodes.
                    let rc_chunks: UnicastMsg = Self::udp_build(
                        &epoch,
                        build_target,
                        outbound_message,
                        self.mtu,
                        &self.signing_key,
                        self.raptor10_redundancy,
                        &known_addresses,
                    );
```

monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -347,16 +430,73 @@ where
         match this.channel_from_primary.try_recv() {
             Ok(inbound_grp_msg) => match &mut this.role {
                 Role::Publisher(publisher) => {
+                    tracing::trace!("RaptorCastSecondary received candidate response");
```
michael-yxchen: remove trace or move to `on_candidate_response`

monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -347,16 +430,73 @@ where
...
                     publisher.on_candidate_response(inbound_grp_msg);
                 }
 
                 Role::Client(client) => {
+                    tracing::trace!("RaptorCastSecondary received group message");
                     // Received group message from validator
+                    if let FullNodesGroupMessage::ConfirmGroup(confirm_msg) = &inbound_grp_msg {
+                        let num_mappings = confirm_msg.name_records.len();
+                        if num_mappings > 0 {
+                            if num_mappings == confirm_msg.peers.len() {
```
michael-yxchen, line 447: can the two `if` be simplified?

monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -347,16 +430,73 @@ where
...
+                                        pubkey: confirm_msg.peers[ii].pubkey(),
+                                        addr: rec.name_record.address,
+                                        signature: rec.signature,
+                                        record_seq_num: rec.name_record.seq,
+                                    };
+                                    peers.push(peer_entry);
+                                }
+                                for rec in &confirm_msg.name_records {
+                                    if let Ok(node_id) = rec.recover_pubkey() {
+                                        let peer_entry = PeerEntry {
```
michael-yxchen: here's crypto primitive benchmark data if you need them: https://www.notion.so/monad-bft-crypto-bench-devnet-spec-15c75b0ba8408175ab06d9d0d1e0cf7c?source=copy_link

remove inline benchmark

monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -347,16 +430,73 @@ where
...
+                                            addr: rec.name_record.address,
+                                            signature: rec.signature,
+                                            record_seq_num: rec.name_record.seq,
+                                        };
+                                        peers.push(peer_entry);
+                                    } else {
+                                        tracing::warn!(
+                                            "RaptorCastSecondary failed to recover pubkey in name record: {:?}, message: {:?}",
+                                            rec, confirm_msg
+                                        );
```
michael-yxchen: we prefer to use kv representation provided by tracing. easier to parse json logs that way
```suggestion
                                        tracing::warn!(
                                            ?rec, ?message,
                                            "RaptorCastSecondary failed to recover pubkey in name record
                                        );
```

monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -347,16 +430,73 @@ where
...
+                                        return Poll::Pending; // Drop the message
+                                    }
+                                }
+                                let mut pd = this.peer_discovery_driver.lock().unwrap();
+                                pd.update(PeerDiscoveryEvent::UpdatePeers { peers });
```
michael-yxchen, line 461: nit: combine into single line
```suggestion
                                this.peer_discovery_driver.lock().unwrap().update(PeerDiscoveryEvent::UpdatePeers { peers });
```

monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -347,16 +430,73 @@ where
...
+                            } else {
+                                tracing::warn!(
+                                    "Number of peers {} does not match the number \
+                                    of name records {} in ConfirmGroup message: {:?}. \
+                                    Skipping PeerDiscovery update",
+                                    confirm_msg.peers.len(),
+                                    confirm_msg.name_records.len(),
+                                    confirm_msg
+                                );
```
michael-yxchen, line 470: we should revisit this. the protocol doesn't require that populated name records to be all or none. publisher can choose to attach a subset, e.g. only new joiners, assuming group churn is low

monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -347,16 +430,73 @@ where
...
                     {
                         // Send back a response to the validator
-                        this.send_single_msg(response_msg, &validator_id);
+                        tracing::trace!(
+                            "RaptorCastSecondary sending back response for group message"
+                        );
+
+                        let known_addresses = {
+                            let pd = this.peer_discovery_driver.lock().unwrap();
+                            &pd.get_known_addresses()
```
michael-yxchen: same here

monad-raptorcast/src/raptorcast_secondary/publisher.rs
```diff
@@ -70,13 +69,20 @@ where
 
         // Remove duplicate entries from always_ask_full_nodes, but making sure
         // we maintain the original order/
+        tracing::trace!(
+            "RaptorCastSecondary initializing Publisher with {} prioritized full nodes",
+            config.full_nodes_prioritized.len()
+        );
```
michael-yxchen: kv format

monad-raptorcast/src/raptorcast_secondary/publisher.rs
```diff
@@ -70,13 +69,20 @@ where
...
             let mut seen = HashSet::new();
             for node in config.full_nodes_prioritized {
                 if seen.insert(node) {
+                    tracing::trace!("  .. prioritized: {:?}", node);
                     always_ask_full_nodes.list.push(node);
+                } else {
+                    tracing::trace!("  .. duplicate: {:?}", node);
                 }
             }
         }
```
michael-yxchen: we should either bump log level higher, to info, or silently dedup for them

monad-raptorcast/src/raptorcast_secondary/publisher.rs
```diff
@@ -109,6 +115,7 @@ where
         &mut self,
         round: Round,
     ) -> Option<(FullNodesGroupMessage<ST>, FullNodesST<ST>)> {
+        tracing::trace!("enter_round_and_step_until: round: {:?}", round);
```
michael-yxchen: kv format

monad-raptorcast/src/raptorcast_secondary/publisher.rs
```diff
@@ -137,6 +144,7 @@ where
     // Populate self.curr_group and clean up expired groups.
     // When we have a real timer, this can be called from UpdateCurrentRound
     fn enter_round(&mut self, round: Round) {
+        tracing::trace!("enter_round: round: {:?}", round);
```
michael-yxchen: ^

monad-raptorcast/src/raptorcast_secondary/publisher.rs
```diff
@@ -201,11 +209,21 @@ where
             None => self.curr_group.get_round_span().end,
         };
 
+        tracing::trace!(
+            "RaptorCastSecondary step_until: time_point: {:?}, schedule_end: {:?}",
+            time_point,
+            schedule_end
+        );
```
michael-yxchen: ^

monad-raptorcast/src/raptorcast_secondary/publisher.rs
```diff
@@ -279,7 +303,6 @@ where
             .retain(|node| !self.always_ask_full_nodes.list.contains(node));
     }
 
-    #[allow(dead_code)] // Placeholder for peer discovery
     pub fn upsert_peer_disc_full_nodes(
         &mut self,
         additional_fn: FullNodes<CertificateSignaturePubKey<ST>>,
```
michael-yxchen, line 305: can this function be removed? we can query peer discovery at prepare group time

monad-raptorcast/src/raptorcast_secondary/publisher.rs
```diff
@@ -428,6 +453,10 @@ where
         );
         let invite_msg = FullNodesGroupMessage::PrepareGroup(prep_grp_data);
         self.num_invites_sent += next_invitees.list.len();
+        tracing::trace!(
+            "RaptorCastSecondary Publisher advance_invites: send Invites to peers: {:?}",
+            next_invitees.list
+        );
```
michael-yxchen: ^

monad-raptorcast/src/config.rs
```diff
@@ -93,7 +94,9 @@ impl Default for RaptorCastConfigSecondaryClient {
             bandwidth_cost_per_group_member: 1,
             bandwidth_capacity: u64::MAX,
             invite_future_dist_min: Round(1),
-            invite_future_dist_max: Round(36000), // ~10 hours into the future, with default round length of 500ms
+            invite_future_dist_max: Round(36000), // ~10 hours into the future, with current round length of 500ms
```
michael-yxchen: 10 hrs is too long. I would imagine this to be on the order of minutes. also calculation is off by 2x

monad-raptorcast/src/lib.rs
```diff
@@ -67,24 +55,26 @@ where
...
 
     current_epoch: Epoch,
 
     udp_state: udp::UdpState<ST>,
     mtu: u16,
 
-    dataplane: Dataplane,
+    dataplane: Arc<Mutex<Dataplane>>,
     pending_events: VecDeque<RaptorCastEvent<M::Event, ST>>,
+    channel_to_secondary: Option<Sender<FullNodesGroupMessage<ST>>>,
```
michael-yxchen, line 77: super nit: empty line before `channel_to_secondary`

monad-raptorcast/src/lib.rs
```diff
@@ -108,27 +98,40 @@ where
...
-    where
-        B: PeerDiscoveryAlgoBuilder<PeerDiscoveryAlgoType = PD>,
-    {
-        assert!(config.redundancy >= 1);
-        let self_id = NodeId::new(config.key.pubkey());
-        let mut builder = DataplaneBuilder::new(&config.local_addr, config.up_bandwidth_mbps);
-        if let Some(buffer_size) = config.buffer_size {
-            builder = builder.with_udp_buffer_size(buffer_size);
+    pub fn new(
+        config: &config::RaptorCastConfig<ST>,
```
michael-yxchen: taking in config by reference is weird. they should be cheap to copy

monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -114,15 +113,29 @@ where
...
+        );
+
+        if raptor10_redundancy < 1 {
+            panic!(
+                "Configuration value raptor10_redundancy must be equal or greater than 1, \
+                but got {}. This is a bug in the configuration for the secondary instance.",
+                raptor10_redundancy
+            );
+        }
+
```
michael-yxchen, line 130: should we check redundancy value in publisher/client constructor instead?

monad-raptorcast/src/lib.rs
```diff
@@ -137,44 +140,109 @@ where
...
+    }
+
+    // Prepares an app message (e.g. proposal) to be sent out via UDP, by signing
+    // it and then splitting into raptorcast chunks fitting MTU (depending on build_target)
+    fn udp_build(
+        epoch: &Epoch,
+        build_target: BuildTarget<ST>,
+        outbound_message: Bytes,
+        mtu: u16,
+        signing_key: &Arc<ST::KeyPairType>,
```
michael-yxchen, line 223: the idiomatic way of using `Arc` is to always clone and pass by value, instead of taking a reference to it, as it's already a reference

monad-raptorcast/src/lib.rs
```diff
@@ -297,60 +384,91 @@ where
...
+                                            .copy_from_slice(&signature);
+                                        signed_message[SIGNATURE_SIZE..]
+                                            .copy_from_slice(&app_message);
+                                        self.dataplane.lock().unwrap().tcp_write(
+                                            address,
+                                            TcpMsg {
+                                                msg: signed_message.freeze(),
+                                                completion,
+                                            },
+                                        );
```
michael-yxchen: what's the motivation to inline this function? i find previous version more readable

monad-raptorcast/src/lib.rs
```diff
@@ -200,12 +274,19 @@ where
                         }
                     }
                     self.peer_discovery_driver
+                        .lock()
+                        .unwrap()
                         .update(PeerDiscoveryEvent::UpdateCurrentEpoch { epoch });
```
michael-yxchen: i think it makes sense to update epoch only if `self.current_epoch < epoch` since introduction of that mutex

monad-raptorcast/src/lib.rs
michael-yxchen: @andr-dev any ideas here

monad-raptorcast/src/lib.rs
```diff
@@ -449,17 +535,28 @@ where
...
-            };
+            let message: RecvMsg;
+            {
+                let mut dataplane = this.dataplane.lock().unwrap();
+                let Poll::Ready(msg) = pin!(dataplane.udp_read()).poll_unpin(cx) else {
+                    //drop(dataplane);
+                    break;
+                };
+                message = msg;
+            }
```
michael-yxchen: this is a weird pattern. is the lock not released with previous version?

monad-node/src/main.rs
fabiotdf: Ok, might be a big change. Thanks for the heads up

monad-node-config/src/fullnode_raptorcast.rs
fabiotdf: Done

monad-raptorcast/src/raptorcast_secondary/client.rs
fabiotdf: cannot borrow `*self` as immutable because it is also borrowed as mutable
immutable borrow occurs here

monad-raptorcast/src/raptorcast_secondary/mod.rs
fabiotdf: Done

monad-raptorcast/src/raptorcast_secondary/mod.rs
fabiotdf: Done

monad-raptorcast/src/raptorcast_secondary/publisher.rs
fabiotdf: Done

monad-raptorcast/src/raptorcast_secondary/publisher.rs
fabiotdf: Done

monad-raptorcast/src/raptorcast_secondary/publisher.rs
fabiotdf: Done

monad-raptorcast/src/raptorcast_secondary/mod.rs
fabiotdf: Done

monad-raptorcast/src/raptorcast_secondary/mod.rs
fabiotdf: Done

monad-raptorcast/src/raptorcast_secondary/publisher.rs
fabiotdf: I think not easily... Can we wait until the peer discovery is passed around instead of calling get_peers?

monad-raptorcast/src/raptorcast_secondary/publisher.rs
fabiotdf: Done

fabiotdf:
> for all tracing logs, we prefer to use syntax `field_name = field_value` to make parsing json logs easier. The short hand for that is `?field_name`

done

monad-raptorcast/src/raptorcast_secondary/mod.rs
fabiotdf: Done

monad-raptorcast/src/raptorcast_secondary/mod.rs
fabiotdf: Yes so that the publisher has an updated view of nodes, as it needs to form random groups

monad-raptorcast/src/raptorcast_secondary/mod.rs
fabiotdf: Ah, I had only temporarily added it to check the numbers

monad-raptorcast/src/raptorcast_secondary/mod.rs
fabiotdf: Done

monad-raptorcast/src/raptorcast_secondary/mod.rs
fabiotdf: Currently the publisher doesn't need to depend on peer discovery... Do we really want to introduce the dependency? I think once the dependency is introduced, its gonna be hard to isolate things for unit testing

monad-raptorcast/src/raptorcast_secondary/mod.rs
fabiotdf: done

monad-raptorcast/src/raptorcast_secondary/mod.rs
fabiotdf: I'm a bit worried about potentially keeping peer_discovery_driver mutex guard alive, like in that deadlock bug we were investigating and we never got to the bottom of how the guard escaped the expression `self.peer_discovery_driver.lock().unwrap()`

fabiotdf: I've tried to minimize lifetimes by using {} where locks are involved

michael-yxchen: I think since `get_known_addresses()` return HashMap by value, it won't actually hold the lock. would you mind testing it quickly to see if the lock is held?

monad-raptorcast/src/lib.rs
```diff
@@ -469,50 +566,89 @@ where
...
+                                            }
+                                        }
+                                        None => {
+                                            tracing::warn!(
+                                                ?from,
+                                                "Received FullNodesGroup message but the primary \
+                                Raptorcast instance is not setup to forward messages \
+                                to a secondary instance."
+                                            );
+                                        }
```
michael-yxchen: malicious parties can send us FullNodeGroup messages. let's nerf this log to debug so we don't spam warnings

monad-raptorcast/src/util.rs
```diff
@@ -479,18 +480,37 @@ where
...
         assert!(self.is_fullnode);
+        let new_grp = group.clone();
         let replaced = self.fullnode_map.insert(*group.get_validator_id(), group);
-        assert!(replaced.is_none());
+        match replaced {
+            Some(old_grp) => {
+                tracing::trace!("Group replace: new: {:?}, old: {:?}", new_grp, old_grp)
+            }
+            None => tracing::trace!("Group insert - new: {:?}, old: None", new_grp),
+        }
```
michael-yxchen: it's cleaner to write
```suggestion
        if let Some(old_grp) = self.fullnode_map.insert(*group.get_validator_id(), group.clone()) {
            tracing::trace!("Group replace: new: {:?}, old: {:?}", group, old_grp)
        } else {
            tracing::trace!("Group insert - new: {:?}, old: None", group),
        }
```

monad-raptorcast/examples/service.rs
```diff
@@ -114,33 +113,21 @@ fn service(
...
-                    },
-                    redundancy: 2,
-                    local_addr: server_address,
-                    up_bandwidth_mbps: 1_000,
-                    mtu: DEFAULT_MTU,
-                    buffer_size: None,
-                };
-
-                let mut service = RaptorCast::<
+                let mut service = new_defaulted_raptorcast_for_tests::<
```
michael-yxchen, line 116: im biased to not include this helper function, as it's only used in two places

monad-router-multi/src/lib.rs
```diff
@@ -0,0 +1,190 @@
...
+
+impl<ST, M, OM, SE, PD> MultiRouter<ST, M, OM, SE, PD>
+where
+    ST: CertificateSignatureRecoverable,
+    M: Message<NodeIdPubKey = CertificateSignaturePubKey<ST>> + Decodable,
+    OM: Encodable + Into<M> + Clone,
+    PD: PeerDiscoveryAlgo<SignatureType = ST>,
+{
+    pub fn new<B>(
+        cfg: &RaptorCastConfig<ST>,
```
michael-yxchen: we can pass cfg by value

monad-router-multi/src/lib.rs
```diff
@@ -0,0 +1,190 @@
...
+        let shared_dataplane = Arc::new(Mutex::new(dp));
+
+        // Create a Channels between primary and secondary raptorcast instances.
+        // Fundamentally this is needed because, while both can send, only the
+        // primary can receive data from the network.
+        type ChannelMsg<ST> = (
+            Sender<FullNodesGroupMessage<ST>>,
+            Receiver<FullNodesGroupMessage<ST>>,
+        );
+        type ChannelGrp<ST> = (Sender<Group<ST>>, Receiver<Group<ST>>);
```
michael-yxchen: are these type annotations strictly necessary to compile? I think we should make variable names (txm, rxm, txg, rxg) more explicit rather than relying on type annotations

monad-router-multi/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+edition = "2021"
+
+[dependencies]
+monad-crypto = { path = "../monad-crypto" }
+monad-dataplane = { path = "../monad-dataplane" }
+monad-executor = { path = "../monad-executor" }
+monad-executor-glue = { path = "../monad-executor-glue" }
+monad-peer-discovery = { workspace = true }
+monad-raptorcast = { path = "../monad-raptorcast" }
+
```
michael-yxchen:
```suggestion
monad-crypto = { workspace = true }
monad-dataplane = { workspace = true }
monad-executor = { workspace = true }
monad-executor-glue = { workspace = true }
monad-peer-discovery = { workspace = true }
monad-raptorcast = { workspace = true }
monad-types = { workspace = true }
```

monad-testground/src/executor.rs
```diff
@@ -94,17 +108,47 @@ where
...
+                Updater::boxed(MultiRouter::<
+                    ST,
+                    MonadMessage<ST, SCT, MockExecutionProtocol>,
+                    VerifiedMonadMessage<ST, SCT, MockExecutionProtocol>,
+                    MonadEvent<ST, SCT, MockExecutionProtocol>,
+                    NopDiscovery<ST>,
+                >::new(
+                    &cfg, dataplane_builder, peer_discovery_builder
+                ))
+            }
```
michael-yxchen: to match what's in monad-node, should we use MultiRouter for both `RaptorCast` and `MultiRaptorCast`?

monad-testground/src/main.rs
```diff
@@ -78,6 +88,7 @@ struct TestgroundArgs {
 enum RouterArgs {
     Local { external_latency_ms: u64 },
     RaptorCast,
+    MultiRaptorCast { num_full_nodes: u16 },
```
michael-yxchen: is `num_full_nodes` used?

monad-raptorcast/src/raptorcast_secondary/publisher.rs
michael-yxchen: yes, i'll create a ticket

monad-router-multi/src/lib.rs
fabiotdf: Done

monad-raptorcast/src/lib.rs
fabiotdf: It looks like this is the same as in C++, one should pass shared_ptr (C++ equivalent of Arc) by reference rather than copy. Check out the generated code below:

https://godbolt.org/z/6PKzKaTc5

image

monad-raptorcast/src/raptorcast_secondary/mod.rs
fabiotdf: Is that the proper place, though, because the publisher and client do not deal with network parameters?
Furthermore, raptorcast primary also does the same check, but doesn't have publisher/client objects. Still the check should be done there because if this value is below 1 the network will silently fail.

monad-raptorcast/src/lib.rs
fabiotdf: I did it by reference because it required less code: the fact that you can't clone KeyPairType (included in the config) is a bit of a pain in the ass! Implemented Clone for RaptorCastConfig (derive didn't work) and switched to copies nevertheless.

monad-raptorcast/src/lib.rs
fabiotdf: Done

monad-raptorcast/src/config.rs
fabiotdf: Nerfed to 5 minutes.

monad-raptorcast/src/raptorcast_secondary/mod.rs
michael-yxchen: https://github.com/category-labs/category-internal/issues/1728

monad-raptorcast/src/raptorcast_secondary/mod.rs
michael-yxchen: https://github.com/category-labs/category-internal/issues/1728

monad-raptorcast/src/raptorcast_secondary/publisher.rs
michael-yxchen: https://github.com/category-labs/category-internal/issues/1728

monad-router-multi/Cargo.toml
fabiotdf: Done

monad-raptorcast/src/raptorcast_secondary/mod.rs
michael-yxchen: i think i was confused because `raptor10_redundancy` is included in both `RaptorCastConfigSecondaryPublisher` and `RaptorCastConfigSecondaryClient`, which are passed to `Publisher/Client`. If publisher/client don't deal with network parameters, we should remove it from that part of the config

monad-testground/src/main.rs
fabiotdf: Aah its part of code I've removed. Cleaned up now.

monad-router-multi/src/lib.rs
fabiotdf: Fixed

monad-testground/src/executor.rs
fabiotdf: Well, I've removed MultiRaptorCast from testground because the code for full-node raptorcasting in testground wasn't working. Should we leave this feature for another PR? What do you think?

monad-raptorcast/src/lib.rs
michael-yxchen: let's punt the discussion. i'll follow up later

monad-testground/src/executor.rs
michael-yxchen: yes let’s deal with that later

monad-raptorcast/src/lib.rs
fabiotdf: IIRC there were issues borrowing self, and also it was only called from one place. I can try again

monad-raptorcast/src/lib.rs
fabiotdf: I agree it looks weird, but since we didn't quite understand how the lock guard leaked when we analyze that deadlock, I didn't want to risk it in production so I though this was safer... At least until we understand that surprise rust language behavior we saw?

monad-raptorcast/examples/service.rs
fabiotdf: Do you have a strong preference for this?
I was thinking `new_defaulted_raptorcast_for_tests` is more readable than copy-pasting this function body to two places, and also reduces the maintenance load. I constantly found myself having to merge changes in 3 places instead of just 1.  The review request around `raptor10_redundancy` earlier is yet another example of this!

monad-raptorcast/src/util.rs

monad-raptorcast/src/lib.rs
fabiotdf: Fixed

monad-raptorcast/src/lib.rs
fabiotdf: Fixed

fabiotdf:
I've removed the scope from the lock in the  "`get_known_addresses`" comment above and tested on flexnet as requested, so the change is included.
https://github.com/category-labs/monad-bft/pull/1870#discussion_r2180401504

monad-raptorcast/src/lib.rs
fabiotdf: No issues this time - fixed

monad-raptorcast/src/raptorcast_secondary/mod.rs
fabiotdf: OK, I will have to create a new config struct `RaptorCastConfigSecondary` that contains just this `raptor10_redundancy` + `pub secondary_instance`  from `RaptorCastConfig` though. Not sure you will like it

monad-raptorcast/src/lib.rs
michael-yxchen: yea sure

monad-raptorcast/examples/service.rs
michael-yxchen: no strong opinion. we can keep existing if that simplifies things for you

monad-raptorcast/src/raptorcast_secondary/mod.rs
michael-yxchen: i feel that's cleaner - everything is cfg is used. don't have to change it now though. we can cleanup later with other issues we punt on

monad-raptorcast/src/lib.rs
michael-yxchen: i think the proper way of doing this is to move ownership of peer discovery driver into `MultiRouter` (same for dataplane), and pass a (mutable) reference to `RaptorCast/Secondary.exec(&mut self, event, pdd)` This means RaptorCast implements a variant of `Executor` or `Stream` interface functions, which are not trait functions anymore

i took a stab at the refactor and it's too complicated - requiring a channel between MultiRouter and RaptorCast primary so polled msgs from peer discovery can get send out

still thinking what's a good interim solution

michael-yxchen: i thought about making a copy of peer discovery metrics and place that in RaptorCast. But when to make that copy becomes a problem. we can't make a copy in `metrics(&self)` because it takes an immutable self. Making copies in exec is too frequent

fabiotdf: How often is metrics() called?

eerkaijun: it's configurable, but generally pretty often (i.e. on testnet we take metrics once every second)

michael-yxchen: https://github.com/category-labs/monad-bft/commit/3fb23192fc0d4bb78ed37d82ed07fc37ca62b76f

temporary workaround. it's updated relatively infrequently (every 30s) I think grafana samples every 15s so it's probably ok

monad-raptorcast/src/raptorcast_secondary/publisher.rs
fabiotdf: Done

monad-raptorcast/src/raptorcast_secondary/publisher.rs
fabiotdf: Done - bumped to info

monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -446,16 +446,7 @@ where
...
-                                        pubkey: confirm_msg.peers[ii].pubkey(),
-                                        addr: rec.name_record.address,
-                                        signature: rec.signature,
-                                        record_seq_num: rec.name_record.seq,
-                                    };
-                                    peers.push(peer_entry);
-                                }
+                                // FIXME: bind name records with peers and ask peer discovery to verify
                                 for rec in &confirm_msg.name_records {
                                     if let Ok(node_id) = rec.recover_pubkey() {
```
fabiotdf: I think this leg containing `recover_pubkey` is the one that should be removed - because it's an expensive call that peer discovery already does

monad-raptorcast/src/raptorcast_secondary/mod.rs
fabiotdf: This made it into merge to master

monad-raptorcast/src/raptorcast_secondary/mod.rs
fabiotdf: SMH this note was hidden. Fixed in new upcoming PR

monad-raptorcast/src/raptorcast_secondary/mod.rs
fabiotdf: SMH this note was hidden. Fixed in new upcoming PR

monad-raptorcast/src/raptorcast_secondary/mod.rs
fabiotdf: SMH this note was hidden. Fixed in new upcoming PR

monad-raptorcast/src/raptorcast_secondary/mod.rs
fabiotdf: I think this made into the merge to master

monad-raptorcast/src/raptorcast_secondary/mod.rs
fabiotdf: Addressed in this PR:
https://github.com/category-labs/monad-bft/pull/1949

PR #1872, opened 2025-06-13
Simplify pacemaker

omegablitz:
Fold consensus.high_qc into pacemaker, and maintain high_certificate in pacemaker.

The only functional change is that Pacemaker::advance_epoch no longer exists. This would only be useful if we entered a round that’s after epoch_start_delay, but before the boundary block is committed. We don't support this use-case, so it's safe to delete.

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -25,6 +25,7 @@ use monad_consensus_types::{
     signature_collection::{SignatureCollection, SignatureCollectionKeyPairType},
     timeout::TimeoutCertificate,
     voting::Vote,
+    RoundCertificate,
 };
 use monad_crypto::certificate_signature::{
     CertificateSignaturePubKey, CertificateSignatureRecoverable,
@@ -62,8 +63,6 @@ where
     vote_state: VoteState<SCT>,
     /// Outgoing prepared votes to send to next leader
     scheduled_vote: Option<OutgoingVoteStatus>,
-    /// The highest QC (QC height determined by Round) known to this node
-    high_qc: QuorumCertificate<SCT>,
     /// Tracks and updates the current round
     pacemaker: Pacemaker<ST, SCT, EPT, CCT, CRT>,
     /// Policy for upholding consensus safety when voting or extending branches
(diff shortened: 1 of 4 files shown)


monad-consensus/src/pacemaker.rs
```diff
@@ -162,24 +170,43 @@ where
...
-        assert!(new_epoch > self.current_epoch || new_round > self.current_round);
+        if let RoundCertificate::Qc(qc) = &certificate {
+            if qc.get_round() > self.high_qc.get_round() {
+                self.high_qc = qc.clone();
+            }
+        }
+
+        if certificate.round() <= self.high_certificate.round() {
+            return Default::default();
+        }
```
michael-yxchen, line 187: what happens if both QC and TC are formed in a round? should QC take precedence over TC?

previous implementation would populate both `QC(r-1)` and `last_round_tc(r-1)` in a proposal, but `last_round_tc` is effectively skipped during safety check because QC takes priority in `safe_to_vote`

monad-consensus/src/pacemaker.rs
```diff
@@ -552,69 +521,79 @@ mod test {
...
+            cmds,
+            vec![
+                PacemakerCommand::EnterRound((Epoch(1), Round(2))),
+                PacemakerCommand::Schedule {
+                    duration: Duration::from_secs(3),
+                },
+            ]
+        );
 
         // in PhaseHonest::SuperMajority, pacemaker doesn't create TC with new timeouts
```
ariqchowdhury: nit: this variant was removed

monad-consensus/src/pacemaker.rs
omegablitz: discussed offline

monad-consensus-state/src/lib.rs
```diff
@@ -1227,15 +1205,26 @@ where
...
+        let last_round_tc = self.consensus.pacemaker.last_round_tc().cloned();
+
+        // make sure we haven't voted or timed out this round
+        if !self.consensus.safety.safe_to_vote(
+            round,
+            self.consensus.pacemaker.high_qc().get_round(),
+            &last_round_tc,
+        ) {
+            return cmds;
+        }
```
michael-yxchen, line 1223: will we ever hit this if clause under normal conditions or does it imply a logic error?

omegablitz: It happens on init, because we do that special-case thing where we immediately locally timeout the pacemaker's current round, but nothing stops us from creating a proposal after timing out.

This doesn't cause any safety issues afaik, but causes the out_of_order_proposals metric to be bumped because we can enter the next round via TC before we receive the proposal.

I think this could theoretically happen in "normal conditions" as well if we locally time out a round in which we're the leader due to lagged execution, but then execution catches up in the same round.

PR #1875, opened 2025-06-16
Add ST and EPT to TC

omegablitz:
This is in preparation for Tip being added to TC as part of the v2 changes. Tip contains an ST and EPT, just as ConsensusBlockHeader does.

### DIFF
--- a/monad-consensus/src/messages/message.rs
+++ b/monad-consensus/src/messages/message.rs
@@ -60,13 +60,23 @@ impl<SCT: SignatureCollection> VoteMessage<SCT> {
 /// The signature is a protocol signature,can be collected into the
 /// corresponding SignatureCollection type, used to create TC from the timeouts
 #[derive(Clone, Debug, PartialEq, Eq, RlpDecodable, RlpEncodable)]
-pub struct TimeoutMessage<SCT: SignatureCollection> {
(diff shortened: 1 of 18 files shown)


monad-consensus/src/messages/message.rs
```diff
@@ -110,18 +125,21 @@ where
...
+    };
     use monad_crypto::{certificate_signature::CertificateSignaturePubKey, NopSignature};
     use monad_multi_sig::MultiSig;
     use monad_testutil::signing::get_certificate_key;
     use monad_types::{Epoch, Round};
 
     use super::*;
 
-    type SignatureType = BlsSignature;
+    type SignatureType = NopSignature;
```
michael-yxchen: shall we use `SecpSignature` for serde tests? also not sure why we had MockSigs types

omegablitz: Fixed to use SecpSignature
Yeah I don't really know why the mocks are here.

omegablitz:
>I feel we don't need all those generic trait bounds in definition, but we can clean up later

Yeah we definitely don't need them everywhere, but I was too lazy to selectively add bounds in the places they were needed, so I just added them everywhere for consistency

PR #1876, opened 2025-06-16
Fix incremental virtual time in txpool

andr-dev:
When building the `TxHeap`, we used to set `virtual_time` to 0 during the constructor since `tracked_txs` is an `IndexMap` and thus produces txs in insertion order. However, `BinaryHeap` does not provide any promises about the order of elements with equal ordering so we must attach the `virtual_time` to enforce the ordering of these elements. Additionally, since `BinaryHeap` is a max heap, we must flip the ordering of the `virtual_time` comparison.

These two issues were unfortunately not detected previously because tests which enforced this class of semantics were not updated after being copied over from the previous implementation of the txpool which purely used insertion order. A test was modified to fully cover the semantics of the intended ordering for which the previous version does not pass.

### DIFF
--- a/monad-eth-txpool/tests/pool.rs
+++ b/monad-eth-txpool/tests/pool.rs
@@ -435,7 +435,7 @@ fn nontrivial_example() {
     let tx3 = make_legacy_tx(S1, 3 * BASE_FEE, GAS_LIMIT, 2, 10);
     let tx4 = make_legacy_tx(S2, 5 * BASE_FEE, GAS_LIMIT, 0, 10);
     let tx5 = make_legacy_tx(S2, 3 * BASE_FEE, GAS_LIMIT, 1, 10);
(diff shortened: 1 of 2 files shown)


monad-eth-txpool/tests/pool.rs
```diff
@@ -478,7 +478,7 @@ fn another_non_trivial_example() {
         TxPoolTestEvent::CreateProposal {
             tx_limit: 128,
             gas_limit: 1024 * GAS_LIMIT,
-            expected_txs: vec![&tx1, &tx5, &tx6, &tx2, &tx3, &tx4],
+            expected_txs: vec![&tx1, &tx5, &tx6, &tx3, &tx2, &tx4],
```
andr-dev, line 481: `tx2` and `tx3` have the same base fee and gas limit but `tx3` is the first tx for that signer in the block while `tx2` is the second.

PR #1890, opened 2025-06-20
Drop full node pings when above max active connections

eerkaijun:
Right now there is a high churn for full nodes because we unconditionally add nodes to our peer list when handling ping request, then prune them again periodically. This is not great as it means there's a high churn and lots of unnecessary signature verification. This PR probabilistically drop incoming pings when above max active connections. This is required rather than dropping all pings as we need validators to be able to add new full nodes in the network for full node broadcasting

More context here: https://github.com/category-labs/category-internal/issues/1619#issuecomment-2961743032

### DIFF
--- a/docker/devnet/monad/config/node.toml
+++ b/docker/devnet/monad/config/node.toml
@@ -47,6 +47,7 @@ request_timeout = 5
 prune_threshold = 5
 min_active_connections = 0
 max_active_connections = 200
+drop_full_nodes_ping_probability = 0.8
 
 [fullnode_dedicated]
 identities = []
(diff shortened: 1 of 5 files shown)


michael-yxchen:
why are we probabilistically dropping instead of not accepting at all? or we could put them in a queue so if we ever prune under max connections, we pop some incoming connections and connect (ping) to them

eerkaijun:
we need a way for full nodes that join the network later in time to still be able to peer with a validator. queueing full nodes can cause a full nodes to be stuck in the queue if the full nodes that join earlier never gets pruned. 

an example scenario below: 
- 2 validators and 2 full nodes. each validator has space to store 1 full node. 
- Full node A first joins the network, it sends ping to both validators, and gets added to the peer list of both validators
- Full node B joins the network later, it sends ping to both validators, gets added to the queue but will never get added to the peer list (as both the validators won't prune node A as it's not above the max active connections)

probabilistically dropping pings / pruning allows full nodes to join later to still be able to peer with a validator

eerkaijun:
discussed offline, gonna try another way

PR #1892, opened 2025-06-23
eip7702 validation support

mikety:
Consensus support for eip7702 validation. Implementing https://github.com/category-labs/category-internal/issues/1616#issuecomment-2981790796

Reserve balance only part is move to https://github.com/category-labs/monad-bft/pull/2026


monad-eth-block-policy/src/lib.rs
```diff
@@ -304,22 +322,41 @@ where
         &self,
         base_seq_num: SeqNum,
         eth_address: &Address,
+        is_delegated: bool,
     ) -> CommittedTxnFeeResult {
         let mut txn_fee: Balance = Balance::ZERO;
         let mut next_validate = base_seq_num + SeqNum(1);
+        let mut is_delegated = is_delegated;
+        let mut last_txn_seq_num = GENESIS_SEQ_NUM;
+        let EXEC_DELAY: u64 = 3;
```
mikety: TODO: read from chain config

monad-consensus-types/src/block.rs
```diff
@@ -278,7 +278,7 @@ where
             return Err(BlockPolicyError::ExecutionResultMismatch);
         }
 
-        Ok(())
+        Ok(block.clone())
```
andr-dev, line 281: Why did we make this change? It seems like we just moved the `.clone()` from the caller to the function and inherently forced every call to `check_coherency` to produce an owned validated block.

monad-eth-block-policy/src/lib.rs
```diff
@@ -130,6 +142,10 @@ struct BlockLookupIndex {
     round: Round,
     is_finalized: bool,
 }
+pub type TxnFee = (Balance, bool, SeqNum); // balance, is_delegated, as of seq_num
```
andr-dev: This should be a struct with named fields rather than having comments specify what each tuple field means. Also, is this a `TxnFee`? I would rename this to something along the lines of `AccountFeeState` since this type is related to accounts and `TxnFee` alone makes me think this should be a u128 or u256. That also means the below map alias could be renamed to `AccountFeeStateMap` or `AccountFeeStates` which makes sense given the map is from address to such state.

monad-eth-block-policy/src/lib.rs
```diff
@@ -304,22 +322,41 @@ where
         &self,
         base_seq_num: SeqNum,
         eth_address: &Address,
+        is_delegated: bool,
     ) -> CommittedTxnFeeResult {
         let mut txn_fee: Balance = Balance::ZERO;
         let mut next_validate = base_seq_num + SeqNum(1);
+        let mut is_delegated = is_delegated;
```
andr-dev: Let's just make the fn arg mut.

monad-eth-block-policy/src/lib.rs
```diff
@@ -558,9 +602,13 @@ where
...
                     mut next_validate,
-                } = self.committed_cache.compute_txn_fee(base_seq_num, address);
+                    mut is_delegated,
+                    mut last_txn_seq_num,
+                } = self
+                    .committed_cache
+                    .compute_txn_fee(base_seq_num, address, account_balance.1);
 
-                if account_balance < txn_fee_committed {
+                if account_balance.0 < txn_fee_committed {
```
andr-dev: Let's rename `account_balance` (based on naming of struct above) now that its type has changed.

monad-eth-txpool/src/pool/transaction.rs
```diff
@@ -17,6 +17,16 @@ use tracing::trace;
...
+pub fn get_apply_fn(
+    is_delegated: bool,
+) -> fn(&ValidEthTransaction, apply_value: Balance) -> Option<Balance> {
+    if is_delegated {
+        ValidEthTransaction::apply_gas_limit
+    } else {
+        ValidEthTransaction::apply_max_value
+    }
+}
+
```
andr-dev: Let's put this in `ValidEthTransaction` and make `apply_max_value` and `apply_gas_limit` private.

monad-eth-txpool/src/pool/transaction.rs
```diff
@@ -121,6 +150,10 @@ impl ValidEthTransaction {
         self.tx.length() as u64
     }
 
+    pub fn is_eip7702(&self) -> bool {
+        self.tx.is_eip7702()
+    }
```
andr-dev: Where is this function needed? In general, I don't want to expose information like this so that downstream users can't do things based on these values. The whole point of the `ValidEthTransaction` is to hide and encapsulate logic into the type so the pool doesn't have to "think" about them.

monad-consensus-types/src/block.rs
mikety: The intention is to modify the txn_fees in `check_coherency`, since they can be computed only with the history of the previous coherent blocks. It does an extra clone , how do you recommend to change it in the immutable & otherwise?

monad-triedb-utils/src/decode.rs
```diff
@@ -44,10 +44,18 @@ pub fn rlp_decode_account(account_rlp: Vec<u8>) -> Option<EthAccount> {
         }
     };
 
+    let is_delegated = false;
+    /*
+    let Ok(is_delegated) = bool::decode(&mut buf) else {
+        info!("rlp is_delegated decode failed: {:?}", buf);
+    };
+    */
```
eerkaijun: is this additional field going to be added to triedb? couldn't we just read the code field of the address to determine whether it's a 7702 delegated account? 7702 delegated account has `0xef0100 || address` set as the code

monad-eth-block-policy/src/lib.rs
```diff
@@ -73,6 +73,18 @@ pub fn compute_txn_max_value(txn: &TxEnvelope) -> U256 {
     txn_value.saturating_add(gas_cost)
 }
 
+pub fn compute_txn_carriage_cost(txn: &TxEnvelope) -> U256 {
+    U256::from(txn.gas_limit() as u128 * txn.max_fee_per_gas())
+}
```
eerkaijun, line 85: here we no longer includes txn.value() as part of the carriage cost. just wanted to confirm that when execution encounters a transaction from an undelegated account that doesn't have enough balance, it will also charge the carriage cost and revert the transaction? (i think this is the behavior for delegated account, just wanted to make sure it's also the same for undelegated account)

monad-eth-block-policy/src/lib.rs
```diff
@@ -745,33 +802,32 @@ where
                 .get_mut(&eth_address)
                 .expect("account_balances should have been populated");
 
-            let txn_fee = compute_txn_max_value(txn);
+            // if an account is not delegated and had no transactions in the last execution_delay blocks, then can charge into reserve.
+            let can_charge_into_reserve = account_balance.1
+                && block_seq_num.0.saturating_sub(account_balance.2 .0) > EXEC_DELAY;
```
eerkaijun: should there be a check here whether the account is delegated or not?

monad-eth-block-policy/src/lib.rs
```diff
@@ -780,8 +836,31 @@ where
...
+                    } else {
+                        is_delegated = true;
+                    }
+                }
+            }
+
+            txn_fee_entry.0 = txn_fee_entry.0.saturating_add(txn_fee);
+
+            txn_fee_entry.1 = is_delegated;
+            txn_fee_entry.2 = block_seq_num;
```
eerkaijun: i think the logic here is incorrect? here we store `is_delegated` to `eth_address`, but `eth_address` is the transaction signer address, not necessary the delegated address. shouldn't we store this to the delegated address? 

for context, the `address` in the `authorization_list` is the delegated address
```
authorization_list = [[chain_id, address, nonce, y_parity, r, s], ...]
```

monad-eth-block-policy/src/lib.rs
```diff
@@ -528,11 +567,12 @@ where
         state_backend: &impl StateBackend,
         extending_blocks: Option<&Vec<&EthValidatedBlock<ST, SCT>>>,
         addresses: impl Iterator<Item = &'a Address>,
-    ) -> Result<BTreeMap<&'a Address, Balance>, StateBackendError>
+    ) -> Result<BTreeMap<&'a Address, TxnFee>, StateBackendError>
     where
         SCT: SignatureCollection,
     {
         trace!(block = consensus_block_seq_num.0, "compute_base_balance");
+        let RESERVE_BALANCE: Balance = Balance::from(100_000_000_u64);
```
eerkaijun: this seems like a network parameter, maybe it should be moved to `monad-chain-config` eventually

eerkaijun: also this doesn't seem like it's used, I thought we need to check the transactions of the last k blocks does not exceed reserve balance?

monad-triedb-utils/src/decode.rs
mikety: Correct,  delegated is a part of the account code. is_delegated is just a placeholder until the execution branch is ready to be integrated.

monad-eth-block-policy/src/lib.rs
mikety: Correct, consensus relies on execution always maintain reserve balance, a transaction is allowed to transfer out of the reserve balance if and only if it's an EOA that is not delegated and there were no transactions for that account in the last `execution_delay`  blocks

monad-eth-block-policy/src/lib.rs
mikety: Reserve balance will be set by execution via precompile, so some users can modify it. Once execution branch is ready can add code to read it.

monad-eth-block-policy/src/lib.rs
mikety: changed the return type

monad-blocktree/src/tree.rs
```diff
@@ -50,6 +50,11 @@ where
         Some(())
     }
 
+    pub(crate) fn replace_coherent_block(&mut self, block: &BPT::ValidatedBlock) -> Option<()> {
```
andr-dev: Why is this producing an `Option<()>`? Doesn't seem to be used anywhere ...

monad-chain-config/src/revision.rs
```diff
@@ -39,20 +39,23 @@ pub struct ChainParams {
     pub proposal_gas_limit: u64,
     // Max proposal size in bytes (average transactions ~400 bytes)
     pub proposal_byte_limit: u64,
+    pub reserve_balance: u64,
```
andr-dev: What is this exactly? I'm guessing something along the lines of `account_reserve_balance_limit` or `max_account_reserve_balance`?

monad-consensus-types/src/block.rs
andr-dev: I'm not seeing any modification of `txn_fees` here. As this is currently implemented, anyone who calls `check_coherency` will clone the block regardless of whether they want to, and there's no logical reason to do this here rather than at the call site.

If this function is no longer checking the coherency, we should rename it, but that doesn't seem to be the case here.

monad-eth-block-policy/src/lib.rs
andr-dev: I meant there's no need to write `let mut is_delegated = is_delegated;` when you can specify `mut is_delegated: bool,` in the arglist.

monad-chain-config/src/revision.rs
mikety: it means max_reserve_balance, renamed.

monad-blocktree/src/tree.rs
mikety: Just to have a way to communicate an error/success. since `self.tree.get_mut(&block.get_id())` can fail

andr-dev: I know it looks kinda nasty but I think `Result<(), ()>` is cleaner here. `Option` implies "this fn might produce a value" while `Result<(), ()>` implies "this fn produces Ok(()) when it succeeds, and an error type of `()` on failure".

monad-eth-block-policy/src/lib.rs
```diff
@@ -304,21 +349,37 @@ where
...
     ) -> CommittedTxnFeeResult {
-        let mut txn_fee: Balance = Balance::ZERO;
+        let mut carriage_cost = Balance::ZERO;
+        let mut max_cost = Balance::ZERO;
         let mut next_validate = base_seq_num + SeqNum(1);
+        let mut last_txn_seq_num = GENESIS_SEQ_NUM;
+
+        // Compute balance from blocks before the base
+        let start_seq_num = next_validate.max(execution_delay) - execution_delay;
+        let mut last_txn_seq_num = GENESIS_SEQ_NUM;
```
andr-dev: Why do we redeclare this here?

monad-eth-block-policy/src/lib.rs
```diff
@@ -704,11 +841,27 @@ where
...
             .iter()
-            .map(|txn| txn.signer())
+            .map(|txn| {
+                if txn.is_eip7702() {
+                    if let Some(auth_list) = txn.authorization_list() {
+                        for delegated_addr in auth_list.iter().map(|a| *a.address()) {
+                            delegated_addresses.push(delegated_addr);
+                        }
+                    }
+                }
```
andr-dev: I think we can remove the `txn.is_eip7702()` check because `authorization_list` should produce `None` in all other cases.

monad-eth-block-policy/src/lib.rs
```diff
@@ -745,43 +934,73 @@ where
                 .get_mut(&eth_address)
                 .expect("account_balances should have been populated");
 
-            let txn_fee = compute_txn_max_value(txn);
-
-            if *account_balance >= txn_fee {
-                *account_balance -= txn_fee;
+            // if an account is not delegated and had no transactions in the last execution_delay blocks, then can charge into reserve.
+            let diff = block_seq_num.max(account_balance.last_txn_seqnum)
```
andr-dev: What does this diff represent? Let's use a more descriptive name.

monad-eth-block-policy/src/lib.rs
```diff
@@ -724,10 +877,46 @@ where
             tx_signers.iter(),
         )?;
 
+        // TODO check for empty acct here
+        let mut is_first_txn = true;
         for txn in block.validated_txns.iter() {
```
andr-dev: Let's use `.enumerate()` and `idx == 0` to check if its the first txn.

monad-eth-block-policy/src/lib.rs
```diff
@@ -131,6 +152,28 @@ struct BlockLookupIndex {
...
+#[derive(Debug, Clone)]
+pub struct AccountBalanceState {
+    pub balance: Balance,
+    pub reserve_balance: Balance,
+    pub is_delegated: bool,
+    pub last_txn_seqnum: SeqNum,
+}
+
+impl AccountBalanceState {
+    pub fn new_notdelegated() -> Self {
```
eerkaijun: nit: `new_not_delegated`

monad-eth-block-policy/src/lib.rs
mikety: I think keeping it improves clarity - since authorization_list is specific for eip7702, so the reader gets the right context. Also it ensures if there is any other eip in the future that has authorization list it will not be used.

monad-eth-block-policy/src/lib.rs
```diff
@@ -304,21 +349,36 @@ where
...
         let mut next_validate = base_seq_num + SeqNum(1);
+        let mut last_txn_seq_num = GENESIS_SEQ_NUM;
+
+        // Compute balance from blocks before the base
+        let start_seq_num = next_validate.max(execution_delay) - execution_delay;
+        for (&cache_seq_num, block) in self.blocks.range(start_seq_num..next_validate) {
+            if block.fees.get(eth_address).is_some() {
+                last_txn_seq_num = cache_seq_num;
+            }
+        }
```
eerkaijun, line 365: i didn't quite get this part - so we want to update `last_txn_seq_num` if the account is touched in blocks in the execution delay period, why do we need that?

monad-eth-block-validator/src/lib.rs
```diff
@@ -129,9 +130,44 @@ where
...
+                .saturating_add(compute_txn_carriage_cost(eth_txn));
+            txn_fee_entry.max_cost = txn_fee_entry
+                .max_cost
+                .saturating_add(compute_txn_max_value(eth_txn));
+
+            if eth_txn.is_eip7702() {
+                if let Some(auth_list) = eth_txn.authorization_list() {
+                    let authorities = auth_list
+                        .iter()
+                        .map(|a| (*a.address(), a.nonce(), a.address))
```
eerkaijun: i think this is incorrect - `a.address()` is the address of the authorized smart contract, not the signer I believe.
the signer needs to be obtained through the recover_authority() function in alloy

monad-eth-block-validator/src/lib.rs
```diff
@@ -129,9 +130,44 @@ where
...
+                        .collect_vec();
+                    for (signer, nonce, address) in authorities {
+                        if let Some(n) = nonces.get(&signer) {
+                            if *n != nonce {
+                                return Err(BlockValidationError::TxnError);
+                            }
+                        }
```
eerkaijun: I thought the nonce of authority should also be incremented by one? i.e. we should check n+1 == nonce

eerkaijun: also we would also need to handle the case for zero nonce. the nonce logic that we use above should be applicable
```
            let maybe_old_nonce = nonces.insert(eth_txn.signer(), eth_txn.nonce());
            // txn iteration is following the same order as they are in the
            // block. A block is invalid if we see a smaller or equal nonce
            // after the first or if there is a nonce gap
            if let Some(old_nonce) = maybe_old_nonce {
                if eth_txn.nonce() != old_nonce + 1 {
                }
            }
```

monad-eth-block-validator/src/lib.rs
```diff
@@ -129,9 +130,44 @@ where
...
+                        nonces
+                            .entry(signer)
+                            .and_modify(|n| *n += 1)
+                            .or_insert(nonce + 1);
+                        if address == Address::ZERO { // undelegate
```
eerkaijun: is it expected that nothing happens during undelegation?

monad-eth-txpool/src/pool/mod.rs
```diff
@@ -144,9 +145,27 @@ where
...
 
-        let addresses = txs.iter().map(ValidEthTransaction::signer).collect_vec();
+        // enrich with delegated EOAs
+        let mut delegated_addresses: Vec<(Address, Address)> = Vec::new();
+        let mut addresses = txs
+            .iter()
+            .map(|txn| {
+                if txn.raw().is_eip7702() {
+                    if let Some(auth_list) = txn.raw().authorization_list() {
+                        for delegated_addr in auth_list.iter().map(|a| (*a.address(), a.address)) {
```
eerkaijun: a.address() and a.address corresponds to the same thing I think, which is the authorized smart contract address

eerkaijun:
can we also add unit tests for 7702 transactions in `monad-eth-block-policy` and `monad-eth-block-validator` crates? that should give us more confidence

monad-eth-block-policy/src/lib.rs
mikety: To validate account withdrawals per https://github.com/category-labs/category-internal/issues/1616#issuecomment-2981790796 . Consensus tracks when last transaction was (i.e. account was touched)

monad-eth-block-validator/src/lib.rs
mikety: Per the Behavior section of the EIP7702:
`Verify the nonce of authority is equal to nonce.` , then run checks and set the code and then `Increase the nonce of authority by one.`
The increment is done on the following line (159 in this commit)

monad-eth-block-validator/src/lib.rs
mikety: I believe execution need to process it, the consensus just needs to know that the account is not delegated.

PR #1893, opened 2025-06-23
Add high_certificate to checkpoint

omegablitz:


monad-executor-glue/src/lib.rs
```diff
@@ -123,10 +123,15 @@ where
...
 #[derive(Clone)]
-pub struct CheckpointCommand<SCT: SignatureCollection> {
+pub struct CheckpointCommand<ST, SCT, EPT>
+where
+    ST: CertificateSignatureRecoverable,
+    SCT: SignatureCollection<NodeIdPubKey = CertificateSignaturePubKey<ST>>,
+    EPT: ExecutionProtocol,
+{
     pub root_seq_num: SeqNum,
     pub high_qc_round: Round,
```
michael-yxchen: nit: we can have multiple forkpoints with same `high_qc_round` but different `high_certificate`. maybe we need both `high_qc_round` and `high_certificate_round` in the filename now?

monad-state/src/lib.rs
```diff
@@ -236,12 +275,49 @@ impl<SCT: SignatureCollection> Forkpoint<SCT> {
...
+            .create(vset_stake)
+            .expect("ValidatorSetTypeFactory failed to init validator set");
+
+        // 2.
+        if verify_tc(&qc_vset, &qc_vmap, tc).is_err() {
+            return Err(ForkpointValidationError::InvalidHighCertificate);
+        }
+
+        Ok(())
+    }
```
michael-yxchen: nit: qc -> tc

monad-executor-glue/src/lib.rs
omegablitz: Implemented, contained in separate commit. Will squash after reviews before merging.

michael-yxchen:
note to update [forkpoint.genesis.toml](https://github.com/category-labs/monad-bft/blob/master/docker/devnet/monad/config/forkpoint.genesis.toml) with v2

Issue #1895, opened 2025-06-24
raptorcast: fail to encode hard-to-compress messages


robertkiel:
**Context**

Raptorcast uses compression to minimize the data transferred to other clients. In general, compression shrinks the data size, i.e. less data gets transferred. However, in case of short messages or hard-to-compress messages, the size might increase.

**Issue**

Before compressing the message, the code [allocates](https://github.com/category-labs/monad-bft/blob/d7159a590d89c0d8a39e70d5b5ce011918f34bbc/monad-raptorcast/src/message.rs#L87) memory whose size is bounded by the RLP-encoded length of the message.

However, in the probably rare case that messages cannot be compressed well, the compression will need more memory than the size of the message, which will make the [compression](https://github.com/category-labs/monad-bft/blob/d7159a590d89c0d8a39e70d5b5ce011918f34bbc/monad-raptorcast/src/message.rs#L88) fail because it exceeds the upper bound of `BoundedWriter`. Although this does not provoke a panic, it prevents the network from making progress, i.e. producing blocks.

**Suggestion**

Use [`zstd::stream::encode_all`](https://docs.rs/zstd/0.13.3/zstd/stream/fn.encode_all.html) which will automatically allocate the right amount of memory - although it might require slightly more time.

**Sample**

```rust
#[derive(RlpEncodable)]
struct MyAppMessage {
    data: Vec<u8>,
}

let msg = MyAppMessage { data: vec![2] };

let mut msg_encoded = BytesMut::new();
// RLP encoding
msg.encode(&mut msg_encoded);

let msg_encoded_len = msg_encoded.len();

let mut compression_writer = BoundedWriter::new(msg_encoded_len);
ZstdCompression::default()
    .compress(&app_encoded, &mut compression_writer)
    .expect("compression failed"); // this will panic
```

dshulyak:
i think it makes more sense to send uncompressed in such case

robertkiel:
> i think it makes more sense to send uncompressed in such case

I agree. However, the logic to automatically send uncompressed messages is missing and this could lead to outages

dshulyak:
> and this could lead to outages

can you elaborate on what can lead to outages? the problem in general or sending uncompressed data if compression "fails" ?
regarding the logic i think we can just fallback to previous encoding format, unless there was a plan to deprecate it, need to ask about that

robertkiel:
Regarding outages:
[...]

PR #1896, opened 2025-06-24
monad-raptorcast: fallback to uncomressed if zstd fails to compress

dshulyak:
primarily to avoid dropping message if compressed size is larger than uncompressed.
added a warn log as i generally don't expect messages to be uncompressable, so it should bring some awareness if it happens

closes: https://github.com/category-labs/monad-bft/issues/1895

### DIFF
--- a/monad-raptorcast/src/message.rs
+++ b/monad-raptorcast/src/message.rs
@@ -61,14 +61,14 @@ pub enum OutboundRouterMessage<OM, ST: CertificateSignatureRecoverable> {
     FullNodesGroup(FullNodesGroupMessage<ST>),
 }
 
-#[derive(Debug)]
-pub struct SerializeError(pub String);
-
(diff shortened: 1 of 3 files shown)


monad-raptorcast/src/message.rs
```diff
@@ -61,14 +61,23 @@ pub enum OutboundRouterMessage<OM, ST: CertificateSignatureRecoverable> {
...
-        let mut buf = BytesMut::new();
-
         let version = NetworkMessageVersion::version();
+        self.try_serialize_with_version(version)
+    }
+
+    fn try_serialize_with_version(
+        &self,
+        version: NetworkMessageVersion,
+    ) -> Result<Bytes, SerializeError> {
```
dshulyak: actually it is non-fallible with fallback

PR #1897, opened 2025-06-24
Implement monad-event-ring library

andr-dev:


monad-event-ring/src/reader/mod.rs
```diff
@@ -0,0 +1,41 @@
+use std::marker::PhantomData;
+
+pub(crate) use self::raw::RawEventReader;
+use crate::{EventDescriptor, EventNextResult, EventRingType};
+
+mod raw;
+
+/// An iterator over events in an [`EventRing`](crate::EventRing).
```
JoeHowarth: Nit: Don't call it an iterator if it doesn't impl Iterator

monad-event-ring/src/descriptor/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+{
+    pub(crate) fn new(raw: RawEventDescriptor<'ring, 'reader>) -> Self {
+        Self {
+            raw,
+            _phantom: PhantomData,
+        }
+    }
+}
+
+impl<'ring, 'reader, T> EventDescriptor<'ring, 'reader, T>
```
JoeHowarth: Nit: Why is this a separate impl block?

monad-event-ring/src/type.rs
```diff
@@ -0,0 +1,112 @@
...
+    /// [`EventReader`](crate::EventReader) have the same underlying ABI as the event ring writer.
+    fn ring_metadata_hash() -> &'static [u8; 32];
+
+    /// Used to check that the event ring matches this [`EventRingType`].
+    fn check_ring_type(c_event_ring: &monad_event_ring) -> Result<(), String> {
+        monad_event_ring_check_type(c_event_ring, Self::ring_ctype(), Self::ring_metadata_hash())
+    }
+
+    /// The flow info of this event ring type.
+    type FlowInfo;
```
JoeHowarth, line 52: This needs more description. Consider an example or how this relates to the other types

monad-event-ring/src/result.rs
```diff
@@ -0,0 +1,76 @@
+use crate::{
+    ffi::{
+        self, monad_event_descriptor, monad_event_iterator_reset, monad_event_iterator_try_next,
+        monad_event_next_result,
+    },
+    RawEventDescriptor, RawEventReader,
+};
+
+/// The result of attempting to retrieve the next event from an [`EventRing`](crate::EventRing).
+pub enum EventNextResult<T> {
```
JoeHowarth, line 10: If you squint, this looks like Poll from std::future::Future 

https://doc.rust-lang.org/std/future/trait.Future.html
```rust
pub enum Poll<T> {
    Ready(T),
    Pending,
}
```

So

```rust
pub type EventNextResult<T> = Poll<Result<T, GapError>>;
pub struct GapError{};
```

I noticed that you mentioned Iterator above and we've talked about that on zoom and I know you're using that just as a shorthand, but we may be able to re-encode this system as a Stream instead since fundamentally it is an async stream with custom error behavior, all of which can be modeled by the Stream trait afaik

monad-event-ring/src/result.rs
```diff
@@ -0,0 +1,76 @@
...
+            EventNextResult::NotReady => EventNextResult::NotReady,
+            EventNextResult::Gap => EventNextResult::Gap,
+        }
+    }
+}
+
+/// The result of attempting to read the payload from an
+/// [`EventDescriptor`](crate::EventDescriptor).
+#[derive(Debug)]
+pub enum EventDescriptorPayload<T> {
```
JoeHowarth, line 59: I'd also remove this enum and use Result + Error instead:
```rust
pub type EventDescriptorPayload<T> = Result<T, PayloadExpired>`
struct PayloadExpired {}
impl std::error::Error for PayloadExpired {..}
```

monad-event-ring/src/lib.rs
```diff
@@ -0,0 +1,75 @@
...
+//! implement a recovery phase which should be triggered when the event ring gaps. This mechanism
+//! should be thoroughly tested, likely using a [`SnapshotEventRing`] in tests, to ensure smooth
+//! operation in the unlikely event the reader falls behind.
+//!
+//! **Failing to recover / backfill missing events after an [`EventNextResult::Gap`] and/or ignoring
+//! the variant will most likely lead to state inconsistency.**
+//!
+//! </div>
+//!
+//! Unlike [`TypedEventRing`]s, [`EventReader`]s are single threaded as they use the lifetime
```
JoeHowarth, line 65: Is this true? RawEventReader has Send + Sync impl'd and iirc that means that Send + Sync are auto impl'd for EventReader as a consequence

monad-event-ring/src/descriptor/mod.rs
andr-dev: Good catch, that was the remains of when these fns were a trait impl.

monad-event-ring/src/lib.rs
andr-dev: Good catch, `RawEventReader` should only implement `Send`. Fixed!

monad-event-ring/src/result.rs
JoeHowarth: Discussed on slack and using stream likely makes sense if the lib exposes a high-level api that handles polling, but not at this level of the api.

monad-event-ring/src/result.rs
JoeHowarth: Or actually just Option

PR #1900, opened 2025-06-25
Add block builders to exec events library

andr-dev:


monad-exec-events/src/block.rs
```diff
@@ -0,0 +1,304 @@
...
+        }
+    }
+
+    /// FIX THIS
+    pub fn to_alloy_rpc_header(&self) -> alloy_rpc_types::Header {
+        alloy_rpc_types::Header {
+            hash: alloy_primitives::FixedBytes(self.execution_result.eth_block_hash.bytes),
+            inner: self.to_alloy_header(),
+            total_difficulty: Some(alloy_primitives::U256::ZERO),
+            size: None,
```
philipglazman: Just making a note that users never get `size`

monad-exec-events/src/block_builder/commit_state/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+
+                let (block, state) = self
+                    .state
+                    .remove(&proposal)
+                    .expect("CommitStateBlockBuilder added proposal to proposals but not state");
+
+                if state != BlockCommitState::Finalized {
+                    warn!(?state, "CommitStateBlockBuilder received verified block number for non-finalized proposal");
+                }
+                // assert_eq!(state, BlockCommitState::Finalized, "CommitStateBlockBuilder received verified block number for non-finalized proposal");
```
philipglazman: unnecessary assert?

monad-exec-events/src/block_builder/executed/mod.rs
```diff
@@ -0,0 +1,367 @@
...
+                    output: None,
+                    call_frames: self.include_call_frames.then(Vec::default),
+                });
+
+                None
+            }
+            ExecEvents::TxnReject { txn_index, reject } => {
+                let state = self.state.as_mut()?;
+
+                unimplemented!("TxnReject {txn_index} {reject}\n{state:#?}");
```
philipglazman: Panic is probably ok here since my understanding is that execution will also crash. However, we might want to give the user the option to gracefully exit.

monad-exec-events/src/block.rs
andr-dev: I'll add a TODO to revisit later like I did with `removed`.

monad-exec-events/src/block.rs
```diff
@@ -0,0 +1,296 @@
...
+            }
+            _ => panic!(
+                "EventServer encountered unknown tx type {}",
+                self.header.txn_type
+            ),
+        }
+    }
+
+    /// Creates a recovered alloy tx envelope.
+    pub fn to_alloy_recocered(
```
kjcamann: Typo

monad-exec-events/src/block.rs
```diff
@@ -0,0 +1,296 @@
...
+        self.logs.iter().map(ExecutedTxnLog::to_alloy).collect_vec()
+    }
+}
+
+/// Transaction log reconstructed from execution events.
+#[allow(missing_docs)]
+#[derive(Clone, Debug)]
+pub struct ExecutedTxnLog {
+    pub address: monad_c_address,
+    pub topic: Box<[u8]>,
```
kjcamann: Shouldn't this by some 256 bit type?

monad-exec-events/src/block_builder/commit_state/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+
+/// Represents a change in a block's commit state.
+#[derive(Debug)]
+pub struct CommitStateBlockUpdate<Block = ExecutedBlock> {
+    /// The block who's commit state changed.
+    pub block: Arc<Block>,
+    /// The new commit state.
+    pub state: BlockCommitState,
+    /// Blocks that were abandoned as a result of the block commit change.
+    pub abandoned: Option<Vec<Arc<Block>>>,
```
kjcamann: Why is this optional rather than there just being zero element in the vector when nothing is abandoned?

monad-exec-events/src/block_builder/commit_state/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+                            .is_none_or(|proposals| !proposals.contains(&proposal_metadata.id)),
+                        "CommitStateBlockBuilder received qc for block in proposals but not in state"
+                    );
+                    return None;
+                };
+
+                if state != &BlockCommitState::Proposed {
+                    warn!(?state, "CommitStateBlockBuilder received voted proposal id for non-proposed proposal");
+                }
+                *state = BlockCommitState::Voted;
```
kjcamann: I'm not sure how it will work in BFTv2 yet, but in the old flow of execution `BLOCK_QC` is recorded every time a QC is seen in a new proposal. So if you keep getting new proposals (that fail), it will keep re-recording the `BLOCK_QC` of the parent block, effectively re-announcing the QC again and again. So you're likely to see this warning and it's not a big deal.

At least I think that is how it will play out, not 100% familiar with the new consensus algorithm yet.

You probably don't want to keep telling them about the update every time though, just the first time.

monad-exec-events/src/block_builder/commit_state/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+                let abandoned = proposals
+                    .drain()
+                    .filter(|block_id| block_id != &proposal_metadata.id)
+                    .map(|abandoned| {
+                        self.state
+                            .remove(&abandoned)
+                            .expect("CommitStateBlockBuilder added block to proposals and blocks")
+                            .0
+                    })
+                    .collect::<Vec<_>>();
```
kjcamann: QC does not cause abandonment of proposals, only finalization does. Abandonment is making explicit the idea that the finalization of a candidate to become block `B` implicitly abandons competing candidates to become this same block. The implicit abandonment is tracked explicitly so that the user can discard their associated state.

I don't know if you are changing the definition because of the BFTv2 tail forking thing or just a copy/paste error?

monad-exec-events/src/block_builder/commit_state/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+            ExecEvents::BlockQC(proposal_metadata) => {
+                let Some((executed_block, state)) = self.state.get_mut(&proposal_metadata.id)
+                else {
+                    assert!(
+                        self.proposals
+                            .get(&proposal_metadata.block_number)
+                            .is_none_or(|proposals| !proposals.contains(&proposal_metadata.id)),
+                        "CommitStateBlockBuilder received qc for block in proposals but not in state"
+                    );
+                    return None;
```
kjcamann: Unless you're doing some kind of check before this (I didn't check in detail), or I am not understanding what this does, it's not correct to say we won't see finalization and QC events for proposals whose execution we never saw.

It's actually pretty common that we will, when first starting the program "in the middle" of stuff happening, or when execution resumes (usually the very first event is `BLOCK_QC`, the announcement of the QC in the proposal header for the previous block). The original implementation returned these in this case:

```
/// Returned when the user attempts to update the consensus state of
/// a proposal that was never added
UnknownProposal {
    block_number: u64,
    block_id: MonadBlockId,
    consensus_state: ConsensusState,
},
```

monad-exec-events/src/block_builder/commit_state/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+            ExecEvents::BlockFinalized(proposal_metadata) => {
+                let Some((executed_block, state)) = self.state.get_mut(&proposal_metadata.id)
+                else {
+                    assert!(
+                        self.proposals
+                            .get(&proposal_metadata.block_number)
+                            .is_none_or(|proposals| !proposals.contains(&proposal_metadata.id)),
+                        "CommitStateBlockBuilder received finalized for block in proposals but not in state"
+                    );
+                    return None;
```
kjcamann: Same comment as for QC

monad-exec-events/src/block_builder/commit_state/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+                };
+
+                if state != &BlockCommitState::Voted {
+                    warn!(?state, "CommitStateBlockBuilder received finalized proposal id for non-voted proposal");
+                }
+                *state = BlockCommitState::Finalized;
```
kjcamann: _Unlike_ for QC, this is fine and you should not see multiple of these.

monad-exec-events/src/block_builder/commit_state/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+                };
+
+                if state != &BlockCommitState::Proposed {
+                    warn!(?state, "CommitStateBlockBuilder received voted proposal id for non-proposed proposal");
+                }
+                *state = BlockCommitState::Voted;
+
+                let (block, state) = (executed_block.clone(), *state);
+
+                let proposals = self
```
kjcamann: Calling the map of block number -> set of proposals `proposals` (as in `self.proposals`) and then also calling the individual set you retrieve `proposals` is confusing.


This might be better as `let candidates = ` as these are specific candidate proposals to become the particular block number.

monad-exec-events/src/block.rs
JoeHowarth: nit: typo

monad-exec-events/src/block_builder/commit_state/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+                *state = BlockCommitState::Finalized;
+
+                let (block, state) = (executed_block.clone(), *state);
+
+                let proposals = self
+                    .proposals
+                    .get_mut(&proposal_metadata.block_number)
+                    .expect("CommitStateBlockBuilder added finalized block to state and proposals");
+                assert!(proposals.contains(&proposal_metadata.id),
+                 "CommitStateBlockBuilder received finalized for block in state but not in proposals");
```
kjcamann: Same comment as for QC, this is likely to happen on startup if I've understood it correctly

monad-exec-events/src/block_builder/commit_state/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+                        abandoned: None,
+                    }
+                }
+            }
+        })
+    }
+}
+
+impl<Block> CommitStateBlockBuilder<Block> {
+    fn select_event_ref(event_ref: ExecEventsRef<'_>) -> Option<ExecEvents> {
```
JoeHowarth: nit: consider a more descriptive name that describes what types of events we're selecting. Perhaps `select_block_event_refs` or smt like that

monad-exec-events/src/block_builder/commit_state/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+                            .0
+                    })
+                    .collect::<Vec<_>>();
+
+                if proposals.insert(proposal_metadata.id) {
+                    warn!(
+                        id =? proposal_metadata.id,
+                        "CommitStateBlockBuilder readded finalized proposal to proposals"
+                    );
+                }
```
kjcamann: Doesn't the above assertion (which I think is incorrect and too aggressive) prevent this from happening?

monad-exec-events/src/block_builder/commit_state/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+    }
+
+    #[test]
+    fn basic_test_monad_testnet() {
+        const SNAPSHOT_NAME: &str = "MONAD_DEVNET_21B_GENESIS";
+        const SNAPSHOT_ZSTD_BYTES: &[u8] =
+            include_bytes!("../../../test/data/exec-events-mdn-65b-genesis.zst");
+
+        run_commit_state_block_builder(SNAPSHOT_NAME, SNAPSHOT_ZSTD_BYTES);
+    }
```
kjcamann: How does this pass, since this file is broken?

monad-exec-events/src/block_builder/commit_state/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+                } => panic!("implicit drop: {reassembly_error:?}"),
+                BlockBuilderResult::Ok(CommitStateBlockUpdate {
+                    block,
+                    state,
+                    abandoned,
+                }) => {
+                    eprintln!("[{}] -> {:?}", block.header.proposal.block_number, state);
+
+                    if let Some(abandoned) = abandoned {
+                        panic!("abandoned: {abandoned:#?}");
```
kjcamann: Why does this panic? There is nothing wrong with abandonment, it's part of the circle of life.

monad-exec-events/src/block.rs
JoeHowarth: There are 4 256bit topics. Can you add a comment describing how this is laid out or change the repr to be more descriptive?

monad-exec-events/src/block_builder/commit_state/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+    Voted,
+    /// The next block has been voted on affirmatively by a supermajority leading to finality of the current block.
+    Finalized,
+    /// The block's execution outputs and state root have been included in a finalized block.
+    Verified,
+}
+
+/// Represents a change in a block's commit state.
+#[derive(Debug)]
+pub struct CommitStateBlockUpdate<Block = ExecutedBlock> {
```
JoeHowarth: What is the use case for the generic here? What does it allow the consumer to do that they couldn't without it?

monad-exec-events/src/block_builder/executed/mod.rs
```diff
@@ -0,0 +1,367 @@
...
+            include_call_frames,
+        }
+    }
+
+    /// Processes the execution event in the provided event descriptor.
+    pub fn process_event_descriptor<'ring, 'reader>(
+        &mut self,
+        event_descriptor: &EventDescriptor<'ring, 'reader, ExecEventRingType>,
+    ) -> Option<BlockBuilderResult<ExecutedBlock>> {
+        match event_descriptor.try_filter_map(if self.include_call_frames {
```
JoeHowarth, line 35: 1. Add comment describing this flow
2. Consider changing `select_event_ref` to be more descriptive. What is it selecting? How does that fit with the logic in this file?

monad-exec-events/src/block_builder/executed/mod.rs
JoeHowarth: Second this - archive never wants to crash for example. More generally, we should really avoid panics in sdks whenever possible and bubble up a typed error.

monad-exec-events/src/block_builder/result.rs
```diff
@@ -0,0 +1,88 @@
+use crate::ffi::monad_exec_block_header;
+
+/// Result produced by [`ExecutedBlockBuilder`](crate::ExecutedBlockBuilder) while attempting to
+/// reassemble a block from execution events.
+///
+/// This type is generic so it can be shared with
+/// [`CommitStateBlockBuilder`](crate::CommitStateBlockBuilder) to produce reassembled block updates
+/// along with the block's commit state.
+pub enum BlockBuilderResult<T> {
```
JoeHowarth: Same comment as event-ring pr:
Consider using a Result<T, BlockBuilderError> instead.

If you really want to go this route instead, please add in-code documentation (not necessarily rustdoc) explaining why we're using a custom return enum instead of the more expected Result + error enum pattern so that future maintainers understand the design choice.

monad-exec-events/src/block.rs
andr-dev: Agreed with both, I originally did the conversion in the alloy type but I should just do it while building the txn log.

monad-exec-events/src/block_builder/commit_state/mod.rs
andr-dev: Minor optimization, `Vec` always incurs an allocation while `Option` doesn't. Can change if we don't like it.

monad-exec-events/src/block_builder/commit_state/mod.rs
andr-dev: Whenever we add a block, we add it to both `self.state` and `self.proposals` so this assertion simply checks that if it's not present in `self.state` then it isn't present in `self.proposals` either. When we start "in the middle", those other proposals will be ignored. I can add the event from above, although I didn't end up needing it elsewhere since there's little you can do without the block itself.

monad-exec-events/src/block_builder/commit_state/mod.rs
andr-dev: The tests didn't have any abandonments so I added the panic so that when they do I can revisit how to "verify" that they are ok.

monad-exec-events/src/block_builder/commit_state/mod.rs
andr-dev: It doesn't, I was going to ask for new files :)

monad-exec-events/src/block_builder/executed/mod.rs
andr-dev: Agreed with the above sentiment of not panicking in an SDK, hence why I picked `unimplemented` which is semantically different from a `panic` in that it's supposed to become implemented in the future.

monad-exec-events/src/block_builder/result.rs
andr-dev: Agreed, here I think it's more correct to use `Result`.

monad-exec-events/src/block_builder/commit_state/mod.rs
andr-dev: The idea here is that the logic contained in this struct isn't specific to attaching commit state info to `ExecutedBlock`s. You could instead specify some other type that you want to attach commit state info to, and I expect this to be a common use case.

Example: Trading application wants all uniswap logs by commit state, when they receive the proposed they start doing processing, when they get the finalized (or soon verified) they send txs to execute against that info.

Since this isn't exactly clear here, I'm going to remove it for now since migrating to include this generic down the line would be pretty simple. I thought it would be nice to have it already but let's keep things simple :)

JoeHowarth: Yeah I'd either have a test/example proving that a consumer can do this with the current public api or keep it out for now

monad-exec-events/src/block_builder/executed/mod.rs
JoeHowarth: Add as an error type instead?

monad-exec-events/src/block_builder/commit_state/mod.rs
andr-dev: This was a misunderstanding on my end, agreed there are no abandonments from QCs.

monad-exec-events/src/block_builder/commit_state/mod.rs
JoeHowarth: Empty Vec doesn't allocate, or are you saying something else?
https://doc.rust-lang.org/std/vec/struct.Vec.html#guarantees

andr-dev: !!! I was not aware, fantastic will change.

monad-exec-events/src/block_builder/executed/mod.rs
andr-dev: Changed it to produce the `Rejected` error for now, will do more precise error handling later to differentiate the different error cases.

PR #1903, opened 2025-06-25
Consolidate certificate verification

omegablitz:
This is in preparation for v2, which adds yet another certificate type.

### DIFF
--- a/monad-consensus/src/validation/signing.rs
+++ b/monad-consensus/src/validation/signing.rs
@@ -14,7 +14,7 @@ use monad_crypto::{
     },
     hasher::{Hash, Hashable, Hasher, HasherType},
 };
-use monad_types::{ExecutionProtocol, NodeId, Stake, GENESIS_ROUND};
+use monad_types::{Epoch, ExecutionProtocol, NodeId, Round, Stake, GENESIS_ROUND};
 use monad_validator::{
     epoch_manager::EpochManager,
     validator_set::{ValidatorSetType, ValidatorSetTypeFactory},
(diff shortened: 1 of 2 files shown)


monad-consensus/src/validation/signing.rs
```diff
@@ -328,10 +328,14 @@ where
     {
         self.well_formed_proposal()?;
         self.verify_epoch(epoch_manager)?;
-        verify_certificates(
-            epoch_manager,
-            val_epoch_map,
-            &self.obj.last_round_tc,
+        if let Some(tc) = &self.obj.last_round_tc {
+            verify_tc(
+                |epoch, round| epoch_to_validators(epoch_manager, val_epoch_map, epoch, round),
```
andr-dev, line 333: Super nit but seeing this pattern repeated in a bunch of places along with the fns having the `impl FnOnce(...)`, might want to add a type like the `ChildState` that bundles it all up nicely or smth.

monad-consensus/src/validation/signing.rs
```diff
@@ -441,63 +449,52 @@ where
     }
 }
 
-fn verify_certificates<ST, SCT, EPT, VTF, VT>(
-    epoch_manager: &EpochManager,
-    val_epoch_map: &ValidatorsEpochMapping<VTF, SCT>,
-    tc: &Option<TimeoutCertificate<ST, SCT, EPT>>,
-    qc: &QuorumCertificate<SCT>,
-) -> Result<(), Error>
+fn epoch_to_validators<'a, SCT, VTF, VT>(
```
andr-dev, line 452: I think you can use `'_` to elide the lifetime and remove it from the generics part.

monad-state/src/lib.rs
```diff
@@ -230,94 +229,52 @@ where
...
+                        .collect::<Vec<_>>(),
+                );
+                (locked.epoch, (vset, vmap))
+            })
+            .collect::<HashMap<_, _>>();
+        let epoch_to_validators = |epoch, _round| {
+            let (vset, vmap) = validators
+                .get(&epoch)
+                .ok_or(validation::Error::ValidatorSetDataUnavailable)?;
+            Ok((vset, vmap))
```
andr-dev, line 259:
```rust
Ok(validators.get(&epoch).ok_or(validation::Error::ValidatorSetDataUnavailable)?)
```

omegablitz: It needs to return a (&A, &B) and not &(A, B)

monad-consensus/src/validation/signing.rs
omegablitz: There are two input lifetimes so it can't be elided afaik

monad-consensus/src/validation/signing.rs
omegablitz: I'll do this later cuz I'm lazy

PR #1904, opened 2025-06-25
Blocksync from high_qc on init

omegablitz:
Currently, consensus-state only decides to blocksync on the first QC it observes above high_qc. This commit changes this behavior to always blocksync root -> high_qc. This is important in the case of a chain halt, because new QCs will not be produced.

This is testable in the mock-swarm forkpoint tests by restarting all nodes, but deferring this for later.

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -279,6 +279,35 @@ where
     pub fn get_current_round(&self) -> Round {
         self.pacemaker.get_current_round()
     }
+
+    #[must_use]
+    pub fn request_blocks_if_missing_ancestor(
+        &mut self,
+    ) -> Vec<ConsensusCommand<ST, SCT, EPT, BPT, SBT>> {
+        let high_qc = self.get_high_qc();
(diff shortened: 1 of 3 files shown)


monad-consensus-state/src/lib.rs
```diff
@@ -279,6 +279,37 @@ where
     pub fn get_current_round(&self) -> Round {
         self.pacemaker.get_current_round()
     }
+
+    #[must_use]
+    pub fn request_blocks_if_missing_ancestor(
+        &mut self,
+    ) -> Vec<ConsensusCommand<ST, SCT, EPT, BPT, SBT>> {
+        let high_qc = self.get_high_qc();
+        let root_seq_num = self.pending_block_tree.get_root_seq_num();
```
ariqchowdhury: this is unused?

omegablitz: Yes, I'll remove it. It was previously used to check if high_qc was too far from root, but that branch was never hit because it's checked separately in maybe_statesync.

PR #1906, opened 2025-06-26
monad-dataplane, monad-raptorcast: integrate disconnect / trusted peers api

dshulyak:
this change ingegrates with discovery to update dataplane with trusted ips anytime
when epoch changes or when validator has a change in its api address.
connections from trusted ips will bypass total and per ip connections limits.

the other change is to disconnect peer that sent invalid message

### DIFF
--- a/monad-dataplane/src/addrlist.rs
+++ b/monad-dataplane/src/addrlist.rs
@@ -81,76 +81,126 @@ impl AddrEntry {
     }
 }
 
+#[derive(Debug)]
+struct Entries(HashMap<IpAddr, AddrEntry>);
+
 #[derive(Debug, Clone)]
 pub(crate) struct Addrlist {
-    entries: Arc<Mutex<HashMap<IpAddr, AddrEntry>>>,
+    entries: Arc<Mutex<Entries>>,
+}
+
+impl Entries {
+    fn new() -> Self {
+        Self(HashMap::new())
+    }
+
+    fn add_trusted(&mut self, addr: &IpAddr) {
+        self.0
+            .entry(*addr)
+            .and_modify(|e| e.set_trusted(true))
+            .or_insert_with(AddrEntry::new_trusted);
+    }
+
+    fn remove_trusted(&mut self, addr: &IpAddr) {
+        if let Some(entry) = self.0.get_mut(addr) {
+            entry.set_trusted(false);
+        }
+
+        match self.status(addr) {
+            Status::Unknown => {
+                self.0.remove(addr);
+            }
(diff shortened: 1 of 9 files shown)


monad-dataplane/src/addrlist.rs
```diff
@@ -96,6 +96,30 @@ impl Addrlist {
...
+    ) {
+        let mut entries = self.entries.lock().unwrap();
+
+        for addr in removed {
+            if let Some(entry) = entries.get_mut(&addr) {
+                entry.set_trusted(false);
+                if !entry.is_banned() {
+                    entries.remove(&addr);
+                }
+            }
```
andr-dev: I think this would be cleaner if we move it into an `entries.remove_trusted(&addr)` so we can explicitly match on the variants and ignore if banned.

monad-raptorcast/src/lib.rs
```diff
@@ -365,6 +377,26 @@ where
...
+                                    let prev_ip = self
+                                        .peer_discovery_driver
+                                        .get_addr(&node_id)
+                                        .map(|sock| sock.ip());
+                                    (new_ip, prev_ip)
+                                })
+                            })
+                            .for_each(|(new_ip, prev_ip)| {
+                                self.dataplane
+                                    .update_trusted(once(new_ip), prev_ip.into_iter());
```
andr-dev: Why is this `prev_ip.into_iter()` instead of `once(prev_ip)`? Does `sock.ip()` above return an iterable?

monad-raptorcast/src/lib.rs
```diff
@@ -550,13 +593,17 @@ where
             }
         }
 
-        while let Poll::Ready((from_addr, message)) = pin!(this.dataplane.tcp_read()).poll_unpin(cx)
-        {
+        loop {
+            let Poll::Ready((from_addr, message)) = pin!(this.dataplane.tcp_read()).poll_unpin(cx)
+            else {
+                break;
+            };
```
andr-dev: What's the motivation for this change? Before looks cleaner.

monad-raptorcast/src/lib.rs
dshulyak: prev_ip is Option<IpAddr> , it provides into_iter() which is basically once(..) or empty()

monad-raptorcast/src/lib.rs
dshulyak: borrow checker. tcp read mutably borrows dataplane and below i need to get another ref for dataplane.disconnect calls

PR #1909, opened 2025-06-26
monad-raptorcast: offload udp validation to a set of worker threads (defaulting to 1 worker)

dshulyak:
it is possible to halt a node by pointing a simple script that sends udp messages,
if scheduled correctly it can halt leaders continuously and halt whole network.

the rationale behind this change is to increase validation throughput to a point where cpu is not a limiting factor. 
it is kept by default to 1 worker as it is only useful under attack, and in normal circumstances udp validation is not a bottleneck.

so if validator observes such behavior and has no means to block traffic in some other way they can increase number of workers. 
also blocking udp traffic is generally harder than tcp, as it is easier to spoof.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1362,9 +1362,9 @@ dependencies = [
 
 [[package]]
 name = "aws-sdk-s3"
-version = "1.93.0"
+version = "1.94.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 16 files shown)


Cargo.lock
```diff
@@ -1362,9 +1362,9 @@ dependencies = [
 
 [[package]]
 name = "aws-sdk-s3"
-version = "1.93.0"
+version = "1.94.0"
```
dshulyak, line 1365: weird that there are so many updates here, need to recheck

dshulyak:
superseded by auth work

PR #1912, opened 2025-06-26
Remove wal-tool

omegablitz:
We've used wal2json instead of wal-tool in recent memory. Removing this to lower maintenance burden.


omegablitz:
Thanks, https://github.com/category-labs/category-infra/pull/1462

PR #1917, opened 2025-06-27
Add ST, SCT to StateBackend

bamrith:
First PR to be able to read validator set from TrieDB.

- Moved SignatureCollection and ValidatorMapping to `monad-validator`
- Added ST, SCT to StateBackend
- Renamed val_set_update_interval to epoch_length
- Renamed all StateRootHashExecutor (and related types) to ValSetUpdater

### DIFF
--- a/monad-state-backend/Cargo.toml
+++ b/monad-state-backend/Cargo.toml
@@ -9,10 +9,15 @@ edition = "2021"
 bench = false
 
 [dependencies]
+monad-crypto = { workspace = true }
 monad-eth-types = { workspace = true }
 monad-types = { workspace = true }
+monad-validator = { workspace = true }
 
 alloy-consensus = { workspace = true }
 alloy-primitives = { workspace = true }
 tracing = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
+
+[dev-dependencies]
+monad-multi-sig = { workspace = true }
--- a/monad-triedb-utils/src/lib.rs
+++ b/monad-triedb-utils/src/lib.rs
@@ -26,10 +26,12 @@ use alloy_primitives::Address;
 use alloy_rlp::Decodable;
 use futures::{channel::oneshot, executor::block_on, future::join_all, FutureExt};
 use key::Version;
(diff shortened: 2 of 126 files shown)


monad-validator/src/validator_mapping.rs
```diff
@@ -0,0 +1,26 @@
+use std::collections::BTreeMap;
+
+use monad_crypto::certificate_signature::{CertificateKeyPair, PubKey};
+use monad_types::NodeId;
+
+/// Map validator NodeId to its Certificate PubKey
+pub struct ValidatorMapping<PT: PubKey, VKT: CertificateKeyPair> {
```
andr-dev, line 22: Why did we move this in this changeset?

monad-triedb-utils/src/lib.rs
```diff
@@ -379,6 +381,10 @@ impl StateBackend for TriedbReader {
         self.get_latest_finalized_block()
     }
 
+    fn read_next_valset(&self, _block_num: SeqNum) -> Vec<(PubKey, BlsPubKey, Stake)> {
```
andr-dev, line 293: Can we separate this into another PR?

monad-validator/src/validator_mapping.rs
bamrith: `ValidatorsEpochMapping` depends on this, which is in `monad-validator`. Can't import `monad-consensus-types` into this crate

monad-triedb-utils/src/lib.rs
bamrith: I'd rather keep this in here to make it clear what the purpose of ST, SCT is. Else, it's just phantom

monad-state-backend/Cargo.toml
```diff
@@ -9,8 +9,11 @@ edition = "2021"
 bench = false
 
 [dependencies]
+monad-crypto = { workspace = true }
 monad-eth-types = { workspace = true }
+monad-multi-sig = { workspace = true }
```
omegablitz: nit: dev-dep

PR #1919, opened 2025-06-28
Use RawValue in RPC response for faster serialization

andr-dev:
Previously, we used `serde_json::Value` to generically pass around in-memory JSON objects and effectively "compose" JSON results. This however has a major downside as every RPC handler result would need to be serialized from `T -> serde_json::Value` and then serialized from `serde_json::Value -> String`. For complex / nested JSON structures (most RPC responses), this results in tons of unnecessary allocations to create `serde_json::Value::Map(...)`s and `serde_json::Value::Array(...)`s in addition to requiring the "double partial serialization" mentioned above.

This PR changes `serde_json::Value` to its sibling type `Box<serde_json::value::RawValue>`. `serde_json::value::RawValue` is a thin wrapper around `str` that is hooked into `serde_json` to allow generically passing around serialized JSON objects. Unlike `str` however, serializing a `serde_json::value::RawValue` produces the JSON object itself rather than being serialized as a JSON string and thus adding double quotes around the child JSON object.

By using `serde_json::value::RawValue`, we still get to compose JSON objects but without the massive performance hit of `serde_json::Value` (see benchmark below). The downside is that calling `to_string_pretty(...)` on an object containing a `serde_json::value::RawValue` will not result in a prettified string (and vice-versa) since the JSON is pre-serialized and thus the serialization is a simple string copy. This is not a concern for RPC.

# Benchmark
```
serialize/block_hashes_1k
                        time:   [37.822 µs 37.834 µs 37.848 µs]
                        thrpt:  [1.7311 GiB/s 1.7317 GiB/s 1.7323 GiB/s]
                 change:
                        time:   [-21.533% -21.330% -21.128%] (p = 0.00 < 0.05)
                        thrpt:  [+26.788% +27.114% +27.442%]

serialize/block_full_1k
                        time:   [823.02 µs 823.37 µs 823.74 µs]
                        thrpt:  [510.96 MiB/s 511.19 MiB/s 511.42 MiB/s]
                 change:
                        time:   [-64.467% -64.446% -64.425%] (p = 0.00 < 0.05)
                        thrpt:  [+181.10% +181.27% +181.43%]

serialize/block_receipts_1k
                        time:   [484.54 µs 484.65 µs 484.78 µs]
                        thrpt:  [1.6695 GiB/s 1.6700 GiB/s 1.6703 GiB/s]
                 change:
                        time:   [-71.511% -71.440% -71.369%] (p = 0.00 < 0.05)
                        thrpt:  [+249.27% +250.14% +251.02%]

serialize/logs_1k
                        time:   [132.28 µs 132.49 µs 132.70 µs]
                        thrpt:  [1.3759 GiB/s 1.3780 GiB/s 1.3802 GiB/s]
                 change:
                        time:   [-85.237% -85.213% -85.191%] (p = 0.00 < 0.05)
                        thrpt:  [+575.25% +576.28% +577.35%]
                        Performance has improved.
```

### DIFF
--- a/monad-rpc/Cargo.toml
+++ b/monad-rpc/Cargo.toml
@@ -61,9 +61,9 @@ pin-project = { workspace = true }
 rand = { workspace = true }
 rayon = { workspace = true }
 reqwest = { workspace = true }
-schemars = { workspace = true, features = ["preserve_order"] }
+schemars = { workspace = true, features = ["preserve_order", "raw_value"] }
 serde = { workspace = true, features = ["derive", "rc"] }
-serde_json = { workspace = true }
+serde_json = { workspace = true, features = ["raw_value"] }
 tokio = { workspace = true, features = ["net", "macros", "rt-multi-thread", "fs"] }
 toml = { workspace = true }
 tracing = { workspace = true, features = ["log"] }
--- a/monad-rpc/src/eth_json_types.rs
+++ b/monad-rpc/src/eth_json_types.rs
@@ -5,7 +5,7 @@ use alloy_primitives::{Address, FixedBytes, LogData, U256};
 use alloy_rpc_types::{Block, FeeHistory, Header, Log, Transaction, TransactionReceipt};
 use schemars::JsonSchema;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
-use serde_json::Value;
+use serde_json::value::RawValue;
 use tracing::debug;
 
 use crate::{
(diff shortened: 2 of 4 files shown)


eerkaijun:
> Unlike str however, serializing a serde_json::value::RawValue produces the JSON object itself rather than being serialized as a JSON string and thus adding double quotes around the child JSON object.

are there any output difference when calling an rpc method? if so, do you have an example? just wanted to make sure we don't break conformance

andr-dev:
> are there any output difference when calling an rpc method? if so, do you have an example? just wanted to make sure we don't break conformance

Output is identical as long as you're not using the serde_json `pretty` variants which we're not in RPC. There are several rust unit tests and flexnet tests that check that this is the case as well as live test usage (two txgens) that have been run against this change on stressnet.

PR #1921, opened 2025-06-28
Add github action concurrency control

andr-dev:
This change makes it so only one of each `rust` or `openrpc` github action can be queued / running for a given pull request. The net effect of this is that pushing to a branch in a PR will cancel the previous run before starting the next.

Example:
https://github.com/category-labs/monad-bft/actions/runs/15947745149
https://github.com/category-labs/monad-bft/actions/runs/15947777608

Additionally, the "edited" trigger type triggers the github actions to run again after editing PR metadata like title / description. Instead of removing just "edited", the entire line was removed which reverts it back to default behavior, that being the remaining three types.

> By default, a workflow only runs when a pull_request event's activity type is `opened`, `synchronize`, or `reopened`.

https://docs.github.com/en/actions/reference/events-that-trigger-workflows

### DIFF
--- a/.github/workflows/openrpc.yml
+++ b/.github/workflows/openrpc.yml
@@ -8,6 +8,14 @@ on:
     paths:
       - "monad-rpc/**"
 
+concurrency:
+  group: >
+    ${{
+      github.event_name == 'pull_request'
+      && format('pr-{0}-openrpc', github.event.pull_request.number)
+      || format('push-{0}-openrpc', github.ref)
+    }}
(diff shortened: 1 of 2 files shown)


michael-yxchen:
i believe "edited" triggers PR run on parent branch change. it's helpful for chained PRs

andr-dev:
> i believe "edited" triggers PR run on parent branch change. it's helpful for chained PRs

Got it, seeing a lot of discussion online about this being annoying because there's no easy fix to avoid on PR title / description changes, we can keep it as is.

PR #1924, opened 2025-06-30
Adds connection limit to websockets

philipglazman:


monad-rpc/src/websocket/handler.rs
```diff
@@ -36,11 +37,31 @@ const CLIENT_TIMEOUT_SECS: u64 = 60;
 #[derive(Debug, PartialEq, Eq, Hash, Clone, Copy, Deserialize, Serialize)]
 pub struct SubscriptionId(pub FixedData<16>);
 
+#[derive(Clone)]
+pub struct ConnectionLimit(Arc<Semaphore>);
```
dshulyak, line 58: i think you can just use Arc itself to enforce a limit, just check the number of strong references in acquire. It will increment on clone and decrement on drop when finished with permit

philipglazman: That can work, but an issue with this approach is the baseline reference count changes depending on the number of workers for actix (clone happens during server startup). I can add starting baseline count to the ConnectionLimit struct but that seems less ideal than using a semaphore.

monad-rpc/Cargo.toml
```diff
@@ -80,6 +80,7 @@ optional = true
 
 [dev-dependencies]
 monad-eth-testutil = { workspace = true }
+awc = "3"
```
omegablitz: Should this be in the workspace cargo.toml?

Issue #1914, opened 2025-06-27
raptorcast: fail to decode large TCP messages


robertkiel:
Nodes understanding the Raptorcast protocol accept RouterMessage messages sent via TCP and UDP. Messages sent to the UDP endpoint use a custom coding scheme whereas messages sent to the TCP sockets get sent as is.

In addition, Raptorcast uses a compression scheme to reduce the size of the "app message" before encoding it (for UDP) and transferring it.

The compression scheme prefixes the compressed "app message" with their decompressed (original) size using a `u32`, which means that "app messages" cannot be larger than 4 GB. However, the embedded length gets derived by casting a `usize` (on most machines a `u64`) to a `u32`, which effectively truncates 32 bits.

```rust
// compress message
let mut rlp_encoded_msg = BytesMut::new();
app_message.encode(&mut rlp_encoded_msg);
// u32::MAX is 4GB
let message_len = rlp_encoded_msg.len() as u32;
```
See [original code](https://github.com/category-labs/monad-bft/blob/622a2ef3941f5979e1d3e9d9feb50b126d414711/monad-raptorcast/src/message.rs#L81)

The UDP part comes with a [length check](https://github.com/category-labs/monad-bft/blob/6462ab2436361f18fa82a3ea24357c49eee39414/monad-raptorcast/src/udp.rs#L508) and therefore prevents encoding large message whereas the TCP counterpart misses a length check and just takes the [raw outcome](https://github.com/category-labs/monad-bft/blob/622a2ef3941f5979e1d3e9d9feb50b126d414711/monad-raptorcast/src/lib.rs#L160) of the serializer and thus allows arbitrarily big messages. Due to the `u32` cast, these messages will be prefixed with a wrong length. Decoding such a big message will produce an error that the encoding party cannot observe.

### Suggestion

Use something like `u32::try_into().map_err(SerializationError)` to prevent large messages from being encoded.

dshulyak:
should the title be: raptorcast: fail to **encode** large TCP messages ?

robertkiel:
> should the title be: raptorcast: fail to **encode** large TCP messages ?

Probably yes. In this case, encoding succeeds but decoding fails.

PR #1926, opened 2025-06-30
monad-raptorcast: add a validation that outbound messages respect protocol size limits

dshulyak:
closes: https://github.com/category-labs/monad-bft/issues/1914

### DIFF
--- a/monad-raptorcast/src/lib.rs
+++ b/monad-raptorcast/src/lib.rs
@@ -412,7 +412,15 @@ where
                                 _ => unreachable!(),
                             };
                             let outbound_message =
-                                OutboundRouterMessage::<OM, ST>::AppMessage(message).serialize();
+                                match OutboundRouterMessage::<OM, ST>::AppMessage(message)
+                                    .try_serialize()
+                                {
+                                    Ok(msg) => msg,
+                                    Err(err) => {
+                                        error!(?err, "failed to serialize a message");
+                                        continue;
+                                    }
+                                };
                             let rc_chunks: UnicastMsg = Self::udp_build(
                                 &epoch,
                                 build_target,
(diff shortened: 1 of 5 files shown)


monad-raptorcast/src/lib.rs
```diff
@@ -317,9 +325,13 @@ where
                                     waker.wake()
                                 }
                             } else {
-                                let outbound_message =
+                                let Ok(outbound_message) =
                                     OutboundRouterMessage::<OM, ST>::AppMessage(message)
-                                        .serialize();
+                                        .try_serialize()
+                                else {
+                                    error!("failed to serialize a message");
```
dshulyak: inconsistent, use error!(?err

monad-raptorcast/src/lib.rs
```diff
@@ -596,10 +618,13 @@ where
...
                 PeerDiscoveryEmit::RouterCommand { target, message } => {
-                    let router_message =
-                        OutboundRouterMessage::serialize(
-                            OutboundRouterMessage::<OM, ST>::PeerDiscoveryMessage(message),
-                        );
+                    let Ok(router_message) =
+                        OutboundRouterMessage::<OM, ST>::PeerDiscoveryMessage(message)
+                            .try_serialize()
+                    else {
+                        error!("failed to serialize peer discovery message");
```
andr-dev: Why don't we log the err like in the other changes?

monad-raptorcast/src/message.rs
```diff
@@ -61,13 +65,22 @@ pub enum OutboundRouterMessage<OM, ST: CertificateSignatureRecoverable> {
     FullNodesGroup(FullNodesGroupMessage<ST>),
 }
 
+#[derive(Debug, Error)]
+pub enum SerializeError {
+    #[error("message too large: {0} bytes exceeds maximum of {1} bytes")]
+    MsgTooLarge(usize, usize),
```
andr-dev: Can we make this two named variants?
```rust
RlpTooLarge {
  rlp_length: usize,
  max_length: usize,
},
DecodedMsgTooLarge {
  msg_length: usize,
  max_length: usize,
}
```

dshulyak: there is another error type for decoder, i didn't change it here.
there are 2 places where MsgTooLarge is emitted:
1. always after final msg size is checked, both compressed / uncompressed
2. and there is additional check in compressed when prefixed with length

also it is always rlp length, i think i will name them InnerrMsgTooLarge and FinalMsgTooLarge to show the difference

PR #1928, opened 2025-06-30
Fix block filter for logs in websockets

philipglazman:
Previously, websocket log filtering ignored any block filters


philipglazman:
Fixes https://github.com/category-labs/monad-bft/issues/1988

PR #1932, opened 2025-06-30
Optimize RPC logs retrieval

andr-dev:


monad-rpc/src/chainstate/mod.rs
```diff
@@ -446,6 +446,8 @@ impl<T: Triedb> ChainState<T> {
         dry_run_get_logs_index: bool,
         max_finalized_block_cache_len: u64,
     ) -> JsonRpcResult<Vec<MonadLog>> {
+        let mut latest_block_number = None;
```
eerkaijun: do we need this? can't we just do `let latest_block_number = self.get_latest_block_number();` here?

andr-dev: There are some cases where we return before calling `let latest_block_number = latest_block_number.unwrap_or(...)` so it's a minor optimization, I can change it back if you have strong feelings about it, not a _significant_ value add.

eerkaijun: getting the latest block number is relatively cheap, so feels like it's clearer this way

PR #1933, opened 2025-07-01
Fractional redundancy factor support for Raptorcast

xinyuan-dev:
This PR adds support for redundancy factor to be a fractional number using a fixed-point data type represented as a u16. Additionally, having a discrete data type allows us to represent all allowed operations with the redundancy factor as a closed set.


monad-raptorcast/src/util.rs
```diff
@@ -499,6 +499,65 @@ where
...
+// | [ I I I I I F F F F F F F F F F F ] |
+// |   (5 bits)  (11 bits)               |
+// +-------------------------------------+
+//
+// I = Integer part (bits 15..11)
+// F = Fractional part (bits 10..0)
+// Value = Integer + Fraction / 2^11
+// Range: 0 to ~31.9995, Increments: ~0.000488
+#[derive(Clone, Copy)]
+pub struct Redundancy(u16);
```
andr-dev: Let's use `https://docs.rs/fixed/latest/fixed/struct.FixedU16.html` so we don't have to think too hard about correctness. Also, is there a reason for constraining this to a 16 bit value?

monad-raptorcast/src/udp.rs
```diff
@@ -578,12 +583,13 @@ where
...
-            * u32::from(redundancy))
-        .try_into()
-        .expect("num_packets doesn't fit in usize");
+        let mut num_packets: usize = (app_message_len as usize)
+            .div_ceil(usize::from(data_size))
+            .max(SOURCE_SYMBOLS_MIN);
+        // amplify by redundancy factor
+        num_packets = redundancy
+            .scale(num_packets)
+            .expect("num_packets doesn't fit in usize");
```
andr-dev: Let's update this expect message.

monad-raptorcast/src/util.rs
xinyuan-dev: > Also, is there a reason for constraining this to a 16 bit value?

8 bit value doesn't provide enough precision, so 16 bit seems like a sweet spot.

andr-dev: Right my question is why not 32bits? 64bits?

xinyuan-dev: We could, but do we need that much precision? 2^-11 delta seems enough. And I can't think of cases we would like >32x redundancy factor.

monad-raptorcast/src/util.rs
```diff
@@ -499,6 +499,65 @@ where
...
+
+impl Redundancy {
+    pub const BITS: u32 = 16;
+    pub const FRAC_BITS: u32 = 11;
+    pub const MIN: Self = Self(u16::MIN);
+    pub const MAX: Self = Self(u16::MAX);
+    pub const DELTA: Self = Self(1);
+    pub const MAX_MULTIPLIER: usize = usize::MAX / (u16::MAX as usize);
+
+    // guaranteed to be lossless for num in [0,31).
```
xinyuan-dev:
```suggestion
    // guaranteed to be lossless for num in [0,32).
```

monad-raptorcast/src/util.rs
andr-dev: Yeah that's fair, I think if we use the `fixed` library it's okay to use something like `FixedU16<11>`. I just want to make sure we're not introducing any unnecessary complexity by using smaller int sizes, but if that's encapsulated by `fixed` which is well tested then I have no issues with it.

monad-raptorcast/Cargo.toml
```diff
@@ -33,6 +33,7 @@ rand = { workspace = true }
 rand_chacha = { workspace = true }
 thiserror = { workspace = true }
 tracing = { workspace = true }
+fixed = { workspace = true}
```
andr-dev: This dep needs to be placed in sorted order.

monad-raptorcast/src/udp.rs
```diff
@@ -261,7 +264,9 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
                     // symbol_len is always greater than zero, so this division is safe
                     let num_source_symbols =
                         app_message_len.div_ceil(symbol_len).max(SOURCE_SYMBOLS_MIN);
-                    let encoded_symbol_capacity = MAX_REDUNDANCY * num_source_symbols;
+                    let encoded_symbol_capacity = MAX_REDUNDANCY
+                        .scale(num_source_symbols)
+                        .expect("num_source_symbols doesn't fit in usize");
```
andr-dev: Let's update this expect, since `num_source_symbols` clearly does fit in a `usize`, if not then we wouldn't be able to pass it to the `scale` fn.

monad-raptorcast/src/util.rs
```diff
@@ -519,6 +520,50 @@ where
...
+pub struct Redundancy(FixedU16<U11>);
+
+impl Redundancy {
+    pub const ZERO: Self = Self(FixedU16::ZERO);
+    pub const BITS: u32 = 16;
+    pub const FRAC_BITS: u32 = 11;
+    pub const MIN: Self = Self(FixedU16::MIN);
+    pub const MAX: Self = Self(FixedU16::MAX);
+    pub const DELTA: Self = Self(FixedU16::DELTA);
+    pub const MAX_MULTIPLIER: usize = usize::MAX / (u16::MAX as usize);
```
andr-dev: Shouldn't this be `usize::MAX >> 11`? Also, I think we can make some of these non-pub. Ideally, I'd only like to see ZERO, MIN, and MAX exposed.

monad-raptorcast/src/util.rs
```diff
@@ -519,6 +520,50 @@ where
...
+    }
+
+    // may round to the nearest representable number when needed
+    pub fn from_f32(num: f32) -> Option<Self> {
+        FixedU16::<U11>::checked_from_num(num).map(Redundancy)
+    }
+
+    pub fn to_f32(&self) -> f32 {
+        self.0.to_num::<f32>()
+    }
```
andr-dev: For these two, I think you can omit the turbofish syntax because in the first the compiler can deduce the typenum `U11` from the `Redundancy` type and for `to_f32` it can deduce it from the return type.

monad-raptorcast/src/util.rs
xinyuan-dev: I want to ensure that `Redundancy::MAX in bits * Redundancy::MAX_MULTIPLIER` does not overflow `usize` during the computation, so I think this is a natural way to write it. if we just consider `usize::MAX >> 11` as `MAX_MULTIPLIER`, it would overflow and requires special handling. The current value of `MAX_MULTIPLIER` is around the size of 48 bits, which should be sufficient for the desired use.

monad-raptorcast/src/udp.rs
xinyuan-dev: I missed this one, let me change it to "redundancy-scaled num_source_symbols doesn't fit in usize".

monad-raptorcast/src/util.rs
xinyuan-dev: You're right, I'll remove the unnecessary type annotation.

monad-raptorcast/src/util.rs
andr-dev: Got it, yeah that makes sense then. Let's put ZERO, MIN, and MAX as is at the top as pub followed by the rest without pub.

monad-raptorcast/src/udp.rs
```diff
@@ -50,7 +53,7 @@ const MAX_NUM_PACKETS: usize = 65535;
 //
 // We pick 7 because that is the largest value that works for all values of K, as K
 // can be at most 8192, and there can be at most 65521 encoding symbol IDs.
-pub const MAX_REDUNDANCY: usize = 7;
+pub const MAX_REDUNDANCY: Redundancy = Redundancy::from_u8(7);
```
andr-dev, line 56: Can we make this a function of `Redundancy::MAX`? I would have expected `Redundancy::MAX / SOURCE_SYMBOLS_MIN`

xinyuan-dev: I suppose you mean `MAX_TRIPLES / SOURCE_SYMBOLS_MAX`? There is a problem though - converting from a fractional number to Fixed is not done in compile time, so we can assign the constant to that expression. Do you suggest we perform the calculation manually and hardcode the bits here?

andr-dev: Ah if there's no pretty compile time way of doing it, let's leave it as is. I thought there might be a way with some of the const fns.

PR #1935, opened 2025-07-02
Fix Raptorcast starvation issues

### LINKED ISSUE #1915: raptorcast: starvation issues during message processing
The Raptorcast instance implements `futures::Stream` such that an async runtime can `poll` it in order to process messages and finally create `MonadEvent`s, if necessary.

Extract from Raptorcast [`futures::Stream::poll_next` implementation](https://github.com/category-labs/monad-bft/blob/1751eb8c34656a5f8f5ae15168a43b8b3b494eb7/monad-raptorcast/src/lib.rs#L456)

```rust
loop {
    let Poll::Ready(message) = pin!(this.dataplane.udp_read()).poll_unpin(cx) else {
        break;
    };

    // do UDP handling
}
// ...
while let Poll::Ready((from_addr, message)) = pin!(this.dataplane.tcp_read()).poll_unpin(cx)
{
    // do TCP messages handling
}
// ...
while let Poll::Ready(Some(peer_disc_emit)) = this.peer_discovery_driver.poll_next_unpin(cx)
{
    // do peer discovery handling
}
```

Although, the aforementioned code correctly handles incoming messages, it *prioritizes* UDP messages over all other messages, namely TCP messages and peer discovery messages. More precisely, handling TCP messages starts once all UDP messages have been handled. This can be exploited by an adversary to prevent the node from processing messages sent over TCP or engaging in peer discovery. The adversary therefore just needs to spam the node with UDP messages that are expensive to decode, i.e. those that contain a correctable error or contain forged signatures that have not been cached. Also advising the node to do expensive broadcasts to other nodes might trigger the issue.

### Suggestion

Use a more fair scheduler to process messages. A simple [`tokio::select!`](https://docs.rs/tokio/latest/tokio/macro.select.html) could do the job. The [`biased`](https://docs.rs/tokio/latest/tokio/macro.select.html#fairness) directive could be used to prioritize UDP messages over TCP messages.

xinyuan-dev:
Closes #1915.

### DIFF
--- a/monad-raptorcast/src/lib.rs
+++ b/monad-raptorcast/src/lib.rs
@@ -18,7 +18,6 @@ use std::{
     future::Future as _,
     marker::PhantomData,
     net::{IpAddr, SocketAddr},
-    ops::DerefMut,
     pin::{pin, Pin},
     sync::Arc,
     task::{Context, Poll},
@@ -27,7 +26,7 @@ use std::{
 
 use alloy_rlp::{Decodable, Encodable};
 use bytes::{Bytes, BytesMut};
-use futures::{channel::oneshot, FutureExt, Stream};
+use futures::{channel::oneshot, Stream};
 use itertools::Itertools;
 use message::{InboundRouterMessage, OutboundRouterMessage};
 use monad_consensus_types::signature_collection::SignatureCollection;
@@ -39,8 +38,8 @@ use monad_crypto::{
 };
 use monad_dataplane::{
     udp::{segment_size_for_mtu, DEFAULT_MTU},
-    BroadcastMsg, DataplaneBuilder, DataplaneReader, DataplaneWriter, RecvTcpMsg, TcpMsg,
-    UnicastMsg,
+    BroadcastMsg, DataplaneBuilder, DataplaneReader, DataplaneWriter, RecvTcpMsg, RecvUdpMsg,
+    TcpMsg, TcpReader, UdpReader, UnicastMsg,
 };
 use monad_executor::{Executor, ExecutorMetrics, ExecutorMetricsChain};
 use monad_executor_glue::{
@@ -89,7 +88,8 @@ where
     udp_state: udp::UdpState<ST>,
     mtu: u16,
 
-    dataplane_reader: DataplaneReader,
+    dataplane_tcp_reader: TcpReader,
+    dataplane_udp_reader: UdpReader,
     dataplane_writer: DataplaneWriter,
 
     pending_event_tx: UnboundedSender<RaptorCastEvent<M::Event, ST>>,
@@ -143,6 +143,7 @@ where
         );
 
         let (pending_event_tx, pending_event_rx) = tokio::sync::mpsc::unbounded_channel();
+        let (dataplane_tcp_reader, dataplane_udp_reader) = dataplane_reader.split();
 
         Self {
             is_dynamic_fullnode,
@@ -163,7 +164,8 @@ where
             udp_state: udp::UdpState::new(self_id, config.udp_message_max_age_ms),
             mtu: config.mtu,
 
-            dataplane_reader,
+            dataplane_tcp_reader,
+            dataplane_udp_reader,
             dataplane_writer,
             pending_event_tx,
             pending_event_rx,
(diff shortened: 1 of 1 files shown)


dshulyak (commented):
this will conflict with this change https://github.com/category-labs/monad-bft/pull/1909 . 
tldr is that i splitted dataplane so that i can read udp in separate workers, tcp is not that easy to read in parallel as the messages are expected to be ordered

beside that it only solves the issue for udp/tcp messages, what about other handles? 
i think the correct approach would be to start using async/await syntax instead of raw state machines.

alternatively we can implement what select! does internally and use that directly in raptorcast loop

monad-raptorcast/src/lib.rs
```diff
@@ -173,6 +174,182 @@ where
...
+                self.handle_dataplane_message(msg);
+                true
+            },
+            emit = self.peer_discovery_driver.next().fuse() => {
+                let Some(emit) = emit else { return false; };
+                self.handle_peer_discovery_emit(emit);
+                true
+            },
+            else => false,
+        }
```
xinyuan-dev: @dshulyak:

> beside that it only solves the issue for udp/tcp messages, what about other handles?

it solves the issue for udp/tcp and peer disc emits. is there anything else i'm missing here?

> i think the correct approach would be to start using async/await syntax instead of raw state machines.

the original state machine is kept because of the waker saved for triggering new polling during internal events emits. other than that, I transformed most of the heavy lifting to async/await here, so the raw stream implementation is a short wrapper around this async function.

dshulyak: sorry i didn't notice that you are actually using select. what is the purpose of merging udp/tcp into one stream then? i will need to roll it back if this is merged before my change

xinyuan-dev: > what is the purpose of merging udp/tcp into one stream then?

`select!` requires to race `udp_read` and `tcp_read` at the same time. if we perform `select!` in raptorcast for both udp and tcp, that will borrow `self.dataplane` mutably twice, which is blocked by the borrow checker. We know it's safe because `udp_read` and `tcp_read` depend on different fields, but such knowledge is limited to within dataplane.

dshulyak: we can split dataplane into tcp reader and udp reader, and writer. it will be compatible with the change i made in linked pr

xinyuan-dev: sure that'll work, i'll make the changes

xinyuan-dev: changed in https://github.com/category-labs/monad-bft/pull/1935/commits/25d8fce5f099875c156e0a19bd197e64c71e1fec.

monad-raptorcast/src/lib.rs
```diff
@@ -589,261 +839,13 @@ where
...
-            }
-
-            if let Poll::Ready(event) = this.pending_event_rx.poll_recv(cx) {
-                let event = event.expect("raptorcast pending_event_tx dropped");
-                return Poll::Ready(Some(event.into()));
+        match self.pending_event_rx.poll_recv(cx) {
+            Poll::Ready(event) => Poll::Ready(event.map(|e| e.into())),
+            Poll::Pending => {
+                let Poll::Pending = pin!(self.poll_input_sources()).poll(cx);
+                Poll::Pending
```
xinyuan-dev, line 843: No need to check pending events after polling the sources, this is safe because in case any events is generated and send down the `pending_event_tx` channel, the channel's internal waker will ensure the scheduler to poll the stream again.

monad-raptorcast/src/lib.rs
```diff
@@ -275,6 +277,254 @@ where
...
+                self.peer_discovery_metrics = executor_metrics;
+            }
+        }
+    }
+
+    //  This async function always returns Poll::Pending when polled.
+    async fn poll_input_sources(&mut self) -> ! {
+        loop {
+            tokio::select! {
+                udp_msg = self.dataplane_udp_reader.read() => {
```
xinyuan-dev, line 496: question: should we prioritize udp messages here? we can use `tokio::select`'s `biased` marker if we need.

dshulyak: makes sense for me. latency sensitive messages (proposal propagation, votes, other consensus messages) all go through udp, and will likely stay that way. 

so for example, if we have 200 votes in the udp channel and 1 statesync request, should we delay that statesync request until all 200 are processed? i think yes

fabiotdf (approved):

Issue #1940, opened 2025-07-03
async: potential locking issues due to missing waker invocations


robertkiel:
While going through the `futures-util::Stream` implementations that are used to wait for emitted events, we noticed some inconsistencies and deviations from best practices that could to lead deadlocks or delayed handling of events (locking). The issue occurs whenever a module can do some progress but also awaits external actions.

During execution, this will make internal events wait until some "external event" gets handled - although they could be handled much earlier, i.e. the Future could make progress but no one informed the scheduler that the Future is ready to do so.

A best practice is to use `Waker::clone_from()` to prevent a deep clone of the waker.
```rust
// impl futures_util::Stream for ...
fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
    // self.waker: Option<Waker>
    match self.waker.as_mut() {
        None => {
            self.waker = Some(cx.waker().clone());
        }
        Some(waker) => {
            waker.clone_from(cx.waker());
        }
    }

    // ... rest of the logic
}
// and wake it once the Future can make progress, i.e.
if let Some(waker) = self.waker.take() {
    waker.wake();
}
```

robertkiel:
PoC:
```rust
use futures_util::{Stream, StreamExt};
use std::{
    collections::VecDeque,
    pin::Pin,
    task::{Context, Poll, Waker},
    time::Duration,
};
use tokio::time::Interval;

struct AsyncStream {
    pub interval: Interval,
    pub events: VecDeque<()>,
    counter: usize,
}

impl AsyncStream {
    pub fn new() -> Self {
        Self {
            interval: tokio::time::interval(Duration::from_secs(30)),
            events: Default::default(),
            counter: 0,
        }
    }

    fn exec(&mut self) {
        print_with_timestamp("Pushing side event");
        self.events.push_back(());
    }
}

impl Stream for AsyncStream {
    type Item = ();

    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
        if let Some(_) = self.events.pop_front() {
            print_with_timestamp("Side event");
            return Poll::Ready(Some(()));
        }

        match self.interval.poll_tick(cx) {
            Poll::Pending => {
                if self.counter == 1 {
                    self.exec();
                }
                Poll::Pending
            }
            Poll::Ready(_) => {
                self.counter += 1;
                print_with_timestamp("Timeout event");
[...]

PR #1941, opened 2025-07-03
Refactor PeerDiscoveryDriver Stream and fix waker usage

andr-dev:
Part 1 in https://github.com/category-labs/category-internal/issues/1727.

### DIFF
--- a/monad-peer-discovery/src/driver.rs
+++ b/monad-peer-discovery/src/driver.rs
@@ -1,8 +1,7 @@
 use std::{
     collections::{HashMap, VecDeque},
     net::SocketAddr,
-    ops::DerefMut,
-    task::{Context, Poll},
+    task::{Context, Poll, Waker},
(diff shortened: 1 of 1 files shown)


dshulyak:
there is an example here https://github.com/category-labs/monad-bft/issues/1940#issuecomment-3033428927

so it is bit different usage pattern. it runs exec inside the poll, which uncovers this edge case.
if you do exec() before poll then obviously stream doesn't get stuck.
and it is impossible to exec() from another task.

i think that it won't be possible to make stream halt using discovery api, but it is somewhat weird mixup overall

andr-dev:
> i think that it won't be possible to make stream halt using discovery api, but it is somewhat weird mixup overall

I haven't dug deep into how it's being used in the discovery api, I'm mostly just concerned with making the module's polling "correct" regardless of parent / child module behavior.

PR #1950, opened 2025-07-05
Reduce full node churn

eerkaijun:
Main changes: 
- Split peer information into `peer_info` and `connection_info`, where `peer_info` contains just the name records and `connection_info` contains information of peers where a ping/pong connection is maintained. I am thinking eventually we will store ConfirmGroup information in `connection_info` too, as a heuristics to determine which full nodes to keep/prune
- A node no longer accept incoming name records if current peer list exceeds `max_num_peers`
- A validators will maintain a ping/pong connection with all other validators, while a full node will only maintain a ping/pong connection with 3 (currently hardcoded but can be configurable) upstream validators

Some risks:
- If bootstrap nodes have peers exceeding `max_num_peers`, it will no longer be able to serve bootstrap requests

### DIFF
--- a/docker/devnet/monad/config/node.toml
+++ b/docker/devnet/monad/config/node.toml
@@ -45,8 +45,8 @@ ping_period = 30
 refresh_period = 120
 request_timeout = 5
 prune_threshold = 5
-min_active_connections = 0
-max_active_connections = 200
+min_num_peers = 0
+max_num_peers = 200
 
 [fullnode_dedicated]
 identities = []
--- a/monad-peer-discovery/src/discovery.rs
+++ b/monad-peer-discovery/src/discovery.rs
@@ -24,10 +24,14 @@ use crate::{
 const MAX_PEER_IN_RESPONSE: usize = 16;
 /// Number of peers to send lookup request to
 const NUM_LOOKUP_PEERS: usize = 3;
+/// Number of validators to connect to if self is a full node
+// TODO: this should be configurable (e.g. a dedicated full node does not need to have 3 upstream validators)
+const NUM_UPSTREAM_VALIDATORS: usize = 3;
 
 /// Metrics constant
 pub const GAUGE_PEER_DISC_SEND_PING: &str = "monad.peer_disc.send_ping";
 pub const GAUGE_PEER_DISC_RECV_PING: &str = "monad.peer_disc.recv_ping";
+pub const GAUGE_PEER_DISC_DROP_PING: &str = "monad.peer_disc.drop_ping";
 pub const GAUGE_PEER_DISC_PING_TIMEOUT: &str = "monad.peer_disc.ping_timeout";
 pub const GAUGE_PEER_DISC_SEND_PONG: &str = "monad.peer_disc.send_pong";
 pub const GAUGE_PEER_DISC_RECV_PONG: &str = "monad.peer_disc.recv_pong";
@@ -46,11 +50,19 @@ pub const GAUGE_PEER_DISC_LOOKUP_TIMEOUT: &str = "monad.peer_disc.lookup_timeout
 pub const GAUGE_PEER_DISC_REFRESH: &str = "monad.peer_disc.refresh";
 pub const GAUGE_PEER_DISC_NUM_CONNECTED_PEERS: &str = "monad.peer_disc.num_connected_peers";
 
+/// validator role is given if the node is a validator in the current or next epoch.
+/// this is to ensure the node starts connecting to other validators even if joining
+/// as a validator only in the next epoch
+#[derive(Debug, Clone, PartialEq, Eq)]
+enum Role {
+    Validator,
+    FullNode,
+}
+
 #[derive(Debug, Clone, Copy)]
-pub struct PeerInfo<ST: CertificateSignatureRecoverable> {
+pub struct ConnectionInfo<ST: CertificateSignatureRecoverable> {
     pub last_ping: Option<Ping<ST>>,
     pub unresponsive_pings: u32,
-    pub name_record: MonadNameRecord<ST>,
 }
 
 #[derive(Debug, Clone, Copy)]
@@ -66,12 +78,20 @@ pub struct LookupInfo<ST: CertificateSignatureRecoverable> {
 pub struct PeerDiscovery<ST: CertificateSignatureRecoverable> {
     pub self_id: NodeId<CertificateSignaturePubKey<ST>>,
     pub self_record: MonadNameRecord<ST>,
+    // role of the node in the current epoch
+    self_role: Role,
     pub current_epoch: Epoch,
+    // mapping of epoch to validators in that epoch
     pub epoch_validators: BTreeMap<Epoch, BTreeSet<NodeId<CertificateSignaturePubKey<ST>>>>,
     // pinned full nodes are dedicated and prioritized full nodes passed in from config that will not be pruned
     pub pinned_full_nodes: BTreeSet<NodeId<CertificateSignaturePubKey<ST>>>,
     // mapping of node IDs to their corresponding name records
-    pub peer_info: BTreeMap<NodeId<CertificateSignaturePubKey<ST>>, PeerInfo<ST>>,
+    pub routing_info: BTreeMap<NodeId<CertificateSignaturePubKey<ST>>, MonadNameRecord<ST>>,
+    // mapping of node IDs to their connection info, used to track ping status
+    // connection_info is a subset of routing_info, i.e. only nodes that are currently connected
+    pub connection_info: BTreeMap<NodeId<CertificateSignaturePubKey<ST>>, ConnectionInfo<ST>>,
+    // mapping of lookup IDs to their corresponding lookup info, node will only entertain lookup response
+    // that matches a local lookup ID
     pub outstanding_lookup_requests: HashMap<u32, LookupInfo<ST>>,
     pub metrics: ExecutorMetrics,
     // duration before sending next ping
@@ -82,10 +102,10 @@ pub struct PeerDiscovery<ST: CertificateSignatureRecoverable> {
     pub request_timeout: Duration,
     // number of unresponsive pings allowed before being pruned
     pub prune_threshold: u32,
-    // minimum number of active connections before actively sending peer lookup requests
-    pub min_active_connections: usize,
-    // maximum number of active connections before pruning
-    pub max_active_connections: usize,
+    // minimum number of peers before actively sending peer lookup requests
+    pub min_num_peers: usize,
+    // maximum number of peers before pruning
+    pub max_num_peers: usize,
     pub rng: ChaCha8Rng,
 }
 
@@ -95,13 +115,13 @@ pub struct PeerDiscoveryBuilder<ST: CertificateSignatureRecoverable> {
     pub current_epoch: Epoch,
     pub epoch_validators: BTreeMap<Epoch, BTreeSet<NodeId<CertificateSignaturePubKey<ST>>>>,
     pub pinned_full_nodes: BTreeSet<NodeId<CertificateSignaturePubKey<ST>>>,
-    pub peer_info: BTreeMap<NodeId<CertificateSignaturePubKey<ST>>, PeerInfo<ST>>,
+    pub routing_info: BTreeMap<NodeId<CertificateSignaturePubKey<ST>>, MonadNameRecord<ST>>,
     pub ping_period: Duration,
     pub refresh_period: Duration,
     pub request_timeout: Duration,
     pub prune_threshold: u32,
-    pub min_active_connections: usize,
-    pub max_active_connections: usize,
+    pub min_num_peers: usize,
+    pub max_num_peers: usize,
     pub rng: ChaCha8Rng,
 }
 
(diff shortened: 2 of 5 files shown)


monad-peer-discovery/src/discovery.rs
```diff
@@ -46,11 +50,19 @@ pub const GAUGE_PEER_DISC_LOOKUP_TIMEOUT: &str = "monad.peer_disc.lookup_timeout
 pub const GAUGE_PEER_DISC_REFRESH: &str = "monad.peer_disc.refresh";
 pub const GAUGE_PEER_DISC_NUM_CONNECTED_PEERS: &str = "monad.peer_disc.num_connected_peers";
 
+/// validator role is given if the node is a validator in the current or next epoch.
+/// this is to ensure the node starts connecting to other validators even if joining
+/// as a validator only in the next epoch
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Role {
```
michael-yxchen: nit: does it need to be public?

monad-peer-discovery/src/discovery.rs
```diff
@@ -66,12 +78,19 @@ pub struct LookupInfo<ST: CertificateSignatureRecoverable> {
...
+    // role of the node in the current epoch
+    pub self_role: Role,
     pub current_epoch: Epoch,
+    // mapping of epoch to validators in that epoch
     pub epoch_validators: BTreeMap<Epoch, BTreeSet<NodeId<CertificateSignaturePubKey<ST>>>>,
     // pinned full nodes are dedicated and prioritized full nodes passed in from config that will not be pruned
     pub pinned_full_nodes: BTreeSet<NodeId<CertificateSignaturePubKey<ST>>>,
     // mapping of node IDs to their corresponding name records
-    pub peer_info: BTreeMap<NodeId<CertificateSignaturePubKey<ST>>, PeerInfo<ST>>,
+    pub peer_info: BTreeMap<NodeId<CertificateSignaturePubKey<ST>>, MonadNameRecord<ST>>,
```
michael-yxchen: nit: peer_info can imply some connection state is kept. maybe rename to `routing_info` or `routing_table`?

monad-peer-discovery/src/discovery.rs
```diff
@@ -226,30 +271,82 @@ impl<ST: CertificateSignatureRecoverable> PeerDiscovery<ST> {
...
+            for &validator in new_upstream_validators {
+                cmds.extend(self.send_ping(validator));
+            }
+        }
+
+        cmds
+    }
+
+    // a helper function to check if a node is a validator in the current or next epoch
+    fn is_validator(&self, node_id: &NodeId<CertificateSignaturePubKey<ST>>) -> bool {
```
michael-yxchen, line 346: naming is a little confusing because `self.is_validator(node_id)` makes it look like we're checking if `self` is a validator. Alternatively, we can implement a member function (`is_member(epoch, node)`) on EpochValidators

can mark a TODO: and fix later

monad-peer-discovery/src/discovery.rs
```diff
@@ -226,30 +271,82 @@ impl<ST: CertificateSignatureRecoverable> PeerDiscovery<ST> {
...
+            self.peer_info.insert(node_id, name_record);
         }
 
         self.metrics[GAUGE_PEER_DISC_NUM_CONNECTED_PEERS] = self.peer_info.len() as u64;
 
         // send pings to newly modified/added peers
+        // full nodes only need to connect to a few validators so they don't need to immediately connect if incoming node is a validator
+        if self.self_role == Role::FullNode && self.is_validator(&node_id) {
+            return vec![];
+        }
```
michael-yxchen: some metrics ideas - we should have separate metrics for `(connection, name record) X (fullnode, validator)`, so we have observability into
- (correctness) full node discovering all validators' name records
- (correctness) full node only maintaining connections with a few validators
- (monitoring) how much resource is validator spending on fullnode connections
- (monitoring) name record number doesn't blow up

monad-peer-discovery/src/discovery.rs
```diff
@@ -226,30 +271,82 @@ impl<ST: CertificateSignatureRecoverable> PeerDiscovery<ST> {
...
+        let mut cmds = Vec::new();
+
+        // full node will try to connect to NUM_UPSTREAM_VALIDATORS
+        // if insufficient connected upstream validators, try to send pings to new validators
+        let connected_validators = self
+            .connection_info
+            .keys()
+            .filter(|&node_id| self.is_validator(node_id))
+            .collect::<HashSet<_>>();
+
```
michael-yxchen, line 311: we might want to record last_recv_ping in ConnectionInfo. That's the only proof that validator has added us to their connection list

monad-peer-discovery/src/discovery.rs
```diff
@@ -665,18 +781,27 @@ where
...
+            } else {
+                debug!(?node_id, "stop sending ping to unresponsive peer");
+                cmds.extend(self.clear_ping_timer(*node_id));
+                if !self.is_validator(node_id) && !self.pinned_full_nodes.contains(node_id) {
+                    // if it's a full node, also remove it from peer info
+                    self.peer_info.remove(node_id);
+                }
+            }
+        }
+        self.connection_info = retained;
```
michael-yxchen: can we prune in-place?

monad-peer-discovery/src/discovery.rs
```diff
@@ -665,18 +781,27 @@ where
...
+        for (node_id, info) in self.connection_info.iter() {
+            if info.unresponsive_pings < self.prune_threshold {
+                retained.insert(*node_id, *info);
+            } else {
+                debug!(?node_id, "stop sending ping to unresponsive peer");
+                cmds.extend(self.clear_ping_timer(*node_id));
+                if !self.is_validator(node_id) && !self.pinned_full_nodes.contains(node_id) {
+                    // if it's a full node, also remove it from peer info
+                    self.peer_info.remove(node_id);
+                }
```
michael-yxchen: can we document the relationship between `peer_info` and `connection_info`? i think in your impl. `connection` is always a subset of `peer`. It's a little hard to reason about because they are modified in different places

monad-peer-discovery/src/discovery.rs
```diff
@@ -747,12 +880,46 @@ where
...
+
+            // if a full node gets promoted to a validator, send pings to all validators in the current and next epoch
+            if self.self_role == Role::FullNode && self.is_validator(&self.self_id) {
+                self.self_role = Role::Validator;
+                if let Some(validators) = self.epoch_validators.get(&epoch).cloned() {
+                    for validator in validators {
+                        cmds.extend(self.send_ping(validator));
+                    }
+                }
+            }
```
michael-yxchen: say you get promoted to validator in epoch 10. The code is run when you enter epoch 9. You start sending ping to all epoch 9 validators. Should you send ping to epoch 10 validators too?

note: this is subject to change if we decide to lock e+1 (instead of e+2) validator set by boundary block of epoch e. A more reasonable place to trigger is on inserting validator set

monad-peer-discovery/src/discovery.rs
```diff
@@ -747,12 +880,46 @@ where
...
+                    .collect();
+
+                let keep = validators
+                    .into_iter()
+                    .choose_multiple(&mut self.rng, NUM_UPSTREAM_VALIDATORS);
+
+                for peer in self.peer_info.keys() {
+                    if !keep.contains(&peer) {
+                        cmds.extend(self.clear_ping_timer(*peer));
+                        self.connection_info.remove(peer);
```
michael-yxchen, line 926: nit: when cleaning up, we should bundle these two calls

docker/devnet/monad/config/node.toml
```diff
@@ -45,8 +45,8 @@ ping_period = 30
 refresh_period = 120
 request_timeout = 5
 prune_threshold = 5
-min_active_connections = 0
-max_active_connections = 200
+min_num_peers = 0
+max_num_peers = 200
```
michael-yxchen, line 49: we should be clear what these numbers include and any recommended values, maybe in that ops/release note page. e.g. i think `max_num_peers` should be minimum all validators + dedicated + prioritized fn.

monad-peer-discovery/src/discovery.rs
```diff
@@ -1284,22 +1445,120 @@ mod tests {
...
+            local_name_record: Some(generate_name_record(peer2, 0)),
+        });
+        assert!(state.peer_info.contains_key(&peer2_pubkey));
+        assert_eq!(state.peer_info.len(), 2);
+        match role {
+            Role::FullNode => {
+                // full node should not add peer2 to connection info and should not send ping
+                assert!(!state.connection_info.contains_key(&peer2_pubkey));
+                assert!(extract_ping(cmds).is_empty());
+            }
```
michael-yxchen: is that changed on refresh? looks like it'd want to replace peer1 with peer2 because peer2 is a validator

monad-peer-discovery/src/discovery.rs
eerkaijun: currently we use pong to indicate whether a validator has added us to their connection list (rather than ping from the validator)

monad-peer-discovery/src/discovery.rs
eerkaijun: refactored in https://github.com/category-labs/monad-bft/pull/1979
this logic is moved to `handle_ping_timeout` so we don't need to wait for refresh to prune

monad-peer-discovery/src/discovery.rs
eerkaijun: also refactored in https://github.com/category-labs/monad-bft/pull/1979 to a separate function
i guess we can move it to this PR if we want

monad-peer-discovery/src/discovery.rs
eerkaijun: yeah peer2 is added to `peer_info` but not immediately sending ping to peer2. upstream validators are selected during refresh and pings are sent then

docker/devnet/monad/config/node.toml
eerkaijun: yeah will document recommended values for peer discovery config

eerkaijun: tracking here https://github.com/category-labs/category-internal/issues/1758

monad-peer-discovery/src/discovery.rs
eerkaijun: updated, now documented in the struct definition

monad-peer-discovery/src/discovery.rs
eerkaijun: yeah it should also send pings to epoch 10 validators, i overlooked this

eerkaijun: is it guaranteed that validator set is inserted at specific period? (e.g. is it guaranteed that validator set of epoch n is inserted during epoch n-2?)

monad-peer-discovery/src/discovery.rs
eerkaijun: agree, some of the metrics is added in https://github.com/category-labs/monad-bft/pull/1979

requires a bit of refactor to add the remaining metrics because we currently store both full nodes and validators in `connection_info`, so getting the number of each role is requires iterating through `connection_info` rather than being a O(1) operation. will address separately if that's fine

monad-peer-discovery/src/discovery.rs
eerkaijun: I guess technically we can also send ping immediately after receiving the peer info, that might allow a full node to connect to an upstream faster

eerkaijun: updated to send ping immediately if there's insufficient upstream validators

monad-peer-discovery/src/discovery.rs
michael-yxchen: yeah pretty much always. delay period is 5000 rounds

eerkaijun: i updated to send pings to both current and next epoch validators, in `update_current_epoch` function. 

putting this logic when inserting validator set has some other implications, since when we do pruning or connections we check whether a node is a validator in current or next epoch, and upon inserting validator set the epoch has not incremented yet, so the pings can be subjected to pruning etc

monad-peer-discovery/src/discovery.rs
```diff
@@ -220,36 +266,89 @@ impl<ST: CertificateSignatureRecoverable> PeerDiscovery<ST> {
         node_id: NodeId<CertificateSignaturePubKey<ST>>,
```
michael-yxchen: super nit: can we rename this `peer_id`?

monad-peer-discovery/src/discovery.rs
```diff
@@ -220,36 +266,89 @@ impl<ST: CertificateSignatureRecoverable> PeerDiscovery<ST> {
...
-        self.metrics[GAUGE_PEER_DISC_NUM_CONNECTED_PEERS] = self.peer_info.len() as u64;
+        self.metrics[GAUGE_PEER_DISC_NUM_CONNECTED_PEERS] = self.routing_info.len() as u64;
 
         // send pings to newly modified/added peers
+        // full nodes only need to connect to a few validators so they don't need to immediately connect if incoming node is a validator
+        if self.self_role == Role::FullNode && self.is_validator(&node_id) {
+            return self.look_for_upstream_validators();
+        }
+
         self.send_ping(node_id)
```
michael-yxchen: note to self: no ping is sent to full nodes. fixed in followup pr

eerkaijun:
merging, CI failure unrelated to changes

PR #1959, opened 2025-07-07
Add node config linter tool

andr-dev:
Request from devops to check if config files are valid for any checked-out BFT. Produces a pretty error displaying what's wrong to help update config files.

Example:
```bash
$ ./monad-node-config-linter --config-type forkpoint forkpoint.toml
error: missing field `parent_id`
   ┌─ forkpoint.toml:6:1
   │
 6 │ ╭ [high_certificate.Qc.info]
 7 │ │ id = "0xfb8c9d462785086a041082b0f55e2c956d66f6ea64dc705eade29ec2ee9c4967"
 8 │ │ round = 1319681
 9 │ │ epoch = 27
10 │ │ block_round = 1319681
   │ ╰─────────────────────^
```
(has coloring too, just can't easily display on GitHub without image)

TODO: Add binary to docker build.


omegablitz:
How needed are the extra deps? What do the default serde errors look like?

PR #1961, opened 2025-07-08
Fix Raptorcast waker usage

xinyuan-dev:
Part 2 in https://github.com/category-labs/category-internal/issues/1727.

Issues not addressed in this PR:

- Deserializing in `poll_next`
  + supposedly the note suggests that deserialization is expensive and should be offloaded to dedicated blocking threads.
  + not sure how to fix this
- Using `try_recv` from sync channel in a `Stream` instead of an async variant
  + raptorcast uses std::sync::mpsc instead of the tokio's variant so there is no async variant. `try_recv` usage is fine here since it's non-blocking. it's unclear how to make the sender wake the receiving task with std::sync::mpsc.
  + Q: should we switch to tokio and use async mpsc for communication between primary and secondary?
  + implemented in https://github.com/category-labs/monad-bft/pull/1961/commits/3d2d705493144af1b40044da41c59f4b46fd6ef6. please review if such change is desirable.

### DIFF
--- a/monad-raptorcast/src/lib.rs
+++ b/monad-raptorcast/src/lib.rs
@@ -1,13 +1,11 @@
 use std::{
     collections::{BTreeMap, HashMap, VecDeque},
+    future::Future as _,
     marker::PhantomData,
     net::{SocketAddr, SocketAddrV4},
     ops::DerefMut,
     pin::{pin, Pin},
-    sync::{
-        mpsc::{Receiver, Sender},
-        Arc, Mutex,
-    },
+    sync::{Arc, Mutex},
     task::{Context, Poll, Waker},
     time::Duration,
 };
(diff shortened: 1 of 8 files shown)


monad-raptorcast/src/lib.rs
```diff
@@ -74,8 +73,8 @@ where
     dataplane: Arc<Mutex<Dataplane>>,
     pending_events: VecDeque<RaptorCastEvent<M::Event, ST>>,
 
-    channel_to_secondary: Option<Sender<FullNodesGroupMessage<ST>>>,
-    channel_from_secondary: Option<Receiver<Group<ST>>>,
+    channel_to_secondary: Option<UnboundedSender<FullNodesGroupMessage<ST>>>,
+    channel_from_secondary: Option<UnboundedReceiver<Group<ST>>>,
```
ariqchowdhury, line 78: should these channels be unbounded?

xinyuan-dev: I translated them from the original `std::sync::mpsc::channel`, which are unbounded. We can argue if we want to use bounded channel here, but if we decide to do so, I recommend we do that in a separate pr.

monad-raptorcast/src/lib.rs
```diff
@@ -1,13 +1,11 @@
 use std::{
     collections::{BTreeMap, HashMap, VecDeque},
+    future::Future as _,
```
fabiotdf, line 3: what does this do?

xinyuan-dev: `pin!(channel_from_secondary.recv()).poll(cx)` it's this line - calling `Future::poll` requires importing the trait.

PR #1967, opened 2025-07-08
Update rust.yml workflow to test self-hosted runner

rtsaimonad:


mhumeSF:
Runner does not look ready with the container fix suggested.

PR #1968, opened 2025-07-09
Include prioritized full nodes as pinned nodes in peer discovery

eerkaijun:
Prioritized full nodes should be pinned in peer discovery to prevents getting pruned

### DIFF
--- a/monad-raptorcast/src/raptorcast_secondary/publisher.rs
+++ b/monad-raptorcast/src/raptorcast_secondary/publisher.rs
@@ -288,12 +288,13 @@ where
     }
 
     // TODO: implement a command to update the always-ask full nodes?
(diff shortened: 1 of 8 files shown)


monad-raptorcast/src/raptorcast_secondary/publisher.rs
```diff
@@ -290,9 +290,9 @@ where
     #[allow(dead_code)]
     pub fn update_always_ask_full_nodes(
```
michael-yxchen, line 293: can we add a TODO: here to remind us to sync any changes to peer discovery?

eerkaijun: done

PR #1973, opened 2025-07-09
signing domains

omegablitz:


monad-secp/benches/secp.rs
```diff
@@ -1,9 +1,13 @@
...
-use monad_crypto::hasher::{Hasher, HasherType};
+use monad_crypto::{
+    hasher::{Hasher, HasherType},
+    signing_domain,
+};
 use monad_secp::{KeyPair, SecpSignature};
 use monad_testutil::signing::get_key;
 use rand::{thread_rng, RngCore};
 
+type SigningDomainType = signing_domain::Vote;
```
michael-yxchen: super nit: we only bls sign vote so i thought ConsensusMessage is more appropriate
```suggestion
type SigningDomainType = signing_domain::ConsensusMessage;
```

monad-secp/src/lib.rs
```diff
@@ -105,10 +108,14 @@ mod test {
...
     // valid certificate signature tests
-    use monad_crypto::certificate_signature::CertificateSignature;
+    use monad_crypto::{
+        certificate_signature::CertificateSignature,
+        signing_domain::{self},
+    };
 
     use crate::SecpSignature;
 
+    type SigningDomainType = signing_domain::Vote;
```
michael-yxchen:
```suggestion
    type SigningDomainType = signing_domain::ConsensusMessage;
```

Issue #1886, opened 2025-06-19
raptorcast: deserialization allows big compressed messages to be processed


robertkiel:
**Context:**

Raptorcast is using compressed messages. To estimate how much space the decompressed message will consume, the serialized message contains the size of the uncompressed message.

**The issue**

The [`decompressed_msg_len`](https://github.com/category-labs/monad-bft/blob/master/monad-raptorcast/src/message.rs#L159) property is under control of the adversary - which means the adversary can put arbitrarily high messages into the property and the decompression code will [allocate](https://github.com/category-labs/monad-bft/blob/master/monad-raptorcast/src/message.rs#L165) the given amount of memory using an underlying `Vec::with_capacity`

**Suggestion**

Limit the range of `decompressed_msg_len` to some sane values, e.g. a few megabytes or "usual_compression_rate * msg_len"

dshulyak:
there are actually 2 duplicates for this same issue in the internal bug tracker. but lets keep this open, i will make sure to close it once it is addressed

Issue #1954, opened 2025-07-07
dataplane: unverified memory allocation for TCP messages


robertkiel:
In the TCP receive part of `dataplane`, the receiving node reads a custom TCP header containing properties like magic bytes and the expected length of the payload.

Upon receiving the TCP packet, the node [allocates](https://github.com/category-labs/monad-bft/blob/4287361bfbe642f15847c92101d7a7347fee3853/monad-dataplane/src/tcp/rx.rs#L229) the specified memory up to a maximum of 1 GB. Although this memory gets eventually freed, this creates a vulnerability since the reception is unauthenticated, i.e. any node can send these messages, and it is sufficient to send a tiny message only containing the custom TCP header to provoke the allocation.

```rust
// message_length is taken from custom TCP header
let message = BytesMut::with_capacity(message_length);
```

### Suggestion

Do some sanity check before allocating the memory.

dshulyak:
we suppose to lower that amount to 1.5MB.
together with enforcing total connection limit and bounded tcp queue size we will put an upper bound for max allocated memory in tcp.

long term it would make sense to rely on io uring provided buffers / ringbuf and do not allocate at runtime, 
that would require significant changes though

robertkiel:
@dshulyak The issue here is that you could combine this one with #1955 and #1953 which means,

1. allocate memory that isn't freed
2. lock the worker thread by sending a short message #1955 
3. repeat this 100x per IP address as this is the maximum per IP #1953

dshulyak:
thanks, i've seen those issues.

i mentioned total cap on connections, it is not in the master, but very obvious that it should exist.
100x per ip is also too optimistic.

0xSqualo:
@dshulyak with #1977 merged I presume this issue can be closed - any objections?

PR #1977, opened 2025-07-10
dataplane, raptorcast: set max message size to 3MB

dshulyak:
related: https://github.com/category-labs/monad-bft/issues/1954
closes: https://github.com/category-labs/monad-bft/issues/1886
closes: https://github.com/category-labs/category-internal/issues/940

3MB limit is applied when receiving data from tcp socket,
and in raptorcast either for full payload or decompressed app message
  
beside changing limits, i updated deserialization error to use thiserror and
add several related tests

### DIFF
--- a/monad-raptorcast/src/message.rs
+++ b/monad-raptorcast/src/message.rs
@@ -404,4 +404,43 @@ mod tests {
             panic!("expected FinalMsgTooLarge error");
         }
     }
+
+    #[test]
+    fn test_outbound_message_serialize_snapshot() {
+        let msg = TestMessage { value: 42 };
+        let outbound_msg = OutboundRouterMessage::<TestMessage, SecpSignature>::AppMessage(msg);
+
+        let serialized = outbound_msg.try_serialize().unwrap();
+        let hex_encoded = hex::encode(&serialized);
+
+        insta::assert_snapshot!(hex_encoded);
+    }
+
+    #[test]
+    fn test_inbound_message_deserialize_snapshot() {
+        let msg = TestMessage { value: 12345 };
+        let outbound_msg = OutboundRouterMessage::<TestMessage, SecpSignature>::AppMessage(msg);
+
+        let serialized = outbound_msg.try_serialize().unwrap();
+        let hex_encoded_original = hex::encode(&serialized);
+
+        let deserialized =
+            InboundRouterMessage::<TestMessage, SecpSignature>::try_deserialize(&serialized)
+                .unwrap();
+
+        match deserialized {
+            InboundRouterMessage::AppMessage(decoded_msg) => {
+                assert_eq!(decoded_msg.value, 12345);
+
+                let reserialize_msg =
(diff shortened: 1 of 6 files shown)


monad-dataplane/src/tcp/tx.rs
```diff
@@ -231,6 +231,22 @@ async fn connect_and_send_messages(
...
+            warn!(
+                conn_id,
+                ?addr,
+                message_id,
+                message_len = len,
+                limit = TCP_MESSAGE_LENGTH_LIMIT,
+                "message exceeds size limit, skipping"
+            );
+            if let Some(completion) = msg.completion {
+                let _ = completion.send(());
```
omegablitz: Is it desired behavior to resolve the completion here if the message is never sent? In the normal send_message path, if the message transmission is interrupted, the completion isn't resolved.

Regardless, I don't think it matters that much because the message won't be able to be transmitted successfully anyways.

dshulyak: updated to not resolve a completion

dshulyak:
updated pr with snapshot tests for app message in raptorcast.rs. this will prevent accidental non-backward compatible changes in wire schema (such as with max_message variable)

monad-raptorcast/src/message.rs
```diff
@@ -8,7 +8,7 @@ use thiserror::Error;
 use super::raptorcast_secondary::group_message::FullNodesGroupMessage;
 
 const SERIALIZE_VERSION: u32 = 1;
-const MAX_MESSAGE_SIZE: usize = u32::MAX as usize;
+const MAX_MESSAGE_SIZE: usize = 3 * 1024 * 1024;
```
bamrith, line 11: Just to confirm, this is also the bound for `ForwardTxs` ?

dshulyak: i think ipc is limited to MAX_BATCH_LEN which is 128 txs.
they are also forwarded on poll_egress, it looks like that in that case it doesn't have a limit

dshulyak: i can add a limit here https://github.com/category-labs/monad-bft/blob/master/monad-eth-txpool-executor/src/forward.rs#L98 but 3MB is too much. even if such amount of txs available better to chunk them into 128-256KB parts

dshulyak: opened a separate pr to limit that to 256KB https://github.com/category-labs/monad-bft/pull/2001 . if there are more txs in pool that will be splited into many packets

dshulyak:
will merge https://github.com/category-labs/monad-bft/pull/2001 in a follow up once reviewed

monad-raptorcast/src/message.rs
andr-dev: Correct, egress did not have a limit. Good to merge for now, #2001 should be wrapped up quickly.

PR #1979, opened 2025-07-10
Pass confirm group to peer discovery for prune heuristics

eerkaijun:
Main changes:
- Pass confirm group to peer discovery. This help a node to prune full nodes that have not participated beyond a threshold (configurable in node.toml).
- Full nodes only send pings to a few upstream validators. Full nodes do not send pings to other full nodes

### DIFF
--- a/monad-node/src/main.rs
+++ b/monad-node/src/main.rs
@@ -170,6 +170,12 @@ async fn run(node_state: NodeState, reload_handle: ReloadHandle) -> Result<(), (
         .high_certificate
         .qc()
         .get_epoch();
+    let current_round = node_state
+        .forkpoint_config
+        .high_certificate
+        .qc()
+        .get_round()
+        + Round(1);
     let router: BoxUpdater<_, _> = {
         let raptor_router = build_raptorcast_router::<
             SignatureType,
@@ -184,6 +190,7 @@ async fn run(node_state: NodeState, reload_handle: ReloadHandle) -> Result<(), (
             &node_state.node_config.fullnode_dedicated.identities,
             locked_epoch_validators.clone(),
             current_epoch,
+            current_round,
         )
         .await;
 
@@ -509,6 +516,7 @@ async fn build_raptorcast_router<ST, SCT, M, OM>(
     full_nodes: &[FullNodeIdentityConfig<CertificateSignaturePubKey<ST>>],
     locked_epoch_validators: Vec<ValidatorSetDataWithEpoch<SCT>>,
     current_epoch: Epoch,
+    current_round: Round,
 ) -> MultiRouter<ST, M, OM, MonadEvent<ST, SCT, ExecutionProtocolType>, PeerDiscovery<ST>>
 where
     ST: CertificateSignatureRecoverable,
@@ -682,6 +690,7 @@ where
     let peer_discovery_builder = PeerDiscoveryBuilder {
         self_id,
         self_record,
+        current_round,
         current_epoch,
         epoch_validators,
         pinned_full_nodes,
@@ -689,7 +698,9 @@ where
         ping_period: Duration::from_secs(peer_discovery_config.ping_period),
         refresh_period: Duration::from_secs(peer_discovery_config.refresh_period),
         request_timeout: Duration::from_secs(peer_discovery_config.request_timeout),
-        prune_threshold: peer_discovery_config.prune_threshold,
+        unresponsive_prune_threshold: peer_discovery_config.unresponsive_prune_threshold,
+        last_participation_prune_threshold: peer_discovery_config
+            .last_participation_prune_threshold,
         min_num_peers: peer_discovery_config.min_num_peers,
         max_num_peers: peer_discovery_config.max_num_peers,
         rng: ChaCha8Rng::from_entropy(),
--- a/monad-peer-disc-swarm/src/driver.rs
+++ b/monad-peer-disc-swarm/src/driver.rs
@@ -152,13 +152,16 @@ where
                 target,
                 lookup_id,
             } => self.algo.handle_peer_lookup_timeout(to, target, lookup_id),
-            PeerDiscoveryEvent::UpdateCurrentEpoch { epoch } => {
-                self.algo.update_current_epoch(epoch)
+            PeerDiscoveryEvent::UpdateCurrentRound { round, epoch } => {
+                self.algo.update_current_round(round, epoch)
             }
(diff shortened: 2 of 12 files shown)


monad-peer-discovery/src/mock.rs
```diff
@@ -179,6 +183,16 @@ where
         Vec::new()
     }
 
+    fn update_peer_participation(
+        &mut self,
+        round: Round,
+        peers: BTreeSet<NodeId<CertificateSignaturePubKey<ST>>>,
+    ) -> Vec<PeerDiscoveryCommand<ST>> {
+        debug!(?round, ?peers, "updating peer participation");
+
```
fabiotdf, line 192: Is something missing here?

fabiotdf: Sorry my bad, this is the mock

monad-peer-disc-swarm/src/driver.rs
```diff
@@ -152,13 +152,19 @@ where
...
             PeerDiscoveryEvent::UpdateValidatorSet { epoch, validators } => {
                 self.algo.update_validator_set(epoch, validators)
             }
             PeerDiscoveryEvent::UpdatePeers { peers } => self.algo.update_peers(peers),
+            PeerDiscoveryEvent::UpdateConfirmGroup {
+                end_round,
+                confirm_group,
+            } => self
+                .algo
+                .update_peer_participation(end_round, confirm_group),
```
fabiotdf: This is so the peer discovery driver knows the round span that it will need to remember the IP addresses/peer info for, correct

fabiotdf: Can the pruning around `refresh()` be sensitive at startup when running as a full-node, since it does not know the current round until it starts receiving proposals?

monad-node/src/main.rs
```diff
@@ -166,6 +166,7 @@ async fn run(node_state: NodeState, reload_handle: ReloadHandle) -> Result<(), (
         .get_locked_validator_sets(&node_state.forkpoint_config);
 
     let current_epoch = node_state.forkpoint_config.high_qc.get_epoch();
+    let current_round = node_state.forkpoint_config.high_qc.get_round();
```
michael-yxchen: nit: i don't think it matters too much for correctness. for more context on consensus, current round is `high_qc.round + 1`. `EpochManager` translates that to epoch number

monad-peer-discovery/src/lib.rs
```diff
@@ -131,6 +132,10 @@ pub enum PeerDiscoveryEvent<ST: CertificateSignatureRecoverable> {
     UpdatePeers {
         peers: Vec<PeerEntry<ST>>,
     },
+    UpdateConfirmGroup {
+        end_round: Round,
+        confirm_group: BTreeSet<NodeId<CertificateSignaturePubKey<ST>>>,
```
michael-yxchen: nit: shall we rename it to `peers` to be consistent with `update_peer_participation` interface?

monad-peer-discovery/src/discovery.rs
```diff
@@ -100,8 +105,10 @@ pub struct PeerDiscovery<ST: CertificateSignatureRecoverable> {
     pub refresh_period: Duration,
     // duration before outstanding pings and lookup requests are dropped
     pub request_timeout: Duration,
-    // number of unresponsive pings allowed before being pruned
-    pub prune_threshold: u32,
+    // number of unresponsive pings allowed before dropping connection
+    pub unresponsive_prune_threshold: u32,
+    // number of rounds since last participation before pruning a full node
+    pub last_participation_prune_threshold: u64,
```
michael-yxchen: should we make it type `Round` instead?

monad-peer-discovery/src/discovery.rs
```diff
@@ -982,6 +1046,37 @@ where
...
+                .and_modify(|connection_entry| {
+                    if connection_entry.last_participation < round {
+                        connection_entry.last_participation = round;
+                    }
+                })
+                .or_insert_with(|| ConnectionInfo {
+                    last_ping: None,
+                    unresponsive_pings: 0,
+                    last_participation: round,
+                });
```
michael-yxchen, line 1072: when would we insert new connection info here? when a full node receives ConfirmGroup with new name records, those records get added to `routing_info`. should we create the "connection (no ping)" object there?

separate question: are `name_records` in confirm group msgs propagated to peer discovery?

eerkaijun: > are name_records in confirm group msgs propagated to peer discovery?

yes, secondary raptorcast calls PeerDiscoveryEvent::UpdatePeers to insert the name records

eerkaijun: > should we create the "connection (no ping)" object there?

UpdatePeers are also used elsewhere, like monad-debug-node to manually insert some peers, i think it's better not to introduce "round" to that event, which is why the events for passing confirm groups are separate

monad-node/src/main.rs
eerkaijun: updated to round + 1

monad-peer-disc-swarm/src/driver.rs
eerkaijun: currently we pass the round number provided in forkpoint to peer discovery on startup, which should be pretty close to the current round

monad-peer-discovery/src/discovery.rs
michael-yxchen: ok this is fine for now. i think the proper way is to have a dedicated event for inserting name records from ConfirmGroup

monad-peer-discovery/src/discovery.rs
```diff
@@ -654,76 +827,87 @@ where
...
-        };
+        // remove nodes that have not participated in secondary raptorcast beyond last_participation_prune_threshold
+        let non_participating_nodes: Vec<_> = self
+            .connection_info
+            .iter()
+            .filter(|(node_id, info)| {
+                self.current_round
+                    .0
+                    .saturating_sub(info.last_participation.0)
+                    >= self.last_participation_prune_threshold.0
```
```suggestion
                self.current_round.max(info.last_participation) - info.last_participation
                    >= self.last_participation_prune_threshold.0
```

monad-peer-disc-swarm/src/driver.rs
michael-yxchen: when you join the first group on startup, your `current_round` will be lower than `end_round`. Nothing would get pruned then

PR #1997, opened 2025-07-13
Relax statesync short-circuit to root executed

andr-dev:
Previously on startup, nodes would check if they had already finalized the current state root. This bound can be relaxed to checking whether the state root has been executed, as the proposed block will eventually get finalized after starting up. This change is expected to reduce the number of unnecessary statesyncs when restarting nodes.

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -234,22 +234,6 @@ where
     pub _phantom: PhantomData<CRT>,
 }
 
-/// Actions after state root validation
-#[derive(Debug)]
-pub enum StateRootAction {
-    /// StateRoot validation is successful, proceed to next steps
-    Proceed,
-    /// StateRoot validation is unsuccessful - there's a mismatch of StateRoots.
-    /// Reject the proposal immediately
-    Reject,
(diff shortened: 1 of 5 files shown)


omegablitz:
Depends on this stack for testing:
https://github.com/category-labs/monad-bft/pull/2196
https://github.com/category-labs/monad-bft/pull/2197
https://github.com/category-labs/monad-bft/pull/2198

The test in #2198 should be un-ignored in this PR (#1997)

monad-consensus-state/src/lib.rs
```diff
@@ -3381,33 +3365,37 @@ mod test {
             execution_delay,
         );
 
-        let mut wrapped_state = ctx[0].wrapped_state();
+        let n0 = &mut ctx[0];
 
         // prepare 5 blocks
+        let mut proposed_block_headers = BTreeMap::new();
```
omegablitz: I think it would read clearer to remove `proposed_block_headers`, and call `n0.ledger_propose` directly after `let _cmds = n0.handle_proposal_message(author, p)` for each proposal in the for loop.

omegablitz (approved):
LGTM, with one test nit

ariqchowdhury (approved):

PR #1998, opened 2025-07-14
Fix RaptorcastSecondary panic from malformed PrepareGroup messages

xinyuan-dev:
Fix two bugs reported by Asymmetric Research:

- raptorcast-secondary: FullNodesGroupMessage::PrepareGroup Bandwidth check bypass
- raptorcast-secondary: Single-packet DoS / Remote node panic via malformed group message

### DIFF
--- a/monad-raptorcast/src/raptorcast_secondary/client.rs
+++ b/monad-raptorcast/src/raptorcast_secondary/client.rs
@@ -11,7 +11,7 @@ use tracing::{debug, error, warn};
 
 use super::{
     super::{config::RaptorCastConfigSecondaryClient, util::Group},
-    group_message::{FullNodesGroupMessage, PrepareGroup, PrepareGroupResponse},
+    group_message::{ConfirmGroup, FullNodesGroupMessage, PrepareGroup, PrepareGroupResponse},
 };
 
 /// Metrics constant
@@ -21,7 +21,6 @@ pub const CLIENT_RECEIVED_INVITES: &str = "monad.bft.raptorcast.secondary.client
 pub const CLIENT_RECEIVED_CONFIRMS: &str =
     "monad.bft.raptorcast.secondary.client.received_confirms";
 
-type Bandwidth = u64;
 type GroupAsClient<ST> = Group<ST>;
 
 // This is for when the router is playing the role of a client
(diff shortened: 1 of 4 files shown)


monad-raptorcast/src/raptorcast_secondary/client.rs
```diff
@@ -126,232 +125,266 @@ where
...
+        if bandwidth.add_and_check(invite_msg.max_group_size).is_none() {
+            log_bandwidth_overflow();
+            return false;
+        };
+
+        // Check confirmed groups
+        for group in self
+            .confirmed_groups
+            .values(invite_msg.start_round..invite_msg.end_round)
+        {
```
xinyuan-dev, line 204: should we limit `invite_msg.end_round` to lie in a reasonable range, too? not sure if this loop may be used for ddos-ing by setting an astronomical `end_round`.

monad-raptorcast/src/raptorcast_secondary/publisher.rs
```diff
@@ -158,44 +160,41 @@ where
         self.group_schedule
             .retain(|_, group| group.end_round > round);
```
xinyuan-dev, line 170: since `end_round` is exclusive, should this condition be `group.end_round >= round`?

monad-raptorcast/src/raptorcast_secondary/client.rs
fabiotdf: Good point. I think in principle, yes a full-node could be denied service if invited into many astronomically long groups up to the point of exhausting its bandwidth config limit, so that it will stop accepting group invites from legit validators

monad-raptorcast/src/raptorcast_secondary/publisher.rs
fabiotdf: Yes, looks right

fabiotdf:
Where exactly was the panic coming from/what exactly fixes it?
Edit: [Here in `ext-audit-zellic` chat `raptorcast-secondary: Single-packet DoS / Remote node panic via malformed group message`] (https://category-labs.slack.com/archives/C094D71KMRC/p1752252600143239) is the original report

monad-raptorcast/src/raptorcast_secondary/client.rs
xinyuan-dev: I'm thinking about creating a new config entry for this purpose called something like `invite_round_span_max`. If the idea sounds reasonable to you, what do you suggest as the default value?

also, to be clear, I intend to implement the config in a separate pr, not in this one.

monad-raptorcast/src/raptorcast_secondary/publisher.rs
xinyuan-dev: sorry, i think i'm wrong. suppose we have a group whose `end_round == round`, then the group shouldn't be **retained** because it would be considered already ended at the current round, so the original condition is correct that we shouldn't retain such group.

monad-raptorcast/src/raptorcast_secondary/client.rs
fabiotdf: I think it sounds great.
Validators have a config entry `round_span` when configured to publish to full-nodes. This defines the target group size that the validator will be scheduling. The current default for this is 1 minute (given the current 500ms vote pace interval). Originally I had this value at 10 hours but that was considered way too long, so it was nerfed.

How about a default of 24h for the full-node's new config entry `invite_round_span_max`? Since this default value will also practically place an upper bound on `round_span` for validators

monad-raptorcast/src/raptorcast_secondary/client.rs
```diff
@@ -413,3 +442,109 @@ where
...
+            ),
+        );
+
+        let malformed_messages = [
+            // group size overflow
+            PrepareGroup {
+                max_group_size: usize::MAX,
+                validator_id: nid(2),
+                start_round: Round(6),
+                end_round: Round(7),
```
michael-yxchen: i noticed that span doesn't overlap with `confirmed_groups`. is that intended?

xinyuan-dev: you're right, it should really be something like `Round(5)..Round(6)`. although this test is more about the handling of the huge `max_group_size` instead. let me fix this before merging.

michael-yxchen: yeah then not overlapping is ok. i was just confused as to why `confirmed_groups` are populated

monad-raptorcast/src/raptorcast_secondary/client.rs
xinyuan-dev: implemented in https://github.com/category-labs/monad-bft/pull/2070.

PR #1999, opened 2025-07-14
Secondary raptorcast metrics

eerkaijun:
closes https://github.com/category-labs/category-internal/issues/1748

### DIFF
--- a/monad-raptorcast/src/raptorcast_secondary/publisher.rs
+++ b/monad-raptorcast/src/raptorcast_secondary/publisher.rs
@@ -6,6 +6,7 @@ use std::{
 use monad_crypto::certificate_signature::{
     CertificateSignaturePubKey, CertificateSignatureRecoverable,
 };
+use monad_executor::ExecutorMetrics;
 use monad_types::{NodeId, Round, RoundSpan};
 use rand::seq::SliceRandom;
 use rand_chacha::ChaCha8Rng;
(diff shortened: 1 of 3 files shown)


monad-raptorcast/src/raptorcast_secondary/publisher.rs
```diff
@@ -18,6 +19,11 @@ use super::{
     group_message::{ConfirmGroup, FullNodesGroupMessage, PrepareGroup},
 };
 
+/// Metrics constant
+pub const PUBLISHER_CURRENT_GROUP_SIZE: &str =
+    "monad.secondary_raptorcast.publisher_current_group_size";
+pub const PUBLISHER_SENT_INVITES: &str = "monad.secondary_raptorcast.publisher_sent_invites";
```
andr-dev: Can we prefix these as `monad.bft.raptorcast.secondary`? Also, if raptorcast also has two modes (publisher and client) we can namespace them accordingly `monad.bft.raptorcast.secondary.publisher`. Also, for the first one maybe something along the lines of `group_peers` is more descriptive.

monad-raptorcast/src/util.rs
```diff
@@ -264,6 +264,10 @@ where
         self.validator_id.as_ref().expect("Validator ID is not set")
     }
 
+    pub fn get_group_size(&self) -> usize {
```
fabiotdf: There's an explicit `size_excl_self()` . Mind you that `sorted_other_peers` do not include self

eerkaijun: oh yeah, didn't realize it's a duplicated function. thanks

PR #2001, opened 2025-07-14
limit forwarded txs from egress to 256KB

dshulyak:
related: https://github.com/category-labs/monad-bft/pull/1977

### DIFF
--- a/monad-eth-txpool-executor/src/forward.rs
+++ b/monad-eth-txpool-executor/src/forward.rs
@@ -21,6 +21,7 @@ const EGRESS_MAX_RETRIES: usize = 2;
 const INGRESS_CHUNK_MAX_SIZE: usize = 128;
 const INGRESS_CHUNK_INTERVAL_MS: u64 = 8;
 const INGRESS_MAX_SIZE: usize = 64 * 1024;
+const EGRESS_MAX_SIZE_BYTES: usize = 256 * 1024;
 
 #[pin_project(project = EthTxPoolForwardingManagerProjected)]
 pub struct EthTxPoolForwardingManager {
@@ -29,7 +30,7 @@ pub struct EthTxPoolForwardingManager {
     ingress_timer: tokio::time::Interval,
     ingress_waker: Option<Waker>,
 
-    egress: Vec<Bytes>,
+    egress: VecDeque<Bytes>,
     egress_waker: Option<Waker>,
 }
 
@@ -45,7 +46,7 @@ impl EthTxPoolForwardingManager {
             ingress_timer,
             ingress_waker: None,
 
-            egress: Vec::default(),
+            egress: VecDeque::default(),
             egress_waker: None,
         }
     }
@@ -95,7 +96,18 @@ impl EthTxPoolForwardingManager {
             return Poll::Pending;
         }
 
-        Poll::Ready(std::mem::take(egress))
+        let mut total_size = 0;
+        let mut drain_count = 0;
+
+        for (i, tx) in egress.iter().enumerate() {
+            let tx_size = tx.len();
+            if total_size + tx_size > EGRESS_MAX_SIZE_BYTES && i > 0 {
+                break;
+            }
+            total_size += tx_size;
+            drain_count = i + 1;
+        }
+        Poll::Ready(egress.drain(..drain_count).collect())
     }
 
     pub fn complete_ingress(self: Pin<&mut Self>) {
@@ -170,13 +182,15 @@ mod test {
 
     use alloy_consensus::{transaction::Recovered, Transaction, TxEnvelope};
     use alloy_primitives::{hex, B256};
+    use bytes::Bytes;
     use futures::task::noop_waker_ref;
     use itertools::Itertools;
     use monad_eth_testutil::{make_legacy_tx, recover_tx};
     use monad_eth_types::BASE_FEE_PER_GAS;
 
     use crate::forward::{
-        EthTxPoolForwardingManager, INGRESS_CHUNK_INTERVAL_MS, INGRESS_CHUNK_MAX_SIZE,
+        EthTxPoolForwardingManager, EGRESS_MAX_SIZE_BYTES, INGRESS_CHUNK_INTERVAL_MS,
+        INGRESS_CHUNK_MAX_SIZE,
     };
 
     // pubkey starts with AAA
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -17,3 +17,4 @@ test-keys
 .gdbinit
 /venv
 /logs
+.devcontainer
```
andr-dev: What's this for? Seems unrelated to this change.

monad-eth-txpool-executor/src/forward.rs
```diff
@@ -95,7 +96,19 @@ impl EthTxPoolForwardingManager {
             return Poll::Pending;
         }
 
-        Poll::Ready(std::mem::take(egress))
+        let mut total_size = 0;
+        let mut count = 0;
+
+        for tx in egress.iter() {
```
andr-dev: We can use `enumerate()` to avoid using a separate `count` variable.

monad-eth-txpool-executor/src/forward.rs
```diff
@@ -21,6 +21,7 @@ const EGRESS_MAX_RETRIES: usize = 2;
 const INGRESS_CHUNK_MAX_SIZE: usize = 128;
 const INGRESS_CHUNK_INTERVAL_MS: u64 = 8;
 const INGRESS_MAX_SIZE: usize = 64 * 1024;
+const EGRESS_MAX_SIZE_BYTES: usize = 256 * 1024;
```
andr-dev, line 24: I was told it's best not to split up payloads since raptorcast handles this. What's the benefit of using a smaller size here?

.gitignore
dshulyak: it is just containerized env that i am using for monad-bft development in vscode. i just thought that it is unnecessary to commit separately

monad-eth-txpool-executor/src/forward.rs
dshulyak: i thought of 2 things:

downloading 256KB is faster then 2MB, so if you send in batches higher chance that batch will end up in a block. if you send whole 2MB you need to download whole thing before you can add even one tx. the same is for signature verification, it is more efficient to start verification sooner for a batch and then download the next one in parallel

the second thing is that nodes are not cooperating, so what if everyone is sending 2MB worth of transactions at the same time. eventually there should be rate limiting to avoid congestion, but with smaller messages it is more likely that they will be downloaded in full

monad-eth-txpool-executor/src/forward.rs
```diff
@@ -95,7 +96,27 @@ impl EthTxPoolForwardingManager {
...
+            drain_count = i + 1;
+        }
+
+        let result: Vec<Bytes> = egress.drain(..drain_count).collect();
+
+        if !egress.is_empty() {
+            if let Some(waker) = egress_waker.take() {
+                waker.wake();
+            }
+        }
```
andr-dev: We don't need this wake here since producing `Poll::Ready(result)` will reschedule the `poll`.

monad-eth-txpool-executor/src/forward.rs
```diff
@@ -95,7 +96,27 @@ impl EthTxPoolForwardingManager {
...
+        for (i, tx) in egress.iter().enumerate() {
+            let tx_size = tx.len();
+            if total_size + tx_size > EGRESS_MAX_SIZE_BYTES && i > 0 {
+                break;
+            }
+            total_size += tx_size;
+            drain_count = i + 1;
+        }
+
+        let result: Vec<Bytes> = egress.drain(..drain_count).collect();
```
andr-dev: `drain` on VecDeque is sometimes inefficient because, from my experience and benchmarks, the compiler can't completely optimize for the range starting at `0`. Since it's a `VecDeque`, let's just create a `Vec` and `push` onto it.

```rust
let mut total_bytes = 0;
let mut egress_ret = Vec::default();

while !egress.is_empty() {
  let tx_size = egress.front().unwrap().len();
  if total_size + tx_size > EGRESS_MAX_SIZE_BYTES && i > 0 {
    break;
  }
  total_size += tx_size;
  egress_ret.push(egress.pop_front().unwrap());
}
```

monad-eth-txpool-executor/src/forward.rs
andr-dev: Agreed, 256KB seems like a reasonably large but not so large value. Let's keep it then!

monad-eth-txpool-executor/src/forward.rs
dshulyak: so what i wanted to avoid is growing vec multiple time. i will double check

monad-eth-txpool-executor/src/forward.rs
dshulyak: i don't think that runtimes are expected to do that. i was going over monoio for some time now. thats the run loop, that picks tasks added to the queue

https://github.com/bytedance/monoio/blob/master/monoio/src/runtime.rs#L157-L158
and this is what it polls
https://github.com/bytedance/monoio/blob/master/monoio/src/task/harness.rs#L53-L90

basically it won't reinsert itself into the task queue

dshulyak: it maybe the case with tokio, i didn't look deeply into that. but i assume there should be a specific contract as for what runtimes are expected to do, and reinserting ready task into task queue is not something i would expect

monad-eth-txpool-executor/src/forward.rs
dshulyak: so this is the bench that i made to check if it works the way i thought. it uses divan=0.1 lib

```
use std::collections::VecDeque;

#[global_allocator]
static ALLOC: divan::AllocProfiler = divan::AllocProfiler::system();

fn main() {
    divan::main();
}

#[divan::bench(sample_count = 1000)]
fn vecdeque_take_1000_elements(bencher: divan::Bencher) {
    bencher
        .with_inputs(|| {
            let mut deque = VecDeque::with_capacity(100_000);
            for i in 0..100_000 {
                deque.push_back(i);
            }
            deque
        })
        .bench_values(|mut deque| {
            let mut rst = vec![];
            for _ in 0..1000 {
                rst.push(deque.pop_front().unwrap());
            }
            rst
        });
}

#[divan::bench(sample_count = 1000)]
fn vecdeque_drain_1000_elements(bencher: divan::Bencher) {
    bencher
        .with_inputs(|| {
            let mut deque = VecDeque::with_capacity(100_000);
            for i in 0..100_000 {
                deque.push_back(i);
            }
            deque
        })
        .bench_values(|mut deque| {
            let vec: Vec<_> = deque.drain(..1000).collect();
            vec
        });
}
```

```
Timer precision: 17 ns
vecdeque_bench                   fastest       │ slowest       │ median        │ mean          │ samples │ iters
├─ vecdeque_drain_1000_elements  699.6 ns      │ 14.26 µs      │ 722.6 ns      │ 752.7 ns      │ 1000    │ 1000
│                                max alloc:    │               │               │               │         │
│                                  1           │ 1             │ 1             │ 1             │         │
│                                  4 KB        │ 4 KB          │ 4 KB          │ 4 KB          │         │
│                                alloc:        │               │               │               │         │
│                                  1           │ 1             │ 1             │ 1             │         │
│                                  4 KB        │ 4 KB          │ 4 KB          │ 4 KB          │         │
│                                dealloc:      │               │               │               │         │
│                                  1           │ 1             │ 1             │ 1             │         │
│                                  400 KB      │ 400 KB        │ 400 KB        │ 400 KB        │         │
╰─ vecdeque_take_1000_elements   950 ns        │ 1.973 µs      │ 1.009 µs      │ 1.033 µs      │ 1000    │ 1000
                                 max alloc:    │               │               │               │         │
                                   1           │ 1             │ 1             │ 1             │         │
                                   4.096 KB    │ 4.096 KB      │ 4.096 KB      │ 4.096 KB      │         │
                                 alloc:        │               │               │               │         │
                                   1           │ 1             │ 1             │ 1             │         │
                                   16 B        │ 16 B          │ 16 B          │ 16 B          │         │
                                 dealloc:      │               │               │               │         │
                                   1           │ 1             │ 1             │ 1             │         │
                                   400 KB      │ 400 KB        │ 400 KB        │ 400 KB        │         │
                                 grow:         │               │               │               │         │
                                   8           │ 8             │ 8             │ 8             │         │
                                   4.08 KB     │ 4.08 KB       │ 4.08 KB       │ 4.08 KB       │         │
```

---

drain is a bit faster, but more importantly i don't grow vector several times

dshulyak: and i increase it to take 50_000, the difference becomes more noticeable

```
Timer precision: 17 ns
vecdeque_bench                    fastest       │ slowest       │ median        │ mean          │ samples │ iters
├─ vecdeque_drain_50000_elements  34.04 µs      │ 98.94 µs      │ 34.05 µs      │ 34.26 µs      │ 1000    │ 1000
│                                 max alloc:    │               │               │               │         │
│                                   1           │ 1             │ 1             │ 1             │         │
│                                   200 KB      │ 200 KB        │ 200 KB        │ 200 KB        │         │
│                                 alloc:        │               │               │               │         │
│                                   1           │ 1             │ 1             │ 1             │         │
│                                   200 KB      │ 200 KB        │ 200 KB        │ 200 KB        │         │
│                                 dealloc:      │               │               │               │         │
│                                   1           │ 1             │ 1             │ 1             │         │
│                                   400 KB      │ 400 KB        │ 400 KB        │ 400 KB        │         │
╰─ vecdeque_take_50000_elements   42.99 µs      │ 115.8 µs      │ 72.03 µs      │ 69.93 µs      │ 1000    │ 1000
                                  max alloc:    │               │               │               │         │
                                    1           │ 1             │ 1             │ 1             │         │
                                    262.1 KB    │ 262.1 KB      │ 262.1 KB      │ 262.1 KB      │         │
                                  alloc:        │               │               │               │         │
                                    1           │ 1             │ 1             │ 1             │         │
                                    16 B        │ 16 B          │ 16 B          │ 16 B          │         │
                                  dealloc:      │               │               │               │         │
                                    1           │ 1             │ 1             │ 1             │         │
                                    400 KB      │ 400 KB        │ 400 KB        │ 400 KB        │         │
                                  grow:         │               │               │               │         │
                                    14          │ 14            │ 14            │ 14            │         │
                                    262.1 KB    │ 262.1 KB      │ 262.1 KB      │ 262.1 KB      │         │
```

monad-eth-txpool-executor/src/forward.rs
dshulyak: you were right it is not needed, it doesn't interact with runtime at all if Ready is returned repeatedly

andr-dev: Yeah this stuff is tricky, I only recently learned that you need to replace wakers on each call to `poll` because the task might get scheduled on another thread and there are some nuances with how the wakers work. In an ideal world, there would be a universal contract that all runtimes follow, but you might want to not do extra computations on specific runtimes because of unique guarantees.

PR #2002, opened 2025-07-14
Refactor message validation

michael-yxchen:
Publish original proposal message (with author signature) to full nodes

Supercedes/closes: https://github.com/category-labs/monad-bft/pull/1558 & https://github.com/category-labs/monad-bft/pull/1559

### DIFF
--- a/monad-state/src/consensus.rs
+++ b/monad-state/src/consensus.rs
@@ -6,7 +6,7 @@ use monad_consensus::{
         consensus_message::{ConsensusMessage, ProtocolMessage},
         message::ProposalMessage,
     },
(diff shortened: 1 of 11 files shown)


monad-state/src/consensus.rs
```diff
@@ -472,10 +476,10 @@ where
                 Vec::new()
             })?;
 
-        let (author, _, verified_message) = verified_message.destructure();
+        // FIXME: compile error
```
omegablitz: What's this for?

michael-yxchen: oops outdated

PR #2004, opened 2025-07-15
add signers length limit to BlsSignatureCollection

mikety:


omegablitz:
At a different layer, we should add a bound on the number of signatures in a SignatureCollection and/or TC.

We can do this in a separate PR

monad-bls/src/aggregation_tree.rs
```diff
@@ -16,6 +16,8 @@ use monad_types::NodeId;
 
 use crate::{bls::BlsAggregatePubKey, BlsAggregateSignature, BlsKeyPair, BlsSignature};
 
+const MAX_SIGNERS_LEN: usize = 16_384;
```
andr-dev: Super nit: I usually like expressing these as `16 * 1024` like you did in the tests because if you were to increase this to 32, 64, 96, or 128 I can't verify it's a power of 2 easily.

Issue #1985, opened 2025-07-11
[BUG] Trace Methods Accept Single Values Instead of Lists for Parameters


nicholas-osec:
The `trace_block`, `trace_call`, and `trace_transaction` methods handle input parameters as single values instead of lists like Geth.

Reported by OtterSec.

PR #2006, opened 2025-07-15
Removes unused code for rpc trace methods

philipglazman:
Also fixes https://github.com/category-labs/monad-bft/issues/1985


eerkaijun:
are we still planning to support these endpoints?

philipglazman:
Yes, just two. There is an open PR for an implementation for `trace_block` that is blocked on a change needed in execution. Blockaid asked for `trace_callMany` once but that is not yet implemented

ariqchowdhury:
`trace_block` is being removed here, but a separate PR (that's block on execution changes) will re-introduce it when ready, is that correct?

philipglazman:
Yep, I'll re-introduce `trace_block` after execution changes are merged

Issue #2007, opened 2025-07-15
tx-pool: panic when setting high `gas_limit` and high `max_fee_per_gas`


robertkiel:
Submitting a transaction through the transaction pool with high `gas_limit` and high `max_fee_per_gas` leads to a panic due to this [code](https://github.com/category-labs/monad-bft/blob/abff695ad7352e0bc274376169877b48cd317d82/monad-eth-block-policy/src/lib.rs#L71)

```rust
let gas_cost = U256::from(txn.gas_limit() as u128 * txn.max_fee_per_gas());
```

Logs from the fuzzer
```
thread '<unnamed>' panicked at /home/robert/.cargo/git/checkouts/monad-bft-7fa603b87da76344/41cc55f/monad-eth-block-policy/src/lib.rs:71:31:
attempt to multiply with overflow
```

PR #2008, opened 2025-07-15
monad-eth-block-policy: prevent overflow in compute max value

dshulyak:
closes: https://github.com/category-labs/monad-bft/issues/2007

it also adds proptest lib for property testing, ideally it would be used more.

in this particular case assertions (the property) are redundant, and can be removed but the test
is still useful as it explores all inputs


monad-eth-block-policy/src/lib.rs
```diff
@@ -67,10 +67,11 @@ fn compute_intrinsic_gas(tx: &TxEnvelope) -> u64 {
 
 #[allow(clippy::unnecessary_fallible_conversions)]
 pub fn compute_txn_max_value(txn: &TxEnvelope) -> U256 {
-    let txn_value = U256::try_from(txn.value()).unwrap();
```
dshulyak, line 85: i didn't understand why it had try_from

robertkiel: @dshulyak the code has many occurrences of unnessary `try_from().unwrap()`. Removing them could improve the readability of the code.

monad-eth-block-policy/src/lib.rs
```diff
@@ -67,10 +67,11 @@ fn compute_intrinsic_gas(tx: &TxEnvelope) -> u64 {
...
-    let gas_cost = U256::from(txn.gas_limit() as u128 * txn.max_fee_per_gas());
-
-    txn_value.saturating_add(gas_cost)
+    let txn_value = txn.value();
+    let gas_limit = U256::from(txn.gas_limit());
+    let max_fee = U256::from(txn.max_fee_per_gas());
+    let gas_cost = gas_limit.checked_mul(max_fee).expect("no overflow");
```
robertkiel, line 87: @dshulyak the issue comes from multiplying high values which lead to an overflow. Adding the `expect(...)` call will not prevent this `panic`

monad-eth-block-policy/src/lib.rs
```diff
@@ -67,10 +67,11 @@ fn compute_intrinsic_gas(tx: &TxEnvelope) -> u64 {
 
 #[allow(clippy::unnecessary_fallible_conversions)]
```
robertkiel: this is probably no longer necessary

monad-eth-block-policy/src/lib.rs
robertkiel: thinking of:
```rust
pub fn compute_txn_max_value(txn: &TxEnvelope) -> U256 {
    let txn_value: U256 = txn.value().into();
    let max_fee_per_gas = U256::from(txn.max_fee_per_gas());
    let gas_limit = U256::from(txn.gas_limit());

    let gas_cost = max_fee_per_gas.saturating_mul(gas_limit);

}
```

dshulyak: gas_cost is u128 * u64. this change converts them to U256 before multiplying, so for gas_cost expect will  not panic, which is also checked by the test below.

the value could be U256 though, i will adjust the test to use U256::MIN .. U256::MAX

PR #2010, opened 2025-07-15
Add exec event iter helpers to event ring library

andr-dev:


philipglazman (dismissed):
Worth adding examples for using the iter helpers

andr-dev:
Execution events library moving to `category-labs/monad`.

Issue #1986, opened 2025-07-11
[BUG] JSON-RPC ID Field Accepts Invalid Types and Missing Version Validation


nicholas-osec:
The current RPC requests accept complex types (lists, dictionaries) in the "id" field instead of only strings/numbers as per JSON-RPC spec. Furthermore, there is no validation that the "jsonrpc" field version number matches the expected version. 

Reported by OtterSec.

philipglazman:
PR https://github.com/category-labs/monad-bft/pull/2013

PR #2013, opened 2025-07-16
Stricter validation for jsonrpc requests

philipglazman:
Fixes https://github.com/category-labs/monad-bft/issues/1986


philipglazman:
Odd. Not sure why they would allow invalid jsonrpc requests, but making inputs stricter should not break compatability with tooling.

Issue #1989, opened 2025-07-11
[BUG] Transaction ChainId Not Validated Against Context


nicholas-osec:
The Monad RPC does not verify that the transaction’s chainId matches the current context’s chainId (`app_state.chain_id`). Instead, it only checks whether the chainId is one of `ETHEREUM_MAINNET`, `MONAD_DEVNET`, or `MONAD_TESTNET`.

Reported by OtterSec.

POC:
```
$ curl $GETH ... "method":"eth_estimateGas","params":[{"chainId":"0x1"}]}'
{"jsonrpc":"2.0","id":1234,"error":{"code":-32000,"message":"chainId does not match node's (have=1, want=1337)"}}
$ curl $MONAD ... "method":"eth_estimateGas","params":[{"chainId":"0x1"}]}'
{"jsonrpc":"2.0","result":"0xd0ac","id":1234}
$ curl $MONAD ... "method":"eth_estimateGas","params":[{"chainId":"0x4eaf"}]}'
{"jsonrpc":"2.0","result":"0xd0ac","id":1234}

$ curl $GETH ... "method":"eth_estimateGas","params":[{"chainId":"0x539"}]}'
{"jsonrpc":"2.0","id":1234,"result":"0xd221"}
$ curl $MONAD ... "method":"eth_estimateGas","params":[{"chainId":"0x539"}]}'
{"jsonrpc":"2.0","error":{"code":-32603,"message":"unsupported chain id","data":"1337"},"id":1234}
```

philipglazman:
PR https://github.com/category-labs/monad-bft/pull/2015

PR #2015, opened 2025-07-16
Validate transaction chain_id in rpc eth_call

philipglazman:
Fixes https://github.com/category-labs/monad-bft/issues/1989

### DIFF
--- a/monad-rpc/src/handlers/eth/call.rs
+++ b/monad-rpc/src/handlers/eth/call.rs
@@ -618,7 +618,14 @@ async fn prepare_eth_call<T: Triedb + TriedbPath>(
     )
     .await?;
 
-    if params.tx().chain_id.is_none() {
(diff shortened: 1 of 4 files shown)


ariqchowdhury:
do we also need to check this in the transaction submit path? It looks like we just verify if there is a chain_id but don't validate if its correct. Consensus later checks this in the static validation path, but its probably good to check in rpc so that the user gets the feedback immediately.

eerkaijun:
we used to call `static_validate_transaction` in rpc, not sure when it gets removed

edit: if i remember correctly, it's because we also do `static_validate_transaction` in txpool, and we are able to feedback that from the txpool to rpc

philipglazman:
Yes, it was moved to txpool and returns and validation errors back to the user. I added the chain_id validation back to rpc to exit early.

PR #2017, opened 2025-07-16
Shorten rounds

omegablitz:
round_timeout = 4 * delta + vote_delay = 400ms + 400ms = 800ms

### DIFF
--- a/monad-chain-config/src/revision.rs
+++ b/monad-chain-config/src/revision.rs
@@ -1,6 +1,6 @@
 use std::{fmt::Debug, time::Duration};
 
-pub const CHAIN_PARAMS_LATEST: ChainParams = CHAIN_PARAMS_V_0_8_0;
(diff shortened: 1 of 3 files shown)


monad-chain-config/src/revision.rs
```diff
@@ -56,6 +58,13 @@ const CHAIN_PARAMS_V_0_8_0: ChainParams = ChainParams {
     vote_pace: Duration::from_millis(500),
 };
 
+const CHAIN_PARAMS_V_0_10_0: ChainParams = ChainParams {
+    tx_limit: 5_000,
+    proposal_gas_limit: 150_000_000,
+    proposal_byte_limit: 2_000_000,
```
ariqchowdhury, line 64: should these values be adjusted as well?

omegablitz: I think we decided not to

Issue #1820, opened 2025-05-22
Missing zeroization for private keys in monad-secp


dnkolegov-ar:
secp256k1 library and `KeyPair` [wrapper](https://github.com/category-labs/monad-bft/blob/master/monad-secp/src/secp.rs#L10)  don't implement zeroization for private keys. Private keys remain in memory after use.
The criticality of this issue is low, but we recommend implementing zeroization to mitigate potential key leakage.

See the original [issue](https://github.com/rust-bitcoin/rust-secp256k1/issues/553).

**Solution**

Some projects use partial mitigation as [here](https://github.com/TDSoftware/substrate-ipfs/blob/5e0a0744c7b510a68c899ba98d7e73f04f90e691/primitives/core/src/ecdsa.rs#L567-L576).

dshulyak:
so practically in monad key can be zeroed only on the node shutdown?
i think in bitcoin it is used in wallet software, which is very different use case

dnkolegov-ar:
The idea of zeroization of the private key is to remove secret keys from memory as soon and robustly as possible in the case of software compromise and extracting sensitive information from memory, as in the Heartbleed attack.

dshulyak:
i understand what it means in general. i will try to elaborate, maybe i misunderstood something else

in monad-node this lib is used for signing network messages, so the private key needs to be kept in memory, or loaded from disk every time the message is signed. so practically when would you zero the memory of the key? thats why i said that it is likely means that it can be done only when monad-node shutdown.

in wallet software it is quite different, since you can zero the key after signing transaction. 

is there anything that can be done better, perhaps something that exists on solana/ethereum validator software?

dnkolegov-ar:
You’re right that a validator/node must keep its identity key resident for the entire process-lifetime.
What we can do is make sure the secret material is handled as defensively as possible while it is in memory and guarantee it’s scrubbed the moment the node no longer needs it.

dshulyak:
it is quite hard to get any guarantees here as discussed in bitcoin thread.
i think the best we can do is to copy secp256k1::KeyPair on heap (with box)  immediately after creation, and erase secp256k1::KeyPair that was on stack. then any moves should not affect boxed key pair, though i am not sure about that.

just doing partial mitigation as in https://github.com/TDSoftware/substrate-ipfs/blob/5e0a0744c7b510a68c899ba98d7e73f04f90e691/primitives/core/src/ecdsa.rs#L567-L576 seems as good as doing nothing

[...]

PR #2018, opened 2025-07-17
monad-secp: add manual zeroise for secret key

dshulyak:
closes: https://github.com/category-labs/monad-bft/issues/1820

this change implements recomendation from auditors. 
in general this doesn't guarantee that secret key will be reliably erased from memory, but some in some cases it will.

### DIFF
--- a/monad-secp/src/secp.rs
+++ b/monad-secp/src/secp.rs
@@ -3,7 +3,7 @@ use monad_crypto::{
     hasher::{Hasher, HasherType},
     signing_domain::SigningDomain,
 };
-use secp256k1::Secp256k1;
+use secp256k1::{ffi::CPtr, Secp256k1};
 use zeroize::Zeroize;
 
 /// secp256k1 public key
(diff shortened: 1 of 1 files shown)


ariqchowdhury:
might have been covered in the discussion with auditors, is there a reason we can't use the [zeroize](https://docs.rs/zeroize/1.8.1/zeroize/) crate?

dshulyak:
support for zeroize needs to be implemented by secp256k1 lib, we just can't do it with wrapper type. also i think secret key is some ffi type so they can't just use zeroize either. there is a link to this particular workaround in the gh issue

monad-secp/src/secp.rs
```diff
@@ -263,4 +278,16 @@ mod tests {
...
+    fn test_keypair_zeroize_on_drop() {
+        let mut privkey: [u8; 32] = [127; 32];
+        let keypair = KeyPair::from_bytes(&mut privkey).unwrap();
+
+        let pubkey = keypair.pubkey();
+
+        drop(keypair);
+        assert_eq!(privkey, [0; 32]);
+        assert_eq!(pubkey.bytes().len(), 65);
+    }
```
michael-yxchen: privkey material is zeroized in `KeyPair::from_bytes`. you want to assert that keypair memory are zeroized. I don't have a good idea on top of my head though..

dshulyak: right, drop doesn't dealloc so ptr will be valid. will just check that

dshulyak: i removed this test. there was a helper function that was developed for this purpose, so i just used in the Drop and custom workaround is not needed

Issue #1883, opened 2025-06-19
keystore: Invalid key range for sampling BLS keys


robertkiel:
The current keystore implementation first [samples](https://github.com/category-labs/monad-bft/blob/master/monad-keystore/src/main.rs#L101) a 12-word mnemonic and then derives a [subkey](https://github.com/category-labs/monad-bft/blob/master/monad-keystore/src/main.rs#L102) using BIP39 key derivation. This samples a `secp256k1` key which can not always be used as a `bls12_381` key. See key ranges below. When provisioning keys for validators in advance, users might end up with an invalid key or, when not filtered, with a weak instance of the crypto system.

Suggestion:

-  sample sepc256k1 keys in the correct range, e.g. using rejection sampling as seen [here](https://docs.rs/elliptic-curve/0.13.8/src/elliptic_curve/scalar/nonzero.rs.html#45). The correct range is from `1..0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141` (exclusive). For reference, see [Bitcoin Wiki](https://en.bitcoin.it/wiki/Secp256k1)
- sample bls12_381 keys in the correct range, e.g. using rejection sampling. The correct range is from `1..0x73eda753299d7d483339d80809a1d80553bda402fffe5bfeffffffff00000001` (exclusive). For reference, see [here](https://hackmd.io/@benjaminion/bls12-381#Curve-equation-and-parameters)

dshulyak:
i think it is not used directly as bls12_381 private key. rather it is used as a secret key material for generating bls12_381 key https://github.com/category-labs/monad-bft/blob/master/monad-bls/src/bls.rs#L281-L292 , which eventually passed to blast gen key.

i don't think that there is anything to fix here, do you agree?

robertkiel:
Cryptographically speaking, it is not the best design to sample keying material for a specific elliptic curve and then rely on a KeyGen mechanism to use the best bits out of it in order to sample a key for a different curve. By doing so, it is likely to unknowingly sacrifice some bits of security which should be avoided at all times.

On the other end, it might not be obvious that `KeyStore` is storing `secp256k1` private keys, and the input for the keygen mechanism for BLS keys.

robertkiel:
The simplest way would be storing randomly filled bits, e.g. `[u8;32]`, as `ikm` and then pass "the key" as IKM to the `BLS::from_bytes()` KeyGen function

PR #2019, opened 2025-07-17
keystore, secp: use ikm from mnemonic seed for bls and secp key derivation

dshulyak:
closes: https://github.com/category-labs/monad-bft/issues/1883

removed mnemonic and bip39 derivation, keystore gets 32 bytes from os that are used as secp and bls key material.

i added v1/v2 to preserve original behavior so that current testnet validators can use same tool


monad-secp/src/secp.rs
```diff
@@ -263,4 +281,56 @@ mod tests {
...
+        let mut max_value = [0xFFu8; 32];
+        let result = KeyPair::from_bytes(&mut max_value);
+        assert!(result.is_err());
+    }
+
+    proptest! {
+        #[test]
+        fn proptest_from_ikm(ikm: [u8; 32]) {
+            let keypair_result = KeyPair::from_ikm(&ikm);
+            prop_assert!(keypair_result.is_ok());
```
dshulyak: this can potentially fail if max number of iterations is not enough

monad-keystore/src/keystore.rs
```diff
@@ -19,6 +19,8 @@ use crate::{
 #[derive(Debug, Deserialize, Serialize)]
 #[serde(deny_unknown_fields)]
 pub struct Keystore {
+    #[serde(default = "default_version")]
+    pub version: u32,
```
dshulyak: can use enum with more descriptive names

monad-keystore/src/main.rs
```diff
@@ -98,28 +97,30 @@ fn main() {
             // create a new randomly generated mnemonic phrase
             let mnemonic = Mnemonic::new(MnemonicType::Words12, Language::English);
```
dshulyak: we can probably get rid of it alltogether, and use just use 32 random bytes

monad-secp/src/secp.rs
```diff
@@ -263,4 +279,99 @@ mod tests {
...
+        let result = KeyPair::from_bytes(&mut above_order);
+        assert!(result.is_err());
+
+        let mut max_value = [0xFFu8; 32];
+        let result = KeyPair::from_bytes(&mut max_value);
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn test_from_ikm_deterministic_profiles() {
```
dshulyak, line 368: this is to prevent accidental modifications in from_ikm

monad-secp/src/secp.rs
```diff
@@ -71,6 +68,24 @@ impl KeyPair {
         keypair
     }
 
+    pub fn from_ikm(ikm: &[u8]) -> Result<Self, Error> {
```
dshulyak, line 101: @robertkiel can you check please if this is reasonable way to derive secp from key material, or there is a better way to do it?

robertkiel: why don't do it the IETF way?
```rust
use k256::{Secp256k1, elliptic_curve::hash2curve::GroupDigest};

pub fn derive_key() {
    let dst = r"Monad key".as_bytes(); // something Monad-specific
    let ikm = [0u8; 32]; 
    let key = <Secp256k1 as GroupDigest>::hash_to_scalar::<ExpandMsgXmd<Sha256>>(&[ikm.as_slice()], &[dst]).unwrap();

    println!("key {:?}", key.to_bytes());
}
```
see https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-13.html#section-5

robertkiel: cc @dshulyak

dshulyak: thanks, will definitely use standard approach

omegablitz (approved):

dshulyak:
@robertkiel please take a look at the change. 
do you have other recommendations for generating secrets or deriving secp/bls keys?

monad-secp/src/secp.rs
```diff
@@ -96,6 +98,17 @@ impl KeyPair {
...
+        let dst = b"monad-ecdsa-keygen";
+        let scalar =
+            <K256Secp256k1 as GroupDigest>::hash_to_scalar::<ExpandMsgXmd<Sha256>>(&[ikm], &[dst])
+                .map_err(|_| Error(secp256k1::Error::InvalidSecretKey))?;
+        let mut scalar_bytes = scalar.to_bytes();
+        let result = Self::from_bytes(scalar_bytes.as_mut_slice());
+        scalar_bytes.zeroize();
```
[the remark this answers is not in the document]
robertkiel: Removing `scalar_bytes.zeroize()` might save some CPU cycles on a method that is rarely executed. However, missing a `zeroize()` invocation after a refactoring of `BLS::from_bytes()` will be worse.

monad-keystore/src/main.rs
```diff
@@ -221,3 +229,18 @@ fn print_public_key(private_key: &[u8], key_type: KeyType) {
         }
     }
 }
+
+fn print_public_key_v2(ikm: &[u8], key_type: KeyType) {
+    match key_type {
+        KeyType::Bls => {
+            let bls_keypair = BlsKeyPair::from_bytes(ikm.to_vec());
+            let pubkey = bls_keypair.unwrap().pubkey();
+            println!("BLS public key: {:?}", pubkey);
```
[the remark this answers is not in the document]
robertkiel: The `panic` is expected. If the key cannot be derived, the node can't start.

robertkiel:
> @robertkiel please take a look at the change. do you have other recommendations for generating secrets or deriving secp/bls keys?

The `keystore` module can use hash2curve as well for BLS keys, as stated in the IETF [proposal](https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-13.html#name-bls12-381-g1-2). However, it seems that the BLS library already does something similar [internally](https://github.com/supranational/blst/blob/4e58a5444298aaa89e9d44f6ee03228cc4e21ba0/src/keygen.c#L173) - following a similar, but different [IETF standard](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature#section-2.3). In this case, it makes sense - imo - to follow the BLS standard.

A small recommendation for the BLS derivation code path: Atm, the derivation function keeps the `key_info` empty. Similar to the Domain Separation Tag (DST) for the hash2curve primitive, this property is intended to prevent using the same key for different purposes. It is a good practice to set this to sth. Monad-specific, e.g. "monad-bls-privkey".

https://github.com/category-labs/monad-bft/blob/c81e53d5e1a8df91b94ff3dd49cec0bbc27292f8/monad-bls/src/bls.rs#L302

Another, but to my knowledge not yet widespread, approach is [ERC-2333](https://eips.ethereum.org/EIPS/eip-2333).

dshulyak:
> A small recommendation for the BLS derivation code path: Atm, the derivation function keeps the key_info empty. Similar to the Domain Separation Tag (DST) for the hash2curve primitive, this property is intended to prevent using the same key for different purposes. It is a good practice to set this to sth. Monad-specific, e.g. "monad-bls-privkey".

as i understand it will prevent accidental secret reuse, for example in different network that uses same bls protocols. indeed minor concern, but since i already updated secp i can update bls for consistency.

omegablitz (approved):

ariqchowdhury (approved):

PR #2021, opened 2025-07-17
Set lower redundancy factor for point-to-point targets

xinyuan-dev:
fixes: https://github.com/category-labs/category-internal/issues/985

todo: change the config entry usage in [monad-integration](https://github.com/search?q=repo%3Acategory-labs%2Fmonad-integration%20redundancy&type=code).

### DIFF
--- a/docker/devnet/monad/config/node.toml
+++ b/docker/devnet/monad/config/node.toml
@@ -27,7 +27,8 @@ chain_id = 20143
 #########################################################
 # Node-specific configuration
 #########################################################
-raptor10_validator_redundancy_factor = 3
+raptor10_validator_redundancy_factor = 3.0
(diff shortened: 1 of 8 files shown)


fabiotdf:
Doesn't this require changes to `udp::build_messages` as well?

xinyuan-dev:
@fabiotdf: i checked usage of `udp::build_message`. it appears that the method is only used in `RaptorCast::udp_build` and `RaptorCastSecondary::udp_build` aside from tests and benchmarking code. in this MR i just changed the corresponding usages in `RaptorCast::udp_build`. do you suggest we have a config for `RaptorCastSecondary` as well? p2p is indeed used in sending group messages.

PR #2023, opened 2025-07-17
Adapt split Dataplane in Raptorcast

xinyuan-dev:
https://github.com/category-labs/monad-bft/pull/1960 splits the dataplane into two halves: a singly owned reader and a cheaply cloned writer handle.

The split allows users of dataplane to be explicit on how they will use the dataplane, and in addition avoids unnecessary locks.

This PR adapts rc primary and rc secondary to use the split dataplane.

### DIFF
--- a/monad-raptorcast/src/lib.rs
+++ b/monad-raptorcast/src/lib.rs
@@ -21,7 +21,8 @@ use monad_crypto::certificate_signature::{
 };
 use monad_dataplane::{
     udp::{segment_size_for_mtu, DEFAULT_MTU},
-    BroadcastMsg, Dataplane, DataplaneBuilder, RecvTcpMsg, RecvUdpMsg, TcpMsg, UnicastMsg,
+    BroadcastMsg, DataplaneBuilder, DataplaneReader, DataplaneWriter, RecvTcpMsg, TcpMsg,
+    UnicastMsg,
 };
 use monad_executor::{Executor, ExecutorMetrics, ExecutorMetricsChain};
 use monad_executor_glue::{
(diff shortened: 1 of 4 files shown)


monad-raptorcast/src/lib.rs
```diff
@@ -102,7 +104,7 @@ where
 {
     pub fn new(
         config: config::RaptorCastConfig<ST>,
-        dataplane: Arc<Mutex<Dataplane>>,
+        dataplane: (DataplaneReader, DataplaneWriter),
```
eerkaijun: can we reuse the `Dataplane` struct here rather than using a tuple?

xinyuan-dev: if raptorcast primary takes the ownership of the `Dataplane`, then we still need to find a way to leak out dataplane writers for other consumers (i.e. rc secondary). one option is to export a `dataplane_writer` method from `RaptorCast` class, but i think doing that is a leaky abstraction since raptorcast primary should only care about its own usage of dataplane.

dshulyak: maybe (dataplane_reader, dataplane_writer): (...) and then use them by name below

xinyuan-dev: fixed & updated. thanks.

monad-raptorcast/src/lib.rs
```diff
@@ -579,7 +583,7 @@ where
         loop {
             let message: RecvUdpMsg;
             {
-                let mut dataplane = this.dataplane.lock().unwrap();
+                let dataplane = &mut this.dataplane_reader;
                 let Poll::Ready(msg) = pin!(dataplane.udp_read()).poll_unpin(cx) else {
                     break;
                 };
```
michael-yxchen: nit: we can probably drop the scope now. It was meant to drop the lock guard

PR #2025, opened 2025-07-17
Remove obsolete get_bft_block function

omegablitz:
This function is no longer supported in execution.


omegablitz:
https://github.com/category-labs/monad/pull/1415

PR #2026, opened 2025-07-17
Reserve balance validation

mikety:
Implement reserve balance validation (with max_reserve_balance set to 1 MON).
Allows to draw account balance to 0 if there were no transactions in the last `execution_delay` blocks


monad-cxx/monad-execution
mikety, line 1: This change is for testing only, it will not be merged

PR #2031, opened 2025-07-20
Reserve balance validation

mikety:
This is same as https://github.com/category-labs/monad-bft/pull/2026/ but rebase on top of v.0.10.0 and should have same result as current consensus due to settings that make it always  charge into reserve balance.


monad-blocktree/src/blocktree.rs
```diff
@@ -251,12 +251,16 @@ where
...
+                Ok(next_block) => {
                     self.tree
                         .set_coherent(&next_block_id, true)
                         .expect("should be in tree");
 
+                    // TODO add an indicator when block fees changed
+                    self.tree
+                        .replace_coherent_block(&next_block)
+                        .expect("should be in tree");
+
```
ariqchowdhury, line 263: seems a little strange to have check_coherency clone and return the block just to update the txn_fees field and then also replace_coherent_block and clone the block. 
Could check_coherency instead return the updated txn_fees and then have one update_coherent_block fn here that sets the flag and updates the txn_fees? 

I think the root problem is that EthValidatedBlock has the txn_fees field but static validation doesn't fully define what txn_fees is, but that's probably a much larger lift to refactor and resolve.

monad-chain-config/src/revision.rs
```diff
@@ -41,27 +43,31 @@ pub struct ChainParams {
...
     pub proposal_byte_limit: u64,
+    pub max_reserve_balance: u128,
     pub vote_pace: Duration,
 }
 
 const CHAIN_PARAMS_V_0_7_0: ChainParams = ChainParams {
     tx_limit: 10_000,
     proposal_gas_limit: 300_000_000,
     proposal_byte_limit: 4_000_000,
+    max_reserve_balance: 0,
```
ariqchowdhury: does the logic in eth_block_policy work if this is 0? How do we treat reserve balance of 0 throughout that code? Wouldn't it just reject everything?

monad-eth-block-policy/src/lib.rs
```diff
@@ -73,6 +80,10 @@ pub fn compute_txn_max_value(txn: &TxEnvelope) -> U256 {
     txn_value.saturating_add(gas_cost)
 }
 
+pub fn compute_txn_carriage_cost(txn: &TxEnvelope) -> U256 {
```
ariqchowdhury: nit: I don't see the term carriage_cost used in the spec or execution code anymore. Seems like its just called `max_gas_cost` in the execution branch. Can we change it to be consistent with their terminology?

monad-eth-block-policy/src/lib.rs
```diff
@@ -130,6 +141,27 @@ struct BlockLookupIndex {
     is_finalized: bool,
 }
 
+#[derive(Debug, Clone)]
+pub struct AccountBalanceState {
+    pub balance: Balance,
+    pub reserve_balance: Balance,
```
ariqchowdhury: nit: maybe `remaining_reserve_balance`?

monad-eth-block-policy/src/lib.rs
```diff
@@ -130,6 +141,27 @@ struct BlockLookupIndex {
...
+    pub max_reserve_balance: Balance,
+    pub last_txn_seqnum: SeqNum,
```
ariqchowdhury: I think renaming to `block_seqnum_of_latest_txn` would help clarify what this is. Or a comment

monad-eth-block-policy/src/lib.rs
```diff
@@ -303,21 +337,36 @@ where
...
     ) -> CommittedTxnFeeResult {
-        let mut txn_fee: Balance = Balance::ZERO;
+        let mut carriage_cost = Balance::ZERO;
+        let mut max_cost = Balance::ZERO;
         let mut next_validate = base_seq_num + SeqNum(1);
+        let mut last_txn_seq_num = GENESIS_SEQ_NUM;
+
+        // Compute balance from blocks before the base
+        let start_seq_num = next_validate.max(execution_delay) - execution_delay;
+        for (&cache_seq_num, block) in self.blocks.range(start_seq_num..next_validate) {
```
ariqchowdhury, line 349: Do you know when/where `self.blocks` gets populated? Its the `CommittedBlkBuffer`
Want to confirm if its populated after statesync; I'm guessing it must be. This was the part where I was wondering if statesync is handled correctly by reserve balance

monad-eth-block-policy/src/lib.rs
```diff
@@ -303,21 +337,36 @@ where
...
         eth_address: &Address,
+        execution_delay: SeqNum,
     ) -> CommittedTxnFeeResult {
-        let mut txn_fee: Balance = Balance::ZERO;
+        let mut carriage_cost = Balance::ZERO;
+        let mut max_cost = Balance::ZERO;
         let mut next_validate = base_seq_num + SeqNum(1);
+        let mut last_txn_seq_num = GENESIS_SEQ_NUM;
+
+        // Compute balance from blocks before the base
```
ariqchowdhury: is this comment correct? the below code iterates and modifies `last_txn_seq_num`, where does it compute any balances?

monad-eth-block-policy/src/lib.rs
```diff
@@ -378,6 +427,10 @@ where
 
     /// Chain ID
     chain_id: u64,
+
+    max_reserve_balance: Balance,
+
+    min_blocks_since_last_txn: SeqNum,
```
ariqchowdhury: this field needs a descriptive comment

monad-eth-block-policy/src/lib.rs
```diff
@@ -536,45 +592,85 @@ where
...
                     mut next_validate,
-                } = self.committed_cache.compute_txn_fee(base_seq_num, address);
+                } = self.committed_cache.compute_txn_fee(
+                    base_seq_num,
+                    address,
+                    self.execution_delay,
+                );
+
+                let mut can_charge_into_reserve = next_validate
+                    >= (balance_state.last_txn_seqnum + self.min_blocks_since_last_txn);
```
ariqchowdhury: I understand what this is doing, but I think it could be more clear. A comment quoting the line in the spec would probably be useful. Maybe another thing that would help would rename `can_charge_into_reserve` to `can_transfer_out_of_reserve_balance`. to match the language used in the spec

monad-eth-block-policy/src/lib.rs
```diff
@@ -536,45 +592,85 @@ where
...
         let account_balances = addresses
             .into_iter()
             .zip_eq(account_balances)
-            .map(|(address, mut account_balance)| {
+            .map(|(address, mut balance_state)| {
                 // Apply Txn Fees for the txns from committed blocks
                 let CommittedTxnFeeResult {
-                    txn_fee: txn_fee_committed,
+                    mut max_cost,
+                    mut carriage_cost,
```
ariqchowdhury: these are cumulative values right? As in, the cumulative carriage_cost of all the txns for an account in the blocks. Can we add `cumulative_` to the name?

monad-eth-block-policy/src/lib.rs
```diff
@@ -536,45 +592,85 @@ where
...
+                    mut last_txn_seq_num,
                     mut next_validate,
-                } = self.committed_cache.compute_txn_fee(base_seq_num, address);
+                } = self.committed_cache.compute_txn_fee(
```
ariqchowdhury, line 624: just to confirm, this always computes txn fees over the last execution_delay number of blocks?

monad-eth-block-policy/src/lib.rs
```diff
@@ -536,45 +592,85 @@ where
...
-                        consensus_block_seq_num,
-                        address
+                }
+
+                if !can_charge_into_reserve
+                    && (balance_state
+                        .balance
+                        .saturating_sub(balance_state.reserve_balance)
+                        < max_cost)
+                    || can_charge_into_reserve && (balance_state.balance < max_cost)
```
ariqchowdhury: this is confusing to read. can we separate these into separate bool statements and name them based on the condition in the spec that its checking

monad-eth-block-policy/src/lib.rs
```diff
@@ -583,37 +679,61 @@ where
...
                 }
 
-                if account_balance < txn_fee_pending {
+                if balance_state.reserve_balance < carriage_cost_pending {
                     panic!(
-                        "Majority extended a block with an incoherent balance \
-                            Not sufficient balance: {:?} \
-                            Txn Fees Pending: {:?} \
+                        "Pending block with insufficient reserve balance: {:?} \
+                            Transaction Carriage Cost Pending {:?} \
```
ariqchowdhury, line 702: I think we need to keep the language that "Majority extended an invalid block" because this panic/crash signals that the rest of network is behaving differently than the local node expects and we are halting progress here because of that divergence.

monad-eth-block-policy/src/lib.rs
```diff
@@ -583,37 +679,61 @@ where
...
+
+                can_charge_into_reserve = next_validate
+                    >= (balance_state.last_txn_seqnum + self.min_blocks_since_last_txn);
+
+                if !can_charge_into_reserve
+                    && (balance_state
+                        .balance
+                        .saturating_sub(balance_state.reserve_balance)
+                        < max_cost_pending)
+                    || can_charge_into_reserve && (balance_state.balance < max_cost_pending)
```
ariqchowdhury: same comments as above

monad-eth-block-policy/src/lib.rs
```diff
@@ -734,43 +866,71 @@ where
...
+            let max_cost = compute_txn_max_value(txn);
+            let carriage_cost = compute_txn_carriage_cost(txn);
+
+            if account_balance.reserve_balance >= carriage_cost
+                && (!can_charge_into_reserve
+                    && (account_balance
+                        .balance
+                        .saturating_sub(account_balance.reserve_balance)
+                        >= max_cost)
+                    || can_charge_into_reserve && (account_balance.balance >= max_cost))
```
ariqchowdhury: similar comment above

monad-eth-block-policy/src/lib.rs
ariqchowdhury, line 1: we need to add unit tests for check_coherency, compute_txn_fee etc that are specific to the reserve balance cases. It looks like only checks were updated for existing tests.

ariqchowdhury:
at a high level, how do we handle the case where there is a long chain of uncommitted blocks that is larger than the execution_delay? For example, every other round timesout, so we have 5 blocks but not consecutive so nothing committed yet. When the next block comes in for validation, the check_coherency function calculate the cumulative max_cost and carriage_cost of all the pending blocks?

monad-chain-config/src/revision.rs
mikety: Correct, "0" will reject everything. Can the chain revision be v0.70? Updated anyways to be the same across all revisions.

monad-eth-block-policy/src/lib.rs
mikety: Updated comment. This code allows to look back past the base block returned from triedb to check when the actual commit has happened.

monad-eth-block-policy/src/lib.rs
mikety: if there is enough committed blocks then yes. If not and there are pending blocks their fees will be included. Total should be over last execution_delay blocks split between committed and extending blocks

monad-blocktree/src/blocktree.rs
mikety: Right, the BlockPolicy::ValidatedBlock used in blocktree does not have txn_fees defined

monad-eth-block-policy/src/lib.rs
mikety: it computes the committed_cache blocks txn fees. then it computes over the remaining pending blocks.

PR #2032, opened 2025-07-21
Update Stake to U256

bamrith:
This PR converts Stake type from u64 to U256 to match with execution. Added a custom serializer and deserializer for U256

Updated existing tests. Added tests for the custom serializer and deserializer. Added test to confirm equivalent leader schedule until fork.

### DIFF
--- a/monad-blocksync/src/blocksync.rs
+++ b/monad-blocksync/src/blocksync.rs
@@ -29,13 +29,14 @@ use monad_crypto::certificate_signature::{
     CertificateSignaturePubKey, CertificateSignatureRecoverable, PubKey,
 };
 use monad_state_backend::StateBackend;
-use monad_types::{Epoch, ExecutionProtocol, NodeId, SeqNum};
+use monad_types::{Epoch, ExecutionProtocol, NodeId, SeqNum, Stake};
 use monad_validator::{
     epoch_manager::EpochManager,
     validator_set::{ValidatorSetType, ValidatorSetTypeFactory},
     validators_epoch_mapping::ValidatorsEpochMapping,
+    weighted_round_robin::{generate_random_validator_with_randomizer, randomize_256_with_rng},
 };
-use rand::{seq::SliceRandom, SeedableRng};
+use rand::{seq::SliceRandom, Rng, SeedableRng};
 use rand_chacha::ChaCha8Rng;
 use tracing::{debug, warn};
 
@@ -423,6 +424,14 @@ where
         cmds
     }
 
+    fn choose_weighted(
+        validators: Vec<(&NodeId<CertificateSignaturePubKey<ST>>, &Stake)>,
+        mut gen: impl Rng,
+    ) -> NodeId<CertificateSignaturePubKey<ST>> {
+        let randomizer = |total_stake| randomize_256_with_rng(&mut gen, total_stake);
+        generate_random_validator_with_randomizer(validators, randomizer)
+    }
+
     fn pick_peer(
         self_node_id: &NodeId<CertificateSignaturePubKey<ST>>,
         current_epoch: Epoch,
@@ -451,10 +460,7 @@ where
                 .filter(|(peer, _)| peer != &self_node_id)
                 .collect_vec();
             assert!(!members.is_empty(), "no nodes to blocksync from");
-            *members
-                .choose_weighted(rng, |(_peer, weight)| weight.0)
-                .expect("nonempty")
-                .0
+            Self::choose_weighted(members, rng)
         }
     }
 
--- a/monad-consensus/src/pacemaker.rs
+++ b/monad-consensus/src/pacemaker.rs
@@ -751,16 +751,16 @@ mod test {
         let mut staking_list = keys
             .iter()
             .map(|k| NodeId::new(k.pubkey()))
-            .zip(std::iter::repeat(Stake(1)))
+            .zip(std::iter::repeat(Stake::ONE))
             .collect::<Vec<_>>();
 
         // total stake = 7, f = 2
         // node1 holds f+1 stake
         // node1 + node2 has 2f+1 stake
-        staking_list[0].1 = Stake(1);
-        staking_list[1].1 = Stake(3);
-        staking_list[2].1 = Stake(2);
-        staking_list[3].1 = Stake(1);
+        staking_list[0].1 = Stake::ONE;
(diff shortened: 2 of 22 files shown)


monad-consensus/src/pacemaker.rs
```diff
@@ -748,16 +749,16 @@ mod test {
...
             .collect::<Vec<_>>();
 
         // total stake = 7, f = 2
         // node1 holds f+1 stake
         // node1 + node2 has 2f+1 stake
-        staking_list[0].1 = Stake(1);
-        staking_list[1].1 = Stake(3);
-        staking_list[2].1 = Stake(2);
-        staking_list[3].1 = Stake(1);
+        staking_list[0].1 = Stake(U256::ONE);
```
xinyuan-dev: can we hide the internal type used in stake from outside caller by exposing constants and conversion methods like `Stake::ZERO`, `Stake::ONE`, `Stake::MAX`, `Stake::from(any numeric type)`?

Also, if we impl `Add<Stake, Output=Stake>` and `Div<Stake, Output=f64>` for `Stake` type, then we may be able to make the users of the stake type completely agnostic of the stake's internal representation. An alternative approach is implement the `checked_{add,div}` methods for extra safety.

bamrith: updated

monad-blocksync/src/blocksync.rs
```diff
@@ -405,6 +406,41 @@ where
...
+                } else {
+                    None
+                }
+            })
+            .collect_vec();
+
+        let max = U256::MAX - (U256::MAX - running_stake + U256::from(1)) % running_stake;
+        let mut stake_index = U256::MAX;
+        while stake_index > max {
+            stake_index = rand::thread_rng().gen();
```
omegablitz: pass rng in as a parameter so that this is deterministic

monad-validator/src/weighted_round_robin.rs
```diff
@@ -76,7 +71,7 @@ impl<PT: PubKey> LeaderElection for WeightedRoundRobin<PT> {
             panic!("no validator has positive stake");
         }
 
-        let stake_index = randomize(round.0, total_stakes);
+        let stake_index = randomize_256(U256::from(round.0), total_stakes);
```
omegablitz: nit: change randomize_256 to take in a u64 for `x` so we don't need to upcast round.0

monad-validator/src/weighted_round_robin.rs
```diff
@@ -36,11 +36,6 @@ impl<PT: PubKey> Default for WeightedRoundRobin<PT> {
     }
 }
 
-fn randomize(x: u64, m: u64) -> u64 {
-    let mut gen = ChaCha20Rng::seed_from_u64(x);
-    gen.gen_range(0..m)
-}
-
 fn randomize_256(x: U256, m: U256) -> U256 {
```
omegablitz: Is this leader schedule equivalent for m <= u64::MAX? If not, we'll need to take care when rolling this out.

monad-blocksync/src/blocksync.rs
```diff
@@ -405,6 +406,41 @@ where
         cmds
     }
 
+    fn choose_weighted(
```
omegablitz, line 427: I see that this is mostly a duplicate of weighted_round_robin.rs. It'd be good to either deduplicate this code, or at least add a comment saying that it's mostly the same.

This is just because the code is sort of hard to grok, and so lowers cognitive overhead to know that it's the same as an existing implementation elsewhere.

bamrith: refactored

monad-types/src/lib.rs
```diff
@@ -489,7 +490,18 @@ impl Debug for BlockId {
     RlpEncodableWrapper,
     RlpDecodableWrapper,
 )]
-pub struct Stake(pub u64);
+pub struct Stake(pub U256);
```
omegablitz: If we implemented Deserialize manually, could we avoid needing to change validators.toml? Parse as either hex string or int

monad-blocksync/src/blocksync.rs
```diff
@@ -423,6 +424,14 @@ where
         cmds
     }
 
+    fn choose_weighted<R: Rng>(
+        validators: Vec<(&NodeId<CertificateSignaturePubKey<ST>>, &Stake)>,
+        mut gen: R,
```
omegablitz: super nit: `mut gen: impl Rng`

monad-types/Cargo.toml
```diff
@@ -11,10 +11,13 @@ bench = false
 [dependencies]
 monad-crypto = { workspace = true }
 
+alloy-primitives = { workspace = true, features = ["rlp"] }
 alloy-rlp = { workspace = true, features = ["derive"] }
 futures = { workspace = true }
 hex = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
+serde_json = { workspace = true }
+serde_test = { workspace = true }
```
omegablitz: nit: dev-dependency?

monad-blocksync/src/blocksync.rs
```diff
@@ -451,10 +460,7 @@ where
                 .filter(|(peer, _)| peer != &self_node_id)
                 .collect_vec();
             assert!(!members.is_empty(), "no nodes to blocksync from");
-            *members
-                .choose_weighted(rng, |(_peer, weight)| weight.0)
-                .expect("nonempty")
-                .0
+            Self::choose_weighted(members, rand::thread_rng())
```
omegablitz: can you pass in the rng parameter from this function (`rng: &mut ChaCha8Rng`)? this is still nondeterministic

monad-validator/src/weighted_round_robin.rs
```diff
@@ -41,8 +42,42 @@ fn randomize(x: u64, m: u64) -> u64 {
...
+                std::cmp::Ordering::Greater
+            } else {
+                std::cmp::Ordering::Less
+            }
+        })
+        .unwrap_err();
+    *stake_bounds[upper_bound].0
+}
+
+pub fn randomize_256_with_rng<R: Rng>(gen: &mut R, m: U256) -> U256 {
```
omegablitz: super nit: `gen: &mut impl Rng`

omegablitz (approved):

PR #2037, opened 2025-07-22
Peer discovery handle

xinyuan-dev:
Implements https://github.com/category-labs/category-internal/issues/1770

This PR move the shared peer-disc driver from the rc primary and secondary instance into router-multi. the interaction with peer-disc is abstracted into two parts:

- a handle for reading the peer address table and sending update events to peer-disc driver
- an emit_rx to receive events from peer-disc (send peer-disc router commands to dataplane)

The handle is optimized for fast peer address lookup and sending events to peer discovery driver without any lock contention.

The changes also remove the need for manual waker handling in the peer discovery driver.

How it's tested: unit tests on the interfacing between the handle and the driver.


dshulyak:
i haven't read the pr yet, just wanted to mention about

> The handle is optimized for fast peer address lookup and sending events to peer discovery driver without any lock contention.

monad-bft runs raptorcast, discovery and almost everything else on the same thread. so it is impossible to have lock contention there. i suspect locking was added as a workaround for borrow checker, or maybe there were some future plans.

xinyuan-dev:
> monad-bft runs raptorcast, discovery and almost everything else on the same thread. so it is impossible to have lock contention there. i suspect locking was added as a workaround for borrow checker, or maybe there were some future plans.

As far as we know, we make no assumption that these components is expected run single-threaded. So we may as well design the code with multithreading in mind. As a precedence, we already use multithreading in dataplane.

monad-raptorcast/src/lib.rs
```diff
@@ -821,8 +798,8 @@ where
         }
 
         {
```
eerkaijun: I think we can now remove this scope

monad-peer-discovery/src/discovery.rs
```diff
@@ -421,12 +420,46 @@ impl<ST: CertificateSignatureRecoverable> PeerDiscovery<ST> {
     }
 }
 
+impl<ST: CertificateSignatureRecoverable> PeerLookup for PeerDiscovery<ST> {
```
eerkaijun: we use peer lookup as a term for a node to request name records from other nodes in the network. we might want to rename this to something else since it's different in context

monad-peer-discovery/src/discovery.rs
```diff
@@ -421,12 +420,46 @@ impl<ST: CertificateSignatureRecoverable> PeerDiscovery<ST> {
...
+    ) -> HashMap<
+        NodeId<CertificateSignaturePubKey<Self::SignatureType>>,
+        MonadNameRecord<Self::SignatureType>,
+    > {
+        self.routing_info.clone().into_iter().collect()
+    }
+
+    fn is_validator(
+        &self,
+        node_id: &NodeId<CertificateSignaturePubKey<Self::SignatureType>>,
```
eerkaijun: there is another `is_validator` function in PeerDiscovery that indicates whether a node is a validator in the current and also next epoch. that's how peer discovery determines which nodes to accept and look up.

if we are only using this in `full_node_addrs` I think we can probably just keep it in the function itself.

monad-peer-discovery/src/lib.rs
```diff
@@ -273,40 +346,89 @@ pub trait PeerDiscoveryAlgo {
...
-        >,
+        Vec<PeerDiscoveryCommand<<Self::PeerDiscoveryAlgoType as PeerLookup>::SignatureType>>,
     );
 }
 
+#[derive(Debug, Clone)]
+pub struct PeerTable<ST: CertificateSignatureRecoverable> {
+    name_records: HashMap<NodeId<CertificateSignaturePubKey<ST>>, MonadNameRecord<ST>>,
+    // If None, all peers are considered validators.
+    validators: Option<HashSet<NodeId<CertificateSignaturePubKey<ST>>>>,
```
eerkaijun: i don't think we should use None to specify all peers are validators, seems error prone. afaik `PeerTable::from_name_records` is only used in mock? if so, I think we can do `validators: HashSet<NodeId<CertificateSignaturePubKey<ST>>>,` and the mock can adjust accordingly

monad-peer-discovery/src/lib.rs
```diff
@@ -194,9 +195,81 @@ pub enum PeerDiscoveryCommand<ST: CertificateSignatureRecoverable> {
...
+    ) -> bool;
+
+    fn lookup_addr_v4(
+        &self,
+        id: &NodeId<CertificateSignaturePubKey<Self::SignatureType>>,
+    ) -> Option<SocketAddrV4> {
+        self.lookup_name_record(id).map(|rec| rec.address())
+    }
+
+    fn known_addrs_v4(
```
eerkaijun: do we need known_addrs_v4 and full_node_addrs_v4?

eerkaijun (commented):
I think the architecture makes sense. Do we need to implement `PeerLookup` trait though for all the different structs, seems excessive? e.g. does PeerTable needs to implement PeerLookup, I would imagine it to be simpler without it.

monad-peer-discovery/src/lib.rs
xinyuan-dev: no, it was added because we do need lookup_addr_v4 so i considered exposing the address variants may be useful. now i removed these methods since there is no actual usage.

monad-peer-discovery/src/discovery.rs
xinyuan-dev: renamed into `NodeAddrLookup`.

monad-peer-discovery/src/discovery.rs
xinyuan-dev: the predicate `is_validator` doesn't actually make any sense without the context of the epoch we're interested in, so the right question to ask is what we want to do with the full nodes. so ideally we should implement the predicate with the epoch as input.

currently, it appears the list of full nodes from here is retrieved and used only in rc secondary publisher to determine the candidate for the next group, so probably we can change it such that secondary publisher just own a peer-disc handle and query the exact list of nodes and full nodes for the prospective epoch when it needs, so there will be no need to maintain a copy of full node lists from peer disc inside the publisher.

in this pr i'm going to implement this predicate (or at least whatever equivalent to this predicate) as is so to be compatible with exact behavior on the master, the mentioned change itself deserves its own pr if it's deemed necessary.

monad-peer-discovery/src/lib.rs
xinyuan-dev: done. please also see the notes in https://github.com/category-labs/monad-bft/pull/2037#discussion_r2275800147.

xinyuan-dev:
> I think the architecture makes sense. Do we need to implement `PeerLookup` trait though for all the different structs, seems excessive? e.g. does `PeerTable` needs to implement `PeerLookup`, I would imagine it to be simpler without it.

PeerLookup (now renamed as NodeAddrLookup) is expected to be implemented by any sources that can offer peer information. if we think about it this way, there are currently three such information sources with different semantics:

- PeerDiscovery (the algo): offer the freshest peer info
- PeerDiscoveryHandle: offer a mostly up to date peer info that is updating every moment or so
- PeerTable: a static snapshot of peer info, unchanging

Although PeerTable is only used as an internal type by other parts, it has all the qualities as such information source as a NodeAddrLookup instance.

PR #2041, opened 2025-07-23
monad-dataplane: add udp socket for explicit direct messages

dshulyak:
dataplane listens on two sockets and tags traffic that it sends to the consumer.
for sending we just select direct socket as source address if user specifies it over api, but otherwise the destination is selected in raptorcast

### DIFF
--- a/monad-dataplane/src/lib.rs
+++ b/monad-dataplane/src/lib.rs
@@ -38,6 +38,8 @@ pub(crate) mod buffer_ext;
 pub mod tcp;
 pub mod udp;
 
+pub(crate) use udp::UdpMessageType;
+
 pub struct DataplaneBuilder {
     local_addr: SocketAddr,
     trusted_addresses: Vec<IpAddr>,
@@ -46,6 +48,7 @@ pub struct DataplaneBuilder {
     udp_buffer_size: Option<usize>,
     tcp_config: TcpConfig,
     ban_duration: Duration,
+    direct_socket_port: Option<u16>,
 }
 
 impl DataplaneBuilder {
@@ -64,6 +67,7 @@ impl DataplaneBuilder {
                 per_ip_connections_limit: 100,
             },
             ban_duration: Duration::from_secs(5 * 60), // 5 minutes
+            direct_socket_port: None,
         }
     }
 
@@ -96,6 +100,12 @@ impl DataplaneBuilder {
         self
     }
 
+    /// with_direct_socket configures an additional UDP socket for direct peer communication
+    pub fn with_direct_socket(mut self, port: u16) -> Self {
+        self.direct_socket_port = Some(port);
+        self
+    }
+
     pub fn build(self) -> Dataplane {
         let DataplaneBuilder {
             local_addr,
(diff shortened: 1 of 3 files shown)


monad-dataplane/tests/tests.rs
```diff
@@ -108,6 +108,67 @@ fn udp_unicast() {
...
+
+    let mut rx = DataplaneBuilder::new(&rx_addr, UP_BANDWIDTH_MBPS)
+        .with_direct_socket(rx_direct_port)
+        .build();
+    let tx = DataplaneBuilder::new(&tx_addr, UP_BANDWIDTH_MBPS)
+        .with_direct_socket(tx_direct_port)
+        .build();
+
+    // Allow Dataplane threads to set themselves up.
+    sleep(Duration::from_millis(10));
```
dshulyak: after rebase can busy wait

monad-dataplane/src/udp.rs
```diff
@@ -44,88 +50,157 @@ pub const DEFAULT_SEGMENT_SIZE: u16 = segment_size_for_mtu(DEFAULT_MTU);
...
+}
 
-    let raw_fd = udp_socket_rx.as_raw_fd();
-    let udp_socket_tx =
-        UdpSocket::from_std(unsafe { std::net::UdpSocket::from_raw_fd(raw_fd) }).unwrap();
+fn create_direct_socket_pair(
+    local_addr: SocketAddr,
+    port: u16,
+    buffer_size: Option<usize>,
+) -> (Option<UdpSocket>, Option<UdpSocket>) {
```
xinyuan-dev: if these two values are never `None`, how about remove the `Option<...>`?

monad-dataplane/src/lib.rs
```diff
@@ -83,6 +94,7 @@ impl DataplaneBuilder {
 
                         udp::spawn_tasks(
                             local_addr,
+                            direct_socket_port,
```
xinyuan-dev: the direct udp is identical to the normal (or broadcast) udp port in all senses, then how about we reuse the existing `udp::spawn_tasks` with different `local_addr`, `ingress_tx` and `egress_rx`? this way we can even configure different buffer size/bandwidth if we want. the two ports do in fact carry different protocols so i think it's reasonable to place the three protocols (tcp/raptorcast_udp/direct_udp) in juxtaposition.

also, it'll be handy to have three pairs of tx/rx channels if we want to deal with starvation issue or prioritization in the future.

dshulyak: bandwidth should be shared between them, instead of configured separately, with consensus traffic prioritized (regardless of which socket it goes through). i think there is no reason to split that into two different tasks, the only thing changes is which source port we use for traffic

for reading i can switch to different channels, though i think there will be no practical implications

dshulyak: i recalled why i kept single udp channel, udp traffic should be always prioritized over tcp in current protocol regardless of socket, hence if i add two channels it becomes harder to do so with select

xinyuan-dev: we can tweak priority as we wish if necessary but i'd prefer it to be more flexible than baked in to that. also, there can be prioritization needs between the two udp protocols, it'll be nice if we have have the priority defined at one single location.

as for the bandwidth, if we're considering the physical bandwidth of the interface, then shouldn't we also take tcp into consideration when calculating the threshold? otherwise, it maybe inconsistent that we only consider udp for bandwidth. also, having three channels may make it easier to implement a centralized module that handles bandwidth and prioritization alike.

dshulyak: ok i will define another channel for reading. technically i can define for writer as well without changing much, but i will keep writing from the same task, so i would rather not

---

in tcp there is a congestion control, so it will not overrun provided bandwidth. but in that case what is important is to prioritize udp traffic over tcp using non-default tc queues. i tested that with prio queue and some programs that share udp/tcp.

as for global tcp/udp pacing, maybe it makes sense but it is definitely unrelated to this change.

omegablitz:
Is the idea that once this change is rolled out, `RouterTarget:PointToPoint` would be switched from udp_write_unicast -> udp_write_direct?

Would we tag `UdpMessageType::Broadcast` for all consensus traffic? Or only `RouterTarget::Broadcast` traffic (eg proposals)

dshulyak:
i will merge to avoid rebases then.

on that question of partitioning, what is important from "perf" perspective is to avoid raptorcast overhead for messages that do not require it. all point to point messages, regardless of origin.

the most straighforward way to achieve that is to move all point to point messages to different socket. we may consider adding more sockets if in addition to that we want to differentiate between consensus point to point, transactions and discovery.

PR #2042, opened 2025-07-23
monad-discovery: extend name record with ports and capabilities

dshulyak:
this change replaces single socket addr with the following fields:
- ip
- ports
- capabilities

ports is a collection, limited to 8 items, with port for tcp, udp.

capabilities is meant to be used for rolling upgrades. so for example
with direct socket, we will be able to use plain protocol selectively
and not necessarily at the same time when socket is enabled.

the other use case that i foresee in the future is the support of
authenticated protocol for udp connection, that would also benefit
from negotiated capability

it doesn't switch to v2 immediately, but makes it possible to
decode it. in decode method of the NameRecord we try to decode v2 first
and if it fails we will try v1. NameRecord::new is using v1, in the
follow up change it will be switched to v2. And then in the next follow
up v1 will be scraped from the codebase.


dshulyak:
@michael-yxchen @eerkaijun take a look please, i think i can make it backward compatible with some hacks in decoding, but just want to confirm if you are ok with this

dshulyak:
- update monad-integration

```
    def _sign_name_record(self, ip: str, private_key: bytes, port: int, seq_num: int = 0) -> str:
        """Sign the name record for peer discovery."""
        ip_bytes = ipaddress.IPv4Address(ip).packed

        name_record = NameRecord(ip_bytes, port, seq_num)
        encoded = rlp.encode(name_record)

        message_hash = bytes.fromhex(blake3.blake3(encoded).hexdigest())
        keys = KeyAPI()
        key = keys.PrivateKey(private_key)
        return str(keys.ecdsa_sign(message_hash, key))
```

eerkaijun:
i think this seems fine to me. 

> so for example with direct socket, we will be able to use plain protocol selectively and not necessarily at the same time when socket is enabled.

can you elaborate what this mean?

dshulyak:
i meant nodes will be able to update at a different time. 

so if node A updated and tries to communicate with node B that didn't update, it will understand it based on capability and will be using original protocol and maybe even original socket (non-direct).

but if A tries communicate with node C that updated and it may use both direct socket and plain protocol

monad-peer-discovery/src/lib.rs
```diff
@@ -21,37 +21,135 @@ pub mod mock;
...
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct NameRecord {
-    pub address: SocketAddrV4,
+    pub ip: Ipv4Addr,
+    pub ports: Vec<Port>,
+    pub capabilities: u64,
     pub seq: u64,
 }
```
xinyuan-dev: how about we put the ports to be contained in a struct like this:

```
struct PortMap {
  udp_raptorcast: u16,
  udp_direct: u16,
  tcp_raptorcast: u16,
  tcp_direct: u16,
  ...
}
```

this way accessing the ports will be easier than traversing/filtering the vec. we can use `Option` fields to indicate certain capabilities are optional.

monad-peer-discovery/src/lib.rs
```diff
@@ -21,37 +21,135 @@ pub mod mock;
...
+            _ => None,
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct NameRecord {
-    pub address: SocketAddrV4,
+    pub ip: Ipv4Addr,
+    pub ports: Vec<Port>,
```
xinyuan-dev: just a minor thing: if we're imposing the limit of a small maximum number of ports, we can consider using smallvec to avoid unnecessary allocation.

monad-peer-discovery/src/lib.rs
dshulyak: the idea behind encoding them with vec is to avoid versioning if new port needs to be added.
we can store them in such struct after parsing

xinyuan-dev: how about we only convert to/from the temporary types in `impl Encodable/Decodable`, this way we can also leverage the error handling of the decoder.

monad-peer-discovery/src/lib.rs
```diff
@@ -21,37 +21,135 @@ pub mod mock;
...
+        })
     }
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, RlpEncodable, RlpDecodable, Eq)]
+impl NameRecord {
+    pub fn validate(&self) -> bool {
+        let has_tcp = self.ports.iter().any(|p| p.tag == PortTag::TCP as u8);
+        let has_udp = self.ports.iter().any(|p| p.tag == PortTag::UDP as u8);
+        has_tcp && has_udp
```
xinyuan-dev: do we allow multiple tcp ports? from the methods below it seems we only care about first one.

dshulyak: i think all tags should be unique, so this validation can be definitely improved

dshulyak:
so regarding backward compatibility the simplest i could come up with is to introduce PingV1/V2 . V1 will use original format recognizable by old and new software, V2 will use new format recognizable only be new software. it creates a bit of the complexity around managing two versions (e.g they can't both be sent in lookup responses)

monad-peer-discovery/src/lib.rs
dshulyak: good approach, i will update this change to use it

monad-peer-discovery/src/lib.rs
dshulyak: improved validation in last changes

monad-peer-discovery/src/lib.rs
dshulyak: updated to use smallvec

dshulyak: i reverted to vec for simplicity

monad-peer-discovery/src/lib.rs
dshulyak: i realized that it doesn't work. we need to preserve unknown ports as they are used for signature. if we discard them we can't send valid record to peers

eerkaijun:
for backwards compatibility, i think we might need to break this into multiple releases if we want an asynchronous upgrade. 
- first upgrade will introduce new decoding for name record V2
- second upgrade to actually switch name record to V2
- (not essential) third upgrade to retire name record V1

or is there any other way you are thinking?

dshulyak:
@eerkaijun i wanted to make a breaking change, if it will land before mainnet.

i explored a bit how to make it work with rolling upgrade, i just wanted to avoid that. 

rlp is just not good codec for doing such things, and it will require either v2 everything (e.g ping/lookup) that works with new structure. so v1 and v2 can both work at the same time during an upgrade period.

or mix v1 and v2 together, e.g try both v1 and v2 decoding path. ping/lookup will use this mixed encoding. this is generally a hack, but will be less changes

dshulyak:
@eerkaijun i tried the 2nd approach that i described, it is simple enough and enables rolling ugprade. so i will push it and follow up with 3 step upgrade

monad-peer-discovery/src/discovery.rs
```diff
@@ -2225,11 +2225,15 @@ mod tests {
     const OLD_ADDR: SocketAddrV4 = SocketAddrV4::new(Ipv4Addr::new(7, 7, 7, 7), 8000);
     const NEW_ADDR: SocketAddrV4 = SocketAddrV4::new(Ipv4Addr::new(8, 8, 8, 8), 8000);
 
-    #[test_case(None, NameRecord { address: NEW_ADDR, seq: 1 }, true, NameRecord { address: NEW_ADDR, seq: 1 }, true; "first record")]
-    #[test_case(Some(NameRecord { address: OLD_ADDR, seq: 1 }), NameRecord { address: NEW_ADDR, seq: 2 }, true, NameRecord { address: NEW_ADDR, seq: 2 }, true; "newer record")]
-    #[test_case(Some(NameRecord { address: OLD_ADDR, seq: 1 }), NameRecord { address: OLD_ADDR, seq: 1 }, true, NameRecord { address: OLD_ADDR, seq: 1 }, false; "same record")]
-    #[test_case(Some(NameRecord { address: NEW_ADDR, seq: 2 }), NameRecord { address: OLD_ADDR, seq: 1 }, false, NameRecord { address: NEW_ADDR, seq: 2 }, false; "older record")]
-    #[test_case(Some(NameRecord { address: OLD_ADDR, seq: 1 }), NameRecord { address: NEW_ADDR, seq: 1 }, false, NameRecord { address: OLD_ADDR, seq: 1 }, false; "conflicting record")]
+    fn make_name_record(addr: SocketAddrV4, seq: u64) -> NameRecord {
+        NameRecord::new(*addr.ip(), addr.port(), seq)
```
dshulyak: this can be inlined

monad-peer-discovery/src/lib.rs
```diff
@@ -40,37 +40,308 @@ pub mod mock;
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+enum VersionedNameRecord {
+    V1(WireNameRecordV1),
+    V2(WireNameRecordV2),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct NameRecord {
+    record: VersionedNameRecord,
+}
```
dshulyak, line 268: VersionedNameRecord can be renamed to NameRecord

monad-peer-discovery/src/lib.rs
```diff
@@ -40,37 +40,308 @@ pub mod mock;
...
+
+    pub fn tcp_port(&self) -> u16 {
+        match &self.record {
+            VersionedNameRecord::V1(v1) => v1.port,
+            VersionedNameRecord::V2(v2) => {
+                v2.ports
+                    .iter()
+                    .find(|p| p.tag_enum() == Some(PortTag::TCP))
+                    .expect("V2 record must have TCP port")
+                    .port
```
dshulyak: ports vector is atmost 8 items, so this iteration is bounded, can use smallvec to make it more obvious

monad-peer-discovery/src/lib.rs
```diff
@@ -40,37 +40,308 @@ pub mod mock;
...
+
+    pub fn udp_port(&self) -> u16 {
+        match &self.record {
+            VersionedNameRecord::V1(v1) => v1.port,
+            VersionedNameRecord::V2(v2) => {
+                v2.ports
+                    .iter()
+                    .find(|p| p.tag_enum() == Some(PortTag::UDP))
+                    .expect("V2 record must have UDP port")
+                    .port
```
dshulyak: same as above

monad-peer-discovery/src/lib.rs
```diff
@@ -40,37 +40,308 @@ pub mod mock;
...
+        (self.capabilities() & (1u64 << (capability as u8))) != 0
+    }
+
+    pub fn set_capability(&mut self, capability: Capability) {
+        match &mut self.record {
+            VersionedNameRecord::V1(_) => {}
+            VersionedNameRecord::V2(v2) => {
+                v2.capabilities |= 1u64 << (capability as u8);
+            }
+        }
```
dshulyak, line 361: check and set can be done specific type, no need to have this methods the name record

dshulyak:
@eerkaijun please take a look when available. the follow up that switches to v2 by default https://github.com/category-labs/monad-bft/pull/2478

monad-peer-discovery/src/lib.rs
```diff
@@ -40,37 +40,308 @@ pub mod mock;
...
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum Capability {}
+
+#[derive(Debug, Clone, PartialEq, RlpEncodable, RlpDecodable, Eq)]
 pub struct MonadNameRecord<ST: CertificateSignatureRecoverable> {
     pub name_record: NameRecord,
     pub signature: ST,
```
dshulyak, line 390: i probably should bump a version for signing domain too

dshulyak:
there is also PeerEntry that distinct from name record, but pretty much tied to the name record and any changes between two will cause errors. it is very error prone to separate it like this, should be refactored to use name record instead

dshulyak:
> i probably should bump a version for signing domain too

i decided not to bump name record signing domain, this will be short living version bump, and encoding is different so it won't be possible to decode v1 encoded value as v2. so for simplicity it should be ok to omit that update, i can update if someone thinks that it might be necessary

dshulyak:
i double checked on stressnet that it decodes original name records as expected

UPD actually i am not sure if it works, it clearly decodes pings just fine, but there are also a lot of timeouts

dshulyak:
i added a commit with snapshot tests to verify that there was no changes in encoding for all messages.
and timeouts are not related to this change

monad-peer-discovery/src/discovery.rs
```diff
@@ -1533,7 +1535,23 @@ where
...
-            .map(|(id, name_record)| (*id, *name_record))
+            .map(|(id, name_record)| (*id, name_record.clone()))
+            .collect()
+    }
+}
+
+impl<ST: CertificateSignatureRecoverable> PeerDiscovery<ST> {
+    fn get_secondary_fullnode_addrs(
+        &self,
+    ) -> HashMap<NodeId<CertificateSignaturePubKey<ST>>, SocketAddrV4> {
```
eerkaijun: I think we can remove this function, not needed anywhere

eerkaijun (dismissed):

monad-peer-discovery/src/discovery.rs
dshulyak: removed, accidentally restored in cherry pick

monad-peer-discovery/src/lib.rs
```diff
@@ -40,37 +41,349 @@ pub mod mock;
...
+            &self.ip.octets(),
+            &self.ports,
+            &self.capabilities,
+            &self.seq,
+        ];
+        encode_list::<_, dyn Encodable>(&enc, out);
+    }
+}
+
+fn decode_port_list(buf: &mut &[u8], limit: usize) -> alloy_rlp::Result<PortList<8>> {
```
michael-yxchen: is it true that we always expect `limit` to equal to PortList's capacity? if so we should make that a constant

dshulyak: i updated code to use ArrayVec::try_push instead of limit, it will fail if capacity is overrun, and moved impl to PortList::decode

eerkaijun (approved):

PR #2047, opened 2025-07-24
Add round & other checks to full-node raptorcast

fabiotdf:
- Adds some extra checks, e.g. that the raptorcast group used matches the round of the proposal received by dynamic fullnode
- Bumped some logs from trace to debug that makes life easier when diagnosing raptorcast
- In raptorcast context, renamed `author` to `sender`, since the message (proposal) already has something called author

### DIFF
--- a/.github/workflows/license.yml
+++ b/.github/workflows/license.yml
@@ -17,7 +17,7 @@ on:
 
 jobs:
   check-license:
-    runs-on: self-hosted
+    runs-on: ubuntu-24.04-32
 
     steps:
       - name: Checkout repository


monad-raptorcast/src/udp.rs
```diff
@@ -718,7 +718,7 @@ where
         BuildTarget::Raptorcast((epoch_validators, full_nodes_view)) => {
             assert!(is_broadcast && is_raptor_broadcast);
 
-            tracing::trace!(
+            tracing::debug!(
```
eerkaijun: this is gonna make the log to be very noisy

monad-raptorcast/src/udp.rs
```diff
@@ -802,7 +802,7 @@ where
         BuildTarget::FullNodeRaptorCast(group) => {
             assert!(is_broadcast && is_raptor_broadcast);
 
-            tracing::trace!(
+            tracing::debug!(
```
eerkaijun: same as above

monad-raptorcast/src/util.rs
```diff
@@ -459,6 +459,36 @@ where
         }
     }
 
+    // Once we receive a message, we want to verify that the round referenced in
+    // the message matches the current group.
+    pub fn check_round(
+        &self,
+        msg_round: Option<Round>,
+        author_node_id: &NodeId<CertificateSignaturePubKey<ST>>,
```
eerkaijun: should this be renamed sender to be consistent with your other changes?

monad-raptorcast/src/raptorcast_secondary/publisher.rs
```diff
@@ -155,52 +155,63 @@ where
...
+        let group_start_round = *next_group.key();
+        if group_start_round < new_round {
+            tracing::debug!(
+                ?new_round,
+                ?group_start_round,
+                "RaptorCastSecondary jumping into middle of future scheduled group after a gap. \
+                Downstream full-nodes might see round gaps as well."
+            );
+        }
+        assert!(group_start_round == next_group.get().start_round);
```
eerkaijun: do we need this assertion? no harm i guess but seems excessive

fabiotdf:
> can we separate out the round gap crash commit to a separate PR?

Sure! Moved the changes here:
https://github.com/category-labs/monad-bft/pull/2090

monad-raptorcast/src/raptorcast_secondary/publisher.rs
fabiotdf: Fixed

monad-raptorcast/src/util.rs

monad-raptorcast/src/udp.rs
fabiotdf: Fixed

monad-raptorcast/src/udp.rs
fabiotdf: Fixed

monad-dataplane/src/lib.rs
```diff
@@ -457,6 +486,7 @@ impl DataplaneWriter {
         let msg_len = msg.payload.len();
 
         for (dst, udp_msg) in msg.into_iter() {
+            tracing::trace!("Dataplane UDP TX bcast {} bytes to {:?}", msg_len, dst);
```
dshulyak, line 489: i would prefer to use structured tracing consistently across codebase instead of formatting
e.g
trace!(msg_len, ?dst, "udp tx broadcast")

monad-dataplane/src/lib.rs
```diff
@@ -364,14 +364,28 @@ impl DataplaneReader {
...
 
     pub async fn udp_read(&mut self) -> RecvUdpMsg {
         match self.udp_ingress_rx.recv().await {
-            Some(msg) => msg,
+            Some(msg) => {
+                tracing::trace!(
+                    "Dataplane UDP RX {} bytes from {:?}",
+                    msg.payload.len(),
+                    msg.src_addr
+                );
```
dshulyak, line 386: just a note that trace! performs checks at runtime to check if message is logged or not, so it will have some overhead. also each separate tracing layer will have to do such check.

generally it is good idea to avoid adding tracing on the hot path, and if tracing is indeed needed then it should be handled with compile flag that is used only in debug builds

Issue #2005, opened 2025-07-15
dataplane: flaky test udp_unicast


dshulyak:
```
failures:

---- udp_unicast stdout ----

thread 'udp_unicast' panicked at monad-dataplane/tests/tests.rs:63:1:
timeout: the function call took 1000 ms. Max time 1000 ms
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
```

https://github.com/category-labs/monad-bft/actions/runs/16285306482/job/45982718950?pr=1983

dshulyak:
unicast_all_strides

https://github.com/category-labs/monad-bft/actions/runs/16293110043/job/46008450146


```
thread '<unnamed>' panicked at monad-dataplane/tests/tests.rs:497:13:
assertion `left == right` failed
[...]

PR #2051, opened 2025-07-24
monad-dataplane: deflake connections setup

dshulyak:
closes: https://github.com/category-labs/monad-bft/issues/2005

### DIFF
--- a/monad-dataplane/src/lib.rs
+++ b/monad-dataplane/src/lib.rs
@@ -1,10 +1,11 @@
 use std::{
     net::SocketAddr,
     sync::{
-        atomic::{AtomicUsize, Ordering},
+        atomic::{AtomicBool, AtomicUsize, Ordering},
         Arc,
     },
     thread,
+    time::Duration,
 };
 
 use bytes::Bytes;
@@ -52,6 +53,9 @@ impl DataplaneBuilder {
         let (udp_ingress_tx, udp_ingress_rx) = mpsc::channel(UDP_INGRESS_CHANNEL_SIZE);
         let (udp_egress_tx, udp_egress_rx) = mpsc::channel(UDP_EGRESS_CHANNEL_SIZE);
 
+        let ready = Arc::new(AtomicBool::new(false));
+        let ready_clone = ready.clone();
+
         thread::Builder::new()
             .name("monad-dataplane".into())
             .spawn(move || {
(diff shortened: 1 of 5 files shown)


dshulyak:
passed several times, it is definitely better than sleeps, but not sure yet if it fixes all flakes

monad-dataplane/src/lib.rs
```diff
@@ -70,6 +74,8 @@ impl DataplaneBuilder {
                             buffer_size,
                         );
 
+                        ready_clone.store(true, Ordering::Release);
```
xinyuan-dev, line 77: if we check it here it's still possible for the readiness to become true before tcp/udp socket gets bound, then the tests can still fail if the transmitter attempts to send packets to unbound socket.

monad-dataplane/src/lib.rs
```diff
@@ -78,13 +84,18 @@ impl DataplaneBuilder {
...
         let reader = DataplaneReader::new(tcp_ingress_rx, udp_ingress_rx);
 
-        Dataplane { writer, reader }
+        Dataplane {
+            writer,
+            reader,
+            ready,
+        }
     }
 }
```
xinyuan-dev, line 93: alternatively we can wait here in `DataplaneBuilder::build`, that should only return if the sockets are properly set up. then no more waiting is necessary. we can also returning a `Result<Dataplane>` here to report any error that happens during socket setup.

monad-dataplane/src/lib.rs
dshulyak: both udp and tcp are bound before a call to spawn

xinyuan-dev: you're right. i overlooked the change in tcp::spawn_tasks.

monad-dataplane/src/lib.rs
```diff
@@ -205,6 +216,21 @@ impl Dataplane {
     pub fn udp_write_unicast(&self, msg: UnicastMsg) {
         self.writer.udp_write_unicast(msg);
     }
+
+    pub fn ready(&self) -> bool {
+        self.ready.load(Ordering::Acquire)
+    }
+
+    pub fn wait_ready(&self, timeout: Duration) -> bool {
```
andr-dev: Most rust functions that do some form of "blocking" make that clear in the fn name with a `block/blocking` prefix, how about `block_until_ready`?

PR #2053, opened 2025-07-24
Archive: indexer eth_getLogs subcommands and capped collections migration

JoeHowarth:
- tested with db created by v0.9.2 
  - migrated both block_level and tx_index successfully
  - successfully detected when already uncapped
  - cleaned up renamed collection
  - was able to continue writing to new collection after migration

### DIFF
--- a/monad-archive/src/bin/monad-indexer/migrate_capped.rs
+++ b/monad-archive/src/bin/monad-indexer/migrate_capped.rs
@@ -1,0 +1,753 @@
+//! Migration helper: make an existing capped collection uncapped
+//! without interrupting writers (rename-first pattern).
+//!
+//! Call it as many times as you like; it is crash-resumable and
+//! no-op after reaching Phase::Done.
+
+use monad_archive::prelude::*;
+use mongodb::{
+    bson::{doc, Bson, Document},
+    Client, Collection,
+};
+use serde::{Deserialize, Serialize};
+
+/// meta-collection name
+const META_COLL: &str = "_migrations";
+
+/// Public entry point -------------------------------------------------------
+pub async fn migrate_to_uncapped(
+    client: &Client,
+    db_name: &str,
+    coll_name: &str,
+    batch_size: u32,
+    free_factor: f64,
+) -> Result<()> {
+    let db = client.database(db_name);
+    let meta: Collection<MetaDoc> = db.collection(META_COLL);
+    let source_capped = format!("{coll_name}_capped");
+
+    // Load or create state document
+    let mut state = meta
+        .find_one(doc! { "_id": coll_name })
+        .await?
+        .unwrap_or_else(|| MetaDoc::new(coll_name));
+
+    loop {
+        match state.phase {
+            Phase::DiskCheck => {
+                ensure_disk_space(&db, coll_name, &source_capped, free_factor).await?;
+                state.phase = Phase::Renamed;
+                meta.upsert(&state).await?;
+                continue; // re-enter loop
+            }
+            Phase::Renamed => {
+                let renamed = maybe_rename(&db, coll_name, &source_capped).await?;
+                if !renamed {
+                    // Collection was already uncapped, we're done
+                    info!(coll_name, "collection already uncapped, migration complete");
+                    meta.delete_one(doc! { "_id": coll_name }).await?;
+                    return Ok(());
+                }
+                create_uncapped_and_indexes(&db, coll_name, &source_capped).await?;
+                state.phase = Phase::Copying;
+                meta.upsert(&state).await?;
+                continue;
+            }
+            Phase::Copying => {
+                // Keep copying until all batches are done
+                loop {
+                    let copied = copy_batches(
+                        &db,
+                        &source_capped,
+                        coll_name,
+                        &mut state,
+                        batch_size,
+                        &meta,
+                    )
+                    .await?;
+
+                    if !copied {
+                        // No more documents to copy
+                        break;
+                    }
+                }
+                state.phase = Phase::Dropping;
+                meta.upsert(&state).await?;
+                continue;
+            }
+            Phase::Dropping => {
+                drop_source(&db, &source_capped, coll_name).await?;
+                meta.delete_one(doc! { "_id": coll_name }).await?;
+                info!(coll_name, "migration complete");
+                return Ok(());
+            }
+            Phase::Done => {
+                return Ok(());
+            }
+        }
+    }
+}
+
+/// --------------------------------------------------------------------------
+///  Implementation helpers
+/// --------------------------------------------------------------------------
+/// Meta-document persisted across restarts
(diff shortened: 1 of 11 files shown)


monad-archive/src/bin/monad-indexer/migrate_capped.rs
```diff
@@ -0,0 +1,716 @@
...
+                    // If we can't get collection stats, estimate based on database stats
+                    warn!("Could not get collection stats, using database stats for estimation");
+                    doc! {}
+                }
+            }
+        }
+    };
+
+    // Calculate space needed
+    let need = if let Ok(coll_storage_size) = get_int_field(&coll_stats, "storageSize") {
```
mikety: Need to also include` totalIndexSize` into the estimate since its creates the collection and indexes in`create_uncapped_and_indexes`

monad-archive/src/bin/monad-indexer/migrate_capped.rs
```diff
@@ -0,0 +1,716 @@
...
+}
+
+/// Step 2 — create destination and copy indexes
+async fn create_uncapped_and_indexes(
+    db: &mongodb::Database,
+    coll: &str,
+    source: &str,
+) -> Result<()> {
+    // create if missing
+    if !db
```
mikety, line 253: Will be good to report or at least log that the collection does exists.

monad-archive/src/bin/monad-indexer/migrate_capped.rs
```diff
@@ -0,0 +1,716 @@
...
+/// Step 0  —  free-space guard
+async fn ensure_disk_space(
+    db: &mongodb::Database,
+    coll: &str,
+    capped_name: &str,
+    factor: f64,
+) -> Result<()> {
+    // Get database stats to understand current storage usage
+    let db_stats = db
+        .run_command(doc! { "dbStats": 1, "scale": 1 })
```
mikety: nit: if you add `freeStorage: 1` it will display the free disk space in the fs where the DB is allocated (dbpath)

monad-archive/src/bin/monad-indexer/migrate_capped.rs
```diff
@@ -0,0 +1,716 @@
...
+                    doc! {}
+                }
+            }
+        }
+    };
+
+    // Calculate space needed
+    let need = if let Ok(coll_storage_size) = get_int_field(&coll_stats, "storageSize") {
+        // Use collection-specific storage size if available
+        (coll_storage_size as f64 * factor) as i64
```
mikety: Need some experiments to determine the `factor` (better name is `scale_factor`) . The data in Mongo is compressed

monad-archive/src/bin/monad-indexer/migrate_capped.rs
```diff
@@ -0,0 +1,716 @@
...
+    coll: &str,
+    source: &str,
+) -> Result<()> {
+    // create if missing
+    if !db
+        .list_collection_names()
+        .await?
+        .contains(&coll.to_string())
+    {
+        db.create_collection(coll).await?;
```
mikety, line 259: Note: $merge will create a collection if it does not exist

monad-archive/src/bin/monad-indexer/migrate_capped.rs
```diff
@@ -0,0 +1,716 @@
...
+                ensure_disk_space(&db, coll_name, &source_capped, free_factor).await?;
+                state.phase = Phase::Renamed;
+                meta.upsert(&state).await?;
+                continue; // re-enter loop
+            }
+            Phase::Renamed => {
+                let renamed = maybe_rename(&db, coll_name, &source_capped).await?;
+                if !renamed {
+                    // Collection was already uncapped, we're done
+                    info!(coll_name, "collection already uncapped, migration complete");
```
mikety, line 47: What if the source being actively changed while the cloning goes? If its an issue a catchup phase should be added with a critical section to block writes to make avoid missed writes.

monad-archive/src/bin/monad-indexer/migrate_capped.rs
```diff
@@ -0,0 +1,716 @@
...
+) -> Result<bool> {
+    let src_coll = db.collection::<Document>(src);
+
+    // 1 — grab the next `batch` _ids after `last_id`  (only _id projected)
+    let mut id_filter = doc! {};
+    if let Some(ref last) = state.last_id {
+        id_filter.insert("_id", doc! { "$gt": last.clone() });
+    }
+    let mut id_cursor = src_coll
+        .find(id_filter)
```
mikety, line 327: Can you comment on the assumptions of what is stored in the _id. This will work only with collections with sequential _ids

monad-archive/src/bin/monad-indexer/migrate_capped.rs
JoeHowarth: The source is renamed and a matching un-capped collection is created with the same indexes. Writers will temporarily see no collection present, then the new empty collection will be available again and writes can proceed. 

This pattern means we don't have to worry about lost writes for more than a couple seconds, which all writers should be resilient against anyways. Notably, we don't block writes for the duration of the copying, that happens in the background async

monad-archive/src/bin/monad-indexer/migrate_capped.rs
JoeHowarth: Is this different than total - used?:
```
    let fs_used = db_stats
        .get_i64("fsUsedSize")
        .or_else(|_| db_stats.get_f64("fsUsedSize").map(|v| v as i64))
        .unwrap_or(0);
    let fs_total = db_stats
        .get_i64("fsTotalSize")
        .or_else(|_| db_stats.get_f64("fsTotalSize").map(|v| v as i64))
        .unwrap_or(0);
```

monad-archive/src/bin/monad-indexer/migrate_capped.rs
JoeHowarth: The data stays in mongo, so as long as the format/compression is the same in the new collection as the old collection, the disk space needed should be the same, right?

monad-archive/src/bin/monad-indexer/migrate_capped.rs
JoeHowarth: Yeah, I wanted to do it here since then we can ensure the indexes are copied over before cloning the data. Also more explicit

monad-archive/src/bin/monad-indexer/migrate_capped.rs
JoeHowarth: The _id field does not have to be incremental or sequential, just orderable, which by definition all bson types are. 

I could be making a mistake with my reasoning though, do you see where orderable is not sufficient?

mikety: should be ok.

PR #2055, opened 2025-07-24
Archive: do not create mongo collections as 'capped'

JoeHowarth:


monad-archive/src/kvstore/mongo.rs
```diff
@@ -134,43 +117,21 @@ impl MongoDbStorage {
         let client = new_client(connection_string).await?;
 
         let db = client.database(database);
-        debug!("Using database: {}", database);
```
mikety, line 152: Why do not log a db name? It can help debug why a collection does not exist.

JoeHowarth: It's logged at the top of the function at `info` level, so deleting the duplicate

Issue #2000, opened 2025-07-14
monad-bls: Universal signatures


robertkiel:
Monad uses BLS signatures as they are easily aggregatable. The basic idea is that a threshold of validators sign a message (i.e. a consensus message) and each validator receiving this message can efficiently verify this signature. To determine which validators have been part of the signing process, the signature comes with a bitmap with those indices set to true that relate to the involved public keys. (For verification, the nodes require a validator map)

```rust
pub struct BlsSignatureCollection {
    pub signers: SignerMap, // bitmap

    pub sig: BlsAggregateSignature, // curvepoint
}
```

However, if the `SignerMap` is set to `[false,..., false]` and the `sig` property is set to `BlsAggregateSignature::infinity()`. A call to `BlsSignatureCollection::verify(validator_set, msg)` will return `Ok(...)` - **independent** of the validator set and **independent** of the actual message.

```rust
let mut bitvector = BitVec::with_capacity(5);
for _ in 0..5 {
    bitvector.push(false);
}
// taken from `monad-bls/src/aggregation_tree.rs`
let (keys, voting_keys, _, valmap) = create_keys_w_validators::<
    SignatureType,
    SignatureCollectionType,
    _,
>(5, ValidatorSetFactory::default());
let sigs = BlsSignatureCollection::<CertificateSignaturePubKey<NopSignature>> {
    signers: SignerMap(bitvector),
    sig: BlsAggregateSignature::infinity(),
    _phantom: PhantomData
};

sigs.verify(&valmap, &[]).unwrap(); // does NOT panic!
sigs.verify(&valmap, &[0x12]).unwrap(); // does NOT panic!
sigs.verify(&valmap, &[0x34]).unwrap(); // does NOT panic!
```

This is the case because the implementation [checks the length](https://github.com/category-labs/monad-bft/blob/d71cd1cb9cc929ac60e232f453cb0b5f662ac5db/monad-bls/src/aggregation_tree.rs#L264) of the bitmap and if none of its the bitmaps values is set to true, the [for loop](https://github.com/category-labs/monad-bft/blob/d71cd1cb9cc929ac60e232f453cb0b5f662ac5db/monad-bls/src/aggregation_tree.rs#L270-L275) is actually a no-op. Which finally makes the function [return](https://github.com/category-labs/monad-bft/blob/d71cd1cb9cc929ac60e232f453cb0b5f662ac5db/monad-bls/src/aggregation_tree.rs#L278) `Ok(signers)`.

[...]

PR #2057, opened 2025-07-25
Fix exploitable universal signature in monad-bls

andr-dev:
Resolves https://github.com/category-labs/monad-bft/issues/2000.


michael-yxchen:
superseded by https://github.com/category-labs/monad-bft/pull/2379

PR #2063, opened 2025-07-25
Track txpool events by tx hash

andr-dev:
This change keys txpool events by tx hash which enables deduplicating double events like `[Insert (Pending), Insert (Tracked)]` or `[Insert (Pending), Drop (some drop reason)]`.

This should improve more nuanced RPC `eth_sendTransaction` and `eth_sendRawTransaction` responses since RPC will respond with the first event received from txpool. In a situation like `[Insert (Pending), Drop (some drop reason)]`, previously, the event would not get deduplicated resulting in an `Ok` response. Now that events are tracked by tx hash, the `Drop` event will overwrite the `Insert` and thus the error will be immediately returned to the user.

### DIFF
--- a/monad-eth-txpool-executor/src/ipc/mod.rs
+++ b/monad-eth-txpool-executor/src/ipc/mod.rs
@@ -14,6 +14,7 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 use std::{
+    collections::BTreeMap,
     future::Future,
     io,
     pin::Pin,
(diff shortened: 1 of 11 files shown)


ariqchowdhury (dismissed):

ariqchowdhury (approved):

omegablitz (approved):

monad-eth-txpool-executor/src/ipc/mod.rs
```diff
@@ -74,13 +76,19 @@ impl EthTxPoolIpcServer {
...
-        self.project().connections.retain(|stream| {
-            match stream.send_tx_events(events.to_owned()) {
+        let events: Vec<EthTxPoolEvent> = events
+            .into_iter()
+            .map(|(tx_hash, action)| EthTxPoolEvent { tx_hash, action })
+            .collect();
+
+        self.project()
+            .connections
+            .retain(|stream| match stream.send_tx_events(events.clone()) {
```
omegablitz, line 91: just noting that there's an extra clone here now of the `Vec<EthTxPoolEvent>`

PR #2066, opened 2025-07-26
Update clear-old-artifacts to new ledger format

omegablitz:


omegablitz:
I noticed that this clear-old-artifacts GCs less frequently than this one:
https://github.com/category-labs/category-infra/blob/e62217fe7b77543fae076121ad0cc2d58b1f35a0/ansible/scripts/node/clear-old-artifacts.sh

I think we were previously using this less frequent GC schedule only for the archivers to increase margin for error, given that their disks are much larger.

It's fine to keep these more conservative params uniform across all nodes if we expect the disk space to be sufficient.
10 hours * 4MB/s = 144 GB

PR #2069, opened 2025-07-28
main reserve

mikety:


monad-chain-config/src/revision.rs
```diff
@@ -2,6 +2,8 @@ use std::{fmt::Debug, time::Duration};
 
 pub const CHAIN_PARAMS_LATEST: ChainParams = CHAIN_PARAMS_V_0_10_0;
 
+pub const RESERVE_BALANCE: u128 = 1_000_000_000_000_000_000; // 1 MON
```
andr-dev: This constant makes it seem like the reserve balance limit never changes except with hard forks, can we remove it?

monad-consensus-types/Cargo.toml
```diff
@@ -14,6 +14,7 @@ monad-state-backend = { workspace = true }
 monad-types = { workspace = true }
 
 alloy-rlp = { workspace = true }
+alloy-primitives = { workspace = true }
```
andr-dev, line 16: This dep is out of order.

monad-eth-block-policy/src/lib.rs
```diff
@@ -73,6 +84,10 @@ pub fn compute_txn_max_value(txn: &TxEnvelope) -> U256 {
     txn_value.saturating_add(gas_cost)
 }
 
+pub fn compute_txn_max_gas_cost(txn: &TxEnvelope) -> U256 {
```
andr-dev, line 92: We can call this fn in the `compute_txn_max_value` above to set `gas_cost`.

monad-eth-block-policy/src/lib.rs
```diff
@@ -303,21 +320,42 @@ where
...
+        let mut cumulative_max_cost = Balance::ZERO;
         let mut next_validate = base_seq_num + SeqNum(1);
+        let mut block_seqnum_of_latest_txn = GENESIS_SEQ_NUM;
+
+        // Compute the block seqnum of the latest txn from the blocks before the base
+        let start_seq_num = next_validate.max(execution_delay) - execution_delay;
+        for (&cache_seq_num, block) in self.blocks.range(start_seq_num..next_validate) {
+            trace!("compute_txn_fee look back: {:?}", cache_seq_num);
+            if block.fees.get(eth_address).is_some() {
+                block_seqnum_of_latest_txn = cache_seq_num;
```
andr-dev: For efficiency, we can `self.blocks.range(start_seq_num..next_validate).rev()` and then break on `is_some()` after setting `block_seqnum_of_latest_txn = cache_seq_num;`.

monad-eth-block-policy/src/lib.rs
```diff
@@ -360,6 +398,89 @@ where
...
+        min_blocks_since_latest_txn: SeqNum,
+    ) -> Result<Self, BlockPolicyError> {
+        Ok(Self {
+            block_seq_num,
+            account_balances,
+            min_blocks_since_latest_txn,
+        })
+    }
+
+    fn try_add_transaction(&mut self, txn: &Self::Transaction) -> Result<(), BlockPolicyError> {
```
andr-dev: This method should produce a `BlockPolicyBlockValidatorError` which specifies what went wrong.
- Instead of `.expect("account_balances should have been populated")` we can produce a `BlockPolicyBlockValidatorError::AccountBalanceMissing` error. This would be useful in the txpool for debugging / reporting.
- Err on 453 isn't an incoherent block, rather the tx can't be added in which case a `BlockPolicyBlockValidatorError::InsufficientEmptyingBalance` error makes sense
- Err on 470 can be `BlockPolicyBlockValidatorError::InsufficientReserveBalance`

monad-eth-block-policy/src/lib.rs
```diff
@@ -536,45 +665,89 @@ where
...
-                            Transaction Fee Committed: {:?} \
-                            consensus block:seq num {:?} \
-                            for address: {:?}",
-                        account_balance,
-                        txn_fee_committed,
-                        consensus_block_seq_num,
-                        address
+                }
+
+                // a transaction is allowed to transfer out of the reserve balance if and only if it's an EOA that is not delegated and there were no transactions for that account in the last k blocks
```
andr-dev: Comments should be multilined

monad-eth-block-policy/src/lib.rs
```diff
@@ -267,7 +282,9 @@ struct CommittedBlkBuffer<ST, SCT> {
 }
 
 struct CommittedTxnFeeResult {
-    txn_fee: Balance,
+    cumulative_max_gas_cost: Balance,
+    cumulative_max_cost: Balance,
+    block_seqnum_of_latest_txn: SeqNum,
     next_validate: SeqNum, // next block number to validate; included for assertions only
```
andr-dev: Can we rename this field? `next_validate` doesn't really say anything about what the value represents, can we just call it `next_expected_seq_num` or just `next_seq_num`

monad-eth-block-policy/src/lib.rs
```diff
@@ -536,45 +665,89 @@ where
...
+                {
+                    warn!(
+                        ?balance_state,
+                        ?can_transfer_out_of_reserve_balance,
+                        ?cumulative_max_cost,
+                        ?cumulative_max_gas_cost,
+                        ?consensus_block_seq_num,
+                        ?address,
+                        "Committed block with insufficient balance",
                     );
```
andr-dev: Why is this a warn when the above is a panic?

monad-eth-block-policy/src/lib.rs
```diff
@@ -583,37 +756,64 @@ where
...
+                    ?consensus_block_seq_num,
+                    ?next_validate,
+                    ?address,
+                    "AccountBalanceState compute pending block: updated balance",
+                );
+
+                (address, balance_state)
             })
             .collect();
         Ok(account_balances)
```
andr-dev: This function was already pretty long and with the new changes is just too much in a single place. Let's split out this entire `map` into a private associated fn or two so it's easier to follow.

monad-eth-block-policy/src/lib.rs
```diff
@@ -378,6 +499,11 @@ where
 
     /// Chain ID
     chain_id: u64,
+
+    max_reserve_balance: Balance,
+
+    // Smallest number of blocks after the latest txn to allow transferring out of reserve balance
+    min_blocks_since_latest_txn: SeqNum,
```
andr-dev, line 629: Shouldn't these be computed values from CCT? Also `min_blocks_since_latest_txn` is a pretty confusing name, how about `reserve_balance_transfer_min_block_delta`?

monad-eth-block-policy/src/lib.rs
```diff
@@ -730,46 +947,10 @@ where
...
-                    seq_num =? block.header().seq_num,
-                    round =? block.header().block_round,
-                    "block not coherent, invalid balance"
-                );
-                return Err(BlockPolicyError::BlockNotCoherent);
-            }
+            validator.try_add_transaction(txn)?;
+            *expected_nonce += 1;
         }
+        let mut res_block = block.clone();
```
andr-dev: I think this line got left behind, let's get rid of it.

monad-eth-block-validator/Cargo.toml
```diff
@@ -10,6 +10,7 @@ bench = false
 
 [dependencies]
 monad-consensus-types = { workspace = true }
+monad-types = { workspace = true }
```
andr-dev: Dep needs to be sorted.

monad-eth-block-validator/src/lib.rs
```diff
@@ -100,7 +99,7 @@ where
 
         // recover the account nonces and txn fee usage in this block
         let mut nonces = BTreeMap::new();
-        let mut txn_fees: BTreeMap<Address, U256> = BTreeMap::new();
+        let mut txn_fees: TxnFees = BTreeMap::default();
```
andr-dev:
```rust
let mut txn_fees = TxnFees::default();
```

monad-eth-testutil/src/lib.rs
```diff
@@ -144,11 +146,32 @@ pub fn generate_block_with_txs(
...
+        .fold(
+            BTreeMap::new(),
+            |mut costs, (address, max_cost, max_gas_cost)| {
+                costs
+                    .entry(address)
+                    .or_insert(TxnFee {
+                        max_cost: Balance::ZERO,
+                        max_gas_cost: Balance::ZERO,
+                        first_txn_nonce: None,
+                    })
```
andr-dev: We should derive `Default` for TxnFee and use `.or_default()`.

monad-eth-txpool/src/pool/mod.rs
```diff
@@ -167,9 +168,20 @@ where
...
+                ValidEthTransaction::apply_max_value(&tx, account_balance.balance)
+            else {
+                event_tracker.drop(tx.hash(), EthTxPoolDropReason::InsufficientBalance);
+                continue;
+            };
+
+            let Some(_new_reserve_balance) = ValidEthTransaction::apply_max_gas_cost(
+                &tx,
+                account_balance.remaining_reserve_balance,
+            ) else {
```
andr-dev: Why can't we just call `tx.apply_max_value(...)` directly here?

monad-eth-txpool/src/pool/tracked/mod.rs
```diff
@@ -351,14 +356,23 @@ where
...
+        min_blocks_since_latest_txn: SeqNum,
+        mut account_balances: BTreeMap<&Address, AccountBalanceState>,
+    ) -> Result<(u64, Vec<Recovered<TxEnvelope>>), BlockPolicyError> {
         assert!(tx_limit > 0);
 
         let mut txs = Vec::new();
         let mut total_gas = 0u64;
         let mut total_size = 0u64;
 
+        let mut balances = account_balances.clone();
```
andr-dev: Why do we clone here?

monad-eth-block-policy/src/lib.rs
mikety: updated

monad-eth-txpool/src/pool/tracked/mod.rs
mikety: lifetimes

monad-eth-block-policy/src/lib.rs
```diff
@@ -428,10 +439,20 @@ where
 
     fn try_add_transaction(&mut self, txn: &Self::Transaction) -> Result<(), BlockPolicyError> {
         let eth_address = txn.signer();
-        let account_balance = self
-            .account_balances
-            .get_mut(&eth_address)
-            .expect("account_balances should have been populated");
+        let mayby_account_balance = self.account_balances.get_mut(&eth_address);
```
ariqchowdhury: `maybe_...`

monad-eth-block-policy/src/lib.rs
```diff
@@ -360,6 +409,103 @@ where
     }
 }
 
+pub struct EthBlockPolicyBlockState {
+    fees: BTreeMap<Address, TxnFees>,
+}
+
+pub struct EthBlockPolicyBlockValidator<'a> {
+    block_seq_num: SeqNum,
+    account_balances: &'a mut BTreeMap<&'a Address, AccountBalanceState>,
```
ariqchowdhury: does the validator actually need to store a mutable ref to the account balances? Can it be passed into try_add_transaction instead? Would allow the removal of the lifetime from the struct

mikety: If its passed as a mutable borrow to `try_add_transaction`  then it will not compile in the loop call from `check_coherency` (line 963) (  the borrow from the previous iteration is considered still active)

PR #2071, opened 2025-07-28
Switch epoch lock from n+2 to n+1

bamrith:
This PR converts validator set locking at the end of an Epoch(N) from Epoch(N+2) to Epoch(N+1)

Updated forkpoint tests, mock swarm tests, and flexnet configs

### DIFF
--- a/monad-mock-swarm/tests/epoch.rs
+++ b/monad-mock-swarm/tests/epoch.rs
@@ -515,11 +515,10 @@ mod test {
         .iter()
         .map(|vdata| vdata.node_id)
         .collect();
-        let (validators_epoch_3, validators_epoch_4) = genesis_validators.split_at(2);
+        let (validators_epoch_2, validators_epoch_3) = genesis_validators.split_at(2);
         // validators for epoch 1 = genesis_validators
-        // validators for epoch 2 = genesis_validators
+        // validators for epoch 2 = validators_epoch_2
         // validators for epoch 3 = validators_epoch_3
-        // validators for epoch 4 = validators_epoch_4
 
         let all_peers: BTreeSet<_> = state_configs
             .iter()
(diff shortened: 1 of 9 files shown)


omegablitz (approved):
LGTM, but needs some care for roll out. @michael-yxchen should review this as well prior to merging.

Also, this should be merged only after v0.10.4 is frozen IMO.

monad-state/src/lib.rs
```diff
@@ -237,11 +229,6 @@ where
             .all(|(locked_vset, forkpoint_vset)| locked_vset.epoch == forkpoint_vset.epoch));
 
         // 3.
-        let Some(_validator_set_0_round) = self.validator_sets[0].round else {
-            return Err(ForkpointValidationError::InvalidValidatorSetStartRound);
```
michael-yxchen, line 248: can remove this error variant

monad-mock-swarm/tests/epoch.rs
```diff
@@ -646,14 +618,13 @@ mod test {
...
+                if full_block.get_block_round() < epoch_2_start_round {
+                    // the first epoch has genesis validators as the validator set
                     assert!(genesis_validators.contains(full_block.get_author()));
-                } else if full_block.get_block_round() < epoch_4_start_round {
-                    assert!(validators_epoch_3.contains(full_block.get_author()));
+                } else if full_block.get_block_round() < epoch_3_start_round {
+                    assert!(validators_epoch_2.contains(full_block.get_author()));
                 } else {
-                    assert!(validators_epoch_4.contains(full_block.get_author()));
+                    assert!(validators_epoch_3.contains(full_block.get_author()));
```
michael-yxchen, line 675: test won't hit this line because simulation never enters epoch 3. we should make it consistent. feel free to edit it either way

bamrith: fixed

omegablitz (approved):

michael-yxchen (approved):

PR #2072, opened 2025-07-28
Remove monad-full-node

bamrith:


dhogaivannan:
debian/usr/lib/systemd/system/monad-bft-fullnode.service -> this might be an artifact if `monad-full-node` is getting depreciated

mhumeSF:
This is amazing!

ariqchowdhury:
the role of full-node vs validator is represented in consensus-state with this change, so does that mean this also works with mock-swarm? Can the fullnode->validator unit tests be implemented in mock-swarm as well?

bamrith:
> the role of full-node vs validator is represented in consensus-state with this change, so does that mean this also works with mock-swarm? Can the fullnode->validator unit tests be implemented in mock-swarm as well?

@ariqchowdhury yes that's right, that should work. it does need some changes to forward messages correctly to full-nodes, i'll try to get it in along with this PR. i think it should make our epoch change mock swarm tests less weird since we can remove the 0 stake hack.

monad-consensus-state/src/lib.rs
```diff
@@ -286,9 +301,63 @@ where
                 None,
             ),
             block_sync_requests: Default::default(),
+        };
+        consensus_state.update_role(epoch_manager, val_epoch_map, node_id);
+        debug!(role=?consensus_state.role, "Consensus role initialized");
+        consensus_state
+    }
+
+    fn update_role<VTF>(
```
omegablitz: It looks like this is only called on init, unless I'm missing something. Was the intention to call this every time round is bumped?

monad-node/src/main.rs
```diff
@@ -246,9 +246,6 @@ async fn run(node_state: NodeState, reload_handle: Box<dyn TracingReload>) -> Re
         )
         .await;
 
-        #[cfg(feature = "full-node")]
-        let raptor_router = monad_router_filter::FullNodeRouterFilter::new(raptor_router);
-
         <_ as Updater<_>>::boxed(raptor_router)
     };
```
omegablitz: nit: simplify to `let router = build_raptorcast_router:: ...` instead of boxing it. We don't need to box anymore because it's always the same type.

monad-state/src/consensus.rs
```diff
@@ -512,6 +518,75 @@ where
...
+        let live = match self.consensus {
+            ConsensusMode::Live(live) => live,
+            ConsensusMode::Sync { .. } => {
+                return Role::FullNode;
+            }
+        };
+
+        // - Validator if a current validator
+        // - UpcomingValidator if will become a validator in the next 10 rounds
+        //   (10 is arbitrary. the exact number of rounds can be as small as 1)
```
omegablitz: Is there any reason to distinguish UpcomingValidator from Validator?

An upcoming validator in round r shouldn't be sending votes/proposals/timeouts for any round < r, which is why I'm not sure this distinction is useful.

If it's not useful, this code would simplify to just be `if is_member(current_round) Validator else FullNode`, and you wouldn't need to look up current_round + some arbitrary round diff.

monad-consensus-types/src/voting.rs
```diff
@@ -21,6 +21,7 @@ use monad_types::*;
 use serde::{Deserialize, Serialize};
 
 /// Map validator NodeId to its Certificate PubKey
+#[derive(Clone)]
```
omegablitz: Is this just for a new test in consensus-state? That's fine if so

bamrith: removed the tests since it's unrelated to this PR

monad-raptorcast/src/udp.rs
```diff
@@ -181,6 +181,17 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
                 }
             };
 
+            // Ignore chunk if self is the author
```
omegablitz, line 184: Does this happen in practice?

omegablitz (dismissed):
```diff
diff --git a/.github/workflows/publish.yaml b/.github/workflows/publish.yaml
index eae067043..3eaff3bd3 100644
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -21,8 +21,6 @@ jobs:
         dockerfile:
           - { name: "monad-bft", file: "docker/devnet/Dockerfile" }
           - { name: "monad-bft", file: "docker/devnet/Dockerfile", target_postfix: -debug }
-          - { name: "monad-full-node", file: "docker/devnet/fullnode.Dockerfile" }
-          - { name: "monad-full-node", file: "docker/devnet/fullnode.Dockerfile", target_postfix: -debug }
           - { name: "monad-rpc", file: "docker/rpc/Dockerfile" }
           - { name: "monad-rpc", file: "docker/rpc/Dockerfile", target_postfix: -debug }
           - { name: "monad-execution", file: "monad-cxx/monad-execution/docker/release.Dockerfile", context: "monad-cxx/monad-execution" }
diff --git a/debian/opt/monad/scripts/reset-workspace.sh b/debian/opt/monad/scripts/reset-workspace.sh
index 4a804ec87..b8a573b74 100644
--- a/debian/opt/monad/scripts/reset-workspace.sh
+++ b/debian/opt/monad/scripts/reset-workspace.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 set -ex
-systemctl stop monad-bft monad-bft-fullnode monad-execution monad-rpc monad-mpt monad-execution-genesis || true
+systemctl stop monad-bft monad-execution monad-rpc monad-mpt monad-execution-genesis || true
 mkdir /home/monad/monad-bft/empty-dir
 rsync -r --delete /home/monad/monad-bft/empty-dir/ /home/monad/monad-bft/ledger/
 rsync -r --delete /home/monad/monad-bft/empty-dir/ /home/monad/monad-bft/config/forkpoint/
diff --git a/debian/usr/lib/systemd/system/monad-bft-fullnode.service b/debian/usr/lib/systemd/system/monad-bft-fullnode.service
deleted file mode 100644
index 3d2466193..000000000
--- a/debian/usr/lib/systemd/system/monad-bft-fullnode.service
+++ /dev/null
@@ -1,39 +0,0 @@
-[Unit]
-Description="Service file for Monad BFT Fullnode"
-After=network.target
-
-[Service]
-Type=simple
-ExecStart=/usr/local/bin/monad-full-node \
-    --secp-identity /home/monad/monad-bft/config/id-secp \
-    --bls-identity /home/monad/monad-bft/config/id-bls \
-    --node-config /home/monad/monad-bft/config/node.toml \
-    --forkpoint-config /home/monad/monad-bft/config/forkpoint/forkpoint.toml \
-    --wal-path /home/monad/monad-bft/wal \
-    --mempool-ipc-path /home/monad/monad-bft/mempool.sock \
-    --control-panel-ipc-path /home/monad/monad-bft/controlpanel.sock \
-    --statesync-ipc-path /home/monad/monad-bft/statesync.sock \
-    --ledger-path /home/monad/monad-bft/ledger \
-    --triedb-path /dev/triedb \
-    --otel-endpoint "http://0.0.0.0:4317" \
-    --statesync-sq-thread-cpu 1 \
-    --record-metrics-interval-seconds 1 \
-    --validators-path /home/monad/monad-bft/config/validators.toml \
-    --keystore-password "${KEYSTORE_PASSWORD}"
-EnvironmentFile=-/home/monad/.env
-Environment="RUST_LOG=debug,h2=warn,tower=warn"
-Restart=no
-User=monad
-Group=monad
-AllowedCPUs=0-4
-DeviceAllow=/dev/triedb rw
-LimitNOFILE=1048576
-# Hugemem configs below
-LimitMEMLOCK=infinity
-MemoryDenyWriteExecute=false
-#SystemCallFilter=~@memlock
-ProtectKernelModules=no
-ProtectKernelLogs=no
-
-[Install]
-WantedBy=multi-user.target
```

monad-raptorcast/src/udp.rs
bamrith: If a full-node has a upstream validator and changes to a validator in an epoch, it will receive all the packets until it is removed as full-node from the validator, including its own. There are no functional issues when that happens but this just avoids the extra log spam and having to go through the `recently_decoded_cache` or `pending_message_cache`

michael-yxchen (approved):

PR #2075, opened 2025-07-28
Separate ping pong from RouterCommand

eerkaijun:
In a future refactor, we want to make sure a ping pong round trip is completed before a name record is inserted into routing_info. This PR separates ping and pong commands from the usual RouterCommand. PingPongCommand provides an additional field which is the socket address to send the message to, so that we can send pings and pongs without inserting them into `routing_info`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -74,7 +74,7 @@ dependencies = [
  "flate2",
  "foldhash",
  "futures-core",
- "h2 0.3.26",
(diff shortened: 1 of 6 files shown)


ariqchowdhury:
is this covered by mock-swarm testing?

eerkaijun:
yeah used in peer disc swarm tests too. some of the tests will also need to be changed in the next pr together with the refactor

PR #2087, opened 2025-07-30
Include last_round_qc if necessary

omegablitz:
When sending a timeout with round=r, it's possible for high_tip.qc.round to be r-2. If the local high_certificate is QC(r-1), we need to support including QC(r-1) instead of TC(r-1).


omegablitz:
Tracked test case end of here: https://www.notion.so/Bftv2-Test-Cases-23175b0ba84080979cf3e3388bbe9399

michael-yxchen (approved):

PR #2090, opened 2025-07-31
Fix round gap crash

fabiotdf:
- Addresses a crash first reported in [slack](https://category-labs.slack.com/archives/C0937BUM5GV/p1753802413547709)
- Adds unit test to reproduce the crash & make sure group scheduling continues as expected after resolution


ariqchowdhury:
can you please add a description here for the cause of the crash?

PR #2091, opened 2025-07-31
spec: authenticated udp

dshulyak:
[rendered](https://github.com/category-labs/monad-bft/blob/dmitry/authspec/docs/0002_authenticated_udp.md)


docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,438 @@
...
+responder.hash = hash(&[&responder.hash, &temp2].concat());
+
+resp.encrypted_nothing = aead_encrypt(&key, 0, &[], &responder.hash);
+responder.hash = hash(&[&responder.hash, &resp.encrypted_nothing].concat());
+
+let mac_key = hash(&[LABEL_MAC1, &initiator_static].concat());
+resp.mac1 = mac(&mac_key, &resp.as_bytes()[..61]);
+resp.mac2 = [0u8; 16];
+```
+
```
dnkolegov-ar, line 267: I do not see the following consideration in the spec, but it is very important:

```
After a handshake is completed, with a message from initiator to responder and then responder back to initiator, the initiator may then send encrypted session packets, but the responder cannot.
The responder must wait to use the new session until it has recieved one encrypted session packet from the initiator, in order to provide key confirmation.
Thus, until the responder receives that first packet using the newly established session,
it must either queue up packets to be sent later, or use the previous session, if one exists and is valid.
Therefore, after the initiator receives the response from the responder,
if it has no data packets immediately queued up to send, it should send an empty packet, so as to provide this confirmation.
```

Source: https://www.wireguard.com/protocol/#subsequent-messages-exchange-of-data-packets

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,438 @@
...
+```
+if peer.has_capability(AUTHENTICATED_RAPTORCAST) AND session_established(peer):
+    send_encrypted(data, full_1472_bytes)
+else:
+    send_plaintext(data, reduced_1440_bytes)
+```
+Subsequent peers forward traffic through the encrypted port only if it was received on the encrypted port as well. This guarantees that encryption will be enabled gradually, without interrupting normal operations.
+
+---
```
dnkolegov-ar, line 523: I would suggest adding a Security Considerations section where you list all important implementation security requirements, though you may want to have a separate document for that: e.g., zeroization, ephemeral keys removing, test vectors, nonce reuse, etc.

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,438 @@
...
+        B->>B: Terminate session
+        Note over B: Clear session state
+    end
+```
+
+With a 10-second keepalive interval and 1-second random jitter, sessions terminate after 33 seconds without receiving any packet, accounting for network delays and jitter.
+
+### Rekeying
+
+For this protocol, rekeying is not required for security purposes. The primary consideration would be nonce exhaustion, but with a 64-bit counter providing 2^64 unique values, this is effectively unlimited for practical network operation.
```
dnkolegov-ar: Do you mean that "forward secrecy" is not the goal of this protocol at all? It would be good to know the protocol goals then. 

If forward secrecy is the goal, then rekeying is required because you probably want to provide forward secrecy and also post-compromise security in some sense.
Forward secrecy means that secrecy is guaranteed even when both the static private keys and the pre-shared key become known to the attacker, but only after the target session has been completed and all its keys (traffic, ephemeral, chaining, etc.) have been deleted by both parties. It is not possible to do if you do not have key rotation.

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,438 @@
...
+The protocol uses secp256k1, allowing validators to use their existing validator keys as static keys in the Noise protocol. This eliminates the need for separate identity management since validators authenticate using the same keys they use for consensus participation.
+
+AEGIS128L was selected based on performance measurements showing at least a 5x improvement over AES-GCM (approximately 200 nanoseconds versus 1 microsecond per packet) on processors with AES-NI extensions<sup>[5](#ref-5)</sup>.
+
+The choice between BLAKE3 and SHA256 is less critical for protocol performance. BLAKE3 was selected for consistency since it is already used in other parts of the client. Key derivation uses HMAC-based key derivation with BLAKE3 as the underlying hash function.
+
+The Noise Protocol Framework<sup>[2](#ref-2)</sup> provides the handshake structure using the IK pattern with pre-shared key support. This pattern assumes the initiator knows the responder's static public key. The complete construction string is:
+
+```
+Noise_IKpsk2_secp256k1_AEGIS128L_BLAKE3
```
dnkolegov-ar, line 53: How do you use PSK?

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,438 @@
...
+  - [Keepalive](#keepalive)
+  - [Rekeying](#rekeying)
+- [Integration](#integration)
+  - [RaptorCast Integration](#raptorcast-integration)
+  - [Discovery Integration](#discovery-integration)
+  - [Upgrade Process](#upgrade-process)
+
+## Introduction
+
+This specification defines an authenticated peer-to-peer protocol for UDP-based communication. The protocol addresses two primary concerns in the existing implementation.
```
dnkolegov-ar: Are there other goals? Encryption, forward secrecy, authenticity, integrity ?

docs/0002_authenticated_udp.md
dshulyak: i don't have a use case in mind, i just decided to keep because of the low implementation cost

docs/0002_authenticated_udp.md
dshulyak: primary goals are authenticity and integrity, i thought that it is implied by specified problems but i can try to make it more clear. encryption is nice to have, and might be more important in future. i don't see why would we care about forward secrecy

docs/0002_authenticated_udp.md
dshulyak: yes i meant forward secrecy. i don't see why would it be important for blockchain protocols, even if data is not immediately public (e.g tx's in mempool) it will become public after being included in the next block.

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,438 @@
...
+        return Err("Invalid cookie");
+    }
+}
+```
+
+## Session Management
+
+### Sessions
+
+The protocol supports two distinct session use cases: long-term sessions for known validator identities, which maintain persistent connections through active keepalive and automatic re-establishment after failures.
```
dnkolegov-ar, line 435: What are these sessions from the handshake protocol perspective? Are they the same? Do they need to have different domains, constants, etc?

docs/0002_authenticated_udp.md
dnkolegov-ar: I agree that authenticated channels are not considered with forward secrecy property, but rekeying, as it is implemented in WireGuard, also about post-compromise security (future secrecy).

If an attacker has compromised a validator and gained access to the transport keys, then the corresponding authenticated channels have also been compromised. If a protocol does not have rekeying, then I do not see how you can revoke these transport keys (revoking means deleting transport keys from memory, right?). In the original protocol, they will be invalidated by REKEY_AFTER_* mechanisms. You can look at the exact mechanism differently:  in practice, it may be important that an attacker can compromise validators several times but not always; in that case, the absence of rekeying (self-healing) also does not allow for secure future sessions.

That said, I think you must enable the original rekeying mechanism.

docs/0002_authenticated_udp.md
dnkolegov-ar: Agree, forward secrecy is not important, but future secrecy (post-compromise security) is important.

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,438 @@
...
+4-7: "Receiver Index"
+8-15: "Counter (8 bytes)"
+16-63: "Encrypted Payload (variable)"
+```
+
+Field descriptions:
+- Type: Message type identifier (4 for data) used to distinguish from handshake packets
+- Reserved: Padding for alignment and future protocol extensions
+- Receiver Index: Session identifier from handshake that routes packet to correct session
+- Counter: 64-bit nonce for AEAD encryption that prevents replay within session
```
dnkolegov-ar: Could you also describe the relation between the cipher nonce (16 bytes) and this counter (8 bytes)? What is the format, and how are they combined?

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,438 @@
...
+
+let encrypted_data = aead_encrypt(&sender.send_key, packet.counter, &plaintext_data, &[]);
+packet.encrypted_data = encrypted_data;
+```
+
+Nonces must never be reused. On the receiver side, replay protection should be implemented to account for possible packet reordering within a limited window. WireGuard sets this window to 2000 messages, requiring around 250 bytes per session using a bitmap. We can follow this approach and adjust if there is evidence of dropped packets.
+
+### Cookies
+
+The protocol implements a cookie mechanism to protect against denial-of-service attacks during handshake processing. When a node experiences high load, it can require initiators to prove recent interaction before allocating resources for expensive handshake operations.
```
dnkolegov-ar: > When a node experiences high load...

I know that is from the WireGuard spec, but I think it is too vague and should be described explicitly

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,438 @@
...
+    else Cookie invalid/expired
+        R->>I: CookieReply (new cookie)
+    end
+```
+
+Under normal operation, the MAC2 field in handshake packets remains zero. When under load, a responder replies with a cookie instead of processing the handshake. The initiator must then include this cookie in the MAC2 field of a retransmitted handshake initiation. This mechanism ensures that attackers cannot overwhelm nodes with spoofed source addresses, as only legitimate initiators who receive the cookie response can complete the handshake.
+
+Cookie replies are triggered when the global handshake request rate exceeds 1,000 requests per second. Known validators receive special treatment with a separate rate limit of 10 requests per second per validator, bypassing the global rate limit. This ensures legitimate validators can establish sessions even during attack scenarios while preventing any single validator from consuming excessive resources.
+
+The cookie construction uses BLAKE3 keyed hashing with a secret that rotates every two minutes, limiting cookie validity while preventing replay attacks across rotation periods. When under load, the responder generates a cookie:
```
dnkolegov-ar: Does cookie use the same AEAD primitive as transport?

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,438 @@
...
+let cookie_secret = get_or_rotate_cookie_secret();
+let cookie = make_cookie(&cookie_secret, &[
+    &initiator_ip.octets(),
+    &msg.sender_index.to_le_bytes()
+].concat());
+
+let mut reply = CookieReply::new();
+reply.message_type = 3;
+reply.receiver_index = msg.sender_index;
+reply.nonce = rand::random::<[u8; 24]>();
```
dnkolegov-ar: If AEGIS128L is used as AEAD here, then the nonce must be a random 16-byte. The original spec uses XChaCha20-Poly1305 for cookie encryption

dshulyak: yes the idea was to use aegis128l, and with random 16 bytes nonce. that line needs to be fixed

docs/0002_authenticated_udp.md
dshulyak: they are the same. in one case we expect consistent communication (validator <-> validator), and in the other it is not necessarily the caase, so i wanted to avoid keeping session open with keapalives

docs/0002_authenticated_udp.md
dshulyak: yes

docs/0002_authenticated_udp.md
dshulyak: i summarized what would be the rate limiting mechanism below. 2 ecdh operations are around 100us in total, so the idea is to keep sessions establishment in under 100ms/1s

> Cookie replies are triggered when the global handshake request rate exceeds 1,000 requests per second. Known validators receive special treatment with a separate rate limit of 10 requests per second per validator, bypassing the global rate limit.

docs/0002_authenticated_udp.md
dshulyak: i should probably rename this field to Nonce. it will be zero padded to 16 bytes

docs/0002_authenticated_udp.md
dshulyak: thanks, post-compromise security is definitely desirable. will implement add rekeying support

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,438 @@
...
+
+The second concern is traffic prioritization for transaction forwarding. The authentication flow guarantees that peers own their claimed IP addresses, enabling efficient prioritization based on either public key identity for known validators or IP address for general traffic. This allows network clients to implement quality-of-service policies that also prevent spam without computatonal overhead.
+
+## Protocol
+
+### Cryptography
+
+The protocol follows the WireGuard design<sup>[4](#ref-4)</sup> with several modifications to cryptographic primitives. These changes enable integration with existing validator keys and achieve performance improvements.
+
+The protocol uses secp256k1, allowing validators to use their existing validator keys as static keys in the Noise protocol. This eliminates the need for separate identity management since validators authenticate using the same keys they use for consensus participation.
```
dnkolegov-ar, line 42: It would be great to add requirements for secp256k1 public keys into the security consideration section:
- Correct encoding (exactly 33-byte compressed form).
- Point lies on the secp256k1 curve and is not the point at infinity.

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,438 @@
...
+
+Data packets use a 16-byte header containing the message type, receiver index, and nonce counter, followed by the encrypted payload and 16-byte authentication tag:
+
+```mermaid
+packet-beta
+0: "4"
+1-3: "Reserved"
+4-7: "Receiver Index"
+8-15: "Counter (8 bytes)"
+16-63: "Encrypted Payload (variable)"
```
dnkolegov-ar: Is it variable with a max size of 63? Also, 16 16-byte tag is not shown

docs/0002_authenticated_udp.md
dnkolegov-ar: The cookie reply message is not described and not shown

dnkolegov-ar: Also, is the cookie reply message 72 bytes as the picture shows?

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,438 @@
...
+## Integration
+
+### RaptorCast integration
+
+RaptorCast integration reserves 32 bytes in each datagram for header and AEAD encryption.
+
+```mermaid
+packet-beta
+0-15: "Reserved Header (16 bytes)"
+16-47: "RaptorCast Header & Data (1440 bytes)"
```
dnkolegov-ar: It is misleading: is it 32 bytes or 1440 bytes?

docs/0002_authenticated_udp.md
dshulyak: added section ### secp256k1 Key Format

docs/0002_authenticated_udp.md
dshulyak: extended a diagram above to clarify that responder should wait for first message and initiator should send zero sized message immediately

docs/0002_authenticated_udp.md
dshulyak: clarified on the diagram

docs/0002_authenticated_udp.md
dshulyak: added diagram, and corrected to 56 bytes

docs/0002_authenticated_udp.md
dshulyak: clarified that this is variable sized, at most 1440

docs/0002_authenticated_udp.md
dshulyak: will create test vectors when we start with implementation. 

nonce reuse i mentioned in sections above.

extended spec with recomendation to zeroize ephemeral private keys once hadnshake is considered complete or timed out. and zeroize symmetric keys once session is closed (for any reason)

what do you mean by ephemeral keys removing? just so that they are kept only in memory?

dnkolegov-ar: > what do you mean by ephemeral keys removing?

I meant that when we have finished rekeying, then we must remove (not zeroize) all key material and session keys from the previous sessions asap.

I think it is essential to separate these two mechanisms:
- zeroization - we are trying to minimise the number of memory regions where key material is stored, but there is memory where the keys are being stored since we use them.
- key removal: we do not need these keys anymore, so we must securely remove them

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,480 @@
...
+    &msg.sender_index.to_le_bytes()
+].concat());
+
+let mut reply = CookieReply::new();
+reply.message_type = 3;
+reply.receiver_index = msg.sender_index;
+reply.nonce = rand::random::<[u8; 16]>();
+
+let temp_key = hash(&[LABEL_COOKIE, &responder.static_public].concat());
+reply.cookie_encrypted = aead_encrypt_cookie(&temp_key, &reply.nonce, &cookie);
```
dnkolegov-ar: This interface is different from `aead_encrypt` and does not contain AD? Original spec used τ as plaintext and msg.mac1 as AD.  Is that intentional?

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,480 @@
...
+
+### Cryptography
+
+The protocol follows the WireGuard design<sup>[4](#ref-4)</sup> with several modifications to cryptographic primitives. These changes enable integration with existing validator keys and achieve performance improvements.
+
+The protocol uses secp256k1, allowing validators to use their existing validator keys as static keys in the Noise protocol. This eliminates the need for separate identity management since validators authenticate using the same keys they use for consensus participation.
+
+AEGIS128L was selected based on performance measurements showing at least a 5x improvement over AES-GCM (approximately 200 nanoseconds versus 1 microsecond per packet) on processors with AES-NI extensions<sup>[5](#ref-5)</sup>.
+
+The choice between BLAKE3 and SHA256 is less critical for protocol performance. BLAKE3 was selected for consistency since it is already used in other parts of the client. Key derivation uses HMAC-based key derivation with BLAKE3 as the underlying hash function.
```
dnkolegov-ar: BLAKE3 already offers keyed and derive_key modes - https://www.ietf.org/archive/id/draft-aumasson-blake3-00.html#section-1.1

docs/0002_authenticated_udp.md
dshulyak: no, it should have been consistent with aead_encrypt function that i used previously. i updated to use cookie as plaintext, and mac1 as AD.

docs/0002_authenticated_udp.md
dshulyak: updated to use keyed hash, as it seems drop in replacement

dnkolegov-ar (approved):

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,483 @@
...
+initiator.hash = hash(&[&initiator.hash, &msg.unencrypted_ephemeral].concat());
+
+let temp = blake3_keyed_hash(&initiator.chaining_key, &msg.unencrypted_ephemeral);
+initiator.chaining_key = blake3_keyed_hash(&temp, &[0x1]);
+
+let temp = blake3_keyed_hash(&initiator.chaining_key, &ecdh(&initiator.ephemeral_private, &responder.static_public));
+initiator.chaining_key = blake3_keyed_hash(&temp, &[0x1]);
+let key = blake3_keyed_hash(&temp, &[&initiator.chaining_key, &[0x2]].concat());
+
+msg.encrypted_static = aead_encrypt(&key, 0, &initiator.static_public, &initiator.hash);
```
dnkolegov-ar, line 201: AEGIS-128L uses 128-bit keys and nonces, but the key is 256 bits, since it is the default output for Blake3.
The spec never says whether and how to truncate to 16 bytes.

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,483 @@
...
+
+The initiator creates a handshake initiation message:
+
+```rust
+const CONSTRUCTION: &[u8] = b"Noise_IKpsk2_secp256k1_AEGIS128L_BLAKE3";
+const IDENTIFIER: &[u8] = b"authenticated udp v1 -- monad";
+const LABEL_MAC1: &[u8] = b"mac1----";
+const LABEL_COOKIE: &[u8] = b"cookie--";
+
+initiator.chaining_key = hash(CONSTRUCTION);
```
dnkolegov-ar, line 179: hash and MAC are not formally defined. 
Suggestion is  to use the similar notations as in the WireGuard whitepaper pages 9 - 10:

For example (from the WireGuard spec):
```
Hash(input) Blake2s(input, 32), returning 32 bytes of output.
```

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,483 @@
...
+
+```rust
+let cookie_secret = get_or_rotate_cookie_secret();
+let cookie = make_cookie(&cookie_secret, &[
+    &initiator_ip.octets(),
+    &msg.sender_index.to_le_bytes()
+].concat());
+
+let mut reply = CookieReply::new();
+reply.message_type = 3;
```
dnkolegov-ar, line 402: `reply.reserved` is missing

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,483 @@
...
+reply.receiver_index = msg.sender_index;
```
dnkolegov-ar, line 404: the original is `msg.receiver_index = little_endian(initiator.sender_index)`

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,483 @@
...
+    else Cookie invalid/expired
+        R->>I: CookieReply (new cookie)
+    end
+```
+
+The cookie construction uses BLAKE3 keyed hashing with a secret that rotates every two minutes, limiting cookie validity while preventing replay attacks across rotation periods. When under load, the responder generates a cookie:
+
+```rust
+let cookie_secret = get_or_rotate_cookie_secret();
+let cookie = make_cookie(&cookie_secret, &[
```
dnkolegov-ar: in WireGuard `cookie = MAC(responder.changing_secret_every_two_minutes, initiator.ip_address)`, here
- `make_cookie ` is not defined
- `msg.sender_index` is also used

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,483 @@
...
+```rust
+let temp = blake3_keyed_hash(&chaining_key, &[]);
+let temp2 = blake3_keyed_hash(&temp, &[0x1]);
+let send_key = blake3_keyed_hash(&temp, &[&temp2, &[0x2]].concat());
+let recv_key = blake3_keyed_hash(&temp, &[&temp2, &[0x3]].concat());
+```
+
+The initiator and responder derive keys in opposite order, ensuring proper bidirectional communication.
+
+After transport keys have been derived, ephemeral private keys should be zeroized as they are no longer needed.
```
dnkolegov-ar: Also, KDF materials (temp, temp2 ...), decrypted cookies

docs/0002_authenticated_udp.md
dshulyak: as i understand, the msg here is handshake_initiation, so it is consistent with original spec

docs/0002_authenticated_udp.md
dshulyak: added such notation

docs/0002_authenticated_udp.md
dshulyak: clarified that we will use first 16 bytes

docs/0002_authenticated_udp.md
dshulyak: simplified that, also removed sender_index from cookie, seems like there is no reason to disallow different sender_index

docs/0002_authenticated_udp.md
dshulyak: added for consistency with other code blocks

docs/0002_authenticated_udp.md
dshulyak: included

docs/0002_authenticated_udp.md
dnkolegov-ar: My comment was about endianess.
This document does not say which endianness it uses. WireGuard spec uses both endianness. I think it is important to clarify that.

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,496 @@
...
+initiator.hash = hash(&[&initiator.hash, &msg.encrypted_static].concat());
+
+let temp = keyed_hash(&initiator.chaining_key, &ecdh(&initiator.static_private, &responder.static_public));
+initiator.chaining_key = keyed_hash(&temp, &[0x1]);
+let key = keyed_hash(&temp, &[&initiator.chaining_key, &[0x2]].concat());
+
+msg.encrypted_timestamp = aead_encrypt(&key, 0, &tai64n(), &initiator.hash);
+initiator.hash = hash(&[&initiator.hash, &msg.encrypted_timestamp].concat());
+
+let mac_key = hash(&[LABEL_MAC1, &responder.static_public].concat());
```
dnkolegov-ar, line 211: `responder` may be misleading here. The original spec uses trickier notation:

```
For messages that can be created by either an initiator
or responder, if the peer creating the message is the initiator, let (m, m') = (i, r), and if the peer creating the message is the responder, let (m, m') = (r, i). 
```

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,496 @@
...
+let temp2 = keyed_hash(&temp, &[&responder.chaining_key, &[0x2]].concat());
+let key = keyed_hash(&temp, &[&temp2, &[0x3]].concat());
+responder.hash = hash(&[&responder.hash, &temp2].concat());
+
+resp.encrypted_nothing = aead_encrypt(&key, 0, &[], &responder.hash);
+responder.hash = hash(&[&responder.hash, &resp.encrypted_nothing].concat());
+
+let mac_key = hash(&[LABEL_MAC1, &initiator_static].concat());
+resp.mac1 = mac(&mac_key, &resp.as_bytes()[..61]);
+resp.mac2 = [0u8; 16];
```
dnkolegov-ar: There is no place in the spec that sets `resp.mac2` if mac2 is not zero bytes

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,496 @@
...
+
+**Note:** AEGIS128L requires a 16-byte nonce. The 8-byte nonce is zero-padded to 16 bytes before use (e.g., nonce value 0x0123456789ABCDEF becomes 0x0123456789ABCDEF0000000000000000).
+
+Nonces must never be reused. On the receiver side, replay protection should be implemented to account for possible packet reordering within a limited window. WireGuard sets this window to 2000 messages, requiring around 250 bytes per session using a bitmap. We can follow this approach and adjust if there is evidence of dropped packets.
+
+When a session is closed (either due to timeout, explicit termination, or rekeying), all symmetric key material (send_key and recv_key) must be zeroized.
+
+### Cookies
+
+Under normal operation, the MAC2 field in handshake packets remains zero. When under load, a responder replies with a cookie instead of processing the handshake. The initiator must then include this cookie in the MAC2 field of a retransmitted handshake initiation. This mechanism ensures that attackers cannot overwhelm nodes with spoofed source addresses, as only legitimate initiators who receive the cookie response can complete the handshake.
```
dnkolegov-ar: Is it correct to say that the cookie is included in MAC2?

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,496 @@
...
+    participant R as Responder (Under Load)
+    
+    I->>R: HandshakeInitiation (MAC2 = 0)
+    Note over R: Detects high load conditions
+    
+    R->>I: CookieReply (56 bytes)
+    Note left of R: type=3, receiver_idx, nonce(16), encrypted_cookie+tag(32)
+    
+    I->>I: Store cookie
+    I->>R: HandshakeInitiation (MAC2 = cookie)
```
dnkolegov-ar: MAC2 is not equal to the cookie

docs/0002_authenticated_udp.md
dshulyak: is big endian only for timestamp in the original spec? i will mention that beside that all integers are little endian

docs/0002_authenticated_udp.md
dshulyak: this section is consistent with wireguard description. i don't think that there is any ambiguity here, initiator creates handshake initiation and responder is whoever will reply to that.

> msg.mac1 = MAC(HASH(LABEL_MAC1 || responder.static_public), msg[0:offsetof(msg.mac1)])

docs/0002_authenticated_udp.md
dshulyak: will update this snippet and for reply as well

docs/0002_authenticated_udp.md
dshulyak: i guess it can be confusing, probably "use cookie to compute mac2" is more accurate

docs/0002_authenticated_udp.md
dnkolegov-ar: correct

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,515 @@
...
+    else Cookie invalid/expired
+        R->>I: CookieReply (new cookie)
+    end
+```
+
+The cookie construction uses BLAKE3 keyed hashing with a secret that rotates every two minutes, limiting cookie validity while preventing replay attacks across rotation periods. When under load, the responder generates a cookie:
+
+```rust
+let cookie_secret = get_or_rotate_cookie_secret();
+let cookie = mac(&cookie_secret, &initiator_ip);
```
dnkolegov-ar: Just the initiator IP without the port?
In WireGuard `Am' represents a concatenation of the subscript’s external IP source address and UDP source port`

docs/0002_authenticated_udp.md
dnkolegov-ar: The misleading originates from the fact that `Cookie Replay` can be sent as a response to `Handshake Response`, not just `Handshake Initiation`:

```
when a message with a valid msg.mac1 is received, but msg.mac2 is invalid or
expired, and the peer is under load, the peer may send a cookie reply message
```
So, `initiator` and `responder` have different meanings.

docs/0002_authenticated_udp.md
dnkolegov-ar: This one is better

docs/0002_authenticated_udp.md
dshulyak: thanks, i didn't notice that in whitepaper that they use full address, i guess we can do as well

docs/0002_authenticated_udp.md
dshulyak: so in the snippet for cookies the **responder.static_public** is the local static key, regardless of the roles in handshake. i think that should be clear from context, but maybe using local/remote keywords in the snippets will make it obvious

> let temp_key = hash(&[LABEL_COOKIE, &responder.static_public].concat());
reply.cookie_encrypted = aead_encrypt(&temp_key, &reply.nonce, &cookie, &msg.mac1);

dshulyak: i added a note below cookies snippet that clarifies it

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,518 @@
...
+```
+
+#### Cryptographic Function Definitions
+
+Following the notation convention from the WireGuard whitepaper:
+
+- hash(input): BLAKE3(input), returning 32 bytes of output
+- mac(key, input): BLAKE3 in keyed mode, returning 16 bytes (first 16 bytes of the 32-byte output)
+- keyed_hash(key, input): BLAKE3 in keyed mode with the specified key, returning 32 bytes
+- aead_encrypt(key, nonce, plaintext, ad): AEGIS128L encryption with 16-byte key (when key derivation produces 32 bytes, use first 16 bytes), nonce (8 bytes zero-padded to 16), plaintext, and associated data, returning ciphertext with 16-byte authentication tag appended
```
dnkolegov-ar: This is inconsistent with `aead_encrypt` for cookies where `nonce` is defined as follows:
```
reply.nonce = rand::random::<[u8; 16]>();
```

docs/0002_authenticated_udp.md
dnkolegov-ar: PSK should be defined as a constant then, or completely removed. There is NoiseIK pattern without PSK at all, so this protocol can be easily adapted for that change; otherwise, it should be defined as a randomly generated constant.

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,518 @@
...
+```
+
+Field descriptions:
+- Type: Message type identifier (1 for initiation) used to distinguish packet types
+- Reserved: Padding for alignment and future protocol extensions
+- Sender Index: Random 4-byte identifier for this session used in response packets
+- Ephemeral Public Key: Per-session ephemeral identity
+- Encrypted Static Key: Encrypted initiator's long-term identity
+- Encrypted Timestamp: TAI64N timestamp to prevent replay attacks across sessions
+- MAC1: Authentication of entire packet using responder's public key hash
```
dnkolegov-ar, line 103: I think the current description doesn't explicitly say the following but it should:

```
... the responder, after computing these MACs as well and comparing them to the ones received in the
message, must always reject messages with an invalid msg.mac1, and when under load may reject messages with
an invalid msg.mac2. 
```

Easy to implement in the wrong order.

Suggestion: add something like

```
Upon receiving a handshake initiation, endpoints MUST verify mac1 before performing any ECDH, decryption, or allocation. Under load, if mac2 is zero or invalid, send CookieReply and return
```

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,518 @@
...
+packet.reserved_zero = [0, 0, 0];
+packet.receiver_index = peer.receiver_index;
+packet.nonce = sender.nonce;
+sender.nonce += 1;
+
+let encrypted_data = aead_encrypt(&sender.send_key, packet.nonce, &plaintext_data, &[]);
+packet.encrypted_data = encrypted_data;
+```
+
+**Note:** AEGIS128L requires a 16-byte nonce. The 8-byte nonce is zero-padded to 16 bytes before use (e.g., nonce value 0x0123456789ABCDEF becomes 0x0123456789ABCDEF0000000000000000).
```
dnkolegov-ar: Now it should be in little-endian, right?

dshulyak: removed this

docs/0002_authenticated_udp.md
dshulyak: added an exception

docs/0002_authenticated_udp.md
dshulyak: i tend to think that such details better to leave for the implementation. 

i agree that it is possible to forget and missimplement it, but i think it should be clear from context that mac1 should be checked right after receiving message, and pointer in cookies section to check mac2 if it is not empty

docs/0002_authenticated_udp.md
dshulyak: my intention was to make it optional and default to [0u8; 32], we will set it to some value if we find use case for it, for example limiting testnet traffic or forks.

it will be consistent with boringtun implementation https://github.com/cloudflare/boringtun/blob/master/boringtun/src/noise/handshake.rs#L605C12-L609

dnkolegov-ar: Sounds good. From the previous message, my impression was that PSK is here because it is in WireGuard

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,516 @@
...
+    
+    I->>R: HandshakeInitiation (MAC2 = 0)
+    Note over R: Detects high load conditions
+    
+    R->>I: CookieReply (56 bytes)
+    Note left of R: type=3, receiver_idx, nonce(16), encrypted_cookie+tag(32)
+    
+    I->>I: Store cookie
+    I->>R: HandshakeInitiation
+    Note right of I: Same packet with cookie in MAC2
```
dnkolegov-ar:
```suggestion
    Note right of I: Retry initiation with MAC2 bound to cookie
```

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,516 @@
...
+## Integration
+
+### RaptorCast integration
+
+RaptorCast integration reserves 32 bytes in each datagram for header and AEAD encryption.
+
+```mermaid
+packet-beta
+0-15: "Reserved Header (16 bytes)"
+16-47: "RaptorCast Header & Data (atmost 1440 bytes)"
```
dnkolegov-ar:
```suggestion
16-47: "RaptorCast Header & Data (at most 1440 bytes)"
```

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,516 @@
...
+48-63: "Reserved Tag (16 bytes)"
+```
+
+**Note:** The actual RaptorCast packet data is variable-sized. The diagram shows the full 1440-byte allocation, individual packets may be smaller.
+
+This reservation applies to all upgraded nodes, regardless of whether specific peers support authentication. When communicating with legacy nodes, the sender uses only 1440 bytes of payload in plaintext, ensuring compatibility. When both peers support authentication, the full 1472-byte packet carries encrypted data.
+
+### Discovery Integration
+
+The peer discovery service advertises authentication capability through a new flag<sup>[6](#ref-6)</sup>:
```
dnkolegov-ar, line 499: Is this a separate discovery service? Or did this service work on top of authenticated UDP?  Could you elaborate more about this?

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,516 @@
...
+
+The choice between BLAKE3 and SHA256 is less critical for protocol performance. BLAKE3 was selected for consistency since it is already used in other parts of the client. Key derivation uses BLAKE3's keyed hash mode for key derivation functions.
+
+The Noise Protocol Framework<sup>[2](#ref-2)</sup> provides the handshake structure using the IK pattern with pre-shared key support. This pattern assumes the initiator knows the responder's static public key. The complete construction string is:
+
+```
+Noise_IKpsk2_secp256k1_AEGIS128L_BLAKE3
+```
+
+#### Cryptographic Function Definitions
```
dnkolegov-ar, line 56: PSK size/default not stated. You don’t specify its size or the disabled default.
I suggest saying that the PSK should have 128-bit entropy, since it is the security level of secp256k1

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,516 @@
...
+let encrypted_data = aead_encrypt(&sender.send_key, packet.nonce, &plaintext_data, &[]);
+packet.encrypted_data = encrypted_data;
+```
+
+Nonces must never be reused. On the receiver side, replay protection should be implemented to account for possible packet reordering within a limited window. WireGuard sets this window to 2000 messages, requiring around 250 bytes per session using a bitmap. We can follow this approach and adjust if there is evidence of dropped packets.
+
+When a session is closed (either due to timeout, explicit termination, or rekeying), all symmetric key material (send_key and recv_key) must be zeroized.
+
+### Cookies
+
```
dnkolegov-ar, line 342:
```suggestion
The cookie mechanism is implemented using the MAC1 and MAC2 fields.  After computing these MACs according to the cryptographic computations as well and comparing them to the ones received in the handshake message, it must always reject messages with an invalid MAC1 field, and when under load, it may reject messages with an invalid MAC2 before any ECDH, decryption, or state allocation.”
```

docs/0002_authenticated_udp.md
dshulyak: this is a separate service that was added recently, there is short description in docs folder.
and yes the plan is to use auth udp for it as well.

basically we need to know if peer supports authentication, as not everyone will update at the same time. therefore we introduce a flag in the discovery message that will be propagated across the network, and if this flag is set then we know that peer supports authentication and we will try to establish session instead of sending data in plaintext

docs/0002_authenticated_udp.md
dshulyak: added a note on importance of mac1/mac2 validation before expensive operations, but in Message Types and Flows section

docs/0002_authenticated_udp.md
dshulyak: boringtun uses 32 bytes (i guess wireguard implementation as well), so i guess i will choose that too

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,520 @@
...
+#### Cryptographic Function Definitions
+
+Following the notation convention from the WireGuard whitepaper:
+
+- hash(input): BLAKE3(input), returning 32 bytes of output
+- mac(key, input): BLAKE3 in keyed mode, returning 16 bytes (first 16 bytes of the 32-byte output)
+- keyed_hash(key, input): BLAKE3 in keyed mode with the specified key, returning 32 bytes
+- aead_encrypt(key, nonce, plaintext, ad): AEGIS128L encryption with 16-byte key (when key derivation produces 32 bytes, use first 16 bytes), nonce (8 bytes zero-padded to 16 bytes, except for cookies which use 16-byte random nonce), plaintext, and associated data, returning ciphertext with 16-byte authentication tag appended
+- aead_decrypt(key, nonce, ciphertext, ad): AEGIS128L decryption with 16-byte key (when key derivation produces 32 bytes, use first 16 bytes), nonce (8 bytes zero-padded to 16 bytes, except for cookies which use 16-byte random nonce), ciphertext (including tag), and associated data, returning plaintext or error
```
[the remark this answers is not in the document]
dnkolegov-ar: @dshulyak I agree that it creates complexity, but it is up to you how to solve that: on design level or implementation level. What do you think about Copilot's suggestion?

docs/0002_authenticated_udp.md
dnkolegov-ar: What if an active attacker changes that flag? The target nodes will never switch to the authenticated protocol. Are these discovery messages signed?

docs/0002_authenticated_udp.md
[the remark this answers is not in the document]
dshulyak: i don't see it it as a useful or important suggestion.
from my point of view the only important detail is how to get 16 bytes for aegis if derivation function produces larger buffer, and we will always truncate to 16. and in the implementation we may have a dedicated function for that to avoid mistakes

as for nonces i think copilots comment is just too random

docs/0002_authenticated_udp.md
dshulyak: yes, they are signed

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,528 @@
...
+
+The rekeying process follows the standard session establishment protocol. Once the new session is established and confirmed (after the initiator sends the first encrypted packet), both participants are free to drop the original session from memory. During the transition period, implementations should accept packets on both the old and new sessions to ensure seamless rekeying without packet loss.
+
+Regarding nonce exhaustion: with a 64-bit nonce providing 2^64 unique values, this is not a practical concern. At 1 Gbps throughput with 1500-byte packets, the maximum packet rate is approximately 83,000 packets per second. With a 64-bit nonce space providing approximately 1.8e19 unique values, nonce exhaustion would occur after 2.2e14 seconds, or approximately 7 million years. Even at theoretical 10 Gbps rates with minimal 64-byte packets, exhaustion would take over 30,000 years.
```
dnkolegov-ar, line 478: What is the goal of decreasing the 128-bit nonce to a 64-bit nonce? 
My concern is that in https://www.ietf.org/archive/id/draft-irtf-cfrg-aegis-aead-03.html#name-quic-and-dtls-13-header-pro, Quic and DTLS use the same nonce size as the original protocol, and at the same time, they operate in the same settings as this protocol

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,528 @@
...
+  - [Discovery Integration](#discovery-integration)
+  - [Upgrade Process](#upgrade-process)
+
+## Introduction
+
+This specification defines an authenticated peer-to-peer protocol for UDP-based communication. The protocol provides authenticity and integrity for UDP traffic, ensuring that packets come from verified peers and have not been tampered with. While encryption is not a hard requirement, the protocol includes it as the additional computational cost is negligible when using modern AEAD ciphers (specifically AEGIS128L).
+
+The protocol addresses two primary concerns in the existing implementation.
+
+The first concern is denial-of-service protection for RaptorCast communication. Current implementations require ECDSA signature verification for each packet, incurring approximately 20-30 microseconds of computational overhead per datagram. This creates a vulnerability where malicious actors can overwhelm validators with invalid signatures. By establishing authenticated sessions using symmetric AEAD operations, the protocol reduces per-datagram overhead to approximately 200 nanoseconds while providing authenticity, integrity, and encryption.
```
dnkolegov-ar, line 32: Have you already conducted experiments that demonstrate the protocol's effectiveness in protecting against DoS attacks? It is especially interesting when the malicious peer is a member of the epoch validators.

dnkolegov-ar, line 32: Do you have any ideas on how this protocol can be combined with RaptorCast DoS protection mechanism (Tiered cache and the eviction algorithm)? For now, they are separate mechanisms; this is my understanding. Is it possible to combine them to provide more protection?

docs/0002_authenticated_udp.md
dshulyak: wireguard uses 64 bits nonce. what is the point of increasing nonce size (and header size for every message) when 64 bits is more than enough?

docs/0002_authenticated_udp.md
dshulyak: no, but i think there is a clear set of items that need to be implemented in order for raptorcast to be safe to use for validators

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,528 @@
...
+
+#### Cryptographic Function Definitions
+
+Following the notation convention from the WireGuard whitepaper:
+
+- hash(input): BLAKE3(input), returning 32 bytes of output
+- mac(key, input): BLAKE3 in keyed mode, returning 16 bytes (first 16 bytes of the 32-byte output)
+- keyed_hash(key, input): BLAKE3 in keyed mode with the specified key, returning 32 bytes
+- aead_encrypt(key, nonce, plaintext, ad): AEGIS128L encryption with 16-byte key (when key derivation produces 32 bytes, use first 16 bytes), nonce (8 bytes zero-padded to 16 bytes, except for cookies which use 16-byte random nonce), plaintext, and associated data, returning ciphertext with 16-byte authentication tag appended
+- aead_decrypt(key, nonce, ciphertext, ad): AEGIS128L decryption with 16-byte key (when key derivation produces 32 bytes, use first 16 bytes), nonce (8 bytes zero-padded to 16 bytes, except for cookies which use 16-byte random nonce), ciphertext (including tag), and associated data, returning plaintext or error
```
dnkolegov-ar, line 64: WireGuard uses a 24-byte random nonce. I think it is a significant difference from the original idea, and it is not possible to use the counter mode anymore. Since the chosen cipher can't use 24-byte nonce, did you check that it is safe to decrease the nonce size? It would be safe to use AEGIS256 for cookies

docs/0002_authenticated_udp.md
dshulyak: i don't think that they will be tightly coupled, but raptorcast will use authentication data to prevent resource exhaustion.

validator will reject any raptorcast-encoded message from non-validator, before doing any computationally expensive work. for misbehaving validator it will be possible to rate limit them, again before any expensive computation.

discovery will be able to get rid of signatures for authentication.

tx ingestion will get rid of signing messages themself as well, and will need to implement rate limiting for tx verification, either based on some sort of reputation scores or design like swqos in solana.

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,528 @@
...
+TYPE_HANDSHAKE_RESPONSE   = 2
+TYPE_COOKIE_REPLY         = 3  
+TYPE_DATA                 = 4
+```
+
+Handshake initiation packets contain 150 bytes:
+
+```mermaid
+packet-beta
+0: "1"
```
dnkolegov-ar: Should it be `Type` as in the `Field descriptions`?

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,528 @@
...
+- Type: Message type identifier (2 for response) used to distinguish packet types
+- Reserved: Padding for alignment and future protocol extensions
+- Sender Index: Random 4-byte identifier for responder's session endpoint
+- Receiver Index: Echo of initiator's sender index used to correlate response
+- Ephemeral Public Key: Per-session ephemeral identity 
+- Encrypted Nothing: Empty payload proves responder can derive shared secrets
+- MAC1: Authentication of entire packet using initiator's static key hash
+- MAC2: Cookie response MAC, empty unless under load (DoS protection)
+
+Session establishment follows the Noise IK handshake pattern:
```
dnkolegov-ar:
```suggestion
Session establishment follows the NoiseIKpsk2 handshake pattern:
```

docs/0002_authenticated_udp.md
dshulyak: in aegis-128l docs https://libsodium.gitbook.io/doc/secret-key_cryptography/aead/aegis-128l

> It has a 128 bit nonce size, allowing random nonces to be safely used up to 2^48 messages for a single key.

but just to be clear about the concern. the risk is that after 2^48 messages we risk to leak the key which is used for encryption, the key is set to `hash!(LABEL_COOKIE, responder_static_public.as_bytes())`. so if requestor is replied every 1s, thats 17,850,570 years. and 1s is clearly too frequent.

in my opinion this is non-issue

docs/0002_authenticated_udp.md
dshulyak: another thought is that dtls and quic by design have to be generic and very future proof. in our case we know approximate load and more flexible (e.g updating quic will be much more involved process).

but again, i just don't see why would we waste 8 more bytes in every datagram

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,528 @@
...
+- Nonce: 64-bit nonce for AEAD encryption that prevents replay within session
+- Authentication Tag: 16-byte AEGIS128L authentication tag in header
+- Encrypted Payload: Application data encrypted with session keys
+
+After the handshake completes, both parties derive transport keys:
+
+```rust
+let temp = keyed_hash(&chaining_key, &[]);
+let temp2 = keyed_hash(&temp, &[0x1]);
+let send_key = keyed_hash(&temp, &[&temp2, &[0x2]].concat());
```
dnkolegov-ar: In WireGuard, these keys are calculated differently. It should be
```
let temp = keyed_hash(&chaining_key, &[]);
let send_key = keyed_hash(&temp, &[0x1]);
let recv_key = keyed_hash(&temp, &[&send_key, &[0x2]].concat());

```
Is it intentional?

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,528 @@
...
+        R->>I: HandshakeResponse
+        Note over I,R: Session Established
+    else Cookie invalid/expired
+        R->>I: CookieReply (new cookie)
+    end
+```
+
+The cookie construction uses BLAKE3 keyed hashing with a secret that rotates every two minutes, limiting cookie validity while preventing replay attacks across rotation periods.
+
+The cookie reply nonce is derived from a nonce secret obtained at startup and a monotonically increasing counter. This provides deterministic nonce generation without relying on system randomness for each cookie reply. When under load, the responder generates a cookie:
```
dnkolegov-ar: What is the reason to do that (secret nonce, counter) instead of just randomly generating it?

dshulyak: to avoid locking when reading randomness from os. i looked up this in boringtun implementation https://github.com/cloudflare/boringtun/blob/master/boringtun/src/noise/rate_limiter.rs#L29-L37

docs/0002_authenticated_udp.md
dshulyak: thanks, i will revert to wireguard specification, i can't recall now if i reused it from somewhere or just mistyped

dshulyak:
i updated spec to bind cookies only to the ip addresss (excluding port). this is to simplify integration with our TCP implementation, we don't use same connections for sending/receiving and changing that is not very simple. i don't think that we lose anything by removing port from the scheme, as the goal is to prove ownership of the ip address

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,529 @@
...
+    if msg.mac2 != expected_mac2 {
+        return Err("Invalid cookie");
+    }
+}
+```
+
+Note that responder and initiator in the context of cookie replies are unrelated to handshake initiation. The initiator sends the message, and the responder replies to it.
+The initiator IP address is the message source IP address (without port). The responder's static public key is always the local static public key.
+
+For TCP, our implementation uses separate connections for sending and receiving, with outbound connections assigned random source ports. This means cookie replies arrive from a different port than the original handshake destination. By binding cookies only to IP addresses, the protocol remains functional as long as replies originate from the same IP address.
```
dnkolegov-ar: Why do you need cookie in the TCP case?

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,529 @@
...
+- Type: Message type identifier (1 for initiation) used to distinguish packet types
+- Reserved: Padding for alignment and future protocol extensions
+- Sender Index: Random 4-byte identifier for this session used in response packets
+- Ephemeral Public Key: Per-session ephemeral identity
+- Encrypted Static Key: Encrypted initiator's long-term identity
+- Encrypted Timestamp: TAI64N timestamp to prevent replay attacks across sessions
+- MAC1: Authentication of entire packet using responder's public key hash
+- MAC2: Cookie response MAC, empty unless under load (DoS protection)
+
+Implementations must validate MAC fields before performing expensive operations such as ECDH, decryption, or state allocation. Messages with invalid MAC1 must always be rejected. When under load, messages with invalid MAC2 should also be rejected (see Cookies section).
```
dnkolegov-ar: Note: Ambiguous about ordering. WireGuard’s rule is: MAC1 first (always); then, only if under load check MAC2, then check rate limiting.

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,529 @@
...
+        R->>I: HandshakeResponse
+        Note over I,R: Session Established
+    else Cookie invalid/expired
+        R->>I: CookieReply (new cookie)
+    end
+```
+
+The cookie construction uses BLAKE3 keyed hashing with a secret that rotates every two minutes, limiting cookie validity while preventing replay attacks across rotation periods.
+
+The cookie reply nonce is derived from a nonce secret obtained at startup and a monotonically increasing counter. This provides deterministic nonce generation without relying on system randomness for each cookie reply. When under load, the responder generates a cookie:
```
dnkolegov-ar: To me, "pseudorandom nonce generation" is correct.

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,529 @@
...
+
+Nonces must never be reused. On the receiver side, replay protection should be implemented to account for possible packet reordering within a limited window. WireGuard sets this window to 2000 messages, requiring around 250 bytes per session using a bitmap. We can follow this approach and adjust if there is evidence of dropped packets.
+
+When a session is closed (either due to timeout, explicit termination, or rekeying), all symmetric key material (send_key and recv_key) must be zeroized.
+
+### Cookies
+
+Under normal operation, the MAC2 field in handshake packets is set to zero and its validation can be skipped. When under load, a responder replies with a cookie instead of processing the handshake. The initiator must use this cookie to compute MAC2 field of a retransmitted handshake initiation. This mechanism ensures that attackers cannot overwhelm nodes with spoofed source addresses, as only legitimate initiators who receive the cookie response can complete the handshake.
+
+Cookie replies are triggered when the global handshake request rate exceeds 2,000 requests per second.
```
dnkolegov-ar: "Under load" is used several times in the spec, but this line clearly defines what "under load" is. Suggestion: define "under load" via a rate exceeding 2,000 requests per second.

docs/0002_authenticated_udp.md
dnkolegov-ar: What is this section trying to explain if this spec is for UDP?

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,529 @@
...
+    
+    R->>R: Verify cookie
+    alt Cookie valid
+        R->>I: HandshakeResponse
+        Note over I,R: Session Established
+    else Cookie invalid/expired
+        R->>I: CookieReply (new cookie)
+    end
+```
+
```
dnkolegov-ar, line 386: Suggested insertion:

Upon receiving a valid CookieReply, the initiator must only store the cookie. Receipt of a CookieReply must not, by itself, trigger retransmission; retransmission of the handshake initiation with MAC2 is governed by the retransmission logic.

docs/0002_authenticated_udp.md
```diff
@@ -0,0 +1,529 @@
...
+After the handshake completes, both parties derive transport keys:
+
+```rust
+let temp = keyed_hash(&chaining_key, &[]);
+let send_key = keyed_hash(&temp, &[0x1]);
+let recv_key = keyed_hash(&temp, &[&send_key, &[0x2]].concat());
+```
+
+The initiator and responder derive keys in opposite order, ensuring proper bidirectional communication.
```
dnkolegov-ar: Suggestion:


After the handshake completes, both parties derive transport keys:

```rust
let t1 = keyed_hash(&chaining_key, &[]);
let t2 = keyed_hash(&t1, &[0x01]);
let t3 = keyed_hash(&t1, &[&t2, &[0x02]].concat());
let k2 = &t2[..16];
let k3 = &t3[..16];
```

The initiator keys are:
```
let send_key = k2;
let recv_key = k3;
```

The responder keys are:
```
let send_key = k3;
let recv_key = k2;
```

docs/0002_authenticated_udp.md
dshulyak: fair enough, tcp part is out of place and i will just document it in code

dshulyak: it is not needed, the idea is to reuse the same protocol as is for tcp as well as right now it depends on signatures for auth. good point, maybe i can disable cookies just by tweaking configuration 

optimal solution would likely be noise session per connection, that makes whole protocol much simpler, but it will require significant refactoring that is not worth it in my opinion

docs/0002_authenticated_udp.md
dshulyak: sorry i am not following, is it about wording?

docs/0002_authenticated_udp.md
dshulyak: i will drop this from the spec and leave up to implementation, i think i actually deviated from this definition already

docs/0002_authenticated_udp.md
dshulyak: inserted

docs/0002_authenticated_udp.md
dshulyak: added

docs/0002_authenticated_udp.md
dnkolegov-ar: I just suggest using "pseudorandom" instead of "deterministic"

PR #2092, opened 2025-07-31
RaptorCast: DOS protection on decoding state cache

xinyuan-dev:
Implements https://github.com/category-labs/category-internal/issues/1801.

### DIFF
--- a/monad-raptorcast/src/decoding.rs
+++ b/monad-raptorcast/src/decoding.rs
@@ -1,0 +1,2156 @@
+// Copyright (C) 2025 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#![allow(clippy::manual_range_contains)]
+use std::{
+    collections::{BTreeMap, BTreeSet, HashMap, HashSet},
+    hash::Hash,
+    num::NonZero,
+    sync::Arc,
+    time::{Duration, Instant},
+};
+
+use bitvec::prelude::*;
+use bytes::Bytes;
+use indexmap::IndexMap;
+use lru::LruCache;
+use monad_crypto::{
+    certificate_signature::PubKey,
+    hasher::{Hasher as _, HasherType},
+};
+use monad_raptor::{ManagedDecoder, SOURCE_SYMBOLS_MIN};
+use monad_types::{Epoch, NodeId, Stake};
+use rand::Rng as _;
+
+use crate::{
+    udp::{ValidatedMessage, MAX_REDUNDANCY},
+    util::{compute_hash, AppMessageHash, HexBytes, NodeIdHash},
+};
+
+pub(crate) const RECENTLY_DECODED_CACHE_SIZE: usize = 10000;
+
+pub(crate) const BROADCAST_TIER_CONFIG: SoftQuotaCacheConfig = SoftQuotaCacheConfig {
+    total_slots: 1000,
+    min_slots_per_author: 5,
+    max_total_size_per_author: 5 * 4 * 1024 * 1024, // 5*4 MB
+    min_slots_per_validator: None,
+    max_total_size_per_validator: None,
+};
+
+pub(crate) const VALIDATOR_TIER_CONFIG: SoftQuotaCacheConfig = SoftQuotaCacheConfig {
+    total_slots: 600,
+    min_slots_per_author: 3,
+    max_total_size_per_author: 1024 * 1024, // 1 MB
+    min_slots_per_validator: None,
+    max_total_size_per_validator: None,
+};
+
+pub(crate) const P2P_TIER_CONFIG: SoftQuotaCacheConfig = SoftQuotaCacheConfig {
+    total_slots: 500,
+    min_slots_per_author: 1,
+    max_total_size_per_author: 1024 * 1024, // 1 MB
+    min_slots_per_validator: None,
+    max_total_size_per_validator: None,
+};
+
+// An abstract size of a message loosely corresponding to the memory
+// usage of its corresponding cache entry. We currently use
+// app_message_length as the value of this size.
+//
+// Required properties: (Copy, Add, Sub, Eq, Ord)
+type MessageSize = usize;
+
+#[derive(Debug, Clone)]
+pub(crate) struct DecoderCacheConfig {
+    // Number of entries to keep in recently decoded cache.
+    pub recently_decoded_cache_size: usize,
+
+    pub broadcast_tier: SoftQuotaCacheConfig,
+    pub validator_tier: SoftQuotaCacheConfig,
+    pub p2p_tier: SoftQuotaCacheConfig,
+}
+
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct SoftQuotaCacheConfig {
+    // Number of entries to keep.
+    pub total_slots: usize,
+
+    // The reserved cache slots for each author each tier
+    pub min_slots_per_author: usize,
+
+    // The cap on the total size of all pending messages per author
+    pub max_total_size_per_author: MessageSize,
+
+    // Set the following two fields to None to use the FixedSlot policy.
+    // The minimal reserved cache slots for each validator
+    pub min_slots_per_validator: Option<usize>,
+
+    // The cap on the total size of all pending messages per validator
+    pub max_total_size_per_validator: Option<MessageSize>,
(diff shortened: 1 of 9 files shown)


monad-raptorcast/src/udp.rs
```diff
@@ -89,23 +69,10 @@ pub(crate) struct UdpState<ST: CertificateSignatureRecoverable> {
     // generate a bunch of linearly dependent chunks and cause unbounded memory usage.
     // TODO strong bound on max amount of memory used per decoder?
     // TODO make eviction more sophisticated than LRU - should look at unix_ts_ms as well
-    pending_message_cache: LruCache<MessageCacheKey<CertificateSignaturePubKey<ST>>, DecoderState>,
+    decoder_cache: DecoderCache<CertificateSignaturePubKey<ST>>,
```
xinyuan-dev, line 94: addressing the comments:

> add a cap on max number of chunks that will be forwarded per message? so that a DOS can't be induced by spamming broadcast chunks to any given node

the `seen_esis` bitmap should ensure only broadcasting once per symbol.

> we also need to cap the max number chunks that are decoded - because an adversary could generate a bunch of linearly dependent chunks and cause unbounded memory usage.

the number of chunks is checked in [Decoder::with_capacity](https://github.com/category-labs/monad-bft/blob/f152a78bc5693b44eb775cd8b7c74c3db40246a8/monad-raptor/src/r10/nonsystematic/decoder/init.rs#L18-L27) ([argument values](https://github.com/category-labs/monad-bft/blob/f152a78bc5693b44eb775cd8b7c74c3db40246a8/monad-raptorcast/src/udp.rs#L270-L277))

> strong bound on max amount of memory used per decoder?

the max memory per decoder is bounded by the max number of chunks and max chunk size. is there anything else?

> make eviction more sophisticated than LRU - should look at unix_ts_ms as well

unix_ts_ms should be unnecessary to as a mean mitigate potential DOS with the changes in this PR. now each LRU has a fixed size and is scoped under (author, epoch), an author cannot spam messages to affect decoding messages from other authors.

monad-raptorcast/src/udp.rs
```diff
@@ -245,210 +201,43 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
...
-            // can we assert!(!decoder_state.decoder.decoding_done()) ?
-
-            decoder_state
-                .decoder
-                .received_encoded_symbol(&parsed_message.chunk, encoding_symbol_id);
 
-            *decoder_state
-                .recipient_chunks
-                .entry(parsed_message.recipient_hash)
-                .or_insert(0) += 1;
```
xinyuan-dev, line 362: is there any particular reason that rebroadcast happens after setting the bitmap but before feeding the symbol to the decoder?

monad-raptorcast/src/udp.rs
```diff
@@ -89,23 +69,10 @@ pub(crate) struct UdpState<ST: CertificateSignatureRecoverable> {
...
-
-#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
-struct MessageCacheKey<PT>
-where
-    PT: PubKey,
-{
-    unix_ts_ms: u64,
-    author: NodeId<PT>,
-    app_message_hash: AppMessageHash,
-    app_message_len: usize,
```
xinyuan-dev, line 128: what's the purpose of the `app_message_len` field? it doesn't seem used. also, the length is implied the hash so it doesn't see to add any additional information.

monad-raptorcast/src/udp.rs
bamrith: i don't think the order makes a difference here, as long as we check the validity of the symbol before this. the batcher guard won't rebroadcast the symbol until it goes out of scope anyway.

monad-raptorcast/src/udp.rs
xinyuan-dev: @bamrith explained that the field is there against attacks where the last symbols's app_message_len field may be used to truncate the app_message to unexpected value if unchecked.

omegablitz:
Looks good to me conceptually. Discussed offline:
1. add a “total_pending_size” (or other better name) that represents the sum of the AppMessageLen of the pending messages
1. bound that per sender, and increase the total number of slots + the number of slots that each author gets
1. this allows us to evict less aggressively while still bounding memory usage to something reasonable

monad-raptorcast/src/udp/decoding.rs
```diff
@@ -0,0 +1,1845 @@
...
+            return;
+        }
+
+        // at this point, all authors are within their quota, and no
+        // keys are considered expired. but we need to make space for
+        // the new entry. this may happen when there are many authors
+        // with min_slots config. now we randomly evict a cache entry
+        // to compensate for the new entry.
+        let (key, author, _decoder_state) = self.cache_store.get_random().expect("cache not empty");
+        self.author_index.remove(author, &key);
```
omegablitz, line 545: It might be worth logging stats on the evicted entry here (and perhaps any other interesting cases). Example on master:
```rust
            let (key, decoder_state) = self.pending_message_cache.pop_lru().expect("nonempty");
            tracing::debug!(
                num_source_symbols = decoder_state.decoder.num_source_symbols(),
                num_encoded_symbols_received = decoder_state.decoder.num_encoded_symbols_received(),
                inactivation_symbol_threshold =
                    decoder_state.decoder.inactivation_symbol_threshold(),
                recipient_chunks =? decoder_state.recipient_chunks,
                ?key,
                "dropped unfinished ManagedDecoder"
            )
```

monad-raptorcast/src/udp/decoding.rs
```diff
@@ -0,0 +1,1845 @@
...
+        let Some(stake) = validator_set.get(&message.author) else {
+            return Quota {
+                max_slots: self.non_validator_slots.min(cache_size),
+                max_size: self.non_validator_max_size,
+            };
+        };
+
+        // quota = proportional to stake
+        let total_stake: Stake = validator_set.values().copied().sum();
+        let stake_fraction = stake.checked_div(total_stake).unwrap_or(0.0);
```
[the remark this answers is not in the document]
xinyuan-dev: `total_stake == 0` implies all validators have `stake == 0` or there is no validator. In either case it's fair to assign the validator with 0% quota, which gets bumped to a non-zero `validator_min_slots`.

monad-raptorcast/src/udp/decoding.rs
```diff
@@ -0,0 +1,1845 @@
...
+
+        let mut decoder_state = DecoderState {
+            decoder,
+            recipient_chunks: BTreeMap::new(),
+            encoded_symbol_capacity,
+            app_message_len,
+            seen_esis: bitvec![usize, Lsb0; 0; encoded_symbol_capacity],
+        };
+
+        decoder_state.validate_symbol(message)?;
```
omegablitz: nit: would it be clearer to directly call `decoder_state.handle_message(message)?;` and then return decoder_state?

monad-raptorcast/src/udp/decoding.rs
```diff
@@ -0,0 +1,1845 @@
...
+            config.non_validator_max_total_size,
+        );
+        let fixed_quota = FixedQuota::new(
+            config.non_validator_slots,
+            config.non_validator_max_total_size,
+        );
+        let prune_config = PruneConfig {
+            // TODO: sync with config.udp_message_max_age_ms
+            max_unix_ts_ms_delta: Some(10 * 1000), // 10 seconds
+            max_epoch_delta: Some(2),              // 2 epochs
```
omegablitz, line 333: One footgun here is that epoch isn't seeded on startup (https://github.com/category-labs/monad-bft/blob/6aaccc2ec51eac66a2cf18ec64ab0df2c1448990/monad-raptorcast/src/lib.rs#L159)

We can seed that from the forkpoint.toml on startup though, after which this should work fine

monad-types/src/lib.rs
```diff
@@ -530,6 +536,13 @@ where
 impl Stake {
     pub const ZERO: Stake = Stake(U256::ZERO);
     pub const ONE: Stake = Stake(U256::ONE);
+
+    pub fn checked_div(self, divisor: Stake) -> Option<f64> {
+        if divisor.0.is_zero() {
+            return None;
+        }
+        Some(self / divisor)
```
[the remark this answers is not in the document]
xinyuan-dev: let me address this issue in a separate branch.

monad-raptorcast/src/udp/decoding.rs
xinyuan-dev: fixed in https://github.com/category-labs/monad-bft/pull/2369.

monad-raptorcast/src/decoding.rs
```diff
@@ -0,0 +1,2159 @@
...
+        let expired_keys = author_index.prune_expired(unix_ts_threshold, epoch_threshold);
+        evicted_keys.extend(expired_keys);
+
+        // if still over quota, compact the cache to fit under quota
+        if author_index.is_overquota() {
+            evicted_keys.extend(author_index.compact());
+        }
+
+        let (evicted_keys, reclaimed_size) = evicted_keys.decompose();
+        self.used_size -= reclaimed_size;
```
omegablitz: super nit: is this the same as reap_pruned?

xinyuan-dev: fixed.

omegablitz (approved):

PR #2093, opened 2025-07-31
Send vote(r) to leader(r) in addition to leader(r+1)

omegablitz:
Sending votes for r to r's leader in addition to r+1's leader, and having leader r broadcast QC(r) saves a lot of time - at least a full timeout period in the happy path.

This is because round r+1 is entered much faster, starting the r+1 timer sooner.

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -26,8 +26,8 @@ use monad_consensus::{
     messages::{
         consensus_message::{ConsensusMessage, ProtocolMessage},
         message::{
-            NoEndorsementMessage, ProposalMessage, RoundRecoveryMessage, TimeoutMessage,
-            VoteMessage,
+            AdvanceRoundMessage, NoEndorsementMessage, ProposalMessage, RoundRecoveryMessage,
+            TimeoutMessage, VoteMessage,
         },
     },
     no_endorsement_state::NoEndorsementState,
@@ -715,11 +715,12 @@ where
         vote_msg: VoteMessage<SCT>,
     ) -> Vec<ConsensusCommand<ST, SCT, EPT, BPT, SBT>> {
         debug!(?author, ?vote_msg, "vote message");
-        if vote_msg.vote.round < self.consensus.pacemaker.get_current_round() {
+        let vote_round = vote_msg.vote.round;
+        if vote_round < self.consensus.pacemaker.get_current_round() {
             self.metrics.consensus_events.old_vote_received += 1;
             return Default::default();
         }
-        if vote_msg.vote.round > self.consensus.pacemaker.get_current_round() + FUTURE_VOTE_BOUND {
+        if vote_round > self.consensus.pacemaker.get_current_round() + FUTURE_VOTE_BOUND {
             self.metrics.consensus_events.future_vote_received += 1;
             return Default::default();
         }
@@ -729,7 +730,7 @@ where
 
         let epoch = self
             .epoch_manager
-            .get_epoch(vote_msg.vote.round)
+            .get_epoch(vote_round)
             .expect("epoch verified");
         let validator_set = self
             .val_epoch_map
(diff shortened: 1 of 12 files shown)


ariqchowdhury (dismissed):

monad-consensus/src/messages/consensus_message.rs
```diff
@@ -183,6 +196,7 @@ where
             ProtocolMessage::Timeout(t) => t.0.tminfo.round,
             ProtocolMessage::RoundRecovery(r) => r.round,
             ProtocolMessage::NoEndorsement(n) => n.msg.round,
+            ProtocolMessage::AdvanceRound(n) => n.last_round_certificate.round() + Round(1),
```
michael-yxchen: `ConsensusMessage::verify` calls `get_round`, where we're checking if author of `AdvanceRound` msg belongs to validator set. we should return `n.last_round_certificate.round()`

monad-consensus/src/messages/message.rs
```diff
@@ -150,6 +150,19 @@ where
     pub last_round_tc: Option<TimeoutCertificate<ST, SCT, EPT>>,
 }
 
+/// This message is broadcasted upon locally constructing QC(r)
+/// This helps other nodes advance their round faster
+#[derive(Clone, Debug, PartialEq, Eq, RlpEncodable, RlpDecodable)]
+#[rlp(trailing)]
```
michael-yxchen: nit: is this needed?

monad-consensus-state/src/lib.rs
```diff
@@ -748,6 +749,25 @@ where
...
+
+            let vote_round_leader = self
+                .election
+                .get_leader(vote_round, validator_set.get_members());
+            if self.nodeid == &vote_round_leader {
+                // Note that if we're also the leader of (vote_round + 1), we'll send QC(r) in
+                // both AdvanceRound(QC(r)) and Proposal(r+1). This is fine, as AdvanceRound(r)
+                // will be propagated faster than Proposal(r+1) due to it being direct broadcast.
+                cmds.push(ConsensusCommand::Publish {
+                    target: RouterTarget::Broadcast(self.consensus.get_current_epoch()),
```
michael-yxchen, line 768: maybe it's also worth noting that we're processing Qc first. so this is broadcasting to `qc.round + 1` validator set

monad-mock-swarm/tests/many_nodes.rs
```diff
@@ -109,3 +109,82 @@ fn many_nodes_noser() {
...
+    let mut max_observed_local_timeouts = swarm
+        .states()
+        .values()
+        .map(|node| node.state.metrics().consensus_events.local_timeout)
+        .max()
+        .unwrap();
+    // subtract 1 for the initial timeout on startup
+    max_observed_local_timeouts -= 1;
+
+    let max_allowed_local_timeouts = num_blocks as u64 / num_online_nodes as u64;
```
michael-yxchen: just for my understanding, why is this `num_blocks / num_online_nodes`? is it because we only have 1 offline node?

i would simulate for X rounds, and make max_allowed `X * offline / total`. this is probably equivalent so is fine to keep

monad-consensus/src/messages/message.rs
omegablitz: nope, good catch

monad-mock-swarm/tests/many_nodes.rs
omegablitz: Your way is way more intuitive lol, changed to yours

monad-consensus/src/messages/consensus_message.rs
omegablitz: Good catch, I looked up usages of `get_round` before and incorrectly only saw the monad-debugger usage

monad-debugger/src/lib.rs
```diff
@@ -101,11 +101,11 @@ pub fn simulation_make() -> *mut Simulation {
                         ),
                         vec![
                             GenericTransformer::Latency(LatencyTransformer::new(
-                                Duration::from_millis(10),
+                                Duration::from_millis(25),
```
omegablitz, line 104: This is done to be more realistic in the frontend.
With the previous settings, it would blocksync occasionally because AdvanceRound(r) would get received before Proposal(r-1)

omegablitz:
https://github.com/user-attachments/assets/7a6b2419-85ad-418b-abba-299258efe924

ariqchowdhury (approved):

PR #2096, opened 2025-07-31
Clean up monad-node-config crypto feature

omegablitz:


omegablitz:
Folded into [47492f5](https://github.com/category-labs/monad-bft/commit/47492f5403aa48d2b7d9e154bc9db078b16bcb77)

PR #2100, opened 2025-08-01
Log leader on local timeout

omegablitz:


omegablitz:
Closing in favor of https://github.com/category-labs/monad-bft/pull/1546

PR #2103, opened 2025-08-04
Fast-fail if NE obviously fails signature checks

omegablitz:


michael-yxchen (approved):

omegablitz:
Add fast-fail tests for votes/timeouts/NEs

omegablitz:
We already had fail-fast tests for votes and timeouts

Issue #2085, opened 2025-07-30
consensus / txpool: preload manager gets triggered on every block


robertkiel:
The transaction pool comes with a preload manager to prefetch data from the database whenever it is expected that the node is going to be the leader in the next round or the one after.

To do so, the consensus event get annotated with a `Vec` that (is supposed) to [contain](https://github.com/category-labs/monad-bft/blob/80fa09910f374d2854145a6c913bc59462b5a4ee/monad-state/src/consensus.rs#L544) all round numbers for which the node is going to be the leader. However, [`compute_upcoming_leader_round_pairs`](https://github.com/category-labs/monad-bft/blob/1ac9472f1b586e758bd37bde7de613190441615f/monad-state/src/consensus.rs#L496) either outputs the next three round numbers and their leaders (when `SKIP_SELF` is set to false) or filters out rounds in which the nodes will be a leader (when `SKIP_SELF` is set to true). 

The outcome of this is that addresses will always get "preloaded", no matter whether the node is a leader or not, see code of [`PreloadManager`](https://github.com/category-labs/monad-bft/blob/1ac9472f1b586e758bd37bde7de613190441615f/monad-eth-txpool-executor/src/preload.rs#L54).

PR #2108, opened 2025-08-04
Fix consensus state upcoming leaders filter

andr-dev:
The upcoming leaders filter is used to tell the txpool in which future rounds the node will be proposing so it can preload account balances. Previously, it would produce all rounds regardless of the leader which causes the txpool preload manager to load account balances in every round. This change fixes the filtering condition so the txpool is only given the node's rounds.

Resolves https://github.com/category-labs/monad-bft/issues/2085.

### DIFF
--- a/monad-state/src/consensus.rs
+++ b/monad-state/src/consensus.rs
@@ -41,12 +41,10 @@ use monad_executor_glue::{
     TimerCommand, TimestampCommand, TxPoolCommand,
 };
 use monad_state_backend::StateBackend;
-use monad_types::{ExecutionProtocol, NodeId, Round, RouterTarget, SeqNum};
+use monad_types::{ExecutionProtocol, NodeId, Round, RouterTarget};
 use monad_validator::{
-    epoch_manager::EpochManager,
-    leader_election::LeaderElection,
-    validator_set::{ValidatorSetType, ValidatorSetTypeFactory},
-    validators_epoch_mapping::ValidatorsEpochMapping,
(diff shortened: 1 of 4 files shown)


philipglazman (dismissed):

monad-state/src/consensus.rs
```diff
@@ -536,16 +536,17 @@ where
                     .get_leader(round, next_validator_set.get_members());
 
                 if SKIP_SELF {
-                    (&leader != self.nodeid).then_some((leader, round))
+                    &leader != self.nodeid
```
omegablitz: I don't think SKIP_SELF describes what this function is doing, unless I'm misunderstanding something.

I think it would be clearer if SKIP_SELF was removed as a parameter from this function, and any desired filtering was done outside this function.

omegablitz (commented):
Could you add the motivation for this change in the PR description or commit message for doc purposes? It's not immediately clear when reading the diff.

monad-state/src/consensus.rs
andr-dev: The `SKIP_SELF` parameter is used when getting leaders for forwarding transactions, I renamed it to `FILTER_ONLY_SELF` to make it clearer.

ariqchowdhury:
is there a test for checking upcoming leaders?

andr-dev:
> is there a test for checking upcoming leaders?

Changed approach and did some minor refactoring to make the individual methods unit testable.

PR #2114, opened 2025-08-04
feat: check validator stake>0

### LINKED ISSUE #2077: consensus: validators process messages from validators with 0 stake
Although, due to other countermeasures it appears to be unlikely that the validator set contains nodes with 0 stake, validators with 0 stake can send messages to other validators. These messages will get processed due to this [check](https://github.com/category-labs/monad-bft/blob/868dacd73272bb013eb85e6e2672508caacd7130/monad-consensus/src/validation/signing.rs#L1002-L1015) and could contain malicious data that can trigger some `assert`s - without the risk of reputation loss.

```rust
// taken from https://github.com/category-labs/monad-bft/blob/868dacd73272bb013eb85e6e2672508caacd7130/monad-consensus/src/validation/signing.rs#L1002-L1015
impl<PT: PubKey> ValidatorPubKey for PT {
    type NodeIdPubKey = PT;
    /// Validate that pubkey is in the validator set
    fn valid_pubkey(
        self,
        validators: &BTreeMap<NodeId<Self::NodeIdPubKey>, Stake>,
    ) -> Result<Self, Error> {
        // why don't add sth. like `&& Stake(...) > 0?
        if validators.contains_key(&NodeId::new(self)) {
            Ok(self)
        } else {
            Err(Error::InvalidAuthor)
        }
    }
}

varun-doshi:
Fixes #2077

### DIFF
--- a/monad-consensus/src/validation/signing.rs
+++ b/monad-consensus/src/validation/signing.rs
@@ -1025,10 +1025,9 @@ impl<PT: PubKey> ValidatorPubKey for PT {
         self,
         validators: &BTreeMap<NodeId<Self::NodeIdPubKey>, Stake>,
     ) -> Result<Self, Error> {
-        if validators.contains_key(&NodeId::new(self)) {
-            Ok(self)
-        } else {
-            Err(Error::InvalidAuthor)
+        match validators.get(&NodeId::new(self)) {
+            Some(stake) if stake.0 > 0 => Ok(self),
+            _ => Err(Error::InsufficientStake),
         }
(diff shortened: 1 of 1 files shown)


Santanavqt:
great

robertkiel:
Any update on this? The fix is simple and effective

bamrith:
I think validator set shouldn't have any zero staked validators when it reaches pubkey validation. Closing in favor of https://github.com/category-labs/monad-bft/pull/2545

PR #2121, opened 2025-08-05
Refactor: Improve Error Handling by Replacing panic! and expect with Result

WONJUNE-LEE:
Currently, the monad-consensus-state crate utilizes panic! and expect() to handle certain failure scenarios within the consensus logic. This approach causes the entire node to crash immediately upon encountering an unrecoverable state or malformed input.

This PR aims to refactor this error handling strategy to be more robust and resilient. By replacing panics with a dedicated Result type, we shift control of how to handle critical errors from the library to the consuming application (the top-level node binary). This prevents a library-level failure from bringing down the entire process, making the node more stable and secure against potential DoS vectors from malicious peers.

### DIFF
--- a/monad-consensus-state/src/error.rs
+++ b/monad-consensus-state/src/error.rs
@@ -1,0 +1,24 @@
+use monad_types::{Epoch, Round, SeqNum};
+use thiserror::Error;
+
+/// Defines errors that can occur within the consensus state machine.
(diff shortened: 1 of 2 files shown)


ariqchowdhury:
these failures are expected and intended to stop the node so that it doesn't try and proceed in consensus with corrupted or invalid state.

PR #2125, opened 2025-08-05
Fix for Publisher large round gap

fabiotdf:
Before fix, publishers (validators) would spend many rounds sending invites to other validators thinking they are full-nodes who want group invites.
This fixes it by adding a `get_known_addrs_fullnodes()` to Peer discovery so that the publisher only gets full-node ids and no validator

### DIFF
--- a/monad-peer-discovery/src/discovery.rs
+++ b/monad-peer-discovery/src/discovery.rs
@@ -1122,6 +1122,19 @@ where
             .collect()
     }
 
+    fn get_fullnode_addrs(&self) -> HashMap<NodeId<CertificateSignaturePubKey<ST>>, SocketAddrV4> {
+        let empty = BTreeSet::new();
+        let curr_validators = self
+            .epoch_validators
+            .get(&self.current_epoch)
+            .unwrap_or(&empty);
+        self.routing_info
+            .iter()
+            .filter(|(id, _)| !curr_validators.contains(id))
+            .map(|(id, name_record)| (*id, name_record.address()))
+            .collect()
+    }
+
     fn get_name_records(
         &self,
     ) -> HashMap<NodeId<CertificateSignaturePubKey<ST>>, MonadNameRecord<ST>> {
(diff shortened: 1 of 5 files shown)


eerkaijun (dismissed):

monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -338,7 +338,7 @@ where
                                 self.peer_discovery_driver
                                     .lock()
                                     .unwrap()
-                                    .get_known_addresses()
+                                    .get_fullnode_addrs()
                             };
                             let nodes: Vec<_> = known_addresses.keys().copied().collect();
```
michael-yxchen: nit: can we combine `let known_addresses = ` and `let nodes =`, and rename it to `full_nodes`?

monad-peer-discovery/src/discovery.rs
```diff
@@ -1122,6 +1122,19 @@ where
             .collect()
     }
 
+    fn get_fullnode_addrs(&self) -> HashMap<NodeId<CertificateSignaturePubKey<ST>>, SocketAddrV4> {
+        let empty = BTreeSet::new();
+        let curr_validators = self
+            .epoch_validators
+            .get(&self.current_epoch)
+            .unwrap_or(&empty);
```
michael-yxchen, line 1130: nit: `unwrap_or_default`

michael-yxchen (dismissed):

monad-peer-discovery/src/discovery.rs
fabiotdf: I had tried that but got 
```
" the trait `Default` is not implemented for `&BTreeSet<NodeId<<... as CertificateKeyPair>::PubKeyType>>"
```

monad-raptorcast/src/raptorcast_secondary/mod.rs
fabiotdf: Done

omegablitz (approved):

PR #2128, opened 2025-08-06
Add support for system transactions

bamrith:
We need a special type of transaction for consensus to be able to communicate with execution about network related changes (epoch change for example). This PR adds the infrastructure on the consensus side to be able to generate such transactions and validate them differently than regular user transactions.

I've added unit tests for checking the validity of the system transactions based on transaction parameters. State specific tests will be added as system calls are added.

### DIFF
--- a/monad-eth-block-policy/src/lib.rs
+++ b/monad-eth-block-policy/src/lib.rs
@@ -32,6 +32,7 @@ use monad_crypto::certificate_signature::{
 use monad_eth_txpool_types::TransactionError;
 use monad_eth_types::{Balance, EthAccount, EthExecutionProtocol, EthHeader, Nonce};
 use monad_state_backend::{StateBackend, StateBackendError};
+use monad_system_calls::SystemTransaction;
 use monad_types::{BlockId, Round, SeqNum, GENESIS_BLOCK_ID, GENESIS_SEQ_NUM};
 use sorted_vector_map::SortedVectorMap;
 use tracing::{trace, warn};
@@ -154,6 +155,7 @@ where
     SCT: SignatureCollection<NodeIdPubKey = CertificateSignaturePubKey<ST>>,
 {
     pub block: ConsensusFullBlock<ST, SCT, EthExecutionProtocol>,
+    pub system_txns: Vec<SystemTransaction>,
     pub validated_txns: Vec<Recovered<TxEnvelope>>,
     pub nonces: BTreeMap<Address, Nonce>,
     pub txn_fees: BTreeMap<Address, Balance>,
@@ -707,11 +709,13 @@ where
             return Err(BlockPolicyError::ExecutionResultMismatch);
         }
 
+        let system_tx_signers = block.system_txns.iter().map(|txn| txn.signer());
         // TODO fix this unnecessary copy into a new vec to generate an owned Address
         let tx_signers = block
             .validated_txns
             .iter()
             .map(|txn| txn.signer())
+            .chain(system_tx_signers)
             .collect_vec();
         // these must be updated as we go through txs in the block
         let mut account_nonces = self.get_account_base_nonces(
@@ -727,6 +731,25 @@ where
             Some(&extending_blocks),
             tx_signers.iter(),
         )?;
+
+        for sys_txn in block.system_txns.iter() {
+            let sys_txn_signer = sys_txn.signer();
+            let sys_txn_nonce = sys_txn.nonce();
+
+            let expected_nonce = account_nonces
+                .get_mut(&sys_txn_signer)
+                .expect("account_nonces should have been populated");
+
+            if &sys_txn_nonce != expected_nonce {
+                warn!(
+                    seq_num =? block.header().seq_num,
+                    round =? block.header().block_round,
+                    "block not coherent, invalid nonce for system transaction"
+                );
+                return Err(BlockPolicyError::BlockNotCoherent);
+            }
+            *expected_nonce += 1;
+        }
 
         for txn in block.validated_txns.iter() {
             let eth_address = txn.signer();
--- a/monad-eth-block-validator/src/lib.rs
+++ b/monad-eth-block-validator/src/lib.rs
@@ -13,7 +13,10 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-use std::{collections::BTreeMap, marker::PhantomData};
+use std::{
+    collections::{BTreeMap, VecDeque},
+    marker::PhantomData,
+};
 
 use alloy_consensus::{
     constants::EMPTY_WITHDRAWALS,
@@ -40,11 +43,13 @@ use monad_eth_types::{
 };
 use monad_secp::RecoverableAddress;
 use monad_state_backend::StateBackend;
+use monad_system_calls::{validator::SystemTransactionValidator, SystemTransaction};
 use rayon::iter::{IntoParallelIterator, ParallelIterator};
 use tracing::{debug, trace_span, warn};
 
 type NonceMap = BTreeMap<Address, Nonce>;
 type TxnFeeMap = BTreeMap<Address, U256>;
+type SystemTransactions = Vec<SystemTransaction>;
 type ValidatedTxns = Vec<Recovered<TxEnvelope>>;
 
 /// Validates transactions as valid Ethereum transactions and also validates that
@@ -77,12 +82,14 @@ where
 
     fn validate_block_body(
         &self,
+        header: &ConsensusBlockHeader<ST, SCT, EthExecutionProtocol>,
         body: &ConsensusBlockBody<EthExecutionProtocol>,
         tx_limit: usize,
         proposal_gas_limit: u64,
         proposal_byte_limit: u64,
         max_code_size: usize,
-    ) -> Result<(ValidatedTxns, NonceMap, TxnFeeMap), BlockValidationError> {
+    ) -> Result<(SystemTransactions, ValidatedTxns, NonceMap, TxnFeeMap), BlockValidationError>
+    {
         let EthBlockBody {
             transactions,
             ommers,
@@ -104,7 +111,7 @@ where
         }
 
         // recovering the signers verifies that these are valid signatures
-        let eth_txns: Vec<Recovered<TxEnvelope>> = transactions
+        let recovered_txns: VecDeque<Recovered<TxEnvelope>> = transactions
             .into_par_iter()
             .map(|tx| {
                 let _span = trace_span!("validator: recover signer").entered();
@@ -114,8 +121,34 @@ where
             .collect::<Result<_, monad_secp::Error>>()
             .map_err(|_err| BlockValidationError::TxnError)?;
 
-        // recover the account nonces and txn fee usage in this block
+        let (system_txns, eth_txns) =
+            match SystemTransactionValidator::validate_and_extract_system_transactions(
+                header,
+                recovered_txns,
+            ) {
+                Ok((system_txns, eth_txns)) => (system_txns, eth_txns),
+                Err(err) => {
+                    debug!(?err, "system transaction validator error");
+                    return Err(BlockValidationError::SystemTxnError);
+                }
+            };
+
+        // recover the account nonces for system txns
         let mut nonces = BTreeMap::new();
+
+        // duplicate check. this is also done in SystemTransactionValidator
+        for sys_txn in &system_txns {
+            let maybe_old_nonce = nonces.insert(sys_txn.signer(), sys_txn.nonce());
+            // A block is invalid if we see a smaller or equal nonce
+            // after the first or if there is a nonce gap
+            if let Some(old_nonce) = maybe_old_nonce {
+                if sys_txn.nonce() != old_nonce + 1 {
+                    return Err(BlockValidationError::SystemTxnError);
+                }
+            }
+        }
+
+        // recover the account nonces and txn fee for user txns
         let mut txn_fees: BTreeMap<Address, U256> = BTreeMap::new();
 
         for eth_txn in &eth_txns {
@@ -152,7 +185,9 @@ where
         }
 
         let total_gas: u64 = eth_txns.iter().map(|tx| tx.gas_limit()).sum();
-        let proposal_size: usize = eth_txns.iter().map(|tx| tx.length()).sum();
+        let system_txns_size: usize = system_txns.iter().map(|tx| tx.length()).sum();
+        let user_txns_size: usize = eth_txns.iter().map(|tx| tx.length()).sum();
+        let proposal_size = system_txns_size + user_txns_size;
         debug!(
             total_gas,
             proposal_size,
@@ -168,7 +203,7 @@ where
             return Err(BlockValidationError::TxnError);
         }
 
-        Ok((eth_txns, nonces, txn_fees))
+        Ok((system_txns, eth_txns, nonces, txn_fees))
     }
 
     fn validate_block_header(
@@ -290,7 +325,8 @@ where
     >{
         self.validate_block_header(&header, &body, author_pubkey, proposal_gas_limit)?;
 
-        if let Ok((validated_txns, nonces, txn_fees)) = self.validate_block_body(
+        if let Ok((system_txns, validated_txns, nonces, txn_fees)) = self.validate_block_body(
+            &header,
             &body,
             tx_limit,
             proposal_gas_limit,
@@ -300,6 +336,7 @@ where
             let block = ConsensusFullBlock::new(header, body)?;
             Ok(EthValidatedBlock {
                 block,
+                system_txns,
                 validated_txns,
                 nonces,
                 txn_fees,
@@ -314,11 +351,15 @@ where
 mod test {
     use alloy_consensus::Signed;
     use alloy_primitives::{PrimitiveSignature, B256};
-    use monad_consensus_types::payload::ConsensusBlockBodyInner;
-    use monad_crypto::NopSignature;
+    use monad_consensus_types::{
+        payload::{ConsensusBlockBodyId, ConsensusBlockBodyInner, RoundSignature},
+        quorum_certificate::QuorumCertificate,
+    };
+    use monad_crypto::{certificate_signature::CertificateKeyPair, NopKeyPair, NopSignature};
     use monad_eth_testutil::make_legacy_tx;
     use monad_state_backend::InMemoryState;
     use monad_testutil::signing::MockSignatures;
+    use monad_types::{Epoch, NodeId, Round, SeqNum, GENESIS_SEQ_NUM};
 
     use super::*;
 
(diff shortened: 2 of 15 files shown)


monad-eth-block-policy/src/lib.rs
```diff
@@ -707,6 +708,9 @@ where
             return Err(BlockPolicyError::ExecutionResultMismatch);
         }
 
+        // TODO check system transation input data here
+        assert!(block.system_txns.is_empty());
```
ariqchowdhury: can we create a `SystemTransactionValidator` and put the validation rules in that struct `fn validate(self, txns: Vec<_>)`

monad-eth-block-validator/src/lib.rs
```diff
@@ -75,14 +79,25 @@ where
...
+        mut transactions: VecDeque<Recovered<TxEnvelope>>,
+    ) -> (Vec<Recovered<TxEnvelope>>, Vec<Recovered<TxEnvelope>>) {
+        let mut system_txns = Vec::new();
+        while transactions.front().is_some_and(is_system_transaction) {
+            system_txns.push(transactions.pop_front().unwrap());
+        }
+
+        (system_txns, transactions.into())
+    }
+
```
ariqchowdhury: where will we be checking that system transactions are always at the head of the block? Is that still a requirement?

monad-eth-block-policy/src/lib.rs
ariqchowdhury: also, is this supposed to be in validate_block_body? Or what part of coherency checking do system txns specifically need to satisfy?

monad-system-calls/src/lib.rs
```diff
@@ -0,0 +1,229 @@
...
+
+pub fn generate_system_transactions(sys_calls: Vec<SystemCall>) -> Vec<Recovered<TxEnvelope>> {
+    sys_calls
+        .into_iter()
+        .map(|sys_call| sys_call.into_txn())
+        .collect()
+}
+
+pub fn is_system_transaction(txn: &Recovered<TxEnvelope>) -> bool {
+    txn.signer() == SYSTEM_SENDER_ETH_ADDRESS
```
ariqchowdhury: this isn't a strong enough check is it?

monad-system-calls/src/lib.rs
```diff
@@ -0,0 +1,229 @@
...
+));
+const SYSTEM_SENDER_ETH_ADDRESS: Address =
+    Address::new(hex!("0x6f49a8F621353f12378d0046E7d7e4b9B249DC9e"));
+
+pub enum SystemCall {}
+
+impl SystemCall {
+    pub fn into_txn(self) -> Recovered<TxEnvelope> {
+        // Placeholder transaction
+        let transaction = TxLegacy {
```
ariqchowdhury: what is this placeholder for?

monad-system-calls/src/lib.rs
```diff
@@ -0,0 +1,229 @@
...
+) -> Result<(), SystemTransactionError> {
+    assert!(is_system_transaction(txn));
+
+    if !txn.tx().is_legacy() {
+        return Err(SystemTransactionError::InvalidTxType);
+    }
+
+    if txn.tx().nonce() != 0 {
+        return Err(SystemTransactionError::NonZeroNonce);
+    }
```
ariqchowdhury: isn't the nonce for sys txns expected to be correct?

monad-eth-block-policy/src/lib.rs
bamrith: will add `SystemTransactionValidator`. we don't need any coherency checks for the system transactions for the staking contract but I wasn't sure if we would require any for new ones that will be added later

monad-system-calls/src/lib.rs
bamrith: once the staking contract is added, this should also check if the destination is the staking contract address. i'm thinking we should also check to see if the input data contains one of the 3 system call function selectors.

monad-system-calls/src/lib.rs
bamrith: just to return any `Recovered<TxEnvelope>` right now since the `SystemCalls` enum is empty. each system call will have a different transaction generated

monad-system-calls/src/lib.rs
bamrith: since execution switched to the new model, we can treat these transactions differently and ignore nonce/gas checks since they won't be running any eth validations.

monad-eth-block-validator/src/lib.rs
bamrith: that is still the case. this function just splits the transaction list into two groups assuming that the system transactions come first. there's a check to make sure that there are no system transactions in the second group (user transactions). once we start validating specific system transactions, we can make sure that the first group has the expected set of transactions exactly.

monad-eth-block-policy/src/lib.rs
ariqchowdhury: I don't think there are any coherency checks that should be necessary for system txns. There's no nonce dependency to check or account balance requirement so there shouldn't be any dependency on previous blocks in the chain. The SystemTransactionValidator should just be part of the block validation in block policy.

monad-eth-block-validator/src/lib.rs
ariqchowdhury: I'm not understanding where the validation happens that sys txns come first. The code above splits into 2 groups by iterating over the list and extracting sys txns so there's obviously never going to be sys txns in the second group.

monad-system-calls/src/lib.rs
ariqchowdhury: until that point, I'm pretty sure this should still check destination and find that there are no matching syscall addresses and then return false. Because if this merges as is, its going to lead to weird behaviour if someone sends to that system address.

monad-system-calls/src/lib.rs
ariqchowdhury: I think system transaction should have their own Type instead of `Recovered<TxEnvelope>`. 
Something like
```
enum SysTxn {
Foobar(Recovered<TxEnvelope>)
}
```
and into_txn should transform it into the correct variant of SysTxn or return an Error if its not valid.

monad-system-calls/src/lib.rs
ariqchowdhury: can you update the spec with this new requirement.

monad-eth-block-validator/src/lib.rs
```diff
@@ -114,11 +129,27 @@ where
             .collect::<Result<_, monad_secp::Error>>()
             .map_err(|_err| BlockValidationError::TxnError)?;
 
+        let (system_txns, eth_txns) = Self::extract_system_transactions(recovered_txns);
```
ariqchowdhury: Related to the comment about having a SysTxn enum, I think the return value of `extract_system_transactions` should be something like `(Result<SysTxn>, Recovered<_>)` so that we extract and validated the sys txns and then have a type going forward that guarantees they are valid sys txns. 

Then we don't have to check later that they valid sys txns, the type guarantees it.

monad-eth-block-validator/src/lib.rs
```diff
@@ -114,11 +129,27 @@ where
...
+
+        for system_txn in &system_txns {
+            match is_valid_system_transaction(system_txn) {
+                Ok(()) => {}
+                Err(system_txn_error) => {
+                    debug!(?system_txn, ?system_txn_error, "invalid system transaction");
+                    return Err(BlockValidationError::TxnError);
```
ariqchowdhury: there should be a special error type for sys txn errors

monad-eth-block-validator/src/lib.rs
```diff
@@ -114,11 +129,27 @@ where
...
+        }
+
         // recover the account nonces and txn fee usage in this block
         let mut nonces = BTreeMap::new();
         let mut txn_fees: BTreeMap<Address, U256> = BTreeMap::new();
 
         for eth_txn in &eth_txns {
+            if is_system_transaction(eth_txn) {
+                return Err(BlockValidationError::TxnError);
+            }
```
ariqchowdhury: the types I suggested should make this situation impossible and we shouldn't have to check it. Unless we are also deciding to have a check that the system address is never part of regular txns. I'm not sure if that's actually in the spec, should clarify there what the intention is.

monad-eth-txpool/src/pool/mod.rs
```diff
@@ -262,12 +265,19 @@ where
         // u64::MAX seconds is ~500 Billion years
         assert!(timestamp_seconds < u64::MAX.into());
 
+        let system_transactions = self.get_system_transactions();
+        let system_txs_gas_limit: u64 = system_transactions.iter().map(|tx| tx.gas_limit()).sum();
```
ariqchowdhury: is the gas_limit still a concern given the change to execution impl?

monad-eth-block-policy/src/lib.rs
bamrith: updated

monad-eth-block-validator/src/lib.rs
bamrith: removed. added `SystemTransactionValidator` to check for expected system transactions and only extract those

monad-system-calls/src/lib.rs
bamrith: updated spec

monad-eth-block-validator/src/lib.rs
bamrith: updated

monad-eth-txpool/src/pool/mod.rs
bamrith: not anymore, they should be 0

monad-eth-block-validator/src/lib.rs
```diff
@@ -114,11 +121,31 @@ where
...
+                    return Err(BlockValidationError::SystemTxnError);
+                }
+            };
+
         // recover the account nonces and txn fee usage in this block
         let mut nonces = BTreeMap::new();
         let mut txn_fees: BTreeMap<Address, U256> = BTreeMap::new();
 
         for eth_txn in &eth_txns {
+            if SystemTransactionValidator::is_system_sender(eth_txn.signer()) {
```
omegablitz: Can these checks be moved to `SystemTransactionValidator::validate_and_extract_system_transactions`? Can you also add a test where a transaction signed by the system signer is at the end of the block payload?

monad-eth-block-validator/src/lib.rs
```diff
@@ -327,6 +360,25 @@ mod test {
...
     const PROPOSAL_SIZE_LIMIT: u64 = 4_000_000;
 
+    fn get_header(
+        payload_id: ConsensusBlockBodyId,
+    ) -> ConsensusBlockHeader<NopSignature, MockSignatures<NopSignature>, EthExecutionProtocol>
+    {
+        let nop_keypair = NopKeyPair::from_bytes(&mut [0_u8; 32]).unwrap();
+        ConsensusBlockHeader::new(
+            NodeId::new(nop_keypair.pubkey()),
+            Epoch(2), // wrong epoch: should be 1
```
omegablitz: nit: switch to Epoch(1), remove comment

monad-system-calls/src/validator.rs
```diff
@@ -0,0 +1,290 @@
...
+                    validated_sys_txns.push(validated_sys_txn);
+                }
+                Err(err) => {
+                    return Err(SystemTransactionValidationError::SystemTransactionError(
+                        err,
+                    ));
+                }
+            }
+        }
+
```
omegablitz, line 163: As mentioned earlier, verify that the rest of the transactions in `txns` are valid, and add a test for this

monad-eth-txpool/src/pool/mod.rs
```diff
@@ -262,12 +265,19 @@ where
...
+            .map(|tx| tx.length() as u64)
+            .sum();
+
         let transactions = self.tracked.create_proposal(
             event_tracker,
             proposed_seq_num,
-            tx_limit,
-            proposal_gas_limit,
-            proposal_byte_limit,
+            tx_limit - system_transactions.len(),
```
omegablitz, line 276: These checks aren't consistent with the block validator. The transaction limit is, but not the gas limit or byte limit. This txpool implementation is over conservative so nothing will break, but I still don't think we should have these diverge.

```rust
// early return if number of transactions exceed limit
// no need to individually validate transactions
if transactions.len() > tx_limit {
}

//
// ...
// validates that among other things, the gas limit of each transaction is less than proposal gas limit
// note that this makes sure that total_gas doesn't overflow when computed later
// ...
//

let total_gas: u64 = eth_txns.iter().map(|tx| tx.gas_limit()).sum();
let proposal_size: usize = eth_txns.iter().map(|tx| tx.length()).sum();
debug!(
    total_gas,
    proposal_size,
    txs = transactions.len(),
    "proposal stats"
);
if total_gas > proposal_gas_limit {
}
if proposal_size as u64 > proposal_byte_limit {
}
```

monad-eth-txpool/src/pool/mod.rs
```diff
@@ -262,12 +265,19 @@ where
         // u64::MAX seconds is ~500 Billion years
         assert!(timestamp_seconds < u64::MAX.into());
 
+        let system_transactions = self.get_system_transactions();
```
omegablitz, line 267: For completeness, these should be prepended to the block body below.
```rust
let body = EthBlockBody {
    transactions: transactions.into_iter().map(|tx| tx.into_tx()).collect(),
    ommers: Vec::new(),
    withdrawals: Vec::new(),
};
```

(I realize that this is always empty right now, but still)

monad-system-calls/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+));
+const SYSTEM_SENDER_ETH_ADDRESS: Address =
+    Address::new(hex!("0x6f49a8F621353f12378d0046E7d7e4b9B249DC9e"));
+
+// A system call is a destination address, system call function selector
+// and function input data
+pub enum SystemCall {}
+
+impl SystemCall {
+    pub fn get_dest_address(&self) -> Address {
```
omegablitz: nit: make these private if they shouldn't be used externally

monad-system-calls/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+    ST: CertificateSignatureRecoverable,
+    SCT: SignatureCollection<NodeIdPubKey = CertificateSignaturePubKey<ST>>,
+{
+    Vec::new()
+}
+
+#[derive(Debug, Clone)]
+pub enum SystemTransaction {}
+
+// impl From<SystemCall> for SystemTransaction
```
omegablitz: nit: should be able to fill this implementation out (will just be matching on empty enum)

monad-system-calls/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+
+const SYSTEM_SENDER_PRIV_KEY: B256 = B256::new(hex!(
+    "b0358e6d701a955d9926676f227e40172763296b317ff554e49cdf2c2c35f8a7"
+));
+const SYSTEM_SENDER_ETH_ADDRESS: Address =
+    Address::new(hex!("0x6f49a8F621353f12378d0046E7d7e4b9B249DC9e"));
+
+// A system call is a destination address, system call function selector
+// and function input data
+pub enum SystemCall {}
```
omegablitz: nit: It might be safer if `SystemCall` and `SystemTransaction` were "sealed", in that they are not constructible outside this crate.

One way of doing this could be:
```rust
pub enum SystemCall(SystemCallInner)

// or pub(crate) enum
enum SystemCallInner {
    // ...
}
```

monad-system-calls/src/validator.rs
```diff
@@ -0,0 +1,290 @@
...
+        address == SYSTEM_SENDER_ETH_ADDRESS
+    }
+
+    // used to check if a user transaction calls a restricted function
+    pub fn is_restricted_system_call(txn: &Recovered<TxEnvelope>) -> bool {
+        // TODO check if txn invokes any supported system call
+        false
+    }
+
+    pub fn static_validate_system_transaction(
```
omegablitz: nit: make this private if it's not intended to be used outside

monad-system-calls/src/validator.rs
```diff
@@ -0,0 +1,290 @@
...
+        }
+
+        if txn.tx().value() != U256::ZERO {
+            return Err(SystemTransactionError::NonZeroValue);
+        }
+
+        Ok(())
+    }
+
+    pub fn validate_system_transaction_input(
```
omegablitz: nit: make this private if it's not intended on being used outside

monad-system-calls/src/validator.rs
```diff
@@ -0,0 +1,290 @@
...
+        expected_sys_call: SystemCall,
+        sys_txn: &Recovered<TxEnvelope>,
+    ) -> Result<SystemTransaction, SystemTransactionError> {
+        // verify destination address, function selector and function input
+
+        // TODO remove error
+        Err(SystemTransactionError::NonZeroValue)
+    }
+
+    pub fn validate_system_transaction(
```
omegablitz: nit: make this private if it's not intended on being used outside

monad-system-calls/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+    }
+}
+
+// Used by a round leader to generate system calls for the proposing block
+pub fn generate_system_calls() -> Vec<SystemCall> {
+    Vec::new()
+}
+
+// Used by a validator to generate expected system calls for a block
+pub fn generate_system_calls_from_header<ST, SCT>(
```
omegablitz: nit: make private if shouldn't be used externally

monad-system-calls/src/validator.rs
omegablitz, line 1: Wrote some individual comments, but the general ask I have is that the `SystemTransaction`/`SystemCall` type should only be able to "leak" out of this crate through a very restricted subset of functions.

It seems like the only two functions that should be used for constructing these types externally are `SystemTransactionValidator::validate_and_extract_system_transactions` and `generate_system_calls`.

The same should be true for the inverse; there should be a restricted set of public functions that are used for validating that a tx is not a system transaction, and this should only be necessary to call in the txpool. I think these are currently the `SystemTransactionValidator::is_system_sender` and `SystemTransactionValidator::is_restricted_system_call` functions (can these be consolidated and rolled into `static_validate_transaction` perhaps?).

It would also be good to document the external functions that should be called and where at the top of one or both of these files.

monad-eth-txpool/src/pool/mod.rs
bamrith: avoided the gas limit check since they have to be 0. added the byte limit check to block validator

monad-system-calls/src/validator.rs
bamrith: I avoided rolling those two functions into `static_validate_transaction` because on top of users not being able sending valid system transactions, they should also not send any transactions from the system sender and not invoke any system calls from other addresses too. We can verify them separately (in TxPool as you mentioned)

monad-system-calls/src/validator.rs
```diff
@@ -0,0 +1,320 @@
...
+    {
+        let mut validated_sys_txns = Vec::new();
+
+        let expected_sys_calls = generate_system_calls_from_header(block_header);
+        for expected_sys_call in expected_sys_calls {
+            let Some(sys_txn) = txns.pop_front() else {
+                return Err(SystemTransactionValidationError::MissingSystemTransaction);
+            };
+
+            match Self::validate_system_transaction(expected_sys_call, &sys_txn) {
```
omegablitz, line 143: nit: do nonce checks here for clarity/safety (even though they already exist in monad-eth-block-policy)
(also add associated test)

omegablitz (approved):

omegablitz (approved):

monad-eth-block-validator/src/lib.rs
```diff
@@ -114,8 +121,33 @@ where
...
+        // recover the account nonces for system txns
         let mut nonces = BTreeMap::new();
+
+        for sys_txn in &system_txns {
+            let maybe_old_nonce = nonces.insert(sys_txn.signer(), sys_txn.nonce());
+            // A block is invalid if we see a smaller or equal nonce
+            // after the first or if there is a nonce gap
+            if let Some(old_nonce) = maybe_old_nonce {
+                if sys_txn.nonce() != old_nonce + 1 {
+                    return Err(BlockValidationError::TxnError);
```
ariqchowdhury: this should be sys txn error, right? Should probably add a debug log too

omegablitz (approved):

ariqchowdhury (approved):

PR #2132, opened 2025-08-06
feat: configurable transaction expiry timeouts to node config

0xObsidian:
So, basically transaction expiry timeouts were hardcoded as `Duration::from_secs(15)` for soft expiry and `Duration::from_secs(5 * 60)` for hard expiry. This prevented operators from tuning these values like for instance for different network conditions.

## Issue Reproduction
Before this fix, adding tx_expiry fields to node.toml would fail: dding soft_tx_expiry_secs = 30 and hard_tx_expiry_secs = 600 to node.toml would result in "unknown field" TOML parsing errors
```bash
cargo test -p monad-node-config test_tx_expiry_fields_are_now_supported_in_toml
```

## Testing the introduced fix
```bash
cargo test -p monad-node-config --lib --features crypto
cargo check -p monad-node-config
```

### DIFF
--- a/monad-node-config/src/lib.rs
+++ b/monad-node-config/src/lib.rs
@@ -75,6 +75,12 @@ pub struct NodeConfig<ST: CertificateSignatureRecoverable> {
     // NETWORK-WIDE CONFIGURATION //
     ////////////////////////////////
     pub chain_id: u64,
+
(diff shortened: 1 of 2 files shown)


0xObsidian:
feel free to let me know if you have any questions

PR #2133, opened 2025-08-06
raptorcast, discovery: update trusted ip addresses when validator ip changes

dshulyak:
discovery is extended with additional event to capture when ip address of the validator change. 
raptorcast waits for this event and sends information to dataplane

### DIFF
--- a/monad-peer-disc-swarm/src/driver.rs
+++ b/monad-peer-disc-swarm/src/driver.rs
@@ -197,6 +197,7 @@ where
                 self.algo.update_peer_participation(end_round, peers)
             }
             PeerDiscoveryEvent::Refresh => self.algo.refresh(),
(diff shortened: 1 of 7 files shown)


xinyuan-dev (commented):
The connection to a trusted peer is not dropped when its address changes. Can the mechanism be exploited to exhaust tcp connections by a malicious validator by repeatedly rotating its addresses?

dshulyak:
you are right that it should be dropped, but also i didn't put any limits on validators connections in general, so that would be even simpler way to exhaust them. so far the idea was to prevent such attacks from random hosts, and make validators unaffected.

practically the cost of staking will be high enough to disincentivize such behavior, but it is somewhat sloppy and eventually there should be some limits in place for validators as well

PR #2139, opened 2025-08-07
Transaction static validation cleanup

eerkaijun:
- Some reformatting to show different validation checks that come from different Ethereum revisions
- Added EIP-7623 check which is recently added to [execution validation check](https://github.com/category-labs/monad/blob/2cdddeb22bb194abfbc4c42be7f0f77eb62c43a6/category/execution/ethereum/validate_transaction.cpp#L97) with unit tests
    - need to double check if we actually want this EIP. does execution already changes the gas cost for calldata tokens?


monad-eth-block-policy/src/lib.rs
```diff
@@ -108,26 +110,29 @@ pub fn static_validate_transaction(
...
 
+    // post Ethereum Prague fork validation
+    // includes EIP-7623 validation
+    EthPragueForkValidation::validate(tx)?;
+
+    // EIP-2 - verify that s is in the lower half of the curve
+    if tx.signature().normalize_s().is_some() {
+        return Err(TransactionError::UnsupportedTransactionType);
+    }
+
```
ariqchowdhury: should this be checked in its original spot?

monad-eth-block-policy/src/lib.rs
```diff
@@ -108,26 +110,29 @@ pub fn static_validate_transaction(
...
-    // max init_code is (2 * max_code_size)
-    let max_init_code_size: usize = 2 * max_code_size;
-    if tx.kind().is_create() && tx.input().len() > max_init_code_size {
-        return Err(TransactionError::InitCodeLimitExceeded);
-    }
+    // post Ethereum Shanghai fork validation
+    // includes EIP-3860 validation
+    EthShanghaiForkValidation::validate(tx, max_code_size)?;
 
     // YP eq. 62 - intrinsic gas validation
```
ariqchowdhury: does it make sense to have a yellow paper validator?

monad-eth-block-policy/src/lib.rs
```diff
@@ -95,11 +102,6 @@ pub fn static_validate_transaction(
     proposal_gas_limit: u64,
     max_code_size: usize,
 ) -> Result<(), TransactionError> {
-    // EIP-2 - verify that s is in the lower half of the curve
-    if tx.signature().normalize_s().is_some() {
-        return Err(TransactionError::UnsupportedTransactionType);
-    }
-
     // EIP-155
```
ariqchowdhury: I think these standalone eips can also be put into functions

monad-eth-block-policy/src/lib.rs
```diff
@@ -139,6 +144,53 @@ pub fn static_validate_transaction(
     Ok(())
 }
 
+struct EthLondonForkValidation {}
+impl EthLondonForkValidation {
+    fn validate(tx: &TxEnvelope) -> Result<(), TransactionError> {
+        Self::eip_1559(tx)
+    }
+    fn eip_1559(tx: &TxEnvelope) -> Result<(), TransactionError> {
+        // EIP-1559
```
ariqchowdhury: can remove these comments now

ariqchowdhury (commented):
can also consider having a separate static validation module that has all the validators so that looking at `monad-eth-block-policy/src/static_validation.rs` would give a quick overview of what is being checked

monad-eth-block-policy/src/lib.rs
eerkaijun: i'm trying to keep to the same ordering as the `static_validate_transaction` function in the execution client

eerkaijun: actually this can be placed into one of the fork

PR #2143, opened 2025-08-08
Only insert name record upon a ping pong round trip

eerkaijun:
Main changes are to only insert name records into routing table upon a successful ping pong round trip. This is to ensure that the incoming peer indeed holds the IP it advertises. Also removed periodic ping pongs -- ping pongs are now only done during initial insertion (to establish handshake). This changes simplify previous separation of `routing_info` and `connection_info`, which makes things very confusing. Tested in unit tests and peer disc swarm tests.

### DIFF
--- a/monad-peer-discovery/src/discovery.rs
+++ b/monad-peer-discovery/src/discovery.rs
@@ -40,7 +40,7 @@ const MAX_PEER_IN_RESPONSE: usize = 16;
 /// Number of peers to send lookup request to
 const NUM_LOOKUP_PEERS: usize = 3;
 /// Number of validators to connect to if self is a full node
-// TODO: this should be configurable (e.g. a dedicated full node does not need to have 3 upstream validators)
+// TODO: this should be configurable
 const NUM_UPSTREAM_VALIDATORS: usize = 3;
 
 /// Metrics constant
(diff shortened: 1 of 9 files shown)


ariqchowdhury:
was there an audit issue associated with this PR?

eerkaijun:
didn't have an issue opened since it's more of a design choice, but it's discussed in this slack thread
https://category-labs.slack.com/archives/C0854M7KF5M/p1752756152238229

ariqchowdhury (approved):

monad-peer-discovery/src/discovery.rs
```diff
@@ -488,59 +390,68 @@ where
...
-        // drop ping if peer list is full and incoming ping is not from a validator or pinned full node
-        if self.routing_info.len() >= self.max_num_peers
-            && !self.is_validator(&from)
+        // do not insert to pending queue if peer list is full and incoming ping is not from a validator or pinned full node
+        // we still respond with pong even if peer list is full
+        let mut peer_list_full = false;
+        if self.routing_info.len() + self.pending_queue.len() >= self.max_num_peers
+            && !self.check_validator_membership(&from)
             && !self.pinned_full_nodes.contains(&from)
             && !self.routing_info.contains_key(&from)
```
xinyuan-dev, line 545: ~~do we need to handle the ping message from a peer that's in pending_queue when peer list is full?~~

no. we're supposed to send pong regardless the peer list is full, as long as the ping message is valid: i.e. pub key valid and seq non-decreasing.

we only ping back when the peer list is not full.

monad-peer-discovery/src/discovery.rs
```diff
@@ -488,59 +390,68 @@ where
...
-                    ?from,
-                    "peer updated name record without bumping sequence number"
-                );
-                return cmds;
             }
         }
 
         // respond to ping
-        if let Some(socket_address) = self.routing_info.get(&from).map(|info| info.address()) {
+        if let Some(socket_address) = ping_msg
```
xinyuan-dev: if the ping message is invalid while the peer_list is full, should we still send a pong?

eerkaijun: by invalid do you mean that the name record attached is incorrect (i.e. backward sequence number)?

xinyuan-dev: yes, either backwards seq or the signature is bad.

eerkaijun: currently we don't drop them because otherwise it would require another signature verification (which is expensive). i think the goal is to avoid signature verification when the peer list is already full

fabiotdf (approved):

PR #2146, opened 2025-08-08
Reject invalid blocksync requests

bamrith:
This PR addresses handling of invalid blocksync header requests that have num_blocks = 0.

Added unit test to cover this case.

### DIFF
--- a/monad-blocksync/src/blocksync.rs
+++ b/monad-blocksync/src/blocksync.rs
@@ -37,7 +37,7 @@ use monad_validator::{
 };
 use rand::{seq::SliceRandom, SeedableRng};
 use rand_chacha::ChaCha8Rng;
-use tracing::debug;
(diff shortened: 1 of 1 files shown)


monad-blocksync/src/blocksync.rs
```diff
@@ -270,6 +270,15 @@ where
         requester: BlockSyncSelfRequester,
         block_range: BlockRange,
     ) -> Vec<BlockSyncCommand<ST, SCT, EPT>> {
+        if block_range.num_blocks == SeqNum(0) {
```
omegablitz, line 273: Is this possible? Even if not, we can leave it in to be defensive.

bamrith: I made sure it isn't possible right now. Added it anyway incase there's future changes that go unnoticed.

PR #2149, opened 2025-08-10
Send AdvanceRound messages to full-nodes

omegablitz:
Full-nodes currently only receive proposals. As a consequence, full-nodes will not bump their round even if the upstream validator has a higher QC/TC if there's no attached proposal.

### DIFF
--- a/monad-consensus-state/src/command.rs
+++ b/monad-consensus-state/src/command.rs
@@ -18,7 +18,7 @@ use std::time::Duration;
 use monad_consensus::{
     messages::{
         consensus_message::{ConsensusMessage, ProtocolMessage},
-        message::TimeoutMessage,
+        message::{AdvanceRoundMessage, TimeoutMessage},
     },
     pacemaker::PacemakerCommand,
     validation::signing::{Validated, Verified},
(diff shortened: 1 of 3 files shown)


eerkaijun:
PublishToFullNodes only send to public full nodes i think. Will dedicated full nodes be sent AdvanceRound messages too?

omegablitz:
I haven't read the code, but would think PublishToFullNodes should be how dedicated full nodes receive messages as well. If it's not right now we should probably change it to.

omegablitz:
Ref: https://github.com/category-labs/monad-bft/pull/1559

eerkaijun:
iiuc this is how dedicated full nodes receive message https://github.com/category-labs/monad-bft/blob/master/monad-raptorcast/src/udp.rs#L806 (in `BuildTarget::Raptorcast`)

public full nodes receive message through `BuildTarget::FullNodeRaptorcast` which is trigger by PublishByFullNodes

ariqchowdhury (approved):

michael-yxchen (approved):

ariqchowdhury (approved):

PR #2151, opened 2025-08-11
More accurate calculation configuration of bandwidth constraints

eerkaijun:
Closes https://github.com/category-labs/category-internal/issues/1818
More detailed description in the issue above. Previously we use number of full nodes across different groups to decide whether to join a group.

Tldr regardless of the group size, bandwidth cost is quite similar. Using max number of group that a client will join is a better configuration for bandwidth constraints (and also more intuitive). Unit tested


fabiotdf (approved):

ariqchowdhury:
is the config for number of groups and group size set via node.toml?

eerkaijun:
yes

omegablitz (approved):

PR #2155, opened 2025-08-12
Publish to full nodes target

xinyuan-dev:
Implements https://github.com/category-labs/category-internal/issues/1831.

The PR is unit tested by ensuring the `PublishToFullNodes` command correctly send the messages over wire to the enlisted dedicated full nodes.

### DIFF
--- a/monad-raptorcast/src/lib.rs
+++ b/monad-raptorcast/src/lib.rs
@@ -56,9 +56,7 @@ use monad_validator::signature_collection::SignatureCollection;
 use raptorcast_secondary::group_message::FullNodesGroupMessage;
 use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender};
 use tracing::{debug, debug_span, error, warn};
-use util::{
-    BuildTarget, EpochValidators, FullNodes, Group, ReBroadcastGroupMap, Redundancy, Validator,
-};
+use util::{BuildTarget, EpochValidators, FullNodes, Group, ReBroadcastGroupMap, Redundancy};
 
 pub mod config;
 pub mod message;
@@ -185,6 +183,10 @@ where
             self.channel_from_secondary = Some(channel_from_secondary);
         }
         self
+    }
+
+    pub fn set_dedicated_full_nodes(&mut self, nodes: Vec<NodeId<CertificateSignaturePubKey<ST>>>) {
+        self.dedicated_full_nodes = FullNodes::new(nodes);
     }
 
     fn enqueue_message_to_self(
@@ -366,31 +368,20 @@ where
                     self.rebroadcast_map
                         .push_group_validator_set(validator_set.clone(), epoch);
                     if let Some(epoch_validators) = self.epoch_validators.get(&epoch) {
-                        assert_eq!(validator_set.len(), epoch_validators.validators.len());
-                        assert!(validator_set.clone().into_iter().all(
-                            |(validator_key, validator_stake)| epoch_validators
-                                .validators
-                                .get(&validator_key)
-                                .map(|v| v.stake)
-                                == Some(validator_stake)
-                        ));
+                        assert_eq!(validator_set.len(), epoch_validators.len());
+
+                        assert!(validator_set
+                            .iter()
+                            .all(|(validator_key, validator_stake)| {
+                                epoch_validators.get(validator_key) == Some(*validator_stake)
+                            }));
+
                         warn!("duplicate validator set update (this is safe but unexpected)")
                     } else {
                         let removed = self.epoch_validators.insert(
                             epoch,
                             EpochValidators {
-                                validators: validator_set
-                                    .clone()
-                                    .into_iter()
-                                    .map(|(validator_key, validator_stake)| {
-                                        (
-                                            validator_key,
-                                            Validator {
-                                                stake: validator_stake,
-                                            },
-                                        )
-                                    })
-                                    .collect(),
+                                validators: validator_set.clone().into_iter().collect(),
                             },
                         );
                         assert!(removed.is_none());
@@ -398,7 +389,7 @@ where
                     self.peer_discovery_driver.lock().unwrap().update(
                         PeerDiscoveryEvent::UpdateValidatorSet {
                             epoch,
-                            validators: validator_set.into_iter().map(|(id, _)| id).collect(),
+                            validators: validator_set.iter().map(|(id, _)| *id).collect(),
                         },
                     );
                 }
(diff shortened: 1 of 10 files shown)


monad-raptorcast/src/lib.rs
```diff
@@ -521,7 +504,40 @@ where
                         }
                     };
                 }
-                RouterCommand::PublishToFullNodes { .. } => {}
+                RouterCommand::PublishToFullNodes { epoch, message } => {
+                    let full_nodes_view = self.dedicated_full_nodes.view();
```
xinyuan-dev, line 526: question: should we also send the message to rc secondary?

monad-raptorcast/src/udp.rs
```diff
@@ -747,74 +747,43 @@ where
...
-                        chunk_data[0..20].copy_from_slice(&compute_hash(node_id).0);
-                        *chunk_symbol_id = Some(chunk_idx);
-                        chunk_idx += 1;
-                    }
-                }
-            }
+            assert!(!epoch_validators.is_empty());
+
+            // generate chunks if epoch validators is not empty
+            // FIXME should self be included in total_stake?
```
xinyuan-dev, line 752: I wanted to ask the same question: should self be included in total_stake?

monad-raptorcast/src/udp.rs
```diff
@@ -747,74 +747,43 @@ where
...
+            let mut chunk_idx = 0_u16;
+            let mut validator_set: Vec<_> = epoch_validators.iter().collect();
+            // Group shuffling so chunks for small proposals aren't always assigned
+            // to the same nodes, until researchers come up with something better.
+            validator_set.shuffle(&mut rand::thread_rng());
+            for (node_id, stake) in validator_set {
+                let start_idx: usize =
+                    (num_packets as f64 * (running_stake / total_stake)) as usize;
+                running_stake += stake;
+                let end_idx: usize = (num_packets as f64 * (running_stake / total_stake)) as usize;
```
xinyuan-dev, line 773: please help review if the type conversion to f64 in the calculation is safe here.

the change here is to avoid making assumption about the underlying types of the Stake (c.f. https://github.com/category-labs/monad-bft/pull/2032) since all we care about is the relative portions of the total stake.

monad-raptorcast/src/lib.rs
```diff
@@ -521,7 +504,40 @@ where
...
+                    if self.is_dynamic_fullnode {
```
xinyuan-dev, line 527: is this condition sufficient for excluding self as a full node?

monad-raptorcast/src/lib.rs
```diff
@@ -521,7 +504,40 @@ where
...
+                        OutboundRouterMessage::<OM, ST>::AppMessage(message).try_serialize();
+                    let outbound_message = match app_message {
+                        Ok(msg) => msg,
+                        Err(err) => {
+                            error!(?err, "failed to serialize a message");
+                            continue;
+                        }
+                    };
+
+                    let build_target = BuildTarget::Broadcast(full_nodes_view.into());
```
xinyuan-dev: if we broadcast the message to all full_nodes, how are we sure if they'll relay the message to each other? or perhaps we should use point to point?

monad-raptorcast/src/lib.rs
fabiotdf: Should be. In utils.rs you have Group<> constructors that exclude self node id, so the object is always filtered from the get-go

monad-raptorcast/src/lib.rs
fabiotdf: May I ask why the primary instance is handling this? Are we merging Primary and Secondary instances into 1?

monad-raptorcast/src/lib.rs
eerkaijun: if self is running as a dedicated full node, this will return false. but i guess it's fine if self does not have other dedicated full nodes because it will exit in the next condition anyways (`if full_nodes_view.is_empty()`)

monad-raptorcast/src/lib.rs
```diff
@@ -521,7 +504,40 @@ where
...
                 }
-                RouterCommand::PublishToFullNodes { .. } => {}
+                RouterCommand::PublishToFullNodes { epoch, message } => {
+                    let full_nodes_view = self.dedicated_full_nodes.view();
+                    if self.is_dynamic_fullnode {
+                        continue; // we're already a full node.
+                    }
+
+                    if full_nodes_view.is_empty() {
+                        warn!("No full nodes to publish message to");
```
eerkaijun: lower log level? I think this is a common path

monad-raptorcast/src/lib.rs
eerkaijun: dedicated full nodes wouldn't relay message to each other. it should be point to point

monad-raptorcast/src/lib.rs
eerkaijun: we don't need to send to rc secondary here, it should be done in multi router  https://github.com/category-labs/monad-bft/blob/d00b37d/monad-router-multi/src/lib.rs#L125

monad-raptorcast/src/udp.rs
fabiotdf: Looks like it makes no difference for the final result

monad-raptorcast/src/udp.rs
fabiotdf: Is there a check somewhere against stake being zero? Maybe in a degenerate case of 1 validator and smh stake is zero, we'd have division by zero here

fabiotdf (dismissed):

monad-raptorcast/src/udp.rs
xinyuan-dev: Indeed. Although I don't think the scenario is likely under normal operations, but let me add a check for that.

omegablitz:
The `forward: impl FnMut(Bytes, u16)` and all the ForwardBatcher stuff should be deletable now, because PublishToFullNodes already encapsulates proposal forwarding. As is, each proposal will be forwarded twice - once when ConsensusCommand::PublishToFullNodes is called on the proposal, and once when the chunks are directly forwarded.

example: https://github.com/category-labs/monad-bft/pull/1559/files#diff-e3f805714fb0c90bbac775b6e694ddbc0a555dcae963060791ca06e38036d826

monad-types/src/lib.rs
```diff
@@ -584,6 +584,14 @@ impl SubAssign for Stake {
     }
 }
 
+impl Div<Stake> for Stake {
```
omegablitz, line 587: Is this easy to misuse elsewhere? One option is to do the division manually, or create a separate `fn lossy_div(self, rhs: Self) -> f64` or something to reduce footguns.
I don't feel too strongly though, I'm fine leaving it as is.

omegablitz (approved):

monad-types/src/lib.rs
xinyuan-dev: or we replace it with a checked_div and document all the cases where a None can be returned: div by zero, overflow, precision loss, etc.

sometimes precision loss is acceptable, then we may have a separate checked_div_lossy function for that purpose. this way the user of the method must be conscious of what they're doing.

i suggest we implement the change from a separate pr.

PR #2158, opened 2025-08-12
Archive: Reimplement bft uploading

JoeHowarth:
Avoids inefficient ListObjectsV2 scan of entire store


monad-archive/src/bin/monad-archiver/cli.rs
```diff
@@ -62,9 +62,12 @@ pub struct Cli {
     #[arg(long)]
     pub bft_block_path: Option<PathBuf>,
 
-    #[arg(long, default_value_t = 5)]
+    #[arg(long, default_value_t = 60)]
```
mikety: Has this increase been tested? Combined with "Delay"  missed tick behavior in tokio will it lead to long hangs and burst of the blocks that needs to be archived?

monad-archive/src/workers/bft_archive_worker.rs
```diff
@@ -0,0 +1,620 @@
...
+    // 3) Process files concurrently using streams
+    stream::iter(local.into_iter())
+        .map(|(key, path)| {
+            let store = store.clone();
+            let metrics = metrics.clone();
+            async move {
+                match process_single_file(store, &key, &path, &metrics).await {
+                    Ok(x) => x,
+                    Err(e) => {
+                        error!(?e, ?key, ?path, "Failed to process BFT block");
```
mikety, line 141: What happens if due to some other error all block stores fail, should it raise a critical error to get an attention to it?

mikety (dismissed):
lgtm with questions

monad-archive/src/workers/bft_archive_worker.rs
JoeHowarth: The thought is the metrics pipeline should pick this up and we alert on it

monad-archive/src/bin/monad-archiver/cli.rs
JoeHowarth: Reverted - may want to bump later to reduce s3 api requests, but fine to omit for now

PR #2159, opened 2025-08-12
Adjust user gas limit to provider gas limit

philipglazman:
Geth adjusts the gas limit to the block gas limit. We currently return an error without attempting estimation. To conform with Geth, we should silently limit the user's gas limit to the provider gas limit. I'm not aware of any workflows that break, so its a minor conformance issue.


eerkaijun:
could this be a bit misleading to the end users? e.g. if block gas limit is 150M, provider gas limit is 30M, and the user sets transaction gas limit as 45M, this silently clamp it to 30M and might returns an out of gas but it's actually sufficient gas
(the user wouldn't know it's transaction gets limited to the provider gas limit)

PR #2160, opened 2025-08-12
Reserve balance tfm spec

mikety:
Prior related PRs with feedback
https://github.com/category-labs/monad-bft/pull/2069
https://github.com/category-labs/monad-bft/pull/2031

### DIFF
--- a/monad-blocktree/src/blocktree.rs
+++ b/monad-blocktree/src/blocktree.rs
@@ -308,7 +308,8 @@ where
                     metrics.consensus_events.rx_base_fee_error += 1;
                 }
                 Err(
-                    BlockPolicyError::BlockNotCoherent
+                    BlockPolicyError::BlockPolicyBlockValidatorError(_)
+                    | BlockPolicyError::BlockNotCoherent
                     | BlockPolicyError::TimestampError
                     | BlockPolicyError::StateBackendError(StateBackendError::NeverAvailable),
                 ) => {
@@ -547,10 +548,10 @@ mod test {
         },
         NopKeyPair, NopSignature,
     };
-    use monad_eth_types::{Balance, EMPTY_RLP_TX_LIST};
+    use monad_eth_types::EMPTY_RLP_TX_LIST;
     use monad_state_backend::{InMemoryState, InMemoryStateInner};
     use monad_testutil::signing::MockSignatures;
-    use monad_types::{Epoch, NodeId, Round, SeqNum, GENESIS_SEQ_NUM};
+    use monad_types::{Balance, Epoch, NodeId, Round, SeqNum, GENESIS_SEQ_NUM};
 
     use super::BlockTree;
     use crate::blocktree::RootInfo;
--- a/monad-chain-config/src/lib.rs
+++ b/monad-chain-config/src/lib.rs
@@ -16,7 +16,7 @@
 use execution_revision::MonadExecutionRevision;
 use monad_types::{Epoch, Round, SeqNum};
 use revision::{
-    ChainParams, ChainRevision, MockChainRevision, MonadChainRevision, CHAIN_PARAMS_LATEST,
+    ChainParams, ChainRevision, MockChainRevision, MonadChainRevision, CHAIN_PARAMS_LATEST_NEXT,
 };
 use serde::Deserialize;
 use thiserror::Error;
@@ -221,7 +221,7 @@ pub struct MockChainConfig {
 
 impl MockChainConfig {
     pub const DEFAULT: Self = Self {
-        chain_params: &CHAIN_PARAMS_LATEST,
+        chain_params: &CHAIN_PARAMS_LATEST_NEXT,
         epoch_length: SeqNum::MAX,
         epoch_start_delay: Round::MAX,
     };
--- a/monad-chain-config/src/revision.rs
+++ b/monad-chain-config/src/revision.rs
@@ -16,6 +16,7 @@
 use std::{fmt::Debug, time::Duration};
 
 pub const CHAIN_PARAMS_LATEST: ChainParams = CHAIN_PARAMS_V_0_10_0;
+pub const CHAIN_PARAMS_LATEST_NEXT: ChainParams = CHAIN_PARAMS_V_0_11_0;
 
 pub trait ChainRevision: Copy + Clone {
     fn chain_params(&self) -> &'static ChainParams;
@@ -27,6 +28,7 @@ pub enum MonadChainRevision {
     V_0_7_0,
     V_0_8_0,
     V_0_10_0,
+    V_0_11_0,
 }
 
 impl ChainRevision for MonadChainRevision {
@@ -35,6 +37,7 @@ impl ChainRevision for MonadChainRevision {
             MonadChainRevision::V_0_7_0 => &CHAIN_PARAMS_V_0_7_0,
             MonadChainRevision::V_0_8_0 => &CHAIN_PARAMS_V_0_8_0,
             MonadChainRevision::V_0_10_0 => &CHAIN_PARAMS_V_0_10_0,
+            MonadChainRevision::V_0_11_0 => &CHAIN_PARAMS_V_0_11_0,
         }
     }
 }
@@ -46,7 +49,7 @@ pub struct MockChainRevision {
 
 impl MockChainRevision {
     pub const DEFAULT: Self = Self {
-        chain_params: &CHAIN_PARAMS_LATEST,
+        chain_params: &CHAIN_PARAMS_LATEST_NEXT,
     };
 }
 
@@ -62,6 +65,7 @@ pub struct ChainParams {
     pub proposal_gas_limit: u64,
     // Max proposal size in bytes (average transactions ~400 bytes)
     pub proposal_byte_limit: u64,
+    pub max_reserve_balance: u128,
     pub vote_pace: Duration,
 
     pub tfm: bool,
@@ -72,6 +76,7 @@ const CHAIN_PARAMS_V_0_7_0: ChainParams = ChainParams {
     tx_limit: 10_000,
     proposal_gas_limit: 300_000_000,
     proposal_byte_limit: 4_000_000,
+    max_reserve_balance: 100_000_000_000_000_000_000, // 100 MON
     vote_pace: Duration::from_millis(1000),
 
     tfm: false,
@@ -82,6 +87,7 @@ const CHAIN_PARAMS_V_0_8_0: ChainParams = ChainParams {
     tx_limit: 5_000,
     proposal_gas_limit: 150_000_000,
     proposal_byte_limit: 2_000_000,
+    max_reserve_balance: 100_000_000_000_000_000_000, // 100 MON
     vote_pace: Duration::from_millis(500),
 
     tfm: false,
@@ -92,10 +98,22 @@ const CHAIN_PARAMS_V_0_10_0: ChainParams = ChainParams {
     tx_limit: 5_000,
     proposal_gas_limit: 150_000_000,
     proposal_byte_limit: 2_000_000,
+    max_reserve_balance: 100_000_000_000_000_000_000, // 100 MON
     vote_pace: Duration::from_millis(400),
 
     tfm: false,
     eip_7702: false,
+};
+
+const CHAIN_PARAMS_V_0_11_0: ChainParams = ChainParams {
+    tx_limit: 5_000,
+    proposal_gas_limit: 150_000_000,
+    proposal_byte_limit: 2_000_000,
+    max_reserve_balance: 100_000_000_000_000_000_000, // 100 MON
+    vote_pace: Duration::from_millis(400),
+
+    tfm: true,
+    eip_7702: true,
 };
 
 #[cfg(test)]
@@ -106,5 +124,6 @@ mod test {
     fn chain_revision_ord() {
         assert!(MonadChainRevision::V_0_7_0 < MonadChainRevision::V_0_8_0);
         assert!(MonadChainRevision::V_0_8_0 < MonadChainRevision::V_0_10_0);
+        assert!(MonadChainRevision::V_0_10_0 < MonadChainRevision::V_0_11_0);
     }
 }
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -1893,9 +1893,7 @@ mod test {
     };
     use monad_eth_block_policy::EthBlockPolicy;
     use monad_eth_block_validator::EthBlockValidator;
-    use monad_eth_types::{
-        Balance, EthBlockBody, EthExecutionProtocol, EthHeader, ProposedEthHeader,
-    };
+    use monad_eth_types::{EthBlockBody, EthExecutionProtocol, EthHeader, ProposedEthHeader};
     use monad_multi_sig::MultiSig;
     use monad_state_backend::{InMemoryState, InMemoryStateInner, StateBackend, StateBackendTest};
     use monad_testutil::{
@@ -1904,7 +1902,7 @@ mod test {
         validators::create_keys_w_validators,
     };
     use monad_types::{
-        BlockId, Epoch, ExecutionProtocol, NodeId, Round, RouterTarget, SeqNum, Stake,
+        Balance, BlockId, Epoch, ExecutionProtocol, NodeId, Round, RouterTarget, SeqNum, Stake,
         GENESIS_SEQ_NUM,
     };
     use monad_validator::{
@@ -1929,11 +1927,14 @@ mod test {
     const BASE_FEE_MOMENT: u64 = 0;
     const EPOCH_LENGTH: SeqNum = SeqNum(100);
     const EPOCH_START_DELAY: Round = Round(20);
+    const GAS_LIMIT: u64 = 30000;
+    const RESERVE_BALANCE: u128 = 1_000_000_000_000_000_000;
 
     static CHAIN_PARAMS: ChainParams = ChainParams {
         tx_limit: 10_000,
         proposal_gas_limit: 300_000_000,
         proposal_byte_limit: 4_000_000,
+        max_reserve_balance: RESERVE_BALANCE,
         vote_pace: Duration::from_millis(1000),
 
         tfm: false,
--- a/monad-consensus-types/src/block.rs
+++ b/monad-consensus-types/src/block.rs
@@ -13,8 +13,9 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-use std::{fmt::Debug, ops::Deref};
+use std::{collections::BTreeMap, fmt::Debug, ops::Deref};
 
+use alloy_primitives::Address;
 use alloy_rlp::{RlpDecodable, RlpEncodable};
 use auto_impl::auto_impl;
 use bytes::Bytes;
@@ -27,7 +28,10 @@ use monad_crypto::{
     hasher::{Hasher, HasherType},
 };
 use monad_state_backend::{InMemoryState, StateBackend, StateBackendError};
-use monad_types::{BlockId, Epoch, ExecutionProtocol, FinalizedHeader, NodeId, Round, SeqNum};
+use monad_types::{
+    Balance, BlockId, Epoch, ExecutionProtocol, FinalizedHeader, NodeId, Round, SeqNum,
+    GENESIS_SEQ_NUM,
+};
 use monad_validator::signature_collection::SignatureCollection;
 
 use crate::{
@@ -170,12 +174,79 @@ pub enum BlockPolicyError {
     TimestampError,
     ExecutionResultMismatch,
     BaseFeeError,
+    BlockPolicyBlockValidatorError(BlockPolicyBlockValidatorError),
 }
 
 impl From<StateBackendError> for BlockPolicyError {
     fn from(err: StateBackendError) -> Self {
         Self::StateBackendError(err)
     }
+}
+
+#[derive(Debug, Clone)]
+pub struct AccountBalanceState {
+    pub balance: Balance,
+    pub remaining_reserve_balance: Balance,
+    pub max_reserve_balance: Balance,
+    pub block_seqnum_of_latest_txn: SeqNum,
+}
+
+impl AccountBalanceState {
+    pub fn new(max_reserve_balance: Balance) -> Self {
+        AccountBalanceState {
+            balance: Balance::ZERO,
+            remaining_reserve_balance: Balance::ZERO,
+            max_reserve_balance,
+            block_seqnum_of_latest_txn: GENESIS_SEQ_NUM,
+        }
+    }
+}
+
+pub type AccountBalanceStates = BTreeMap<Address, AccountBalanceState>;
+
+#[derive(Debug, PartialEq)]
+pub enum BlockPolicyBlockValidatorError {
+    AccountBalanceMissing,
+    InsufficientBalance,
+    InsufficientReserveBalance,
+}
+
+#[derive(Debug, Clone)]
+pub struct TxnFee {
+    pub first_txn_value: Balance,
+    pub first_txn_gas: Balance,
+    pub max_gas_cost: Balance,
+    pub max_txn_cost: Balance, // Used for pre TFM validation
+}
+
+pub type TxnFees = BTreeMap<Address, TxnFee>;
+
+pub trait BlockPolicyBlockValidator<CRT>
+where
+    Self: Sized,
+    CRT: ChainRevision,
+{
+    type Transaction;
+
+    fn new(
+        block_seq_num: SeqNum,
+        execution_delay: SeqNum,
+        base_fee: u64,
+        chain_revision: &CRT,
+    ) -> Result<Self, BlockPolicyError>;
+
+    fn try_apply_block_fees(
+        &self,
+        account_balance: &mut AccountBalanceState,
+        fees: &TxnFee,
+        eth_address: &Address,
+    ) -> Result<(), BlockPolicyError>;
+
+    fn try_add_transaction(
+        &self,
+        account_balances: &mut BTreeMap<&Address, AccountBalanceState>,
+        txn: &Self::Transaction,
+    ) -> Result<(), BlockPolicyError>;
 }
 
 /// Trait that represents how inner contents of a block should be validated
--- a/monad-eth-block-policy/src/lib.rs
+++ b/monad-eth-block-policy/src/lib.rs
@@ -13,7 +13,11 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-use std::{collections::BTreeMap, marker::PhantomData, ops::Deref};
+use std::{
+    collections::BTreeMap,
+    marker::PhantomData,
+    ops::{Deref, Range, RangeFrom},
+};
 
 use alloy_consensus::{
     transaction::{Recovered, Transaction},
@@ -23,19 +27,24 @@ use alloy_primitives::{Address, TxHash, U256};
 use itertools::Itertools;
 use monad_chain_config::{revision::ChainRevision, ChainConfig};
 use monad_consensus_types::{
-    block::{BlockPolicy, BlockPolicyError, ConsensusFullBlock},
+    block::{
+        AccountBalanceState, BlockPolicy, BlockPolicyBlockValidator,
+        BlockPolicyBlockValidatorError, BlockPolicyError, ConsensusFullBlock, TxnFee, TxnFees,
+    },
     checkpoint::RootInfo,
 };
 use monad_crypto::certificate_signature::{
     CertificateSignaturePubKey, CertificateSignatureRecoverable,
 };
-use monad_eth_types::{Balance, EthAccount, EthExecutionProtocol, EthHeader, Nonce};
+use monad_eth_types::{EthAccount, EthExecutionProtocol, EthHeader};
 use monad_state_backend::{StateBackend, StateBackendError};
 use monad_system_calls::SystemTransaction;
-use monad_types::{BlockId, Epoch, Round, SeqNum, GENESIS_BLOCK_ID, GENESIS_SEQ_NUM};
+use monad_types::{
+    Balance, BlockId, Epoch, Nonce, Round, SeqNum, GENESIS_BLOCK_ID, GENESIS_ROUND, GENESIS_SEQ_NUM,
+};
 use monad_validator::signature_collection::SignatureCollection;
 use sorted_vector_map::SortedVectorMap;
-use tracing::{trace, warn};
+use tracing::{debug, trace, warn};
 
 pub mod validation;
 
@@ -49,17 +58,36 @@ pub enum ReserveBalanceCheck {
     Validate,
 }
 
-pub fn compute_txn_max_value(txn: &TxEnvelope) -> U256 {
+pub fn compute_max_txn_cost(txn: &TxEnvelope) -> U256 {
     let txn_value = txn.value();
     let gas_limit = U256::from(txn.gas_limit());
     let max_fee = U256::from(txn.max_fee_per_gas());
-    let gas_cost = gas_limit.checked_mul(max_fee).expect("no overflow");
+    let priority_fee = U256::from(txn.max_priority_fee_per_gas().unwrap_or(0));
+    let max_gas_cost = gas_limit
+        .checked_mul(max_fee.checked_add(priority_fee).expect("no overflow"))
+        .expect("no overflow");
+    txn_value.saturating_add(max_gas_cost)
+}
+
+pub fn compute_txn_max_value(txn: &TxEnvelope, base_fee: u64) -> U256 {
+    let txn_value = txn.value();
+    let gas_cost = compute_txn_max_gas_cost(txn, base_fee);
     txn_value.saturating_add(gas_cost)
+}
+
+pub fn compute_txn_max_gas_cost(txn: &TxEnvelope, base_fee: u64) -> U256 {
+    let gas_limit = U256::from(txn.gas_limit());
+    let max_fee = U256::from(txn.max_fee_per_gas());
+    let priority_fee = U256::from(txn.max_priority_fee_per_gas().unwrap_or(0));
+    let base_fee = U256::from(base_fee);
+    let gas_bid = max_fee.min(base_fee.saturating_add(priority_fee));
+    gas_limit.checked_mul(gas_bid).expect("no overflow")
 }
 
 struct BlockLookupIndex {
     block_id: BlockId,
     seq_num: SeqNum,
+    round: Round,
     is_finalized: bool,
 }
 
@@ -75,7 +103,7 @@ where
     pub system_txns: Vec<SystemTransaction>,
     pub validated_txns: Vec<Recovered<TxEnvelope>>,
     pub nonces: BTreeMap<Address, Nonce>,
-    pub txn_fees: BTreeMap<Address, Balance>,
+    pub txn_fees: TxnFees,
 }
 
 impl<ST, SCT> AsRef<EthValidatedBlock<ST, SCT>> for EthValidatedBlock<ST, SCT>
@@ -183,12 +211,12 @@ impl BlockAccountNonce {
 }
 
 #[derive(Debug)]
-struct BlockTxnFees {
-    txn_fees: BTreeMap<Address, Balance>,
+struct BlockTxnFeeStates {
+    txn_fees: TxnFees,
 }
 
-impl BlockTxnFees {
-    fn get(&self, eth_address: &Address) -> Option<Balance> {
+impl BlockTxnFeeStates {
+    fn get(&self, eth_address: &Address) -> Option<TxnFee> {
         self.txn_fees.get(eth_address).cloned()
     }
 }
@@ -198,9 +226,9 @@ struct CommittedBlock {
     block_id: BlockId,
     round: Round,
     epoch: Epoch,
-
+    seq_num: SeqNum,
     nonces: BlockAccountNonce,
-    fees: BlockTxnFees,
+    fees: BlockTxnFeeStates,
 
     base_fee: u64,
     base_fee_trend: u64,
@@ -211,14 +239,9 @@ struct CommittedBlock {
 #[derive(Debug)]
 struct CommittedBlkBuffer<ST, SCT, CCT, CRT> {
     blocks: SortedVectorMap<SeqNum, CommittedBlock>,
-    size: usize, // should be execution delay
+    min_buffer_size: usize, // should be 2 * execution delay
 
     _phantom: PhantomData<(ST, SCT, fn(&CCT, &CRT))>,
-}
-
-struct CommittedTxnFeeResult {
-    txn_fee: Balance,
-    next_validate: SeqNum, // next block number to validate; included for assertions only
 }
 
 impl<ST, SCT, CCT, CRT> CommittedBlkBuffer<ST, SCT, CCT, CRT>
@@ -228,10 +251,10 @@ where
     CCT: ChainConfig<CRT>,
     CRT: ChainRevision,
 {
-    fn new(size: usize) -> Self {
+    fn new(min_buffer_size: usize) -> Self {
         Self {
             blocks: Default::default(),
-            size,
+            min_buffer_size,
 
             _phantom: Default::default(),
         }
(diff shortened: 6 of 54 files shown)


monad-eth-block-policy/src/lib.rs
```diff
@@ -270,31 +280,26 @@ struct CommittedBlock {
...
     nonces: BlockAccountNonce,
-    fees: BlockTxnFees,
+    fees: BlockTxnFeeStates,
 }
 
 #[derive(Debug)]
 struct CommittedBlkBuffer<ST, SCT> {
     blocks: SortedVectorMap<SeqNum, CommittedBlock>,
-    size: usize, // should be execution delay
+    min_buffer_size: usize, // should be execution delay
```
eerkaijun: is this now 2 * execution delay

monad-eth-block-policy/src/lib.rs
```diff
@@ -314,27 +319,43 @@ where
...
-
-        // start iteration from base_seq_num (non inclusive)
-        for (&cache_seq_num, block) in self.blocks.range(next_validate..) {
-            assert_eq!(next_validate, cache_seq_num);
-            if let Some(account_txn_fee) = block.fees.get(eth_address) {
-                txn_fee = txn_fee.saturating_add(account_txn_fee);
-            }
-            next_validate += SeqNum(1);
+        execution_delay: SeqNum,
+        only_seqnum: bool,
```
eerkaijun: what does `only_seqnum` mean?

monad-eth-block-policy/src/lib.rs
```diff
@@ -314,27 +319,43 @@ where
...
-            txn_fee,
-            next_validate,
+                validator.try_apply_block_fees(
+                    account_balance,
+                    &block_txn_fees,
+                    eth_address,
+                    only_seqnum,
+                )?;
+            }
+            *base_seq_num = cache_seq_num;
```
eerkaijun: should there be an assertion that cache_seq_num is always increasing by 1?

monad-eth-block-policy/src/lib.rs
```diff
@@ -551,92 +789,89 @@ where
...
+                        true,
+                    )?;
+
+                    self.committed_cache.update_account_balance(
+                        &mut next_seq_num,
+                        &mut balance_state,
+                        &address,
+                        self.execution_delay,
+                        false,
+                    )?;
```
eerkaijun: am I understanding correctly that the first `.update_account_balances()` where only_seqnum == true is for computing account balances from block n-k+1 up till n, and the second `.update_account_balances()` where only_seqnum == false is for computing balances from block n-2k+2 up till n-k+1? if so, can we add a comment to make it clearer?

eerkaijun: also is there any reason that these two `.update_account_balances()` cannot be combined?

monad-eth-block-policy/src/lib.rs
```diff
@@ -551,92 +789,89 @@ where
...
+
+                    // Apply Txn Fees for txns in extending blocks
+                    if let Some(blocks) = extending_blocks {
+                        // handle the case where base_seq_num is a pending block
+                        let next_blocks = blocks
+                            .iter()
+                            .skip_while(move |block| block.get_seq_num() < next_seq_num);
+                        for extending_block in next_blocks {
+                            trace!(?next_seq_num, "looking for txn fees in extanding block");
```
eerkaijun: nit: extending spelling

monad-eth-block-policy/src/lib.rs
```diff
@@ -375,6 +399,211 @@ where
...
 }
 
+pub struct EthBlockPolicyBlockState {
+    fees: BTreeMap<Address, TxnFees>,
+}
+
+pub struct EthBlockPolicyBlockValidator {
+    block_seq_num: SeqNum,
+    min_blocks_since_latest_txn: SeqNum,
+}
```
eerkaijun: i feel like it's a little weird that we need to initialize this struct every time we want to check a block -- `min_blocks_since_latest_txn` is just a constant right? 

i feel like all the functions in this struct can be folded into perhaps EthBlockPolicy and `block_seq_num` will just be a function argument. But I guess we can also refactor later

monad-eth-block-policy/src/lib.rs
mikety: yes

monad-eth-block-policy/src/lib.rs
mikety: To compute seq_num of the last txn block, so can compute if the blocks following the state from tiredb have emptying transactions

monad-eth-block-policy/src/lib.rs
mikety: added another assert to match the block seq_num. increment by 1  might not hold if statesync is needed

monad-eth-block-policy/src/lib.rs
mikety: update_account_balances can be combined. Will add more comments.

monad-eth-block-policy/src/lib.rs
mikety: Its better to refactor later if needed. The validator is separating validation logic from block policy.
the min_blocks_since_latest_txn  changed to be a function

monad-eth-block-validator/src/lib.rs
```diff
@@ -179,9 +176,19 @@ where
...
-            *txn_fee_entry = txn_fee_entry.saturating_add(compute_txn_max_value(eth_txn));
+            let txn_fee_entry = txn_fees.entry(eth_txn.signer()).or_insert(TxnFee {
+                first_txn_value: Balance::ZERO,
+                first_txn_gas: Balance::ZERO,
+                max_gas_cost: Balance::ZERO,
+            });
+            if i == 0 {
+                txn_fee_entry.first_txn_value = eth_txn.value();
+                txn_fee_entry.first_txn_gas = compute_txn_max_gas_cost(eth_txn);
+            } else {
```
eerkaijun: only the first transaction in the block (regardless of signer) has `first_txn_value` and `first_txn_gas` populated, is that expected?

mikety: Per the TFM spec, only the first txn in the last k blocks can be "emptying" . only if txn is emptying its value is needed to recompute the reserve_balance. The fees  needs to store the map of signer -> first txn instead.

monad-eth-block-policy/src/lib.rs
```diff
@@ -377,6 +430,206 @@ where
...
+                seq_num =?self.block_seq_num,
+                ?eth_address,
+                "account balance have not been populated"
+            );
+            return Err(BlockPolicyError::BlockPolicyBlockValidatorError(
+                BlockPolicyBlockValidatorError::AccountBalanceMissing,
+            ));
+        }
+
+        let account_balance = maybe_account_balance.unwrap();
```
eerkaijun: this can be rewritten into 
```
        let Some(account_balance) = maybe_account_balance else {
            warn!(
                seq_num =?self.block_seq_num,
                ?eth_address,
                "account balance have not been populated"
            );
            return Err(BlockPolicyError::BlockPolicyBlockValidatorError(
                BlockPolicyBlockValidatorError::AccountBalanceMissing,
            ));
        };
````

File: monad-eth-block-policy/src/lib.rs [hunk 9]
```diff
@@ -377,6 +430,206 @@ where
...
+        block_txn_fees: &TxnFee,
+        eth_address: &Address,
+    ) -> Result<(), BlockPolicyError> {
+        // txn T is emptying if there is no "prior txn" i.e. a txn from the same sender sent from block P so that P >= block_number(T) - k + 1.
+        let blocks_since_latest_txn = self
+            .block_seq_num
+            .max(account_balance.block_seqnum_of_latest_txn)
+            - account_balance.block_seqnum_of_latest_txn;
+
+        let has_emptying_transaction = blocks_since_latest_txn > self.execution_delay - SeqNum(1);
```
eerkaijun: should this be >= instead of >?

monad-eth-block-validator/src/lib.rs
eerkaijun: fixed in [acf29c2](https://github.com/category-labs/monad-bft/pull/2160/commits/acf29c257d0576770f1e835dc7b8d289532de8c4)

monad-eth-block-policy/src/lib.rs
```diff
@@ -553,92 +814,76 @@ where
...
+                                let mut validator = EthBlockPolicyBlockValidator::new(
+                                    next_seq_num,
+                                    self.execution_delay,
+                                )?;
+
+                                validator.try_apply_block_fees(
+                                    &mut balance_state,
+                                    txn_fee,
+                                    &address,
+                                )?;
```
eerkaijun: why are `extending_blocks` being treated differently from blocks in `self.committed_cache`? we do `self.committed_cache.update_account_balance` but here we do `try_apply_block_fees` for an extending block. Shouldn't blocks be treated the same regardless whether it's a committed block or an extending block?

monad-eth-block-policy/src/lib.rs
mikety: Per tfm   for the txn T to be emptying  (algo 11)  there should be no prior sender.
"prior sender" is defined as a txn if  `blocknumber(Tp) >= blocknumber(T) - k + 1`
hence "no prior sender" means `blocknumber(Tp) < blocknumber(T) - k + 1`
in the code :
`block_since_latest_txn  =  blocknumber(T)  - blocknumber(Tp)` 
hence the condition. 

I will double check with Kushal that its correct.

monad-eth-block-policy/src/lib.rs
mikety: It should be the same result. there is no common method cause `extending_blocks` data structure is different from the `committed_cache`.

monad-eth-block-policy/src/lib.rs
eerkaijun: aaa ok i think you're right

monad-eth-block-policy/src/lib.rs
```diff
@@ -852,188 +1065,334 @@ mod test {
...
         let secret_key = B256::repeat_byte(0xAu8).to_string();
         let signer = &secret_key.parse::<PrivateKeySigner>().unwrap();
         signer.sign_hash_sync(signature_hash).unwrap()
     }
 
+    fn make_test_tx(value: u64, nonce: u64, signer: FixedBytes<32>) -> Recovered<TxEnvelope> {
+        recover_tx(make_legacy_tx(
+            signer,
+            BASE_FEE_PER_GAS as u128,
+            value,
```
eerkaijun, line 1334: the third argument of `make_legacy_tx` takes in gas_limit, but in your tests it seems like you are trying to pass in the transaction value

monad-eth-block-policy/src/lib.rs
```diff
@@ -316,46 +321,93 @@ where
...
-                txn_fee = txn_fee.saturating_add(account_txn_fee);
+        execution_delay: SeqNum,
+    ) -> Result<(), BlockPolicyError> {
+        let next_seq_num = *base_seq_num + SeqNum(1);
+
+        // base_seq_num = N - k, where N is the proposed block and k is execution delay.
+        // transactions inclusion starting block = N - k + 1
+        // check for emptying txn starting block = N - 2k + 2
+        let txn_inclusion_start = next_seq_num;
+        let emptying_txn_check_start =
```
eerkaijun: there are various instance of this code that checks for empty transaction, e.g. it also appears in `try_add_transaction` and `try_apply_block_fees`. I would recommend putting it into a function, so when we add 7702 transactions that's the only place we need to check for account delegation

monad-eth-block-policy/src/lib.rs
eerkaijun: fixed

monad-blocktree/src/blocktree.rs
```diff
@@ -290,6 +290,9 @@ where
                 Err(BlockPolicyError::ExecutionResultMismatch) => {
                     metrics.consensus_events.rx_bad_state_root += 1;
                 }
+                Err(BlockPolicyError::BlockPolicyBlockValidatorError(_)) => {
+                    // TODO add metrics
+                }
                 Err(
                     BlockPolicyError::BlockNotCoherent
                     | BlockPolicyError::TimestampError
```
eerkaijun: nit: put them in the same arm

monad-eth-txpool-types/src/lib.rs
```diff
@@ -83,9 +84,8 @@ impl EthTxPoolDropReason {
                 TransactionError::InitCodeLimitExceeded => "Init code size limit exceeded",
                 TransactionError::GasLimitTooLow => "Gas limit too low",
                 TransactionError::GasLimitTooHigh => "Exceeds block gas limit",
-                TransactionError::UnsupportedTransactionType => {
-                    "EIP4844 and EIP7702 transactions unsupported"
-                }
+                TransactionError::InvalidSetCodeTx => "Invalid EIP 7702 transaction",
+                TransactionError::UnsupportedTransactionType => "EIP4844 transaction unsupported",
```
andr-dev: nit: Let's flip the order of these so they match the enum definition order

monad-node/src/main.rs
```diff
@@ -290,6 +290,17 @@ async fn run(node_state: NodeState, reload_handle: Box<dyn TracingReload>) -> Re
             GENESIS_SEQ_NUM, // FIXME: MonadStateBuilder is responsible for updating this to forkpoint root if necessary
             EXECUTION_DELAY,
             node_state.node_config.chain_id,
+            node_state
```
omegablitz: Why is max_reserve_balance a chain parameter if it's fixed in EthBlockPolicy?

monad-eth-txpool/src/pool/tracked/mod.rs
```diff
@@ -365,15 +369,20 @@ where
...
+    ) -> Result<(u64, Vec<Recovered<TxEnvelope>>), BlockPolicyError> {
         assert!(tx_limit > 0);
 
         let mut txs = Vec::new();
         let mut total_gas = 0u64;
         let mut total_size = 0u64;
 
-        tx_heap.drain_in_order_while(|sender, tx| {
+        let mut balances = account_balances;
+        let mut validator = EthBlockPolicyBlockValidator::new(proposed_seq_num, execution_delay)?;
```
andr-dev: We're passing in `proposed_seq_num: SeqNum` and `execution_delay: SeqNum` just to build a `EthBlockPolicyBlockValidator`, let's just create it before calling `self.create_proposal_tx_list` (line 228) and pass it in.

monad-eth-txpool/src/pool/tracked/mod.rs
```diff
@@ -365,15 +369,20 @@ where
...
+        account_balances: BTreeMap<Address, AccountBalanceState>,
+    ) -> Result<(u64, Vec<Recovered<TxEnvelope>>), BlockPolicyError> {
         assert!(tx_limit > 0);
 
         let mut txs = Vec::new();
         let mut total_gas = 0u64;
         let mut total_size = 0u64;
 
-        tx_heap.drain_in_order_while(|sender, tx| {
+        let mut balances = account_balances;
```
andr-dev: This is unnecessary, let's make `account_balances` mut again and delete this line.

monad-types/src/lib.rs
```diff
@@ -70,6 +70,9 @@ impl Round {
     }
 }
 
+pub type Balance = U256;
```
omegablitz, line 73: Why did these have to be moved here? Is this because of monad-state-backend/src/mock.rs?

monad-eth-txpool/src/pool/tracked/mod.rs
```diff
@@ -389,23 +398,21 @@ where
...
-                error!(
-                    ?sender,
-                    "txpool create_proposal account_balances lookup failed"
-                );
-                return TrackedTxHeapDrainAction::Skip;
-            };
+            let res = validator.try_add_transaction(&mut balances, tx.raw());
 
-            let Some(new_account_balance) = tx.apply_max_value(*account_balance) else {
+            if res.is_err() {
```
andr-dev:
```rust
if let Err(error) = validator.try_add_transaction(&mut balances, tx.raw()); {
  ...
}
```

monad-eth-txpool/src/pool/transaction.rs
```diff
@@ -118,6 +122,23 @@ impl ValidEthTransaction {
         None
     }
 
+    pub fn apply_max_gas_cost(&self, reserve_balance: Balance) -> Option<Balance> {
```
andr-dev: This argument should be called `balance`, it's an implementation detail that the (currently only) caller of this function passes in `reserve_balance`.

monad-eth-txpool/src/pool/transaction.rs
```diff
@@ -118,6 +122,23 @@ impl ValidEthTransaction {
...
+        }
+
+        trace!(
+            "AccountBalance insert_tx 3 \
+                            do not add txn to the pool. insufficient reserve balance after applying carriage cost: {reserve_balance:?} \
+                            max_gas_cost: {max_gas_cost:?} \
+                            for address: {address:?}",
+            max_gas_cost = self.max_gas_cost,
+            address = self.tx.signer()
+        );
```
andr-dev: This trace should be moved to the callsite as it does not belong to this function.

monad-eth-block-validator/src/lib.rs
```diff
@@ -568,7 +575,7 @@ mod test {
         // ECDSA signature is malleable
         // given a signature, we can form a second signature by computing additive inverse of s and flips v
         let original_signature = valid_txn.signature();
-        let secp256k1_n = U256::from_str_radix(
+        let secp256k1_n = Balance::from_str_radix(
```
omegablitz: why use the Balance alias?

monad-eth-txpool/src/pool/transaction.rs
```diff
@@ -118,6 +122,23 @@ impl ValidEthTransaction {
         None
     }
 
+    pub fn apply_max_gas_cost(&self, reserve_balance: Balance) -> Option<Balance> {
+        if let Some(reserve_balance) = reserve_balance.checked_sub(self.max_gas_cost) {
+            return Some(reserve_balance);
+        }
+
+        trace!(
+            "AccountBalance insert_tx 3 \
```
andr-dev: What does `insert_tx 3` mean?

monad-eth-txpool/tests/pool.rs
```diff
@@ -1111,3 +1116,42 @@ fn test_exceed_byte_limit() {
...
+        .map(|txn| txn.signer())
+        .collect_vec();
+
+    let balances = block_policy.compute_account_base_balances(
+        block.get_seq_num(),
+        &state_backend,
+        extending_blocks,
+        tx_signers.iter(),
+    );
+}
```
andr-dev: This test doesn't seem to use the txpool, is there a reason for putting it here?

monad-state/src/lib.rs
```diff
@@ -1256,13 +1256,15 @@ where
         let mut commands = Vec::new();
 
         let delay = self.consensus_config.execution_delay;
+        /// TFM reserve balance checking requires N-2*state_root_delay blocks to validate N
         let last_delay_committed_blocks: Vec<_> = root_parent_chain
```
omegablitz: last_two_delay_committed_blocks

eth-block-policy should be reset with last_two_delay_committed_blocks

Only `last_delay_committed_blocks` should be re-proposed/finalized. The changes in monad-state-backend/src/in_memory.rs should be deletable after this.

monad-state-backend/src/in_memory.rs
```diff
@@ -63,6 +63,7 @@ pub struct InMemoryBlockState {
     round: Round,
     parent_id: BlockId,
     nonces: BTreeMap<Address, Nonce>,
+    balances: Option<BTreeMap<Address, Balance>>,
```
omegablitz: what's this for?

monad-state-backend/src/in_memory.rs
omegablitz, line 1: After reading the code further, I don't think these changes should be necessary. Explained further in a different comment.

monad-state/src/lib.rs
```diff
@@ -1256,13 +1256,15 @@ where
         let mut commands = Vec::new();
 
         let delay = self.consensus_config.execution_delay;
+        /// TFM reserve balance checking requires N-2*state_root_delay blocks to validate N
```
omegablitz: Please also update this comment: https://github.com/category-labs/monad-bft/blob/29fcde61eeb244fc1cfd5b355e4b4c03d68e0947/monad-state/src/lib.rs#L1324-L1329

monad-state/src/statesync.rs
```diff
@@ -68,7 +70,7 @@ where
         Self {
             max_buffered_proposals: resync_threshold.0 as usize,
             resync_threshold,
-            state_root_delay,
```
omegablitz, line 71: nit: while you're here, rename state_root_delay -> execution_delay

monad-state-backend/src/mock.rs
omegablitz, line 1: What is this for?

eerkaijun: for unit testing, to inject arbitrary values to balances and nonces at a particular block

monad-eth-txpool/src/pool/transaction.rs
mikety: for debugging, removed as the log also moved

monad-node/src/main.rs
mikety: Please clarify, how you suggest to set it? I think max reserve balance will be changing.

monad-types/src/lib.rs
mikety: there is a circular dependency if it stays.

monad-state-backend/src/in_memory.rs
mikety: removed

monad-node/src/main.rs
omegablitz: If it's hard-coded, it should be outside of ChainConfig.

If it's in chain-config, then eth-block-policy should take chain-config in as a parameter and use the chain-config for deciding which reserve balance to check against.

Right now, it's sort of a mix of the two, where it's specified in Chain-config but can't actually fork on it

monad-eth-txpool/src/pool/tracked/mod.rs
```diff
@@ -365,15 +370,16 @@ where
         proposal_gas_limit: u64,
         proposal_byte_limit: u64,
         tx_heap: TrackedTxHeap<'_>,
-        mut account_balances: BTreeMap<&Address, Balance>,
-    ) -> (u64, Vec<Recovered<TxEnvelope>>) {
+        validator: &mut EthBlockPolicyBlockValidator,
+        balances: &mut BTreeMap<Address, AccountBalanceState>,
```
andr-dev:
```rust
mut account_balances: BTreeMap<&Address, AccountBalanceState>,
```

monad-eth-txpool/src/pool/tracked/mod.rs
```diff
@@ -365,15 +370,16 @@ where
...
+    ) -> Result<(u64, Vec<Recovered<TxEnvelope>>), BlockPolicyError> {
```
andr-dev: This method doesn't produce an error, let's remove the Result.

monad-eth-txpool/src/pool/transaction.rs
```diff
@@ -118,6 +122,13 @@ impl ValidEthTransaction {
         None
     }
 
+    pub fn apply_max_gas_cost(&self, balance: Balance) -> Option<Balance> {
+        if let Some(balance) = balance.checked_sub(self.max_gas_cost) {
+            return Some(balance);
+        }
+        None
```
andr-dev:
```rust
balance.checked_sub(self.max_gas_cost)
```

monad-node/src/main.rs
mikety: Thanks, I think its will be better to have an ability to fork on it, but its better to do in a separate change.

monad-state/src/lib.rs
mikety: discussed both approaches on slack, will keep this one.

monad-node/src/main.rs
omegablitz: Can you remove it from ChainConfig in the interim then?

monad-eth-txpool/src/pool/tracked/mod.rs
mikety: Changed to use BTreeMap<&Address, AccountBalanceState>, but still need &mut IMO to avoid copying the sate.

monad-eth-txpool/src/pool/transaction.rs
```diff
@@ -39,6 +40,8 @@ pub struct ValidEthTransaction {
     forward_last_seqnum: SeqNum,
     forward_retries: usize,
     max_value: Balance,
+    max_gas_cost: Balance,
+    effective_tip_per_gas: u128,
```
andr-dev: This field was removed in a recent commit, I think it may have snuck its way back in after the rebase. Let's remove it.

monad-eth-txpool/tests/pool.rs
```diff
@@ -1206,3 +1210,44 @@ fn test_proposal_tx_low_base_fee() {
...
```
andr-dev: I believe this is the test we moved earlier because it shouldn't be here, let's move it again.

monad-consensus-state/src/lib.rs
```diff
@@ -1346,6 +1347,15 @@ where
...
             // epoch manager records
             self.metrics.consensus_events.commit_block += 1;
+
+            let tfm_enabled = self
+                .config
+                .chain_config
+                .get_chain_revision(block.header().block_round)
+                .chain_params()
+                .tfm_enabled;
+            self.block_policy.set_tfm_enabled(tfm_enabled);
```
mikety: As an alternative to threading chain_config into BlockPolicy.

monad-chain-config/src/revision.rs
```diff
@@ -72,6 +80,7 @@ const CHAIN_PARAMS_V_0_7_0: ChainParams = ChainParams {
     tx_limit: 10_000,
     proposal_gas_limit: 300_000_000,
     proposal_byte_limit: 4_000_000,
+    max_reserve_balance: 100_000_000_000_000_000_000, // 100 MON
```
andr-dev, line 79: Shouldn't these be 0? If `tfm` is disabled, the `max_reserve_balance` should never be used so it's likely safer to set it to min.

monad-chain-config/src/lib.rs
```diff
@@ -226,6 +227,12 @@ impl MockChainConfig {
         epoch_start_delay: Round::MAX,
     };
 
+    pub const DEFAULT_NEXT_RELEASE: Self = Self {
```
andr-dev: I think it's better if we change `DEFAULT ` to have `tfm` enabled instead of adding a new one, this is for MockChainConfig anyways.

monad-consensus-types/src/block.rs
```diff
@@ -178,6 +183,83 @@ impl From<StateBackendError> for BlockPolicyError {
...
+}
+
+impl Default for TxnFee {
+    fn default() -> Self {
+        TxnFee {
+            first_txn_value: Balance::ZERO,
+            first_txn_gas: Balance::ZERO,
+            max_gas_cost: Balance::ZERO,
+            max_txn_cost: Balance::ZERO,
+        }
```
andr-dev: Does `Balance` not impl `Default`?

monad-consensus-types/src/block.rs
```diff
@@ -178,6 +183,83 @@ impl From<StateBackendError> for BlockPolicyError {
...
+    }
+}
+
+pub type TxnFees = BTreeMap<Address, TxnFee>;
+
+pub trait BlockPolicyBlockValidator
```
andr-dev: Should `BlockPolicy` define an associated `BlockPolicyBlockValidator` type? In txpool, we're using a hard-coded `EthBlockPolicyBlockValidator` when it should be taking a generic `BlockPolicyBlockValidator` that is type-related to the `BlockPolicy` it is using.

monad-consensus-types/src/block.rs
```diff
@@ -178,6 +183,83 @@ impl From<StateBackendError> for BlockPolicyError {
...
+where
+    Self: Sized,
+{
+    type Transaction;
+
+    fn new(
+        block_seq_num: SeqNum,
+        execution_delay: SeqNum,
+        base_fee: u64,
+        tfm_enabled: bool,
```
andr-dev: We've already passed `ChainConfig` to `BlockPolicy`, so this should likely take a `CRT: ChainRevision` or use `ChainParams` directly.

monad-chain-config/src/revision.rs
mikety: it is a good test to catch issues, but it will be better to do so in a test for a soft fork scenario

monad-consensus-types/src/block.rs
mikety: Balance is U256, I think its better to set it explicitly to 0.

monad-consensus-types/src/block.rs
mikety: may be - we can do it in a follow up once all features are in?

monad-consensus-types/src/block.rs
mikety: I can change it as a follow up.

monad-consensus-types/src/block.rs
andr-dev: All signed/unsigned integers defaults to 0.

monad-eth-block-policy/src/lib.rs
```diff
@@ -49,17 +58,36 @@ pub enum ReserveBalanceCheck {
...
+    let gas_cost = compute_txn_max_gas_cost(txn, base_fee);
     txn_value.saturating_add(gas_cost)
 }
 
+pub fn compute_txn_max_gas_cost(txn: &TxEnvelope, base_fee: u64) -> U256 {
+    let gas_limit = U256::from(txn.gas_limit());
+    let max_fee = U256::from(txn.max_fee_per_gas());
+    let priority_fee = U256::from(txn.max_priority_fee_per_gas().unwrap_or(0));
+    let base_fee = U256::from(base_fee);
+    let gas_bid = max_fee.min(base_fee.saturating_add(priority_fee));
```
andr-dev, line 83: Isn't this just `effective_gas_price`?

monad-eth-block-policy/src/lib.rs
```diff
@@ -326,6 +398,305 @@ where
...
+}
+
+fn is_possibly_emptying_transaction(
+    block_seq_num_of_curr_txn: SeqNum,
+    block_seqnum_of_latest_txn: SeqNum,
+    execution_delay: SeqNum,
+) -> bool {
+    // txn T is emptying if there is no "prior txn" i.e. a txn from the same sender sent from block P so that P >= block_number(T) - k + 1.
+    let blocks_since_latest_txn =
+        block_seq_num_of_curr_txn.max(block_seqnum_of_latest_txn) - block_seqnum_of_latest_txn;
```
andr-dev: `block_seq_num_of_curr_txn.saturating_sub(block_seqnum_of_latest_txn)`

monad-consensus-types/src/block.rs
mikety: ok, removed so its implied

monad-eth-block-policy/src/lib.rs
```diff
@@ -326,6 +398,305 @@ where
...
+            block_seq_num,
+            execution_delay,
+            base_fee,
+            tfm_enabled,
+            max_reserve_balance: Balance::from(max_reserve_balance),
+        })
+    }
+
+    fn try_apply_block_fees(
+        &mut self,
```
andr-dev: Why is this method taking a mutable reference? It doesn't seem to mutate self.

monad-eth-block-policy/src/lib.rs
```diff
@@ -326,6 +398,305 @@ where
...
+        account_balance: &mut AccountBalanceState,
+        block_txn_fees: &TxnFee,
+        eth_address: &Address,
+    ) -> Result<(), BlockPolicyError> {
+        if !self.tfm_enabled {
```
andr-dev: Why do we have separate logic for whether TFM is enable or not? I would have expected that we do the same computation regardless of TFM activation, but when pulling balances from an `AccountBalanceState` we pull the reserve balance when TFM is enabled and regular balance otherwise.

This approach also avoids bugs at the activation edges since we're always computing balance values in the same way, but when validating a new block we pick which balance based on the activation rather than changing the computation itself. For example, below we set `account_balance.remaining_reserve_balance = estimated_balance.min(self.max_reserve_balance);`. If we transitioned from TFM off to TFM on, could that cause a stale `account_balance.remaining_reserve_balance` that would be different if TFM was always enabled?

monad-eth-block-policy/src/lib.rs
```diff
@@ -326,6 +398,305 @@ where
...
+            ?account_balance,
+            ?self.block_seq_num,
+            ?eth_address,
+            "try_apply_block_fees updated balance state",
+        );
+        Ok(())
+    }
+
+    fn try_add_transaction(
+        &mut self,
```
andr-dev: Am I missing a mutation of self in this fn as well, or should this be changed to immutable ref?

monad-eth-block-policy/src/lib.rs
```diff
@@ -498,16 +874,24 @@ where
...
     where
         SCT: SignatureCollection,
     {
-        trace!(block = consensus_block_seq_num.0, "compute_base_balance");
-
         // calculation correct only if GENESIS_SEQ_NUM == 0
         assert_eq!(GENESIS_SEQ_NUM, SeqNum(0));
-        let base_seq_num = consensus_block_seq_num.max(self.execution_delay) - self.execution_delay;
+        let base_seq_num =
+            (consensus_block_seq_num).max(self.execution_delay) - self.execution_delay;
```
andr-dev: Extra parens?

monad-state-backend/src/mock.rs
```diff
@@ -0,0 +1,91 @@
...
+use monad_types::{Balance, BlockId, Nonce, SeqNum, Stake};
+use monad_validator::signature_collection::{SignatureCollection, SignatureCollectionPubKeyType};
+
+use crate::{StateBackend, StateBackendError};
+
+#[derive(Debug, Default, Clone)]
+pub struct NopStateBackend {
+    pub nonces: BTreeMap<Address, Nonce>,
+    pub balances: BTreeMap<Address, Balance>,
+}
```
andr-dev, line 32: We already have `InMemoryBlockState` which is used in mock swarm tests and in txpool for manually setting per-account nonces which can be trivially extended to also set per-account balances.

monad-eth-block-policy/src/lib.rs
```diff
@@ -257,46 +285,89 @@ where
...
+                "Reserve balance check range is not contiguous"
+            );
+
+            if let Some(block_txn_fees) = block.fees.get(eth_address) {
+                let mut validator = EthBlockPolicyBlockValidator::new(
+                    block.seq_num,
+                    execution_delay,
+                    block.base_fee,
+                    get_tfm(chain_config, block.round),
+                    get_max_reserve_balance(chain_config, block.round),
```
andr-dev: `get_tfm` and `get_max_reserve_balance` both fetch the chain params for the current block round and then pull out a specific parameter. This is a strong indicator that `EthBlockPolicyBlockValidator` should take a `ChainParam`s or even better take a `&CCT` and get the `ChainParam` fields it wants itself.

monad-eth-block-policy/src/lib.rs
```diff
@@ -517,87 +901,103 @@ where
...
+                                } else {
+                                    let mut validator = EthBlockPolicyBlockValidator::new(
+                                        extending_block.get_seq_num(),
+                                        self.execution_delay,
+                                        extending_block.get_base_fee(),
+                                        get_tfm(chain_config, extending_block.get_block_round()),
+                                        get_max_reserve_balance(
+                                            chain_config,
+                                            extending_block.get_block_round(),
+                                        ),
```
andr-dev: Here's another place where we're double calling `chain_config.get_chain_revision()`, let's just pass `&CCT` to `EthBlockPolicyBlockValidator`.

monad-consensus-types/src/block.rs
andr-dev: Sorry, should have been clearer. I think having a generic `BlockPolicyBlockValidator` in txpool must be changed in this PR because having the hard-coded `EthBlockPolicyBlockValidator` is incorrect. Whether `BlockPolicy` gets an associated type is the question and that could be done in a future PR but maybe it makes more sense to do it here.

monad-eth-block-policy/src/lib.rs
mikety: We need separate logic because the computation is separate and keeping it separate helps to avoid bugs - the logic is too different. Each block is processed per the TFM setting so it should be consistent across the nodes.

monad-eth-block-policy/src/lib.rs
mikety: I follow the TFM spec as a reference, I think its better to use explicit formula for gas_bid.

monad-eth-block-policy/src/lib.rs
mikety: its SeqNum, no defined saturationg_sub.   If you think its clearer I can use `SeqNum(block_seq_num_of_curr_txn.0.saturating_sub(block_seqnum_of_latest_txn.0));`

monad-state-backend/src/mock.rs
mikety: I think it can be done as a follow up.

monad-eth-block-policy/src/lib.rs
mikety: Ok, but it should ber `CRT` since the TxPool does not know the current round and it needs to be passed with the CCT to get the revision.

monad-eth-block-policy/src/lib.rs
mikety: As explained below it should be `&CRT`

monad-consensus-types/src/block.rs
mikety: done

monad-consensus-types/src/block.rs
mikety: Totally makes sense but I think it will be better to make a follow up small commit instead. I will send you a PR soon.

monad-eth-block-policy/src/lib.rs
```diff
@@ -925,143 +1308,377 @@ where
...
+            block_policy,
+            incoming_block,
+            extending_blocks,
+            state_backend,
+            signers,
+        )
+    }
+
+    // #[test_case(3; "three committed blocks, one extending block")]
+    // #[test_case(0; "no committed blocks, four extending block")]
```
eerkaijun: why are tests commented out? is there a way we can still enable these with the tfm flag enabled?

monad-eth-block-validator/src/lib.rs
```diff
@@ -246,6 +244,14 @@ where
...
         };
 
+        /*
+        =======
+                chain_id: u64,
+                chain_params: &ChainParams,
+                execution_chain_params: &ExecutionChainParams,
+            ) -> Result<(SystemTransactions, ValidatedTxns, NonceMap, TxnFees), BlockValidationError> {
+        >>>>>>> 50193403 (reserve balance)
+                */
```
eerkaijun: remove?

monad-consensus-types/src/block.rs
```diff
@@ -178,6 +183,72 @@ impl From<StateBackendError> for BlockPolicyError {
...
+    InsufficientBalance,
+    InsufficientReserveBalance,
+}
+
+#[derive(Debug, Clone)]
+pub struct TxnFee {
+    pub first_txn_value: Balance,
+    pub first_txn_gas: Balance,
+    pub max_gas_cost: Balance,
+    pub max_txn_cost: Balance, // Used for pre TFM validation
```
eerkaijun, line 219: is the plan to remove this field eventually?

monad-eth-block-policy/src/lib.rs
mikety: they work , good catch, enabled

monad-consensus-types/src/block.rs
mikety: I think so - it technically should be there if there is a possibility to run with <= V10

andr-dev: I think it makes sense to keep them so we can always toggle between TFM on/off. Let's say there's a bug in the TFM implementation, a node config change could be quickly pushed out to disable TFM.

mikety: TFM validation is more relaxed than current balance validation, so the node switched from  TFM to the old validation may not validate the transaction. The transition to TFM model should work since its from more strict to less.

monad-eth-block-policy/src/lib.rs
```diff
@@ -326,6 +397,317 @@ where
...
+                "account balance have not been populated"
+            );
+            return Err(BlockPolicyError::BlockPolicyBlockValidatorError(
+                BlockPolicyBlockValidatorError::AccountBalanceMissing,
+            ));
+        };
+
+        let tfm_enabled = self.chain_revision.chain_params().tfm;
+        let max_reserve_balance =
+            Balance::from(self.chain_revision.chain_params().max_reserve_balance);
```
eerkaijun: this is not used - is that expected?

mikety: No, can be removed.

ariqchowdhury (dismissed):

eerkaijun (approved):

ariqchowdhury (approved):

monad-consensus-types/src/block.rs
andr-dev: Right, what I'm saying is that we might want to revert to more restricted balance validation if there's a critical bug that isn't easily solvable (in bft or in execution).

andr-dev (approved):

PR #2162, opened 2025-08-12
Full nodes connection

eerkaijun:
Closes https://github.com/category-labs/category-internal/issues/1772 (more details in the issue linked)
This introduces two new message types in peer discovery for (1) a full node to indicate that it wants to look for an upstream validator running as Publisher and (2) a validator indicating it's running as Publisher and has space

This PR allows a full node to keep looking for upstream validators until it finds 3 upstream that is running as Publisher mode. 

Tested in peer disc swarm

### DIFF
--- a/monad-peer-discovery/src/discovery.rs
+++ b/monad-peer-discovery/src/discovery.rs
@@ -40,7 +40,7 @@ const MAX_PEER_IN_RESPONSE: usize = 16;
 /// Number of peers to send lookup request to
 const NUM_LOOKUP_PEERS: usize = 3;
 /// Number of validators to connect to if self is a full node
-// TODO: this should be configurable (e.g. a dedicated full node does not need to have 3 upstream validators)
+// TODO: this should be configurable
 const NUM_UPSTREAM_VALIDATORS: usize = 3;
 
 /// Metrics constant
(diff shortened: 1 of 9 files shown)


fabiotdf (approved):

monad-peer-discovery/src/discovery.rs
```diff
@@ -190,9 +212,14 @@ impl<ST: CertificateSignatureRecoverable> PeerDiscoveryAlgoBuilder for PeerDisco
...
-        if state.check_validator_membership(&self.self_id) {
-            state.self_role = Role::Validator;
+        // if self is a validator in the current or next epoch, make sure it's recorded correctly
+        if state.self_role == PeerDiscoveryRole::ValidatorNone
+            || state.self_role == PeerDiscoveryRole::ValidatorPublisher
+        {
+            assert!(
+                state.check_validator_membership(&self.self_id),
+                "incorrectly set as validator"
+            );
```
ariqchowdhury: can you explain what this check is for? I also don't understand the comment

eerkaijun: it's to check that if a node sets [fullnode_raptorcast] as Publisher, it is indeed in the validator set. this check is on startup to prevent a full node from setting as a publisher

PR #2163, opened 2025-08-13
consensus eip7702

mikety:


monad-eth-block-policy/src/lib.rs
```diff
@@ -142,10 +142,20 @@ pub fn static_validate_transaction(
...
     }
 
-    if tx.is_eip4844() || tx.is_eip7702() {
+    if tx.is_eip4844() {
         return Err(TransactionError::UnsupportedTransactionType);
     }
 
+    if tx.is_eip7702() {
+        if let Some(auth_list) = tx.authorization_list() {
+            if auth_list.is_empty() || auth_list.len() >= 20 {
```
philipglazman: Why is there a cap on the list of authorization tuples?

monad-eth-block-validator/src/lib.rs
```diff
@@ -157,6 +154,36 @@ where
                     .max_gas_cost
                     .saturating_add(compute_txn_max_gas_cost(eth_txn));
             }
+
+            if eth_txn.is_eip7702() {
+                if let Some(auth_list) = eth_txn.authorization_list() {
+                    let authorities = auth_list
+                        .iter()
```
philipglazman, line 199: Potentially use rayon's `into_par_iter` for key recovery in `recover_authority`

monad-eth-block-validator/src/lib.rs
```diff
@@ -157,6 +154,36 @@ where
...
+                            if let Some(n) = nonces.get(&signer) {
+                                if *n != nonce {
+                                    return Err(BlockValidationError::TxnError);
+                                }
+                            }
+                            // TODO: what if nonces updated later in this block?
+                            nonces
+                                .entry(signer)
+                                .and_modify(|n| *n += 1)
+                                .or_insert(nonce + 1);
```
eerkaijun: as per my previous [comment](https://github.com/category-labs/monad-bft/pull/1892#discussion_r2210566011) i don't think this nonce check is correct

in the EIP it says 
```
6. Verify the nonce of authority is equal to nonce
```
nonces.get(&signer) is the previous nonce though, not the current nonce (i.e. it has been used in a previous transaction in the same block)

monad-triedb-utils/src/decode.rs
```diff
@@ -47,11 +47,17 @@ pub fn rlp_decode_account(account_rlp: Vec<u8>) -> Option<EthAccount> {
...
         None
     } else {
         match <[u8; 32]>::decode(&mut buf) {
-            Ok(x) => Some(x),
+            Ok(x) => {
+                if x.starts_with(b"ef0100") {
+                    is_delegated = true;
+                }
+                Some(x)
+            }
```
eerkaijun: i think this is incorrect - a delegated account has `(0xef0100 || address)` stored in its code, this is the code hash. we'll need the code hash to get the code

eerkaijun: there's a get_code function in triedb_env.rs

monad-eth-block-policy/tests/eip7702.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    #[test]
+    fn test_static_validate_no_authorization_list() {
+        let tx = TxEip7702 {
+            chain_id: CHAIN_ID,
+            nonce: 0,
+            gas_limit: PROPOSAL_GAS_LIMIT,
+            max_fee_per_gas: 1000,
+            max_priority_fee_per_gas: 0,
+            ..Default::default()
```
eerkaijun: nit: the default just set authorization_list to be `vec![]`, so this is essentially the same test case as `test_static_validate_empty_authorization_list()`

monad-eth-block-policy/src/lib.rs
```diff
@@ -142,10 +142,20 @@ pub fn static_validate_transaction(
...
+                return Err(TransactionError::InvalidSetCodeTx);
+            }
+        } else {
+            return Err(TransactionError::InvalidSetCodeTx);
```
ariqchowdhury: would rather use match here
```
        match tx.authorization_list() {
            Some(auth_list) => {
                if auth_list.is_empty() {
                    return Err(TransactionError::InvalidSetCodeTx);
                }
            }
            None => return Err(TransactionError::InvalidSetCodeTx),
        }
```
and echoing phil's comment, I don't see anything in the eip that restricts the length of the authorization list

monad-eth-block-policy/src/lib.rs
```diff
@@ -944,11 +964,37 @@ where
         }
 
         // TODO fix this unnecessary copy into a new vec to generate an owned Address
-        let tx_signers = block
-            .validated_txns
-            .iter()
-            .map(|txn| txn.signer())
-            .collect_vec();
+        let mut delegated_addresses: Vec<Address> = Vec::new();
```
ariqchowdhury: the eip uses the term `authority` for the address recovered from an authorization tuple so I think this vec should be `authority_addresses`

monad-eth-block-policy/src/lib.rs
```diff
@@ -944,11 +964,37 @@ where
...
+            if txn.is_eip7702() {
+                if let Some(auth_list) = txn.authorization_list() {
+                    for result in auth_list.iter().map(|a| a.recover_authority()) {
+                        match result {
+                            Ok(address) => {
+                                delegated_addresses.push(address);
+                            }
+                            Err(error) => {
+                                warn!(?error, "Can not process authorization list");
+                                return Err(BlockPolicyError::BlockNotCoherent);
```
ariqchowdhury: I think this should have a new error variant in BlockPolicyError

monad-eth-block-policy/src/lib.rs
```diff
@@ -965,6 +1011,15 @@ where
...
 
+        for signer in &delegated_addresses {
+            let account_balance = account_balances
+                .get_mut(signer)
+                .expect("account_balances should have been populated for delegated accounts");
+
+            debug!(delegated_account = ?signer, "Setting account to is_delegated: true");
+            account_balance.is_delegated = true;
+        }
+
```
ariqchowdhury, line 1082: don't we need to check the nonces of the authority accounts somewhere here? The nonce of the authority account has to be equal to the nonce included in the authorization tuple. That check has to be part of the coherency check because it needs to consider the branch of blocks being extended, right?

monad-eth-block-policy/src/lib.rs
```diff
@@ -944,11 +964,37 @@ where
...
-        let tx_signers = block
-            .validated_txns
-            .iter()
-            .map(|txn| txn.signer())
-            .collect_vec();
+        let mut delegated_addresses: Vec<Address> = Vec::new();
+        let mut tx_signers: Vec<Address> = Vec::new();
+
+        for tx_signer in block.validated_txns.iter().map(|txn| {
+            if txn.is_eip7702() {
```
ariqchowdhury: the eip says that the code of authority must be empty or already delegated. Don't we need to do that validation somewhere in coherency? Could there be a pending block that changed the code of an account before getting to the current block?

monad-eth-block-validator/src/lib.rs
```diff
@@ -157,6 +154,36 @@ where
...
             }
+
+            if eth_txn.is_eip7702() {
+                if let Some(auth_list) = eth_txn.authorization_list() {
+                    let authorities = auth_list
+                        .iter()
+                        .map(|a| (a.recover_authority(), a.nonce(), a.address))
+                        .collect_vec();
+                    for (result, nonce, address) in authorities {
+                        if let Ok(signer) = result {
```
ariqchowdhury: nit: rename signer to authority to match eip terminology

monad-eth-block-validator/src/lib.rs
ariqchowdhury: can you clarify what you mean by previous_nonce vs current_nonce here?

eerkaijun: E.g a user submits transaction with nonce 9 previously. we do `nonces.insert(9)` in the code above. Here we do `nonces.get()` which returns 9 and we check whether the delegation nonce is equal to 9. This is incorrect because the delegation nonce should be 10

mikety: the spec states that after the verification 6 it 

> 9. Increase the nonce of authority by one.

so as I understand the spec the nonce of the previous txn was incremented. Note that its incremented on line 175

monad-triedb-utils/src/decode.rs
mikety: good point,  but making 2 requests will slow down everything , the best would be reading from the db if its is_delegated in the get_account

monad-eth-block-policy/src/lib.rs
mikety: There is a way to get this checked, but first I wanted to get what can be in EOA code other then "empty" or "already delegated" i.e. with the prefix

monad-eth-block-policy/src/lib.rs
mikety: moved nonce check from block validator to here

monad-triedb-utils/src/decode.rs
eerkaijun: `x.starts_with(b"ef0100")`
but this wouldn't be correct because the code hash will be something else and wouldn't start with `ef0100`

monad-eth-block-validator/src/lib.rs
eerkaijun: in the example above, we are reusing nonce 9 twice

monad-triedb-utils/src/decode.rs
mikety: I agree - was talking to Bruce to see what is possible on the DB end to avoid a call to read code. Its pretty large change so Im adding the get_code call

monad-triedb-utils/src/lib.rs
```diff
@@ -170,7 +170,35 @@ impl TriedbReader {
...
-                maybe_rlp_account.and_then(rlp_decode_account)
+                let res = maybe_rlp_account.and_then(rlp_decode_account);
+                match res {
+                    Some(mut eth_account) => {
+                        match eth_account.code_hash {
+                            Some(code_hash) => {
+                                // Request code
+                                let (triedb_key, key_len_nibbles) =
+                                    create_triedb_key(version, KeyInput::CodeHash(&code_hash));
+                                let res = self.handle.read(&triedb_key, key_len_nibbles, seq_num.0);
```
mikety, line 184: use sync read for now to test correctness. Will do async after.

monad-eth-block-policy/src/lib.rs
```diff
@@ -945,14 +966,50 @@ where
...
+                    return Err(err);
+                }
+            }
+        }
+
+        
+        tx_signers.extend(authority_addresses.iter().cloned());
+
+        let mut system_tx_signers = block.system_txns.iter().map(|txn| txn.signer());
+        tx_sginers.append(&mut system_tx_signers);
```
philipglazman: typo. should be `tx_signers.extend`

monad-eth-block-policy/src/lib.rs
```diff
@@ -945,14 +966,50 @@ where
...
+
+                                // First increment than validate against account's nonce+1.
+                                authority_nonces
+                                    .entry(authority)
+                                    .and_modify(|n| *n += 1)
+                                    .or_insert(nonce + 1);
+                            }
+                            Err(error) => {
+                                warn!(?error, "Can not process authorization list");
+                                return Err(BlockPolicyError::Eip7702Error);
```
philipglazman: Should we return an error or continue processing the next tuple? According to the eip they suggest the latter

monad-eth-block-policy/src/lib.rs
```diff
@@ -1018,6 +1087,22 @@ where
                 return Err(BlockPolicyError::BlockNotCoherent);
             }
 
+            if authority_addresses.contains(&eth_address) {
+                let maybe_nonce = authority_nonces.get(&eth_address);
+                match maybe_nonce {
+                    Some(n) => {
+                        if *n != txn_nonce + 1 {
```
philipglazman: We increment the authority nonce earlier so I think this will always fail if authority_address contains eth_address. For example, I submit a transaction with nonce 2, and authority nonce 3. It is incremented to 4 earlier in the code path, so 4 != 2 +1

monad-eth-block-policy/src/lib.rs
```diff
@@ -709,14 +980,49 @@ where
...
         }
 
-        let system_tx_signers = block.system_txns.iter().map(|txn| txn.signer());
         // TODO fix this unnecessary copy into a new vec to generate an owned Address
-        let tx_signers = block
-            .validated_txns
-            .iter()
-            .map(|txn| txn.signer())
-            .chain(system_tx_signers)
-            .collect_vec();
```
eerkaijun: why are the system_tx_signers being deleted?

monad-triedb-utils/src/lib.rs
```diff
@@ -172,7 +172,37 @@ impl TriedbReader {
...
+                                debug!(?res, block_id = ?seq_num.0, ?eth_account, "account code_data");
+                                match res {
+                                    Some(data) => {
+                                        if data.len() >= 3 {
```
eerkaijun: should we check that data length is 23 bytes? from my understanding a delegated account will always be 0xef0100 || address where address is 20 bytes

monad-triedb-utils/src/decode.rs
```diff
@@ -63,6 +64,7 @@ pub fn rlp_decode_account(account_rlp: Vec<u8>) -> Option<EthAccount> {
         nonce,
         balance,
         code_hash: code_hash.map(B256::from),
+        is_delegated,
```
eerkaijun: putting is_delegated inside EthAccount is kinda weird imo because it's not part of the rlp encoding, can we add a TODO here so we can fix it later?

monad-eth-txpool/src/pool/tracked/mod.rs
```diff
@@ -192,7 +194,30 @@ where
...
+                                    .drop(txn.hash(), EthTxPoolDropReason::InvalidSignature);
+                            }
+                        }
+                    }
+                }
+                txn.signer()
+            })
+            .collect_vec();
+
+        addresses.append(&mut authority_addresses);
```
eerkaijun: is `addresses.extend(authority_addresses)` a cleaner way of writing this?

monad-eth-txpool/src/pool/tracked/mod.rs
```diff
@@ -207,12 +232,21 @@ where
...
+                    addresses.iter(),
                 )?,
                 state_backend.total_db_lookups() - total_db_lookups_before,
             )
         };
 
+        for signer in &authority_addresses {
+            let account_balance = account_balances
+                .get_mut(signer)
+                .expect("account_balances should have been populated for delegated accounts");
```
eerkaijun: just wanted to double check this is always the case?

monad-eth-block-policy/src/lib.rs
```diff
@@ -768,45 +1086,22 @@ where
...
-                    seq_num =? block.header().seq_num,
-                    round =? block.header().block_round,
-                    "block not coherent, invalid balance"
-                );
-                return Err(BlockPolicyError::BlockNotCoherent);
+            if authority_addresses.contains(&eth_address) {
+                let maybe_nonce = authority_nonces.get(&eth_address);
+                match maybe_nonce {
+                    Some(n) => {
+                        if *n != txn_nonce + 1 {
```
eerkaijun: i think there are some missing logic here. an example scenario (all in the same block): 
- alice sends a transaction with nonce n
- alice sends an authorization with nonce n + 1
- alice sends another transaction with nonce n + 2

this should be a valid block, but i think with our current code we separate out transaction nonce and authority nonce so we can't validate the sequence

eerkaijun: The code first checks transaction nonce, then checks authority nonce. But the sequencing matter, eg a user first sends a tx, then an authorization, then another tx, and the nonce needs to keep track of this ordering

ariqchowdhury:
do we check the chain_id of authorization_tuples somewhere? I don't see it here

monad-eth-block-policy/src/validation.rs
```diff
@@ -46,10 +46,21 @@ pub fn static_validate_transaction(
         return Err(TransactionError::GasLimitTooHigh);
     }
 
-    if tx.is_eip4844() || tx.is_eip7702() {
+    if tx.is_eip4844() {
         return Err(TransactionError::UnsupportedTransactionType);
     }
 
+    if tx.is_eip7702() {
+        match tx.authorization_list() {
```
andr-dev, line 54:
```rust
if tx.authorization_list().is_none_or(|auth_list| auth_list.is_empty()) {
  return Err(TransactionError::InvalidSetCodeTx);
}
```

eerkaijun:
more up to date branch https://github.com/category-labs/monad-bft/pull/2218

PR #2171, opened 2025-08-14
try_propose on process_qc/process_tc

omegablitz:
Now that process_qc and process_tc are consolidated, try_propose can be moved into the associated functions. This also fixes suboptimal behavior where a node would not try proposing when it bumped round via an AdvanceRound message.


michael-yxchen (approved):

ariqchowdhury:
the code depends on `try_propose` being idempotent (before this change too). Can you add a comment to the function that it needs to be idempotent for future.

ariqchowdhury (dismissed):

ariqchowdhury (approved):

PR #2174, opened 2025-08-14
Enable trace_calls

omegablitz:
Enable trace_calls so execution populates traces. Also, disable the /debian exclusion rule in CI for now, because that stops PRs from being mergeable.

https://github.com/category-labs/monad/commit/aa5c5fe89664ea31c9ea8eb26c4c11cacc32c56f

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -3,17 +3,11 @@ name: Rust
 on:
   merge_group:
     branches: ["master"]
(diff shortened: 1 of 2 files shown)


mhumeSF (approved):

omegablitz:
We don't need to merge this, but need to do this: https://github.com/category-labs/category-internal/issues/1840

PR #2182, opened 2025-08-17
Add staking support

bamrith:
First PR: https://github.com/category-labs/monad-bft/pull/1917

Changes:
- Implemented `read_next_valset` for TrieDB handle
- Added `SystemTransactionGenerator` and 3 new system calls (snapshot, epoch change and reward)
- Added staking activation epoch to chain config
- Use staking activation epoch for
    - Reading validator set from state backend
    - Generating and validating system transactions
    - Starting U256 leader schedule
- Added `epoch_length` and `epoch_start_delay` to chain config for testing
- Added `ExtractEthAddress` trait to get the block author's eth address for rewards
- Added writing new validator sets to `validators.toml`


monad-crypto/src/certificate_signature.rs
```diff
@@ -41,6 +42,9 @@ pub trait PubKey:
     type Error: Display + Debug + Send + Sync;
     fn from_bytes(pubkey: &[u8]) -> Result<Self, Self::Error>;
     fn bytes(&self) -> Vec<u8>;
+
+    // FIXME this is required to get Address from NodeId
+    fn get_eth_address(&self) -> Option<Address>;
```
ariqchowdhury: this doesn't seem correct. monad-crypto, monad-bls shouldn't know about eth. Can the monad-eth-* crates that use this have function to extract?

monad-system-calls/src/lib.rs
```diff
@@ -33,70 +40,249 @@ pub mod validator;
...
+            }
+        };
+
+        let transaction = TxLegacy {
+            chain_id: Some(chain_id),
+            nonce,
+            gas_price: 0,
+            gas_limit: 0,
+            to,
+            value: Default::default(),
```
ariqchowdhury: value should be 0 or whatever specific constant was defined in the spec, not default

monad-system-calls/src/lib.rs
```diff
@@ -33,70 +40,249 @@ pub mod validator;
...
+    EpochChange { new_epoch: Epoch },
+}
+
+impl SystemCallInner {
+    fn into_signed_transaction(self, chain_id: u64, nonce: u64) -> Recovered<TxEnvelope> {
+        let (to, input) = match self {
+            SystemCallInner::Reward {
+                block_author_address,
+            } => {
+                let to = TxKind::Call(STAKING_CONTRACT_ADDRESS);
```
ariqchowdhury: should we consider having nested enums for the different syscall groups (that go to the same address)? I don't feel strongly about this, it would cut down the repeated `let to = TxKind...` See if you think it makes things more readable.

monad-system-calls/src/lib.rs
```diff
@@ -33,70 +40,249 @@ pub mod validator;
...
+// System transactions related to staking
+pub(crate) const STAKING_CONTRACT_ADDRESS: Address =
+    Address::new(hex!("0x0000000000000000000000000000000000001000"));
+pub(crate) const REWARD_FUNCTION_SELECTOR: FixedBytes<4> = FixedBytes::new(hex!("0x00000064"));
+pub(crate) const SNAPSHOT_FUNCTION_SELECTOR: FixedBytes<4> = FixedBytes::new(hex!("0x00000065"));
+pub(crate) const EPOCH_CHANGE_FUNCTION_SELECTOR: FixedBytes<4> =
+    FixedBytes::new(hex!("0x00000066"));
 
-enum SystemCallInner {}
+pub struct SystemCall(SystemCallInner);
```
ariqchowdhury: what's the reason for the inner type?

monad-system-calls/src/lib.rs
```diff
@@ -33,70 +40,249 @@ pub mod validator;
...
+    let mut system_calls = Vec::new();
+
+    if proposed_seq_num.is_epoch_end(epoch_length)
+        && proposed_seq_num.get_locked_epoch(epoch_length) >= staking_activation
+    {
+        system_calls.push(SystemCall(SystemCallInner::SnapshotStakingContract));
+    }
+
+    if proposed_epoch >= staking_activation {
+        if parent_block_epoch != proposed_epoch {
```
ariqchowdhury: do we validate the proposed_epoch value somewhere?

monad-system-calls/src/lib.rs
```diff
@@ -33,70 +40,249 @@ pub mod validator;
...
+impl From<SystemTransaction> for Recovered<TxEnvelope> {
+    fn from(sys_txn: SystemTransaction) -> Self {
+        sys_txn.0.into()
     }
 }
 
-// Used by a round leader to generate system calls for the proposing block
-pub fn generate_system_calls() -> Vec<SystemCall> {
-    Vec::new()
+fn generate_system_calls(
```
ariqchowdhury: this function is going to expand in the future if new sys calls are added. it looks kind of disorganized right now and will definitely become messy if more stuff is added like this inline. Also the activation logic is mixed with the syscall logic. Maybe there's a better way of structuring this?

monad-system-calls/src/validator.rs
```diff
@@ -132,15 +204,28 @@ impl SystemTransactionValidator {
...
         for expected_sys_call in expected_sys_calls {
             let Some(sys_txn) = txns.pop_front() else {
                 return Err(SystemTransactionValidationError::MissingSystemTransaction);
             };
 
-            match Self::validate_system_transaction(expected_sys_call, &sys_txn) {
+            match Self::validate_system_transaction(expected_sys_call, sys_txn) {
                 Ok(validated_sys_txn) => {
+                    // system sender nonce must be sequential
+                    if let Some(old_nonce) = curr_sys_sender_nonce {
```
ariqchowdhury, line 150: this checks that the sender nonces are sequential, but do we check the initial nonce value is correct anywhere?

monad-system-calls/src/lib.rs
```diff
@@ -33,70 +40,249 @@ pub mod validator;
...
+            staking_activation,
+        }
     }
 
-    pub fn nonce(&self) -> u64 {
-        // TODO use actual nonce from transaction
-        0
-    }
+    // Used by a round leader to generate system calls for the proposing block
+    pub fn generate_system_transactions(
```
ariqchowdhury: It looks like there are 2 generate functions, one that the proposer uses and one that validators use. I was expecting that there would be 1 generation function (generate_sys_from_blk_hdr) and that the proposer would use the same thing. And then also related to the comment about checking the first nonce, the 2 functions here have different fn signatures; this one has next_sys_txn_nonce. I thought both paths (proposal and validation) should check that. That seems to be one of the differences between the 2 generate functions.

monad-system-calls/src/validator.rs
```diff
@@ -102,24 +122,76 @@ impl SystemTransactionValidator {
...
-        Err(SystemTransactionError::NonZeroValue)
+                Ok(SystemTransaction(
+                    SystemTransactionInner::SnapshotStakingContract(sys_txn),
+                ))
+            }
+            SystemCallInner::EpochChange { new_epoch } => {
+                if to != Some(STAKING_CONTRACT_ADDRESS) {
+                    return Err(SystemTransactionError::UnexpectedDestAddress);
+                }
+                let expected_epoch_input = new_epoch.0.to_be_bytes();
```
ariqchowdhury: nit: move the declaration closer to where its used

monad-system-calls/src/lib.rs
bamrith: the intention was to only let the crate have access to inner system call types. not required anymore

monad-system-calls/src/validator.rs
bamrith: that's checked in block policy since it needs the list of extending blocks

monad-updaters/Cargo.toml
```diff
@@ -10,6 +10,7 @@ bench = false
 
 [dependencies]
 monad-blocksync = { workspace = true }
+monad-bls = { workspace = true }
```
omegablitz: should monad-bls and monad-secp be optional deps enabled only if monad-triedb feature is activated?

most ideal would be if this could be moved out of monad-updaters and delete the monad-triedb feature, but that sounds higher lift

monad-state-backend/Cargo.toml
```diff
@@ -11,6 +11,7 @@ bench = false
 [dependencies]
 monad-crypto = { workspace = true }
 monad-eth-types = { workspace = true }
+monad-multi-sig = { workspace = true }
```
omegablitz: Where is this used? I can't find it, but if it's for a test, should be a dev-dep

monad-updaters/src/parent.rs
```diff
@@ -118,7 +118,7 @@ where
             .metrics
             .record(GAUGE_LEDGER_TOTAL_EXEC_US, || self.ledger.exec(ledger_cmds));
         guard.metrics.record(GAUGE_CHECKPOINT_TOTAL_EXEC_US, || {
```
omegablitz: rename the GAUGE

monad-chain-config/src/lib.rs
```diff
@@ -186,6 +226,18 @@ impl ChainConfig<MockChainRevision> for MockChainConfig {
         20143
     }
 
+    fn get_epoch_length(&self) -> SeqNum {
+        SeqNum::MAX
```
omegablitz: I'm sort of surprised that these don't cause any overflows - is that because we use saturating arithmetic when using these?

monad-node/examples/ledger-tail.rs
```diff
@@ -142,8 +142,12 @@ async fn main() {
                 });
 
                 let skipped_round = tc.round;
-                let skipped_leader =
-                    WeightedRoundRobin::default().get_leader(skipped_round, validators);
+                // TODO make this aware of staking_activation fork ?
```
ariqchowdhury: @omegablitz what needs to be done here for ledger-tail?

monad-system-calls/src/staking_contract.rs
```diff
@@ -0,0 +1,190 @@
...
+    pub const STAKING_CONTRACT_ADDRESS: Address =
+        Address::new(hex!("0x0000000000000000000000000000000000001000"));
+
+    // System transactions related to staking
+    pub const REWARD_FUNCTION_SELECTOR: FixedBytes<4> = FixedBytes::new(hex!("0x00000064"));
+    pub const SNAPSHOT_FUNCTION_SELECTOR: FixedBytes<4> = FixedBytes::new(hex!("0x00000065"));
+    pub const EPOCH_CHANGE_FUNCTION_SELECTOR: FixedBytes<4> = FixedBytes::new(hex!("0x00000066"));
+
+    pub const MON: u64 = 1_000_000_000_000_000_000;
+    pub const BLOCK_REWARD_MON: u64 = 1;
```
ariqchowdhury, line 43: is this just for the initial version? Is it expected to be a const in the future?

monad-triedb/src/lib.rs
```diff
@@ -462,6 +466,35 @@ impl TriedbHandle {
...
+            let secp_pubkey =
+                monad_secp::PubKey::from_bytes(validator.secp_pubkey.as_slice()).unwrap();
+            let bls_pubkey = BlsPubKey::from_bytes(validator.bls_pubkey.as_slice()).unwrap();
+            let stake = Stake::from(U256::from_be_bytes(validator.stake));
+            validator_data.push((secp_pubkey, bls_pubkey, stake));
+        }
+
+        unsafe { bindings::monad_free_valset(result) };
+
+        // TODO assert validators are unique
```
ariqchowdhury: add this check?

monad-node/examples/ledger-tail.rs
omegablitz: The fork epoch here isn't set based on ChainConfig. We could fix that properly by adding a chain-config parameter to ledger-tail so it gets populated appropriately.

monad-eth-block-validator/src/lib.rs
```diff
@@ -73,10 +76,12 @@ where
     ST: CertificateSignatureRecoverable,
     SCT: SignatureCollection<NodeIdPubKey = CertificateSignaturePubKey<ST>>,
     SBT: StateBackend<ST, SCT>,
+    CertificateSignaturePubKey<ST>: ExtractEthAddress,
 {
-    pub fn new(chain_id: u64) -> Self {
+    pub fn new(chain_id: u64, epoch_length: SeqNum, staking_activation: Epoch) -> Self {
```
omegablitz: Would it be a big diff to make this take a `ChainConfig` instead of a chain_id, epoch_length, and staking_activation? Or have the parameters be named via a struct. It would read better but not worth doing right now if it's annoying.

monad-system-calls/src/lib.rs
```diff
@@ -19,84 +19,192 @@
...
+
+#[derive(Clone, Debug)]
+pub struct SystemTransactionGenerator {
+    pub chain_id: u64,
+    pub epoch_length: SeqNum,
+    pub staking_activation: Epoch,
+}
+
+impl SystemTransactionGenerator {
+    pub fn new(chain_id: u64, epoch_length: SeqNum, staking_activation: Epoch) -> Self {
```
omegablitz: It would be easier to read if this took a ChainConfig instead, but we can defer if it's annoying. A lower lift option could be to remove the `::new()` ctor so that the parameters are named.

monad-state-backend/src/in_memory.rs
```diff
@@ -333,7 +333,7 @@ where
         &self,
         _block_num: SeqNum,
     ) -> Vec<(SCT::NodeIdPubKey, SignatureCollectionPubKeyType<SCT>, Stake)> {
-        unimplemented!()
+        unimplemented!("can't fetch validator set from InMemoryState")
```
omegablitz, line 339: Should `monad-updaters/src/val_set.rs` be updated to read from here?

Or if that's annoying, can you add a comment or a more detailed error message that validator set updates are done in-line in the mock ValSetCommand executor?

monad-node/src/main.rs
```diff
@@ -395,14 +396,18 @@ async fn run(node_state: NodeState, reload_handle: Box<dyn TracingReload>) -> Re
...
+            node_state.chain_config.get_epoch_length(),
+            node_state.chain_config.get_staking_activation(),
+        ),
         block_policy: create_block_policy(),
         state_backend,
         key: node_state.secp256k1_identity,
         certkey: node_state.bls12_381_identity,
-        epoch_length,
-        epoch_start_delay,
+        epoch_length: node_state.chain_config.get_epoch_length(),
```
omegablitz: We're also passing `chain_config` into ConsensusConfig, so the epoch_length and epoch_start delay parameters seem repetitive now. Can `epoch_length` and `epoch_start_delay` be removed from MonadStateBuilder now? This would also affect the make_state_configs mock-swarm function.

monad-eth-types/src/lib.rs
```diff
@@ -28,6 +30,22 @@ pub const BASE_FEE_PER_GAS: u64 = 50_000_000_000;
 pub type Nonce = u64;
 pub type Balance = U256;
 
+pub trait ExtractEthAddress {
```
omegablitz, line 33: This trait (and the monad-crypto/monad-secp dependency) seem sort of weird. Is this only for the EthTxPool? If so, can self_eth_address just be passed as a parameter in the EthTxPool ctor and remove NodeId from CreateProposal?

monad-chain-config/src/lib.rs
bamrith: It's because we use `epoch_length` only for two purposes. To convert a `SeqNum` to an `Epoch` or check if a `SeqNum` is a boundary block, neither of which requires an addition.

monad-system-calls/src/staking_contract.rs
bamrith: just for initial version I believe

monad-updaters/Cargo.toml
bamrith: i've made it optional for now. will move it out after this.

monad-eth-block-validator/src/lib.rs
bamrith: Made to take it `ChainConfig`

monad-system-calls/src/lib.rs
bamrith: updated. I'd prefer storing the chain config in the struct itself but the type params propagate quite high, I'll do that after

monad-state-backend/src/in_memory.rs
bamrith: added a comment. moving the validator set reads to here is on my list. but I didn't do that since it's used only for testing now. will update after.

monad-eth-types/src/lib.rs
bamrith: if it was used only in the `EthTxPool`, it think `self_eth_address` is better. but, it is also needed in the block validator where we need to retrieve the address from the block author (`NodeId<P>`) which is why this is required.

monad-node/examples/ledger-tail.rs
omegablitz: We can actually parse this straight from the node.toml, eg https://github.com/category-labs/monad-bft/blob/98a81f09a68aca662fe4a9866b1aef5dc2eb97f4/docker/devnet/monad/config/node.toml#L13

omegablitz: https://github.com/category-labs/monad-bft/blob/98a81f09a68aca662fe4a9866b1aef5dc2eb97f4/monad-node/examples/ledger-tail.rs#L105

monad-cxx/monad-execution
omegablitz, line 1: If you rebase on master now, you shouldn't need to bump the submodule

monad-mock-swarm/tests/nonces.rs
```diff
@@ -153,7 +154,7 @@ mod test {
             num_nodes,
             ValidatorSetFactory::default,
             SimpleRoundRobin::default,
-            || EthValidator::new(1337),
+            || EthValidator::new(MockChainConfig::new(&CHAIN_PARAMS)),
```
omegablitz: I assume it doesn't matter too much, but the ChainConfig here doesn't match the one in make_state_configs. I didn't check all the tests so don't know if the same mismatch exists elsewhere.

monad-system-calls/Cargo.toml
```diff
@@ -5,12 +5,15 @@ edition = "2024"
 
 [dependencies]
 monad-consensus-types = { workspace = true }
+monad-chain-config = { workspace = true }
```
omegablitz, line 7: nit: sort

monad-updaters/src/triedb_val_set.rs
```diff
@@ -16,86 +16,137 @@
...
     SCT: SignatureCollection<NodeIdPubKey = CertificateSignaturePubKey<ST>>,
 {
     validators_path: PathBuf,
 
-    next_val_data: Option<ValidatorSetDataWithEpoch<SCT>>,
-    last_emitted_val_data: Option<SeqNum>,
     epoch_length: SeqNum,
+    staking_activation: Epoch,
+
+    // sends seq num to state backend thread to initate a valset read
```
omegablitz: nit: initiate

monad-updaters/src/config_file.rs
```diff
@@ -0,0 +1,170 @@
...
+                }
+                ConfigFileCommand::ValidatorSetData { validator_set_data } => {
+                    let epoch = validator_set_data.epoch;
+                    let validators_config_file = ValidatorsConfigFile {
+                        validator_sets: vec![validator_set_data],
+                    };
+                    let validators_str = toml::to_string_pretty(&validators_config_file)
+                        .expect("failed to serialize validators");
+                    std::fs::write(
+                        format!("{}.{}", self.validators_path.to_string_lossy(), epoch.0),
```
omegablitz: What's the point of this? If I understand correctly, the validator set for epoch n is written to `validators.toml.n`. But on init, it reads from `validators.toml`. I assume I'm missing something.

monad-triedb/triedb-driver/src/triedb.cpp
```diff
@@ -427,3 +437,51 @@ uint64_t triedb_earliest_finalized_block(triedb *db)
...
+    return monad_validator_set{.valset = output, .length = length};
+}
+
+void monad_free_valset(monad_validator_set valset)
+{
+    delete[] valset.valset;
+}
+
+monad_validator_set
+monad_read_valset(triedb *db, size_t const block_num, bool get_next)
```
omegablitz: Can you call the function here instead? https://github.com/category-labs/monad/blob/69218050fbc73f46c7a1888f2c2cb9f2b2674dde/category/execution/monad/staking/read_valset.cpp#L30-L63

Would prefer there to be one implementation. Also, this allows the function to be unit-tested on execution side (cc @kkuehlz)

monad-triedb/src/lib.rs
```diff
@@ -462,6 +467,44 @@ impl TriedbHandle {
             Some(maybe_earliest_finalized_block)
         }
     }
+
+    pub fn get_next_valset(
+        &self,
+        block_num: SeqNum,
+    ) -> Vec<(monad_secp::PubKey, BlsPubKey, Stake)> {
+        let result =
+            unsafe { bindings::monad_read_valset(self.db_ptr, block_num.0 as usize, true) };
```
omegablitz: nit: `block_num.0.try_into().expect("seq_num doesn't fit in usize")`

monad-triedb/src/lib.rs
```diff
@@ -462,6 +467,44 @@ impl TriedbHandle {
...
+        let val_set_ptr = result.valset;
+        let val_set_length = result.length as usize;
```
omegablitz: nit: `let val_set_length: usize = result.length.try_into().expect("val_set_length doesn't fit in usize")`

monad-triedb/triedb-driver/include/triedb.h
```diff
@@ -81,6 +81,36 @@ uint64_t triedb_latest_verified_block(triedb *);
 // returns MAX if doesn't exist
 uint64_t triedb_earliest_finalized_block(triedb *);
 
+#pragma pack(push, 1)
+
+typedef struct monad_validator
+{
+    uint8_t secp_pubkey[33];
+    uint8_t bls_pubkey[48];
+    uint8_t stake[32];
```
omegablitz, line 91: nit: add comment that stake is big endian

monad-triedb/triedb-driver/src/triedb.cpp
```diff
@@ -427,3 +437,51 @@ uint64_t triedb_earliest_finalized_block(triedb *db)
...
+    vm::VM vm;
+    TrieDb tdb(db->db_);
+    BlockState block_state{tdb, vm};
+    Incarnation const incarnation{block_num, Incarnation::LAST_TX - 1u};
+    State state{block_state, incarnation};
+    staking::StakingContract contract(state);
+    state.add_to_balance(staking::STAKING_CA, 0);
+
+    if (!contract.vars.in_epoch_delay_period.load()) {
+        get_next = false;
```
omegablitz: I don't like how this silently floors. For example, if the boundary block is block n, and monad_read_valset(n-1, true) is called.

Either returning a result type or asserting here would make more sense IMO. Can you make a ticket and tag @kkuehlz? We don't need to block merging this in a meanwhile.

monad-eth-txpool/src/pool/mod.rs
```diff
@@ -405,9 +424,47 @@ where
...
+        proposed_seq_num: SeqNum,
+        proposed_epoch: Epoch,
+        block_author: Address,
+        extending_blocks: &Vec<&EthValidatedBlock<ST, SCT>>,
+        block_policy: &EthBlockPolicy<ST, SCT>,
+        state_backend: &SBT,
+    ) -> Result<Vec<Recovered<TxEnvelope>>, StateBackendError> {
+        // TODO this should be inside SystemTransactionGenerator to prevent
+        // exposing SYSTEM_SENDER_ETH_ADDRESS outside the crate
+        let next_system_txn_nonce = *block_policy
```
omegablitz, line 461: This should probably be stored in memory eventually. We don't need to change this right now.

monad-eth-txpool/src/pool/mod.rs
```diff
@@ -405,9 +424,47 @@ where
...
+            .get(&SYSTEM_SENDER_ETH_ADDRESS)
+            .unwrap();
+
+        let sys_txns = self.sys_tx_generator.generate_system_transactions(
+            proposed_seq_num,
+            proposed_epoch,
+            extending_blocks
+                .last()
+                .map(|parent_block| parent_block.get_epoch())
+                .unwrap_or(QuorumCertificate::<SCT>::genesis_qc().get_epoch()),
```
omegablitz: This looks suspicious to me. Extending blocks can be empty if proposed_seq_num = last_finalized_seq_num + 1

monad-updaters/src/triedb_val_set.rs
```diff
@@ -16,86 +16,137 @@
...
-                    .clone(),
-            });
-            self.last_emitted_val_data = Some(seq_num);
-        }
+impl<ST, SCT> ValSetUpdater<ST, SCT>
+where
+    ST: CertificateSignatureRecoverable,
+    SCT: SignatureCollection<NodeIdPubKey = CertificateSignaturePubKey<ST>>,
+{
+    fn valset_update_from_validators_toml(&mut self, seq_num: SeqNum) {
```
omegablitz, line 132: nit: you can move this into the upper `impl ValSetUpdater<SecpSignature, BlsSignatureCollection<PubKey>>`

monad-updaters/src/triedb_val_set.rs
```diff
@@ -16,86 +16,137 @@
...
+    ) -> Self
+    where
+        SBT: StateBackend<SecpSignature, BlsSignatureCollection<PubKey>> + Send + 'static,
+    {
+        let (valset_send, valset_recv) = tokio::sync::mpsc::unbounded_channel();
+        let (seq_num_send, seq_num_recv) = std::sync::mpsc::channel();
+
+        let next_valset_send = valset_send.clone();
+        std::thread::spawn(move || loop {
+            let seq_num_to_read = seq_num_recv.recv().expect("channel never closed");
```
omegablitz: nit: assert `seq_num.is_epoch_end(self.epoch_length)`

monad-updaters/src/triedb_val_set.rs
```diff
@@ -16,86 +16,137 @@
...
+            // read the validator set
+            while state_backend
+                .raw_read_latest_finalized_block()
+                .is_none_or(|latest_finalized| latest_finalized < seq_num_to_read)
+            {
+                info!(?seq_num_to_read, "next valset not ready, sleeping");
+                std::thread::sleep(Duration::from_millis(500));
+            }
+
+            let next_valset = state_backend.read_next_valset(seq_num_to_read);
```
omegablitz: /// [0, epoch_length-1] -> Epoch 1
/// [epoch_length, (2 * epoch_length)-1] -> Epoch 2

We're calling read_next_valset(get_next: true) on epoch_length-1, which is the last block in Epoch 1. I would have thought that epoch_start_delay started in block epoch_length.

Is the state at`epoch_length-1` the state after executing `epoch_length-1`?
Maybe that makes sense because you start `epoch_start_delay` after executing the last block of the previous epoch, but before the first block of the next epoch?


Is this understanding correct? If so, I think it'd be more intuitive and obviously correct if this operated in terms of epochs instead of seq_nums, and if get_next didn't silently floor.

monad-updaters/src/triedb_val_set.rs
```diff
@@ -132,22 +195,25 @@ where
...
-                    self.valset_update(seq_num);
-                    wake = true;
+                    if seq_num.is_epoch_end(self.epoch_length) {
+                        if !self.valset_recv.is_empty() {
+                            error!("Validator set data is not consumed");
+                        }
+
+                        if seq_num.get_locked_epoch(self.epoch_length) >= self.staking_activation {
+                            self.seq_num_send
+                                .send(seq_num)
```
omegablitz: IMO this should be operating in terms of epochs. This is confusing, because seq_num is the last block in `seq_num.get_locked_epoch(self.epoch_length) - 1`.

omegablitz (commented):
I still need to go over monad-system-calls, but finished going through the rest. Going over monad-system-calls now.

monad-system-calls/src/lib.rs
```diff
@@ -19,84 +19,197 @@
...
+}
+
+enum SystemCall {
+    StakingContractCall(StakingContractCall),
+}
 
 impl SystemCall {
-    fn get_dest_address(&self) -> Address {
-        Address::new([0_u8; 20])
+    pub fn is_restricted_system_call(txn: &Recovered<TxEnvelope>) -> bool {
```
omegablitz, line 60: Can you add comments on each of the pub validation functions in this crate on where they're expected to be called?

omegablitz (commented):
finished going over everything

monad-system-calls/src/staking_contract.rs
```diff
@@ -0,0 +1,190 @@
...
+                input[0..4].copy_from_slice(Self::EPOCH_CHANGE_FUNCTION_SELECTOR.as_slice());
+                input[4..12].copy_from_slice(&new_epoch.0.to_be_bytes());
+                transaction.input = input.into();
+
+                StakingContractTransaction::EpochChange(sign_with_system_sender(transaction))
+            }
+        }
+    }
+
+    pub fn validate_system_transaction_input(
```
omegablitz, line 96: I'm worried we forgot or will forget to verify stuff in here. Would it be possible to generate the set of expected transactions as if we're the proposer, and then compare the system transactions in the block to the set of transactions we generate internally?

omegablitz: For example, check that `sys_txn` and `self.into_signed_transaction` are byte equivalent

monad-triedb/triedb-driver/src/triedb.cpp
bamrith: it might be better to have the function here instead of execution then. else, we'd need to write custom types (exactly as what exists in execution) just to generate rust bindings. and i think the wrapper function we write here needs to copy the set from execution type to our bindings type. (or maybe there's a way to generate bindings for execution code directly ? i didn't think there was)

monad-system-calls/src/staking_contract.rs
bamrith: I'd prefer doing that too. It's written this way because we don't have the correct nonce for the system sender to generate the exact system transactions (don't have the extending blocks since this is static validation). And then we separately check that the nonce is correct in block policy.

monad-updaters/src/config_file.rs
bamrith: added writing the last two validator sets to the file (if over staking activation epoch)

monad-eth-txpool/src/pool/mod.rs
bamrith: made a note. will fix later

omegablitz (approved):

PR #2183, opened 2025-08-18
Eagerly push all groups to primary raptorcast instance

fabiotdf:
This is a more generic fix that replaces PR [2126](https://github.com/category-labs/monad-bft/pull/2126) 
Here, all groups are always eagerly sent to primary. The primary instance will then queue these groups and pop them off on demand. This simplifies the logic a bit and does not rely on the `is_receiving_proposal` mechanism, and should work in more corner cases than the original PR

### DIFF
--- a/monad-raptorcast/src/util.rs
+++ b/monad-raptorcast/src/util.rs
@@ -14,7 +14,8 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 use std::{
-    collections::{BTreeMap, HashSet},
+    cmp::Ordering,
+    collections::{BTreeMap, BinaryHeap, HashSet},
     fmt,
 };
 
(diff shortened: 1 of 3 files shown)


eerkaijun (dismissed):

omegablitz (dismissed):

monad-raptorcast/src/util.rs
```diff
@@ -575,7 +604,14 @@ where
     ) -> Option<GroupIterator<ST>> {
         let maybe_group = match broadcast_mode {
             BroadcastMode::Primary => self.validator_map.get(&msg_epoch),
-            BroadcastMode::Secondary => self.fullnode_map.get(msg_author),
+            BroadcastMode::Secondary => {
+                let maybe_group_queue = self.fullnode_map.get(msg_author);
+                if let Some(group_queue) = maybe_group_queue {
+                    group_queue.peek() // Take earliest among all future groups for msg_author
```
eerkaijun, line 610: is this always the current group? i.e. do we immediately delete a group once it expired?

fabiotdf: We do immediately delete a group once it is expired.

It is possible that group could be in the future. We won't be receiving anything from that future group before this group's publisher starts to actually send chunks to us (and that should happen only when the validator's current_round == group.start_round)

This is a priority queue ordered by start_round, and validators always schedule groups in chronological order, so the front entry in this queue is always the next to be used when we receive a chunk from that validator.

PR #2190, opened 2025-08-19
Pace forwarded tx sender recovery

andr-dev:
Previously, forwarded txs were recovered in the txpol upon receiving the `InsertForwardedTxs` command. Under high forwarding pressure, the number of commands can grow unboundedly leading to txpool starvation from recovering the txs. This change delays sender recovery until a chunk of txs is being inserted to the txpool, effectively pacing the sender recovery and avoiding starvation.

Additionally, the max size of queued ingress txs was reduced significantly to disallow large bursts of txs which cannot be processed within a round.


monad-eth-txpool-executor/src/forward.rs
```diff
@@ -131,7 +128,7 @@ impl EthTxPoolForwardingManager {
 }
 
 impl EthTxPoolForwardingManagerProjected<'_> {
-    pub fn add_ingress_txs(&mut self, txs: Vec<Recovered<TxEnvelope>>) {
+    pub fn add_ingress_txs(&mut self, txs: Vec<TxEnvelope>) {
         let Self {
             ingress,
             ingress_waker,
```
omegablitz, line 135: nit: add a warning with the number of dropped transactions if `txs` isn't fully drained into `ingress`

monad-eth-txpool-executor/src/lib.rs
```diff
@@ -536,13 +518,32 @@ where
...
+                let (recovered_txs, dropped_txs): (Vec<_>, Vec<_>) =
+                    forwarded_txs.into_par_iter().partition_map(|tx| {
+                        let _span = trace_span!("txpool: recover signer").entered();
+                        match tx.secp256k1_recover() {
+                            Ok(signer) => {
+                                rayon::iter::Either::Left(Recovered::new_unchecked(tx, signer))
+                            }
+                            Err(_) => rayon::iter::Either::Right(EthTxPoolEvent::Drop {
+                                tx_hash: *tx.tx_hash(),
+                                reason: EthTxPoolDropReason::InvalidSignature,
```
omegablitz, line 529: nit: add comment clarifying that there's no corresponding insertion event for this yet

monad-eth-txpool-executor/src/forward.rs
omegablitz: or rather than a warning, perhaps a `debug`

monad-eth-txpool-executor/src/lib.rs
```diff
@@ -536,13 +518,32 @@ where
 
         while let Poll::Ready(forwarded_txs) = forwarding_manager.as_mut().poll_ingress(cx) {
```
omegablitz, line 516: Should this be moved after `preload_manager.as_mut().poll_requests(cx)`? Isn't preloading more important?

omegablitz: There's also the promote_pending after this, but not sure how important that is

monad-eth-txpool-executor/src/lib.rs
andr-dev: > nit: add comment clarifying that there's no corresponding insertion event for this yet

This is normal and happens naturally within txpool as well, it isn't exactly an edge case.

monad-eth-txpool-executor/src/forward.rs
```diff
@@ -131,17 +129,24 @@ impl EthTxPoolForwardingManager {
...
-        ingress.extend(
-            txs.into_iter()
-                .take(INGRESS_MAX_SIZE.saturating_sub(ingress.len())),
-        );
+        let capacity_remaining = INGRESS_MAX_SIZE.saturating_sub(ingress.len());
+
+        if txs.len() > capacity_remaining {
+            debug!(
+                capacity =? INGRESS_MAX_SIZE,
+                ?capacity_remaining,
```
omegablitz, line 146: super nit: can you add the # dropped = `txs.len() - capacity_remaining` as well

monad-eth-txpool-executor/src/lib.rs
omegablitz: Not a real issue, because the loop only loops once in practice: https://github.com/category-labs/monad-bft/pull/2190/files#diff-16c58c6d5efc6627c5a79a37f60ad9e4eb1d520f3e343d611c9cdaea2f88cfb2L554

monad-eth-txpool-executor/src/lib.rs
```diff
@@ -536,13 +515,32 @@ where
...
             let _span = debug_span!("forwarded txs", len = forwarded_txs.len()).entered();
+
+            let recovered_txs = {
+                let (recovered_txs, dropped_txs): (Vec<_>, Vec<_>) =
+                    forwarded_txs.into_par_iter().partition_map(|tx| {
+                        let _span = trace_span!("txpool: recover signer").entered();
```
omegablitz: super nit: name this span something different than the other one

omegablitz (approved):
LGTM, to be merged after repro'ing performance degradation before this change, and observing it fixed after this change.

ariqchowdhury (approved):

omegablitz (approved):

PR #2191, opened 2025-08-19
RPC: Treat Archive NotFound the same as TrieDB NotFound

JoeHowarth:
- Use try_get_x methods instead of get_x on ArchiveReader
- Improve chainstate -> handler mappings for readability via extension traits on error types
- Fix some warns in RPC

fixed https://github.com/category-labs/category-internal/issues/1845

### DIFF
--- a/monad-rpc/src/chainstate/mod.rs
+++ b/monad-rpc/src/chainstate/mod.rs
@@ -44,7 +44,7 @@ use crate::{
         block::block_receipts,
         txn::{parse_tx_receipt, FilterError},
     },
-    jsonrpc::{JsonRpcError, JsonRpcResult},
+    jsonrpc::{ArchiveErrorExt, JsonRpcError, JsonRpcResult},
 };
 
 pub mod buffer;
@@ -59,7 +59,16 @@ pub struct ChainState<T> {
 #[derive(Debug)]
 pub enum ChainStateError {
     Triedb(String),
+    Archive(String),
     ResourceNotFound,
+}
+
+impl From<monad_archive::prelude::Report> for ChainStateError {
+    fn from(e: monad_archive::prelude::Report) -> Self {
+        // Log with debug to get error chain, but return only top level error in response
+        error!("Archive Error: {e:?}");
+        ChainStateError::Archive(e.to_string())
+    }
 }
 
 pub fn get_block_key_from_tag<T: Triedb>(triedb_env: &T, tag: BlockTags) -> BlockKey {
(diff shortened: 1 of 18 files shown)


JoeHowarth:
Stupid copilot autoreview - changed these myself using "Files Changed" tab then saw the exact same comments here.

On second thought, maybe that would have been faster than me doing it manually?

monad-rpc/src/handlers/eth/block.rs
```diff
@@ -151,14 +143,10 @@ pub async fn monad_eth_getBlockTransactionCountByHash<T: Triedb>(
     params: MonadEthGetBlockTransactionCountByHashParams,
 ) -> JsonRpcResult<Option<String>> {
     trace!("monad_eth_getBlockTransactionCountByHash: {params:?}");
-    match chain_state
+    chain_state
         .get_block(BlockTagOrHash::Hash(params.block_hash), true)
         .await
-    {
-        Ok(block) => Ok(Some(format!("0x{:x}", block.transactions.len()))),
-        Err(ChainStateError::ResourceNotFound) => Ok(None),
```
eerkaijun, line 159: do we still ever return null?

philipglazman (dismissed):

monad-rpc/src/chainstate/mod.rs
```diff
@@ -59,9 +59,18 @@ pub struct ChainState<T> {
...
     Triedb(String),
+    Archive(String),
     ResourceNotFound,
 }
 
+impl From<monad_archive::prelude::Report> for ChainStateError {
+    fn from(e: monad_archive::prelude::Report) -> Self {
+        // Log with debug to get error chain, but return only top level error in response
+        error!("Archive Error: {e:?}");
+        ChainStateError::Archive(e.to_string())
```
philipglazman, line 70: Is there any concern about unsanitized errors from the archiver ? (e.g. end user seeing url of aws or mongo instances)

JoeHowarth: I added test assertions that all `get` methods on ArchiveReader do not contain any details of the underlying store when formatted with Display. Full error chain is present on Debug which is good to log internally.

monad-rpc/src/handlers/eth/block.rs
JoeHowarth: Yeah, same behavior, just reduced the verbosity by packaging the pattern into a combinator.

I renamed the combinator from `map_ok` to `map_present_and_no_err` for clarity. Basically the closure only is applied for the Ok(Some(T)) case

eerkaijun (approved):

PR #2203, opened 2025-08-21
reserve balance consecutive blocks unit tests

eerkaijun:


monad-eth-block-policy/src/lib.rs
```diff
@@ -508,6 +508,7 @@ where
             block_gas_cost = block_txn_fees
                 .max_gas_cost
                 .saturating_add(block_txn_fees.first_txn_gas);
+            account_balance.block_seqnum_of_latest_txn = self.block_seq_num;
```
mikety: I think this update should be done after validation  on line 513 - otherwise it will be looking like this block has been added even if validation has failed

PR #2207, opened 2025-08-21
monad-tfm: dynamic base fee

michael-yxchen:


monad-tfm/src/base_fee.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    let factor = U256::from(factor);
+    let numerator = I256::try_from(numerator).expect("i64 always fit in i256");
+    let denominator = U256::from(denominator);
+
+    let mut i = U256::from(1);
+    let mut output = I256::ZERO;
+    let mut numerator_accum =
+        I256::try_from(factor * denominator).expect("product of two 64 bit numbers fits in i256");
+    while !numerator_accum.is_zero() {
```
ezm0n, line 114: maybe we need to add a robustness check that we do not loop more than a bounded number 
of times like 256 ?

monad-tfm/src/base_fee.rs
```diff
@@ -0,0 +1,145 @@
...
+        output += numerator_accum;
+
+        // denominator is not zero, asserted at the beginning of the function
+        // denominator * i is never (-1) so this division never overflows
+        numerator_accum = (numerator_accum * numerator)
```
ezm0n, line 119: use checked multiplication to handle potential overflow gracefully ?

monad-tfm/src/base_fee.rs
```diff
@@ -0,0 +1,145 @@
...
+    let factor = U256::from(factor);
+    let numerator = I256::try_from(numerator).expect("i64 always fit in i256");
+    let denominator = U256::from(denominator);
+
+    let mut i = U256::from(1);
+    let mut output = I256::ZERO;
+    let mut numerator_accum =
+        I256::try_from(factor * denominator).expect("product of two 64 bit numbers fits in i256");
+    while !numerator_accum.is_zero() {
+        output += numerator_accum;
```
ezm0n, line 115: use checked_add to handle possible overflow gracefully ?

monad-tfm/src/base_fee.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    // Convert trend to unsigned representation (2's complement bit pattern)
+    let trend_u64 = trend.cast_unsigned();
+
+    (base_fee, trend_u64, moment)
+}
+
+/// https://eips.ethereum.org/EIPS/eip-4844#helpers)
+#[inline]
+fn fake_exponential(factor: u64, numerator: i64, denominator: u64) -> u64 {
```
ezm0n, line 98: please note that due to the way the base_fee is implemented we currently only call fake_exponential
with values of the inputs such that |numerator| < denominator. Therefore we are guaranteed that the inner loop will. 
not overflow.
I am wondering if we would want to add this as a guard for fake_exponential in order to clearly
signal to the user that we wish to use this function to calculate factor * exp(x) only in the domain that
|x| <= 1 ?

michael-yxchen: added assertion

monad-dataplane/tests/address_family_mismatch.rs
michael-yxchen, line 1: changed ports because those are occupied on dev servers

ipv6 is not supported by dataplane or available on dev servers

monad-dataplane/tests/tests.rs
michael-yxchen, line 1: port occupied by services

monad-dataplane/tests/address_family_mismatch.rs
```diff
@@ -21,9 +21,9 @@ use tracing::debug;
 /// 1_000 = 1 Gbps, 10_000 = 10 Gbps
 const UP_BANDWIDTH_MBPS: u64 = 1_000;
 
-const BIND_ADDRS: [&str; 3] = ["0.0.0.0:9100", "127.0.0.1:9101", "[::1]:9102"];
+const BIND_ADDRS: [&str; 2] = ["0.0.0.0:19100", "127.0.0.1:19101"];
```
ariqchowdhury: why is this change here?

monad-eth-block-policy/src/lib.rs
```diff
@@ -274,6 +279,12 @@ where
             assert_eq!(self.blocks.len(), self.size);
         }
 
+        let block_gas_usage = block
+            .validated_txns
+            .iter()
+            .map(|txn| txn.gas_limit())
+            .sum::<u64>();
```
ariqchowdhury: this could be a helper function in `EthValidatedBlock`, `get_total_gas`

monad-dataplane/tests/address_family_mismatch.rs
michael-yxchen: https://github.com/category-labs/monad-bft/pull/2207#discussion_r2299235489

these values cause unit tests to fail on dev machines because ports are occupied (specifically tcp:9100) and ipv6 is not supported

michael-yxchen: i can split that into a different PR if preferred

ariqchowdhury: yeah, can we make a separate PR for those please

michael-yxchen:
will create a struct for base fee fields later: https://github.com/category-labs/category-internal/issues/1869

ariqchowdhury (approved):

PR #2208, opened 2025-08-22
Dynamic secondary raptorcast role switch

eerkaijun:
Allow role switching from full node -> validator if the node joins the validator set in the next epoch. Approach is that upon epoch increment, if a node detects that it is now in the active validator set, it will change its role from to validator in peer discovery and secondary raptorcast. Tested on stressnet by making a full node joins the validator set through staking. 

Role switching validator -> full node not included in this PR, that requires some additional changes for the node that falls out of the validator set to be able to bump its epoch.


monad-router-multi/src/lib.rs
```diff
@@ -140,6 +274,44 @@ where
...
+                        } else if !is_validator_before_epoch_increment
+                            && is_validator_after_epoch_increment
+                        {
+                            // full node promoted to validator, update role to be Publisher
+                            self.update_role(SecondaryRaptorCastModeConfig::Publisher);
+                        }
+
+                        // clean old epoch validators
+                        self.epoch_validators
+                            .retain(|epoch, _| *epoch + Epoch(1) >= self.current_epoch);
```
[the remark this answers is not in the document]
eerkaijun: isn't this the same?

monad-peer-discovery/src/discovery.rs
```diff
@@ -1218,11 +1228,17 @@ where
             self.current_epoch = epoch;
 
             // if a full node gets promoted to a validator, advertise name records to all validators in the current and next epoch
+            // default to ValidatorPublisher
             if (self.self_role == PeerDiscoveryRole::FullNodeNone
                 || self.self_role == PeerDiscoveryRole::FullNodeClient)
-                && self.check_validator_membership(&self.self_id)
+                && self.check_current_epoch_validator(&self.self_id)
```
ariqchowdhury, line 1233: if the node is a full_node in epoch N and will become a validator in N+1, will this check start advertising name records in N or N+1? Ideally the name records are shared in advance

monad-peer-discovery/src/discovery.rs
```diff
@@ -1218,11 +1228,17 @@ where
...
             {
-                self.self_role = PeerDiscoveryRole::ValidatorNone;
+                debug!(?epoch, "full node promoted to validator");
+                self.self_role = PeerDiscoveryRole::ValidatorPublisher;
+                // clear connection info
+                self.participation_info.iter_mut().for_each(|(_, info)| {
+                    info.status = SecondaryRaptorcastConnectionStatus::None;
+                });
```
ariqchowdhury: should we make a helper function to clear secondary raptorcast status? Will make it easier to verify that everything that needs to be cleaned up is contained in one place

monad-peer-discovery/src/discovery.rs
eerkaijun: good point, i think we can start sharing name records at epoch boundary block (when validator set is updated). will update

monad-peer-discovery/src/discovery.rs

monad-peer-discovery/src/discovery.rs
eerkaijun: updated, done in update_validator_set

omegablitz (approved):

PR #2215, opened 2025-08-22
Feature: round up RaptorCast chunk assignments

tklenze:
**Round up the number of chunks assigned to each validator in the first phase of validator-to-validator RaptorCast.**  
This change is required to guarantee Byzantine Fault Tolerance. It implements the [publicly documented behavior](https://www.category.xyz/blogs/raptorcast-designing-a-messaging-layer) of RaptorCast, ensuring documentation and implementation are aligned.

This feature only affects the leader when sending out a block proposal using RaptorCast. The leader operates as follows:

- First, we compute the number of packets (`num_packets`) without rounding up. This is essentially the payload size divided by the payload per chunk times the redundancy factor.
- For each validator, we determine their _obligation_: the number of chunks they should be assigned, calculated as `num_packets * stake / total_stake`. This value is a float.
- Each validator receives a number of chunks equal to their obligation, rounded up. We distinguish between _initial chunks_ (the obligation rounded down, denoted `ic[i]`) and a _rounding chunk_ (each validator receives exactly one). While this distinction is not currently important, it may become relevant in the future.
- The new total number of packets is the sum of all `ic[i]` plus the number of validators `n` (since each receives one rounding chunk).

As a safeguard, we check that the rounded-up total `rounded_up_num_packets` falls between the original `num_packets` and `num_packets + n`. This should always be true; if not, we issue a warning. In case `num_packets` is larger, we use that value instead.

### Other minor features

- The validator stake map is shuffled deterministically for chunk assignment using a seed. This does not yet affect chunk assignment, but will likely be important in future updates.

### Performance impact
The solution of rounding up requires sending up to `n` more chunks than necessary -- each one of which will get re-broadcasted by a validator. Each validator sends and receives `~n` more chunks than required. For small messages, the overhead is very large. As such, it is only a temporary solution; a more permanent solution is planned.

We simulated the overhead created by rounding up. The following graph shows the additional bandwidth consumed at validators due to rounding up, for various values of `num_packets` and numbers of validators `n`. 4740 chunks represents a full block, whereas 474 chunks present a block that is only 10% full. (Note that the graph says "current RaptorCast", but this is actually the solution proposed in this PR.)

For the leader's block proposal message, the additional overhead is likely not prohibitive for small-medium network sizes (e.g., `n = 200`).
image

### Required actions before merge
- Before deploying this feature, we must ensure that this only affects the leader's block proposal message, and not any other messages, in particular, if they are sent more frequently than block proposals.
- Testing could be extended. Manual testing has been done locally on monad-testground, with 5-8 validators, and various payload sizes (yielding 6-100 packets). These tests were all successful, i.e., I confirmed manually that every validator received the expected number of packets. No testing has been done outside of these parameters yet.
- Need to think about whether this should affect the `encoded_symbol_capacity` in the decoder. Right now, this capacity is calculated as `max(app_message_len / symbol_len, SOURCE_SYMBOLS_MIN) * MAX_REDUNDANCY`. Due to rounding up, for small messages, more chunks than that could be created, and they would not be added to the decoder state. I think this might not be an issue, because (i) those chunks are not required for decoding and (ii) we still broadcast those chunks if they are assigned to us (`self_hash == parsed_message.recipient_hash`), even if we don't end up storing them.

Closes https://github.com/category-labs/category-internal/issues/898 (but a follow-up issue is needed, to address the inefficiency of rounding up).

### DIFF
--- a/monad-raptorcast/src/udp.rs
+++ b/monad-raptorcast/src/udp.rs
@@ -20,6 +20,7 @@ use std::{
     ops::Range,
 };
 
+use alloy_primitives::U256;
 use bitvec::prelude::*;
 use bytes::{Bytes, BytesMut};
 use itertools::Itertools;
@@ -35,7 +36,7 @@ use monad_dataplane::RecvUdpMsg;
 use monad_merkle::{MerkleHash, MerkleProof, MerkleTree};
 use monad_raptor::{ManagedDecoder, SOURCE_SYMBOLS_MIN};
 use monad_types::{Epoch, NodeId, Stake};
-use rand::seq::SliceRandom;
+use rand::{rngs::StdRng, seq::SliceRandom, SeedableRng};
 use tracing::warn;
 
 use crate::{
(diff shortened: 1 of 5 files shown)


babak1369:
If we have 100 equally staked validators and 100 chunks. and say  redundancy 3, does it set total n_chunks to  400 chunks ? 
Probably not a big issue , but I would add a check : if ic[i] is integer then we don't need to add +1.

monad-raptorcast/src/udp.rs
```diff
@@ -748,53 +921,50 @@ where
...
-                }
+            let mut trace_string = String::new();
+            for (node_id, num_chunks) in validator_num_chunks.iter() {
+                let end_idx: usize = start_idx + num_chunks;
                 if let Some(addr) = known_addresses.get(node_id) {
                     outbound_gso_idx.push((
                         *addr,
                         start_idx * segment_size as usize..end_idx * segment_size as usize,
                     ));
+                    trace_string.push_str(&format!("{:?}: {} chunks, ", node_id, num_chunks));
```
xinyuan-dev, line 945: we probably do not want to leave have the tracing code run conditionlessly. how about put a `trace` statement here instead?

monad-raptorcast/src/udp.rs
```diff
@@ -463,6 +464,170 @@ where
...
+        .map(|(_, stake)| -> U256 { stake.0 })
+        .sum();
+
+    let num_receivers = validator_stake_map_without_leader.len();
+
+    // Compute obligations: num_packets_without_rounding * stake / total_weight for each validator.
+    // The obligations are scaled by the basis points (bps) to avoid floating point arithmetic on
+    // the large values of stake and total_weight. Only at the very end we convert to a float.
+    // In the future, we might want to avoid floats alltogether.
+    let unit_bias: f64 = 10_000_000.0;
```
xinyuan-dev, line 575: is this bias necessary? `stake / total_stake` is a fractional number between 0 and 1. In IEEE floating point representation, there is roughly same number of representable real numbers between 0 and 1, versus between 1 and infinity. ([source](https://chadnauseam.com/coding/random/floating-points-between-zero-and-one)) so there should be enough precision if `stake / total_stake` is precise enough.

omegablitz:
https://github.com/category-labs/monad-bft/pull/2483

PR #2218, opened 2025-08-23
eip7702

eerkaijun:
rebased on top of test harness

### DIFF
--- a/monad-consensus-types/src/block.rs
+++ b/monad-consensus-types/src/block.rs
@@ -175,6 +175,7 @@ pub enum BlockPolicyError {
     ExecutionResultMismatch,
     BaseFeeError,
     BlockPolicyBlockValidatorError(BlockPolicyBlockValidatorError),
+    Eip7702Error,
 }
 
 impl From<StateBackendError> for BlockPolicyError {
@@ -189,6 +190,7 @@ pub struct AccountBalanceState {
     pub remaining_reserve_balance: Balance,
     pub max_reserve_balance: Balance,
     pub block_seqnum_of_latest_txn: SeqNum,
+    pub is_delegated: bool,
 }
 
 impl AccountBalanceState {
@@ -198,6 +200,7 @@ impl AccountBalanceState {
             remaining_reserve_balance: Balance::ZERO,
             max_reserve_balance,
             block_seqnum_of_latest_txn: GENESIS_SEQ_NUM,
+            is_delegated: false,
         }
     }
 }
@@ -217,6 +220,7 @@ pub struct TxnFee {
     pub first_txn_gas: Balance,
     pub max_gas_cost: Balance,
     pub max_txn_cost: Balance, // Used for pre TFM validation
+    pub is_delegated: bool,
 }
 
 pub type TxnFees = BTreeMap<Address, TxnFee>;
--- a/monad-eth-block-policy/src/lib.rs
+++ b/monad-eth-block-policy/src/lib.rs
@@ -23,6 +23,7 @@ use alloy_consensus::{
     transaction::{Recovered, Transaction},
     TxEnvelope,
 };
+use alloy_eips::eip7702::SignedAuthorization;
 use alloy_primitives::{Address, TxHash, U256};
 use itertools::Itertools;
 use monad_chain_config::{revision::ChainRevision, ChainConfig};
@@ -337,16 +338,16 @@ where
 
 fn is_possibly_emptying_transaction(
     block_seq_num_of_curr_txn: SeqNum,
-    block_seqnum_of_latest_txn: SeqNum,
+    balance_state: &AccountBalanceState,
     execution_delay: SeqNum,
 ) -> bool {
     // txn T is emptying if there is no "prior txn" i.e. a txn from the same sender sent from block P so that P >= block_number(T) - k + 1.
     let blocks_since_latest_txn = SeqNum(
         block_seq_num_of_curr_txn
             .0
-            .saturating_sub(block_seqnum_of_latest_txn.0),
+            .saturating_sub(balance_state.block_seqnum_of_latest_txn.0),
     );
-    blocks_since_latest_txn > execution_delay - SeqNum(1)
+    !balance_state.is_delegated && blocks_since_latest_txn > execution_delay - SeqNum(1)
 }
 
 impl<CRT> BlockPolicyBlockValidator<CRT> for EthBlockPolicyBlockValidator<CRT>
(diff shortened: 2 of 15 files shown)


monad-eth-block-policy/src/lib.rs
```diff
@@ -340,13 +341,15 @@ pub struct EthBlockPolicyBlockValidator {
...
 ) -> bool {
     // txn T is emptying if there is no "prior txn" i.e. a txn from the same sender sent from block P so that P >= block_number(T) - k + 1.
-    let blocks_since_latest_txn =
-        block_seq_num_of_curr_txn.max(block_seqnum_of_latest_txn) - block_seqnum_of_latest_txn;
-    blocks_since_latest_txn > execution_delay - SeqNum(1)
+    let blocks_since_latest_txn = block_seq_num_of_curr_txn
+        .max(balance_state.block_seqnum_of_latest_txn)
+        - balance_state.block_seqnum_of_latest_txn;
+
+    !balance_state.is_delegated && blocks_since_latest_txn > (execution_delay - SeqNum(1))
```
eerkaijun: I think this logic is incomplete -- there's a few more checks in the TFM spec

monad-eth-block-validator/src/lib.rs
```diff
@@ -188,8 +188,46 @@ where
...
+                    is_delegated: false,
                 });
-            debug!(seq_num = ?header.seq_num, address = ?eth_txn.signer(), nonce = ?eth_txn.nonce(), ?txn_fee_entry, "TxnFeeEntry");
+            debug!(?txn_fee_entry, address = ?eth_txn.signer(), "TxnFeeEntry");
+
+            if eth_txn.is_eip7702() {
+                if let Some(auth_list) = eth_txn.authorization_list() {
+                    for (result, nonce, code_address, chain_id) in auth_list
+                        .iter()
+                        .map(|a| (a.recover_authority(), a.nonce(), a.address(), a.chain_id()))
```
eerkaijun: we are doing recover_authority in monad-eth-block-validator and monad-eth-block-policy. i think in monad-eth-block-policy, we only need to check that each starting nonce of the each addresses in the incoming block is +1 from `block_policy.get_account_base_nonces`. in that case, we wouldn't need recover_authority in monad-eth-block-policy

monad-eth-block-policy/src/lib.rs
eerkaijun: actually i might be wrong, i'll double check

eerkaijun: we also need to set is_delegated in compute_account_base_balances

monad-eth-txpool/src/pool/tracked/mod.rs
```diff
@@ -194,7 +194,30 @@ where
...
+                            } else {
+                                event_tracker
+                                    .drop(txn.hash(), EthTxPoolDropReason::InvalidSignature);
+                            }
+                        }
+                    }
+                }
+                txn.signer()
+            })
+            .collect_vec();
```
andr-dev: Authority recovery should not be performed in the `create_proposal` hot path. Additionally, the use of `event_tracker.drop` is incorrect here since the transaction has not been dropped.

monad-eth-txpool/src/pool/tracked/tx_heap.rs
```diff
@@ -106,6 +106,17 @@ impl<'a> TrackedTxHeap<'a> {
...
+    pub fn txs<'s>(&'s self) -> impl Iterator<Item = &'a ValidEthTransaction> + 's {
+        self.heap.iter().map(
+            |OrderedTxGroup {
+                 tx,
+                 virtual_time: _,
+                 address: _,
+                 queued: _,
+             }| *tx,
+        )
+    }
```
andr-dev: This method is likely incorrect since it produces only the next (by nonce) tx for each sender, not all the txs that are available from that sender for proposal inclusion.

monad-eth-block-policy/src/lib.rs
eerkaijun: already set

eerkaijun:
keeping track of some outstanding performance considerations https://github.com/category-labs/category-internal/issues/1864

eerkaijun:
txpool nonce tracking needs to keep also take into account authorization nonces, andre taking a look

monad-eth-txpool/src/pool/tracked/mod.rs
```diff
@@ -199,7 +199,9 @@ where
         let sequencer = ProposalSequencer::new(&self.txs, &extending_blocks, base_fee);
         let sequencer_len = sequencer.len();
 
-        let (account_balances, account_balance_lookups) = {
+        // TODO: enrich with delegated EOAs
```
eerkaijun: to be populated by andre

monad-eth-txpool/src/pool/tracked/mod.rs
```diff
@@ -220,6 +222,15 @@ where
...
         };
 
+        // for signer in &authority_addresses {
+        //     let account_balance = account_balances
+        //         .get_mut(signer)
+        //         .expect("account_balances should have been populated for delegated accounts");
+
+        //     debug!(authority_account = ?signer, "Setting account to is_delegated: true");
+        //     account_balance.is_delegated = true;
+        // }
```
eerkaijun: likewise

monad-consensus-types/src/block.rs
```diff
@@ -217,6 +220,19 @@ pub struct TxnFee {
...
+impl Default for TxnFee {
+    fn default() -> Self {
+        TxnFee {
+            first_txn_value: Balance::ZERO,
+            first_txn_gas: Balance::ZERO,
+            max_gas_cost: Balance::ZERO,
+            max_txn_cost: Balance::ZERO,
+            is_delegated: false,
+        }
+    }
```
mikety: nit: strictly speaking not necessary due to built-in defaults

monad-triedb-utils/src/lib.rs
```diff
@@ -172,7 +172,37 @@ impl TriedbReader {
...
+                                let (triedb_key, key_len_nibbles) =
+                                    create_triedb_key(version, KeyInput::CodeHash(&code_hash));
+                                let res = self.handle.read(&triedb_key, key_len_nibbles, seq_num.0);
+                                debug!(?res, block_id = ?seq_num.0, ?eth_account, "account code_data");
+                                match res {
+                                    Some(data) => {
+                                        if data.len() >= 3 {
+                                            let delegation_code = &data[0..3];
+                                            eth_account.is_delegated =
+                                                delegation_code == [0xef, 0x01, 0x00];
```
mikety, line 191: nit: also can add  a more strict validation that:  `data.len() == 23`

mikety (approved):
lgtm with optional nits

mikety (approved):

ariqchowdhury (approved):

PR #2222, opened 2025-08-25
fix

mikety:


monad-eth-block-policy/src/lib.rs
```diff
@@ -258,7 +258,9 @@ where
         );
 
         let mut next_validate = emptying_txn_check_block_range.start;
-        for (_, block) in self.blocks.range(emptying_txn_check_block_range) {
+        for (seq_num, block) in self.blocks.range(emptying_txn_check_block_range) {
+            assert_eq!(*seq_num, next_validate, "Emptying range is not contiguous");
+
             if block.fees.get(eth_address).is_some()
```
ariqchowdhury, line 264: this is what we should assert on. if `block.fees.get` fails

eerkaijun: block.fees.get fails is valid i think, it means the address did not send any transaction in the block

PR #2235, opened 2025-08-26
Archive: aws cli parsing works with any s3-api-compatible provider

JoeHowarth:
- Introduced kv connection string args
- Tidied cli parsing
- Added local Bucket testing against minio
- Now supports multi region s3, self-hosted minio, r2 and any other object store provider that has s3 api compatibility

This is an incremental step to more configurability. Next I'll introduce a config file instead of relying exclusively on complicated cli arguments

### DIFF
--- a/monad-archive/src/kvstore/s3.rs
+++ b/monad-archive/src/kvstore/s3.rs
@@ -16,7 +16,9 @@
 use core::str;
 
 use aws_config::SdkConfig;
-use aws_sdk_s3::{error::SdkError, primitives::ByteStream, Client};
+use aws_sdk_s3::{
+    error::SdkError, operation::create_bucket::CreateBucketError, primitives::ByteStream, Client,
+};
 use bytes::Bytes;
 use eyre::{Context, Result};
 use tracing::trace;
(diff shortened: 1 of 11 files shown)


monad-archive/src/kvstore/s3.rs
```diff
@@ -25,27 +25,48 @@ use super::{kvstore_get_metrics, KVStoreType, MetricsResultExt};
 use crate::{metrics::Metrics, prelude::*};
 
 #[derive(Clone)]
-pub struct S3Bucket {
-    client: Client,
+pub struct Bucket {
```
JoeHowarth, line 30: Renamed S3Bucket -> Bucket since we're no longer s3 specific

monad-archive/src/test_utils.rs
```diff
@@ -80,14 +80,35 @@ pub fn mock_block(number: u64, transactions: Vec<TxEnvelopeWithSender>) -> Block
...
+    pub async fn new() -> Result<Self> {
+        let port = NEXT_PORT.fetch_add(1, std::sync::atomic::Ordering::SeqCst);
+        Command::new("docker")
+            .arg("run")
+            .arg("-p")
+            .arg(format!("{port}:9000"))
+            .arg("minio/minio")
+            .arg("server")
+            .arg("/tmp/minio")
+            .spawn()
```
JoeHowarth: docker subprocess dies once test process dies

monad-archive/src/test_utils.rs
```diff
@@ -80,14 +80,35 @@ pub fn mock_block(number: u64, transactions: Vec<TxEnvelopeWithSender>) -> Block
...
+
 pub struct TestMongoContainer {
     pub container_id: String,
     pub uri: String,
     pub port: u16,
     pub temp_dir: Option<tempfile::TempDir>,
 }
 
-static NEXT_PORT: AtomicU16 = AtomicU16::new(27017);
+static NEXT_PORT: AtomicU16 = AtomicU16::new(2812);
```
JoeHowarth, line 153: Changed to deconflict with standard mongo port that may be running on device

monad-archive/src/test_utils.rs
JoeHowarth: I was wrong here, reverted to killing the container in a custom `Drop` impl like I did for mongo

eerkaijun (dismissed):

PR #2237, opened 2025-08-27
Enable validator to full node role switch

eerkaijun:
Approach: if a validator finds out that it's gonna fall out of the active set in the next epoch (this is known when next epoch validator set is available during epoch boundary block), then the validator will switch to FullNodeClient and starts participating in secondary raptorcast. After the epoch switch, it can then transition to a full node. 

Since this allows a node to participate in usual validator raptorcast and secondary raptorcast at the same time, it requires an additional bit to differentiate whether a chunk it receives is for the validator raptorcast or the secondary raptorcast in order to determine which other nodes to rebroadcast to. An additional bit is used in the raptorcast header for this purpose

Tested on stressnet with staking

### DIFF
--- a/monad-peer-discovery/src/discovery.rs
+++ b/monad-peer-discovery/src/discovery.rs
@@ -217,16 +217,6 @@ impl<ST: CertificateSignatureRecoverable> PeerDiscoveryAlgoBuilder for PeerDisco
             rng: self.rng,
         };
 
-        // if self initializes config as a validator, make sure it's indeed in the validator set
-        if state.self_role == PeerDiscoveryRole::ValidatorNone
-            || state.self_role == PeerDiscoveryRole::ValidatorPublisher
-        {
-            assert!(
-                state.check_current_epoch_validator(&state.self_id),
-                "incorrectly set as validator"
-            );
-        }
-
         let mut cmds = Vec::new();
         self.bootstrap_peers
             .into_iter()
@@ -1226,7 +1216,7 @@ where
         round: Round,
         epoch: Epoch,
     ) -> Vec<PeerDiscoveryCommand<ST>> {
-        let mut cmds = Vec::new();
+        let cmds = Vec::new();
 
         if round > self.current_round {
             trace!(?round, "updating current round in peer discovery");
@@ -1246,18 +1236,6 @@ where
                 self.self_role = PeerDiscoveryRole::ValidatorPublisher;
                 // clear secondary raptorcast connection info
                 self.clear_connection_info();
-            }
-
-            // when a validator is demoted to a full node, defaults to FullNodeClient
-            if (self.self_role == PeerDiscoveryRole::ValidatorNone
-                || self.self_role == PeerDiscoveryRole::ValidatorPublisher)
-                && !self.check_current_epoch_validator(&self.self_id)
-            {
-                debug!(?epoch, "validator demoted to full node");
-                self.self_role = PeerDiscoveryRole::FullNodeClient;
-                // clear secondary raptorcast connection info
-                self.clear_connection_info();
-                cmds.extend(self.look_for_upstream_validators());
             }
         }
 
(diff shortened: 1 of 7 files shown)


monad-raptorcast/src/util.rs
```diff
@@ -536,28 +526,42 @@ where
...
-            if !author_found {
-                tracing::debug!(?author_node_id, ?self.validator_map,
-                    "Validator author for v2v group not found in validator_map");
-            }
             author_found
         } else {
-            tracing::debug!(?msg_epoch, ?self.validator_map,
-                "Epoch not found in validator_map");
-            false
+            unreachable!("Either broadcast or secondary_broadcast must be set");
```
fabiotdf: Maybe its better to log here rather than panic, since `secondary_broadcast` and `broadcast` are a form of user input?
I.e. if these fields are wrong, its not a logic error, it could be a malicious node crafting the packet, no?

monad-raptorcast/src/util.rs
```diff
@@ -605,42 +623,35 @@ where
...
+        // clear old validator map
+        self.validator_map
+            .retain(|key, _| *key + Epoch(1) >= curr_epoch);
+
+        debug!(
             epoch=?curr_epoch,
             round=?curr_round,
-            ?old_count,
-            ?new_count,
+            validator_map_len_diff=?self.validator_map.len(),
```
fabiotdf: Could we somehow keep the old count in the log, so we can see how many groups got deleted?

fabiotdf (commented):
LGTM with a couple of minor comments

monad-raptorcast/src/util.rs
eerkaijun: we only enter this function if either bit is set, we checked this earlier in the code

monad-raptorcast/src/udp.rs
```diff
@@ -914,7 +924,9 @@ where
                 let (cursor_version, cursor) = cursor.split_at_mut(2);
                 cursor_version.copy_from_slice(&version.to_le_bytes());
                 let (cursor_broadcast_merkle_depth, cursor) = cursor.split_at_mut(1);
-                cursor_broadcast_merkle_depth[0] = ((is_raptor_broadcast as u8) << 7) | tree_depth;
+                cursor_broadcast_merkle_depth[0] = ((is_raptor_broadcast as u8) << 7)
+                    | ((is_secondary_raptor_broadcast as u8) << 6)
+                    | (tree_depth & 0b0000_1111); // tree_depth max 4 bits
```
omegablitz, line 949: Should add something like this at the top of the file:
`const _: () = assert!(MAX_MERKLE_TREE_DEPTH <= 0xF, "merkle tree depth must be <= 4 bits");`

monad-raptorcast/src/util.rs
```diff
@@ -570,11 +574,25 @@ where
         &self,
         msg_epoch: Epoch, // for validator-to-validator re-raptorcasting only
         msg_author: &NodeId<CertificateSignaturePubKey<ST>>, // skipped when iterating RaptorCast group
+        broadcast: bool,
```
omegablitz: Can this take in an enum that's either PrimaryRaptorcast or SecondaryRaptorcast instead of two bools?

monad-raptorcast/src/udp.rs
eerkaijun: added

monad-peer-discovery/src/discovery.rs
```diff
@@ -1403,20 +1394,13 @@ where
...
             .filter(|(id, _)| {
-                !curr_validators.contains(id)
-                    && self
-                        .participation_info
-                        .get(id)
-                        .map(|p| p.status == SecondaryRaptorcastConnectionStatus::Connected)
-                        .unwrap_or(false)
+                self.participation_info
+                    .get(id)
+                    .map(|p| p.status == SecondaryRaptorcastConnectionStatus::Connected)
```
omegablitz: nit: `Option::is_some_and` would be easier to read

omegablitz (approved):

PR #2242, opened 2025-08-28
Fix peer discovery round update

eerkaijun:
observed on stressnet that peer discovery only updates round number when epoch is bumped, which is incorrect

### DIFF
--- a/monad-raptorcast/src/lib.rs
+++ b/monad-raptorcast/src/lib.rs
@@ -355,11 +355,11 @@ where
                                 break;
                             }
                         }
-                        self.peer_discovery_driver
-                            .lock()
-                            .unwrap()
-                            .update(PeerDiscoveryEvent::UpdateCurrentRound { round, epoch });
                     }
+                    self.peer_discovery_driver
+                        .lock()
+                        .unwrap()
+                        .update(PeerDiscoveryEvent::UpdateCurrentRound { round, epoch });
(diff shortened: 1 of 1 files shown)


monad-raptorcast/src/lib.rs
```diff
@@ -355,11 +355,11 @@ where
...
                         }
-                        self.peer_discovery_driver
-                            .lock()
-                            .unwrap()
-                            .update(PeerDiscoveryEvent::UpdateCurrentRound { round, epoch });
                     }
+                    self.peer_discovery_driver
+                        .lock()
+                        .unwrap()
+                        .update(PeerDiscoveryEvent::UpdateCurrentRound { round, epoch });
```
dshulyak, line 362: would be good to check in test to emphasize that this is intentional and make regressions less likely

eerkaijun: it's harder to unit test this right now since it involves peer discovery too... i guess eventually it would be good to build up some test harness for the interactions of raptorcast and peer discovery since they are closely related

ariqchowdhury: can you create an issue to track and we can create the tests later

eerkaijun: tracked here https://github.com/category-labs/category-internal/issues/1871

PR #2243, opened 2025-08-28
Bump execution and disable event ring tests

andr-dev:


andr-dev:
Commit being cherry-picked into another PR.

PR #2244, opened 2025-08-28
Change txpool account balance check on insertion

andr-dev:
This change relaxes the account balance check during txpool insertion to only check if the account balance is non-zero.


dshulyak (approved):
does it make any publicly open rpc trivially dos'able? perhaps config option to use stricter validation will make it less risky

ariqchowdhury (approved):

PR #2248, opened 2025-08-29
Add nonce usages for cached nonce computations

andr-dev:
Previously, the account nonce of an address after executing a block could be statically determined by taking the highest tx nonce and incrementing it by one. After EIP7702, nonce increments for authorization authorities is conditioned on whether the authorization is valid which can only be determined if the the nonce of the authority immediately before the block is known. This makes it impossible to cache the nonces of a block simply using a map from address to highest tx nonce.

This change introduces the NonceUsagesMap and the concept of a NonceUsage, which is `NonceUsage::Known(nonce)` when the last usage of a nonce in a block is statically determined to be `nonce`. For authorization lists, the `NonceUsage::Possible(possible_nonces)` variant which lists (in order) which nonces might have been used, depending on what the previous account nonce was. Given a previous account nonce, it is always possible to determine the nonce after a block by either using the `Known(nonce)` variant or by applying the `Possible(possible_nonces)` on top of the known account nonce.


monad-eth-block-policy/src/lib.rs
```diff
@@ -730,21 +787,44 @@ where
...
-        let pending_block_nonces = extending_blocks.get_account_nonces();
+
+        let mut cached_nonce_usages = NonceUsageMap::default();
+
+        for nonce_usages in self
+            .committed_cache
+            .blocks
+            .values()
+            .map(|block| &block.nonce_usages)
+            .chain(extending_blocks.iter().map(|block| &block.nonce_usages))
```
eerkaijun, line 678: i think we just need to limit these to the last k blocks here

ariqchowdhury (approved):

omegablitz (approved):

PR #2250, opened 2025-08-30
Use chain config in txpool

andr-dev:


monad-eth-txpool/src/pool/mod.rs
```diff
@@ -327,19 +314,71 @@ where
...
+            info!(chain_params =? self.chain_revision.chain_params(), "updating chain revision");
+
+            self.static_validate_all_txs();
+        }
+    }
+
     pub fn update_committed_block(
         &mut self,
         event_tracker: &mut EthTxPoolEventTracker<'_>,
+        chain_config: &impl ChainConfig<CRT>,
```
omegablitz, line 331: Rather than being passed in on commit/enter_round, would it make sense for this to exist as a member in EthTxPool? The chain config should be static for the lifetime of the EthTxPool I would think

andr-dev: The model I'm using for txpool stuff is that `EthTxPool` is only a data structure that manages txs given resouces (like event_tracker, block_policy, etc.). `EthTxPoolExecutor` on the other hand schedules tasks on its `EthTxPool`, hence why it's the owner of the following:
```
    block_policy: EthBlockPolicy<ST, SCT>,
    state_backend: SBT,
    chain_config: CCT,
```

mikety (approved):

monad-eth-txpool/src/pool/mod.rs
mikety: This approach adds an extra parameter  that is passed  to each call, the other approach is to use a reference in the structures that use it locally.

eerkaijun (approved):
probably doesn't need to block this PR but it feels a bit weird to me that `chain_config` has to be passed in during enter_round. couldn't that be stored by EthTxPool?

PR #2253, opened 2025-08-31
Add ChainRevision to EthBlockPolicy

andr-dev:
To gate reserve balance and EIP7702 behind activations, EthBlockPolicy will need a ChainRevision. This change adds the generic in preparation for its use.

Additionally, the `chain_id` field from `EthBlockPolicy` was removed.


monad-eth-block-policy/src/lib.rs
```diff
@@ -317,10 +318,11 @@ where
 
 /// A block policy for ethereum payloads
 #[derive(Debug)]
-pub struct EthBlockPolicy<ST, SCT>
+pub struct EthBlockPolicy<ST, SCT, CRT>
```
mikety: Please use reference to ChainConfig so its available to all calling methods in BLockPolicy

mikety (dismissed):

ariqchowdhury (approved):

bamrith (approved):

PR #2258, opened 2025-09-01
Add v0.11.0 activation

bamrith:


monad-chain-config/src/revision.rs
```diff
@@ -63,27 +66,39 @@ pub struct ChainParams {
...
     vote_pace: Duration::from_millis(400),
+    validate_system_txs: false,
+};
+
+const CHAIN_PARAMS_V_0_11_0: ChainParams = ChainParams {
+    tx_limit: 5_000,
+    proposal_gas_limit: 150_000_000,
+    proposal_byte_limit: 2_000_000,
+    vote_pace: Duration::from_millis(400),
+    validate_system_txs: true,
```
mikety, line 120: it also need params for activating reserve_balance : 
tfm and eip7702, possibly with timestamp and round acotivation.

michael-yxchen (approved):

PR #2263, opened 2025-09-02
Apply proposal limits to system txs in txpool

andr-dev:


omegablitz:
I don't think this is desirable behavior; system txs should never be silently dropped. If anything we should be asserting that system txs never overflow the block limits (they won't in practice).

@bamrith correct me if I'm wrong

andr-dev:
> I don't think this is desirable behavior; system txs should never be silently dropped. If anything we should be asserting that system txs never overflow the block limits (they won't in practice).
> 
> @bamrith correct me if I'm wrong

Yep, we were just discussing this as you posted 😂. We decided on forcing the node to timeout with an error in the logs, although this should never happen in practice.

omegablitz:
How do we think about timing out vs asserting? As currently implemented, if 1 node fails validation on this, wouldn't every other node also be stuck failing validation on this?

I would have thought that under the current paradigm, there's a fatal design error somewhere if the system transactions don't fit in the proposal limits

andr-dev:
> How do we think about timing out vs asserting? As currently implemented, if 1 node fails validation on this, wouldn't every other node also be stuck failing validation on this?

My thoughts were that nodes will run different versions and if one of those versions has a fatal design error then, although those nodes would cause a timeout for their leader rounds, they would continue participating in consensus and voting on other proposals.

omegablitz:
Actually, I think we should assert. A node that's failing these checks won't be able to vote properly either as far as I can tell

PR #2267, opened 2025-09-02
Txgen: allow multiple generators and multiple rpcs

JoeHowarth:


monad-eth-testutil/examples/txgen/workers/rpc_sender.rs
```diff
@@ -223,14 +222,21 @@ impl RpcSender {
...
-            return; // unnecessary?
+            return;
         }
-        trace!(batch_size = batch.len(), "Sending batch of txs...");
 
-        let recipient_sender = self.recipient_sender.clone();
-        let client = self.client.clone();
+        // round robin through clients
+        self.round_robin_idx = (self.round_robin_idx + 1) % self.clients.len();
+        let client = self.clients[self.round_robin_idx].clone();
```
eclark0426, line 232: I wonder if it might make sense to, instead of having a round-robin where each batch goes to a different client, if it might make sense to split the batch into `n` parts and send a part of the batch to each client. Maybe I'm thinking about it too linearly when ultimately this is running in parallel with the other tasks, but it seems like splitting the batch between servers would better test the ability of nodes to communicate transactions between each other

monad-eth-testutil/examples/txgen/sample_configs/sequential_phases.toml
```diff
@@ -0,0 +1,62 @@
+rpc_urls = ["http://localhost:33332", "http://localhost:8080"]
```
eclark0426, line 1: Would it make sense for the RPC URLs to be configurable an a workload group level, or even on the level of each traffic gen? Then there could be a configuration which, say, sends a lot of transactions to one RPC server and just a few to a second, to ensure the second server is still able to get transactions in.

This probably would be more than just a couple line change, so doesn't necessarily need to be implemented as part of this PR. More just a thought for how we might expand on the idea of multiple RPC clients

monad-eth-testutil/examples/txgen/workers/rpc_sender.rs
JoeHowarth: Each batch is designed to be sent as a single rpc call so that we can push as many txs as possible with the lowest overhead (this is useful for when we don't want to stress teh rpc or network, but mainly hit the txpool and execution). If we wanted to load balance traffic better the way to do that is to have smaller batches imo. 

All this should be put into a readme / operator's guide though... It's currently too hard to understand how the params map to behavior.

monad-eth-testutil/examples/txgen/sample_configs/sequential_phases.toml
JoeHowarth: I think that's a good idea and could be useful.
It's quite a bit more invasive than this change so I'd like to leave it for a follow-up, but worth revisiting 👍

eclark0426 (approved):

PR #2268, opened 2025-09-02
Assert locked epoch validators on startup

bamrith:


monad-state/src/lib.rs
```diff
@@ -1303,6 +1318,37 @@ where
...
+                    .iter()
+                    .map(|val_data| (val_data.node_id, (val_data.stake, val_data.cert_pubkey)))
+                    .collect();
+
+                let db_val_data: BTreeMap<
+                    NodeId<SCT::NodeIdPubKey>,
+                    (Stake, SignatureCollectionPubKeyType<SCT>),
+                > = self
+                    .state_backend
+                    .read_valset_at_block(delay_seq_num, locked_epoch) // TODO use root_seq_num here
```
omegablitz, line 1340: to confirm, this panics/aborts if the locked_epoch can't be found?

omegablitz (approved):

monad-state/src/lib.rs
bamrith: correct

ariqchowdhury (approved):

PR #2269, opened 2025-09-02
Revalidate txs on txpool create proposal

andr-dev:


monad-eth-txpool-executor/src/lib.rs
```diff
@@ -380,7 +380,8 @@ where
                     round,
                     upcoming_leader_rounds,
                 } => {
-                    self.pool.enter_round(&self.chain_config, round);
+                    self.pool
```
mikety, line 385: It does not catch TFM change in ExecutionChainParams.

mikety (commented):
It should revalidate the transaction on TFM change as well which is triggered by a timestamp change. 
Also a test case will be helpful (can be done as a follow up)

monad-eth-txpool-executor/src/lib.rs
andr-dev: Change in `ExecutionChainParams` is caught during `create_proposal` because the timestamp isn't known until either
1. You are proposing
2. You receive a committed block

monad-eth-txpool/src/pool/mod.rs
```diff
@@ -267,10 +268,41 @@ where
...
+                self.chain_revision = chain_revision;
+                self.execution_revision = execution_revision;
+
+                info!(
+                    chain_params =? chain_revision.chain_params(),
+                    execution_revision =? execution_revision,
+                    "updating chain params and execution revision in create_proposal"
+                );
+
+                self.static_validate_all_txs(event_tracker);
```
mikety, line 298: Is this intention that the balances are not re-validated, i.e if reserve balance changes and a transactions does not pass tfm balance verification then it will not be revalidated and will be included in the proposal?

andr-dev: Reserve balance is only checked at proposal creation time, the only thing that needs to be rerun is static validation which changes as a function of the chain params / execution params.

ariqchowdhury (approved):

PR #2271, opened 2025-09-03
activate tfm on timestamp

mikety:
Activate TFM on timestamp to match execution.

### DIFF
--- a/monad-chain-config/src/execution_revision.rs
+++ b/monad-chain-config/src/execution_revision.rs
@@ -19,10 +19,11 @@ pub enum MonadExecutionRevision {
     V_ZERO,
     V_ONE,
     V_TWO,
+    V_FOUR,
 }
 
 impl MonadExecutionRevision {
-    pub const LATEST: Self = Self::V_TWO;
+    pub const LATEST: Self = Self::V_FOUR;
 }
 
 impl MonadExecutionRevision {
@@ -31,6 +32,7 @@ impl MonadExecutionRevision {
             Self::V_ZERO => &EXECUTION_CHAIN_PARAMS_V_ZERO,
             Self::V_ONE => &EXECUTION_CHAIN_PARAMS_V_ONE,
             Self::V_TWO => &EXECUTION_CHAIN_PARAMS_V_TWO,
+            Self::V_FOUR => &EXECUTION_CHAIN_PARAMS_V_FOUR,
         }
     }
 }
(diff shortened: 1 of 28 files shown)


andr-dev (dismissed):
Looks good!
As follow up, I wonder if we can cleanup the timestamp/1_000_000_000, get_chain_revision, get_execution_chain_revision boilerplate. Would also be nice to somehow assert that if `tfm_enabled` for execution params, then `tfm_enabled` for chain parms.

monad-chain-config/src/execution_revision.rs
```diff
@@ -19,10 +19,12 @@ pub enum MonadExecutionRevision {
     V_ZERO,
     V_ONE,
     V_TWO,
+    V_FOUR,
 }
 
 impl MonadExecutionRevision {
     pub const LATEST: Self = Self::V_TWO;
+    pub const LATEST_NEXT: Self = Self::V_FOUR;
```
michael-yxchen: LATEST is only used in unit tests. why do we need a LATEST_NEXT?

monad-chain-config/src/revision.rs
```diff
@@ -70,55 +70,50 @@ pub struct ChainParams {
...
-    pub tfm: bool,
     pub eip_7702: bool,
 }
 
 const CHAIN_PARAMS_V_0_7_0: ChainParams = ChainParams {
     tx_limit: 10_000,
     proposal_gas_limit: 300_000_000,
     proposal_byte_limit: 4_000_000,
-    max_reserve_balance: 100_000_000_000_000_000_000, // 100 MON
+    max_reserve_balance: 10_000_000_000_000_000_000, // 10 MON
```
michael-yxchen, line 80: is the change from 100 to 10 MON intentional?

monad-eth-block-policy/src/lib.rs
```diff
@@ -417,6 +425,12 @@ fn is_possibly_emptying_transaction(
     blocks_since_latest_txn > execution_delay - SeqNum(1)
 }
 
+fn convert_timestamp_to_seconds(timestamp_ns: u128) -> u64 {
+    let timestamp_seconds = timestamp_ns / 1_000_000_000;
+    assert!(timestamp_seconds < u64::MAX.into());
```
mikety, line 363: Ensure timestamp is validated for sanity earlier

monad-chain-config/src/revision.rs
ariqchowdhury: yes, its changed on execution side as well

monad-chain-config/src/lib.rs
```diff
@@ -200,6 +205,7 @@ const MONAD_TESTNET2_CHAIN_CONFIG: MonadChainConfig = MonadChainConfig {
 
     execution_v_one_activation: 0,
     execution_v_two_activation: 0,
+    execution_v_four_activation: 0,
```
michael-yxchen: we should use u64::MAX as tn2/mn activation placeholder values

ariqchowdhury (approved):

PR #2272, opened 2025-09-03
Fix extending blocks range when resolving nonces

eerkaijun:
We also need to filter the extending block range to last execution_delay blocks (similar to filtering the committed cache blocks) when resolving nonces


omegablitz:
Can you apply the filter after chaining committed_cache and extending_blocks? It would be more obviously correct that way.

ariqchowdhury (approved):

PR #2274, opened 2025-09-03
Disallow system transaction sender to sign authorization

eerkaijun:
skip authorization that is signed by system transaction sender. need corresponding execution changes too

### DIFF
--- a/monad-eth-block-validator/src/lib.rs
+++ b/monad-eth-block-validator/src/lib.rs
@@ -46,7 +46,9 @@ use monad_eth_block_policy::{
 use monad_eth_types::{EthBlockBody, EthExecutionProtocol, ExtractEthAddress, ProposedEthHeader};
 use monad_secp::RecoverableAddress;
 use monad_state_backend::StateBackend;
-use monad_system_calls::{validator::SystemTransactionValidator, SystemTransaction};
+use monad_system_calls::{
+    validator::SystemTransactionValidator, SystemTransaction, SYSTEM_SENDER_ETH_ADDRESS,
+};
 use monad_types::Balance;
 use monad_validator::signature_collection::{SignatureCollection, SignatureCollectionPubKeyType};
 use rayon::iter::{IntoParallelIterator, ParallelIterator};
@@ -372,6 +374,11 @@ where
                             .map(|authority| (authority, authorization.inner()))
                     }) {
                         trace!(address =? authorization.address, nonce =? authorization.nonce, ?authority, "Signed authority");
+
+                        // do not allow system account from sending authorization
+                        if authority == SYSTEM_SENDER_ETH_ADDRESS {
+                            return Err(BlockValidationError::TxnError);
+                        }
 
                         if authorization.chain_id != 0_u64
                             && authorization.chain_id != chain_config.chain_id()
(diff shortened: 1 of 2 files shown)


bamrith (dismissed):

monad-eth-block-validator/src/lib.rs
```diff
@@ -373,9 +373,11 @@ where
                     }) {
                         trace!(address =? authorization.address, nonce =? authorization.nonce, ?authority, "Signed authority");
 
-                        if authorization.chain_id != 0_u64
-                            && authorization.chain_id != chain_config.chain_id()
-                        {
+                        if !static_validate_authorization(
```
mikety: It needs to reject this txn in txpool and do not validate the blocks with this txn in check_coherency.

mikety (changes requested):

monad-eth-block-policy/src/validation.rs
```diff
@@ -245,6 +247,24 @@ fn compute_floor_data_gas(tx: &TxEnvelope) -> u64 {
...
+    chain_id: u64,
+    authorization_chain_id: u64,
+    authority: Address,
+) -> bool {
+    if authorization_chain_id != 0_u64 && authorization_chain_id != chain_id {
+        return false;
+    }
+
+    // system sender are not allowed to set authorization, skip authorization
+    if authority == SYSTEM_SENDER_ETH_ADDRESS {
```
bamrith: I think we should start this check on v0.11.0 activation

monad-eth-txpool/src/pool/tracked/sequencer.rs
```diff
@@ -47,26 +48,24 @@ impl<'a> OrderedTx<'a> {
     fn new(tx: &'a ValidEthTransaction, chain_id: u64, base_fee: u64) -> Option<Self> {
         let effective_tip_per_gas = tx.raw().effective_tip_per_gas(base_fee)?;
 
+        let mut valid_signed_authorizations = Vec::new();
+        for signed_authorization in tx.raw().authorization_list().into_iter().flatten() {
+            if signed_authorization.chain_id != 0 && signed_authorization.chain_id != chain_id {
+                continue;
```
eerkaijun, line 54: @andr-dev this was previously return None, but I think it should be continue?

andr-dev: Yep that's correct, the `return None` previously was for the scope of ther `flat_map`, this looks good :)

andr-dev (dismissed):

monad-eth-txpool/src/pool/tracked/sequencer.rs
eerkaijun: aa ok i'm dumb haha didn't look closely

mikety (dismissed):

monad-eth-block-policy/src/validation.rs
```diff
@@ -28,6 +29,11 @@ pub fn static_validate_transaction(
         return Err(TransactionError::UnsupportedTransactionType);
     }
 
+    // do not allow transaction from the system account
+    if tx.signer() == SYSTEM_SENDER_ETH_ADDRESS {
```
bamrith: this is duplicate. can be removed

bamrith (approved):

PR #2275, opened 2025-09-03
[rpc] Prestate and statediff tracers

dhil:


monad-rpc/src/handlers/eth/call.rs
```diff
@@ -706,6 +723,14 @@ pub async fn monad_eth_call<T: Triedb + TriedbPath>(
     }
 }
 
+// TODO(dhil): Bizarrely substituting this type alias in below causes a type error.
```
philipglazman: I think the `rpc` macro is complaining because it can't parse generic return types potentially. Can be fixed in a followup.

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -720,12 +745,14 @@ pub async fn monad_debug_traceCall<T: Triedb + TriedbPath>(
     chain_id: u64,
     eth_call_gas_limit: u64,
     params: MonadDebugTraceCallParams,
-) -> JsonRpcResult<Option<MonadCallFrame>> {
-    trace!("monad_debug_traceCall: {params:?}");
+) -> JsonRpcResult<BoxRawValue> {
```
philipglazman: Returning `BoxRawValue` will break rpc documentation since we use the return type to generate user-facing docs and the return type will be ambigious.

andr-dev: I think this is fine for now since debug_traceCall might return one of many types which do not yet exist in rust. They are CBOR serialized which we deserialize into a generic "Value" (like serde_json::Value) which then gets JSON serialized and returned.

philipglazman (approved):

PR #2276, opened 2025-09-03
Simplify EthBlockPolicy nonce retrieval

andr-dev:
The `seq_num` param for `get_account_base_nonces` was an unfortunate side effect of a previous invariant of the committed cache where it sometimes wouldn't have enough blocks and would require passing the desired sequence number + 1 to force the underlying triedb lookup to always succeed. Now that the committed cache stores all blocks within the range (N-2k, N], this parameter is no longer required.

### DIFF
--- a/monad-eth-block-policy/src/lib.rs
+++ b/monad-eth-block-policy/src/lib.rs
@@ -648,10 +648,9 @@ where
         }
     }
 
-    /// returns account nonces at the start of the provided consensus block
+    /// returns account nonces at the end of the extending blocks
     pub fn get_account_base_nonces<'a>(
         &self,
-        consensus_block_seq_num: SeqNum,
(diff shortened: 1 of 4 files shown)


monad-eth-block-policy/src/lib.rs
```diff
@@ -665,14 +664,14 @@ where
 
         let addresses = addresses.unique().collect::<HashSet<&'a Address>>();
 
-        let base_seq_num = consensus_block_seq_num.max(self.execution_delay) - self.execution_delay;
+        let base_seq_num = self.last_commit.max(self.execution_delay) - self.execution_delay;
```
eerkaijun, line 667: this is incorrect? self.last_commit is not the same as consensus_block_seq_num I think?

PR #2280, opened 2025-09-03
Add chain config to compute base fee

michael-yxchen:
Fixes a bug that current proposal_gas_limit is used when calculating base fee. Parent proposal gas limit should be used there. This PR adds chain config to compute_base_fee so it can fetch corresponding block gas limit


ariqchowdhury (dismissed):

omegablitz:
Does qc.get_round() work for forking if it's a reproposal?

omegablitz (dismissed):

ariqchowdhury (approved):

andr-dev (dismissed):

omegablitz (approved):

PR #2290, opened 2025-09-04
Unit test for multiple delegations and emptying tx

eerkaijun:
multiple delegations unit test

### DIFF
--- a/monad-eth-block-policy/src/lib.rs
+++ b/monad-eth-block-policy/src/lib.rs
@@ -2000,6 +2000,59 @@ mod test {
             CoherencyCheckMode::ReserveBalanceCoherency,
         );
         assert!(result.is_ok(), "Block coherency check failed: {:?}", result);
+
+        /////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


monad-eth-block-policy/src/lib.rs
```diff
@@ -2000,6 +2000,59 @@ mod test {
             CoherencyCheckMode::ReserveBalanceCoherency,
         );
         assert!(result.is_ok(), "Block coherency check failed: {:?}", result);
+
+        /////////////////////////////////////////////////////////////////////////////
+        // Case7: Multiple 7702 delegations and attempted emptying transaction    ///
+        /////////////////////////////////////////////////////////////////////////////
+
+        // first tx has multiple delegation and undelegation authorization from the signer
```
philipglazman, line 2008: nit: from a different signer. Auth lists signed by S1 but tx signer is S2

eerkaijun: aaa I meant signer for tx2

PR #2292, opened 2025-09-04
Dynamic base fee activation

michael-yxchen:
Activates dynamic base fee with execution `tfm_enabled` flag. Minimum base fee is bumped from 50 gwei to 100 gwei


monad-tfm/src/base_fee.rs
```diff
@@ -19,6 +19,7 @@ pub const GENESIS_BASE_FEE: u64 = 0;
 // trend is a signed value in 2's complement representation
 pub const GENESIS_BASE_FEE_TREND: u64 = (0i64).cast_unsigned();
 pub const GENESIS_BASE_FEE_MOMENT: u64 = 0;
+pub const PRE_TFM_BASE_FEE: u64 = 50_000_000_000; // 50 gwei
```
omegablitz, line 22: nit: remove `// pub const BASE_FEE_PER_GAS: u64 = 50_000_000_000;` from monad-eth-types/src/lib.rs

monad-eth-block-validator/src/lib.rs
```diff
@@ -324,7 +324,10 @@ where
                 return Err(BlockValidationError::TxnError);
             }
 
-            if eth_txn.max_fee_per_gas() < header.base_fee.into() {
+            let block_base_fee = header
+                .base_fee
+                .unwrap_or(monad_tfm::base_fee::PRE_TFM_BASE_FEE);
+            if eth_txn.max_fee_per_gas() < block_base_fee.into() {
```
mikety, line 341: It would be nice to add similar to `check_coherency` validation that  if from `block.get_timestamp()` computes that  `tfm_enbled  == true than block. base_fee.is_some()` and if `tfm_enabled == false than block. base_fee.is_none()`

monad-eth-block-policy/src/lib.rs
```diff
@@ -358,8 +359,9 @@ fn is_possibly_emptying_transaction(
     !balance_state.is_delegated && blocks_since_latest_txn > execution_delay - SeqNum(1)
 }
 
-fn convert_timestamp_to_seconds(timestamp_ns: u128) -> u64 {
-    let timestamp_seconds = timestamp_ns / 1_000_000_000;
+pub fn timestamp_ns_to_secs(timestamp_ns: u128) -> u64 {
+    const NSEC_PER_SEC: u128 = 1_000_000_000;
+    let timestamp_seconds = timestamp_ns / NSEC_PER_SEC;
     assert!(timestamp_seconds < u64::MAX.into());
```
mikety: can `timestamp_ns` be trusted to be sane ?

monad-eth-txpool/src/pool/transaction.rs
```diff
@@ -67,14 +67,22 @@ impl ValidEthTransaction {
...
+        };
+        if tx.max_fee_per_gas() < min_base_fee.into() {
             event_tracker.drop(tx.tx_hash().to_owned(), EthTxPoolDropReason::FeeTooLow);
             return None;
         }
 
-        // TODO(misha): use a method that can put readl base_fee to txn at some point.
-        let max_value = compute_txn_max_value(&tx, last_commit.base_fee);
-        let max_gas_cost = compute_txn_max_gas_cost(&tx, last_commit.base_fee);
+        // TODO(misha): use a method that can put real base_fee to txn at some point.
```
mikety: this comment can be removed now

monad-eth-block-policy/src/lib.rs
```diff
@@ -976,54 +978,79 @@ where
...
+                    parent_block.base_fee_trend,
+                    parent_block.base_fee_moment,
+                    parent_block.block_gas_usage,
+                )
+            }
+        };
+
+        // if parent block doesn't have base_fee fields, it must be pre-tfm
+        // block and we return genesis values
+        let (parent_base_fee, parent_trend, parent_moment) = if maybe_parent_base_fee.is_none()
```
omegablitz: super nit: you could avoid the unwraps if you `match (maybe_parent_base_fee, maybe_parent_trend, maybe_parent_moment)`, but the current impl is also fine.

omegablitz (dismissed):

mikety (commented):
I think the validation of the base fee should be done earlier in validate_block_body

omegablitz (dismissed):

monad-eth-block-policy/src/lib.rs
michael-yxchen: right we can't trust it. im removing the assertion and clamping the value at u64::MAX

monad-eth-block-validator/src/lib.rs
michael-yxchen: is this what you're referring to? 

https://github.com/category-labs/monad-bft/blob/7da2756014bf22c8e9f77652312c8ede1d17d3e1/monad-eth-block-policy/src/lib.rs#L1344-L1356

mikety: yes

PR #2294, opened 2025-09-04
[monad-event-ring] fix for very large snapshots

kjcamann:
Snapshots are decompressed into a Vec<u8> by the zstd library, then manually copied over to an anonymous memory region associated with a memfd_create(2) file descriptor, using the write(2) system call. For very large snapshots, short writes tend to occur, so this loops until we copy the whole thing.


kjcamann:
This is implemented in a different way, in a later PR

PR #2296, opened 2025-09-04
Add authorization list processing fee to intrinsic gas

mikety:


monad-eth-block-policy/src/validation.rs
```diff
@@ -65,6 +65,7 @@ pub fn static_validate_transaction(
 
 pub const TFM_MAX_EIP2718_ENCODED_LENGTH: usize = 384 * 1024;
 pub const TFM_MAX_GAS_LIMIT: u64 = 30_000_000;
+pub const TFM_PER_EMPTY_ACCOUNT_COST: u64 = 25_000;
```
ariqchowdhury: nit: this should be `EIP_7702_PER_EMPTY...` I don't think this value is defined in the TFM spec, its from the EIP7702 spec

monad-eth-block-policy/src/validation.rs
```diff
@@ -541,4 +560,36 @@ mod test {
...
+            0,
+            vec![
+                make_signed_authorization(S2, secret_to_eth_address(S1), 0),
+                make_signed_authorization(S2, secret_to_eth_address(S1), 0),
+            ],
+            0,
+        );
+
+        let result_2_auth = compute_intrinsic_gas(&tx_2_auth);
+        assert_eq!(result_2_auth, result_1_auth + TFM_PER_EMPTY_ACCOUNT_COST);
```
ariqchowdhury: nit: can you use the expected value for the second term instead of result_1_auth + ...

mikety: I wanted to highlight that the only change is the processing fee of another entry, do you think it makes the test intention more clear?

ariqchowdhury (approved):

philipglazman (approved):

andr-dev (dismissed):

andr-dev (approved):

ariqchowdhury (approved):

PR #2297, opened 2025-09-04
Accept 7702 auth list in rpc eth_call

philipglazman:


eerkaijun:
also need to support here https://github.com/category-labs/monad-bft/blob/master/monad-rpc/src/handlers/eth/call.rs#L299

eerkaijun (approved):

omegablitz (approved):

PR #2299, opened 2025-09-05
Consolidate recover_authority usage

eerkaijun:
recover_authority is an expensive operation, we should not redo them across monad-eth-block-policy and monad-eth-block-validator

### DIFF
--- a/monad-eth-block-validator/src/lib.rs
+++ b/monad-eth-block-validator/src/lib.rs
@@ -24,6 +24,7 @@ use alloy_consensus::{
     transaction::{Recovered, Transaction},
     TxEnvelope, EMPTY_OMMER_ROOT_HASH,
 };
+use alloy_eips::eip7702::{RecoveredAuthority, RecoveredAuthorization};
 use alloy_rlp::Encodable;
 use monad_chain_config::{
     revision::{ChainParams, ChainRevision},
@@ -44,7 +45,9 @@ use monad_eth_block_policy::{
     validation::static_validate_transaction,
     EthBlockPolicy, EthValidatedBlock,
 };
-use monad_eth_types::{EthBlockBody, EthExecutionProtocol, ExtractEthAddress, ProposedEthHeader};
+use monad_eth_types::{
+    EthBlockBody, EthExecutionProtocol, ExtractEthAddress, ProposedEthHeader, ValidatedTx,
+};
 use monad_secp::RecoverableAddress;
 use monad_state_backend::StateBackend;
 use monad_system_calls::{
@@ -52,11 +55,11 @@ use monad_system_calls::{
 };
 use monad_types::Balance;
 use monad_validator::signature_collection::{SignatureCollection, SignatureCollectionPubKeyType};
-use rayon::iter::{IntoParallelIterator, ParallelIterator};
+use rayon::iter::{IntoParallelIterator, IntoParallelRefIterator, ParallelIterator};
 use tracing::{debug, trace, trace_span, warn};
 
 type SystemTransactions = Vec<SystemTransaction>;
-type ValidatedTxns = Vec<Recovered<TxEnvelope>>;
+type ValidatedTxns = Vec<ValidatedTx>;
 
 /// Validates transactions as valid Ethereum transactions and also validates that
 /// the list of transactions will create a valid Ethereum block
@@ -275,6 +278,12 @@ where
             return Err(BlockValidationError::TxnError);
         }
 
+        // early return if sum of transaction gas limits exceed block gas limit
+        let total_gas: u64 = transactions.iter().map(|tx| tx.gas_limit()).sum();
+        if total_gas > chain_params.proposal_gas_limit {
+            return Err(BlockValidationError::TxnError);
+        }
+
         // recovering the signers verifies that these are valid signatures
         let recovered_txns: VecDeque<Recovered<TxEnvelope>> = transactions
             .into_par_iter()
@@ -299,6 +308,21 @@ where
                     return Err(BlockValidationError::SystemTxnError);
                 }
             };
+
+        // early return if proposal size exceed limit
+        let system_txns_size: usize = system_txns.iter().map(|tx| tx.length()).sum();
+        let user_txns_size: usize = eth_txns.iter().map(|tx| tx.length()).sum();
+        let proposal_size = system_txns_size + user_txns_size;
+        debug!(
+            total_gas,
+            proposal_size,
+            txs = eth_txns.len(),
+            "proposal stats"
+        );
+
+        if proposal_size as u64 > chain_params.proposal_byte_limit {
+            return Err(BlockValidationError::TxnError);
+        }
 
         let mut nonce_usages = NonceUsageMap::default();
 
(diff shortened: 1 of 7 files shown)


monad-eth-testutil/src/lib.rs
```diff
@@ -366,9 +368,27 @@ pub fn generate_consensus_test_block(
...
 
+    let mut validated_txns: Vec<ValidatedTx> = Vec::new();
+    for tx in txs.into_iter() {
+        let authorizations: Vec<RecoveredAuthorization> = tx
+            .authorization_list()
+            .unwrap_or(&Vec::new())
+            .iter()
+            .filter_map(|a| {
+                a.recover_authority().ok().map(|authority| {
+                    RecoveredAuthorization::new_unchecked(
```
omegablitz, line 379: nit: `authority.clone().into_recovered()` to dodge the `new_unchecked` perhaps?
https://docs.rs/alloy-eip7702/0.6.1/alloy_eip7702/struct.SignedAuthorization.html#method.into_recovered

monad-eth-block-validator/src/lib.rs
```diff
@@ -370,70 +373,92 @@ where
...
+                    first_txn_gas: compute_txn_max_gas_cost(&eth_txn, block_base_fee),
                     max_gas_cost: Balance::ZERO,
-                    max_txn_cost: pre_tfm_compute_max_txn_cost(eth_txn),
+                    max_txn_cost: pre_tfm_compute_max_txn_cost(&eth_txn),
                     is_delegated: false,
                 });
 
             trace!(seq_num = ?header.seq_num, address = ?eth_txn.signer(), nonce = ?eth_txn.nonce(), ?txn_fee_entry, "TxnFeeEntry");
 
             if eth_txn.is_eip7702() {
```
omegablitz, line 444: nit: `if let Some(eth_txn_7702) = eth_txn.as_eip7702() { ... }`
to avoid the unwrap on authorization_list

monad-eth-block-validator/src/lib.rs
```diff
@@ -370,70 +373,92 @@ where
...
-                        if authority == SYSTEM_SENDER_ETH_ADDRESS {
-                            return Err(BlockValidationError::TxnError);
-                        }
+                let authorizations: Vec<RecoveredAuthorization> = eth_txn
+                    .authorization_list()
+                    .expect("EIP7702 tx must have authorization list")
+                    .into_par_iter()
+                    .filter_map(|signed_auth| {
+                        signed_auth.recover_authority().ok().map(|authority| {
+                            RecoveredAuthorization::new_unchecked(
```
omegablitz: nit: authority.clone().into_recovered() to dodge the new_unchecked perhaps?
https://docs.rs/alloy-eip7702/0.6.1/alloy_eip7702/struct.SignedAuthorization.html#method.into_recovered

monad-eth-block-validator/src/lib.rs
```diff
@@ -370,70 +373,92 @@ where
...
-                        // do not allow system account from sending authorization
-                        if authority == SYSTEM_SENDER_ETH_ADDRESS {
-                            return Err(BlockValidationError::TxnError);
-                        }
+                let authorizations: Vec<RecoveredAuthorization> = eth_txn
+                    .authorization_list()
+                    .expect("EIP7702 tx must have authorization list")
+                    .into_par_iter()
+                    .filter_map(|signed_auth| {
+                        signed_auth.recover_authority().ok().map(|authority| {
```
omegablitz: There's no parallelism across transactions here.

I think all of the following should be done in one top-level par_iter:
1. recover tx sender
2. static validate tx
3. parallel recover authorizations (if exists)
4. `collect::<Result<ValidatedTx>, _>()`

And then all the other stateful validation (nonce, system tx, etc) should be done afterwards.

monad-eth-block-validator/src/lib.rs
omegablitz: For that matter, why do we need to check is_eip7702 vs directly checking authorization_list?
Wouldn't the current approach be brittle if there's a new tx type past 7702 that also contains authorization_list?

ariqchowdhury: I think if some new txn type had auth lists too, we would want to check those separately. The rules for auth_list (can't be empty, the intrinsic gas cost etc) are part of the txn type's EIP so we probably don't want to have generic auth_list validation code.

monad-eth-block-validator/src/lib.rs
eerkaijun: I was using new_unchecked because I wanted to filter out the invalid authority (which we will eventually skip so I think it's better to just filter out in the beginning). into_recovered still keep the invalid authority

monad-eth-testutil/src/lib.rs
eerkaijun: https://github.com/category-labs/monad-bft/pull/2299#discussion_r2326708124

monad-eth-block-validator/src/lib.rs
eerkaijun: updated to make recover tx sender and recover authorizations in the top level par_iter. `static_validate_transaction` is still done afterwards as we haven't split into system calls and normal transactions at that point, but let me know if you think it should be done otherwise

omegablitz: I was thinking that intrinsic gas should be checked before recover_authorities, otherwise a validator could include a transaction in a block with a ton of authorizations in order to drive up the signature verification time. For that matter, the same dos vector exists for the tx limit as well, since we do tx limit validation at the end.

The following would accomplish fixing that:
1. check tx limit
1. parallel recover tx senders
1. split/validate alidate system/normal txs
1. static validate normal txs
1. parallel recover authorizations (if exists) on normal txs

However, now that I think about it more, we aren't actually validating that the senders can pay for the intrinsic gas before we start doing the computation... I guess this would need to be moved to eth-block-policy to be able to check intrinsic gas before authorizations, but that sounds more annoying, and we can do that later.

monad-eth-block-validator/src/lib.rs
```diff
@@ -326,15 +350,47 @@ where
...
         for eth_txn in eth_txns.iter() {
             if static_validate_transaction(eth_txn, chain_id, chain_params, execution_chain_params)
                 .is_err()
             {
                 return Err(BlockValidationError::TxnError);
             }
+        }
 
+        let validated_txns: Vec<ValidatedTx> = eth_txns
+            .into_par_iter()
```
eerkaijun, line 363: to be honest not sure if this is overkill, i would expect very few transactions in a block to be 7702 txs

monad-eth-block-validator/src/lib.rs
eerkaijun: tx limit is already checked at the beginning. i also moved block gas limit check to be at the beginning

monad-eth-block-validator/src/lib.rs
omegablitz: The concern isn't about the likeliness - it's more just making sure that our validation time is reflective of that the way stuff is priced, to prevent any DOS vectors.
The worst case without this is that someone constructs 5000 txs each with a few authorizations each. Without the top-level par_iter, at least 5000 of these will be done serially, which is at least 5000 * 30us = 150ms.

eerkaijun: yeah, that makes sense

monad-eth-block-validator/src/lib.rs
```diff
@@ -386,71 +442,50 @@ where
             trace!(seq_num = ?header.seq_num, address = ?eth_txn.signer(), nonce = ?eth_txn.nonce(), ?txn_fee_entry, "TxnFeeEntry");
 
             if eth_txn.is_eip7702() {
```
omegablitz, line 444: It seems like ValidatedTx.authorizations is specifically only populated for 7702 transactions (I understand that 7702 are the only txs that contain authorizations right now).

If so, it would be clearer if they were named authorizations_7702, and if we skipped the is_eip7702 branching everywhere except where ValidatedTx.authorization_list is populated.

Currently we're:
1. Branching on _specifically_ 7702 to populate ValidatedTx.authorizations (note that we're not using [TxEnvelope::authorization_list](https://docs.rs/alloy-consensus/latest/alloy_consensus/transaction/trait.Transaction.html#tymethod.authorization_list) purposefully here)
2. Branching on `eth_txn.is_eip7702()` before using `eth_txn.authorizations` even though it's explicitly only 7702 authorizations

omegablitz (approved):

monad-eth-block-validator/src/lib.rs
eerkaijun: renamed to authorizations_7702

omegablitz (approved):

PR #2305, opened 2025-09-07
Update system transaction functions

bamrith:


monad-system-calls/src/staking_contract.rs
```diff
@@ -69,9 +73,19 @@ impl StakingContractCall {
...
             } => {
-                let mut input = [0_u8; 24];
-                input[0..4].copy_from_slice(Self::REWARD_FUNCTION_SELECTOR.as_slice());
-                input[4..24].copy_from_slice(block_author_address.as_slice());
+                let function_selector_len = Self::REWARD_FUNCTION_SELECTOR.len();
+                let block_author_input = block_author_address.abi_encode();
+                let block_author_len = block_author_input.len();
+                assert_eq!(function_selector_len + block_author_len, 36);
+
+                let mut input = [0_u8; 36];
```
omegablitz: Could you use [concat](https://doc.rust-lang.org/std/primitive.slice.html#method.concat) instead of constructing it manually?

It looks fine to me but it seems less error-prone to do it the other way, and this code is sensitive

monad-system-calls/src/staking_contract.rs
```diff
@@ -109,16 +132,23 @@ impl StakingContractCall {
                 if to != Some(Self::STAKING_CONTRACT_ADDRESS) {
                     return Err(SystemTransactionError::UnexpectedDestAddress);
                 }
-                if input.len() != Self::REWARD_FUNCTION_SELECTOR.len() + block_author_address.len()
-                {
+
+                let function_selector_len = Self::REWARD_FUNCTION_SELECTOR.len();
```
omegablitz: For the validation, would it be clearer/less error-prone to just concat construct the expected input and then check for direct equality?

monad-system-calls/src/staking_contract.rs
omegablitz: After, you could assert that the result of the concat is 36 bytes, for documentation purposes, but doesn't really matter

omegablitz (dismissed):

ariqchowdhury (dismissed):

omegablitz (approved):

PR #2306, opened 2025-09-08
Recover eip7702 authorities during tx insertion

andr-dev:
By recovering authorities during tx insertion (which is paced to not overload the consensus node), this mitigates part of the DoS vector of needing to recover a large number of authorizations at proposal time.

Additionally, the max number of authorizations per tx is limited to 4 in the txpool.

### DIFF
--- a/monad-eth-txpool/src/pool/tracked/sequencer.rs
+++ b/monad-eth-txpool/src/pool/tracked/sequencer.rs
@@ -19,7 +19,6 @@ use std::{
 };
 
 use alloy_consensus::{transaction::Recovered, Transaction, TxEnvelope};
-use alloy_eips::eip7702::Authorization;
 use alloy_primitives::Address;
 use indexmap::IndexMap;
 use monad_chain_config::{revision::ChainRevision, ChainConfig};
@@ -30,43 +29,25 @@ use monad_crypto::certificate_signature::{
 use monad_eth_block_policy::{
     nonce_usage::NonceUsageRetrievable, EthBlockPolicyBlockValidator, EthValidatedBlock,
 };
-use monad_system_calls::SYSTEM_SENDER_ETH_ADDRESS;
 use monad_validator::signature_collection::SignatureCollection;
 use rand::seq::SliceRandom;
 use tracing::{debug, error, trace, warn};
 
 use super::list::TrackedTxList;
-use crate::pool::transaction::ValidEthTransaction;
+use crate::pool::transaction::{ValidEthRecoveredAuthorization, ValidEthTransaction};
 
 #[derive(Debug, PartialEq, Eq)]
 struct OrderedTx<'a> {
     tx: &'a ValidEthTransaction,
-    valid_signed_authorizations: Vec<(Address, &'a Authorization)>,
     effective_tip_per_gas: u128,
 }
 
 impl<'a> OrderedTx<'a> {
-    fn new(tx: &'a ValidEthTransaction, chain_id: u64, base_fee: u64) -> Option<Self> {
+    fn new(tx: &'a ValidEthTransaction, base_fee: u64) -> Option<Self> {
         let effective_tip_per_gas = tx.raw().effective_tip_per_gas(base_fee)?;
-
-        let mut valid_signed_authorizations = Vec::new();
-        for signed_authorization in tx.raw().authorization_list().into_iter().flatten() {
-            if signed_authorization.chain_id != 0 && signed_authorization.chain_id != chain_id {
-                continue;
-            }
-
-            if let Ok(authority) = signed_authorization.recover_authority() {
-                // system account cannot be used to sign authorizations
-                if authority == SYSTEM_SENDER_ETH_ADDRESS {
-                    return None;
-                }
-                valid_signed_authorizations.push((authority, signed_authorization.inner()));
-            }
-        }
 
         Some(Self {
             tx,
-            valid_signed_authorizations,
             effective_tip_per_gas,
         })
     }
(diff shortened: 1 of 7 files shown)


omegablitz:
I think it'd be useful to know the full plan for dealing with the authorization_list DOS before merging this

monad-eth-txpool/src/pool/tracked/sequencer.rs
```diff
@@ -30,43 +29,25 @@ use monad_crypto::certificate_signature::{
...
 #[derive(Debug, PartialEq, Eq)]
 struct OrderedTx<'a> {
     tx: &'a ValidEthTransaction,
-    valid_signed_authorizations: Vec<(Address, &'a Authorization)>,
     effective_tip_per_gas: u128,
 }
 
 impl<'a> OrderedTx<'a> {
-    fn new(tx: &'a ValidEthTransaction, chain_id: u64, base_fee: u64) -> Option<Self> {
+    fn new(tx: &'a ValidEthTransaction, base_fee: u64) -> Option<Self> {
```
omegablitz, line 46: nit: can this return Self now?

monad-eth-txpool/src/pool/tracked/sequencer.rs
```diff
@@ -245,7 +225,11 @@ impl<'a> ProposalSequencer<'a> {
                     self.push(address, next_tx, queued);
                 }
 
-                for (authority, authorization) in tx.valid_signed_authorizations {
+                for ValidEthRecoveredAuthorization {
+                    authority,
+                    authorization,
+                } in tx.tx.iter_valid_recovered_authorizations()
+                {
                     if authorization.chain_id != 0
```
omegablitz, line 233: This shouldn't be possible with the recovery done in iter_valid_recovered_authorizations, correct?

Maybe warn -> error to make it clear that this should never happen? or add "unexpected" to the error message to make it clear that an invariant is broken.

monad-eth-txpool/src/pool/transaction.rs
```diff
@@ -26,13 +28,13 @@ use monad_eth_block_policy::{
...
 use monad_eth_types::EthExecutionProtocol;
-use monad_system_calls::validator::SystemTransactionValidator;
+use monad_system_calls::{validator::SystemTransactionValidator, SYSTEM_SENDER_ETH_ADDRESS};
 use monad_tfm::base_fee::{MIN_BASE_FEE, PRE_TFM_BASE_FEE};
 use monad_types::{Balance, Nonce, SeqNum};
 use monad_validator::signature_collection::SignatureCollection;
 use tracing::trace;
 
-use crate::EthTxPoolEventTracker;
+const EIP7702_MAX_AUTHORIZATION_LIST_LENGTH: usize = 4;
```
omegablitz: nit: this name makes it sound like the limit of 4 is part of EIP7702, but I also don't have a better name suggestion, so we can leave it.

monad-eth-txpool/src/pool/transaction.rs
```diff
@@ -83,24 +90,60 @@ impl ValidEthTransaction {
...
+            return Err((tx, EthTxPoolDropReason::NotWellFormed(error)));
+        }
+
+        let valid_recovered_authorizations = if let Some(signed_authorizations) =
+            tx.authorization_list()
+        {
+            if signed_authorizations.len() > EIP7702_MAX_AUTHORIZATION_LIST_LENGTH {
+                return Err((
+                    tx,
+                    EthTxPoolDropReason::NotWellFormed(TransactionError::AuthorizationListEmpty),
```
omegablitz: does this need a new error type?

monad-eth-txpool/src/pool/transaction.rs
```diff
@@ -83,24 +90,60 @@ impl ValidEthTransaction {
...
+                ));
+            }
+
+            signed_authorizations
+                .iter()
+                .filter_map(|signed_authorization| {
+                    if signed_authorization.chain_id != 0
+                        && signed_authorization.chain_id != chain_id
+                    {
+                        return None;
```
omegablitz: Is a 7702 tx that includes an invalid chain_id considered an invalid transaction? If so, this should be dropped here.

The corresponding proposal creation logic refuses to include these transactions if I understand right

omegablitz (commented):
Tests that the following are dropped with the appropriate error message would be good to prevent regressions:
1. txs with authorizations with invalid chain_id are dropped with the appropriate err
2. txs with long authorization lists are dropped with the appropriate err

Do we already have infrastructure for checking drop reasons in unit-tests? If not, maybe better suited for flexnet?

monad-eth-txpool/src/pool/tracked/sequencer.rs
andr-dev: Nope, the `Option` is being used for filtering out txs that don't meet the block's base fee.
```rust
let effective_tip_per_gas = tx.raw().effective_tip_per_gas(base_fee)?;
```

monad-eth-txpool/src/pool/transaction.rs
andr-dev: Authorizations can have invalid chain ids in which case the authorization is considered invalid but not the tx.

monad-eth-txpool/src/pool/transaction.rs
andr-dev: Moved the `MAX` to be the prefix so it seems less like it's an EIP7702 thing but rather a limit that applies to an EIP7702 thing.

monad-eth-txpool/src/pool/transaction.rs
omegablitz: i read this wrong

monad-eth-txpool/src/pool/tracked/sequencer.rs
omegablitz: read this wrong

omegablitz (approved):

omegablitz (approved):

ariqchowdhury (approved):

PR #2309, opened 2025-09-09
Add EIP7702 tx generator

eclark0426:


monad-eth-testutil/examples/txgen/config.rs
```diff
@@ -414,6 +473,14 @@ fn default_tx_type() -> TxType {
     TxType::ERC20
 }
 
+fn default_total_authorizations() -> usize {
```
JoeHowarth: nit: using functions for default values is non-idiomatic. Prefer inline values if it's only used once or a named const when used multiple times (i.e. `DEFAULT_AUTHORIZATIONS_PER_TX: usize = 1;`

eerkaijun:
https://github.com/category-labs/monad-bft/blob/master/monad-eth-testutil/examples/txgen/shared/erc20.rs#L131 bump this to higher

PR #2312, opened 2025-09-09
Update txpool constants and insert balance check

andr-dev:


omegablitz (approved):

omegablitz:
TODO: we should be able to delete pending now
when polling out of forwarded/ipc, we do the sender recovery, nonce lookup, balance lookup, and then dump straight in tracked

omegablitz:
unit test: populate tracked with 20k addresses, and assert the # of statebackend lookups is < tx_limit

monad-eth-txpool/src/pool/tracked/mod.rs
```diff
@@ -49,20 +49,16 @@ use crate::EthTxPoolEventTracker;
...
 mod sequencer;
 
-// To produce 10k tx blocks, we need the tracked tx map to hold at least 20k addresses so that if
-// the block in the pending blocktree has 10k txs with 10k unique addresses that are also in the
-// tracked tx map then we still have 10k other addresses to use when creating the next block.
-const MAX_ADDRESSES: usize = 20 * 1024;
+// To produce 5k tx blocks, we need the tracked tx map to hold at least 10k addresses so that if the
+// block in the pending blocktree has 5k txs with 5k unique addresses that are also in the tracked
+// tx map then we still have 5k other addresses to use when creating the next block.
+const MAX_ADDRESSES: usize = 10 * 1024;
```
omegablitz: Can leave this as is, because create_proposal state lookups are now bounded by tx_limit

omegablitz (dismissed):

ariqchowdhury (approved):

PR #2316, opened 2025-09-10
Decode delegated flag

mikety:
Implemented reading delegated flag that does not require sync with db changes. It supports the current format with synchronous reading account code from the db.

### DIFF
--- a/monad-triedb-utils/src/lib.rs
+++ b/monad-triedb-utils/src/lib.rs
@@ -176,30 +176,35 @@ impl TriedbReader {
                 let res = maybe_rlp_account.and_then(rlp_decode_account);
                 match res {
                     Some(mut eth_account) => {
-                        debug!(?eth_account, block_id = ?seq_num.0, "account code_hash");
-                        match eth_account.code_hash {
-                            Some(code_hash) => {
-                                // Request code
-                                let (triedb_key, key_len_nibbles) =
-                                    create_triedb_key(version, KeyInput::CodeHash(&code_hash));
-                                let res = self.handle.read(&triedb_key, key_len_nibbles, seq_num.0);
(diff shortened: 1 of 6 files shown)


omegablitz:
Can you add tests for rlp_decode_account with/without is_delegated?

monad-triedb-utils/src/lib.rs
```diff
@@ -176,30 +176,35 @@ impl TriedbReader {
...
-                                    Some(data) => {
-                                        if data.len() >= 3 {
-                                            let delegation_code = &data[0..3];
-                                            eth_account.is_delegated =
-                                                delegation_code == [0xef, 0x01, 0x00];
-                                            if eth_account.is_delegated {
-                                                debug!(?eth_account, block_id = ?seq_num.0, "is_delegated == true");
+                        match eth_account.is_delegated {
+                            Some(_) => Some(eth_account),
+                            None => {
```
eerkaijun, line 181: if this goes in before MONAD_FOUR, is it safe to assume that all `None` means the account is not delegated?

mikety: Yes.

eerkaijun: ok so we can remove everything in the `None` arm if this gets merged before MONAD_FOUR

mikety: `is_delegated` type was changed to add a `None`  for backward compatibility to simplify upgrade. it works even if it merged now to avoid sync on timestamp with execution when it will have the change. The code in decode is also backward compatible for the same reason. If it turnes out to be not needed due to commit merge to main schedule Ill be happy to remove it.

mikety:
due to change on the execution side encoder.

PR #2317, opened 2025-09-10
monad-dataplane: use io_uring batching instead of GSO

dshulyak:
replaced GSO with io_uring batching to split payload and preserve 64KB burst size. multiple UDP sends are batched into single io_uring submission.

change doesn't introduce any new features and tested with original test suite, to verify monoio integration with io uring i run:

```
sudo bpftrace -e 'tracepoint:syscalls:sys_enter_io_uring_enter /args->to_submit > 1/ { printf("%-16s %d ops\n", comm, args->to_submit); }'

bpftrace output:
monad-dataplane  46 ops
monad-dataplane  22 ops
monad-dataplane  21 ops
```

### DIFF
--- a/monad-dataplane/src/udp.rs
+++ b/monad-dataplane/src/udp.rs
@@ -22,9 +22,10 @@ use std::{
 };
 
 use bytes::{Bytes, BytesMut};
+use futures::future::join_all;
 use monoio::{net::udp::UdpSocket, spawn, time};
 use tokio::sync::mpsc;
-use tracing::{debug, error, warn};
+use tracing::{debug, error, trace, warn};
 
 use super::{RecvUdpMsg, UdpMsg};
 use crate::buffer_ext::SocketBufferExt;
@@ -195,18 +196,24 @@ async fn tx(
     mut udp_egress_rx: mpsc::Receiver<(SocketAddr, UdpMsg)>,
     up_bandwidth_mbps: u64,
 ) {
-    let mut udp_segment_size: u16 = DEFAULT_SEGMENT_SIZE;
-    set_udp_segment_size(&socket_tx, udp_segment_size);
-    if let Some(ref direct_socket) = direct_socket_tx {
-        set_udp_segment_size(direct_socket, udp_segment_size);
-    }
-    let mut max_chunk: u16 = max_write_size_for_segment_size(udp_segment_size);
-
     let mut next_transmit = Instant::now();
 
     let mut messages_to_send: VecDeque<(SocketAddr, Bytes, u16, UdpMessageType)> = VecDeque::new();
 
+    let max_batch_bytes = max_write_size_for_segment_size(DEFAULT_SEGMENT_SIZE) as usize;
+    let mut send_futures = Vec::with_capacity(MAX_AGGREGATED_SEGMENTS as usize);
+
     loop {
+        let now = Instant::now();
+        if next_transmit > now {
+            time::sleep(next_transmit - now).await;
+        } else {
+            let late = now - next_transmit;
+            if late > PACING_SLEEP_OVERSHOOT_DETECTION_WINDOW {
+                next_transmit = now;
+            }
+        }
+
         while messages_to_send.is_empty() || !udp_egress_rx.is_empty() {
             let Some((addr, udp_msg)) = udp_egress_rx.recv().await else {
                 return;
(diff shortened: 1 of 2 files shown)


monad-dataplane/src/udp.rs
```diff
@@ -215,116 +222,88 @@ async fn tx(
...
+        {
+            let (addr, mut payload, stride, msg_type) = messages_to_send.pop_front().unwrap();
 
-        // Transmit the first max_chunk bytes of this (addr, payload) pair.
-        let chunk = payload.split_to(payload.len().min(max_chunk.into()));
-        let chunk_len = chunk.len();
+            let chunk_size = payload.len().min(stride as usize);
 
-        let now = Instant::now();
+            if chunk_size + total_bytes > max_batch_bytes {
```
omegablitz, line 237: do we ever assert that egress udp message payloads are < max_write_size_for_segment_size(DEFAULT_SEGMENT_SIZE)?

dshulyak: original behavior was to split message on max_batch_size boundary if stride was set too high, like for example to u16::MAX. i restored that behavior and added a test for that.

that would be a programmer mistake, i think setting a zero will also cause looping.
i will introduce a Stride type in a followup to protect against such misuse

monad-dataplane/src/udp.rs
```diff
@@ -215,116 +222,87 @@ async fn tx(
...
+                            );
+                        }
                     }
                 }
             }
         }
 
-        // the remainder of the message is re-queued only if the send is succesful
-        if ret.is_ok() {
+        if total_bytes > 0 {
```
omegablitz, line 299: We shouldn't see any of the errors above (from send_futs) in practice, correct?
Asking because of the changes to how pacing works when sends fail

monad-dataplane/src/udp.rs
```diff
@@ -215,116 +222,87 @@ async fn tx(
...
+        let mut total_bytes = 0usize;
+        let mut batch_count = 0usize;
+        send_futures.clear();
+
+        while !messages_to_send.is_empty()
+            && total_bytes < max_batch_bytes
+            && batch_count < MAX_AGGREGATED_SEGMENTS as usize
+        {
+            let (addr, mut payload, stride, msg_type) = messages_to_send.pop_front().unwrap();
+            let chunk_size = payload.len().min(stride as usize).min(max_batch_bytes);
```
xinyuan-dev, line 235: does the correctness of the `chunk_size` rely on that `DEFAULT_SEGMENT_SIZE == stride`? consider `DEFAULT_SEGMENT_SIZE < stride`, then `max_batch_bytes` could take on a value that is not an integer multiple of `stride`, thus causing misaligned packet boundary.

monad-dataplane/src/udp.rs
xinyuan-dev, line 299: if no bytes are sent, will this becomes a busy loop?

monad-dataplane/src/udp.rs
dshulyak: i don't understand, what kind of correctness of chunk_size or misalignment of packet boundary? can you give an example?

max_batch_bytes is set to ~64KB to avoid overruning queues with large burst, it can be raised to 128KB or 256KB, but higher might be risky. and also single datagram should not be larger than 64KB.

monad-dataplane/src/udp.rs
dshulyak: yes, all of them are local error, they will not indicate that something is wrong with network

dshulyak: it will wait for new messages if queue is empty

monad-dataplane/src/udp.rs
dshulyak: i think DEFAULT_SEGMENT_SIZE doesn't matter for correctness, for example in this change https://github.com/category-labs/monad-bft/pull/2438 updated nodes will temporarily will use larger chunk size

xinyuan-dev: Suppose we take the extreme value at `(MAX_AGGREGATED_WRITE_SIZE / DEFAULT_SEGMENT_SIZE).min(MAX_AGGREGATED_SEGMENTS) == 1` (i.e. no aggregation), and we let `stride == DEFAULT_SEGMENT_SIZE + 1` and `payload.len() == stride * 2` (two packets). Then, here `chunk_size` would be assigned to `max_batch_bytes == max_write_size_for_segment_size(DEFAULT_SEGMENT_SIZE) == DEFAULT_SEGMENT_SIZE`. Trying to slice the message at this boundary will create three packets: `[0..DEFAULT_SEGMENT_SIZE]`, `[DEFAULT_SEGMENT_SIZE..(DEFAULT_SEGMENT_SIZE*2)]`, and `[(DEFAULT_SEGMENT_SIZE*2)..]` whereas the correctly sliced packets should be at `[0..stride]` and `[stride..]`. The culprit is the `.min(max_batch_bytes)` part.

In general, the boundary could fail if if `(MAX_AGGREGATED_WRITE_SIZE / segment_size).min(MAX_AGGREGATED_SEGMENTS) == N` and `stride > DEFAULT_SEGMENT_SIZE * N`.

On the other hand, if we assume this `N` never <= some largeer number (e.g. 128) such that `stride` is definitely always smaller than `max_batch_bytes`, then `.min(max_batch_bytes)` may also be redundant.

monad-dataplane/src/udp.rs
xinyuan-dev: You're right. The line `Some((addr, udp_msg)) = udp_egress_rx.recv().await` will run if `messages_to_send` is empty.

monad-dataplane/src/udp.rs
```diff
@@ -215,116 +222,87 @@ async fn tx(
...
-        if ret.is_ok() {
+        if total_bytes > 0 {
             next_transmit +=
-                Duration::from_nanos((chunk_len as u64) * 8 * 1000 / up_bandwidth_mbps);
-
-            // Re-queue (addr, payload) at the end of the list if there are bytes left to transmit.
-            if !payload.is_empty() {
-                messages_to_send.push_back((addr, payload, stride, msg_type));
-            }
+                Duration::from_nanos((total_bytes as u64) * 8 * 1000 / up_bandwidth_mbps);
```
xinyuan-dev, line 301: To saturate the maximum bandwidth, should we scale down this duration considering it can take some time to process through the messages? I'm thinking about subtracting this duration by `(Instant::now() - start_of_loop).as_sec()`. If it takes long enough to process through all the packets, we may not need to sleep for that long.

monad-dataplane/src/udp.rs
dshulyak: .min(max_batch_bytes) was added to preserve original behavior, in order to avoid looping on programmer mistake. i think the better fix is to introduce stride type that enforces values are in correct range, e.g they should never before above maximum datagram size

> (MAX_AGGREGATED_WRITE_SIZE /DEFAULT_SEGMENT_SIZE).min(MAX_AGGREGATED_SEGMENTS) == 1
> Then, here chunk_size would be assigned to max_batch_bytes == max_write_size_for_segment_size(DEFAULT_SEGMENT_SIZE) == DEFAULT_SEGMENT_SIZE

do you mean that we changed DEFAULT_SEGMENT_SIZE to be actual maximal possible datagram size? then any stride value above is incorrect and should be either invalidated or capped at DEFAULT_SEGMENT_SIZE. 

i will invalidate it in the follow up, in this change it is capped as in original code.
also note that in the implementation as it exists today we can't have that

monad-dataplane/src/udp.rs
dshulyak: there is no expensive computation anywhere, it should take atmost 1-2us to prepare packets for submission. i will take a look how long it takes in simulation and whether we actually saturate configured bandwidth

monad-dataplane/src/udp.rs
dshulyak: also one more detail that dataplane introduces a limit on received datagram to be equal to ETHERNET_SEGMENT_SIZE, so actually we should never use stride larger than that, but it is somewhat unrelated to this change

monad-dataplane/src/udp.rs
dshulyak: i re-examined the logic and i think it is correct as it is, e.g we account for arbitrary long processing time.
given 1000mbps rate we sleep between sending data for ~522us, this timer is called next_transmit and it is incremented independently. at each iteration we take a current time, check the difference between next_transmit and current time and sleep only for that difference

so if processing took lets say 500us, we will sleep for 522us-500us ~22us

PR #2318, opened 2025-09-10
monad-raptorcast: switch to round robin ordering

dshulyak:
requires: https://github.com/category-labs/monad-bft/pull/2317

to test it i was running simulation on 100 raptorcast processes, and measured delivery latency. in the simulation processes
are divided into 10 buckets, with 20ms latency between each bucket, and proportional latency based on the distance.
so for example from bucket 1 to bucket 5 the latency is 4 * 20ms, 80ms.

distribution across all nodes

image

statistics across all nodes

  | Config               | Count  | Mean  | Std  | P50   | P70   | P90   | P95   | P99   |
  |----------------------|--------|-------|------|-------|-------|-------|-------|-------|
  | Original 2MB 20ms    | 16,533 | 277.4 | 28.1 | 278.1 | 295.6 | 315.0 | 321.4 | 332.9 |
  | Round Robin 2MB 20ms | 46,115 | 260.5 | 28.5 | 261.4 | 277.2 | 296.8 | 304.4 | 321.6 |

### DIFF
--- a/monad-raptorcast/src/udp.rs
+++ b/monad-raptorcast/src/udp.rs
@@ -933,7 +933,6 @@ mod tests {
 
         let mut used_ids: HashMap<SocketAddr, HashSet<_>> = HashMap::new();
 
-        let messages_len = messages.len();
         for (to, mut aggregate_message) in messages {
             while !aggregate_message.is_empty() {
                 let message = aggregate_message.split_to(DEFAULT_SEGMENT_SIZE.into());
@@ -948,7 +947,6 @@ mod tests {
             }
         }
 
-        assert_eq!(used_ids.len(), messages_len);
         let ids = used_ids.values().next().unwrap().clone();
         assert!(used_ids.values().all(|x| x == &ids)); // check that all recipients are sent same ids
         assert!(ids.contains(&0)); // check that starts from idx 0
--- a/monad-raptorcast/src/packet/assembler.rs
+++ b/monad-raptorcast/src/packet/assembler.rs
@@ -36,7 +36,7 @@ use crate::{util::Redundancy, SIGNATURE_SIZE};
 pub enum AssembleMode {
     // Compatible with existing build_messages logic, does not support
     // streaming per merkle batch
-    #[default]
+    #[expect(unused)]
     GsoFull,
 
     // Gso concatenated chunks only within a merkle batch.
@@ -44,7 +44,7 @@ pub enum AssembleMode {
     GsoBestEffort,
 
     // Each recipient gets its own packet in round-robin order.
-    #[expect(unused)]
+    #[default]
     RoundRobin,
 }
 


dixietradesx:
i need to learn

tklenze:
I read through the code and it looks good to me. The idea of using round robin make sense to me. I haven't executed or tested the code.

monad-raptorcast/src/udp.rs
```diff
@@ -409,9 +413,13 @@ pub fn build_messages_with_length<ST>(
     unix_ts_ms: u64,
     build_target: BuildTarget<ST>,
     known_addresses: &HashMap<NodeId<CertificateSignaturePubKey<ST>>, SocketAddr>,
+    rng: &mut R,
```
omegablitz: super nit: in-line the impl blocks into the function params to reduce generic params

monad-raptorcast/src/udp.rs
```diff
@@ -409,9 +413,13 @@ pub fn build_messages_with_length<ST>(
...
+    mut orderer: F,
 ) -> Vec<(SocketAddr, Bytes)>
 where
     ST: CertificateSignatureRecoverable,
+    R: rand::Rng,
+    F: FnMut(&Bytes, usize, u16, VecDeque<(SocketAddr, Range<usize>)>) -> Vec<(SocketAddr, Bytes)>,
```
omegablitz: super nit: FnOnce might describe this more accurately

monad-raptorcast/src/udp.rs
```diff
@@ -365,32 +365,36 @@ const CHUNK_HEADER_LEN: u16 = 20 // Chunk recipient hash
...
+    num_packets: usize,
+    segment_size: u16,
+    mut outbound_gso_idx: VecDeque<(SocketAddr, Range<usize>)>,
+) -> Vec<(SocketAddr, Bytes)> {
+    let mut result = Vec::with_capacity(num_packets);
+    let segment_size = segment_size as usize;
+
+    while let Some((sock, range)) = outbound_gso_idx.pop_front() {
+        let start = range.start;
+        let end = (range.start + segment_size).min(range.end);
```
xinyuan-dev: when is this `min` necessary? if `range.end - range.start < segment_size`, it would suggest a violation of fixed segment size.

dshulyak: i might be wrong, my assumption was that the output is not necessarily the multiple of segment size. e.g it depends on the input, lets say if we encode payload with size 1300 and segment_size is set to 1452, what will be the output with default parameters? i will do a test as well

dshulyak: looks like it wasn't necessary, i pushed update with some tests

monad-raptorcast/src/udp.rs
dshulyak: inlined

monad-raptorcast/src/udp.rs
dshulyak: updated to FnOnce

dshulyak:
@xinyuan-dev @omegablitz rebased pr, please take a look. this change still uses older build_messages, i can close it in favor of one that will use packet::build_messages

monad-raptorcast/src/udp.rs
```diff
@@ -442,9 +409,11 @@ pub fn build_messages_with_length<ST>(
     build_target: BuildTarget<ST>,
     known_addresses: &HashMap<NodeId<CertificateSignaturePubKey<ST>>, SocketAddr>,
     rng: &mut impl rand::Rng,
+    orderer: F,
```
omegablitz: nit: this one can be in-lined too

dshulyak:
@xinyuan-dev @omegablitz updated to use round robin assembler mode, please take a look

PR #2320, opened 2025-09-10
dataplane: batch udp messages for efficient io_uring submission

dshulyak:
small API change to better utilize io_uring batched submissions.
instead of sending multiple potentially small messages to tokio channel, 
we send unicast/broadcast directly into channel and then update the queue.


omegablitz (dismissed):

dshulyak:
superseded by another change

PR #2321, opened 2025-09-10
cli: fix CLI examples and flags for keystore and monad-rpc

FredPhilipy:
I’ve been exploring the project and tried to run a few commands from the docs. I hit some outdated CLI examples, so I’ve fixed them to match the current binaries and flags.

- Update monad-keystore examples in `monad-keystore/src/main.rs` to use subcommands (create/recover/import) with required flags.
- Add copy-pasteable examples for keystore usage.
- Fix `monad-rpc/README.md` to use `--node-config` (hyphen), not `--node_config`.
- Update `monad-keystore/README.md` from legacy `--mode` syntax to subcommand syntax.

Verification:
- Ran `cargo run --bin monad-keystore -- create/recover --help` and commands locally.
- Confirmed clap exposes `--node-config` for monad-rpc via the CLI struct.

### DIFF
--- a/monad-keystore/README.md
+++ b/monad-keystore/README.md
@@ -5,7 +5,7 @@ The keystore CLI tool can be used to generate keystore json files for the BLS ke
 ### Getting Started
 
 ```sh
-cargo run --release -- --mode [create|recover] --key-type [bls|secp] --keystore-path <path_to_output_file>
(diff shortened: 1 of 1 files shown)


FredPhilipy:
Hi there, I noticed that there is no auto assigned reviewer. Would you guys have a look at this? 100% sure that Im not the only one who are struggling here of these wrong CLI example. cc @omegablitz @michael-yxchen @ariqchowdhury (sorry for tagging)

PR #2322, opened 2025-09-10
Remove authority nonce lookup during proposal creation

andr-dev:
During proposal creation, possible nonce usages from authorizations need to be handled to avoid building an invalid tx proposal list. Previously, this was done by loading the nonces of all authorization authorities so that when an authorization is included, we can immediately determine if the nonce of the authority gets increased.

Similar to the NonceUsage change however, instead of resolving the nonce increase immediately, we can store the possible nonce usages in a map and if a tx from the same signer is included, we can then resolve the nonce usages using the invariant that the heap always produces txs that are valid for inclusion after the provided extending blocks and thus the next tx for a signer must have the signer's current account nonce. This reduces the need to lookup the authority nonces, thus bounding the number of lookups at proposal time back to `tx_limit`.

TLDR: This change resolves authority nonce changes in the sequencer, removing the need to query authorization authority nonces and re-bounding the number of lookups at proposal time to `tx_limit`.

### DIFF
--- a/monad-eth-txpool/src/pool/mod.rs
+++ b/monad-eth-txpool/src/pool/mod.rs
@@ -273,7 +273,6 @@ where
         self.tracked.evict_expired_txs(event_tracker);
 
         let timestamp_seconds = timestamp_ns_to_secs(timestamp_ns);
-        let execution_revision = chain_config.get_execution_chain_revision(timestamp_seconds);
 
         {
             let chain_id = chain_config.chain_id();
@@ -286,6 +285,7 @@ where
             }
 
             let chain_revision = chain_config.get_chain_revision(round);
+            let execution_revision = chain_config.get_execution_chain_revision(timestamp_seconds);
 
             if chain_revision.chain_params() != self.chain_revision.chain_params()
                 || self.execution_revision != execution_revision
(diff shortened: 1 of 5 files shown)


monad-eth-txpool/src/pool/mod.rs
```diff
@@ -339,7 +339,11 @@ where
             withdrawals: Vec::new(),
         };
 
-        let maybe_request_hash = if execution_revision.execution_chain_params().prague_enabled {
+        let maybe_request_hash = if self
+            .execution_revision
+            .execution_chain_params()
+            .prague_enabled
+        {
```
andr-dev, line 351: For context, this is a tiny nit from a commit that recently got merged. `execution_revision` used to be in the scope above so it doesn't get used elsewhere, so we use `self.chain_revision` and `self.execution_revision` after they get updated (for `maybe_request_hash`).

omegablitz:
This LGTM, but i have low confidence in my own understanding of 7702 stuff.

omegablitz (approved):

ariqchowdhury:
are there any tests for this?

omegablitz:
Can you also add a test that random authorization addresses don't cause extra nonce lookups?

PR #2336, opened 2025-09-11
Expose event ring descriptor info

andr-dev:
This change allows SDK users to retrieve the `EventDescriptorInfo` associated with an event descriptor to access fields like `record_epoch_nanos`.


omegablitz:
Could you add context to the PR description on the motivation of this? I don't have context on what the use-case of this is

andr-dev:
Execution events SDK is moving to `category-labs/monad`.

PR #2344, opened 2025-09-12
Refactor keystore

bamrith:
- Print the correct private keys for validators to use for staking
- Rename CLI arg (since we only support importing IKMs) and other misleading variable names

### DIFF
--- a/monad-bls/src/bls.rs
+++ b/monad-bls/src/bls.rs
@@ -276,6 +276,15 @@ impl BlsAggregatePubKey {
 #[derive(ZeroizeOnDrop)]
 struct BlsSecretKey(blst_core::SecretKey);
 
+#[derive(ZeroizeOnDrop)]
+pub struct BlsSecretKeyView(Vec<u8>);
+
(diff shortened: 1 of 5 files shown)


monad-bls/src/bls.rs
```diff
@@ -323,6 +327,10 @@ impl BlsKeyPair {
         self.secretkey.0.sign(&msg, DST, &[]).into()
     }
 
+    pub fn privkey_bytes(&self) -> Vec<u8> {
+        self.secretkey.sk_bytes()
+    }
```
bamrith: Exposing this looks bad. Is there a better option ? (same for secp key)

dshulyak: if specifically for printing, maybe custom type with Display implementation that writes hex and zeroizes if it had to store key during print operation. just not on the original type, so that it is harder to misuse

otherwise, if there is no way to get &[u8], then better add custom type with derived zeroize/zeroize on drop so that they don't have be tracked at the callsite

PR #2347, opened 2025-09-12
Txpool forwarding IPC pacing

andr-dev:


omegablitz:
Is it possible to add a test that tries sending an invalid transaction, and checks that no ForwardTxs is emitted? That would help prevent regressions

monad-eth-txpool/src/pool/mod.rs
```diff
@@ -485,17 +485,12 @@ where
...
-                .chain(self.tracked.iter_mut_txs())
-                .filter_map(move |tx| {
-                    tx.get_if_forwardable::<MIN_SEQNUM_DIFF, MAX_RETRIES>(
-                        last_commit_seq_num,
-                        last_commit_base_fee,
-                    )
-                }),
-        )
+        Some(self.tracked.iter_mut_txs().filter_map(move |tx| {
+            tx.get_if_forwardable::<MIN_SEQNUM_DIFF, MAX_RETRIES>(
```
omegablitz, line 489: We can do this separately, but only txs that are guaranteed to be includable would be forwarded (no nonce gaps).

This could be complicated somewhat if there's am extending block that fills in a nonce you're missing, but I don't think that should happen in practice because people are using the same RPC servers between multiple requests.

monad-eth-txpool-executor/src/lib.rs
```diff
@@ -538,17 +536,26 @@ where
...
+            forwarding_manager
+                .as_mut()
+                .project()
+                .add_egress_txs(inserted_txs.iter());
+
             metrics.update(executor_metrics);
             ipc.as_mut().broadcast_tx_events(ipc_events);
-            preload_manager.add_requests(inserted_addresses.iter());
 
+            return Poll::Ready(None);
```
omegablitz: I don't get why this returns Poll::Ready(None) here. Doesn't that indicate the stream is terminated?

monad-eth-txpool-executor/tests/executor.rs
```diff
@@ -0,0 +1,166 @@
...
+        EthTxPoolIpcClient::new(bind_path).await.unwrap();
+
+    assert!(pending.is_empty());
+    assert!(tracked.is_empty());
+
+    (txpool_executor, ipc_client)
+}
+
+#[tokio::test]
+async fn test_ipc_tx_forwarding_pacing() {
```
omegablitz, line 117: I think I'm confused on what this is testing. Is it trying to check that the txs are all forwarded?

Could you also create a test where you feed in an invalid tx, and that no ForwardTxs is emitted? Maybe both one garbage transaction, and then one with too low of a base fee.

monad-eth-txpool/src/pool/mod.rs
andr-dev: Yep agreed, I was going to tackle this one after adding the OOM limit for the txpool. I think it's best to separate it as I'll also want to write quite a few tests for that method.

monad-eth-txpool-executor/tests/executor.rs
andr-dev: The test is showing that when you insert txs over IPC, it doesn't exceed the 256KB limit which is the issue being addressed in this PR. It sends a large number of txs over IPC (all flushed at the same time) and asserts that txpool paces the forwards based on the 256KB limit.

I will add the invalid tx and low base fee tests in the PR where I add these checks to `get_tx_if_forwardable`.

monad-eth-txpool-executor/src/lib.rs
andr-dev: Yeah the better way is to call `cx.waker().wake_by_ref()`. I initially returned `None` because although `Stream` mandates that once a call to `poll_next` produces  `None` all future calls return `None`, that's not how I was using it making this ok although confusing. Updated.

monad-eth-txpool-executor/src/lib.rs
```diff
@@ -214,8 +214,13 @@ where
                     self.exec(commands);
                 }
 
-                event = self.next() => {
-                    if let Err(err) = event_tx.send(event.unwrap()).await {
+                result = self.next() => {
+                    let Some(event) = result else {
```
omegablitz, line 218: is this diff still necessary now that it doesn't return Poll::Ready(None)?

Can it ever return Poll::Ready(None)?

omegablitz (approved):

monad-eth-txpool-executor/src/lib.rs
andr-dev: `EthTxPoolExecutor` never produces `Poll::Ready(None)`, but I think it's better future-proofing to gracefully exit with an error than to panic in the task. In practice, both will eventually result in the node crashing because the `TokioTaskUpdater` has checks for when the `JoinHandle` terminates but at least this way the "executor" is more graceful.

PR #2350, opened 2025-09-14
Shorten testnet2 epoch length

bamrith:


omegablitz:
Can also batch in staking_activation, MONAD_FOUR, and 11 activation 

And submodule

omegablitz (dismissed):

omegablitz (approved):

PR #2352, opened 2025-09-15
dataplane: priority queues for udp traffic

dshulyak:
introduces two priority modes: high and regular.
original api uses regular priority.
all traffic marked as high priority will be delivered before regular, therefore starvation is expected.

the intention is to use high priority for consensus traffic

### DIFF
--- a/monad-dataplane/src/udp.rs
+++ b/monad-dataplane/src/udp.rs
@@ -21,19 +21,60 @@ use std::{
     time::{Duration, Instant},
 };
 
-use bytes::{Bytes, BytesMut};
+use bytes::BytesMut;
 use futures::future::join_all;
 use monoio::{net::udp::UdpSocket, spawn, time};
 use tokio::sync::mpsc;
 use tracing::{debug, error, trace, warn};
 
-use super::{RecvUdpMsg, UdpMsg};
+use super::{RecvUdpMsg, UdpMsg, UdpPriority};
 use crate::buffer_ext::SocketBufferExt;
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Copy)]
 pub(crate) enum UdpMessageType {
     Broadcast,
     Direct,
+}
+
+impl TryFrom<usize> for UdpPriority {
+    type Error = &'static str;
+
+    fn try_from(value: usize) -> Result<Self, Self::Error> {
+        match value {
+            0 => Ok(UdpPriority::High),
+            1 => Ok(UdpPriority::Regular),
+            _ => Err("invalid priority index"),
+        }
+    }
+}
+
+struct PriorityQueues {
+    queues: [VecDeque<UdpMsg>; 2],
+}
+
+impl PriorityQueues {
+    fn new() -> Self {
+        Self {
+            queues: [VecDeque::new(), VecDeque::new()],
+        }
+    }
+
+    fn push(&mut self, msg: UdpMsg) {
+        self.queues[msg.priority as usize].push_back(msg);
+    }
+
+    fn pop_highest_priority(&mut self) -> Option<UdpMsg> {
+        for queue in self.queues.iter_mut() {
+            if let Some(msg) = queue.pop_front() {
+                return Some(msg);
+            }
+        }
+        None
+    }
+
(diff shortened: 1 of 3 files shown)


monad-dataplane/src/udp.rs
```diff
@@ -35,6 +35,47 @@ pub(crate) enum UdpMessageType {
     Direct,
 }
 
+impl From<usize> for UdpPriority {
+    fn from(value: usize) -> Self {
+        match value {
+            0 => UdpPriority::High,
+            _ => UdpPriority::Regular,
```
dshulyak: need to either use try_from and expect on the caller side or panic here

monad-dataplane/src/udp.rs
```diff
@@ -302,10 +342,28 @@ async fn tx(
 
             // Re-queue (addr, payload) at the end of the list if there are bytes left to transmit.
             if !payload.is_empty() {
-                messages_to_send.push_back((addr, payload, stride, msg_type));
+                priority_queues.push(priority, (addr, payload, stride, msg_type.clone()));
```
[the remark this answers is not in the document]
omegablitz: implementing copy on UdpMessageType seems reasonable

monad-dataplane/src/udp.rs
```diff
@@ -204,18 +245,29 @@ async fn tx(
...
+        let now = Instant::now();
+        if next_transmit > now {
+            time::sleep(next_transmit - now).await;
+        } else {
+            let late = now - next_transmit;
 
-            messages_to_send.push_back((addr, udp_msg.payload, udp_msg.stride, udp_msg.msg_type));
+            if late > PACING_SLEEP_OVERSHOOT_DETECTION_WINDOW {
+                next_transmit = now;
+            }
```
omegablitz, line 254: just curious, what was the motivation for moving the pacing block up?

monad-dataplane/src/udp.rs
omegablitz: another option is to:
```rust
    queues: [(UdpPriority, VecDeque<(SocketAddr, Bytes, u16, UdpMessageType)>); 2],
```

i'm fine with what you already have though

omegablitz: another option is `zip_eq(queues, [UdpPriority::High, UdpPriority::Regular])`

monad-dataplane/src/udp.rs
dshulyak: it gives more time for channel to fill up with messages, so if there are concurrent sends with higher priority we are more likely to pick them up. i also did some experiments with such workloads and p99 propagation was consistently slightly lower

and i didn't find any downsides of moving pacing

monad-dataplane/src/udp.rs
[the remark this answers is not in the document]
dshulyak: added copy

monad-dataplane/src/udp.rs
dshulyak: i added try_from

dshulyak:
@omegablitz updated change, please check it out

monad-dataplane/src/udp.rs
```diff
@@ -27,15 +27,59 @@ use monoio::{net::udp::UdpSocket, spawn, time};
...
+        match value {
+            0 => Ok(UdpPriority::High),
+            1 => Ok(UdpPriority::Regular),
+            _ => Err("invalid priority index"),
+        }
+    }
+}
+
+struct PriorityQueues {
+    queues: [VecDeque<(SocketAddr, Bytes, u16, UdpMessageType)>; 2],
```
xinyuan-dev: how about use `(SocketAddr, UdpMsg)` instead of the four-tuple? this way not only we save the unnecessary packing/unpacking steps, it's also easier to read than the bare types.

also, how about moving `SocketAddr` into `UdpMsg` as a field? since they always show up together.

xinyuan-dev (dismissed):

monad-dataplane/src/udp.rs
dshulyak: updated to store UdpMsg

xinyuan-dev (approved):

PR #2355, opened 2025-09-15
More reserve balance unit tests

eerkaijun:
A few more tests for transactions that exceed balance estimate and max reserve balance

### DIFF
--- a/monad-eth-block-policy/src/lib.rs
+++ b/monad-eth-block-policy/src/lib.rs
@@ -1459,7 +1459,7 @@ mod test {
     type ChainConfigType = MockChainConfig;
     type ChainRevisionType = MockChainRevision;
 
-    const RESERVE_BALANCE: u128 = 1_000_000_000_000_000_000;
+    const RESERVE_BALANCE: u128 = 10_000_000_000_000_000_000; // 10 MON
     const EXEC_DELAY: SeqNum = SeqNum(3);
 
     // pubkey starts with AAA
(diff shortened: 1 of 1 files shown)


omegablitz (dismissed):

monad-eth-block-policy/src/lib.rs
```diff
@@ -1765,13 +1765,39 @@ mod test {
...
+
+        let result = setup_block_policy_with_txs(
+            txs,
+            vec![signer],
+            &state_backend,
+            num_committed_blocks,
+            CoherencyCheckMode::ReserveBalanceCoherency,
+        );
+        assert!(
+            result.is_err(),
```
mikety: nit: it will be more informative to check the error explicitly:
 `InsufficientBalance` or  `InsufficientReserveBalance`

monad-eth-block-policy/src/lib.rs
```diff
@@ -1841,14 +1867,41 @@ mod test {
...
```
mikety: ditto re explicit assert on the result.

monad-eth-block-policy/src/lib.rs
```diff
@@ -1920,6 +1973,36 @@ mod test {
...
```
mikety: ditto

monad-eth-block-policy/src/lib.rs
eerkaijun: cool, good point, updated

ariqchowdhury (approved):

PR #2357, opened 2025-09-15
fix(consensus): prevent infinite loop in vote signature verification

### LINKED ISSUE #2345: [BUG] monad-consensus: Signature verification exhaustion via malicious looping in delayed verification
The vulnerability exists in `VoteState::process_vote` implementation where the system uses delayed verification mechanism: after a vote arrives, its signature is not verified immediately. Instead, the `(NodeId, Signature)` pair is first stored in the corresponding vote bucket; only when the total stake in that bucket reaches ≥ 2f + 1 does the code call `SignatureCollection::new()` to batch-verify every signature.

```rust
// Loop to check if super majority votes are reached
while validators
    .has_super_majority_votes(&round_pending_votes.keys().copied().collect::<Vec<_>>())
{
    assert!(round >= self.earliest_round);
    let vote_enc = alloy_rlp::encode(vote);
    match SCT::new::<signing_domain::Vote>(
        round_pending_votes
            .iter()
            .map(|(node, signature)| (*node, *signature)),
        validator_mapping,
        vote_enc.as_ref(),
    ) {
        Ok(sigcol) => {
            let qc = QuorumCertificate::<SCT>::new(vote, sigcol);
            self.earliest_round = round + Round(1);
            return (Some(qc), ret_commands);
        }
        Err(SignatureCollectionError::InvalidSignaturesCreate(invalid_sigs)) => {
            let cmds = Self::handle_invalid_vote(round_pending_votes, invalid_sigs);
            ret_commands.extend(cmds);
        }
        Err(_) => {
            unreachable!("InvalidSignaturesCreate is only expected error from creating SC");
        }
    }
}
```

The core issue is that the system uses delayed verification mechanism, first storing votes in buckets without immediately verifying signature validity. Only when the bucket's weight reaches super majority (≥2f+1) does it trigger batch verification. When `SignatureCollection::new()` returns `InvalidSignaturesCreate` error, the system removes invalid signatures but continues to loop and check if super majority condition is still satisfied, with no limit on the number of repeated verifications within the same round.

An attacker first injects a forged signature into a vote bucket, then waits for honest validators to naturally accumulate votes until the bucket's weight reaches ≥2f+1. This triggers batch verification which will fail due to the invalid signature. The system then removes the malicious entry, but since other honest votes remain, the weight may still satisfy the super majority condition, allowing the attacker to immediately inject another forged signature, triggering verification again and forming an infinite loop.

In each loop iteration, the system executes expensive operations including RLP encoding vote data (`alloy_rlp::encode(vote)`), batch BLS signature verification (`SignatureCollection::new()`), and traversing and processing all pending votes.

A malicious validator can cause temporary CPU exhaustion by repeatedly injecting invalid signatures when vote buckets reach super majority, forcing expensive batch verification operations in a loop.

hoank101:
fix #2345


michael-yxchen:
https://github.com/category-labs/monad-bft/issues/2345#issuecomment-3294006353

if you think otherwise, can you write a test to demonstrate?

hoank101:
hi @michael-yxchen i added some test, help me check it

michael-yxchen:
I don't believe the tests demonstrate the attack. I made `process_vote` panic here
```rust
  if verification_attempts >= MAX_VERIFICATION_ATTEMPTS {
      panic!("failed test");
      error!(
          round = ?vote.round,
          epoch = ?vote.epoch,
          verification_attempts,
          "Maximum verification attempts reached, aborting QC creation to prevent DoS"
      );
      break;
  }
```
None of the tests fail with that change. In fact, you can lower `MAX_VERIFICATION_ATTEMPTS` to 2 and all tests still pass. That's proof that the loop's failure path is executed at most once in the given scenarios

michael-yxchen:
closing this PR as the "infinite loop" mentioned in the issue is not within `process_vote` function, but in the main event loop if the attacker repeatedly submit votes with invalid signatures. Please see the issue for discussion

PR #2358, opened 2025-09-15
perf: optimize transaction pool hot paths and memory layout

hoank101:


ariqchowdhury:
there's no benchmarks showing that anything was improved.

PR #2361, opened 2025-09-16
Reject reserved ips in peer discovery

eerkaijun:
Closes https://github.com/category-labs/category-internal/issues/1777, audit issue raised by Zellic. This PR rejects IP addresses that are special addresses. For reference, geth also has similar checks https://github.com/ethereum/go-ethereum/blob/master/p2p/netutil/net.go#L185-L222

### DIFF
--- a/monad-peer-discovery/src/ipv4_validation.rs
+++ b/monad-peer-discovery/src/ipv4_validation.rs
@@ -1,0 +1,101 @@
+// Copyright (C) 2025 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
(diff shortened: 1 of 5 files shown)


monad-peer-discovery/src/ip_check.rs
```diff
@@ -0,0 +1,148 @@
...
+#[derive(Debug, PartialEq)]
+pub enum IpCheckError {
+    UnspecifiedIp,
+    SpecialIP,
+    PrivateIp,
+    LoopbackIp,
+    LinkLocalIp,
+}
+
+pub fn check_socket_address(
```
ariqchowdhury: nit: rename to `validate_socket_ip4_address`

ariqchowdhury: file name should also be re-named, ipv4_validation.rs

eerkaijun: updated

monad-peer-discovery/src/ipv4_validation.rs
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
```
dshulyak: rstest is very convenient to use for such tests

dshulyak (approved):

monad-peer-discovery/src/ipv4_validation.rs
eerkaijun: great point, updated

dshulyak (approved):

ariqchowdhury (approved):

PR #2363, opened 2025-09-16
Remove pending pool from txpool

andr-dev:


omegablitz:
Could you add a blurb on the new flow of transactions and what's done where (static validation, forwarding, eviction, etc)?
Would make it easier to go through the PR

omegablitz:
Thanks - with the new model, what's the plan on dealing with nonce-gapped transactions? Assuming we move towards a model where only transactions that are guaranteed to be includable are forwarded.

Would we only accept transactions into tracked if they have a valid includable nonce?

andr-dev:
> Thanks - with the new model, what's the plan on dealing with nonce-gapped transactions?

The `TrackedTxList` has a `get_queued` method which is used during sequencing to get the list of non-nonce-gapped transactions. It takes an optional account nonce override to handle extending block nonce changes, but for forwarding we can pass `None` and it will produce the list of in-nonce-order transactions (without gaps) based on the last committed block.

monad-eth-txpool/src/pool/mod.rs
```diff
@@ -188,65 +178,64 @@ where
...
-                .tracked
-                .try_insert_tx(event_tracker, tx)
-                .unwrap_or_else(|tx| {
-                    self.pending
-                        .try_insert_tx(event_tracker, tx, last_commit_base_fee)
-                })
-            else {
+        for (address, txs) in txs {
+            let Some(account_nonce) = account_nonces.remove(&address) else {
+                event_tracker.drop_all(
```
omegablitz, line 234: drop_all and then continue seems weird

Although iiuc, this branch should never get hit

omegablitz (dismissed):

monad-eth-txpool/src/pool/mod.rs
omegablitz: nvm, txs is shadowed

ariqchowdhury (approved):

ariqchowdhury (approved):

PR #2364, opened 2025-09-17
Add support to dynamically reload prioritized full nodes

eerkaijun:
Closes https://github.com/category-labs/category-internal/issues/1919. Tested on a single node. Extend support for the monad-debug-node reload config functionality such that we can also update prioritized full nodes without needing to restart the node.

Usage:
```
monad-debug-node --control-panel-ipc-path controlpanel.sock reload-config
```

### DIFF
--- a/monad-peer-discovery/src/discovery.rs
+++ b/monad-peer-discovery/src/discovery.rs
@@ -1335,6 +1335,17 @@ where
         cmds
     }
 
+    fn update_pinned_nodes(
+        &mut self,
+        pinned_nodes: BTreeSet<NodeId<CertificateSignaturePubKey<ST>>>,
+    ) -> Vec<PeerDiscoveryCommand<ST>> {
+        debug!(?pinned_nodes, "updating pinned nodes");
+
+        self.pinned_full_nodes = pinned_nodes;
+
+        Vec::new()
+    }
+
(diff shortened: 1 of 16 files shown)


fabiotdf (dismissed):

monad-peer-discovery/src/discovery.rs
```diff
@@ -1306,7 +1306,11 @@ where
         cmds
     }
 
-    fn update_peers(&mut self, peers: Vec<PeerEntry<ST>>) -> Vec<PeerDiscoveryCommand<ST>> {
+    fn update_peers(
+        &mut self,
+        peers: Vec<PeerEntry<ST>>,
+        pinned_full_nodes: Option<BTreeSet<NodeId<CertificateSignaturePubKey<ST>>>>,
+    ) -> Vec<PeerDiscoveryCommand<ST>> {
```
michael-yxchen: i think we can create a new function for updating pinned_full_nodes. won't need the `option` if done separately

eerkaijun: you mean to use a separate PeerDiscoveryEvent?

michael-yxchen: yeah separate event & interface function

PR #2366, opened 2025-09-17
Fix eth_FeeHistory

philipglazman:
Implement a todo feature for returning fee history for a range of blocks.

### DIFF
--- a/monad-rpc/src/handlers/eth/gas.rs
+++ b/monad-rpc/src/handlers/eth/gas.rs
@@ -18,9 +18,10 @@ use std::{
     sync::Arc,
 };
 
-use alloy_consensus::{Header, Transaction as _, TxEnvelope};
+use alloy_consensus::{Header, Transaction, TxEnvelope};
 use alloy_primitives::{Address, TxKind, U256, U64};
-use alloy_rpc_types::FeeHistory;
+use alloy_rpc_types::{FeeHistory, TransactionReceipt};
+use itertools::Itertools;
 use monad_ethcall::{CallResult, EthCallExecutor, MonadTracer, StateOverrideSet};
 use monad_rpc_docs::rpc;
 use monad_triedb_utils::triedb_env::{BlockKey, FinalizedBlockKey, ProposedBlockKey, Triedb};
@@ -422,12 +423,7 @@ pub async fn monad_eth_feeHistory<T: Triedb>(
         .await
         .map_err(|_| JsonRpcError::internal_error("could not get block data".into()))?;
 
-    let base_fee_per_gas = header.base_fee_per_gas.unwrap_or_default();
-    let gas_used_ratio = (header.gas_used as f64).div(header.gas_limit as f64);
-    let blob_gas_used = header.blob_gas_used.unwrap_or_default();
-    let blob_gas_used_ratio = (blob_gas_used as f64).div(header.gas_limit as f64);
-
-    let reward = match params.reward_percentiles {
+    let percentiles = match params.reward_percentiles {
         Some(percentiles) => {
             // Check percentiles are between 0-100
             if percentiles.iter().any(|p| *p < 0.0 || *p > 100.0) {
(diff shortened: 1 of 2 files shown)


michael-yxchen:
Cross-posting offline discussion: `eth_feeHistory` first fetches `blockCount` entries for everything, then calculate `baseFeePerGas/baseFeePerBlobGas` for the next block, as all information is known from the previous block. In our case, the fields to compute next base fee are in consensus blocks, which RPC can't access right now

Latest state we're exposing through RPC is `voted`, implying that the next block is always executed and base_fee is available in eth blocks in db. In the short term, we will fetch the base_fee directly from the next block.

When we expose `proposed` block, RPC can either read from consensus ledger folder or from the event system (optionally enabled currently)

monad-rpc/src/handlers/eth/gas.rs
```diff
@@ -446,27 +442,128 @@ pub async fn monad_eth_feeHistory<T: Triedb>(
...
 
-    // TODO: retrieve fee parameters from historical blocks. For now, return a hacky default
+    let oldest_block = header.number.saturating_sub(block_count);
+    let mut base_fee_per_gas_history = Vec::with_capacity(block_count as usize + 1);
+    let mut gas_used_ratio_history = Vec::with_capacity(block_count as usize);
+    let mut rewards = Vec::with_capacity(block_count as usize + 1);
+
+    for (idx, blk_num) in (oldest_block..=header.number).enumerate() {
+        let block = chain_state
+            .get_block(
```
eerkaijun: just a thought: will block count be something high like 100? if so, do we need to read these blocks concurrently like how we do it for eth_getLogs?

monad-rpc/src/handlers/eth/gas.rs
```diff
@@ -446,27 +442,128 @@ pub async fn monad_eth_feeHistory<T: Triedb>(
...
+    // Get the reward and gas used for each transaction using receipt.
+    let gas_and_rewards = transactions
+        .iter()
+        .zip(receipts)
+        .map(|(tx, receipt)| {
+            let gas_used = receipt.gas_used;
+            let reward = tx.effective_tip_per_gas(base_fee).unwrap_or_default();
+            (gas_used, reward)
+        })
+        .sorted_by_key(|(gas_used, _)| *gas_used)
```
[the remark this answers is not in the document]
philipglazman: Geth sorts by gas used

monad-rpc/src/handlers/eth/gas.rs
```diff
@@ -446,27 +442,128 @@ pub async fn monad_eth_feeHistory<T: Triedb>(
...
+        .collect::<Vec<_>>();
+
+    let mut idx = 0;
+    let mut cumulative_gas_used: u128 = 0;
+    let mut rewards = Vec::new();
+
+    for pct in percentiles.unwrap() {
+        let gas_threshold = (block_gas_used as f64 * pct / 100.0).round() as u128;
+        while cumulative_gas_used < gas_threshold && idx < transactions.len() - 1 {
```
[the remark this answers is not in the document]
philipglazman: fixed with clamping the index range after

omegablitz:
Is returning the base fee from the highest available eth block in db sufficient? Do we need to calculate it for the next block?

monad-rpc/src/handlers/eth/gas.rs
```diff
@@ -435,27 +437,182 @@ pub async fn monad_eth_feeHistory<T: Triedb>(
...
 
-    // TODO: retrieve fee parameters from historical blocks. For now, return a hacky default
+    let oldest_block = header.number.saturating_sub(block_count - 1);
+    let mut base_fee_per_gas_history: Vec<u128> = Vec::with_capacity(block_count as usize + 1);
+    let mut gas_used_ratio_history = Vec::with_capacity(block_count as usize);
+    let mut rewards = Vec::with_capacity(block_count as usize + 1);
+
+    let block_range = oldest_block..=header.number;
+
+    let block_data_futures = block_range.map(|blk_num| async move {
```
eerkaijun, line 456: this can read up to a max of 1024 blocks concurrently, might block other rpc call? would it make sense to do something like `stream.iter(block_range).map(...).buffered(10)` so that we limit concurrent reads at once?

philipglazman: Good point, updated to make it buffered.

eerkaijun (dismissed):

monad-rpc/src/handlers/eth/gas.rs
andr-dev: Although most RPC provider docs suggest only [1, 1024] block_counts are supported, I believe implementations support `0` block_counts, in which case `alloy_rpc_types_eth::FeeHistory::default()` should be produced.

eerkaijun (dismissed):

michael-yxchen (approved):

PR #2367, opened 2025-09-17
Fix waker usage

michael-yxchen:
- use clone_from when updater has a waker in place
- wake up the task when relevant internal state changes

https://github.com/category-labs/category-internal/issues/1727
https://github.com/category-labs/monad-bft/issues/1940

### DIFF
--- a/monad-updaters/src/ledger.rs
+++ b/monad-updaters/src/ledger.rs
@@ -217,6 +217,9 @@ where
                             self.get_headers(block_range),
                         ),
                     });
+                    if let Some(waker) = self.waker.take() {
+                        waker.wake();
+                    }
                 }
                 LedgerCommand::LedgerFetchPayload(payload_id) => {
                     self.events.push_back(BlockSyncEvent::SelfResponse {
@@ -224,6 +227,9 @@ where
                             self.get_payload(payload_id),
                         ),
                     });
+                    if let Some(waker) = self.waker.take() {
+                        waker.wake();
+                    }
                 }
             }
         }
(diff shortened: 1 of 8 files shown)


monad-updaters/src/ledger.rs
```diff
@@ -241,12 +247,16 @@ where
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         let this = self.deref_mut();
 
+        if let Some(waker) = this.waker.as_mut() {
+            waker.clone_from(cx.waker());
+        } else {
+            this.waker = Some(cx.waker().clone());
+        }
```
andr-dev, line 261: This doesn't need to be moved up before the `pop_front` since it always returns a `Poll::Ready` which schedules the `poll_next` fn to be called again so there's no need for a waker in that case.

monad-updaters/src/val_set.rs
```diff
@@ -194,14 +192,12 @@ where
...
+            waker.clone_from(cx.waker());
+        } else {
             this.waker = Some(cx.waker().clone());
         }
 
-        if this.ready() {
-            this.waker.take().unwrap().wake();
-        }
-
         event
```
andr-dev: The way this method was structured is confusing, and with this change we're always setting the waker even when we produce `Poll::Ready` which is useless. Let's refactor to something like:
```rust
if let Some(next_val_data) = this.next_val_data.take() else {
    return Poll::Ready(Some(MonadEvent::ValidatorEvent(
        monad_executor_glue::ValidatorEvent::<SCT>::UpdateValidators(next_val_data),
    )));
}

if let Some(waker) = this.waker.as_mut() {
    waker.clone_from(cx.waker());
} else {
    this.waker = Some(cx.waker().clone());
}

Poll::Pending
```

monad-updaters/src/val_set.rs
```diff
@@ -361,14 +355,12 @@ where
             Poll::Pending
         };
```
andr-dev: Same refactor here.

PR #2369, opened 2025-09-18
Propagate RaptorCast's initial epoch from forkpoint.toml

xinyuan-dev:
a small fix to https://github.com/category-labs/monad-bft/pull/2092#discussion_r2343186514.

### DIFF
--- a/monad-router-multi/src/lib.rs
+++ b/monad-router-multi/src/lib.rs
@@ -109,6 +109,7 @@ where
             shared_pdd.clone(),
             recv_net_messages,
             send_group_infos,
+            current_epoch,
         );
 
         let mut rc_primary = RaptorCast::new(
@@ -116,6 +117,7 @@ where
             dp_reader,
             dp_writer.clone(),
             shared_pdd.clone(),
+            current_epoch,
         );
         rc_primary.bind_channel_to_secondary_raptorcast(send_net_messages, recv_group_infos);
 
@@ -132,8 +134,12 @@ where
         }
     }
 
-    fn update_role(&mut self, new_role: SecondaryRaptorCastModeConfig) {
-        debug!(?new_role, "Updating secondary raptorcast role");
+    fn update_role(&mut self, current_epoch: Epoch, new_role: SecondaryRaptorCastModeConfig) {
+        debug!(
+            ?new_role,
+            ?current_epoch,
+            "Updating secondary raptorcast role"
+        );
         self.rc_config.secondary_instance.mode = new_role;
 
         // create new channels
@@ -156,6 +162,7 @@ where
             self.shared_pdd.clone(),
             recv_net_messages,
             send_group_infos,
+            current_epoch,
         );
         self.rc_secondary = rc_secondary;
     }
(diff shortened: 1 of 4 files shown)


monad-router-multi/src/lib.rs
```diff
@@ -270,7 +275,7 @@ where
                             .is_some_and(|set| set.contains(&self.self_node_id))
                     {
                         // validator demoted to full node, update role to be Client
-                        self.update_role(SecondaryRaptorCastModeConfig::Client);
+                        self.update_role(epoch, SecondaryRaptorCastModeConfig::Client);
```
eerkaijun: to be more accurate, this would be `self.current_epoch` instead of `epoch`. when the new validator set is made available, epoch is not yet bumped

monad-router-multi/src/lib.rs
```diff
@@ -132,8 +134,8 @@ where
         }
     }
 
-    fn update_role(&mut self, new_role: SecondaryRaptorCastModeConfig) {
-        debug!(?new_role, "Updating secondary raptorcast role");
+    fn update_role(&mut self, new_epoch: Epoch, new_role: SecondaryRaptorCastModeConfig) {
```
eerkaijun: i think this can be named as `current_epoch` instead of `new_epoch`

monad-router-multi/src/lib.rs
```diff
@@ -317,9 +325,14 @@ where
                             && is_validator_after_epoch_increment
                         {
                             // full node promoted to validator, update role to be Publisher
-                            self.update_role(SecondaryRaptorCastModeConfig::Publisher);
+                            self.update_role(
+                                self.current_epoch,
```
eerkaijun: this would be `epoch`. it's a bit confusing because we call `update_role` before epoch is bumped when transitioning from validator to full node, but we only call `update_role` after epoch is bumped when transitioning from full node to validator

xinyuan-dev: I see. how about it now? https://github.com/category-labs/monad-bft/pull/2369/commits/4fcbb00a44e75b449ccfc039c5b5fc1ae8dbe3db

monad-router-multi/src/lib.rs
```diff
@@ -294,6 +303,8 @@ where
                     if epoch > self.current_epoch {
                         // epoch increment
                         debug!(?self.current_epoch, ?epoch, "Epoch incremented in multi router");
+                        self.current_epoch = epoch;
+
```
eerkaijun: self.current_epoch needs to be updated between `is_validator_before_epoch_increment` and `is_validator_after_epoch_increment`

monad-router-multi/src/lib.rs
```diff
@@ -302,7 +313,7 @@ where
                             ?is_validator_before_epoch_increment,
                             "Is validator before epoch increment"
                         );
-                        self.current_epoch = epoch;
+
```
eerkaijun: if we recover this line and remove line 306 it would be correct

monad-router-multi/src/lib.rs
eerkaijun: https://github.com/category-labs/monad-bft/pull/2369#discussion_r2359838085

monad-router-multi/src/lib.rs
xinyuan-dev: ah I restored it to the wrong place. please check again.

eerkaijun (approved):

PR #2371, opened 2025-09-18
Populate eth header fields in events header

eerkaijun:
These are fields that are in [ProposedEthHeader](https://github.com/category-labs/monad-bft/blob/653c193fa1410810f8ae29369af1ca9dc44dd648/monad-eth-types/src/lib.rs#L57). The block header from events set them as None while the block header from triedb set them based on ProposedEthHeader. I think we should be consistent as our actual block header for rpc response. 

We don't use these fields though, so they are always set to the default zero. These fields should perhaps come from `monad_exec_block_start`, but i'm not sure if that's good use of memory since these are always set to the hardcoded defaults anyways.


philipglazman (dismissed):

omegablitz:
I'm confused, are those fields not accessible in the events header?

eerkaijun:
i don't see those fields in the events stream, @kjcamann correct me if i'm wrong? 

edit: spoke with ken and at some point the execution events schema will be changed to include these fields. for now, I just hardcode these values since it's always constant

eerkaijun:
closes https://github.com/category-labs/category-external/issues/96

philipglazman (approved):

eerkaijun:
@omegablitz at some point ken will change the schema of the execution events to include this field, for now i think it's safe to hardcode since these are unused fields in the block header and will always be set to zero

monad-exec-events/src/block.rs
```diff
@@ -70,10 +70,10 @@ impl ExecutedBlock {
 
                 (!withdrawals_root.const_is_zero()).then_some(withdrawals_root)
             },
-            blob_gas_used: None,
-            excess_blob_gas: None,
-            parent_beacon_block_root: None,
-            requests_hash: None,
+            blob_gas_used: Some(0),
```
omegablitz, line 73: For context, why isn't this passed in via events?
https://github.com/category-labs/monad/blob/90f8b796061aeaf78a2943c45eae5303f6ff7900/category/execution/ethereum/core/eth_ctypes.h#L127-L144

omegablitz: never mind, I just saw the above discussion.

omegablitz (approved):

PR #2372, opened 2025-09-18
Set txpool forwarding max byte limit to pool limit

andr-dev:
Previously, the `EGRESS_MAX_SIZE_BYTES` limit was soft in nature to allow a single large tx that exceeds the size limit to still get forwarded.

Since there is a tx limit checked in `ValidEthTransaction`, this change makes `EGRESS_MAX_SIZE_BYTES` a function of that max limit, raising the previous limit of 256KB to 384KB. Now, exceeding this limit will cause the tx to get dropped with a warn (as this would constitute a forwarding bug).

Resolves https://github.com/category-labs/category-internal/issues/1914.

### DIFF
--- a/monad-eth-txpool-executor/src/forward.rs
+++ b/monad-eth-txpool-executor/src/forward.rs
@@ -22,16 +22,18 @@ use std::{
 
 use alloy_consensus::TxEnvelope;
 use bytes::Bytes;
-use monad_chain_config::{revision::ChainRevision, ChainConfig};
+use monad_chain_config::{
+    execution_revision::ExecutionChainParams, revision::ChainRevision, ChainConfig,
+};
 use monad_crypto::certificate_signature::{
     CertificateSignaturePubKey, CertificateSignatureRecoverable,
 };
-use monad_eth_txpool::EthTxPool;
+use monad_eth_txpool::{max_eip2718_encoded_length, EthTxPool};
 use monad_eth_types::ExtractEthAddress;
 use monad_state_backend::StateBackend;
 use monad_validator::signature_collection::SignatureCollection;
 use pin_project::pin_project;
-use tracing::debug;
+use tracing::{debug, error};
 
 const EGRESS_MIN_COMMITTED_SEQ_NUM_DIFF: u64 = 5;
 const EGRESS_MAX_RETRIES: usize = 3;
@@ -39,7 +41,10 @@ const EGRESS_MAX_RETRIES: usize = 3;
 const INGRESS_CHUNK_MAX_SIZE: usize = 128;
 const INGRESS_CHUNK_INTERVAL_MS: u64 = 8;
 const INGRESS_MAX_SIZE: usize = 8 * 1024;
-const EGRESS_MAX_SIZE_BYTES: usize = 256 * 1024;
+
+pub fn egress_max_size_bytes(execution_params: &ExecutionChainParams) -> usize {
+    max_eip2718_encoded_length(execution_params)
+}
 
 #[pin_project(project = EthTxPoolForwardingManagerProjected)]
 pub struct EthTxPoolForwardingManager {
@@ -52,8 +57,8 @@ pub struct EthTxPoolForwardingManager {
     egress_waker: Option<Waker>,
 }
 
-impl EthTxPoolForwardingManager {
-    pub fn new() -> Self {
+impl Default for EthTxPoolForwardingManager {
+    fn default() -> Self {
         let mut ingress_timer =
             tokio::time::interval(Duration::from_millis(INGRESS_CHUNK_INTERVAL_MS));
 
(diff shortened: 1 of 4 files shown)


monad-eth-txpool-executor/src/forward.rs
```diff
@@ -103,26 +106,39 @@ impl EthTxPoolForwardingManager {
...
-        for (i, tx) in egress.iter().enumerate() {
-            let tx_size = tx.len();
-            if total_size + tx_size > EGRESS_MAX_SIZE_BYTES && i > 0 {
-                break;
+            let mut drain_count = 0;
+
+            for tx in egress.iter() {
+                let tx_size = tx.len();
+
+                let Some(new_remaining_size) = remaining_size.checked_sub(tx_size) else {
```
ariqchowdhury: nit: just use tx.len() directly

monad-eth-txpool-executor/src/forward.rs
```diff
@@ -103,26 +106,39 @@ impl EthTxPoolForwardingManager {
...
+                }
+                return Poll::Pending;
+            };
 
-        let mut total_size = 0;
-        let mut drain_count = 0;
+            let Some(mut remaining_size) = EGRESS_MAX_SIZE_BYTES.checked_sub(tx.len()) else {
+                error!("txpool forwarding manager detected tx larger than max tx byte size, skipping forwarding");
+                continue;
+            };
```
ariqchowdhury: why do we have to pop_front and check the first txn individually? 
why not `let mut remaining_size = EGRESS_MAX;` and then just have the for loop below cover all txns?
if you're concerned about individual txns that exceed the size, can't the for loop check and log an error on any of the txns exceeding the limit?

andr-dev: We're using `drain` to efficiently remove the first `drain_count` elements from the front of the `VecDeque`. If we log an error and "skip" in the middle, we'd have to build the Vec manually by pushing bytes onto it. The performance loss with that method is likely negligible so I'm happy to go with that.

monad-eth-txpool-executor/src/forward.rs
```diff
@@ -477,44 +515,109 @@ mod test {
...
+        let (forwarding_manager, mut cx) = setup();
+        let mut forwarding_manager = pin!(forwarding_manager);
+
+        let tx1 = make_legacy_tx(S1, BASE_FEE_PER_GAS, 100_000, 0, 0);
+        assert!(
+            tx1.eip2718_encoded_length()
+                <= egress_max_size_bytes(EXECUTION_REVISION.execution_chain_params())
+        );
+
+        let tx2 = make_legacy_tx(
```
mikety: nit: adding eip1559 txs would increase coverage

mikety (approved):
lgtm with nits

monad-eth-txpool-executor/src/forward.rs
```diff
@@ -96,33 +103,61 @@ impl EthTxPoolForwardingManager {
...
-        for (i, tx) in egress.iter().enumerate() {
-            let tx_size = tx.len();
-            if total_size + tx_size > EGRESS_MAX_SIZE_BYTES && i > 0 {
                 break;
             }
-            total_size += tx_size;
-            drain_count = i + 1;
+
+            if txs.is_empty() {
+                let tx = egress.pop_front();
```
omegablitz, line 151: when can egress be non-empty here?

andr-dev: If it's ever non-empty, then there's a bug somewhere.

This line provides insurance against a class of DOS attacks where a tx can "block" the forwarding manager from scheduling transactions. If that ever happens, this line acts as an "insurance" that we'll skip said tx from being forwarded and continue trying to forward other transactions.

omegablitz: How could a tx block the forwarding manager? I guess I'm just confused at what the class of DOS vector this protects against is.

To me it looks like this branch happening would an imply a bug in this file, and not a bug anywhere else.

andr-dev: If a tx is larger than `egress_max_size_bytes`, without the error + **pop** + continue at L141-145, forwarding would halt.
```rust
if tx.len() > egress_max_size_bytes {
    error!("txpool forwarding manager detected tx larger than max tx byte size, skipping forwarding");
    egress.pop_front();
    continue;
}
```
This is a secondary safeguard to that.

It would be a bug that manifest through this file's definition of `egress_max_size_bytes` that is most likely to be caused by txpool accepting a tx larger than `egress_max_size_bytes`. Theoretically, `egress_max_size_bytes` should be a function of that upper txpool limit, but it could diverge.

omegablitz: What are the tradeoffs btwn checking egress_max_size_bytes here vs in add_egress_txs?

omegablitz (approved):
LGTM, I think poll_egress might be simpler if egress_max_size_bytes is checked upon adding a tx instead of upon polling

monad-eth-txpool-executor/src/forward.rs
andr-dev: No real tradeoffs, I'll just sleep better at night knowing super cheap if statements like this one that are very easy to reason about exist to catch possible issues down the line. In fact, I do like the idea of adding another check in `egress_max_size_bytes` as well.

PR #2374, opened 2025-09-19
Only limit tx byte size in txpool

andr-dev:


omegablitz (dismissed):

monad-eth-txpool/src/pool/transaction.rs
```diff
@@ -35,6 +35,10 @@ use tracing::trace;
 
 const MAX_EIP7702_AUTHORIZATION_LIST_LENGTH: usize = 4;
 
+pub const fn max_eip2718_encoded_length(execution_params: &ExecutionChainParams) -> usize {
+    2 * execution_params.max_code_size + 128 * 1024
```
mikety, line 39: Is this confirmed to depend on the chain params ( that can change ? )

monad-eth-txpool/tests/pool.rs
```diff
@@ -1365,6 +1367,101 @@ fn test_large_batch_many_senders() {
...
+            .get_execution_chain_revision(0)
+            .execution_chain_params(),
+    );
+
+    let tx1 = make_legacy_tx(
+        S1,
+        BASE_FEE,
+        TFM_MAX_GAS_LIMIT,
+        0,
+        max_eip2718_encoded_length - 111,
```
mikety, line 1434: Is there any name for the magic number 110?

mikety (approved):
lgtm with nits

monad-eth-txpool/tests/pool.rs
andr-dev: It could have a name, it's just the overhead of the additional bytes required to encode everything but the input data length hence why I didn't make it a named variable.

monad-eth-txpool/src/pool/transaction.rs
andr-dev: Yep! The idea is to limit transaction size as a function of input size. The max code size is used to give a max length for the input data, and we add 128KB over that to account for things like access lists, authorization lists, etc.

PR #2376, opened 2025-09-22
Improve eth mock-swarm StateBackendTest trait

ariqchowdhury:
We currently have a trait StateBackendTest which requires the implementer to provide a ledger_propose and ledger_commit function. The InMemory state backend implementation implements both the StateBackend and StateBackendTest trait for use in tests. 

The StateBackendTest trait is trying to emulate some aspects of execution, but does not do it in a general way. It only takes in block headers and requires the user to generate the valid set of accounts nonces to pass in. So for nonces, the “execution” logic of nonces is split between the called of ledger_propose and the ledger_propose implementation. Balances are defaulted to MAX so that balance calculations don’t interfere with other tests. 

This PR proposes to change the StateBackendTest trait to MockExecution and to move any execution related work to the implementers of MockExecution. The ledger_propose signature is modified to take in the list of transaction in the block along with the block header. 

This initial change only implements nonces in the existing InMemory backend to ensure that all existing tests pass. mock-swarm already contained nonce checking tests and this PR also added a basic 7702 sanity nonce test. 

The mapping for accounts to nonce in InMemory is changed to map account to AccountState so that in the future, balance can also be tracked. This changes the signature for InMemory’s genesis function so many of the diffs in the PR are just updating the existing tests. 

The next change will to be add proper balance accounting as well as the validations of the TFM spec. 

Additionally, a future change will split the InMemory backend into a InMemoryTFM and an InMemoryNoExec. Many of the existing tests that use StateBackendTest now do not care about the nonce/balance execution results, so they can use the simpler InMemoryNoExec.

interesting changes are in:
monad-eth-ledger/src/lib.rs
monad-mock-swarm/tests/nonces.rs
monad-state-backend/*

rest are updating the existing tests for the new interface


monad-state-backend/src/in_memory.rs
```diff
@@ -172,20 +209,63 @@ where
...
+            "block N={:?}, parent account state: {:?}",
+            seq_num,
+            accounts
+        );
+
+        for tx in txns.iter() {
+            let addr = tx.recover_signer().expect("invalid eth tx in block");
+
+            let account_entry = accounts.entry(addr).or_insert(AccountState {
+                balance: Balance::from(100_000_000_000_000_000_u64),
```
ariqchowdhury, line 237: the default behaviour before was for accounts to have large balances since most tests only cared about nonces, so maintaining that behaviour with this. Future PR that does balance/tfm simulation will have new accounts at 0.

eerkaijun (approved):

monad-state-backend/src/in_memory.rs
```diff
@@ -172,20 +209,63 @@ where
...
+                let auth_list = tx
+                    .authorization_list()
+                    .expect("valid 7702 must have auth list");
+                for tuple in auth_list {
+                    if let Ok(addr) = tuple.recover_authority() {
+                        let auth_account_entry = accounts.entry(addr).or_insert(AccountState {
+                            balance: Balance::from(100_000_000_000_000_000_u64),
+                            reserve_balance: self.default_reserve_balance,
+                            nonce: 0,
+                            is_delegated: false,
```
michael-yxchen, line 259: Not sure if matters for mock, but when do we update account state is_delegated to true?

eerkaijun: great point, i updated the logic in a subsequent pr, but let me just pull it over here for completeness (it's a one line change)

eerkaijun: updated

michael-yxchen (approved):

PR #2377, opened 2025-09-23
RaptorCast: Modular packet assembly

xinyuan-dev:
Design doc: https://www.notion.so/Modular-packet-assembly-28075b0ba8408067b8f9ecf9092d56d4

Benchmark result:

| Test Case              | Time (mean) | Throughput (mean) |
|------------------------|-------------|-------------------|
| Raptorcast 128K / old  | 546.91 µs   | 228.56 MiB/s      |
| Raptorcast 128K / new  | 551.27 µs   | 226.75 MiB/s      |
| Raptorcast 2M / old    | 9.0575 ms   | 220.81 MiB/s      |
| Raptorcast 2M / new    | 8.9950 ms   | 222.34 MiB/s      |
| Broadcast 128K / old   | 52.624 ms   | 237.53 MiB/s      |
| Broadcast 128K / new   | 50.306 ms   | 248.48 MiB/s      |

<details>
<summary>Full Benchmark Logs</summary>

```
Raptorcast 128K/udp::build_messages
                        time:   [541.17 µs 546.91 µs 555.23 µs]
                        thrpt:  [225.13 MiB/s 228.56 MiB/s 230.98 MiB/s]
                 change:
                        time:   [+0.8834% +2.1325% +4.2982%] (p = 0.00 < 0.05)
                        thrpt:  [-4.1210% -2.0880% -0.8757%]
Found 13 outliers among 100 measurements (13.00%)
  5 (5.00%) high mild
  8 (8.00%) high severe
Raptorcast 128K/packet::build_messages
                        time:   [549.25 µs 551.27 µs 554.04 µs]
                        thrpt:  [225.62 MiB/s 226.75 MiB/s 227.58 MiB/s]
                 change:
                        time:   [+0.9739% +1.6640% +2.7696%] (p = 0.00 < 0.05)
                        thrpt:  [-2.6949% -1.6368% -0.9645%]
Found 9 outliers among 100 measurements (9.00%)
  3 (3.00%) high mild
  6 (6.00%) high severe

Raptorcast 2M/udp::build_messages
                        time:   [8.9445 ms 9.0575 ms 9.1865 ms]
                        thrpt:  [217.71 MiB/s 220.81 MiB/s 223.60 MiB/s]
                 change:
                        time:   [-0.9076% +0.9565% +2.8338%] (p = 0.33 > 0.05)
                        thrpt:  [-2.7557% -0.9474% +0.9159%]
                        No change in performance detected.
Found 6 outliers among 100 measurements (6.00%)
  2 (2.00%) high mild
  4 (4.00%) high severe
Raptorcast 2M/packet::build_messages
                        time:   [8.9289 ms 8.9950 ms 9.0688 ms]
                        thrpt:  [220.54 MiB/s 222.34 MiB/s 223.99 MiB/s]
                 change:
                        time:   [-2.3773% -0.6885% +0.9216%] (p = 0.43 > 0.05)
                        thrpt:  [-0.9132% +0.6932% +2.4352%]
                        No change in performance detected.
Found 15 outliers among 100 measurements (15.00%)
  7 (7.00%) high mild
  8 (8.00%) high severe

Benchmarking Broadcast 128K/udp::build_messages: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 5.3s, or reduce sample count to 90.
Broadcast 128K/udp::build_messages
                        time:   [52.438 ms 52.624 ms 52.810 ms]
                        thrpt:  [236.70 MiB/s 237.53 MiB/s 238.38 MiB/s]
                 change:
                        time:   [-2.7506% -1.1749% -0.1138%] (p = 0.07 > 0.05)
                        thrpt:  [+0.1139% +1.1888% +2.8284%]
                        No change in performance detected.
Benchmarking Broadcast 128K/packet::build_messages: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 5.1s, or reduce sample count to 90.
Broadcast 128K/packet::build_messages
                        time:   [50.093 ms 50.306 ms 50.522 ms]
                        thrpt:  [247.42 MiB/s 248.48 MiB/s 249.54 MiB/s]
                 change:
                        time:   [-0.1627% +0.6978% +1.4534%] (p = 0.09 > 0.05)
                        thrpt:  [-1.4326% -0.6929% +0.1629%]
                        No change in performance detected.
```
</details>

How it's tested:

- existing features are tested to be compatible with the old `udp::build_message` function, which itself is extensively tested in `udp.rs`
- benchmarking metrics is attached above to show no performance regression
- new features are checked with unit tests
  + round-robin packet assembly (implements https://github.com/category-labs/monad-bft/pull/2318)
  + stake-based with rounding (implements https://github.com/category-labs/monad-bft/pull/2215)

### DIFF
--- a/monad-raptorcast/src/packet/mod.rs
+++ b/monad-raptorcast/src/packet/mod.rs
@@ -1,0 +1,417 @@
+// Copyright (C) 2025 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+mod assembler;
+mod assigner;
+
+use std::{collections::HashMap, net::SocketAddr};
+
+use assembler::AssembleMode;
+use bytes::Bytes;
+use monad_crypto::certificate_signature::{
+    CertificateKeyPair as _, CertificateSignaturePubKey, CertificateSignatureRecoverable, PubKey,
+};
+use monad_types::NodeId;
+use rand::seq::SliceRandom as _;
+
+pub(crate) use self::{
+    assembler::{assemble, BroadcastType, Chunk, PacketLayout, Recipient},
+    assigner::ChunkAssigner,
+};
+use crate::util::{BuildTarget, Redundancy};
+
+#[derive(Debug)]
+pub struct UdpMessage {
+    pub dest: SocketAddr,
+    pub payload: Bytes,
+    pub stride: usize,
+}
+
+#[derive(Debug)]
+pub enum BuildError {
+    // merkle tree depth does not fit in u4
+    MerkleTreeTooDeep,
+    // chunk id does not fit in u16
+    ChunkIdOverflow,
+    // failed to create encoder
+    EncoderCreationFailed,
+    // too many chunks
+    TooManyChunks,
+    // app message is too large
+    AppMessageTooLarge,
+    // total stake is zero
+    ZeroTotalStake,
+    // redundancy is too high
+    RedundancyTooHigh,
+}
+
+pub(crate) trait PeerAddrLookup<PT: PubKey> {
+    fn lookup(&self, node_id: &NodeId<PT>) -> Option<SocketAddr>;
+}
+
+// Similar to std::iter::Extend trait but implemented for FnMut as
+// well.
+pub(crate) trait Collector<T> {
+    fn push(&mut self, item: T);
+    fn reserve(&mut self, _additional: usize) {}
+}
+
+type Result<A, E = BuildError> = std::result::Result<A, E>;
+
+// A compatible interface to crate::udp::build_messages that uses
+// packet::assemble underneath.
+#[allow(clippy::too_many_arguments)]
+pub fn build_messages<ST>(
+    key: &ST::KeyPairType,
+    segment_size: u16,
+    app_message: Bytes,
+    redundancy: Redundancy,
+    epoch_no: u64,
+    unix_ts_ms: u64,
+    build_target: BuildTarget<ST>,
+    known_addresses: &HashMap<NodeId<CertificateSignaturePubKey<ST>>, SocketAddr>,
+    rng: &mut impl rand::Rng,
+) -> Vec<(SocketAddr, Bytes)>
+where
+    ST: CertificateSignatureRecoverable,
+{
+    use self::assigner::{Partitioned, Replicated};
+
+    let broadcast_type = match build_target {
+        BuildTarget::Raptorcast { .. } => BroadcastType::Primary,
+        BuildTarget::FullNodeRaptorCast { .. } => BroadcastType::Secondary,
+        _ => BroadcastType::Unspecified,
+    };
+    let peer_lookup = known_addresses;
+
+    let assigner: Box<dyn ChunkAssigner<_>> = match build_target {
+        BuildTarget::PointToPoint(to) => Box::new(Replicated::from_unicast(*to)),
+        BuildTarget::Broadcast(ref nodes_view) => Box::new(Replicated::from_broadcast(
+            nodes_view.iter().copied().collect(),
+        )),
+        BuildTarget::Raptorcast(ref validators_view) => {
+            let mut validator_set: Vec<_> = validators_view
+                .iter()
+                .map(|(node_id, stake)| (*node_id, stake))
+                .collect();
+            validator_set.shuffle(rng);
+            Box::new(Partitioned::from_validator_set(validator_set))
+        }
+        BuildTarget::FullNodeRaptorCast(group) => {
+            let self_id = NodeId::new(key.pubkey());
+            let seed = rng.gen::<usize>();
+            let nodes = group
+                .iter_skip_self_and_author(&self_id, seed)
+                .copied()
+                .collect();
+            Box::new(Partitioned::from_homogeneous_peers(nodes))
+        }
+    };
+
+    let app_message_len = app_message.len();
+    let mut packets = Vec::new();
+    let result = assemble::<ST, _, _>(
+        key,
+        segment_size as usize,
+        app_message,
+        redundancy,
+        epoch_no,
+        unix_ts_ms,
+        broadcast_type,
+        AssembleMode::GsoFull,
+        peer_lookup,
+        &*assigner,
+        &mut packets,
+    );
+
+    let packets = match result {
+        Ok(()) => packets,
+        Err(BuildError::TooManyChunks) => {
+            tracing::error!(
+                ?app_message_len,
+                ?redundancy,
+                ?build_target,
+                "Too many chunks generated."
+            );
+            return vec![];
+        }
+        Err(BuildError::AppMessageTooLarge) => {
+            tracing::error!(?app_message_len, "App message too large");
+            return vec![];
+        }
+        Err(BuildError::ZeroTotalStake) => {
+            tracing::error!(?build_target, "Total stake is zero");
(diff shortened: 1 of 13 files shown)


monad-raptorcast/src/udp.rs
```diff
@@ -410,9 +446,11 @@ pub fn build_messages_with_length<ST>(
     unix_ts_ms: u64,
     build_target: BuildTarget<ST>,
     known_addresses: &HashMap<NodeId<CertificateSignaturePubKey<ST>>, SocketAddr>,
+    rng: &mut R,
```
omegablitz: super nit: you can elide this generic parameter `rng: &mut impl rand::Rng`

dshulyak:
i noticed in simulations and while working on authentication that it will be beneficial perf-wise to output chunks incrementally. one example is to avoid waiting for all chunks to be signed. once first chunk is signed and ready to be sent it can be offloaded to dataplane and assembler will proceed in parallel. ofcourse, some parts will have to complete before any of that.

this approach will utilize bandwidth slightly better and will result in lower latency for proposals delivery. i tried that with encryption as well. parallelization, that you mentioned in doc will be good too, but outputting first chunk as soon as you can will give best results overall.

would it be possible to provide iterator-like interface to consume chunks?

monad-raptorcast/src/packet/mod.rs
omegablitz, line 1: Just to be consistent, could you move this to monad-raptorcast/src/packet/mod.rs?
We tend to use mod.rs elsewhere

monad-raptorcast/src/udp.rs
xinyuan-dev: changed!

xinyuan-dev:
> would it be possible to provide iterator-like interface to consume chunks?

Implemented! The last `assemble` function now accepts a callback function or a container (via `collector: &mut impl Collector<UdpMessage>`) that will be called as soon as a merkle batch is signed and assembled. This interface should be equivalent to an iterator.

Note that the streaming mode is only validly defined for certain "serialization modes". For example, the round-robin mode would be unavailable for the streaming because it would be impossible to implement round-robin on small batches.

dshulyak:
> Note that the streaming mode is only validly defined for certain "serialization modes". For example, the round-robin mode would be unavailable for the streaming because it would be impossible to implement round-robin on small batches.

thats unfortunate, i had a hunch that it should be possible by caching signature and outputting first chunk early, but i might be completely wrong. i am also reviewing a change in general, will approve later in a day

monad-raptorcast/src/packet/mod.rs
```diff
@@ -0,0 +1,418 @@
...
+};
+use crate::util::{BuildTarget, Redundancy};
+
+#[derive(Debug)]
+pub struct UdpMessage {
+    pub dest: SocketAddr,
+    pub payload: Bytes,
+    pub stride: usize,
+    // 0 - highest priority, 255 - lowest priority, None - unspecified
+    pub priority: Option<u8>,
```
dshulyak: where priority is used? 

or what is the idea for it? i personally don't see why different chunks of the same payload will be prioritized differently, if they need to be sent in a specific order thats another matter

monad-raptorcast/src/packet/mod.rs
```diff
@@ -0,0 +1,418 @@
...
+    }
+
+    #[allow(unreachable_code)]
+    fn assert_compatible_eq(expected: &Packets, actual: &Packets) {
+        assert_eq!(expected.len(), actual.len());
+
+        // Set the environment variable to show detailed difference in
+        // packet bytes.
+        let (expected, actual) = (normalize(expected), normalize(actual));
+        if option_env!("SCRUTIZE_PACKET") != Some("1") {
```
dshulyak: is it a typo ? SCRUTINIZE ?

monad-raptorcast/src/packet/serializer.rs
```diff
@@ -0,0 +1,374 @@
...
+                    stride,
+                }
+            }
+        }
+
+        let stride = layout.segment_len();
+        let mut agg_chunk = None;
+
+        for chunk in chunks {
+            let Some(dest) = chunk.recipient.get_addr() else {
```
dshulyak: original code also skipped chunks silently? i assume thats when we don't have information in local peer table

problems in this area will be very hard to debug without some information

monad-raptorcast/src/packet/serializer.rs
```diff
@@ -0,0 +1,374 @@
...
+        chunks: Vec<Chunk<PT>>,
+        layout: PacketLayout,
+    ) {
+        #[derive(Hash, PartialEq, Eq, Clone, Copy)]
+        struct GroupKey {
+            dest: SocketAddr,
+            priority: Option<u8>,
+        }
+
+        let mut messages: HashMap<GroupKey, BytesMut> = HashMap::new();
```
dshulyak: original implementation was deterministic or it was using hash map as well like this? what it was using for order if it was deterministic?

monad-raptorcast/src/packet/serializer.rs
```diff
@@ -0,0 +1,374 @@
...
+        for chunk in chunks {
+            let Some(dest) = chunk.recipient.get_addr() else {
+                continue;
+            };
+
+            buckets.entry(dest).or_default().push_back(chunk);
+        }
+
+        // Each recipient get their own queue of chunks.
+        let mut queues: Vec<VecDeque<Chunk<PT>>> = buckets.into_values().collect();
```
dshulyak: it converts to vec from hashmap for order? would it make a difference if it was stored in btree and then iteration below is on btree

xinyuan-dev: yes I used vec mainly for the ability of going through it over repeatedly while deleting recipients with empty queues. `retain_mut` is perfect for this purpose. it should be able to do the same on btreemap with its `retain` method. however, i expect using btreemap to allocating more than hashmap/vec approach due to its lack of `with_capacity` constructor, also i expect hashmap/vec to perform better generally due to cache efficiency. that said, those are my speculations and i can benchmark both versions.

monad-raptorcast/src/packet/serializer.rs
xinyuan-dev: original implementation was deterministic because it relies on the chunks assigned to be in a specific order so that each recipients' chunks are always adjacent. the new framework allows more versatile chunk assignment algorithms by lifting this restraint. therefore the order of recipients outputted from the round-robin serializer would not be the same as the original order. but the round-robin property is inaffected, and for each recipient, their received packets are still in the correct order.

monad-raptorcast/src/packet/serializer.rs
xinyuan-dev: sure, let me add some debugging logs to this.

monad-raptorcast/src/packet/mod.rs
xinyuan-dev: fixed, thanks.

monad-raptorcast/src/packet/serializer.rs
dshulyak: i just wanted to be sure that we don't break some implicit order assumption, that wasn't well documented.
like for example ordering them by stake-weight

monad-raptorcast/src/packet/mod.rs
xinyuan-dev: one place i know priority is expected is with rounding chunks. in the researcher proposal they suggested that those rounding chunks can be assigned a lower priority with a future raptorcast improvement.

meanwhile i think it's a useful field to have because we may assign different priority to different types of messages. or we can assign those redundant chunks with a lower priority so those be chunks are delivered first. of course we'll need to have dataplane to prioritize different packets based on the field first, how exactly it may work can be discussed in the future.

dshulyak: i think if some chunks can be sent later they shouldn't have a priority field that dataplane will have to be aware of, instead they should just appear later in the final order of the packet assembly.

distinct priority is useful for payloads that are sent concurrently with each other and so can't be aware of each other, like discovery vs consensus.

i would rather pull it out from this pr if there are no features that depend on it at the moment. in my opinion we have a use  case to prioritize based on protocols, like consensus vs rest. and within single payload assembler can just output them in the desired order instead

xinyuan-dev: okay it makes sense to add it later when needed. i removed priority support altogether.

PR #2378, opened 2025-09-23
Refactor to allow opt in to secondary raptorcast

eerkaijun:
Closes https://github.com/category-labs/category-internal/issues/1872

### DIFF
--- a/monad-node-config/src/fullnode_raptorcast.rs
+++ b/monad-node-config/src/fullnode_raptorcast.rs
@@ -21,16 +21,9 @@ use super::fullnode::FullNodeConfig;
 
 #[derive(Debug, Deserialize, Serialize, Clone)]
 #[serde(deny_unknown_fields)]
-pub enum SecondaryRaptorCastModeConfig {
-    Client,
-    Publisher,
-    None, // Disables secondary raptorcast
-}
-
-#[derive(Debug, Deserialize, Serialize, Clone)]
-#[serde(deny_unknown_fields)]
 pub struct FullNodeRaptorCastConfig<P: PubKey> {
-    pub mode: SecondaryRaptorCastModeConfig,
+    pub enable_publisher: bool,
+    pub enable_client: bool,
 
     #[serde(bound = "P:PubKey")]
     pub full_nodes_prioritized: FullNodeConfig<P>,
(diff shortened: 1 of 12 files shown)


monad-peer-disc-swarm/tests/peer_discovery.rs
```diff
@@ -1023,13 +1004,7 @@ fn test_validator_demoted_to_full_node() {
...
     let config = TestConfig {
         num_nodes: 5,
-        roles: BTreeMap::from([
-            (0, PeerDiscoveryRole::ValidatorPublisher),
-            (1, PeerDiscoveryRole::ValidatorPublisher),
-            (2, PeerDiscoveryRole::ValidatorPublisher),
-            (3, PeerDiscoveryRole::ValidatorNone),
-            (4, PeerDiscoveryRole::ValidatorNone),
-        ]),
+        roles: BTreeMap::from([(0, true), (1, true), (2, true), (3, false), (4, true)]),
```
michael-yxchen, line 1007: nit: I think it should be `(4, false)`?

monad-node-config/src/fullnode_raptorcast.rs
```diff
@@ -19,7 +19,7 @@ use serde::{Deserialize, Serialize};
 
 use super::fullnode::FullNodeConfig;
 
-#[derive(Debug, Deserialize, Serialize, Clone)]
+#[derive(Debug, Deserialize, Serialize, Clone, Copy)]
 #[serde(deny_unknown_fields)]
 pub enum SecondaryRaptorCastModeConfig {
```
michael-yxchen: super nit (can ignore): looks like we can move this out of `monad-node-config` crate?

michael-yxchen (dismissed):

monad-peer-disc-swarm/tests/peer_discovery.rs
eerkaijun: setting as true so that it becomes a public full node when falling out of the validator set (previously it always default to becoming a public full node when falling out of active set)

monad-node-config/src/fullnode_raptorcast.rs
eerkaijun: great point, moved into secondary raptorcast

monad-peer-disc-swarm/tests/peer_discovery.rs
michael-yxchen: cool got it

fabiotdf (approved):

eerkaijun:
tested on stressnet

PR #2383, opened 2025-09-24
Txpool Limits

andr-dev:


omegablitz:
The current structuring is something like this:
```rust
pub struct TrackedTxMap {
    last_commit: Option<ConsensusBlockHeader<ST, SCT, EthExecutionProtocol>>,
    txs: IndexMap<Address, TrackedTxList>,
    limits: TrackedTxLimits,

    _phantom: PhantomData<(SBT, CCT, CRT)>,
}

pub struct TrackedTxLimits {
    config: TrackedTxLimitsConfig,

    addresses: usize,
    txs: usize,
}

impl TrackedTxLimits {
    pub fn prepare_add_tx_to_new_address<'et, 'l, 't>(
        &'l mut self,
        event_tracker: &'et mut EthTxPoolEventTracker<'t>,
        tx: ValidEthTransaction,
    ) -> Option<TrackedTxLimitAddToken<'et, 'l, 't>> {
        if !self.can_increase_limits(1, 1) {
            event_tracker.drop(tx.hash(), EthTxPoolDropReason::PoolFull);
            return None;
        }

        Some(TrackedTxLimitAddToken::new(
            TrackedTxLimitAddTokenKind::NewAddress,
            self,
            event_tracker,
            tx,
        ))
    }
}
```

I think it's confusing as currently designed how TrackedTxMap::txs and TrackedTxLimits::limits have their own standalone statefulness that need to be kept in sync by calling the functions at the right times. I understand the TrackedTxLimitAddToken ties them together to some extent, but I'm thinking it'd be simpler to just tie the limits together with the datastructure.

Something like this perhaps:
```rust
pub struct TrackedTxMapState {
    last_commit: Option<ConsensusBlockHeader<ST, SCT, EthExecutionProtocol>>,
    txs: TrackedTxMap,

    _phantom: PhantomData<(SBT, CCT, CRT)>,
}

pub struct TrackedTxMap {
    // stateless config
    config: TrackedTxLimitsConfig,
    
    // stateful stuff
    txs: IndexMap<Address, TrackedTxList>,
    num_tx: usize,
}
```

What do you think?

PR #2400, opened 2025-09-29
Log support in RPC

Baltoli:


monad-rpc/src/handlers/debug.rs
```diff
@@ -268,6 +294,31 @@ pub struct MonadDebugTraceTransactionParams {
     tracer: TracerObject,
 }
 
+#[derive(Serialize, Debug, schemars::JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct MonadCallFrameLog {
+    address: EthAddress,
+    topics: Vec<FixedData<32>>,
+    data: UnformattedData,
+    position: Quantity,
```
philipglazman, line 294: `position` should be u64. Quantity serializes to a hex string, but the response should be an integer

monad-rpc/src/handlers/debug.rs
```diff
@@ -268,6 +294,31 @@ pub struct MonadDebugTraceTransactionParams {
     tracer: TracerObject,
 }
 
+#[derive(Serialize, Debug, schemars::JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct MonadCallFrameLog {
```
philipglazman, line 290: Missing `index` field which is the index of the log in the trace. 

`Position` is the position of the log relative to subcalls in the same trace

monad-rpc/src/handlers/debug.rs
Baltoli: I had it as a `u64` but that conflicted with Anvil. Should I carve it out as a trace comparison exception?

monad-rpc/src/handlers/debug.rs
Baltoli: Will add, thanks

monad-archive/src/model/block_data_archive.rs
```diff
@@ -535,6 +535,28 @@ pub enum CallKind {
     StaticCall,
 }
 
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct CallFrameLog {
```
omegablitz, line 539: `#[derive(Clone, Debug, PartialEq, Eq, RlpEncodable, RlpDecodable)]`

and can remove the manual implementation

monad-rpc/src/handlers/debug.rs
```diff
@@ -161,6 +161,21 @@ pub async fn monad_debug_getRawTransaction<T: Triedb>(
     }
 }
 
+#[derive(Clone, Debug)]
```
omegablitz: ```#[derive(Clone, Debug. RlpEncodable, RlpDecodable)]```

and can remove the manual impl

monad-rpc/src/handlers/debug.rs
```diff
@@ -567,6 +624,22 @@ pub async fn decode_call_frame<T: Triedb>(
         .flatten()
         .collect::<Vec<_>>();
 
+    for frame in call_frames.iter_mut() {
+        if tracer.config.with_log {
+            // If logs were requested and there were none stored with the call
```
philipglazman, line 620: Is there a reason we return an error? Anvil/Geth does not return an error if the transaction's call traces do not have logs

Baltoli: To avoid requiring a hard execute-from-genesis reset for this change, I made the RLP backwards-compatible. This error distinguishes "old tx we didn't record events for, but might have had them" from "tx with no logs". Can change this behaviour to whatever you think is most reasonable

monad-rpc/src/handlers/debug.rs
```diff
@@ -292,6 +343,8 @@ pub struct MonadCallFrame {
     #[serde(skip_serializing_if = "Vec::is_empty")]
     #[schemars(skip)] // TODO: handle recursive generation in jsonrpc schema
     calls: Vec<std::rc::Rc<std::cell::RefCell<MonadCallFrame>>>,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    logs: Vec<MonadCallFrameLog>,
```
omegablitz, line 338: If with_log is true, is it possible for logs to be empty? If yes, then should logs be emitted as an empty list or no?

Am just checking that this shouldn't be Option<Vec<_>>

omegablitz (dismissed):

andr-dev (dismissed):

andr-dev (approved):

PR #2401, opened 2025-09-29
Issue #1771: Full nodes blocksync peers

fabiotdf:


monad-blocksync/src/blocksync.rs
```diff
@@ -162,8 +162,16 @@ where
 
     self_request_mode: BlockSyncSelfRequester,
 
+    // When false, the set of override peers comes from config and can only be
+    // changed with set_override_peers(). Otherwise, it will be changing dynamically
+    // with sef_fallback_override_peers(), as the node receives ConfirmGroup messages.
+    override_peers_is_dynamic: bool,
```
eerkaijun: I think we can make this to be something like `secondary_raptorcast_peers`, and then we select blocksync peers based on the following rule: 
- if blocksync override peer is set, always blocksync from these override peers
- if they are not set, and secondary raptorcast peers is not empty, blocksync from secondary raptorcast peers (validators should have `secondary_raptorcast_peers` to be empty, i.e. we don't need to send the groups for Publisher) 
- otherwise randomly select from current validator set and blocksync (which is already the current behavior before this PR)

monad-blocksync/src/blocksync.rs
```diff
@@ -189,28 +197,52 @@ where
...
+        override_peers_inc_self: Vec<NodeId<CertificateSignaturePubKey<ST>>>,
     ) {
-        let override_peers = override_peers.into_iter().unique().collect();
+        let override_peers: Vec<_> =
+            override_peers_inc_self
+            .iter()
+            .cloned()
+            .filter(|peer| peer != &self.our_nodeid)
+            .collect();
+        self.override_peers_is_dynamic = override_peers.is_empty();
```
eerkaijun: why do we need this?

monad-executor-glue/src/lib.rs
```diff
@@ -1917,6 +1924,11 @@ where
     ConfigUpdate(ConfigUpdate<SCT>),
     KnownPeersUpdate(KnownPeersUpdate<ST>),
     LoadError(String),
+    // These peers are used as a fallback for requesting blocksync from when
+    // blocksync_override_peers is empty, rather than using from random validators
+    // which we don't have valid peer entries for yet. These are filled when
+    // we receive a ConfirmGroup message.
+    DynamicOverridePeers(DynamicOverridePeersUpdate<SCT>),
```
eerkaijun: can this be put into RaptorcastEvent? feels like it's not part of ConfigEvent

monad-blocksync/src/blocksync.rs
```diff
@@ -190,26 +201,51 @@ where
...
+            .iter()
+            .cloned()
+            .filter(|peer| peer != &self.our_nodeid)
+            .collect();
+        self.override_peers = peers_excl_self;
+    }
+
+    pub fn set_secondary_raptorcast_peers(
+        &mut self,
+        override_peers_inc_self: Vec<NodeId<CertificateSignaturePubKey<ST>>>,
```
eerkaijun: nit: rename this to differentiate with override peers?

monad-blocksync/src/blocksync.rs
```diff
@@ -190,26 +201,51 @@ where
...
+    ) {
+        let peers_excl_self: Vec<_> = override_peers_inc_self
+            .iter()
+            .cloned()
+            .filter(|peer| peer != &self.our_nodeid)
+            .collect();
+        debug!( num_choices =? peers_excl_self.len(),
+                "Updated blocksync peer secondary_raptorcast_peers");
+        self.secondary_raptorcast_peers = peers_excl_self;
```
eerkaijun, line 240: say a full node is in 3 groups, each time a new confirm group is formed, the other two existing groups get wiped?

monad-blocksync/src/blocksync.rs
```diff
@@ -446,18 +482,21 @@ where
...
+            // Note that self nodeid is already excluded
+            debug!( num_choices =? override_peers.len(),
+                "Picking blocksync peer from override_peers");
+            *override_peers.choose(rng).expect("non empty")
+        } else if !secondary_raptorcast_peers.is_empty() {
+            // Second choice: a peer mentioned in last ConfirmGroup message
+            // Note that self nodeid is also here already excluded
+            debug!( num_choices =? secondary_raptorcast_peers.len(),
+                "Picking blocksync peer from secondary_raptorcast_peers");
+            *secondary_raptorcast_peers.choose(rng).expect("non empty")
```
eerkaijun: i think we need to take into account role change as well -- e.g. if a node is first a public full node, which then later gets promoted to be a validator, we never clear `secondary_raptorcast_peers`, so it will still try to blocksync from its old group although it should blocksync from other validators

monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -529,12 +529,20 @@ where
...
-                                peers: participated_nodes,
+                                peers: participated_nodes.clone(),
                             },
                         );
+                        return Poll::Ready(Some(
+                            RaptorCastEvent::DynamicOverridePeersUpdate(
+                                current_confirm_group_peers,
+                            )
+                            .into(),
+                        ));
```
eerkaijun: is this correct? if we return here, `client.on_receive_group_message` is never called

monad-blocksync/src/blocksync.rs
fabiotdf: It was needed for switching back and forth between using explicit override peers and dynamic, but now no longer needed, as we now have 2 vectors, one of which doesn't get used depending on config

monad-blocksync/src/blocksync.rs
fabiotdf: Implemented as suggested

monad-executor-glue/src/lib.rs
fabiotdf: I tried to but it was very messy to get monad-blocksync to access this, so I've put it in MonadEvent for now

monad-blocksync/src/blocksync.rs
fabiotdf: Yes, as it is now, only the latest group is used. This results in a skewed probability IRT what nodes are candidate for being picked up for blocksync, depending on the the current skew in group scheduling between the publishers.
Do we need a more elaborate mechanism to prune out old groups?

monad-raptorcast/src/raptorcast_secondary/mod.rs
fabiotdf: Good point, let me fix

monad-blocksync/src/blocksync.rs
fabiotdf: done

monad-blocksync/src/blocksync.rs
fabiotdf: I see. What is a simple way to react to a role switch from `MonadState` ? 
MonadEvent::ValidatorEvent UpdateValidators is nearby but is for a future event before the strate transition

monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -529,12 +531,21 @@ where
                             NodeId<CertificateSignaturePubKey<ST>>,
                         > = confirm_msg.peers.clone().into_iter().collect();
                         participated_nodes.insert(confirm_msg.prepare.validator_id);
+                        let current_confirm_group_peers =
+                            participated_nodes.iter().cloned().collect::<Vec<_>>();
```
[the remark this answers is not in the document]
fabiotdf: removed cloned() instead, as suggestion doesn't compile

monad-blocksync/src/blocksync.rs
eerkaijun: i made a commit here https://github.com/category-labs/monad-bft/pull/2423/commits/0aff2d6e000cc7d937490964afc5dbf762296a40

eerkaijun:
deployed on stressnet and testnet2

monad-blocksync/src/blocksync.rs
eerkaijun: i guess it's fine for now but seems a bit weird that current groups are not treated equally. we can address that later i guess

eerkaijun (approved):
can you squash the commits and remove the "Issue 1771 prefix" from the commit message?

monad-blocksync/src/blocksync.rs
michael-yxchen: i think we should address this now. counting number of group references should work for gc?

monad-executor-glue/src/lib.rs
```diff
@@ -1892,6 +1892,14 @@ where
     pub pinned_nodes: Vec<NodeId<CertificateSignaturePubKey<ST>>>,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, RlpEncodable, RlpDecodable, Serialize)]
+pub struct SecondaryRaptorcastEvent<SCT>
+where
+    SCT: SignatureCollection,
+{
+    pub confirm_group_peers: Vec<NodeId<SCT::NodeIdPubKey>>,
+}
```
michael-yxchen: we can add it to `BlockSyncEvent` enum. blocksync module consumes this

monad-blocksync/src/blocksync.rs
fabiotdf: I gave it a try here: https://github.com/category-labs/monad-bft/pull/2427

monad-executor-glue/src/lib.rs
fabiotdf: Done here: https://github.com/category-labs/monad-bft/pull/2436

PR #2407, opened 2025-09-30
Allow higher provider gas limit than block gas limit

philipglazman:
Operators can set a higher provider gas limit for eth_call and eth_estimateGas and meter usage themselves.

### DIFF
--- a/monad-ethcall/src/lib.rs
+++ b/monad-ethcall/src/lib.rs
@@ -20,7 +20,7 @@ use std::{
     sync::Arc,
 };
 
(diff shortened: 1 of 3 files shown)


eerkaijun (dismissed):

eerkaijun (approved):
if provider gas limit is larger than block gas limit, will execution respect the provider gas limit?

philipglazman:
> if provider gas limit is larger than block gas limit, will execution respect the provider gas limit?

Yes, the only guard is in monad_ethcall. I added a flexnet test to demonstrate this too.

PR #2409, opened 2025-10-01
archiver: fix block public access setting to true (aka 'set to private')

alexpotato:
Backstory:
- we have a script to create buckets
- that script was set to make new buckets publicly accessible (aka "block public access" was set to false"
- this now corrects the above

### DIFF
--- a/monad-archive/scripts/create-bucket.sh
+++ b/monad-archive/scripts/create-bucket.sh
@@ -28,7 +28,7 @@ echo "Disabling block public access settings..."
 # Disable block public access
 aws s3api put-public-access-block \
     --bucket "$S3_BUCKET_NAME" \
-    --public-access-block-configuration "BlockPublicAcls=false,IgnorePublicAcls=false,BlockPublicPolicy=false,RestrictPublicBuckets=false"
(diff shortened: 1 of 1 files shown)


monad-archive/scripts/create-bucket.sh
```diff
@@ -28,7 +28,7 @@ echo "Disabling block public access settings..."
 # Disable block public access
 aws s3api put-public-access-block \
     --bucket "$S3_BUCKET_NAME" \
-    --public-access-block-configuration "BlockPublicAcls=false,IgnorePublicAcls=false,BlockPublicPolicy=false,RestrictPublicBuckets=false"
```
JoeHowarth, line 31: This script is primarily for creating archive replica buckets, which generally should be public by default. Consider either add a cli flag for non-public

alexpotato: related question:

if we are making a public bucket, do we want to ensure that it's requestor pays?

I would assume "yes" which means:
- public -> requestor pays
- private (doesn't matter about requestor)

JoeHowarth: I think we want requester pays on all buckets actually, since if they're internal we pay pay either way

alexpotato: I'll update the script to have the switch and set the buckets to requestor pays

omegablitz (approved):

PR #2416, opened 2025-10-02
Remove unnecessary lock around request submission

maxkozlovsky:
There is no need to lock the executor to submit request to pool. The submission channel is MPMC queue which handles synchronization internally. There was a non-atomic counter of requests processed, this was fixed in execution.

### DIFF
--- a/monad-ethcall/src/lib.rs
+++ b/monad-ethcall/src/lib.rs
@@ -32,7 +32,6 @@ use monad_chain_config::{
     MONAD_TESTNET2_CHAIN_ID, MONAD_TESTNET_CHAIN_ID,
 };
 use serde::{Deserialize, Serialize};
-use tokio::sync::Mutex;
 use tracing::{info, warn};
 
 #[allow(dead_code, non_camel_case_types, non_upper_case_globals)]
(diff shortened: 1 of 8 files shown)


monad-triedb-utils/src/decode.rs
```diff
@@ -47,7 +47,7 @@ pub fn rlp_decode_account(account_rlp: Vec<u8>) -> Option<EthAccount> {
         return None;
     };
 
-    let mut is_delegated = false;
+    let is_delegated = false;
```
andr-dev, line 50: I wonder why this used to be mut, maybe we're missing something here?

andr-dev (approved):

monad-triedb-utils/src/decode.rs
mikety: Its fine. will be updated once execution change the representation

monad-ethcall/src/lib.rs
```diff
@@ -308,15 +307,11 @@ pub async fn eth_call(
...
     let sender_ctx = Box::new(SenderContext { sender: send });
 
-    // hold lock on executor while submitting the task
-    let executor_lock = eth_call_executor.lock().await;
-    let eth_call_executor = executor_lock.eth_call_executor;
-
     unsafe {
         let sender_ctx_ptr = Box::into_raw(sender_ctx);
 
         bindings::monad_eth_call_executor_submit(
```
omegablitz, line 313: nit: for documentation purposes, can you add a comment noting that `monad_eth_call_executor_submit` may be called concurrently by multiple threads, and is safe because it handles synchronization internally?

omegablitz (approved):

monad-ethcall/src/lib.rs
maxkozlovsky: Ok, I'll add the comments on the execution side to document the monad_eth_call api.

PR #2417, opened 2025-10-03
authentication protocol for udp

dshulyak:
implementation of the https://github.com/category-labs/monad-bft/pull/2091

### DIFF
--- a/monad-wireauth/src/api.rs
+++ b/monad-wireauth/src/api.rs
@@ -1,0 +1,716 @@
+// Copyright (C) 2025 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+use std::{
+    collections::{BTreeSet, VecDeque},
+    net::SocketAddr,
+    time::{Duration, Instant},
+};
+
+use bytes::Bytes;
+use monad_executor::{ExecutorMetrics, ExecutorMetricsChain};
+use monad_secp::PubKey;
+use tracing::{debug, error, instrument, trace, warn, Level};
+
+use crate::{
+    config::Config,
+    context::Context,
+    cookie::Cookies,
+    error::{Error, Result},
+    filter::{Filter, FilterAction},
+    messages::MacMessage,
+    metrics::*,
+    protocol::messages::{
+        ControlPacket, CookieReply, DataPacket, DataPacketHeader, HandshakeInitiation,
+        HandshakeResponse, Plaintext,
+    },
+    session::{InitiatorState, RenewedTimer, ResponderState, SessionIndex},
+    state::State,
+};
+
+pub struct API<C: Context> {
+    state: State,
+    timers: BTreeSet<(Duration, SessionIndex)>,
+    packet_queue: VecDeque<(SocketAddr, Bytes)>,
+    config: Config,
+    local_static_key: monad_secp::KeyPair,
+    // Cached compressed public key to avoid recomputing when logging
+    local_serialized_public: CompressedPublicKey,
+    cookies: Cookies,
+    filter: Filter,
+    context: C,
+    metrics: ExecutorMetrics,
+    last_tick: Option<Duration>,
+}
+
+impl<C: Context> API<C> {
+    /// Creates a new API instance, it should be created for an individual socket.
+    pub fn new(config: Config, local_static_key: monad_secp::KeyPair, mut context: C) -> Self {
+        let local_static_public = local_static_key.pubkey();
+        let cookies = Cookies::new(
+            context.rng(),
+            local_static_public,
+            config.cookie_refresh_duration,
+        );
+
+        let filter = Filter::new(
+            config.handshake_rate_limit,
+            config.handshake_rate_reset_interval,
+            config.ip_rate_limit_window,
+            config.ip_history_capacity,
+            config.max_sessions_per_ip,
+            config.low_watermark_sessions,
+            config.high_watermark_sessions,
+        );
+        let local_serialized_public = CompressedPublicKey::from(&local_static_public);
+        debug!(local_public_key=?local_serialized_public, "initialized manager");
+        Self {
+            state: State::new(),
+            timers: BTreeSet::new(),
+            packet_queue: VecDeque::new(),
+            config,
+            local_static_key,
+            local_serialized_public,
+            cookies,
+            filter,
+            context,
+            metrics: ExecutorMetrics::default(),
+            last_tick: None,
+        }
+    }
+
+    pub fn metrics(&self) -> ExecutorMetricsChain {
+        ExecutorMetricsChain::default()
+            .push(&self.metrics)
+            .push(self.state.metrics())
+            .push(self.filter.metrics())
+    }
+
+    /// Returns the next packet to send over the network.
+    ///
+    /// Note: There are no limits for the internal queue, so it is better to use a separate
+    /// queue for pacing.
+    #[instrument(level = Level::TRACE, skip(self), fields(local_public_key = ?self.local_serialized_public))]
+    pub fn next_packet(&mut self) -> Option<(SocketAddr, Bytes)> {
+        self.metrics[GAUGE_WIREAUTH_API_NEXT_PACKET] += 1;
+        let result = self.packet_queue.pop_front();
+        self.metrics[GAUGE_WIREAUTH_STATE_PACKET_QUEUE_SIZE] = self.packet_queue.len() as u64;
+        result
+    }
+
+    fn enqueue_packet(&mut self, addr: SocketAddr, pkt: impl Into<Bytes>) {
+        self.packet_queue.push_back((addr, pkt.into()));
+        self.metrics[GAUGE_WIREAUTH_STATE_PACKET_QUEUE_SIZE] = self.packet_queue.len() as u64;
+    }
+
+    /// Returns the next deadline.
+    #[instrument(level = Level::TRACE, skip(self), fields(local_public_key = ?self.local_serialized_public))]
+    pub fn next_deadline(&self) -> Option<Instant> {
+        let session_deadline = self.timers.iter().next().map(|&(deadline, _)| deadline);
+
+        let filter_deadline = self.filter.next_reset_time();
+
+        let deadline = match session_deadline {
+            Some(sd) => sd.min(filter_deadline),
+            None => filter_deadline,
+        };
+
+        Some(
+            self.context
+                .convert_duration_since_start_to_deadline(deadline),
+        )
+    }
+
+    fn insert_timer(&mut self, timer: Duration, session_id: SessionIndex) {
+        self.timers.insert((timer, session_id));
+        self.metrics[GAUGE_WIREAUTH_STATE_TIMERS_SIZE] = self.timers.len() as u64;
+    }
+
+    fn replace_timer(&mut self, timer: RenewedTimer, session_index: SessionIndex) {
+        if let Some(previous) = timer.previous {
+            self.timers.remove(&(previous, session_index));
+        }
+        self.timers.insert((timer.current, session_index));
+        self.metrics[GAUGE_WIREAUTH_STATE_TIMERS_SIZE] = self.timers.len() as u64;
+    }
+
+    #[instrument(level = Level::TRACE, skip(self), fields(local_public_key = ?self.local_serialized_public))]
+    pub fn tick(&mut self) {
+        self.metrics[GAUGE_WIREAUTH_API_TICK] += 1;
+        let duration_since_start = self.context.duration_since_start();
+
+        self.filter.tick(duration_since_start);
+
+        let expired_timers: Vec<(Duration, SessionIndex)> = self
+            .timers
+            .range(..=(duration_since_start, SessionIndex::MAX))
+            .copied()
+            .collect();
+
+        if let Some(last_tick) = self.last_tick {
+            let checked_duration = duration_since_start.saturating_sub(last_tick);
+            trace!(
+                checked_duration_ms = checked_duration.as_millis(),
+                expired_timers = expired_timers.len(),
+                "tick"
+            );
+        } else {
+            trace!(expired_timers = expired_timers.len(), "tick");
+        }
+
+        for (duration, session_id) in expired_timers {
+            self.timers.remove(&(duration, session_id));
+            self.metrics[GAUGE_WIREAUTH_STATE_TIMERS_SIZE] = self.timers.len() as u64;
+
+            if let Some(elapsed) = duration_since_start.checked_sub(duration) {
+                let elapsed_ms = elapsed.as_millis();
+                trace!(
+                    session_id=?session_id,
+                    elapsed_ms=elapsed_ms,
+                    "timer triggered"
+                );
+                if elapsed_ms > 100 {
+                    warn!(
+                        session_id=?session_id,
+                        elapsed_ms=elapsed_ms,
+                        "deadline is too old"
+                    );
+                }
+            } else {
+                error!(
+                    session_id=?session_id,
+                    deadline_duration=?duration,
+                    duration_since_start=?duration_since_start,
+                    "deadline is in the future"
+                );
+            }
+
+            let tick_result = if let Some(s) = self.state.get_initiator_mut(&session_id) {
+                s.tick(duration_since_start)
+                    .map(|(timer, r)| (timer, None, r.rekey, Some(r.terminated)))
+            } else if let Some(s) = self.state.get_responder_mut(&session_id) {
+                s.tick(duration_since_start)
+                    .map(|(timer, r)| (timer, None, r.rekey, Some(r.terminated)))
+            } else if let Some(transport) = self.state.get_transport_mut(&session_id) {
+                Some(transport.tick(self.context.rng(), &self.config, duration_since_start))
+            } else {
+                None
+            };
+
+            let Some((timer, message, rekey, terminated)) = tick_result else {
+                continue;
+            };
+
+            if let Some(message) = message {
+                self.metrics[GAUGE_WIREAUTH_ENQUEUED_KEEPALIVE] += 1;
+                self.enqueue_packet(message.remote_addr, message.header);
+            }
+
+            if let Some(rekey) = rekey {
+                if let Ok((new_session_index, timer, message)) = self.init_session_with_cookie(
+                    rekey.remote_public_key,
+                    rekey.remote_addr,
+                    rekey.stored_cookie,
+                    rekey.retry_attempts,
+                ) {
+                    self.metrics[GAUGE_WIREAUTH_ENQUEUED_HANDSHAKE_INIT] += 1;
+                    self.enqueue_packet(rekey.remote_addr, message);
+                    self.insert_timer(timer, new_session_index);
+                }
+            }
+
+            if let Some(timer) = timer {
+                self.insert_timer(timer, session_id);
+            }
+
+            if let Some(terminated) = terminated {
+                self.state.terminate_session(
+                    session_id,
+                    &terminated.remote_public_key,
+                    terminated.remote_addr,
+                );
+            }
+        }
+
+        self.last_tick = Some(duration_since_start);
+    }
+
+    /// Initiates a connection with a peer.
+    ///
+    /// This will initiate a connection even if the peer is already connected.
+    /// To avoid this, the caller can check for existing sessions before calling this method.
+    #[instrument(level = Level::TRACE, skip(self, remote_static_key), fields(local_public_key = ?self.local_serialized_public, remote_addr = ?remote_addr))]
+    pub fn connect(
+        &mut self,
+        remote_static_key: monad_secp::PubKey,
+        remote_addr: SocketAddr,
+        retry_attempts: u64,
+    ) -> Result<()> {
+        self.metrics[GAUGE_WIREAUTH_API_CONNECT] += 1;
+        debug!(retry_attempts, "initiating connection");
+
+        // Cookies are looked up from initiated sessions for simplicity.
+        // In the future, this can be improved to look up from both initiated and accepted sessions.
+        let cookie = self
+            .state
+            .lookup_cookie_from_initiated_sessions(&remote_static_key);
+
+        let (local_index, timer, message) = self
+            .init_session_with_cookie(remote_static_key, remote_addr, cookie, retry_attempts)
+            .inspect_err(|_| {
+                self.metrics[GAUGE_WIREAUTH_ERROR_CONNECT] += 1;
+            })?;
+
+        self.metrics[GAUGE_WIREAUTH_ENQUEUED_HANDSHAKE_INIT] += 1;
+        self.enqueue_packet(remote_addr, message);
+        self.insert_timer(timer, local_index);
+
+        Ok(())
+    }
+
+    fn init_session_with_cookie(
+        &mut self,
+        remote_static_key: monad_secp::PubKey,
+        remote_addr: SocketAddr,
+        cookie: Option<[u8; 16]>,
+        retry_attempts: u64,
+    ) -> Result<(SessionIndex, Duration, HandshakeInitiation)> {
+        trace!(%remote_addr, ?remote_static_key, cookie = cookie.is_some(), ?retry_attempts, "init session");
+
+        // reservation should be committed when code is no longer fallible
+        let reservation = self.state.reserve_session_index().ok_or_else(|| {
+            self.metrics[GAUGE_WIREAUTH_ERROR_SESSION_EXHAUSTED] += 1;
+            Error::SessionIndexExhausted
+        })?;
+        trace!(local_session_id=?reservation.index(), "allocating session index for new connection");
+        let system_time = self.context.system_time();
+        let duration_since_start = self.context.duration_since_start();
+        let (session, (timer, message)) = InitiatorState::new(
+            self.context.rng(),
+            system_time,
+            duration_since_start,
+            &self.config,
+            reservation.index(),
+            &self.local_static_key,
+            remote_static_key,
+            remote_addr,
+            cookie,
+            retry_attempts,
+        );
+        let index = reservation.index();
+        reservation.commit();
+
+        self.state
+            .insert_initiator(index, session, remote_static_key);
+
+        Ok((index, timer, message))
+    }
+
+    fn is_under_load(
+        &mut self,
+        remote_addr: SocketAddr,
+        sender_index: u32,
+        message: &impl MacMessage,
+    ) -> bool {
+        let duration_since_start = self.context.duration_since_start();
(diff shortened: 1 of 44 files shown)


dshulyak:
this doesn't include integration code (e.g discovery, dataplane and raptorcast changes) and i still need to add more docs to the code itself

monad-wireauth/src/protocol/cookies.rs
```diff
@@ -0,0 +1,379 @@
...
+        msg_mac1,
+    )
+    .map_err(CookieError::CookieDecryptionFailed)?;
+
+    Ok(reply.encrypted_cookie)
+}
+
+pub fn generate_cookie(cookie_secret: &[u8; 32], nonce: u64, remote_addr: &SocketAddr) -> [u8; 16] {
+    let mut address_bytes = [0u8; 16];
+    match remote_addr.ip() {
```
dnkolegov-ar: The remote port is not used. According to the spec: `The initiator address is a concatenation of the message source IP address and source port.`

monad-wireauth-protocol/src/common.rs
```diff
@@ -0,0 +1,337 @@
...
+    }
+}
+
+impl AsRef<[u8; 16]> for CipherNonce {
+    fn as_ref(&self) -> &[u8; 16] {
+        &self.0
+    }
+}
+
+#[derive(Clone, Debug, Zeroize)]
```
dnkolegov-ar: There is no `ZeroizeOnDrop`. Is it intentional? It could be useful for `tmp_x` variables

monad-wireauth/src/protocol/handshake.rs
```diff
@@ -0,0 +1,849 @@
...
+    state.chaining_key = keyed_hash!(temp.as_ref(), &[0x1]).into();
+
+    let ecdh_ee = ecdh(
+        state
+            .ephemeral_private
+            .as_ref()
+            .expect("ephemeral private key must be set"),
+        &msg.ephemeral_public,
+    )?;
+    let temp = keyed_hash!(state.chaining_key.as_ref(), ecdh_ee.as_ref());
```
dnkolegov-ar, line 311: The ephemeral private key of the initiator's state can be erased here.

monad-wireauth-protocol/src/common.rs
dshulyak: yes, i am using Zeroizing<HashOutput> for chaining_key and hash output. i think there was no need to zeroize every tmp variable, will recheck

monad-wireauth/src/protocol/handshake.rs
dshulyak: i guess i can `.take()` it from the state, but the state itself is dropped (and private key is erased at that point) soon after exiting this function

monad-wireauth/src/protocol/cookies.rs
dshulyak: thanks, i forgot about port, will add

monad-wireauth-api/src/api.rs
```diff
@@ -0,0 +1,594 @@
...
+                        Err(_) => None,
+                    }
+                })
+            } else if self.state.get_responder(&session_id).is_some() {
+                self.state.get_responder_mut(&session_id).and_then(|s| {
+                    match s.tick(duration_since_start) {
+                        Ok(result) => {
+                            result.map(|(timer, r)| (timer, None, r.rekey, Some(r.terminated)))
+                        }
+                        Err(_) => None,
```
dnkolegov-ar: Why is it safe to ignore an error here and at similar places below?

dshulyak: tick() shouldn't be fallible, i will remove Result from every such function in a follow up

monad-wireauth-session/src/common.rs
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    pub fn handle_cookie(
+        &mut self,
+        cookie_reply: &mut monad_wireauth_protocol::messages::CookieReply,
+    ) -> Result<(), SessionError> {
+        let Some(mac1) = self.last_handshake_mac1 else {
+            debug!("no last_handshake_mac1 stored");
+            return Err(SessionError::InvalidCookie(
+                monad_wireauth_protocol::errors::CookieError::InvalidCookieMac(
```
dnkolegov-ar: What is the relation between mac1 and cookie? Mac1 shouldn't be related to cookie, right?

monad-wireauth/src/protocol/cookies.rs
```diff
@@ -0,0 +1,379 @@
...
+    decrypt_in_place(
+        &(&temp_key).into(),
+        &reply.nonce,
+        &mut reply.encrypted_cookie,
+        &reply.encrypted_cookie_tag,
+        msg_mac1,
+    )
+    .map_err(CookieError::CookieDecryptionFailed)?;
+
+    Ok(reply.encrypted_cookie)
```
dnkolegov-ar, line 76: Maybe it is worth adding a comment that this function decrypts the cookie and returns its value. No comments, and a returned `reply.encrypted_cookie` is misleading.

monad-wireauth-session/src/common.rs
```diff
@@ -0,0 +1,304 @@
...
+            remote_public_key: self.remote_public_key.clone(),
+            remote_addr: self.remote_addr,
+            retry_attempts: self.retry_attempts,
+            stored_cookie: self.stored_cookie,
+        });
+
+        Ok((terminated, rekey))
+    }
+
+    pub fn handle_rekey_timer(&mut self) -> RekeyEvent {
```
dnkolegov-ar: It is called `handle_rekey_timer`, but nothing is handled here.

monad-wireauth/src/session/initiator.rs
```diff
@@ -0,0 +1,168 @@
...
+        .map_err(SessionError::InvalidHandshake)?;
+
+        let mac1 = init_msg.mac1.0;
+        let mut common = CommonSessionData::new(
+            remote_addr,
+            remote_static_key,
+            local_session_index,
+            duration_since_start,
+            retry_attempts,
+            None,
```
dnkolegov-ar, line 75: Why is system time set to None?

monad-wireauth/src/session/initiator.rs
```diff
@@ -0,0 +1,168 @@
...
+            true,
+        );
+        common.stored_cookie = cookie_secret;
+        common.last_handshake_mac1 = Some(mac1);
+
+        let mut session = Initiator {
+            handshake_state,
+            common,
```
dnkolegov-ar, line 83: Maybe it should be called session state?

monad-wireauth/src/cookie.rs
```diff
@@ -0,0 +1,285 @@
...
+use std::net::SocketAddr;
+use std::time::Duration;
+
+use monad_wireauth_protocol::common::SerializedPublicKey;
+use monad_wireauth_protocol::errors::ProtocolError;
+use monad_wireauth_protocol::messages::CookieReply;
+
+use crate::error::{ProtocolErrorContext, Result};
+
+pub struct Cookies {
```
dnkolegov-ar, line 20: There is no zeroization for secrets

monad-wireauth/src/cookie.rs
```diff
@@ -0,0 +1,285 @@
...
+}
+
+impl Cookies {
+    pub fn new<R: RngCore + CryptoRng>(
+        rng: &mut R,
+        local_static_public: SerializedPublicKey,
+        refresh_duration: Duration,
+    ) -> Self {
+        let mut cookie_secret = [0u8; 32];
+        rng.fill_bytes(&mut cookie_secret);
```
dnkolegov-ar, line 35: What happens if the responder is under load, sends a handshake response with a cookie, and then immediately crashes? The initiator is an honest node, and it will send the new handshake initiation with the cookie that can't be verified after reloading. Will the reloaded node accept that handshake after timeout?

P.S. It is easier to ask that explicitly than to go through the code

monad-wireauth-session/src/common.rs
```diff
@@ -0,0 +1,304 @@
...
+    pub handshake_rate_limit: u64,
+    pub handshake_rate_reset_interval: Duration,
+    pub cookie_refresh_duration: Duration,
+    pub low_watermark_sessions: usize,
+    pub high_watermark_sessions: usize,
+    pub max_sessions_per_ip: usize,
+    pub ip_rate_limit_window: Duration,
+    pub max_requests_per_ip: usize,
+    pub ip_history_capacity: usize,
+    pub psk: [u8; 32],
```
dnkolegov-ar: Do we need to add zeroization for Config?

monad-wireauth/src/session/common.rs
```diff
@@ -0,0 +1,304 @@
...
+    #[error("session timed out: exceeded maximum duration or idle time")]
+    SessionTimeout,
+}
+
+pub struct CommonSessionData {
+    pub keepalive_deadline: Option<Duration>,
+    pub rekey_deadline: Option<Duration>,
+    pub session_timeout_deadline: Option<Duration>,
+    pub max_session_duration_deadline: Option<Duration>,
+    pub stored_cookie: Option<[u8; 16]>,
```
dnkolegov-ar, line 81: Do we need to add zeroization for the cookie secret?

monad-wireauth-session/src/common.rs
dshulyak: from wireguard description

```
msg.encrypted_cookie = XAEAD(HASH(LABEL_COOKIE || responder.static_public), msg.nonce, cookie, last_received_msg.mac1)
```

as i understand it is used to bind cookie reply to the specific message

monad-wireauth/src/cookie.rs
dshulyak: i don't think that it accomplishes anything. those secrets can be erased only when authentication protocol is terminated (e.g program closed for example).
and it doesn't matter if they are stolen when protocol is no longer working

monad-wireauth/src/cookie.rs
dshulyak: if responder is under load after crash then such message will have invalid mac2 and will be rejected.
if responder is not under load then invlalid mac2 will be ignored and initiation will pass.

do you think it is worth it to persist cookie secret to handle this case? to me it seems not worth it

monad-wireauth/src/protocol/cookies.rs
dshulyak: will add

monad-wireauth-session/src/common.rs
dshulyak: i think it makes sense to erase psk, as it meant to be long term secret

monad-wireauth/src/session/common.rs
dshulyak: i think it only allows attacker to bypass dos-filter. 
it doesn't cost much to wrap this in Zerozing, but practically it seemed to me unnecessary

monad-wireauth-session/src/common.rs
dshulyak: thanks, i will refactor this

monad-wireauth/src/session/initiator.rs
dshulyak: it is used only on responder side to prevent handshake replay, so it is set to None for initiator

monad-wireauth/src/session/initiator.rs
dshulyak: do you mean whole struct, e.g rename CommonSessionData to SessionState and all related variables?
i can rename, seems like a better choice

monad-wireauth-session/src/common.rs
dnkolegov-ar: Right, I forgot about that one; it is named `M` in the paper.

monad-wireauth/src/session/initiator.rs
dnkolegov-ar: Correct:
`common -> session_state`
`CommonSessionData -> SessionState`

The Noise [protocol](http://www.noiseprotocol.org/noise.html) framework uses `CipherState`, `SymmetricState`, `HandshakeState`,  you can follow that naming for other relevant types: `SessionState`, `ProtocolState`, ...

monad-wireauth/src/cookie.rs
dnkolegov-ar: So, if the responder is still under load after its crash, the honest initiator will not be able to establish the connection until some timer is triggered. In WireGuard, it seems to me, this would be `Reject-After-Time × 3` seconds.  
I am trying to understand which timeout defines that, and in how many seconds the honest peer will be able to establish a connection

monad-wireauth/src/api.rs
```diff
@@ -0,0 +1,594 @@
...
+
+use crate::context::Context;
+use crate::cookie::Cookies;
+use crate::error::{Error, ProtocolErrorContext, Result, SessionErrorContext};
+use crate::filter::{Filter, FilterAction};
+use crate::state::State;
+use crate::{Initiator, Responder, Transport};
+use monad_wireauth_session::{Config, SessionIndex};
+
+pub struct API<C: Context> {
```
dnkolegov-ar, line 43: In WireGuard:
```
after an initiator receives a handshake response message (section 5.4.3), 
the responder cannot send transport data messages (section 5.4.6)
until it has received the first transport data message from the initiator.
And, further, transport data messages encrypted using the previous secure session might be in transit after a new secure session has been created.
For these reasons, WireGuard keeps in memory the current secure session,
the previous secure session, and the next secure session for the case of an unconfirmed session 
```

Could you point me to the code where this logic is implemented? Or explain why this is not implemented and why the protocol does not need that functionality.

I think it should be in https://github.com/category-labs/monad-bft/blob/98a6b9b070f7f6931652d8139ab59d0b23669bf5/monad-wireauth-api/src/api.rs#L400, but I do not see the entire workflow, only the mechanism that ensures that the initiator must send some data to the responder before the responder starts sending data to the initiator.

monad-wireauth-api/src/state.rs
```diff
@@ -0,0 +1,1003 @@
...
+    accepted_sessions_by_peer: BTreeSet<(SerializedPublicKey, SessionIndex)>,
+}
+
+impl State {
+    pub fn new() -> Self {
+        Self {
+            initiating_sessions: HashMap::new(),
+            responding_sessions: HashMap::new(),
+            transport_sessions: HashMap::new(),
+            last_open_session_by_public_key: HashMap::new(),
```
dnkolegov-ar: What does `open` mean? Who opened it? Can we use `initiator`/`responder`-based word?
Does `Open` mean `not yet established`?

monad-wireauth/src/cookie.rs
dshulyak: initiator will retry if it doesn't get reply in `session_timeout` with some jitter

monad-wireauth-api/src/state.rs
```diff
@@ -0,0 +1,1003 @@
...
+            if !terminated_sessions.contains(&existing_id) {
+                terminated_sessions.push(existing_id);
+            }
+        }
+
+        arr[slot_index] = Some((session_id, created));
+
+        self.transport_sessions.insert(session_id, transport);
+
+        terminated_sessions
```
dnkolegov-ar: It is not obvious without comments why the function called `handle_established ` returns `terminated_sessions` also in https://github.com/category-labs/monad-bft/blob/98a6b9b070f7f6931652d8139ab59d0b23669bf5/monad-wireauth-api/src/api.rs#L204

monad-wireauth/src/api.rs
dshulyak:
```
after an initiator receives a handshake response message (section 5.4.3), 
the responder cannot send transport data messages (section 5.4.6)
until it has received the first transport data message from the initiator.
```

implemented by having a separate type for Responder, that type can't send messages, but can accept them (decrypt method), and on first successfully decrypted message it transitions to Transport type (establish method)

```
And, further, transport data messages encrypted using the previous secure session might be in transit after a new secure session has been created.
For these reasons, WireGuard keeps in memory the current secure session,
the previous secure session, and the next secure session for the case of an unconfirmed session 
```

in api crate (state.rs) i track sessions with different state separately, sessions in Transport state can't be replaced by unconfirmed sessions, once Transport is created there is a call handle_established that evicts older sessions. that covers current secure session (Transport type) and separate unconfirmed session. so initiator or responder will never evict established session until new session transitions to Transport state

there is currently an expectation that both sides will initiate `connect` before sending messages, and i track 1 of each initiated and accepted instead of previous secure session. it addresses a problem that older session may get evicted during rekey while there are inflight messages

dshulyak: though this last part is not very intuitive, i will check how to transition to simply storing 2 last established sessions instead

monad-wireauth-api/src/state.rs
dshulyak: should be `established` or `transport` last_transport_session_by_public_key

monad-wireauth-api/src/state.rs
dshulyak: those are replaced sessions, i will take a look if it can be simplified

monad-wireauth-api/src/api.rs
```diff
@@ -0,0 +1,594 @@
...
+        if !self.check_under_load(remote_addr, response.sender_index.get(), response)? {
+            debug!(?remote_addr, "handshake response dropped under load");
+            return Ok(());
+        }
+
+        let receiver_session_index = response.receiver_index.into();
+
+        let mut initiator = self
+            .state
+            .remove_initiator(&response.receiver_index.get().into())
```
dnkolegov-ar: Why is the State removed *before* the response is authenticated/validated?

monad-wireauth/src/api.rs
```diff
@@ -0,0 +1,594 @@
...
+
+        Ok(is_keepalive)
+    }
+
+    fn complete_handshake(
+        &mut self,
+        response: &mut HandshakeResponse,
+        remote_addr: SocketAddr,
+    ) -> Result<()> {
+        if !self.check_under_load(remote_addr, response.sender_index.get(), response)? {
```
dnkolegov-ar: In WireGuard, verify MAC1 first — always. Check the MAC2 if the host is under load.
Here, first check is whether we are under load, and then check MAC1

monad-wireauth-api/src/api.rs
```diff
@@ -0,0 +1,594 @@
...
+            remote_addr,
+            handshake_packet.sender_index.get(),
+            handshake_packet,
+        )? {
+            debug!(?remote_addr, "handshake initiation dropped under load");
+            return Ok(());
+        }
+
+        let duration_since_start = self.context.duration_since_start();
+        let local_index = self.state.allocate_session_index();
```
dnkolegov-ar: We allocate and insert a new index here. What if we allocate it and then fail on `validate_init`? My understanding is that the error is returned, but `handle_terminate` is not called, leaving the index permanently marked as “allocated.”

monad-wireauth/src/protocol/crypto.rs
```diff
@@ -0,0 +1,124 @@
...
+        use $crate::common::HashOutput;
+        let mut hasher = blake3::Hasher::new_keyed($key);
+        hasher.update($data1);
+        hasher.update($data2);
+        hasher.update($data3);
+        HashOutput(hasher.finalize().into())
+    }};
+}
+
+pub fn verify_keyed_hash(key: &HashOutput, data: &[u8], tag: &[u8; 16]) -> Result<(), CryptoError> {
```
dnkolegov-ar, line 92: Just to be sure, you intentionally chose to make `verify_keyed_hash`, `keyed_hash` and `MacTag` non-constant-time

dshulyak: i used constant time implementation at first, 
but i wasn't sure if the rest of the implementation is constant time and i didn't want to pull another lib for that.

i decided that for monad network purposes non constant time is good enough

monad-wireauth-api/src/api.rs
dshulyak: thats a mistake

monad-wireauth/src/api.rs
dshulyak: i will update to run mac1 check before this, the same way it is done in accept_handshake_init

monad-wireauth-api/src/api.rs
dshulyak: thanks, i forgot to update this method to use get -> validate -> remove , similarly to decrypt_data_packet

monad-wireauth/src/protocol/crypto.rs
dnkolegov-ar: I agree, just wanted to check it was intentional.

dnkolegov-ar:
@dshulyak When can I review the fixes for the findings? I think it makes sense to do one more round of the security review after you fix all current findings/comments.

dshulyak:
@dnkolegov-ar i will push changes by wed 29th

monad-wireauth-protocol/src/common.rs
dshulyak: my reasoning was that those values are always on the stack. so they will be overwritten shortly after function returns.
adding ZeroizeOnDrop overwrites them only a few ns earlier

it likely costs nothing to zeroize them, just feels weird.
boringtun doesn't zeroize intermidiate variables, i am curious how kernel implementation handles those.

monad-wireauth/src/protocol/handshake.rs
dshulyak: i kept it as is, dropping it slightly earlier should not make a difference

dshulyak:
@dnkolegov-ar i pushed a separate commit with changes for comments

monad-wireauth/src/protocol/cookies.rs
dnkolegov-ar: fixed

monad-wireauth-api/src/api.rs
dnkolegov-ar: fixed

monad-wireauth/src/session/initiator.rs
dnkolegov-ar: Fixed

monad-wireauth-api/src/api.rs
dnkolegov-ar: fixed

monad-wireauth/src/api.rs
dnkolegov-ar: fixed

monad-wireauth-api/src/api.rs
dnkolegov-ar: fixed

monad-wireauth-api/src/state.rs
dnkolegov-ar: fixed

monad-wireauth/src/api.rs
dnkolegov-ar: Fixed

monad-wireauth/src/protocol/handshake.rs
```diff
@@ -0,0 +1,831 @@
...
+
+    (msg, inititiator)
+}
+
+pub fn accept_handshake_init(
+    responder_static_keypair: &monad_secp::KeyPair,
+    msg: &mut HandshakeInitiation,
+) -> Result<(HandshakeState, SystemTime), ProtocolError> {
+    let responder_static_public = responder_static_keypair.pubkey();
+    crate::crypto::verify_mac1(msg, &responder_static_public)
```
mawalu: For both accepting and processing handshakes MAC1 seems to be verified twice per message? Once in `API` and once here

monad-wireauth/src/filter.rs
```diff
@@ -0,0 +1,462 @@
...
+
+        if total_sessions < self.low_watermark_sessions {
+            return FilterAction::Pass;
+        }
+
+        if !cookie_valid {
+            return FilterAction::SendCookie;
+        }
+
+        let ip = remote_addr.ip();
```
mawalu: Since this potentionally handles IPv6 it could be wise to support ratelimiting based on subnets. For v6 devices comonly get a `/64` net and can easily rotate IPs.

monad-wireauth/src/protocol/messages.rs
```diff
@@ -0,0 +1,661 @@
...
+    }
+}
+
+#[repr(C, packed)]
+#[derive(FromBytes, IntoBytes, Immutable, KnownLayout, Clone)]
+pub struct DataPacketHeader {
+    pub message_type: u8,
+    pub reserved: [u8; 3],
+    pub receiver_index: U32<LE>,
+    pub counter: U64<LE>,
```
mawalu: This is called `nonce` in the spec and for all other packet types. Could be unified for clarity.

monad-wireauth/src/filter.rs
dshulyak: i will add a note that it is not properly supported despite the types. there are lots of assumptions in monad codebase that prevent usage of ipv6 which is not trivial to get rid of, so i am not concerned about it being misused

monad-wireauth/src/protocol/handshake.rs
dshulyak: i didn't remove as there is almost no overhead of having that, but yes probably doesn't add any real misuse safety

monad-wireauth/src/api.rs
```diff
@@ -0,0 +1,723 @@
...
+                self.metrics[GAUGE_WIREAUTH_ERROR_MAC1_VERIFICATION_FAILED] += 1;
+                Error::Mac1VerificationFailed {
+                    addr: remote_addr,
+                    source,
+                }
+            },
+        )?;
+
+        if !self.is_under_load(remote_addr, response.sender_index.get(), response) {
+            debug!(?remote_addr, "handshake response dropped under load");
```
ariqchowdhury, line 557: from the spec, is the responder supposed to respond with a cookie_reply and initiator does handshake_init again with mac2 which should be verified? 
checking if that is intentionally omitted from this version or if I missed something

dshulyak: this is done inside of this method https://github.com/category-labs/monad-bft/blob/9308d534788192cda464a7ea8c307427f213c975/monad-wireauth/src/api.rs#L311-L343

omegablitz (approved):

monad-wireauth/src/api.rs
```diff
@@ -0,0 +1,716 @@
...
+        &mut self,
+        handshake_packet: &mut HandshakeInitiation,
+        remote_addr: SocketAddr,
+    ) -> Result<()> {
+        crate::protocol::crypto::verify_mac1(handshake_packet, &self.local_static_key.pubkey())
+            .inspect_err(|_| {
+                self.metrics[GAUGE_WIREAUTH_ERROR_MAC1_VERIFICATION_FAILED] += 1;
+            })?;
+
+        if !self.is_under_load(
```
dnkolegov-ar, line 368: The name `is_under_load` is misleading (it means “filter_passed” or "not_under_load").

Current behaviour:

FilterAction::Pass → `is_under_load` returns true → you do not return early → you process the handshake.

FilterAction::SendCookie / Drop → `is_under_load` returns false → you return early → you do not process the handshake.

monad-wireauth/src/filter.rs
```diff
@@ -0,0 +1,538 @@
...
+        remote_addr: SocketAddr,
+        duration_since_start: Duration,
+        cookie_valid: bool,
+    ) -> FilterAction {
+        info!(remote_addr = %remote_addr, cookie_valid = cookie_valid, "applying filter");
+        self.counter += 1;
+        let total_sessions = state.total_sessions();
+        let ip = remote_addr.ip();
+
+        let action = self
```
dnkolegov-ar, line 115: The spec says "During rekeying, the responder must not apply the global rate limit to handshake packets from peers with existing sessions". I do not see how this is implemented here

PR #2418, opened 2025-10-03
Clean up logs and current number of groups metric

eerkaijun:
`CLIENT_NUM_CURRENT_GROUPS` is not showing up correctly because an interval in `confirmed_groups` is removed immediately when entering round.

### DIFF
--- a/monad-raptorcast/src/raptorcast_secondary/client.rs
+++ b/monad-raptorcast/src/raptorcast_secondary/client.rs
@@ -130,11 +130,13 @@ where
 
         // Remove all groups that should already have been consumed
         let mut keys_to_remove = Vec::new();
-        for iv in self.confirmed_groups.intervals(..self.curr_round) {
-            keys_to_remove.push(iv);
+        for interval in self.confirmed_groups.intervals(..self.curr_round) {
+            if interval.end <= self.curr_round {
+                keys_to_remove.push(interval.clone());
+            }
         }
-        for iv in keys_to_remove {
-            self.confirmed_groups.remove(iv);
+        for interval_key in keys_to_remove {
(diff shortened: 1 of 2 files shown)


monad-raptorcast/src/udp.rs
```diff
@@ -198,12 +198,6 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
                     &parsed_message.author,
                     broadcast_mode,
                 ) {
-                    tracing::debug!(
```
eerkaijun, line 201: removing this log because it's already logged once inside `check_source`

monad-raptorcast/src/raptorcast_secondary/client.rs
```diff
@@ -130,11 +130,13 @@ where
 
         // Remove all groups that should already have been consumed
         let mut keys_to_remove = Vec::new();
-        for iv in self.confirmed_groups.intervals(..self.curr_round) {
-            keys_to_remove.push(iv);
+        for interval in self.confirmed_groups.intervals(..self.curr_round) {
+            if interval.end <= self.curr_round {
```
fabiotdf, line 134: This should always be true because of the `(..self.curr_round)` range specified in the for loop, no?

fabiotdf (approved):

monad-raptorcast/src/raptorcast_secondary/client.rs
eerkaijun: `intervals(..self.curr_round)` returns all the intervals that overlap with `self.curr_round`, e.g if a group is round 1-5 and current round is round 2. but it should only be pruned when current round is round 5. 

if this line is commented, the unit test i added will fail

PR #2427, opened 2025-10-06
Fabio/fn blocksync peers.prune old

fabiotdf:


monad-blocksync/src/blocksync.rs
```diff
@@ -494,8 +508,13 @@ where
         let self_is_validator = validators.keys().contains(self_node_id);
 
         if !secondary_raptorcast_peers.is_empty() && !self_is_validator {
-            // secondary peers is set and self is a full node
-            *secondary_raptorcast_peers.choose(rng).expect("non empty")
+            // Choose a random peer from secondary_raptorcast_peers
+            let candidate_peers: Vec<_> = secondary_raptorcast_peers.values().collect();
```
eerkaijun: do we need to deduplicate here?

monad-blocksync/src/blocksync.rs
```diff
@@ -232,12 +233,25 @@ where
...
+            .retain(|expiry_round, _| *expiry_round > round_span.start);
+
+        // Trim peers that are already included in peers_excl_self, as we will
+        // be re-adding them with new expiry keys.
+        self.secondary_raptorcast_peers
+            .retain(|_, peer| !peers_excl_self.contains(peer));
+
+        for peer in peers_excl_self {
+            self.secondary_raptorcast_peers.insert(round_span.end, peer);
+        }
```
eerkaijun, line 254: can this be unit tested?

monad-blocksync/src/blocksync.rs
fabiotdf: Well, the whole point with this PR was to provide a more even probability distribution for peer selection. Without this dedup, some peers will have 2x or 3x the probability of some other, similar to the skew without these changes

eerkaijun: yeah i meant we should deduplicate, but is this the case since we are collecting into a vec and not a set?

monad-blocksync/src/blocksync.rs
```diff
@@ -167,7 +167,8 @@ where
     override_peers: Vec<NodeId<CertificateSignaturePubKey<ST>>>,
 
     /// Excludes self node id
-    secondary_raptorcast_peers: Vec<NodeId<CertificateSignaturePubKey<ST>>>,
+    /// Expiry round -> NodeId
+    secondary_raptorcast_peers: BTreeMap<Round, NodeId<CertificateSignaturePubKey<ST>>>,
```
michael-yxchen: i dont believe the choice of data structure is correct. value type should be vec/set of NodeId

fabiotdf: Yeah I was thinking the same. I changed it to be a map of NodeId -> Round instead. This makes most sense when updating or pushing back a peer's expiry round upon receipt of new batch of peers from secondary raptorcast

monad-blocksync/src/blocksync.rs
fabiotdf: Duplication is not an issue here, as `secondary_raptorcast_peers` is now a nodeId -> Round map

monad-blocksync/src/blocksync.rs
fabiotdf: Done

monad-blocksync/src/blocksync.rs
```diff
@@ -232,12 +233,28 @@ where
...
+
+        // Push back existing peer's expiry round, or insert new if not found
+        for peer in peers_excl_self {
+            let expiry = self
+                .secondary_raptorcast_peers
+                .get(&peer)
+                .copied()
+                .unwrap_or(Round(0));
+            let max_expiry = std::cmp::max(expiry, round_span.end);
+            self.secondary_raptorcast_peers.insert(peer, max_expiry);
```
michael-yxchen: nit
```suggestion
            self.secondary_raptorcast_peers
                .entry(peer)
                .and_modify(|expiry| *expiry = (*expiry).max(round_span.end))
                .or_insert(round_span.end);
```

monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -546,8 +546,14 @@ where
...
                         );
 
+                        let round_span = RoundSpan {
+                            start: confirm_msg.prepare.start_round,
+                            end: confirm_msg.prepare.end_round,
+                        };
+
                         ret = Poll::Ready(Some(
                             RaptorCastEvent::SecondaryRaptorcastPeersUpdate(
+                                round_span,
```
eerkaijun: is it true we only need round_span.end in blocksync module? if so, I think we can just pass `end` instead of needing to pass the entire RoundSpan

monad-state/src/lib.rs
```diff
@@ -361,6 +361,18 @@ where
...
     }
+
+    fn current_round(&self) -> Round {
+        match self {
+            ConsensusMode::Sync {
+                high_certificate, ..
+            } => match high_certificate {
+                RoundCertificate::Qc(qc) => qc.get_round(),
+                RoundCertificate::Tc(tc) => tc.round,
+            },
```
michael-yxchen:
```suggestion
            ConsensusMode::Sync {
                high_certificate, ..
            } => high_certificate.round() + Round(1),
```

monad-raptorcast/src/raptorcast_secondary/mod.rs
fabiotdf: Fixed. Originally I didn't have current_round(), so I was using round_span.start instead

monad-blocksync/src/blocksync.rs
fabiotdf: fixed

monad-state/src/lib.rs
fabiotdf: fixed

monad-types/src/lib.rs
```diff
@@ -115,7 +115,7 @@ impl Debug for Round {
     }
 }
 
-#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+#[derive(Copy, Clone, PartialEq, Eq, Debug, RlpEncodable, RlpDecodable, Serialize)]
```
michael-yxchen: super nit: revert

michael-yxchen (approved):

ariqchowdhury (approved):

PR #2429, opened 2025-10-06
Txgen Extreme Values Generator

eclark0426:


monad-eth-testutil/examples/txgen/generators/extreme_values.rs
```diff
@@ -0,0 +1,203 @@
...
+impl Generator for ExtremeValuesGenerator {
+    fn handle_acct_group(
+        &mut self,
+        accts: &mut [SimpleAccount],
+        ctx: &GenCtx,
+    ) -> Vec<(TxEnvelope, Address)> {
+        let mut txs = Vec::with_capacity(self.tx_per_sender * accts.len());
+
+        for sender in accts {
+            for _ in 0..self.tx_per_sender {
```
JoeHowarth: You're sending 2x the tx per sender here

monad-eth-testutil/examples/txgen/generators/extreme_values.rs
```diff
@@ -0,0 +1,203 @@
...
+            combination.max_fee_per_gas,
+            ctx.chain_id,
+            Some(combination.gas_limit),
+            Some(combination.max_priority_fee_per_gas as u64),
+        );
+
+        if combination.nonce.is_none() {
+            sender.nonce += 1;
+        } else {
+            sender.nonce = nonce + 1;
```
JoeHowarth: This means 0 is never used for nonce. If the combo gives a specific nonce, why don't we use it without adding 1?

monad-eth-testutil/examples/txgen/workers/rpc_sender.rs
```diff
@@ -114,7 +116,11 @@ impl RpcSender {
                     interval = self.maybe_update_interval(interval, now);
                 }
 
-                self.spawn_send_batch(batch);
+                // Server only accepts payloads under 2MB - split into multiple smaller batches if needed
+                let size_aware_batches = self.create_size_aware_batches(batch);
+                for sub_batch in size_aware_batches {
+                    self.spawn_send_batch(&sub_batch);
```
JoeHowarth, line 122: Good change - I've hit this issue before and always just adjusted the group size and tx per sender to avoid the limit, but this is much cleaner 👍

monad-eth-testutil/examples/txgen/workers/rpc_sender.rs
```diff
@@ -222,6 +228,48 @@ impl RpcSender {
...
+        );
+
+        batches
+    }
+
+    fn calculate_tx_payload_size(&self, tx: &TxEnvelope) -> usize {
+        let mut rlp_encoded_tx = Vec::new();
+        tx.encode_2718(&mut rlp_encoded_tx);
+        // Encoding adds "0x" prefix and hex encoding doubles the size
+        rlp_encoded_tx.len() * 2 + 2
```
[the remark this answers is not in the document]
JoeHowarth: Add a comment saying this is calculating the json/hex encoding instead of binary?

JoeHowarth (approved):
🚢

omegablitz (approved):

monad-eth-testutil/examples/txgen/generators/extreme_values.rs
eclark0426: At one point I had it as offsets from the account nonce, instead of the exact nonce specified by the combination. Probably a typo when changing it to explicit values

PR #2431, opened 2025-10-06
Txgen: HighCallData GenMode uses erc20 call with 2^32 bytes calldata

JoeHowarth:
Previously high call data txs with too low gas limit were not rejected until getting to execution, but now they are rejected at rpc. This means that we cannot fill blocks to stress proposal creation and raptor cast. This pr fixes this


omegablitz (approved):

omegablitz:
HighCallData also needs to require the Erc20 contract to be deployed

ariqchowdhury (approved):

PR #2433, opened 2025-10-07
secp: update secp256k1 to 0.31

dshulyak:
using newer secp in the authentication protocol, there were some changes in api.
also erase can be now used on inner type without additional unsafe block

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1912,6 +1912,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "bitcoin-io"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0b47c4ab7a93edb0c7198c5535ed9b52b63095f4e9b45279c6736cec4b856baf"
+
+[[package]]
+name = "bitcoin_hashes"
+version = "0.14.0"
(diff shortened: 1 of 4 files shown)


omegablitz (approved):

dshulyak:
wasm compilation is broken in 0.31. looks like version bump is blocked until new secp version is released that will include this fix https://github.com/rust-bitcoin/rust-secp256k1/pull/818

omegablitz:
Weird, secp shouldnt be a transitive dep in the wasm build. That seems unintended.

omegablitz:
It's because of this in monad-eth-types:
```rust
pub trait ExtractEthAddress {
    fn get_eth_address(&self) -> Address;
}

impl ExtractEthAddress for NopPubKey {
    fn get_eth_address(&self) -> Address {
        Address::new([0_u8; 20])
    }
}

impl ExtractEthAddress for SecpPubkey {
    fn get_eth_address(&self) -> Address {
        Address::from_raw_public_key(&Self::bytes(self)[1..])
    }
}
```

dshulyak:
moved implementation of ExtractEthAddress for Pubkey to monad-secp to avoid secp imports in monad-eth-types

PR #2436, opened 2025-10-07
Move SecondaryRaptorcastEvent from MonadEvent to BlockSyncEvent

fabiotdf:


monad-state/src/blocksync.rs
```diff
@@ -135,6 +135,11 @@ where
                 block_sync_wrapper.handle_peer_response(sender, response)
             }
             BlockSyncEvent::Timeout(request) => block_sync_wrapper.handle_timeout(request),
+            BlockSyncEvent::SecondaryRaptorcastEvent { event } => {
+                self.block_sync
+                    .set_secondary_raptorcast_peers(event.confirm_group_peers);
+                vec![]
```
eerkaijun: nit: use Vec::new() to be consistent

monad-raptorcast/src/lib.rs
```diff
@@ -981,10 +981,10 @@ where
                 }
             }
             RaptorCastEvent::SecondaryRaptorcastPeersUpdate(peer_list) => {
-                let event_data = SecondaryRaptorcastEvent {
+                let event = SecondaryRaptorcastEvent {
```
eerkaijun: I think we can delete this entire SecondaryRaptorcastEvent from the codebase

monad-executor-glue/src/lib.rs
```diff
@@ -867,6 +867,10 @@ where
     SelfResponse {
         response: BlockSyncResponseMessage<ST, SCT, EPT>,
     },
+    /// Events for secondary raptorcast updates
+    SecondaryRaptorcastEvent {
+        event: SecondaryRaptorcastEvent<SCT>,
```
eerkaijun: this would then just be something like
```
SecondaryRaptorcastPeersUpdate {
    peers: Vec<NodeId<SCT::NodeIdPubKey>>,
},
```

monad-raptorcast/src/lib.rs
fabiotdf: In a different branch, I'm adding a RoundSpan field to this struct

monad-executor-glue/src/lib.rs
fabiotdf: Yes but I'm adding a round span elsewhere, so the struct would be cleaner in that case

monad-raptorcast/src/lib.rs
eerkaijun: i don't think we need a separate SecondaryRaptorcastEvent struct though. it can just be passed directly to the BlocksyncEvent

monad-executor-glue/src/lib.rs
```diff
@@ -867,6 +867,10 @@ where
     SelfResponse {
         response: BlockSyncResponseMessage<ST, SCT, EPT>,
     },
+    /// Events for secondary raptorcast updates
+    SecondaryRaptorcastEvent {
```
eerkaijun: suggested name: SecondaryRaptorcastPeersUpdate

monad-raptorcast/src/lib.rs
fabiotdf: Alright

monad-state/src/blocksync.rs
fabiotdf: Done

eerkaijun (approved):

monad-executor-glue/src/lib.rs
fabiotdf: Done

PR #2438, opened 2025-10-08
monad-dataplane: restore to full 1500 mtu

dshulyak:
additional 20 bytes of useful space

### DIFF
--- a/monad-dataplane/src/udp.rs
+++ b/monad-dataplane/src/udp.rs
@@ -35,9 +35,7 @@ pub(crate) enum UdpMessageType {
     Direct,
 }
 
-// When running in docker with vpnkit, the maximum safe MTU is 1480, as per:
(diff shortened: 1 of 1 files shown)


dshulyak:
@eclark0426 you mentioned some time ago that we don't rely on this for testing anymore, is that right?

eclark0426:
Correct, this was needed when we used rootless Docker, which added a 20-byte networking overhead. Most things are now on systemd, removing Docker entirely, but for the things that do still use Docker, they now have the daemon running as root, which does not have the same 20 byte overhead

omegablitz:
This is backwards compatible because it only affects the chunking on the egress side, correct?

dshulyak:
@omegablitz yes, any updated node will start sending larger chunks, old nodes will be able to receive and re-transmit as long as they are smaller than ETHERNET_MTU

dshulyak:
@eclark0426 do you mind approving it?

PR #2441, opened 2025-10-08
Move broadcast bit validation to parse message

michael-yxchen:


monad-raptorcast/src/udp.rs
```diff
@@ -173,22 +173,16 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
...
-                (true, true) => {
-                    // invalid to have both primary and secondary broadcast bit set
-                    debug!(
-                        ?parsed_message.author,
-                        "Receiving invalid message with both broadcast and secondary broadcast bit set"
-                    );
-                    continue;
-                }
-            };
+            let maybe_broadcast_mode =
```
omegablitz: Can maybe_broadcast_mode be moved to the ValidatedMessage type now? So we don't need to do this match here

omegablitz (approved):

monad-raptorcast/src/udp.rs
```diff
@@ -506,8 +498,9 @@ where
         unix_ts_ms,
         app_message_hash,
         app_message_len,
-        broadcast,
-        secondary_broadcast,
+        // broadcast,
+        // secondary_broadcast,
```
eerkaijun: remove comments?

michael-yxchen: oops thanks. removed

PR #2442, opened 2025-10-08
Overload Raptor Chunk Header Epoch

michael-yxchen:
A full node may join new RaptorCast groups from the same validator while it's syncing or stuck. In such cases, it may rebroadcast to an expired group because its local round hasn't advanced, and expired groups are not garbage-collected. As a result, many full nodes end up receiving unsolicited Raptor chunks from groups they no longer belong to.

The root cause is that the node selects the rebroadcast group based on its local round. This PR resolves the issue by overloading the epoch field in the Raptor chunk header setting it to the round number for secondary raptorcast. This allows syncing or stuck full nodes to use the round number to identify and service the correct group

### DIFF
--- a/monad-raptorcast/src/raptorcast_secondary/mod.rs
+++ b/monad-raptorcast/src/raptorcast_secondary/mod.rs
@@ -48,6 +48,7 @@ use crate::{
     config::{RaptorCastConfig, SecondaryRaptorCastMode},
     message::OutboundRouterMessage,
     packet::{RetrofitResult as _, UdpMessageBatcher},
+    udp::GroupId,
     util::{BuildTarget, FullNodes, Group, Redundancy},
     OwnedMessageBuilder, RaptorCastEvent, UNICAST_MSG_BATCH_SIZE,
 };
@@ -149,7 +150,7 @@ where
         let message_builder =
             OwnedMessageBuilder::new(config.shared_key, peer_discovery_driver.clone())
                 .segment_size(segment_size_for_mtu(config.mtu))
-                .epoch_no(current_epoch)
+                .group_id(GroupId::Primary(current_epoch))
                 .redundancy(redundancy);
 
         Self {
@@ -317,7 +318,7 @@ where
                             "RaptorCastSecondary UpdateCurrentRound (Publisher)"
                         );
                         self.curr_epoch = epoch;
-                        self.message_builder.set_epoch_no(epoch);
+                        self.message_builder.set_group_id(GroupId::Primary(epoch));
                         // The publisher needs to be periodically informed about new nodes out there,
                         // so that it can randomize when creating new groups.
                         let full_nodes = self
@@ -353,7 +354,11 @@ where
                     }
                 },
 
-                Self::Command::PublishToFullNodes { epoch, message } => {
+                Self::Command::PublishToFullNodes {
+                    epoch: _,
+                    round,
+                    message,
+                } => {
                     let _timer = DropTimer::start(Duration::from_millis(20), |elapsed| {
                         warn!(?elapsed, "long time to publish message")
                     });
@@ -412,7 +417,7 @@ where
                     // send to full nodes.
                     self.message_builder
                         .prepare()
-                        .epoch_no(epoch)
+                        .group_id(GroupId::Secondary(round))
                         .build_into(&outbound_message, &build_target, &mut sink)
                         .unwrap_log_on_error(&outbound_message, &build_target);
                 }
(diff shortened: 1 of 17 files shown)


monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -163,7 +163,7 @@ where
     }
 
     fn udp_build(
-        epoch: &Epoch,
+        group_id: u64,
```
xinyuan-dev: how about make it obvious which variant this is for with an enum type? e.g. `enum GroupId { Primary(Epoch), Secondary(Round) }`

monad-raptorcast/src/util.rs
```diff
@@ -559,17 +562,19 @@ where
     // before calling iterate_rebroadcast_peers().
     pub fn check_source(
         &self,
-        msg_epoch: Epoch,
+        msg_group_id: u64,
         author_node_id: &NodeId<CertificateSignaturePubKey<ST>>,
+        src_addr: &SocketAddr,
```
xinyuan-dev, line 579: just want to confirm, is `src_addr` used only for logging? or is there any planned future uses e.g. interface with peer-disc for source ip verification

xinyuan-dev (approved):

monad-raptorcast/src/util.rs
michael-yxchen: it's only used for logging. it's been useful to help us identify that unwanted rebroadcasted chunks came from stuck full nodes instead of the validator. after authenticated udp, we can remove this log

monad-raptorcast/src/udp.rs
```diff
@@ -173,22 +173,16 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
...
+            let maybe_broadcast_mode =
+                match (parsed_message.broadcast, parsed_message.secondary_broadcast) {
+                    (true, false) => Some(BroadcastMode::Primary),
+                    (false, true) => Some(BroadcastMode::Secondary),
+                    (false, false) => None,
+                    (true, true) => {
+                        // InvalidBroadcastBits
+                        unreachable!()
+                    }
+                };
```
ariqchowdhury: can the type of the valid parsed_message guarantee that the broadcast bits were valid so that we don't have this unreachable. A method on the type `fn broadcast_mode(&self) -> Option<BroadcastMode>` The new func for the ValidatedMessage can enforce that they are correct on construction

michael-yxchen: yeah i moved maybe_broadcast_mode to `ValidatedMessage` in the previous PR https://github.com/category-labs/monad-bft/pull/2441. i'm working on this PR, marking as draft

xinyuan-dev (approved):

monad-raptorcast/src/util.rs
```diff
@@ -588,15 +593,30 @@ where
...
                 // Source node id (validator) is already the key to the map, so
                 // we don't need to look into the group itself.
-                let author_found = self.fullnode_map.contains_key(author_node_id);
-                if !author_found {
-                    debug!(?author_node_id, ?self.fullnode_map,
+                if let Some(groups) = self.fullnode_map.get(author_node_id) {
+                    if let Some((_start_round, group)) = groups.range(..=msg_round).next_back() {
+                        if msg_round >= group.round_span.start && msg_round < group.round_span.end {
+                            return true;
+                        }
```
omegablitz, line 603: The else branch here should never get hit unless the upstream is malicious, correct?
Just asking because it's the only branch that isn't logged

monad-raptorcast/src/util.rs
omegablitz, line 1: I'm not that familiar with this file, but at a high level it looks reasonable to me

eerkaijun:
so just to confirm this change will be backwards compatible?

monad-raptorcast/src/util.rs
michael-yxchen: yeah that's if upstream is malicious or we lost a confirm group message

michael-yxchen:
here's how the change is backwards compatible. Old server put epoch number in group_id, and client ignores that field. New server put round number in group_id and client use that field to pick the rebroadcast group
1. if server (validator) updates first, client ignores the field -> no change in behavior
2. if client (FN) updates first, it will interpret epoch number as a round number and fail to find a group in its group map because epoch number is much smaller. Client will default to old behavior and rebroadcast to the first group

monad-raptorcast/src/util.rs
michael-yxchen: i edited it slightly so we also produce a log in that case

omegablitz (approved):

PR #2444, opened 2025-10-09
Upper bound full node selection pool at max_group_size

eerkaijun:
addresses https://github.com/category-labs/category-internal/issues/1819

### DIFF
--- a/monad-peer-discovery/src/discovery.rs
+++ b/monad-peer-discovery/src/discovery.rs
@@ -129,7 +129,9 @@ pub struct PeerDiscovery<ST: CertificateSignatureRecoverable> {
     pub epoch_validators: BTreeMap<Epoch, BTreeSet<NodeId<CertificateSignaturePubKey<ST>>>>,
     // initial bootstrap peers set in config file
     pub initial_bootstrap_peers: BTreeSet<NodeId<CertificateSignaturePubKey<ST>>>,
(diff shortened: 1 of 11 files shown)


monad-peer-discovery/src/discovery.rs
```diff
@@ -1164,9 +1192,16 @@ where
...
+            if self.is_pinned_node(&node_id) {
+                debug!(?node_id, "clearing participation info for pinned node");
+                if let Some(info) = self.participation_info.get_mut(&node_id) {
+                    info.status = SecondaryRaptorcastConnectionStatus::None;
+                }
+            } else {
+                debug!(?node_id, "removing non-participating peer");
+                self.participation_info.remove(&node_id);
+                self.routing_info.remove(&node_id);
+            }
```
michael-yxchen, line 1204: what's the effect of clearing participation info for prioritized nodes?

eerkaijun: this is for the case of full nodes transitioning into a validator, rather than keeping them as Connected status, it's better to clear them. before this change, we do nothing about pinned nodes and keep them as Connected (and continuously sending group invites to them).

so technically it's not related to this PR, but is something i observed on testnets and feels like it's relevant enough to be bundled together. A prioritized full node will always be sent group invite whether Connected or not so this diff doesn't affect prioritized nodes

PR #2446, opened 2025-10-09
systemd: update allowedCPU range

dhogaivannan:


debian/usr/lib/systemd/system/monad-bft.service
```diff
@@ -16,16 +16,17 @@ ExecStart=/usr/local/bin/monad-node \
     --ledger-path /home/monad/monad-bft/ledger \
     --triedb-path /dev/triedb \
     --otel-endpoint "http://0.0.0.0:4317" \
-    --statesync-sq-thread-cpu 1 \
-    --record-metrics-interval-seconds 1 \
+    --statesync-sq-thread-cpu 8 \
+    --record-metrics-interval-seconds 9 \
```
omegablitz: unintentional?

bhuan: Should be both 8?

bhuan:
I bumped execution to facilitate building a release for mfnet (pick up Bruce’s commit for SSTORE) off Vannan’s branch. Not sure if that’s desired for merging to master

debian/usr/lib/systemd/system/monad-bft.service
dhogaivannan: I think it was typo, fixed

PR #2447, opened 2025-10-09
Make Triedb::get_block_key stricter

omegablitz:
Previously, when looking up a block with a given sequence number, Triedb::get_block_key would silently default the block to the Finalized nibble. This is not incorrect per se, but would require the caller to also check for state_availability every time the caller wanted to operate on this block key.

Instead, we can make Triedb::get_block_key stricter so that an unfinalized block is never upcasted to a Finalized one.

Concretely, before this change, this was the safe way of using the api:
1. get_block_key(block_num)
    - if None, return Err
2. get_state_availability(block_num)
    - if None, return Err
3. get_account(block_num)
4. get_state_availability(block_num)
   - if None, return Err

After this change, the following usage of the api is safe:
1. get_block_key(block_num)
    - if None, return Err
2. get_account(block_num)
3. get_state_availability(block_num)
   - if None, return Err

The intuition here is that get_block_key will only return block keys that were at one point valid, which is an implicit get_state_availability check.

### DIFF
--- a/monad-rpc/src/chainstate/mod.rs
+++ b/monad-rpc/src/chainstate/mod.rs
@@ -41,7 +41,7 @@ use crate::{
     chainstate::buffer::{block_height_from_tag, ChainStateBuffer},
     eth_json_types::{BlockTagOrHash, BlockTags, FixedData, MonadLog, Quantity},
     handlers::eth::{
-        block::block_receipts,
+        block::{block_receipts, get_block_key_from_tag_or_hash},
         txn::{parse_tx_receipt, FilterError},
     },
     jsonrpc::{ArchiveErrorExt, JsonRpcError, JsonRpcResult},
@@ -71,12 +71,19 @@ impl From<monad_archive::prelude::Report> for ChainStateError {
     }
 }
 
-pub fn get_block_key_from_tag<T: Triedb>(triedb_env: &T, tag: BlockTags) -> BlockKey {
+// BlockTags::Latest
+pub fn get_latest_block_key(triedb_env: &impl Triedb) -> BlockKey {
+    triedb_env.get_latest_voted_block_key()
+}
+
+pub fn get_block_key_from_tag(triedb_env: &impl Triedb, tag: BlockTags) -> Option<BlockKey> {
     match tag {
         BlockTags::Number(n) => triedb_env.get_block_key(SeqNum(n.0)),
-        BlockTags::Latest => triedb_env.get_latest_voted_block_key(),
-        BlockTags::Safe => triedb_env.get_latest_voted_block_key(),
-        BlockTags::Finalized => BlockKey::Finalized(triedb_env.get_latest_finalized_block_key()),
+        BlockTags::Latest => Some(get_latest_block_key(triedb_env)),
+        BlockTags::Safe => Some(triedb_env.get_latest_voted_block_key()),
+        BlockTags::Finalized => Some(BlockKey::Finalized(
+            triedb_env.get_latest_finalized_block_key(),
+        )),
     }
 }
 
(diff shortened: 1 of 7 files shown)


monad-rpc/src/handlers/eth/gas.rs
```diff
@@ -257,14 +257,8 @@ pub async fn monad_eth_estimateGas<T: Triedb>(
...
     }
 
-    let block_key = get_block_key_from_tag(triedb_env, params.block);
-    let version_exist = triedb_env
-        .get_state_availability(block_key)
-        .await
-        .map_err(JsonRpcError::internal_error)?;
-    if !version_exist {
-        return Err(JsonRpcError::block_not_found());
-    }
```
andr-dev, line 267: This doesn't seem related to the block key change, why did we remove it?

omegablitz: This comment in triedb-env is relevant:
```rust
    /// returns None if block number was never known to be valid
    /// a known-to-be-finalized BlockKey will never return None
    ///
    /// the following sequence is safe:
    /// 1. get_block_key(block_num)
    ///    - if None, return Err
    /// 2. get_account(block_num)
    /// 3. get_state_availability(block_num)
    ///    - if None, return Err early
```

The point of this change is that now, you should never need to do state_availability checks immediately after looking up a block_key from triedb-env.

The race was that most of our rpcs did not do this state availability check. estimateGas is one that did, but it is no longer necessary.

omegablitz: Ah but I should change the `JsonRpcError::internal_error` immediately below this to `JsonRpcError::block_not_found` to account for this

andr-dev: Ah I might have misunderstood that comment then, I thought it meant that doing the checks in that order is a safe thing to do rather than the stricter `get_block_key` change makes it so you don't need to check for state availability.

omegablitz: The comment describes a safe thing to do _now_, but it wasn't before.
Before the change, this was the safe equivalent:
1. get_block_key(block_num)
    - if None, return Err
1. get_state_availability(block_num)
    - if None, return Err early
1. get_account(block_num)
1. get_state_availability(block_num)
    - if None, return Err early

With the change, 2 is no longer needed. So the above change is effectively just deleting 2.

monad-rpc/src/chainstate/mod.rs
```diff
@@ -158,7 +168,8 @@ impl<T: Triedb> ChainState<T> {
                     }
                 }
 
-                let block_key = get_block_key_from_tag(&self.triedb_env, block);
+                let block_key = get_block_key_from_tag(&self.triedb_env, block)
+                    .ok_or(ChainStateError::ResourceNotFound)?;
```
philipglazman, line 172: It's correct that if a user passes a block number, it should only fail if the number > voted? Making sure that we only exit early if the block is definitely not in archiver.

omegablitz: Yes, because voted blocks are unfinalized and won't be in the archiver

omegablitz:
I checked usages of that, but didn't see anything problematic. its only used for accessing the ChainState buffer which doesn't have this race afaik

monad-rpc/src/chainstate/mod.rs
```diff
@@ -559,16 +589,20 @@ impl<T: Triedb> ChainState<T> {
...
-                )
+                let from_block = get_block_key_from_tag(&self.triedb_env, from_block_tag)
+                    .ok_or(JsonRpcError::block_not_found())?
+                    .seq_num()
+                    .0;
+                let to_block = get_block_key_from_tag(&self.triedb_env, to_block_tag)
+                    .map(|key| key.seq_num().0)
+                    // to_block should be floored to latest_block
+                    .unwrap_or(latest_block_number)
+                    .min(latest_block_number);
```
andr-dev, line 600:
```rust
.map_or(latest_block_number, |key| key.seq_num().0.min(latest_block_number));
```

PR #2450, opened 2025-10-10
Collect metrics on chainstate data source

philipglazman:


monad-tracing-timing/src/lib.rs
```diff
@@ -173,50 +209,92 @@ where
...
+                        timing.histogram.record(elapsed, &attributes);
+                        
+                        let mut wait_attributes = vec![
+                            KeyValue::new("main", span.name()),
+                            KeyValue::new("type", "wait"),
+                        ];
+                        if let Some(ref source) = data_source {
+                            wait_attributes.push(KeyValue::new("chainstate_source", source.clone()));
+                        }
+                        timing.histogram.record(timing.wait_time, &wait_attributes);
```
dshulyak, line 297: i think this looks somewhat hackish, if needed urgently it is probably ok.

i plugged into tracing, so that i can get wait time for async spans, but otherwise spans can be extended with necessary API.

i think i would try to add Span::current().add_extra_timing_label(key, value), that can be accumulated into some array and then exported. impl for add_extra_timing_label should be somewhat similar to update_secondary. there are also tests that can be used to verify that it works as expected

PR #2456, opened 2025-10-13
docker/single-node: adding --use-prebuilt flag

mhumeSF:
Adds ability to reference existing images using a docker compose override file.


bhuan:
I was able to build with `--pre-built` last night (used the default `latest`) with WSL.

ariqchowdhury (approved):

PR #2457, opened 2025-10-13
Validate group message sender

eerkaijun:
Closes https://github.com/category-labs/category-internal/issues/1954


monad-raptorcast/src/lib.rs
```diff
@@ -988,3 +1000,17 @@ where
...
+    group_message: &FullNodesGroupMessage<ST>,
+) -> bool
+where
+    ST: CertificateSignatureRecoverable,
+{
+    match group_message {
+        FullNodesGroupMessage::PrepareGroup(msg) => &msg.validator_id == sender,
+        FullNodesGroupMessage::PrepareGroupResponse(msg) => &msg.node_id == sender,
+        FullNodesGroupMessage::ConfirmGroup(msg) => &msg.prepare.validator_id == sender,
+    }
```
dshulyak, line 1015: why not to just use sender instead of those fields? it seems those fields are redundant

eerkaijun: easier to make the message backwards compatible for now but agree they are redundant

PR #2458, opened 2025-10-13
dataplane: refactor to use socket handles

dshulyak:
this pr makes dataplane api to be based around socket handles, rather than writer/reader. caller will be able to setup multiple udp sockets and interact with them through distinct handles. specifically this is meant to simplify integration of the following changes:

- authentication protocol will be listening on a distinct socket, and at least during rolling upgrade both sockets will coexist
- point to point traffic (specifically tx ingestion) will be eventually moved to a different socket, that will allow using different framing instead of raptorcast and applying independent firewall rules for it

### DIFF
--- a/monad-dataplane/src/lib.rs
+++ b/monad-dataplane/src/lib.rs
@@ -14,6 +14,7 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 use std::{
+    fmt::Debug,
     net::{IpAddr, SocketAddr},
     num::NonZeroU32,
     sync::{
@@ -39,7 +40,10 @@ pub(crate) mod buffer_ext;
 pub mod tcp;
 pub mod udp;
 
-pub(crate) use udp::UdpMessageType;
+pub struct UdpSocketConfig {
+    pub socket_addr: SocketAddr,
+    pub label: String,
+}
 
 pub struct DataplaneBuilder {
     local_addr: SocketAddr,
@@ -49,7 +53,7 @@ pub struct DataplaneBuilder {
     udp_buffer_size: Option<usize>,
     tcp_config: TcpConfig,
     ban_duration: Duration,
-    direct_socket_port: Option<u16>,
+    udp_sockets: Vec<UdpSocketConfig>,
 }
 
 impl DataplaneBuilder {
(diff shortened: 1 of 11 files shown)


ariqchowdhury (approved):

monad-dataplane/src/lib.rs
```diff
@@ -39,7 +40,10 @@ pub(crate) mod buffer_ext;
 pub mod tcp;
 pub mod udp;
 
-pub(crate) use udp::UdpMessageType;
+pub struct UdpSocketConfig {
+    pub socket_addr: SocketAddr,
+    pub label: String,
```
xinyuan-dev: Just want to point out that the three concepts:

- socket id
- socket label
- udp port

seems to have some overlaps semantics - all must be unique, and all are somehow used for addressing which socket to send. but there are some differences:

- socket id: used internally for routing udp message to specific socket
- socket label: used externally but mostly for logging, i suppose it could also be used in name records?
- udp port: unique, used internally, protocol

I'm wondering if we can merge the first two concepts - that we can define a registry in dataplane, of the expected udp services (raptorcast, direct, auth, etc) and support these services only, each designated with an integer id that can be used in the messaging.

monad-dataplane/src/lib.rs
```diff
@@ -630,31 +545,204 @@ impl DataplaneWriter {
...
+    pub src_addr: SocketAddr,
+    pub payload: Bytes,
+}
+
+pub struct TcpMsg {
+    pub msg: Bytes,
+    pub completion: Option<oneshot::Sender<()>>,
+}
+
+pub(crate) struct UdpMsg {
```
xinyuan-dev, line 627: I'm hoping that this message (or something like this) can be exposed to public that can be sent to a `UdpSocketWriter` directly - it may save the unnecessary grouping of chunks produced by raptorcast builder.

monad-dataplane/src/lib.rs
```diff
@@ -174,421 +193,317 @@ impl DataplaneBuilder {
...
-    pub completion: Option<oneshot::Sender<()>>,
-}
-
-pub(crate) struct UdpMsg {
-    pub(crate) dst: SocketAddr,
-    pub(crate) payload: Bytes,
-    pub(crate) stride: u16,
-    pub(crate) msg_type: UdpMessageType,
-    pub(crate) priority: UdpPriority,
+pub struct UdpSocketHandle {
```
xinyuan-dev, line 252: nit: this message may be called something else than UdpSocketHandle because it doesn't contain and socket and doesn't necessarily correspond to a (single) udp socket. `UdpHandle`/`UdpReader`/`UdpWriter` may be more precise.

monad-dataplane/src/udp.rs
```diff
@@ -251,8 +213,7 @@ async fn rx_single_socket(socket: UdpSocket, udp_ingress_tx: mpsc::Sender<RecvUd
 const PACING_SLEEP_OVERSHOOT_DETECTION_WINDOW: Duration = Duration::from_millis(100);
 
 async fn tx(
-    socket_tx: UdpSocket,
-    direct_socket_tx: Option<UdpSocket>,
+    tx_sockets: Vec<UdpSocket>,
```
xinyuan-dev, line 216: Just to confirm, we want all udp sockets to share the same message dispatch loop because we want them to share the bw control, right?

xinyuan-dev (approved):

monad-dataplane/src/udp.rs
dshulyak: yes, there is also prioritization, but it is essentially over bw

monad-dataplane/src/lib.rs
dshulyak: i don't mind renaming it to UdpHandle, but i am not sure why is it not necessarily correspond to a single socket, right now it does. the point of this refactoring was to separate how we handle protocols for different sockets

xinyuan-dev: It's my mistake here. When i was writing this, i was probably thinking that all messages are sent through the same channel, not directly to individual sockets. but it's indeed the case that the messages will eventually got sent through their corresponding socket. i hold no objection to the naming now. please ignore my comment here.

PR #2459, opened 2025-10-13
Build all docker images with debug symbols

omegablitz:


.github/workflows/publish.yaml
```diff
@@ -7,6 +7,7 @@ on:
       - 'staging**'
     tags:
       - "v[0-9]+.[0-9]+.[0-9]+"
+      - "v[0-9]+.[0-9]+.[0-9]+-.*"
```
omegablitz: this should be `"v[0-9]+.[0-9]+.[0-9]+-*"`
https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#filter-pattern-cheat-sheet

mhumeSF (approved):

PR #2462, opened 2025-10-15
Add defaults for AwsCliArgs

mikety:


monad-archive/src/cli.rs
```diff
@@ -282,8 +287,11 @@ pub struct AwsCliArgs {
     // TODO: remove me, concurrency should be handled elsewhere
     pub concurrency: usize,
     // If these are not provided, uses timeout_secs for all
+    #[serde(default = "get_default_bucket_timeout")]
```
mikety, line 290: Another approach is to use the init config if s3 parse have errors due to missing fields etc.

PR #2464, opened 2025-10-16
Fix known addresses when sending group invites

eerkaijun:
I notice that there is an inaccuracy at https://github.com/category-labs/monad-bft/blob/master/monad-raptorcast/src/raptorcast_secondary/mod.rs#L390 where the `known_addresses` passed in might not contain the addresses of the prioritized full nodes. We should use the full known addresses here. Tested on stressnet

### DIFF
--- a/monad-raptorcast/src/raptorcast_secondary/mod.rs
+++ b/monad-raptorcast/src/raptorcast_secondary/mod.rs
@@ -362,13 +362,12 @@ where
                             .peer_discovery_driver
                             .lock()
                             .unwrap()
-                            .get_secondary_fullnode_addrs();
-                        let full_nodes_vec: Vec<_> = full_nodes.keys().copied().collect();
+                            .get_secondary_fullnodes();
                         trace!(
                             "RaptorCastSecondary updating {} full nodes from PeerDiscovery",
-                            full_nodes_vec.len()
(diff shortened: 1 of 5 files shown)


monad-raptorcast/src/raptorcast_secondary/mod.rs
```diff
@@ -387,7 +386,12 @@ where
                                 );
                             }
 
-                            self.send_group_msg(group_msg, full_nodes_set, &full_nodes);
+                            let known_addresses = self
+                                .peer_discovery_driver
+                                .lock()
+                                .unwrap()
+                                .get_known_addresses();
+                            self.send_group_msg(group_msg, full_nodes_set, &known_addresses);
```
dshulyak, line 394: is it possible to assert in tests that the right nodes are included?

eerkaijun: we are discussing to build a swarm environment for secondary raptorcast testing (similar to `monad-peer-disc-swarm` or extending it), which would allow testing protocol logic without having to deal with actual raptorcast messages. right now the protocol logic and message handling are coupled together

michael-yxchen (approved):

dshulyak (approved):

PR #2468, opened 2025-10-17
read delegated status of eoa from the account

mikety:
Changes:

Adds ability to read delegation status from EOA accounts using the inline delegated account format in vicky/account-delegated.

Performance Testing:
Created flexnet tests with 1k, 3k, and 10k delegated accounts sending transactions to force BFT delegation status reads. Tests ran on  16 cores and 192 cores with 1 and 4 node topologies comparing legacy vs inline reads.
Results (average read time in ns):
1-node topology: nproc = 16
┌──────────┬─────────────┬─────────────┬─────────────┐
│ Accounts │ Legacy (ns) │ Inline (ns) │ Improvement │
├──────────┼─────────────┼─────────────┼─────────────┤
│ 1k              │ 8,748          │ 6,254        │ 29% faster  │
│ 3k             │ 8,193          │ 4,685         │ 43% faster  │
│ 10k           │ 6,455          │ 2,555         │ 60% faster  │
└──────────┴─────────────┴─────────────┴─────────────┘
1-node topology: nproc = 192
┌──────────┬─────────────┬─────────────┬─────────────┐
│ Accounts │ Legacy (ns) │ Inline (ns) │ Improvement │
├──────────┼─────────────┼─────────────┼─────────────┤
│ 1k              │ 30,340       │ 19,077        │ 37% faster  │
│ 3k             │ 30,457         │ 18,060      │ 41% faster  │
│ 10k           │ 28,484         │ 17,821       │ 37% faster  │
└──────────┴─────────────┴─────────────┴─────────────┘
4-node topology: nproc = 192
┌──────────┬─────────────┬─────────────┬─────────────┐
│ Accounts │ Legacy (ns) │ Inline (ns) │ Improvement │
├──────────┼─────────────┼─────────────┼─────────────┤
│ 1k             │ ~40,000     │ ~28,000     │ 30% faster  │
│ 3k            │ ~39,000     │ ~27,000      │ 31% faster  │
│ 10k           │ ~38,000     │ ~25,000     │ 34% faster  │
└──────────┴─────────────┴─────────────┴─────────────┘
Conclusion: Inline reads consistently outperform legacy reads across all test configurations (29-60% improvement).

Dependent on https://github.com/category-labs/monad/pull/1612

### DIFF
--- a/monad-eth-types/src/lib.rs
+++ b/monad-eth-types/src/lib.rs
@@ -18,17 +18,24 @@ use std::{fmt::Debug, ops::Deref};
 use ::serde::{Deserialize, Serialize};
 use alloy_consensus::{transaction::Recovered, Header, TxEnvelope};
 use alloy_eips::eip7702::RecoveredAuthorization;
-use alloy_primitives::{Address, B256};
+use alloy_primitives::{b256, Address, FixedBytes, B256};
 use alloy_rlp::{
     Decodable, Encodable, RlpDecodable, RlpDecodableWrapper, RlpEncodable, RlpEncodableWrapper,
 };
 use monad_crypto::NopPubKey;
 use monad_types::{Balance, ExecutionProtocol, FinalizedHeader, Nonce, SeqNum};
 use serde_with::{serde_as, DisplayFromStr};
+use tracing::warn;
 
 pub mod serde;
 
 pub const EMPTY_RLP_TX_LIST: u8 = 0xc0;
+
+pub const EIP7702_DELEGATION_PREFIX: [u8; 3] = [0xef, 0x01, 0x00];
+pub const NULL_HASH: B256 =
+    b256!("0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470");
+const HASH_LENGTH: usize = 32;
+const INLINE_CODE_LENGTH: usize = 23;
 
 pub trait ExtractEthAddress {
     fn get_eth_address(&self) -> Address;
@@ -41,10 +48,60 @@ impl ExtractEthAddress for NopPubKey {
 }
 
 #[derive(Debug, Copy, Clone)]
+pub enum AccountCodeOrHash {
+    InlineCode(FixedBytes<INLINE_CODE_LENGTH>), // code for delegated per EIP7702 accounts.
+    CodeHash(B256),
+    IsEmpty,
+}
+
+impl Decodable for AccountCodeOrHash {
+    fn decode(buf: &mut &[u8]) -> alloy_rlp::Result<Self> {
+        const RLP_HASH_LENGTH: usize = HASH_LENGTH + 1;
+        const RLP_INLINE_CODE_LENGTH: usize = INLINE_CODE_LENGTH + 1;
+
+        let account: AccountCodeOrHash = match buf.len() {
+            0 => AccountCodeOrHash::IsEmpty,
+            RLP_HASH_LENGTH => match <[u8; HASH_LENGTH]>::decode(buf) {
+                Ok(x) => AccountCodeOrHash::CodeHash(B256::from(x)),
+                Err(e) => {
+                    warn!("rlp code_hash decode failed: {:?}", e);
+                    return Err(alloy_rlp::Error::Custom("code_hash decode failed"));
+                }
+            },
+            RLP_INLINE_CODE_LENGTH => match <[u8; INLINE_CODE_LENGTH]>::decode(buf) {
+                Ok(x) => {
+                    if x[..3] != EIP7702_DELEGATION_PREFIX {
+                        return Err(alloy_rlp::Error::Custom("unexpected inline code prefix"));
+                    }
+                    AccountCodeOrHash::InlineCode(FixedBytes::<INLINE_CODE_LENGTH>::from(x))
+                }
+                Err(e) => {
+                    warn!("rlp inline code decode failed: {:?}", e);
+                    return Err(alloy_rlp::Error::Custom("inline code decode failed"));
+                }
+            },
+            _ => {
+                return Err(alloy_rlp::Error::Custom(
+                    "stored code_hash length is unexpected",
+                ));
+            }
+        };
+
+        Ok(account)
+    }
+}
+
+impl Default for AccountCodeOrHash {
+    fn default() -> Self {
+        AccountCodeOrHash::IsEmpty
+    }
+}
+
+#[derive(Debug, Copy, Clone)]
 pub struct EthAccount {
     pub nonce: Nonce,
     pub balance: Balance,
-    pub code_hash: Option<B256>,
+    pub code_or_hash: AccountCodeOrHash,
     pub is_delegated: bool,
 }
 
--- a/monad-rpc/src/handlers/debug.rs
+++ b/monad-rpc/src/handlers/debug.rs
@@ -23,6 +23,7 @@ use alloy_primitives::{
 };
 use alloy_rlp::{Decodable, Encodable, RlpDecodable};
 use monad_archive::prelude::{ArchiveReader, BlockDataReader, IndexReader};
+use monad_eth_types::AccountCodeOrHash;
 use monad_rpc_docs::rpc;
 use monad_triedb_utils::triedb_env::{BlockKey, FinalizedBlockKey, Triedb};
 use monad_types::SeqNum;
@@ -723,13 +724,21 @@ async fn include_code_output<T: Triedb>(
             .get_account(block_key, contract_addr.0.into())
             .await
             .map_err(JsonRpcError::internal_error)?;
-        let code = triedb_env
-            .get_code(block_key, account.code_hash)
-            .await
-            .map_err(JsonRpcError::internal_error)?;
 
-        let decoded_code = hex::decode(&code)
-            .map_err(|_| JsonRpcError::internal_error("could not decode code".to_string()))?;
+        let decoded_code: Vec<u8> = match account.code_or_hash {
+            AccountCodeOrHash::CodeHash(hash) => {
+                let code = triedb_env
+                    .get_code_db(block_key, hash.into())
+                    .await
+                    .map_err(JsonRpcError::internal_error)?;
+                hex::decode(&code).map_err(|_| {
+                    JsonRpcError::internal_error("could not decode code".to_string())
+                })?
+            }
+            AccountCodeOrHash::InlineCode(inline_code) => inline_code.to_vec(),
+            AccountCodeOrHash::IsEmpty => vec![],
+        };
+
         frame.output = decoded_code.into();
     }
 
--- a/monad-rpc/src/handlers/eth/account.rs
+++ b/monad-rpc/src/handlers/eth/account.rs
@@ -14,6 +14,7 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 use alloy_primitives::B256;
+use monad_eth_types::AccountCodeOrHash;
 use monad_rpc_docs::rpc;
 use monad_triedb_utils::triedb_env::Triedb;
 use serde::Deserialize;
@@ -77,10 +78,14 @@ pub async fn monad_eth_getCode<T: Triedb>(
         .await
         .map_err(JsonRpcError::internal_error)?;
 
-    let code = triedb_env
-        .get_code(block_key, account.code_hash)
-        .await
-        .map_err(JsonRpcError::internal_error)?;
+    let code = match account.code_or_hash {
+        AccountCodeOrHash::InlineCode(inline_code) => format!("0x{:x}", inline_code),
+        AccountCodeOrHash::CodeHash(hash) => triedb_env
+            .get_code_db(block_key, hash.into())
+            .await
+            .map_err(JsonRpcError::internal_error)?,
+        AccountCodeOrHash::IsEmpty => "0x".to_string(),
+    };
 
     match triedb_env
         .get_state_availability(block_key)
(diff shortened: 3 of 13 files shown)


monad-triedb-utils/src/decode.rs
```diff
@@ -47,23 +47,49 @@ pub fn rlp_decode_account(account_rlp: Vec<u8>) -> Option<EthAccount> {
...
-    } else {
+    } else if buf.len() == 33 {
         match <[u8; 32]>::decode(&mut buf) {
             Ok(x) => Some(x),
             Err(e) => {
                 warn!("rlp code_hash decode failed: {:?}", e);
                 return None;
             }
         }
+    } else if buf.len() != 24 {
```
eerkaijun: what's the significance of length 33 and 24?

monad-triedb-utils/src/decode.rs
```diff
@@ -47,23 +47,49 @@ pub fn rlp_decode_account(account_rlp: Vec<u8>) -> Option<EthAccount> {
...
+        warn!(
+            "The stored code_hash length is not expected {:?}",
+            buf.len()
+        );
+        return None;
+    } else {
+        match <[u8; 23]>::decode(&mut buf) {
+            Ok(x) => {
+                is_delegated = true;
```
eerkaijun: this doesn't seem right -- is_delegated should only be true if it's prefixed with 0xef0100 right?

monad-triedb-utils/src/decode.rs
```diff
@@ -47,23 +47,49 @@ pub fn rlp_decode_account(account_rlp: Vec<u8>) -> Option<EthAccount> {
         return None;
     };
 
-    let is_delegated = false;
+    let mut is_delegated = false;
```
eerkaijun: i feel like having this `is_delegated` field here is error prone. it's defaulted to false for legacy accounts and only updated in `get_accounts_async`, whereas for new accounts it's the actual value here. I think they should be updated at the same place

monad-triedb-utils/src/decode.rs
Chen-Yifan: when the decoded code field is inline, they are always 23 bytes and represents delegated code based on how execution stores it, but yeah we should add an assertion checking the prefix here.

monad-eth-types/src/lib.rs
```diff
@@ -44,6 +44,7 @@ pub struct EthAccount {
     pub nonce: Nonce,
     pub balance: Balance,
     pub code_hash: Option<B256>,
+    pub inline_code: Option<FixedBytes<23>>,
     pub is_delegated: bool,
```
Chen-Yifan: I would suggest consolidating these three fields to a single `Vec<u8>` field but provide helper methods like `has_code()`, `is_inline_delegated()` and `get_code_hash()` (similar to the Account in execution). 
The only con of this is this will require computing code hash on demand for delegated accounts, which adds a keccak operation. However, this should only impact eth_getAccount in RPC iiuc, and delegated accounts are expected to be a small portion, thus the impact is minimal.

monad-triedb-utils/src/decode.rs
mikety: this is datatype length + 1 (for RLP), 
`23` bytes length indicates its the new format to store delegated code
`32 `bytes is the legacy code hash, so delegation status is determined after pulling legacy account code.

monad-eth-types/src/lib.rs
mikety: I  encapsulated code_hash and inline_code, and can decide later cache or recompute `code_hash`.
I think its better to keep `is_delegated` explicit.

monad-rpc/src/handlers/eth/account.rs
```diff
@@ -77,10 +77,14 @@ pub async fn monad_eth_getCode<T: Triedb>(
         .await
         .map_err(JsonRpcError::internal_error)?;
 
-    let code = triedb_env
-        .get_code(block_key, account.code_hash)
-        .await
-        .map_err(JsonRpcError::internal_error)?;
+    let code = if !account.code_or_hash.is_legacy() {
+        format!("0x{:x}", account.code_or_hash.inline_code.unwrap())
```
eerkaijun: would prefer using 
```
if let Some(code) = account.code_or_hash.inline_code else {
    ...
}
```
rather than using unwrap here

monad-rpc/src/handlers/debug.rs
```diff
@@ -711,11 +711,15 @@ async fn include_code_output<T: Triedb>(
...
             .await
             .map_err(JsonRpcError::internal_error)?;
-        let code = triedb_env
-            .get_code(block_key, account.code_hash)
-            .await
-            .map_err(JsonRpcError::internal_error)?;
 
+        let code = if !account.code_or_hash.is_legacy() {
+            format!("0x{:x}", account.code_or_hash.inline_code.unwrap())
+        } else {
```
eerkaijun: same as above

monad-triedb-utils/src/decode.rs
```diff
@@ -47,24 +47,22 @@ pub fn rlp_decode_account(account_rlp: Vec<u8>) -> Option<EthAccount> {
...
-                warn!("rlp code_hash decode failed: {:?}", e);
-                return None;
-            }
+    let code_or_hash = match AccountCodeOrHash::decode(&mut buf) {
+        Ok(v) => v,
+        Err(e) => {
+            warn!(
+                "inline code or hash decode failed: {:?}, buffer: {:?}",
+                e, buf
+            );
```
eerkaijun: for a new account, i.e. an account with no code deployed at its address, is this the codepath that will be reached? if so, `warn` seems excessive

eerkaijun: just want to confirm: a new account should have no code and code hash, or should it have code hash zero initialized?

monad-eth-types/src/lib.rs
```diff
@@ -39,11 +44,94 @@ impl ExtractEthAddress for NopPubKey {
...
+            let delegation_code = &inline_code[0..3];
+            delegation_code == EIP7702_DELEGATION_PREFIX
+        } else {
+            false
+        }
+    }
+
+    pub fn is_legacy(&self) -> bool {
+        self.inline_code.is_none()
+    }
```
eerkaijun: if we use the if let Some(code) matching pattern based on above comments, i think we can remove this function

monad-triedb-utils/src/lib.rs
```diff
@@ -177,29 +178,34 @@ impl TriedbReader {
...
-                                match res {
-                                    Some(data) => {
-                                        if data.len() >= 3 {
-                                            let delegation_code = &data[0..3];
-                                            eth_account.is_delegated =
-                                                delegation_code == [0xef, 0x01, 0x00];
-                                            if eth_account.is_delegated {
-                                                trace!(?eth_account, block_id = ?seq_num.0, "is_delegated == true");
+                        if eth_account.code_or_hash.is_legacy() {
+                            match eth_account.code_or_hash.code_hash {
```
eerkaijun: i think this can use the if let Some(...) match pattern again. if I understand correctly, only one of `code_hash` or `inline_code` will ever be populated, never both

monad-triedb-utils/src/decode.rs
mikety: For a new account both inline_code and code are `None` the `decode` should return `Ok`. This code is reached only when there is an error in processing data.

monad-triedb-utils/src/lib.rs
mikety: correct.

eerkaijun: if so, does it make sense to make it into an enum instead of struct?

mikety: it has a case when neither inline_code nor code_hash are set.

monad-eth-types/src/lib.rs
```diff
@@ -39,11 +44,94 @@ impl ExtractEthAddress for NopPubKey {
...
+
+impl AccountCodeOrHash {
+    pub fn code_hash(&self) -> [u8; HASH_LENGTH] {
+        self.code_hash.map_or([0u8; HASH_LENGTH], |bytes| bytes.0)
+    }
+
+    pub fn is_delegated(&self) -> bool {
+        if let Some(inline_code) = self.inline_code {
+            let delegation_code = &inline_code[0..3];
+            delegation_code == EIP7702_DELEGATION_PREFIX
```
Chen-Yifan: if inline_code contains value, the prefix check should be an assertion because execution enforces inline code to always be delegated code.

monad-eth-types/src/lib.rs
Chen-Yifan: agreed, your is_legacy() function is basically !is_delegated() if you turn the if there to an assertion. And it's not clear to me how to decide whether an account is in legacy format or not, because account stored with null hash and a code hash can exist in both format and new format.

monad-eth-types/src/lib.rs
Chen-Yifan: when will one need delegated account's code_hash? I don't think there's a use case for that (we don't have eth_getAccount that returns code_hash but only eth_getCode, but i could be missing something), then there's no keccak overhead for computing code_hash. 
Your new `AccountCodeOrHash` struct still seems quite complicated for checking account format and account type, it adds some mental overhead since it diverges from how execution structures it. Also it computes code_hash for inline delegated code which is not used. Any reason to not use a `Vec<u8>`? It would simplify the check where len = 0 means null hash, len = 23 means inline delegated code, len = 32 means it's a non null code hash.

monad-triedb-utils/src/lib.rs
Chen-Yifan: I think with your current implementation it sets both `inline_code` and `code_hash` when decoded code is 23-byte delegated code. Not that I agree with the design, but just stating what it currently does.
I think structuring the code_or_hash as an enum is better than what it currently is, just need to make the default to a NULL_HASH code hash, and in fn code_hash(), compute keccak when it's of inline_code type. Or more clearly, I would suggest changing code_or_hash to a simple Vec<u8>, where the byte length would be sufficient to indicate which code format it is. This would match how execution defines it as well.

monad-eth-types/src/lib.rs
```diff
@@ -18,17 +18,24 @@ use std::{fmt::Debug, ops::Deref};
...
 use monad_types::{Balance, ExecutionProtocol, FinalizedHeader, Nonce, SeqNum};
+use tracing::warn;
 
 pub mod serde;
 
 pub const EMPTY_RLP_TX_LIST: u8 = 0xc0;
 
+pub const EIP7702_DELEGATION_PREFIX: [u8; 3] = [0xef, 0x01, 0x00];
+pub const NULL_HASH: B256 =
+    b256!("0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470");
```
mikety, line 36: @Chen-Yifan - should NULL_HASH use the same value as in execution.

monad-triedb-utils/src/lib.rs
mikety: `enum` works better than `Vec` for representation, updated

monad-triedb-utils/src/lib.rs
```diff
@@ -186,30 +187,30 @@ impl TriedbReader {
...
-                                    Some(data) => {
-                                        if data.len() >= 3 {
-                                            let delegation_code = &data[0..3];
-                                            eth_account.is_delegated =
-                                                delegation_code == [0xef, 0x01, 0x00];
-                                            if eth_account.is_delegated {
-                                                trace!(?eth_account, block_id = ?seq_num.0, "is_delegated == true");
-                                            }
+                                if let Some(data) = res {
+                                    if data.len() >= 3 {
```
eerkaijun: it's probably more accurate to check length is 23 and the first 3 bytes match EIP7702_DELEGATION_PREFIX

monad-eth-types/src/lib.rs
```diff
@@ -40,11 +47,85 @@ impl ExtractEthAddress for NopPubKey {
...
+                    Ok(x) => {
+                        // validate data
+                        let res: Result<B256, _> = keccak256(x).as_slice().try_into();
+                        match res {
+                            Ok(_) => {}
+                            Err(e) => {
+                                warn!("validation error extracting code_hash {:?}", e);
+                                return Err(alloy_rlp::Error::Custom("error extracting code_hash"));
+                            }
+                        }
```
eerkaijun: do we need to perform hashing here? seems unneccessary

keccak always return 32 bytes anyways

monad-eth-types/src/lib.rs
```diff
@@ -40,11 +47,85 @@ impl ExtractEthAddress for NopPubKey {
     }
 }
 
+#[derive(Debug, Copy, Clone)]
+pub enum AccountCodeOrHash {
+    InlineCode(FixedBytes<INLINE_CODE_LENGTH>), // code for delegated per EIP7702 accounts.
+    CodeHash(B256),
+    IsEmpty(()),
```
eerkaijun: this can just be `IsEmpty` (instead of `IsEmpty(())`)

monad-eth-types/src/lib.rs
```diff
@@ -40,11 +47,85 @@ impl ExtractEthAddress for NopPubKey {
...
+
+impl AccountCodeOrHash {
+    pub fn code_hash(&self) -> [u8; HASH_LENGTH] {
+        match &self {
+            Self::CodeHash(data) => (*data).into(),
+            Self::InlineCode(data) => keccak256(data).into(),
+            Self::IsEmpty(()) => NULL_HASH.into(),
+        }
+    }
+}
```
eerkaijun: is this used anywhere?

monad-triedb-utils/src/decode.rs
eerkaijun: we already log once in `decode` function, do we need to log here again?

monad-eth-types/src/lib.rs
mikety: no, removed

PR #2469, opened 2025-10-17
Switch to modular  packet builder

xinyuan-dev:
This PR replaces `udp::build_messages` with `packet::build_messages`.

This PR adds a new builder-style API for building udp messages, intended for reusing parameters. I replaced the `RaptorCast{,Secondary}::udp_build` method with this new API.

Note on testing:

- I removed the regression test for oversized message because it relies on building invalid udp message with unusual length, which is non-trivial to do with the new builder. This test is no longer relevant because the current decoder cache [will not panic](https://github.com/category-labs/monad-bft/blob/e35648e5f6b39b4117bdecc437aa1fc17bcfa74e/monad-raptorcast/src/decoding.rs#L1417-L1418) on failed decoder initialization, as long as the decoder creation itself does not panic, which is covered by [a new test](https://github.com/category-labs/monad-bft/pull/2469/files#diff-26e9efe2d911d609a77921fe02741b938fd804059721df614a564bea3d925aa4R93).
- The compatibility test is removed since the original version is removed. The correctness of the builder is covered by the numerous tests of the original `udp::build_messages`.
- The correctness of the new builder API is covered transitively through those original tests.

### DIFF
--- a/monad-raptorcast/src/lib.rs
+++ b/monad-raptorcast/src/lib.rs
@@ -40,7 +40,6 @@ use monad_crypto::{
 use monad_dataplane::{
     udp::{segment_size_for_mtu, DEFAULT_MTU},
     BroadcastMsg, DataplaneBuilder, DataplaneReader, DataplaneWriter, RecvTcpMsg, TcpMsg,
-    UnicastMsg,
 };
 use monad_executor::{Executor, ExecutorMetrics, ExecutorMetricsChain};
 use monad_executor_glue::{
@@ -59,9 +58,7 @@ use monad_validator::signature_collection::SignatureCollection;
 use raptorcast_secondary::{group_message::FullNodesGroupMessage, SecondaryRaptorCastModeConfig};
 use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender};
 use tracing::{debug, debug_span, error, trace, warn};
-use util::{
-    unix_ts_ms_now, BuildTarget, EpochValidators, FullNodes, Group, ReBroadcastGroupMap, Redundancy,
-};
+use util::{BuildTarget, EpochValidators, FullNodes, Group, ReBroadcastGroupMap, Redundancy};
 
 pub mod config;
 pub mod decoding;
@@ -73,6 +70,9 @@ pub mod util;
 
 const SIGNATURE_SIZE: usize = 65;
 
+pub(crate) type OwnedMessageBuilder<ST, PD> =
+    packet::MessageBuilder<'static, ST, Arc<Mutex<PeerDiscoveryDriver<PD>>>>;
+
 pub struct RaptorCast<ST, M, OM, SE, PD>
 where
     ST: CertificateSignatureRecoverable,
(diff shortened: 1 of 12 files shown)


monad-raptorcast/src/packet/assembler.rs
```diff
@@ -481,7 +475,7 @@ impl<'a, PT: PubKey> MerkleBatch<'a, PT> {
         debug_assert!(self.chunks.len() <= 2usize.pow((depth - 1) as u32));
 
         for (leaf_index, chunk) in self.chunks.iter_mut().enumerate() {
-            let leaf_index = u8::try_from(leaf_index).map_err(|_| BuildError::MerkleTreeTooDeep)?;
+            let leaf_index = leaf_index as u8;
```
dshulyak, line 478: assertion on the number of chunks was moved to some other place?

monad-raptorcast/src/lib.rs
```diff
@@ -385,17 +351,15 @@ where
...
-                        self.redundancy,
-                        &known_addresses,
-                    );
-                    self.dataplane_writer
-                        .udp_write_unicast_with_priority(rc_chunks, priority);
+                    let build_target = BuildTarget::<ST>::PointToPoint(&to);
+                    if let Some(rc_chunks) = self
+                        .message_builder
+                        .clone()
+                        .build_unicast_msg(&outbound_message, &build_target)
```
dshulyak, line 357: is it possible to move immutable parts from message builder into some other type, and then instantiate builder with that type instead of copying builder every time?

monad-raptorcast/src/lib.rs
```diff
@@ -510,6 +474,8 @@ where
                         }
 
                         self.current_epoch = epoch;
+                        self.message_builder = self.message_builder.clone().epoch_no(epoch);
```
dshulyak: why epoch_no is updated like this and then below it is passed after clone? is it possible to pass self.current_epoch explicitly instead?

xinyuan-dev: fixed, now it's specified with a setter method operating on &mut self.

monad-raptorcast/src/lib.rs
xinyuan-dev: the clone should be cheap because there is no deep structs, but it's true that still have some extra unnecessary cost. please check out the the updated code that implements the immutable base + override pattern, which should avoid the necessary cloning.

monad-raptorcast/src/packet/assembler.rs
xinyuan-dev: the check should be no longer necessary as the merkle tree depth is capped to 9 at most, which translates to leaf index of `0..=255`.

monad-raptorcast/src/lib.rs
dshulyak: it is not so much about cost. builder doesn't need any state, it is initialized and then drops everything at the end of execution, but proposed api makes it so that it needs to be correctly updated. 

i still see that it is updated with current epoch and then sometimes it is overwritten. my preference would be to pass it explicitly:

```
MessageBuilder::new(cfg, current_epoch).init()
or
MessageBuilder::new(cfg, epoch).with_peer_lookup(...).init()
```

xinyuan-dev:
I see. The problem with `cfg` as the base struct is that we have to make deliberate decision in the builder on which fields to keep as the base (e.g. segment size) and which are expected to change (e.g. epoch). This decision however should be a logic inside raptorcast instance based on the pattern how it uses the builder. My solution to reconcile is with this base + override pattern, with the overriding struct to be extended based on all possible needs.

PR #2475, opened 2025-10-21
Refactor txpool last commit state

andr-dev:
This change pulls the `last_commit` field out of the `TrackedTxMap` and into the parent `EthTxPool` structure. The end goal is to make `TrackedTxMap` a datastructure that operates independently of monad txpool specific behaviors.

### DIFF
--- a/monad-eth-txpool/src/pool/mod.rs
+++ b/monad-eth-txpool/src/pool/mod.rs
@@ -27,7 +27,7 @@ use monad_chain_config::{
     ChainConfig, MockChainConfig,
 };
 use monad_consensus_types::{
-    block::{BlockPolicyError, ProposedExecutionInputs},
+    block::{BlockPolicyError, ConsensusBlockHeader, ProposedExecutionInputs},
     payload::RoundSignature,
 };
 use monad_crypto::certificate_signature::{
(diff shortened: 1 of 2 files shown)


monad-eth-txpool/src/pool/tracked/mod.rs
```diff
@@ -152,6 +143,7 @@ where
     pub fn create_proposal(
         &mut self,
         event_tracker: &mut EthTxPoolEventTracker<'_>,
+        last_commit: &ConsensusBlockHeader<ST, SCT, EthExecutionProtocol>,
```
omegablitz: can last_commit be deleted from here? conceptually, create_proposal shouldn't need this

monad-eth-txpool/src/pool/mod.rs
```diff
@@ -323,21 +334,27 @@ where
...
-            proposal_gas_limit,
-            proposal_byte_limit - system_txs_size,
-            block_policy,
-            extending_blocks.iter().collect(),
-            state_backend,
-            chain_config,
-            &self.chain_revision,
-            &self.execution_revision,
-        )?;
+        let user_transactions = if let Some(last_commit) = self.last_commit.as_ref() {
```
omegablitz, line 337: this wouldn't need to change if last_commit was omitted (and don't see why last_commit is relevant to create_proposal)

bamrith (approved):

PR #2477, opened 2025-10-22
RaptorCast: Stream chunks into dataplane

xinyuan-dev:
In raptorcasting, a message is first raptor-coded into chunks, then batched into merkle trees (<=32 chunks per tree), signed, and assembled into UDP messages for delivery to the dataplane.

Although the merkle tree structure reduces the number of signing operations by up to 32x, signing each merkle batch remains the most time-consuming step in raptorcasting, often accounting for up to 90% of message build time. As a result, the dataplane can remain underutilized while waiting for the packet builder to generate all messages.

This PR improves dataplane utilization by streaming assembled UDP messages as soon as each merkle batch is constructed and signed. This reduces the latency for a raptorcasted message to reach the dataplane and distributes message delivery more evenly across the entire building operation.

### DIFF
--- a/monad-raptorcast/src/lib.rs
+++ b/monad-raptorcast/src/lib.rs
@@ -55,10 +55,14 @@ use monad_peer_discovery::{
 };
 use monad_types::{DropTimer, Epoch, ExecutionProtocol, NodeId, Round, RouterTarget, UdpPriority};
 use monad_validator::signature_collection::SignatureCollection;
-use raptorcast_secondary::{group_message::FullNodesGroupMessage, SecondaryRaptorCastModeConfig};
 use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender};
 use tracing::{debug, debug_span, error, trace, warn};
 use util::{BuildTarget, EpochValidators, FullNodes, Group, ReBroadcastGroupMap, Redundancy};
+
+use crate::{
+    packet::{RetrofitResult as _, UdpMessageBatcher},
+    raptorcast_secondary::{group_message::FullNodesGroupMessage, SecondaryRaptorCastModeConfig},
+};
 
 pub mod config;
 pub mod decoding;
@@ -69,6 +73,10 @@ pub mod udp;
 pub mod util;
 
 const SIGNATURE_SIZE: usize = 65;
+
+// Number of chunks to aggregate into udp messages before sending to
+// the dataplane.
+pub const UNICAST_MSG_BATCH_SIZE: usize = 32;
 
 pub(crate) type OwnedMessageBuilder<ST, PD> =
     packet::MessageBuilder<'static, ST, Arc<Mutex<PeerDiscoveryDriver<PD>>>>;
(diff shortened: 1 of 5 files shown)


monad-raptorcast/src/lib.rs
```diff
@@ -320,15 +324,16 @@ where
                             return;
                         }
                     };
-                if let Some(rc_chunks) = self
-                    .message_builder
+
+                let mut sink = UdpMessageBatcher::new(32, |rc_chunks| {
+                    self.dataplane_writer.udp_write_unicast(rc_chunks);
```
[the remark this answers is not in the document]
xinyuan-dev: good bot!

dshulyak:
i thought you mentioned that we can't stream with round robin policy, we should switch to round robin first

monad-raptorcast/src/lib.rs
```diff
@@ -320,15 +324,17 @@ where
...
                         }
                     };
-                if let Some(rc_chunks) = self
-                    .message_builder
-                    .prepare()
-                    .epoch_no(epoch)
-                    .build_unicast_msg(&outbound_message, &build_target)
-                {
+
+                let mut sink = UdpMessageBatcher::new(32, |rc_chunks| {
```
dshulyak: does it have to be 32 because of the special signing scheme or why 32? maybe move into a constant with a comment

monad-raptorcast/src/packet/mod.rs
```diff
@@ -210,3 +210,154 @@ where
...
+    fn drop(&mut self) {
+        self.flush();
+
+        let now = Instant::now();
+        if now > self.deadline {
+            let elapsed = now - self.deadline + Self::TIMEOUT;
+            tracing::warn!(
+                "UdpMessageBatcher dropped after deadline exceeded, total bytes flushed: {}, took {:?}",
+                self.total_bytes,
+                elapsed
```
dshulyak: it feels out of place, why it can't be done outside of the batcher drop method? 
i think there is even a special helper for that in the codebase

monad-raptorcast/src/lib.rs
xinyuan-dev: it's merely a hint based on the fact we currently produce 32 udp messages at once. I can move it to a constant.

is this batching necessary from dataplane's perspective? since for each `UnicastMsg` it still gets separated out into `UdpMsg` (which is equivalent to `packet::UdpMessage`), just later in the pipeline. If batching not necessary it would be much simpler to just have dataplane to ingest these udp messages (chunks) individually.

monad-raptorcast/src/packet/mod.rs
xinyuan-dev: >  i think there is even a special helper for that in the codebase

I think you're talking about [`DropTimer`](https://github.com/category-labs/monad-bft/pull/2477/files#diff-165e3fb6cde39ab9afa89714f29231d1f9e13d6df6a0d119e83c01e39f85eec5L495). If we want to use DropTimer in UdpMessageBatcher we will need to put it as a field of UdpMessageBatcher so that it drops at the right time. However, if we want to output some information (e.g. app message len, etc) we need it to borrow fields inside `UdpMessageBatcher`. Doing so would be unnecessarily complicated (with Rc<RefCell<_>> or something like that), so I just opt to implement it from scratch here.

> it feels out of place, why it can't be done outside of the batcher drop method?

I felt the same. Let me try move it outside the batcher.

monad-raptorcast/src/lib.rs
dshulyak: it is somewhat useful, but not necessary. 
it will allow to batch submissions to io_uring predictably, but probably only for the first few, after that queue will fill up and we will use full batches anyway

michael-yxchen (approved):

PR #2478, opened 2025-10-22
peer-discovery: switch to v2 name record format as default

dshulyak:
requires: https://github.com/category-labs/monad-bft/pull/2042
integration: https://github.com/category-labs/monad-integration/pull/363

change switches to v2 format, should be released once required change is rolled out to the network


dshulyak:
superseded by auth work

PR #2483, opened 2025-10-23
RaptorCast: Round up chunk assignment

xinyuan-dev:
This PR implements the rounding up chunk assignment algorithm proposed in https://github.com/category-labs/monad-bft/pull/2215.

### DIFF
--- a/monad-raptorcast/src/util.rs
+++ b/monad-raptorcast/src/util.rs
@@ -230,9 +230,19 @@ pub fn compute_hash<PT>(id: &NodeId<PT>) -> NodeIdHash
 where
     PT: PubKey,
 {
+    let full_hash = compute_full_hash(&id.pubkey().bytes());
+    HexBytes(full_hash.0[..20].try_into().expect("20 bytes"))
+}
+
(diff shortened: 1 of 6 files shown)


monad-raptorcast/src/util.rs
```diff
@@ -230,9 +230,19 @@ pub fn compute_hash<PT>(id: &NodeId<PT>) -> NodeIdHash
 where
     PT: PubKey,
 {
+    let full_hash = compute_full_hash(&id.pubkey().bytes());
+    HexBytes(full_hash.0[..20].try_into().expect("20 bytes"))
+}
+
+pub fn compute_app_message_hash(app_msg: &[u8]) -> NodeIdHash {
```
michael-yxchen: nit: return type should be `AppMessageHash`

michael-yxchen (approved):

monad-raptorcast/src/util.rs
xinyuan-dev: fixed, thanks :)

omegablitz (approved):

michael-yxchen (approved):

PR #2484, opened 2025-10-23
Try fetching remote forkpoint.toml and validators.toml on startup

bamrith:


ariqchowdhury (approved):

monad-node/src/state.rs
```diff
@@ -159,9 +180,171 @@ impl NodeState {
...
+
+                    if remote_forkpoint_round < local_forkpoint_round - Round(200) {
+                        // warn user if remote configs are stale
+                        warn!(
+                            ?remote_forkpoint_round,
+                            ?local_forkpoint_round,
+                            "remote forkpoint 200 rounds older than local forkpoint"
+                        );
+                    }
+                }
```
michael-yxchen, line 366: i think we should log here so all possible branches are covered

monad-node/src/state.rs
```diff
@@ -159,9 +180,171 @@ impl NodeState {
...
+
+                    // if remote config is more recent, use that over local config
+                    if remote_forkpoint_round > local_forkpoint_round {
+                        info!(
+                            ?remote_forkpoint_round,
+                            ?local_forkpoint_round,
+                            "remote forkpoint newer than local forkpoint, using remote configs"
+                        );
+
+                        return Ok((remote_forkpoint_config, remote_validators_config));
```
michael-yxchen, line 349: if we decide to use remote forkpoints/validators, should we write them (especially validators.toml b/c it's updated less frequently) to disk and back up old ones? operator might want to inspect those files if forkpoints fail to validate later in startup for example

this also prevents storing mismatched forkpoint/validators on disk. say we fetched forkpoint/validators files from a future epoch. after consensus is live, we'll write new forkpoints to file every round but validators.toml will be stale, until we hit epoch boundary

bamrith: updated to overwrite `forkpoint.toml` and `validators.toml` with the remote configs if they are newer. also save old forkpoint/validators to `forkpoint.toml.local` and `validators.toml.local` if that happens.

michael-yxchen (approved):

PR #2491, opened 2025-10-27
Stricter balance check for eth_sendRawTransaction

eerkaijun:
Closes https://github.com/category-labs/category-external/issues/121

Reserve balance allows a transaction with insufficient transfer balance to be included in the block, as long as the balance is sufficient to cover the gas cost of the transaction. However, in most cases, this is due to user error. This PR changes this by introducing a flag passed into txpool, where the default is for RPC (`eth_sendRawTransaction`) to reject such transaction where the transfer balance is insufficient. Since this is a RPC change and not a protocol change, the flag can be toggled to allow such transaction. 

Flexnet tests cover both with and without the flag.

### DIFF
--- a/monad-eth-txpool/src/pool/mod.rs
+++ b/monad-eth-txpool/src/pool/mod.rs
@@ -42,7 +42,7 @@ use monad_eth_txpool_types::{EthTxPoolDropReason, EthTxPoolInternalDropReason, E
 use monad_eth_types::{EthBlockBody, EthExecutionProtocol, ExtractEthAddress, ProposedEthHeader};
 use monad_state_backend::{StateBackend, StateBackendError};
 use monad_system_calls::{SystemTransactionGenerator, SYSTEM_SENDER_ETH_ADDRESS};
-use monad_types::{DropTimer, Epoch, NodeId, Round, SeqNum};
+use monad_types::{Balance, DropTimer, Epoch, NodeId, Round, SeqNum};
 use monad_validator::signature_collection::SignatureCollection;
 use rayon::iter::{IntoParallelIterator, ParallelIterator};
 use tracing::{debug, error, info, warn};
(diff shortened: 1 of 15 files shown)


monad-eth-txpool/src/pool/mod.rs
```diff
@@ -735,6 +741,27 @@ where
 
         Ok(proposal.txs)
     }
+
+    /// Checks whether the sender has sufficient balance for the transaction.
+    /// If `txpool_allow_insufficient_transfer_balance_tx` is true, only the gas cost is checked.
+    /// Otherwise, both the gas cost and the value transfer are checked.
+    fn check_sender_balance(
```
omegablitz: nit: a name like `sender_balance_is_sufficient` or something similar would indicate the truthiness of this better imo, but don't feel too strongly

eerkaijun: sounds good, will update

omegablitz:
Thinking about this a bit more, should this stricter check only be done for rpc servers? It seems a bit weird for some validators to have this on and some validators off

eerkaijun:
only for rpc servers as in validators would not use this flag while full nodes will use this flag?

eerkaijun:
ok i think a better implementation is to have the flag passed in the rpc request, so the check is per request

omegablitz:
As an alternative, what if instead of threading it through rpc, it's a setting in monad-node (similar to before), but it only does the stricter check on *owned* txs?

eerkaijun:
i think the motivation is we eventually want to provide an interface for users / mev protocols to specify their inclusion preference, there will be a few more criterias where the users / mev protocols can pass in to indicate their preference, so it's more of a per user preference rather than the node operator preference. right now we don't have a clearly defined interface yet (@andr-dev has been collecting some feedback from external teams). we can probably see how this fit into the interface

andr-dev:
I've written up a patch to enable sidecars to prioritize tx insertion into the pool. This patch allows passing additional fields to the pool over IPC in a significantly cleaner way than what is proposed in this changeset. If this is not super urgent, let's wait until that lands and it'll help cut this change down significantly.

PR #2493, opened 2025-10-28
Improve wal2json performance

omegablitz:
Does the following:
1. Split out WALReader, WALWriter
2. Remove locking when writing to stdout
3. Add configurable parallelism, defaulted to 1


omegablitz:
I can split this if it's too hard to read.

dshulyak (approved):

PR #2498, opened 2025-10-29
txpool: limit forwarded transactions by proposal byte limit

dshulyak:


dshulyak:
will be superseded by https://github.com/category-labs/monad-bft/pull/2560/

PR #2499, opened 2025-10-29
activity diagram for full-node groups

fabiotdf:
This diagram was made with Visual Paradigm. Should I include the .vpp file to documents/ or just the .png file?


fabiotdf:
Note1: Is a commercial license needed?

PR #2500, opened 2025-10-29
count only known nonces in 7702 auth 

mikety:
This code was generated Claude Sonnet 4.5

### DIFF
--- a/monad-eth-testutil/src/lib.rs
+++ b/monad-eth-testutil/src/lib.rs
@@ -34,6 +34,7 @@ use alloy_signer_local::PrivateKeySigner;
 use monad_chain_config::{revision::ChainRevision, ChainConfig, MockChainConfig};
 use monad_consensus_types::{
     block::{ConsensusBlockHeader, ConsensusFullBlock, TxnFee},
+    nonce_usage::{NonceUsage, NonceUsageMap},
     payload::{ConsensusBlockBody, ConsensusBlockBodyInner, RoundSignature},
     quorum_certificate::QuorumCertificate,
 };
@@ -44,9 +45,7 @@ use monad_crypto::{
     NopKeyPair, NopSignature,
 };
 use monad_eth_block_policy::{
-    compute_txn_max_gas_cost,
-    nonce_usage::{NonceUsage, NonceUsageMap},
-    pre_tfm_compute_max_txn_cost, EthValidatedBlock,
+    compute_txn_max_gas_cost, pre_tfm_compute_max_txn_cost, EthValidatedBlock,
 };
 use monad_eth_types::{EthBlockBody, EthExecutionProtocol, ProposedEthHeader, ValidatedTx};
 use monad_secp::KeyPair;
@@ -272,7 +271,9 @@ fn compute_expected_txn_fees_and_nonce_usages(
     txs: &[Recovered<TxEnvelope>],
 ) -> (BTreeMap<Address, TxnFee>, NonceUsageMap) {
     let mut txn_fees: BTreeMap<_, TxnFee> = BTreeMap::new();
+    let mut nonce_usages = NonceUsageMap::default();
 
+    // Process transactions in forward order, matching validator behavior
     for eth_txn in txs.iter() {
         txn_fees
             .entry(eth_txn.signer())
@@ -286,58 +287,56 @@ fn compute_expected_txn_fees_and_nonce_usages(
                 first_txn_gas: compute_txn_max_gas_cost(eth_txn, BASE_FEE),
                 max_gas_cost: Balance::ZERO,
                 max_txn_cost: pre_tfm_compute_max_txn_cost(eth_txn),
-                is_delegated: false,
+                valid_authorization: None,
+                authorization_nonce: None,
             });
-    }
 
-    let nonce_usages = txs
-        .iter()
-        .flat_map(|t| {
-            let mut pairs = vec![(t.signer(), NonceUsage::Known(t.nonce()))];
+        // Mimic validator: add_known for transaction signer
+        nonce_usages.add_known(eth_txn.signer(), eth_txn.nonce());
 
-            if t.is_eip7702() {
-                if let Some(auth_list) = t.authorization_list() {
-                    for auth in auth_list {
-                        let authority = auth.recover_authority().unwrap();
+        // Process authorizations (matches validator logic at monad-eth-block-validator/src/lib.rs:456-504)
+        if eth_txn.is_eip7702() {
+            if let Some(auth_list) = eth_txn.authorization_list() {
+                for auth in auth_list {
+                    let authority = auth.recover_authority().unwrap();
 
-                        pairs.push((
-                            authority,
-                            NonceUsage::Possible(VecDeque::from_iter([auth.nonce()])),
-                        ));
-
-                        txn_fees
-                            .entry(authority)
-                            .and_modify(|e| {
-                                e.is_delegated = true;
-                            })
-                            .or_insert(TxnFee {
-                                first_txn_value: Balance::ZERO,
-                                first_txn_gas: Balance::ZERO,
-                                max_gas_cost: Balance::ZERO,
-                                max_txn_cost: Balance::ZERO,
-
-                                is_delegated: true,
-                            });
+                    // Mimic validator nonce_usages.entry(authority) logic
+                    match nonce_usages.entry(authority) {
+                        std::collections::btree_map::Entry::Occupied(mut o) => {
+                            match o.get_mut() {
+                                NonceUsage::Known(nonce) => {
+                                    // If auth nonce matches known nonce, increment it
+                                    if *nonce == auth.nonce {
+                                        *nonce += 1;
+                                    }
+                                }
+                                NonceUsage::Possible(possible_nonces) => {
+                                    possible_nonces.push_back(auth.nonce);
+                                }
+                            }
+                        }
+                        std::collections::btree_map::Entry::Vacant(v) => {
+                            v.insert(NonceUsage::Possible(VecDeque::from_iter([auth.nonce])));
+                        }
                     }
+
+                    txn_fees
+                        .entry(authority)
+                        .and_modify(|e| {
+                            e.valid_authorization = None;
+                        })
+                        .or_insert(TxnFee {
+                            first_txn_value: Balance::ZERO,
+                            first_txn_gas: Balance::ZERO,
+                            max_gas_cost: Balance::ZERO,
+                            max_txn_cost: Balance::ZERO,
+                            valid_authorization: None,
+                            authorization_nonce: None,
+                        });
                 }
             }
-            pairs
-        })
-        .rev()
-        .fold(
-            NonceUsageMap::default(),
-            |mut map, (address, nonce_usage)| {
-                match map.entry(address) {
-                    std::collections::btree_map::Entry::Vacant(v) => {
-                        v.insert(nonce_usage);
-                    }
-                    std::collections::btree_map::Entry::Occupied(mut o) => {
-                        o.get_mut().merge_with_previous(&nonce_usage);
-                    }
-                }
-                map
-            },
-        );
+        }
+    }
 
     (txn_fees, nonce_usages)
 }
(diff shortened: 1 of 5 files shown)


monad-eth-block-validator/src/lib.rs
```diff
@@ -473,16 +472,21 @@ where
...
+                    txn_fee.authorization_nonce = Some(recovered_auth.nonce());
+                    let mut is_valid_authorization = false;
+
                     match nonce_usages.entry(authority) {
                         BTreeMapEntry::Occupied(nonce_usage) => match nonce_usage.into_mut() {
                             NonceUsage::Known(nonce) => {
-                                if let Some(next) = nonce.checked_add(1) {
-                                    if next == recovered_auth.nonce() {
+                                if *nonce == recovered_auth.nonce() {
+                                    if let Some(next) = nonce.checked_add(1) {
```
mikety, line 483: match the spec.

monad-consensus-types/src/lib.rs
```diff
@@ -28,6 +28,7 @@ pub mod block_validator;
 pub mod checkpoint;
 pub mod metrics;
 pub mod no_endorsement;
+pub mod nonce_usage;
```
andr-dev, line 31: Nonce usage shouldn't live in consensus-types because it isn't a consensus type. The only reason the concept exists is because of an eth-specific nuance (EIP7702 authorizations). I see that in another commit, we used the NonceUsage map in a consensus-types function. Let's find another way of solving this.

monad-consensus-types/src/block.rs
```diff
@@ -356,7 +356,8 @@ pub struct TxnFee {
     pub first_txn_gas: Balance,
     pub max_gas_cost: Balance,
     pub max_txn_cost: Balance, // Used for pre TFM validation
-    pub is_delegated: bool,
+    pub valid_authorization: Option<bool>,
+    pub authorization_nonce: Option<Nonce>,
```
andr-dev, line 360: The next commit adds logic so that on every authorization we override this field with the authorization nonce being processed. This means that previous authorizations will be thrown away. This seems like an incorrect approach.

We likely want to keep track of all possible authorization nonces, similar to the way `NonceUsage::Possible` does.

monad-eth-block-validator/src/lib.rs
andr-dev, line 483: I think we might be confusing tx nonce and account nonce here. As the name suggests, NonceUsage tracks account nonces that have been used. Specifically, there are two variants:

NonceUsage::Known -> The highest used nonce
NonceUsage::Possible(...) -> The set of possibly used nonces (from authorizations)

This means that given a NonceUsage::Known(nonce), we _know_ that `nonce` has been used. Thus if there is an authorization, we want to check and see if it is equal to the nonce _after_ it (`next`)

This change however checks if the used nonce matches the authorization nonce, in which case it bumps the nonce. This is incorrect behavior.

monad-eth-block-validator/src/lib.rs
```diff
@@ -473,16 +472,21 @@ where
                         continue;
                     }
 
+                    let txn_fee = txn_fees.entry(authority).or_default();
+                    txn_fee.authorization_nonce = Some(recovered_auth.nonce());
```
andr-dev, line 476: This is overridden on each iteration of the loop, so previous authorizations will be thrown away. This does not seem desirable.

monad-eth-block-policy/src/lib.rs
```diff
@@ -262,6 +262,9 @@ where
                 "Reserve balance check range is not contiguous"
             );
 
+            // Merge nonce_usages from previous blocks
+            merged_nonce_usages.merge_with_previous_block(block.nonce_usages.map.iter());
```
andr-dev, line 266: As the name suggests, the `merge_with_previous_block` function merges a `NonceUsagesMap` with that of the previous block. The for loop however iterates over blocks in increasing order, merging the `NonceUsageMap`s forward. This will cause incorrect merging when the range iterator produces more than one block.

Additionally, it is unclear whether the `merged_nonce_usages` comes from a block being extended or from a previous block so that may also cause an incorrect merging in the one block case.

monad-eth-block-policy/src/lib.rs
```diff
@@ -444,13 +453,35 @@ where
             return Ok(());
         }
 
+        let mut is_delegated_now = false;
+
+        if let Some(is_valid_authorization) = block_txn_fees.valid_authorization {
+            if is_valid_authorization {
+                is_delegated_now = true;
+            } else if let Some(authorization_nonce) = block_txn_fees.authorization_nonce {
```
andr-dev, line 461: As alluded to previously, since we throw out all authorization nonces except for the last seen authorization, we only check the last authorization which is (probably) incorrect.

monad-eth-block-policy/src/lib.rs
```diff
@@ -902,90 +936,101 @@ where
...
-
-                    if emptying_txn_check_start == GENESIS_SEQ_NUM {
-                        emptying_txn_check_start += SeqNum(1);
-                    }
+        // Initialize merged_nonce_usages for tracking across all committed and extending blocks
+        // Start with all account nonces from the base block
+        let mut merged_nonce_usages = NonceUsageMap::default();
+        for (address, maybe_status) in addresses.iter().zip_eq(account_statuses.iter()) {
+            if let Some(status) = maybe_status {
+                merged_nonce_usages.add_known(**address, status.nonce);
```
andr-dev, line 944: We are again confusing account nonces and tx nonces. When we do account balance lookups, we retrieve the current account nonce which is one greater than the highest used tx nonce. Thus, if we want to build a NonceUsageMap then we need to add `status.nonce - 1` since this is the used nonce.

monad-eth-block-policy/src/lib.rs
```diff
@@ -902,90 +936,101 @@ where
...
+                let next_blocks = blocks
+                    .iter()
+                    .skip_while(move |block| block.get_seq_num() < next_validate);
+
+                for extending_block in next_blocks {
+                    assert_eq!(next_validate, extending_block.get_seq_num());
+
+                    // Merge nonce_usages from previous extending blocks
+                    merged_nonce_usages
+                        .merge_with_previous_block(extending_block.nonce_usages.map.iter());
```
andr-dev, line 995: This is merging blocks forward and will result in an invalid `NonceUsageMap`.

monad-eth-block-policy/src/lib.rs
```diff
@@ -3733,4 +3792,399 @@ mod test {
...
+        // Verify nonces are stored correctly
+        assert_eq!(
+            merged_nonce_usages.get(&address1),
+            Some(&NonceUsage::Known(5))
+        );
+        assert_eq!(
+            merged_nonce_usages.get(&address2),
+            Some(&NonceUsage::Known(10))
+        );
+    }
```
andr-dev, line 3817: Why are we unit testing the `NonceUsageMap`?

monad-eth-block-policy/src/lib.rs
```diff
@@ -3733,4 +3792,399 @@ mod test {
...
+        let address1 = secret_to_eth_address(S1);
+        let address2 = secret_to_eth_address(S2);
+
+        let mut merged_nonce_usages = NonceUsageMap::default();
+        merged_nonce_usages.add_known(address1, 5);
+
+        // Create a block with nonce usage
+        let mut block_nonce_usages = NonceUsageMap::default();
+        block_nonce_usages.add_known(address1, 6); // Won't override Known value
+        block_nonce_usages.add_known(address2, 10); // New address will be added
```
andr-dev, line 3832: This seems to be testing invalid behavior. The test states that in a previous block (`block_nonce_usages`) the highest used nonce of `address1` was `6` but in a future block (`merged_nonce_usages`) the highest used nonce of `address1` is 5.

Again, why are we unit testing nonce usages here???

monad-eth-block-policy/src/lib.rs
```diff
@@ -3733,4 +3792,399 @@ mod test {
...
+
+        // Should set is_delegated to true because sender's nonce matches authorization_nonce
+        let result =
+            validator.try_apply_block_fees(&mut account_balance, &fees, &address, &nonce_usages);
+
+        assert!(result.is_ok());
+        assert!(
+            account_balance.is_delegated,
+            "Account should be marked as delegated when sender nonce matches authorization_nonce"
+        );
```
andr-dev, line 3894: In the fees struct created above, `valid_authorization` is set to `Some(false)` with a comment indicating that the authorization is invalid. If so, why would the authorization be considered?

monad-eth-block-policy/src/lib.rs
```diff
@@ -3733,4 +3792,399 @@ mod test {
...
+            .map
+            .insert(address, NonceUsage::Possible(possible_nonces));
+
+        // Create a previous block with Known nonce
+        let mut previous_block_nonces = NonceUsageMap::default();
+        previous_block_nonces.add_known(address, 5);
+
+        // Merge: Possible([6, 7, 8]) with previous Known(5) should convert to Known(8)
+        // because 6 = 5+1, 7 = 6+1, 8 = 7+1 (all sequential)
+        current_nonce_usages.merge_with_previous_block(previous_block_nonces.map.iter());
```
andr-dev, line 4046: After this line, `current_nonce_usages` only has a `NonceUsage::Known` so the new code isn't being tested for possible -> known conversions. What exactly is this meant to be testing?

monad-eth-block-policy/src/lib.rs
```diff
@@ -3733,4 +3792,399 @@ mod test {
...
+        if let Some(NonceUsage::Possible(nonces)) = test_merge_possible.get(&address) {
+            assert!(
+                nonces.contains(&10),
+                "Possible nonces should still contain 10 after merging Possible with Possible"
+            );
+            assert_eq!(
+                nonces.clone(),
+                VecDeque::from(vec![3, 4, 6, 7, 10]),
+                "Merging Possible([6,7,10]) with Possible([3,4]) should prepend to get [3,4,6,7,10]"
+            );
```
andr-dev, line 4129: This is a duplicate check that can be removed.

monad-eth-block-policy/src/lib.rs
```diff
@@ -3733,4 +3792,399 @@ mod test {
...
+            panic!("Expected Possible nonces after Possible-to-Possible merge");
+        }
+
+        // Create a previous block with Known nonce
+        let mut previous_block_nonces = NonceUsageMap::default();
+        previous_block_nonces.add_known(address, 5);
+
+        // Merge: Possible([6, 7, 10]) with previous Known(5) should convert to Known(7)
+        // because 6 = 5+1, 7 = 6+1, but 10 != 7+1 so it stops
+        current_nonce_usages.merge_with_previous_block(previous_block_nonces.map.iter());
```
andr-dev, line 4140: Again, after this line the nonce usages only contains `NonceUsage::Known` so what is this testing?

These tests seem to be re-testing the `NonceUsage`/`NonceUsageMap` instead of the actual changes introduced in this commit. For example, what happens if there is more than one authorization? Valid authorization followed by invalid and vice versa? Multiple blocks? How about showing that `is_delegated` switches from true -> false based on how many blocks ago the authorization was present?

monad-eth-block-validator/src/lib.rs
mikety: Per https://eips.ethereum.org/EIPS/eip-7702  for each authorization tuple it verifies that nonce of authority  is equal to nonce from the auth tuple and after increases the nonce of authority.

monad-eth-block-validator/src/lib.rs
mikety: for each iteration it uses an authority address recovered from the current tuple. Is the case you refer when authorization account is the same for different tuples?

monad-eth-block-validator/src/lib.rs
andr-dev: Right, and as mentioned above the nonce usage map does **not** track the nonce of the authority. It tracks the last used nonce for that authority which implies that the nonce of the authority is the last used nonce + 1.

PR #2501, opened 2025-10-29
Remove unnecessary triedb code read in rpc traces

philipglazman:
Traces now include code in the response. RPC should only fetch code from triedb if the trace's output field is empty.


omegablitz:
Will this break historical traces?

Baltoli:
> Will this break historical traces?

Right, I think it needs to keep the lookup logic if there's no code in the RLP-deserialised call frame

Baltoli (approved):

ariqchowdhury (approved):

PR #2504, opened 2025-10-30
Only retry forwarding txs without a nonce gap

andr-dev:


omegablitz:
I didn't read this yet, but does this apply only to retrying forwarding? What's the downside of being overly strict and rejecting transactions with gaps from tracked in the first place?

PR #2506, opened 2025-10-31
Use txpool bridge channel size for ratelimiting

andr-dev:
Previously, we used a const in the eth txn handler file when ratelimiting `eth_sendRawTransaction`. Instead, we can use the capacity of the `EthTxPoolBridgeClient` channel since at most that many transactions can be sent to the bridge at a time. This prevents the `try_send` method from ever failing.


michael-yxchen:
The change itself is fine. Regarding the original issue raised by auditor, does returning "overloaded, try again later" expose internal state of that RPC endpoint?

michael-yxchen (approved):

PR #2514, opened 2025-11-03
Archive: archive generic dirs - supports historical exec-event archiving

JoeHowarth:
Ensures that all files in the specified directory and recursively through child dirs have been uploaded. 

This worker can run as part of a normal archive writer process OR as a standalone worker (i.e. does not read blocks/traces/receipts from a source and write to a sink)

Notes:
- gc's s3 known keys that are not present locally 
- gc's local known files in dirs that have not been modified recently (protect against cold dirs)
- protects against partial file writes (only archives after delay since last modified)
- generalized version of bft upload code

### DIFF
--- a/monad-archive/src/workers/generic_folder_archiver.rs
+++ b/monad-archive/src/workers/generic_folder_archiver.rs
@@ -1,0 +1,670 @@
+// Copyright (C) 2025 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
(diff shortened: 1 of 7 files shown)


monad-archive/src/workers/generic_folder_archiver.rs
```diff
@@ -0,0 +1,670 @@
...
+        Self {
+            known_in_s3: HashSet::new(),
+            last_seen_modified: None,
+            last_hot: Instant::now(),
+        }
+    }
+
+    fn mark_processed(&mut self, modified: Option<SystemTime>) {
+        self.last_hot = Instant::now();
+        self.last_seen_modified = modified.or(Some(SystemTime::now()));
```
andr-dev, line 42:
```rust
modified.unwrap_or_else(SystemTime::now)
```

PR #2520, opened 2025-11-05
First skip system sender

eerkaijun:
consensus rejects block that contains signed authorization from system sender. however txpool incorrectly skips authorization with invalid chain id, so an authorization with invalid chain id signed by system sender will get skipped instead of getting discarded. in this pr, the ordering of the checks are switched so we immediately discard the transaction if there's an authorization by system sender

### DIFF
--- a/monad-eth-txpool/src/pool/transaction.rs
+++ b/monad-eth-txpool/src/pool/transaction.rs
@@ -120,12 +120,6 @@ impl ValidEthTransaction {
                 match signed_authorizations
                     .iter()
                     .filter_map(|signed_authorization| {
-                        if signed_authorization.chain_id != 0
-                            && signed_authorization.chain_id != chain_id
-                        {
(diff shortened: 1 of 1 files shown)


monad-eth-txpool/src/pool/transaction.rs
```diff
@@ -135,6 +129,12 @@ impl ValidEthTransaction {
                             return Some(Err(EthTxPoolDropReason::InvalidSignature));
                         }
 
+                        if signed_authorization.chain_id != 0
+                            && signed_authorization.chain_id != chain_id
+                        {
+                            return None;
+                        }
```
dshulyak, line 136: could you please document somewhere why the previous version was wrong? maybe in the pr body

eerkaijun: updated

PR #2521, opened 2025-11-05
Serialize forkpoints as rlp

omegablitz:
Previously, forkpoints were only serialized as toml. If the toml serialization failed, the node would panic.

Now, forkpoints are serialized as both toml and rlp. The rlp serialization is considered the source of truth. If the toml serialization fails, the node will no longer panic. To maintain operational backwards compatibility, a node can be started up from either the toml or rlp formats.


michael-yxchen (dismissed):

bamrith (approved):

omegablitz:
Rebased

michael-yxchen (approved):

PR #2524, opened 2025-11-05
Validators to limit statesync requests

eerkaijun:
Statesync service rules:
 - If self is a validator: only service requests from validators or whitelisted full nodes
 - If self is a full node: service all requests

### DIFF
--- a/monad-state/src/lib.rs
+++ b/monad-state/src/lib.rs
@@ -14,7 +14,7 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 use std::{
-    collections::{BTreeMap, HashMap},
+    collections::{BTreeMap, HashMap, HashSet},
     fmt::Debug,
     marker::PhantomData,
     ops::Deref,
@@ -57,8 +57,8 @@ use monad_executor_glue::{
     BlockSyncEvent, ClearMetrics, Command, ConfigEvent, ConfigReloadCommand, ConsensusEvent,
     ControlPanelCommand, ControlPanelEvent, GetFullNodes, GetMetrics, GetPeers, LedgerCommand,
     MempoolEvent, Message, MonadEvent, ReadCommand, ReloadConfig, RouterCommand, StateSyncCommand,
-    StateSyncEvent, StateSyncNetworkMessage, TxPoolCommand, ValSetCommand, ValidatorEvent,
-    WriteCommand,
+    StateSyncEvent, StateSyncNetworkMessage, StateSyncRequest, TxPoolCommand, ValSetCommand,
+    ValidatorEvent, WriteCommand,
 };
 use monad_state_backend::StateBackend;
 use monad_types::{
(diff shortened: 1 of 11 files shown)


monad-statesync/src/outbound_requests.rs
```diff
@@ -275,6 +275,14 @@ impl<PT: PubKey> OutboundRequests<PT> {
         }
     }
 
+    pub fn handle_not_whitelisted(&mut self, from: NodeId<PT>) {
+        tracing::debug!(
+            ?from,
+            "peer does not serve statesync request, removing from peer list"
+        );
+        self.peers.remove(&from);
```
omegablitz, line 283: Maybe should clear out requests from in_flight_requests and add them to pending_requests here, but not sure if strictly necessary. I'll look more closely tmrw

omegablitz: Can you append this block:
```rust
let requests_to_remove: Vec<_> = self
    .in_flight_requests
    .iter()
    .filter(|(request, in_flight_request)| {
        in_flight_request.peer == from
    })
    .map(|(request, _)| *request)
    .collect();

for request in requests_to_remove {
    self.in_flight_requests.remove(&request);
    self.pending_requests.insert(request);
}
```

monad-state/src/lib.rs
```diff
@@ -997,8 +1039,25 @@ where
             }
             MonadEvent::StateSyncEvent(state_sync_event) => match state_sync_event {
                 StateSyncEvent::Inbound(sender, message) => {
-                    // TODO we need to add some sort of throttling to who we service... right now
-                    // we'll service indiscriminately
+                    // Filter statesync requests based on sender
```
omegablitz, line 1044: need to check that `message` is `StateSyncNetworkMessage::Request` before doing the `should_service_statesync_request` check

omegablitz: idk if this compiles, but something like this:
```rust
 match state_sync_event {
    StateSyncEvent::Inbound(sender, StateSyncNetworkMessage::Request(_)) if !self.should_service_statesync_request(&sender) => {
        tracing::debug!(
            ?sender,
            "dropping statesync request from non-whitelisted sender"
        );

        // Send NotWhitelisted response to sender so that it can look for other peers
            target: RouterTarget::TcpPointToPoint {
                to: sender,
                completion: None,
            },
            message: VerifiedMonadMessage::StateSyncMessage(
                StateSyncNetworkMessage::NotWhitelisted,
            ),
        })]
    }
    StateSyncEvent::Inbound(sender, message) => {
        vec![Command::StateSyncCommand(StateSyncCommand::Message((
            sender, message,
        )))]
    }
    // ...
}
```

omegablitz (approved):

monad-state/src/lib.rs
```diff
@@ -502,6 +505,42 @@ where
...
+    /// - If self is a validator: only service requests from validators or whitelisted full nodes
+    /// - If self is a full node: service all requests
+    fn should_service_statesync_request(
+        &mut self,
+        sender: &NodeId<CertificateSignaturePubKey<ST>>,
+    ) -> bool {
+        // Check if self is a validator
+        let self_role = self.get_role();
+
+        if self_role == Role::FullNode {
```
ariqchowdhury: can we match on this in case it ever expands

omegablitz: In the same vein, if you end up changing this, can you add &StateSyncRequest as a param to `should_service_statesync_request`, even if it's unused rn

ariqchowdhury (approved):

PR #2525, opened 2025-11-06
Evict txs from txpool when limits exceeded

andr-dev:
This change introduces an eviction policy for the txpool to mitigate DoS vectors when the pool hits its limits. When evicting, the pool selects the address with the lowest priority (highest value) and removes all the txs associated with that address.

The current eviction policy first looks at the size of the nonce gap to the first transaction (if any). After that, it prioritizes based on the max fee of the first 7 nonce-sequential transactions.

### DIFF
--- a/monad-eth-txpool/src/pool/tracked/list.rs
+++ b/monad-eth-txpool/src/pool/tracked/list.rs
@@ -18,16 +18,18 @@ use std::{
     time::{Duration, Instant},
 };
 
+use alloy_consensus::Transaction;
 use alloy_primitives::Address;
 use indexmap::map::VacantEntry;
 use monad_chain_config::{execution_revision::MonadExecutionRevision, revision::ChainRevision};
 use monad_eth_block_policy::nonce_usage::NonceUsage;
 use monad_eth_txpool_types::EthTxPoolDropReason;
+use monad_tfm::base_fee::MIN_BASE_FEE;
 use monad_types::Nonce;
 use tracing::error;
 
 use super::{limits::TrackedTxLimits, ValidEthTransaction};
-use crate::EthTxPoolEventTracker;
+use crate::{pool::tracked::priority::Priority, EthTxPoolEventTracker};
 
 /// Stores byte-validated transactions alongside the an account_nonce to enforce at the type level
 /// that all the transactions in the txs map have a nonce at least account_nonce. Similar to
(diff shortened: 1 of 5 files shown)


monad-eth-txpool/src/pool/tracked/list.rs
```diff
@@ -81,6 +82,29 @@ impl TrackedTxList {
...
+            .first_key_value()
+            .and_then(|(tx_nonce, _)| tx_nonce.checked_sub(self.account_nonce))
+            .unwrap_or(u64::MAX)
+            .min(u16::MAX as u64);
+
+        let mut max_fees = [u64::MAX; 7];
+
+        for (idx, tx) in self.get_queued(None).take(7).enumerate() {
+            max_fees[idx] = u64::MAX
+                - tx.max_fee_per_gas()
```
Haxatron: For non legacy txs, prefer to use effective tip = min(max_priority_fee_per_gas, max_fee_per_gas − MIN_BASE_FEE) instead? Otherwise non-legacy tx can set 0 max priority fee and extremely high max_fee_per_gas to crowd out txs from the pool.

monad-eth-txpool/src/pool/tracked/mod.rs
```diff
@@ -154,13 +178,17 @@ where
...
+            match self.txs.entry(address.clone()) {
+                IndexMapEntry::Occupied(tx_list) => {
+                    if TrackedTxList::update_committed_nonce_usage(
+                        event_tracker,
+                        &mut self.limits,
+                        tx_list,
+                        nonce_usage,
+                    ) {
+                        self.update_priority(event_tracker, address);
+                    }
```
Haxatron, line 202: Addresses remain dangling in the PriorityMap if updated_committed_nonce_usage returns false here and the tx list was deleted

monad-eth-txpool/src/pool/tracked/mod.rs
```diff
@@ -180,16 +208,22 @@ where
                 break;
             };
 
+            let address = entry.key().clone();
+
             if TrackedTxList::evict_expired_txs(event_tracker, &mut self.limits, entry, tx_expiry) {
                 continue;
             }
```
Haxatron, line 228: Dangling addresses in PriorityMap if returns true and the tx list was deleted

monad-eth-txpool/src/pool/tracked/mod.rs
```diff
@@ -208,5 +242,9 @@ where
                 execution_revision,
             )
         });
+
```
Haxatron, line 258: Some tx lists will be retained here after static_validate_all_txs is called, and for those txs, they are updated in the next few lines.

But for the tx lists that were removed, their addresses are left dangling in PriorityMap

monad-eth-txpool/src/pool/tracked/list.rs
msetrin: Agreed on this

msetrin: Just noting, that I think the most correct thing would be to compute the effective tip relative to the base fee of the most recent block rather than MIN_BASE_FEE, but MIN_BASE_FEE might be good enough for now if it is significantly easier to implement.

Haxatron: Agree on the above ^

Haxatron: To do that would require updating priorities for all addresses in the PriorityMap though which would be an O(n log n) operation.

andr-dev: Yeah that's the issue I ran into during implementation, we could keep the "stale" priority using the committed base fee from the last insertion time but that generally doesn't seem like a better solution than just using `MIN_BASE_FEE`.

omegablitz (dismissed):

ariqchowdhury (approved):

omegablitz (approved):

PR #2529, opened 2025-11-06
Persist name records for Peer Discovery

fabiotdf:
Related issue:
https://github.com/category-labs/category-internal/issues/1659

### DIFF
--- a/monad-peer-discovery/src/discovery.rs
+++ b/monad-peer-discovery/src/discovery.rs
@@ -30,6 +30,9 @@ use rand::{RngCore, seq::IteratorRandom};
 use rand_chacha::ChaCha8Rng;
 use tracing::{debug, info, trace, warn};
 
+use monad_types::{deserialize_pubkey, serialize_pubkey};
+use serde::{Deserialize, Serialize};
+
 use crate::{
     MonadNameRecord, NameRecord, PeerDiscoveryAlgo, PeerDiscoveryAlgoBuilder, PeerDiscoveryCommand,
     PeerDiscoveryEvent, PeerDiscoveryMessage, PeerDiscoveryMetricsCommand,
@@ -191,6 +194,22 @@ pub struct PeerDiscoveryBuilder<ST: CertificateSignatureRecoverable> {
     pub persisted_peers_path: PathBuf,
 }
 
+#[derive(Debug, Deserialize, Serialize, Clone)]
+#[serde(deny_unknown_fields)]
+#[serde(bound = "ST: CertificateSignatureRecoverable")]
+pub struct PersistedPeers<ST: CertificateSignatureRecoverable> {
+    pub address: String,
+
+    pub record_seq_num: u64,
+
+    #[serde(serialize_with = "serialize_pubkey::<_, CertificateSignaturePubKey<ST>>")]
+    #[serde(deserialize_with = "deserialize_pubkey::<_, CertificateSignaturePubKey<ST>>")]
+    pub secp256k1_pubkey: CertificateSignaturePubKey<ST>,
+
+    #[serde(bound = "ST: CertificateSignatureRecoverable")]
+    pub name_record_sig: ST,
+}
+
 impl<ST: CertificateSignatureRecoverable> PeerDiscoveryAlgoBuilder for PeerDiscoveryBuilder<ST> {
     type PeerDiscoveryAlgoType = PeerDiscovery<ST>;
 
@@ -272,42 +291,47 @@ impl<ST: CertificateSignatureRecoverable> PeerDiscoveryAlgoBuilder for PeerDisco
                     cmds.extend(cmds_from_insert);
                 }
             });
-
-        // Attempt to load persisted peers from JSON file and restore into pending queue
+/*
         match std::fs::read_to_string(&state.persisted_peers_path) {
             Ok(contents) => {
-                match toml::from_str::<
-                    BTreeMap<NodeId<CertificateSignaturePubKey<ST>>, MonadNameRecord<ST>>,
-                >(&contents)
-                {
+                match toml::from_str::<Vec<PersistedPeers<ST>>>(&contents) {
                     Ok(loaded) => {
                         debug!(path =? state.persisted_peers_path, loaded_len = loaded.len(), "loaded persisted peers");
-                        for (peer_id, name_record) in loaded.into_iter() {
-                            if peer_id == state.self_id {
-                                continue;
-                            }
+                        for persisted in loaded.into_iter() {
+                            match persisted.address.parse::<SocketAddrV4>() {
+                                Ok(addr) => {
+                                    let peer_id = NodeId::new(persisted.secp256k1_pubkey);
+                                    if peer_id == state.self_id {
+                                        continue;
+                                    }
 
-                            match name_record.recover_pubkey() {
-                                Ok(recovered) if recovered == peer_id => {}
-                                _ => {
-                                    warn!(
-                                        ?peer_id,
-                                        "skipping persisted peer with invalid signature"
-                                    );
-                                    continue;
-                                }
-                            }
+                                    let name_record = MonadNameRecord {
+                                        name_record: NameRecord::new(*addr.ip(), addr.port(), persisted.record_seq_num),
+                                        signature: persisted.name_record_sig,
+                                    };
 
-                            match state.insert_peer_to_pending(peer_id, name_record) {
-                                Ok(cmds_from_insert) => {
-                                    cmds.extend(cmds_from_insert);
-                                    debug!(?peer_id, "inserted persisted peer");
+                                    // verify signature matches stored pubkey
+                                    match name_record.recover_pubkey() {
+                                        Ok(recovered) if recovered == peer_id => {
+                                            match state.insert_peer_to_pending(peer_id, name_record) {
+                                                Ok(cmds_from_insert) => {
+                                                    cmds.extend(cmds_from_insert);
+                                                    debug!(?peer_id, "inserted persisted peer");
+                                                }
+                                                Err(err) => {
+                                                    warn!(?err, ?peer_id, "skipping persisted peer due to IP/validation error");
+                                                    continue;
+                                                }
+                                            }
+                                        }
+                                        _ => {
+                                            warn!(?peer_id, "skipping persisted peer with invalid signature");
+                                            continue;
+                                        }
+                                    }
                                 }
                                 Err(err) => {
-                                    warn!(
-                                        ?err,
-                                        "skipping persisted peer due to IP/validation error"
-                                    );
+                                    warn!(path =? state.persisted_peers_path, ?err, "invalid persisted peer address, skipping");
                                     continue;
                                 }
                             }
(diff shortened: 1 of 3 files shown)


monad-peer-discovery/src/lib.rs
```diff
@@ -656,6 +665,8 @@ pub trait PeerDiscoveryAlgo {
         NodeId<CertificateSignaturePubKey<Self::SignatureType>>,
         MonadNameRecord<Self::SignatureType>,
     >;
+
+    fn persist_peers(&self);
```
eerkaijun: i think this function shouldn't go into the PeerDiscoveryAlgo trait. it will just be a function under impl PeerDiscovery

monad-peer-discovery/src/mock.rs
```diff
@@ -279,4 +279,6 @@ where
     ) -> HashMap<NodeId<CertificateSignaturePubKey<ST>>, MonadNameRecord<ST>> {
         HashMap::new()
     }
+
+    fn persist_peers(&self) {}
```
eerkaijun: remove from mock when we remove the function from PeerDiscoveryAlgo trait

monad-peer-discovery/src/discovery.rs
```diff
@@ -1302,6 +1374,27 @@ where
         cmds
     }
 
+    fn persist_peers(&self) {
+        // Persist both routing_info and pending_queue as PersistedPeers (TOML)
+        let persisted = PersistedPeers {
+            routing_info: self.routing_info.clone(),
+            pending_queue: self.pending_queue.clone(),
```
eerkaijun: we only need to persist the name records in `pending_queue`. the other information doesn't matter. I don't think we need to separate into two fields (`routing_info` and `pending_queue`). basically we just want to collect all the current known name records and put them in a file.

monad-peer-discovery/src/discovery.rs
```diff
@@ -269,6 +285,60 @@ impl<ST: CertificateSignatureRecoverable> PeerDiscoveryAlgoBuilder for PeerDisco
...
+                        ) {
+                            Ok(_) => {
+                                state.pending_queue.insert(peer_id, conn_info);
+                                debug!(?peer_id, "Loaded persisted pending peer");
+                            }
+                            Err(err) => {
+                                debug!(?peer_id, ?err, "Persisted pending peer rejected");
+                            }
+                        }
+                    }
```
eerkaijun: as per above comment, i think we don't need to separate `pending_queue`. all the name records will just go through the `insert_peer_to_pending` code path

monad-peer-discovery/src/discovery.rs
```diff
@@ -1302,6 +1374,27 @@ where
         cmds
     }
 
+    fn persist_peers(&self) {
+        // Persist both routing_info and pending_queue as PersistedPeers (TOML)
+        let persisted = PersistedPeers {
+            routing_info: self.routing_info.clone(),
+            pending_queue: self.pending_queue.clone(),
+        };
+        match toml::to_string(&persisted) {
```
eerkaijun: maybe we can even format to the same format as [[bootstrap.peers]] format, that would probably make it consistent

monad-peer-discovery/src/message.rs
```diff
@@ -114,7 +115,9 @@ impl<ST: CertificateSignatureRecoverable> Decodable for PeerDiscoveryMessage<ST>
     }
 }
 
-#[derive(Debug, Clone, PartialEq, RlpDecodable, RlpEncodable)]
+#[derive(Debug, Clone, PartialEq, RlpDecodable, RlpEncodable, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+#[serde(bound = "ST: CertificateSignatureRecoverable")]
```
eerkaijun: we probably don't need this if we don't keep the unneeded info in `pending_queue`

docker/devnet/monad/config/node.toml
```diff
@@ -45,6 +45,7 @@ unresponsive_prune_threshold = 5
 last_participation_prune_threshold = 5000
 min_num_peers = 0
 max_num_peers = 200
+persisted_peers_path = "/tmp/peers.toml"
```
eerkaijun: do we want this extra field in node.toml? feels like the path should be an additional cli argument for monad-node. perhaps it can be optional, i.e. a new node wouldn't have a peers.toml

fabiotdf: A new node wouldn't have a peers.toml at startup, but when its time to save the peers, a path is required

With optional, do you mean opt-out or opt-in? I'd say opt-out is more appropriate, as I don't see much reason for users to disable the feature, since these files are pretty small

eerkaijun: optional as in if there's no file at the path, we will just skip inserting these name records during startup (my bad, optional is not a good word here)

docker/single-node/nets/compose.yaml
```diff
@@ -74,6 +74,7 @@ services:
       --node-config /monad/config/node.toml
       --forkpoint-config /monad/config/forkpoint.genesis.toml
       --validators-path /monad/config/validators.toml
+      --persisted-peers-path /monad/peers.toml
```
eerkaijun: nit: /monad/config/peers.toml as the default path to keep all the config files under the config directory

monad-node/src/state.rs
```diff
@@ -72,6 +72,7 @@ pub struct NodeState {
     // should be kept as long as node is alive, tracing listener is stopped when handle is dropped
     #[allow(unused)]
     manytrace_agent: Option<agent::Agent>,
+    pub persisted_peers_path: PathBuf,
```
eerkaijun: nit: can we move this line to the same section as other paths cli arg? for e.g. line 68

monad-peer-discovery/src/discovery.rs
```diff
@@ -269,6 +273,58 @@ impl<ST: CertificateSignatureRecoverable> PeerDiscoveryAlgoBuilder for PeerDisco
                 }
             });
 
+        // Attempt to load persisted peers from JSON file and restore into pending queue
+        match std::fs::read_to_string(&state.persisted_peers_path) {
+            Ok(contents) => {
+                match toml::from_str::<
+                    BTreeMap<NodeId<CertificateSignaturePubKey<ST>>, MonadNameRecord<ST>>,
+                >(&contents)
```
eerkaijun: As discussed, is it possible to reuse the struct `NodeBootstrapPeerConfig` when writing and reading from peers.toml file? so that these name records are recorded in the same format as node.toml

eerkaijun: I think it should be pretty straightforward to parse `MonadNameRecord` into `NodeBootstrapPeerConfig`, but let me know if that's not the case

monad-peer-discovery/src/discovery.rs
```diff
@@ -269,6 +273,58 @@ impl<ST: CertificateSignatureRecoverable> PeerDiscoveryAlgoBuilder for PeerDisco
...
+                {
+                    Ok(loaded) => {
+                        debug!(path =? state.persisted_peers_path, loaded_len = loaded.len(), "loaded persisted peers");
+                        for (peer_id, name_record) in loaded.into_iter() {
+                            if peer_id == state.self_id {
+                                continue;
+                            }
+
+                            match name_record.recover_pubkey() {
```
eerkaijun: if we use `NodeBootstrapPeerConfig` to store the entries in peers.toml, we do not need to do recover_pubkey here

monad-peer-discovery/src/discovery.rs
```diff
@@ -608,6 +664,26 @@ impl<ST: CertificateSignatureRecoverable> PeerDiscovery<ST> {
     fn is_pinned_node(&self, peer_id: &NodeId<CertificateSignaturePubKey<ST>>) -> bool {
         self.check_validator_membership(peer_id) || self.pinned_full_nodes.contains(peer_id)
     }
+
+    fn persist_peers(&self) {
+        // Persist both routing_info and pending_queue to TOML file
+        let mut routing_info = self.routing_info.clone();
+        for (peer_id, conn_info) in self.pending_queue.iter() {
+            routing_info.insert(*peer_id, conn_info.name_record.clone());
+        }
```
eerkaijun: we can parse these name records into `NodeBootstrapPeerConfig` here

monad-peer-discovery/src/lib.rs
```diff
@@ -47,7 +48,8 @@ pub enum PortTag {
     UDP = 1,
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, RlpEncodable, RlpDecodable)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, RlpEncodable, RlpDecodable, Deserialize, Serialize)]
```
eerkaijun, line 51: https://github.com/category-labs/monad-bft/pull/2529/files#r2547199504
we wouldn't need `Deserialize` and `Serialize` if we do the above

monad-peer-discovery/src/lib.rs
```diff
@@ -70,7 +72,8 @@ impl Port {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
```
eerkaijun, line 75: same as above

monad-peer-discovery/src/lib.rs
```diff
@@ -112,7 +115,8 @@ impl WireNameRecordV1 {
...
```
eerkaijun, line 118: same as above

monad-node/src/state.rs
fabiotdf: done

monad-peer-discovery/src/discovery.rs
```diff
@@ -186,6 +191,23 @@ pub struct PeerDiscoveryBuilder<ST: CertificateSignatureRecoverable> {
...
     pub enable_client: bool,
     pub rng: ChaCha8Rng,
+    pub persisted_peers_path: PathBuf,
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone)]
+#[serde(deny_unknown_fields)]
+#[serde(bound = "ST: CertificateSignatureRecoverable")]
+pub struct PersistedPeers<ST: CertificateSignatureRecoverable> {
+    pub address: String,
```
dshulyak, line 201: please don't introduce another type that is not future proof.

there is v1 and v2 name record already in the codebase, i think new code should support both of them

dshulyak (changes requested):

monad-peer-discovery/src/discovery.rs
```diff
@@ -608,6 +738,44 @@ impl<ST: CertificateSignatureRecoverable> PeerDiscovery<ST> {
...
+
+        // Persist both routing_info and pending_queue to TOML file
+        let mut persisted: Vec<PersistedPeers<ST>> = self
+            .routing_info
+            .iter()
+            .map(|(peer_id, name_record)| PersistedPeers {
+                address: name_record.udp_address().to_string(),
+                record_seq_num: name_record.seq(),
+                secp256k1_pubkey: peer_id.pubkey(),
+                name_record_sig: name_record.signature.clone(),
```
dshulyak, line 752: this is not going to work for v2 record, also it is better to introduce From trait for casts

monad-peer-discovery/src/discovery.rs
fabiotdf: I tried to reuse that struct for serialization but I can't get it to work.
I get `"error":"Error { inner: UnsupportedType(None) }"` from `toml::to_string()`

This is an example of a logged `persisted` struct:

```
{"message":"persist_peers aggregated routing_info / pending_queue","persisted":"[PersistedPeers { address: \"10.50.37.52:8888\", record_seq_num: 0, secp256k1_pubkey: 02537dd8d905d0e72467fef17fbd1ef5505b1e50993d2be41d3c726d11f750c710, name_record_sig: SecpSignature(RecoverableSignature(574fb64ff66e3d661e303e35618a31a3209b69bfbb93ef13a9ac68eb87cd328a67f2356f28db6ab8745196b2cbc639141d3a72176f66837f533e6a55d1cd2dd901)) }, PersistedPeers { address: \"10.50.37.50:8888\", record_seq_num: 0, secp256k1_pubkey: 0272ac793d36a71739e7080ee5e62029b665aa19a373e7efbc5d28660ad944ab4e, name_record_sig: SecpSignature(RecoverableSignature(f74bd97302961f7c9828963bd5a3394d88d1ac5136004c52c094459537aedf0f2363ad660ca0df738dc77c335e35a519c1049d642fd13cf9e25da393d1ca1fd200)) }]"},"target":"monad_peer_discovery::discovery"}
```

Attempted method:

```
    fn persist_peers(&self) {

        // Persist both routing_info and pending_queue to TOML file
        let mut persisted: Vec<PersistedPeers<ST>> = self
            .routing_info
            .iter()
            .map(|(peer_id, name_record)| PersistedPeers {
                address: name_record.udp_address().to_string(),
                record_seq_num: name_record.seq(),
                secp256k1_pubkey: peer_id.pubkey(),
                name_record_sig: name_record.signature.clone(),
            })
            .collect();
        for (peer_id, conn_info) in self.pending_queue.iter() {
            persisted.push( PersistedPeers {
                address: conn_info.name_record.udp_address().to_string(),
                record_seq_num: conn_info.name_record.seq(),
                secp256k1_pubkey: peer_id.pubkey(),
                name_record_sig: conn_info.name_record.signature.clone(),
            });
        }

        match toml::to_string(&persisted) {
            Ok(serialized) => {
                if let Err(error) = std::fs::write(&self.persisted_peers_path, serialized) {
                    warn!(path =? self.persisted_peers_path, ?error, "failed to persist peers");
                } else {
                    debug!(path =? self.persisted_peers_path, persisted_len = persisted.len(), "persisted peers to disk");
                }
            }
            Err(error) => {
                warn!(path =? self.persisted_peers_path, ?error, "failed to serialize persisted peers");
            }
        }
```

PR #2534, opened 2025-11-07
Add threshold for using remote configs

bamrith:


monad-node/src/state.rs
```diff
@@ -351,16 +356,18 @@ fn get_latest_configs(
...
                             ?remote_forkpoint_round,
                             ?local_forkpoint_round,
-                            "local forkpoint newer than remote forkpoint, using local configs"
+                            "local forkpoint within {} rounds of remote forkpoint, using local configs",
+                            remote_configs_threshold.0
                         );
                     }
 
-                    if remote_forkpoint_round < local_forkpoint_round - Round(200) {
+                    if remote_forkpoint_round < local_forkpoint_round - remote_configs_threshold {
```
omegablitz: nit: switch the order so it can't underflow

nit because it'll only make a difference near genesis, but could be possibly hit on test networks

omegablitz (approved):

omegablitz (approved):

PR #2538, opened 2025-11-09
monad-peer-discovery: changes related to authentication

dshulyak:
- new_with_authentication initializer for MonadNameRecord
- optional auth udp port for PeerEntry and added TryFrom<&PeerEntry<ST>> for MonadNameRecord<ST> for less error prone conversion
- name_records to NoopDiscovery for additional testing in raptorcast and get_name_record on the trait

### DIFF
--- a/monad-executor-glue/src/lib.rs
+++ b/monad-executor-glue/src/lib.rs
@@ -269,17 +269,31 @@ pub struct PeerEntry<ST: CertificateSignatureRecoverable> {
 
     pub signature: ST,
     pub record_seq_num: u64,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub auth_port: Option<u16>,
 }
 
 impl<ST: CertificateSignatureRecoverable> Encodable for PeerEntry<ST> {
     fn encode(&self, out: &mut dyn alloy_rlp::BufMut) {
-        let enc: [&dyn Encodable; 4] = [
-            &self.pubkey,
-            &self.addr.to_string(),
-            &self.signature,
-            &self.record_seq_num,
-        ];
-        encode_list::<_, dyn Encodable>(&enc, out);
+        if let Some(auth_port) = self.auth_port {
+            let enc: [&dyn Encodable; 5] = [
+                &self.pubkey,
+                &self.addr.to_string(),
+                &self.signature,
+                &self.record_seq_num,
+                &auth_port,
+            ];
+            encode_list::<_, dyn Encodable>(&enc, out);
+        } else {
+            let enc: [&dyn Encodable; 4] = [
+                &self.pubkey,
+                &self.addr.to_string(),
+                &self.signature,
+                &self.record_seq_num,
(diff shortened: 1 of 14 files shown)


michael-yxchen (approved):

monad-executor-glue/src/lib.rs
```diff
@@ -295,11 +309,18 @@ impl<ST: CertificateSignatureRecoverable> Decodable for PeerEntry<ST> {
         let signature = ST::decode(&mut payload)?;
         let record_seq_num = u64::decode(&mut payload)?;
 
+        let auth_port = if !payload.is_empty() {
+            Some(u16::decode(&mut payload)?)
+        } else {
+            None
+        };
```
eerkaijun, line 316: thinking about the backwards compatibility part - nodes that have not upgraded will not be able to decode the auth port of the upgraded nodes?

dshulyak: this is not sent over the network but will be in the local config, is it correct?

so in this case they will have to update to the version with auth if they want to use peer entry with auth udp port in their config. they can continue using non-auth peer entry as long as the other node didn't change the ip address together with enabling auth

alternatively we can do one more intermediate release that will rollout this code to the whole network before introducing config changes

dshulyak: actually yes it is better to have one more intermediate release, then we we will guarantee that casts always correct between this two structs

i updated couple of nodes, and there is no need for old nodes decode to udp port from peer entry as long they can continue to use previous name record. one case where they actually need to *encode* is if they try to export records with

> monad-debug-node -c monad-bft/controlpanel.sock get-peers

```
[[peers]]
address = "188.214.131.5:8000"
record_seq_num = 1
secp256k1_pubkey = "0x0342f3e447bd6951b2cf7cd717958a84e58dbe8cfbe2780f0247c69591216d2d7c"
name_record_sig = "0x215e845dade986e48f6fcb1f065f3ff993ddf0d38b0ef41a05f6621d71bf4c6c2d25747629ac8a3d583418857bc0b4c75140202e14919faf643f431ade4b944d00"
```

note that udp port is not exported because they don't know about it, with intermediate release they will be able to export it with auth port

```
[[peers]]
address = "188.214.131.5:8000"
record_seq_num = 1
secp256k1_pubkey = "0x0342f3e447bd6951b2cf7cd717958a84e58dbe8cfbe2780f0247c69591216d2d7c"
name_record_sig = "0x215e845dade986e48f6fcb1f065f3ff993ddf0d38b0ef41a05f6621d71bf4c6c2d25747629ac8a3d583418857bc0b4c75140202e14919faf643f431ade4b944d00"
auth_port = 8001
```

eerkaijun (approved):

monad-executor-glue/src/lib.rs
eerkaijun: probably a rare use case so I think it's fine

PR #2542, opened 2025-11-10
Feat: EIP-7966

haythemsellami:
This PR introduce the `eth_sendRawTransactionSync` RPC endpoint.
For more details: https://eips.ethereum.org/EIPS/eip-7966

Closes https://github.com/category-labs/category-external/issues/119


monad-rpc/src/handlers/eth/txn.rs
```diff
@@ -230,45 +231,135 @@ pub async fn monad_eth_sendRawTransaction(
...
+
+    return Err(JsonRpcError::custom("txpool not responding".to_string()));
+
+}
+
+#[derive(Deserialize, Debug, schemars::JsonSchema)]
+pub struct MonadEthSendRawTransactionSyncParams {
+    hex_tx: UnformattedData,
+    #[serde(rename = "timeoutMs")]
+    timeout_ms: Option<u64>,
```
philipglazman, line 303: Why not just use `#[serde(default = 2_000)]`?

monad-rpc/src/handlers/eth/txn.rs
```diff
@@ -230,45 +231,135 @@ pub async fn monad_eth_sendRawTransaction(
...
+    chain_state: &ChainState<T>,
+    params: MonadEthSendRawTransactionSyncParams,
+    chain_id: u64,
+    allow_unprotected_txs: bool,
+) -> JsonRpcResult<MonadTransactionReceipt> {
+    trace!("monad_eth_sendRawTransactionSync: {params:?}");
+
+    let timeout_ms = params.timeout_ms
+    .filter(|&t| t > 0 && t <= RECOMMENDED_TIMEOUT_MS)
+    .unwrap_or(RECOMMENDED_TIMEOUT_MS);
```
philipglazman: We probably want to allow operators to set the maximum timeout allowed. See `monad-rpc/src/cli.rs`

monad-rpc/src/handlers/eth/txn.rs
haythemsellami: Initially I had that, but I thought the EIP implying that node operator can configure the default timeout and max timeout, so planning to have this one configurable also through the cli. Or is it only a max time that node operator set and default one is 2s ?

philipglazman: Right, both should be configurable through cli. We can set the default 2_000 in cli.

monad-rpc/src/handlers/eth/txn.rs
```diff
@@ -230,45 +235,132 @@ pub async fn monad_eth_sendRawTransaction(
...
+    eth_send_raw_transaction_sync_max_timeout_ms: u64,
+) -> JsonRpcResult<MonadTransactionReceipt> {
+    trace!("monad_eth_sendRawTransactionSync: {params:?}");
+
+    let timeout_ms = params
+        .timeout_ms
+        .filter(|&t| t > 0 && t <= eth_send_raw_transaction_sync_max_timeout_ms)
+        .unwrap_or(eth_send_raw_transaction_sync_default_timeout_ms);
+
+    let tx = validate_and_decode_tx(&params.hex_tx.0, chain_id, allow_unprotected_txs)?;
```
philipglazman: return error code `5` ?

haythemsellami: Just added it by checking `validate_and_decode_tx` returned error, if it is the `txn_decode_error` then return the async error code 5.

monad-rpc/src/handlers/eth/txn.rs
```diff
@@ -226,48 +234,141 @@ pub async fn monad_eth_sendRawTransaction(
...
+    allow_unprotected_txs: bool,
+    eth_send_raw_transaction_sync_default_timeout_ms: u64,
+    eth_send_raw_transaction_sync_max_timeout_ms: u64,
+) -> JsonRpcResult<MonadTransactionReceipt> {
+    trace!("monad_eth_sendRawTransactionSync: {params:?}");
+
+    let timeout_ms = params
+        .timeout_ms
+        .filter(|&t| t > 0 && t <= eth_send_raw_transaction_sync_max_timeout_ms)
+        .unwrap_or(eth_send_raw_transaction_sync_default_timeout_ms);
```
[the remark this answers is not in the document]
haythemsellami: Per the EIP, on invalid timeout, just fallback to default one.

monad-rpc/src/handlers/eth/txn.rs
```diff
@@ -226,48 +234,141 @@ pub async fn monad_eth_sendRawTransaction(
...
+
+    let tx =
+        validate_and_decode_tx(&params.hex_tx.0, chain_id, allow_unprotected_txs).map_err(|e| {
+            match e.code {
+                -32603 => JsonRpcError::tx_sync_unready(),
+                _ => e,
```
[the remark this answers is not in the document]
haythemsellami: Hmm, `validate_and_decode_tx` will only return that error code in case of decoding failure or when unprotected tx is rejected, so I guess no need to look by error message.

monad-rpc/src/handlers/eth/txn.rs
```diff
@@ -226,48 +234,141 @@ pub async fn monad_eth_sendRawTransaction(
...
+            }
+        })?;
+
+    let tx_hash = *tx.tx_hash();
+    debug!(name = "sendRawTransactionSync", txn_hash = ?tx_hash);
+    submit_to_txpool(txpool_bridge_client, tx).await?;
```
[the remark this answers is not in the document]
haythemsellami:
```
On standard error. Node implementations MUST return a JSON-RPC error object consistent with existing RPC error formats.
```

monad-rpc/src/handlers/eth/txn.rs
```diff
@@ -226,48 +234,141 @@ pub async fn monad_eth_sendRawTransaction(
...
+        .filter(|&t| t > 0 && t <= eth_send_raw_transaction_sync_max_timeout_ms)
+        .unwrap_or(eth_send_raw_transaction_sync_default_timeout_ms);
+
+    let tx =
+        validate_and_decode_tx(&params.hex_tx.0, chain_id, allow_unprotected_txs).map_err(|e| {
+            match e.code {
+                -32603 => JsonRpcError::tx_sync_unready(),
+                _ => e,
+            }
+        })?;
```
andr-dev: Instead of matching on the error variant, we can pass a lambda to the `validate_and_decode_tx(...)` method that is used to generate the error for this case. This gives us the flexibility of using the existing error for `eth_sendRawTransaction` while supporting this new error for `eth_sendRawTransactionSync`.

haythemsellami: Updated.

monad-rpc/src/handlers/eth/txn.rs
```diff
@@ -226,48 +234,141 @@ pub async fn monad_eth_sendRawTransaction(
...
+                tokio::time::sleep(poll_interval).await;
+            }
+            Err(ChainStateError::Archive(e)) => {
+                return Err(JsonRpcError::internal_error(format!("Archive error: {e}")));
+            }
+            Err(ChainStateError::Triedb(e)) => {
+                return Err(JsonRpcError::internal_error(format!("Triedb error: {e}")));
+            }
         }
     }
```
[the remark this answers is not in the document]
haythemsellami: I think for now the polling mechanism is fine to use, maybe later on we switch to an poll-less events listening implementation.

monad-rpc/src/cli.rs
```diff
@@ -136,6 +136,14 @@ pub struct Cli {
     #[arg(long, default_value_t = 30_000_000)]
     pub eth_estimate_gas_provider_gas_limit: u64,
 
+    /// Set the default timeout (in milliseconds) for eth_sendRawTransactionSync
+    #[arg(long, default_value_t = 2_000)]
+    pub eth_send_raw_transaction_sync_default_timeout_ms: u64,
+
+    /// Set the maximum timeout (in milliseconds) for eth_sendRawTransactionSync
```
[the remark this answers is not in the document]
haythemsellami: This is all documented in the EIP url.

PR #2550, opened 2025-11-12
Add more logging for block validation errors

bamrith:
This PR only includes better logging for `EthBlockValidator`. Changes to `EthBlockPolicy` will be in a separate PR.


monad-eth-block-policy/src/validation.rs
```diff
@@ -61,16 +76,25 @@ impl TfmValidator {
...
-    ) -> Result<(), TransactionError> {
+    ) -> Result<(), StaticValidationError> {
         if execution_chain_params.tfm_enabled {
             // tfm-specific gating
             if tx.gas_limit() > TFM_MAX_GAS_LIMIT {
-                return Err(TransactionError::GasLimitTooHigh);
+                debug!(
+                    tx_gas_limit = tx.gas_limit(),
+                    TFM_MAX_GAS_LIMIT, "Tx gas limit over tfm max gas limit"
+                );
```
eerkaijun: since txpool also uses `static_validate_transaction`, will these debug logs be too spammy? i.e. any invalid transactions in the txpool also get debug logged. I'm thinking just logging the error in monad-eth-block-validator is sufficient, or is there some additional info that are provided by these logs?

monad-eth-block-validator/src/lib.rs
```diff
@@ -366,11 +434,11 @@ where
...
         // early return if any user transaction fails static validation
         for eth_txn in eth_txns.iter() {
-            if let Err(e) =
+            if let Err(txn_err) =
                 static_validate_transaction(eth_txn, chain_id, chain_params, execution_chain_params)
             {
-                tracing::debug!("block body static validation failed: {:?}", e);
-                return Err(BlockValidationError::TxnError);
+                debug!(?eth_txn, ?txn_err, "transaction static validation failed");
+                return Err(TxnError::StaticValidationError(txn_err).into());
```
eerkaijun, line 441: on above comment, this seems sufficient for me without needing the debug logs in static_validate_transaction

michael-yxchen: i agree that we should embed the offending value in the error type. the caller can decide how to handle/log the error

monad-eth-block-policy/src/validation.rs
bamrith: that's a good point. Ariq suggested we add logs here too so that it is clear what the expected and actual value was in case of errors.

bamrith: removed the debug logs and made the returned enums more descriptive

monad-eth-block-validator/src/lib.rs
bamrith: resolved

PR #2552, opened 2025-11-13
Keystore: clone IKM if printing keys

bamrith:


monad-keystore/src/keystore.rs
```diff
@@ -130,6 +130,7 @@ impl KeystoreSecret {
         self.0.is_empty()
     }
 
+    /// zeroizes the keystore secret
     pub fn to_bls(&mut self, version: KeystoreVersion) -> Result<BlsKeyPair, KeystoreError> {
```
dshulyak, line 134: i would remove manual zeroization and &mut from those methods, secret is zeroized on drop, i can do that in separate commit if you prefer

dshulyak: or instead take self to force a clone on the secret

bamrith:
https://github.com/category-labs/monad-bft/pull/2553

PR #2557, opened 2025-11-13
Add support for generic txpool sidecars

andr-dev:
This set of changes adds the following txpool features:
1. Allow txpool IPC clients to specify per-tx priorities, txpool then ingests txs based on these priorities
2. Order transactions by said priority when proposing, enabling sidecars to place transactions at the top of the block
3. Send transactions over IPC after insertion, enabling sidecars to create a view of the pool

These three features enable generic sidecars to be implemented without having to modify the Category Labs `monad-bft` client.

### DIFF
--- a/monad-eth-txpool/src/pool/transaction.rs
+++ b/monad-eth-txpool/src/pool/transaction.rs
@@ -261,6 +261,16 @@ impl ValidEthTransaction {
         }
     }
 
+    pub fn is_owned_and_forwardable(&self) -> bool {
+        match &self.kind {
+            PoolTransactionKind::Owned {
+                priority,
+                extra_data: _,
+            } => priority <= &DEFAULT_TX_PRIORITY,
+            PoolTransactionKind::Forwarded => false,
+        }
+    }
+
     pub fn has_higher_priority(&self, other: &Self, _base_fee: u64) -> bool {
         if self.tx_kind_priority() != other.tx_kind_priority() {
             // If either tx has a custom tx kind priority, the pool will ignore all other tx related
@@ -301,7 +311,7 @@ impl ValidEthTransaction {
         last_commit_seq_num: SeqNum,
         last_commit_base_fee: u64,
     ) -> Option<&TxEnvelope> {
-        if !self.is_owned() {
+        if !self.is_owned_and_forwardable() {
             return None;
         }
 
--- a/monad-eth-txpool-executor/src/lib.rs
+++ b/monad-eth-txpool-executor/src/lib.rs
@@ -557,7 +557,7 @@ where
             };
 
             let mut inserted_addresses = HashSet::<Address>::default();
-            let mut inserted_txs = Vec::default();
+            let mut immediately_forwardable_txs = Vec::default();
 
             pool.insert_txs(
                 &mut EthTxPoolEventTracker::new(&metrics.pool, &mut ipc_events),
@@ -567,7 +567,10 @@ where
                 recovered_txs,
                 |tx| {
                     inserted_addresses.insert(tx.signer());
-                    inserted_txs.push(tx.raw().tx().clone());
+
+                    if tx.is_owned_and_forwardable() {
+                        immediately_forwardable_txs.push(tx.raw().tx().clone());
+                    }
                 },
             );
 
@@ -576,7 +579,7 @@ where
             forwarding_manager
                 .as_mut()
                 .project()
-                .add_egress_txs(inserted_txs.iter());
+                .add_egress_txs(immediately_forwardable_txs.iter());
 
             metrics.update(executor_metrics);
             ipc.as_mut().broadcast_tx_events(ipc_events);


dshulyak:
maybe it was already considered, but just wanted to share what i would prefer to work with if i had to implement mev searcher for monad. instead of learning extension points and finding creative ways of using them, i would rather take building blocks (e.g libraries) and make an executable that interacts with consensus and execution in an expected and performance aware manner.

it would mean that default pool implementation is completely swapable, and i may build on top of that or i may not depending on what exactly i want to achieve, so e.g libraries and reusable components. beside that i would like to have defined communication protocol with rpc (e.g tx format) and consensus (events/notifications format, proposal format). i think existing implementation already sortof works this way (in terms of defined protocol on both sides), but it is not swapable and protocol itself is not formalized.

ariqchowdhury:
> maybe it was already considered, but just wanted to share what i would prefer to work with if i had to implement mev searcher for monad. instead of learning extension points and finding creative ways of using them, i would rather take building blocks (e.g libraries) and make an executable that interacts with consensus and execution in an expected and performance aware manner.
> 
> it would mean that default pool implementation is completely swapable, and i may build on top of that or i may not depending on what exactly i want to achieve, so e.g libraries and reusable components. beside that i would like to have defined communication protocol with rpc (e.g tx format) and consensus (events/notifications format, proposal format). i think existing implementation already sortof works this way (in terms of defined protocol on both sides), but it is not swapable and protocol itself is not formalized.

yeah, longer term this is the goal. Proposal creation and the txpool sorting/eviction etc will be swappable components. For the short term we want to do this so that some basic support exists

andr-dev:
> Note: Need to add `owned` flag to IPC tx message to prevent sidecar txs from being forwarded if desired.

This has been implemented by skipping forwarding for transactions sent through IPC when their priority is higher than RPC's default priority (0xFFFFF).

monad-eth-txpool-ipc/src/message.rs
```diff
@@ -0,0 +1,38 @@
...
+
+use alloy_consensus::TxEnvelope;
+use alloy_primitives::U256;
+use alloy_rlp::{RlpDecodable, RlpEncodable};
+use monad_eth_txpool_types::DEFAULT_TX_PRIORITY;
+
+#[derive(RlpEncodable, RlpDecodable)]
+pub struct EthTxPoolIpcTx {
+    pub tx: TxEnvelope,
+    pub priority: U256,
```
michael-yxchen, line 24: just curious why do we need an U256 type for priority?

monad-eth-txpool/src/pool/transaction.rs
```diff
@@ -253,8 +288,9 @@ impl ValidEthTransaction {
         //     to forward the transaction, then there is little cost to the network in allowing
         //     transactions with small fee bumps.
 
-        self.tx.max_fee_per_gas() > other.tx.max_fee_per_gas()
-            && self.tx.max_priority_fee_per_gas() >= other.tx.max_priority_fee_per_gas()
+        self.priority() > other.priority()
+            || (self.tx.max_fee_per_gas() > other.tx.max_fee_per_gas()
```
michael-yxchen: I have the same question as copilot here, as the `has_higher_priority` behavior is different from `OrderedTx.cmp`. are there docs explaining the role of priority?

monad-eth-txpool-types/src/lib.rs
```diff
@@ -28,7 +36,13 @@ pub struct EthTxPoolEvent {
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub enum EthTxPoolEventType {
     /// The tx was inserted into the txpool.
-    Insert { address: Address, owned: bool },
+    Insert {
+        address: Address,
+        owned: bool,
+
+        #[serde(with = "eth_tx_pool_event_type_tx_serde")]
+        tx: TxEnvelope,
```
michael-yxchen, line 44: why do we need the full tx in this event type? the event appears to be only used by the tracker

monad-eth-txpool-ipc/src/message.rs
ariqchowdhury: external team request

monad-eth-txpool-types/src/lib.rs
andr-dev: External team request to see all txs being added to the pool. Previously, only tx hashes would be sent back from the pool over IPC since there was only one "sidecar" (RPC) so it already knew which txs it was sending. Now that there will be multiple sidecars (RPC + other teams), we need to send full txs so every sidecar can see every tx.

monad-eth-txpool/src/pool/transaction.rs
ariqchowdhury: can we add a comment here to clarify

monad-eth-txpool-ipc/src/message.rs
andr-dev: Priority bids are likely to be expressed in native token amounts which for Monad mainnet is 18 decimals. Using a u64 would require sidecars to compress priorities derived from bid amounts into the smaller range, adding complexity to the sidecar.

PR #2559, opened 2025-11-14
dataplane: print warnings if dataplane is overrun

dshulyak:
sets total priority queue capacity to 100MB per priority queue, so 200MB in total max. 
if capacity is exceeded for that priority, messages will be dropped with a warning logged.


omegablitz:
Is there a benefit to making the byte limits be by priority bucket?

dshulyak:
updated to be per priority

PR #2560, opened 2025-11-14
Add egress bytes ratelimit to txpool forwarding manager

andr-dev:


monad-eth-txpool-executor/src/forward.rs
```diff
@@ -71,6 +80,9 @@ impl Default for EthTxPoolForwardingManager {
 
             egress: VecDeque::default(),
             egress_waker: None,
+            egress_bytes_rate_limit: RateLimiter::direct(Quota::per_second(
+                EGRESS_RATE_LIMIT_BYTES_PER_SECOND,
```
dshulyak, line 81: maybe rate limiter with burst would be a better fit, better to deliver txs on time before block proposal rather than slowly feed them

PR #2565, opened 2025-11-14
Update execution

ariqchowdhury:
includes changes for removing sqpoll threads for statesync


andr-dev:
Resolved in https://github.com/category-labs/monad-bft/pull/2566.

PR #2567, opened 2025-11-14
Txgen: add nonce gaps and legacy tx opts; refactor batch transformations

JoeHowarth:
Reorganizes the existing random tx mutation logic into a centralized `transform.rs` module that is called in `gen_harness.rs` after the generator produces a batch.

Also adds new transformations:
- randomly dropping txs to create nonce gaps
- convert eip-1559 txs to Legacy txs

All transform opts are specified at the workload level

### DIFF
--- a/monad-eth-testutil/examples/txgen/cli.rs
+++ b/monad-eth-testutil/examples/txgen/cli.rs
@@ -174,6 +174,14 @@ pub struct CliConfig {
     #[arg(long, global = true)]
     pub random_priority_fee_range: Option<String>,
 
+    /// Percentage of transactions to drop after generation (0-100)
+    #[arg(long, global = true)]
+    pub drop_percentage: Option<f64>,
+
+    /// Percentage of EIP-1559 transactions to convert to legacy after mutation (0-100)
+    #[arg(long, global = true)]
+    pub convert_eip1559_to_legacy_percentage: Option<f64>,
+
     /// Override for native contract address
     #[arg(long, global = true)]
     pub native_contract: Option<String>,
@@ -337,12 +345,20 @@ impl From<CliConfig> for Config {
 
 impl From<CliConfig> for WorkloadGroup {
     fn from(value: CliConfig) -> Self {
-        WorkloadGroup {
+        let mut workload = WorkloadGroup {
             // Effectively infinite runtime
             runtime_minutes: 100_000_000_000.0,
-            traffic_gens: vec![value.into()],
+            traffic_gens: vec![value.clone().into()],
             ..Default::default()
+        };
+
+        if let Some(drop_percentage) = value.drop_percentage {
+            workload.drop_percentage = drop_percentage.clamp(0.0, 100.0);
         }
+        if let Some(convert_percentage) = value.convert_eip1559_to_legacy_percentage {
+            workload.convert_eip1559_to_legacy = convert_percentage.clamp(0.0, 100.0);
+        }
+        workload
     }
 }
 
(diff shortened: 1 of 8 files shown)


monad-eth-testutil/examples/txgen/cli.rs
```diff
@@ -163,6 +163,10 @@ pub struct CliConfig {
     #[arg(long, global = true)]
     pub random_priority_fee_range: Option<String>,
 
+    /// Percentage of transactions to drop after generation (0-100)
+    #[arg(long, global = true)]
+    pub drop_percentage: Option<u64>,
+
```
eclark0426, line 184: Is legacy conversion intentionally not being added to the CLI?

monad-eth-testutil/examples/txgen/workers/transform.rs
```diff
@@ -0,0 +1,335 @@
...
+    };
+
+    let sig = original_key.sign_transaction(&legacy);
+    Some((
+        TxEnvelope::Legacy(legacy.into_signed(sig)),
+        original_key.clone(),
+    ))
+}
+
+fn mutate_eip1559_transaction(tx: &TxEnvelope, original_key: &PrivateKey) -> TxEnvelope {
```
eclark0426, line 319: You've moved eip1559 mutation here, but left 7702 where it was. If the goal is to centralize all the mutation logic in `transform.rs` then 7702 mutation should probably also be moved here

monad-eth-testutil/examples/txgen/workers/transform.rs
```diff
@@ -0,0 +1,335 @@
...
+    prelude::*,
+    shared::{eip7702::mutate_eip7702_transaction, private_key::PrivateKey},
+};
+
+/// Batched transformations configured for a generator workload.
+#[derive(Debug, Clone)]
+pub struct TransformOptions {
+    pub mutation_percentage: f64,
+    pub drop_percentage: f64,
+    pub convert_eip1559_to_legacy: bool,
```
eclark0426: Would it make sense for legacy conversion to also be a float so we can convert only a certain percentage to legacy while leaving others unchanged?

monad-eth-testutil/examples/txgen/cli.rs
JoeHowarth: Yeah, idk about the cli growing ever larger... It's probably "correct" but feels bad

monad-eth-testutil/examples/txgen/workers/transform.rs
JoeHowarth: it's supposed to be in range [0,100] right? imo neither need to be floats, but I guess that gives more flexibility to acceptable inputs so no harm

eclark0426: Float, int, whatever format you want - the point was more that right now, conversion to legacy transactions is all or nothing. Mutation definitely should stay a float - when I was testing it, anything above about 0.02% (not 2%, but 0.02%) resulted in a large decrease in successful tps

JoeHowarth: ahh I think we're talking past each other - I was referring to drop % and you're talking about 1559 -> legacy, right?

Also ack regarding tps drop. Fwiw, when we test this we should increase --senders, --sender-group-size and lower --tx-per-sender so that nonce gaps are less of an issue, then increase mutation percent to something higher like 2%. That way we see the effect of mutations on tps rather than it being dominated by nonce gaps

JoeHowarth: When you/I update the config after this gets merged we should change those for the new workflow

michael-yxchen (approved):

PR #2571, opened 2025-11-15
Update logging for block policy errors

bamrith:


monad-eth-block-policy/src/lib.rs
```diff
@@ -528,20 +500,18 @@ where
         &self,
```
michael-yxchen: txpool proposal sequencer calls this function. i think we can put the logs on a lower level, but don't feel strongly

monad-eth-block-policy/src/lib.rs
```diff
@@ -528,20 +500,18 @@ where
...
         txn: &ValidatedTx,
-    ) -> Result<(), BlockPolicyError> {
+    ) -> Result<(), EthBlockPolicyBlockValidatorError> {
         let eth_address = txn.signer();
 
         let maybe_account_balance = account_balances.get_mut(&eth_address);
 
         let Some(account_balance) = maybe_account_balance else {
-            warn!(
+            debug!(
```
eerkaijun, line 509: i think this should be a warn - if this happens, it's likely there is a bug in our code (i.e. account balances for tx signers should always be populated)

PR #2572, opened 2025-11-17
monad-version: added crate for git version info and used in the selected binaries

dshulyak:
```
./target/debug/monad-keystore --version

monad-keystore {"commit":"0f2db414bbad61e5a3d8a4845e5e82c7c20bf3b8","tag":"test-again","branch":"dmitry/monad-node-version","modified":false}
```

i would remove binary name from the output, but this is done by clap and i can't overwrite that
and just using clap macro seems cleaner

fields (git command with env var fallback):
- commit: git rev-parse HEAD (or GIT_COMMIT env)
- tag: git describe --tags --exact-match HEAD (or GIT_TAG env)
- branch: git branch --show-current (or GIT_BRANCH env)
- modified: git status --porcelain (or GIT_MODIFIED env)

rebuild triggers if any of the following changes: .git/HEAD, .git/index, .git/refs

binaries updated: monad-node, monad-rpc, monad-keystore, wal2json, txgen

Edit: this PR closes https://github.com/category-labs/category-internal/issues/2052

### DIFF
--- a/monad-version/src/lib.rs
+++ b/monad-version/src/lib.rs
@@ -1,0 +1,133 @@
+// Copyright (C) 2025 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+use std::{
+    ffi::OsStr,
+    io,
+    path::Path,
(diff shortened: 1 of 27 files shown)


monad-version/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+use std::{
+    ffi::OsStr,
+    io,
+    path::Path,
+    process::{Command, Output},
+};
+
+fn git_command<D, I, S>(work_tree: D, args: I) -> io::Result<Output>
```
omegablitz, line 23: Will this work in Dockerfile if .git is dockerignored? If not, maybe [GIT_COMMIT_HASH](https://github.com/category-labs/monad-bft/blob/d379893c50d5f13647de9bcdc8f0f6a8976b7131/.github/workflows/publish.yaml#L101) can be used instead

dshulyak: it won't, i will update to compute in .github jobs and pass it using env variables

dshulyak:
tested that staging job builds correct versions {"commit":"9f1b443462a9180cc0e8b42c66ab0ef4e8c378d3","tag":"","branch":"staging-version-helper","modified":false}

PR #2574, opened 2025-11-17
Improve raptorcast group invite response checks

fabiotdf:
This adds a sanity check to the raptorcast group scheduler.
Previously, it was possible for the validator to process an accept for a group invite from a fullnode after already receiving a reject from same.

Related issue https://github.com/category-labs/category-internal/issues/2022

### DIFF
--- a/monad-raptorcast/src/raptorcast_secondary/publisher.rs
+++ b/monad-raptorcast/src/raptorcast_secondary/publisher.rs
@@ -377,12 +377,12 @@ where
     ST: CertificateSignatureRecoverable,
 {
     full_nodes_accepted: FullNodesST<ST>,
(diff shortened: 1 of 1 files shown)


ariqchowdhury:
does the full_nodes_rejected list need to be cleared at any point? Is it possible for a node to be rejected but eligible later?

ariqchowdhury (approved):

fabiotdf:
> does the full_nodes_rejected list need to be cleared at any point? Is it possible for a node to be rejected but eligible later?

Yes because the scheduler will send invites to other nodes until deadline or until the target group size has been reached. This is the current behavior and this PR doesn't change that. A well-behaving node won't ever send 2 responses to the same invite (which is what this PR sanity-checks against)

PR #2575, opened 2025-11-17
Update txpool insertion policy

andr-dev:
This change updates the txpool insertion balance check to be stricter, rejecting any tx that doesn't have sufficient balance from the last committed block to pay for the max cost of the transaction. This is the minimal change required to correctly mitigate https://github.com/category-labs/category-internal/issues/2039.

In the future, this can be relaxed by updating the eviction policy, although this will require a bit more work and thorough testing to ensure the DoS vector isn't re-introduced.

### DIFF
--- a/monad-eth-txpool/tests/pool.rs
+++ b/monad-eth-txpool/tests/pool.rs
@@ -915,8 +915,8 @@ fn test_insert_unknown_account() {
 
 #[test]
 #[traced_test]
-fn test_insert_low_balance() {
-    let tx1 = make_legacy_tx(S1, BASE_FEE, GAS_LIMIT, 0, 10);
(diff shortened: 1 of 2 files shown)


monad-eth-txpool/tests/pool.rs
```diff
@@ -945,11 +945,86 @@ fn test_insert_low_balance() {
...
+                expected_txs: vec![&tx],
+                add_to_blocktree: true,
+            },
+        ],
+    );
+}
+
+#[test]
+#[traced_test]
+fn test_insert_1559_low_balance() {
```
mikety, line 993: nit: eip1559 cases do not use priority fee which also included in computing gas fee.

mikety (approved):
lgtm with nit

PR #2576, opened 2025-11-17
monad-node-config: reuse default mtu from monad-dataplane

dshulyak:
this doesn't backward compatibility, as long as configured mtu is not higher than ethernet mtu

### DIFF
--- a/monad-node-config/src/network.rs
+++ b/monad-node-config/src/network.rs
@@ -15,6 +15,7 @@
 
 use std::net::Ipv4Addr;
 
+use monad_types::DEFAULT_MTU;
 use serde::Deserialize;
 
 #[derive(Debug, Clone, Deserialize)]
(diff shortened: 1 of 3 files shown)


monad-node-config/src/network.rs
```diff
@@ -48,10 +49,8 @@ pub struct NodeNetworkConfig {
     pub tcp_rate_limit_burst: u32,
 }
 
-// When running in docker with vpnkit, the maximum safe MTU is 1480, as per:
-// https://github.com/moby/vpnkit/tree/v0.5.0/src/hostnet/slirp.ml#L17-L18
 fn default_mtu() -> u16 {
-    1480
+    DEFAULT_MTU
 }
```
ariqchowdhury: does everything in flexnet still work after this?

omegablitz (approved):

PR #2584, opened 2025-11-18
Add cli argument for number of fibers in trace tx execution pool

maxkozlovsky:
Add cli argument to configure trace block transaction execution pool in executor. The pool shares threads with trace block execution and is used to parallelize transaction execution within a block.


maxkozlovsky:
also contains execution bump to the corresponding commit

ariqchowdhury (approved):

bamrith (approved):

PR #2588, opened 2025-11-19
Simplify RPC chainstate get logs

andr-dev:
- Improves legibility of `get_logs` method by flattening conditional branches in `stream_with_triedb`
- Removes unnecessary clones of all transactions and receipts (caused by `iter()` => `to_vec()`)

### Stressnet before/after [fullnode, eth_getLogs last block (10k TPS, 2.5 BPS, 4k tx per block)]:
image
image


JoeHowarth (approved):
Nice

PR #2597, opened 2025-11-21
Firewall scripts

fabiotdf:
A script to help creating firewall rules whitelisting monad nodes


michael-yxchen:
as discussed offline, can we make a separate script to fetch the IPs?

PR #2601, opened 2025-11-24
monad-raptorcast: add signature verification rate limit with authenticated peer bypass

dshulyak:
we apply rate limiting of 4000 new sig verifications with a bypass for validator, based on auth information.
on stressnet normal rate is ~200 sigs/s, so 4000 is significantly higher than what we can expect.

verified that with this limit in place i can't starve raptorcast thread using my spam scripts, cpu usage peak around 50% and i am unable to prevent a node from participating in consensus.

### DIFF
--- a/monad-raptorcast/src/udp.rs
+++ b/monad-raptorcast/src/udp.rs
@@ -16,6 +16,7 @@
 use std::{collections::BTreeMap, num::NonZero, ops::Range};
 
 use bytes::Bytes;
+use governor::{DefaultDirectRateLimiter, Quota, RateLimiter};
 use lru::LruCache;
 use monad_crypto::{
     certificate_signature::{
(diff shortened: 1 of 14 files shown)


monad-raptorcast/src/udp.rs
```diff
@@ -157,13 +170,36 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
...
+                    };
+                    if allowed || is_validator {
+                        Ok(())
+                    } else {
+                        Err(MessageValidationError::RateLimited)
+                    }
+                },
             ) {
                 Ok(message) => message,
+                Err(MessageValidationError::RateLimited) => {
```
dshulyak, line 198: keep debug here

xinyuan-dev (approved):

PR #2606, opened 2025-11-24
Archive: async backfill; only traces archive

JoeHowarth:
2 features related to backfilling traces quickly
1. `--traces-only` archiver flag allows skipping blocks and receipts and only uploading traces. This avoids clobbering correct blocks and receipts as well as speeding up the backfill process.
2. `--async-backfill` archiver and indexer flag allows a second process to backfill a given range while the first (main) process continues to process the chain tip. This is useful for RPC operators so that they do not have downtime

Copilot description arguably better:

> This PR adds two new features to the monad-archive system: asynchronous backfill capability and traces-only archiving mode. The async backfill feature allows a second archiver/indexer to backfill a historical range while the primary one continues processing new blocks, using separate tracking keys to avoid conflicts. The traces-only mode enables archiving block traces without storing blocks or receipts.
> 
> Key Changes
> Added async_backfill and traces_only configuration options with corresponding CLI flags
> Introduced new LatestKind enum variants (UploadedAsyncBackfill, IndexedAsyncBackfill) to track async backfill progress separately
> Updated archive and index workers to conditionally skip block/receipt operations when in traces-only mode

### DIFF
--- a/monad-archive/src/workers/block_archive_worker.rs
+++ b/monad-archive/src/workers/block_archive_worker.rs
@@ -38,6 +38,10 @@ pub struct ArchiveWorkerOpts {
     pub unsafe_skip_bad_blocks: bool,
     /// If set, archiver will require traces to be present for all blocks
     pub require_traces: bool,
+    /// If set, archiver will only archive traces
+    pub traces_only: bool,
+    /// If set, archiver will perform an asynchronous backfill of the archive
+    pub async_backfill: bool,
 }
 
 /// Main worker that archives block data from the execution database to durable storage.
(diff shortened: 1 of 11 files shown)


monad-archive/src/workers/block_archive_worker.rs
```diff
@@ -38,6 +38,10 @@ pub struct ArchiveWorkerOpts {
     pub unsafe_skip_bad_blocks: bool,
     /// If set, archiver will require traces to be present for all blocks
     pub require_traces: bool,
+    /// If set, archiver will only archive traces
+    pub traces_only: bool,
+    /// If set, archiver will perform an asynchronous backfill of the archive
+    pub async_backfill: bool,
```
philipglazman, line 44: Is there a workflow where we set `traces_only` to true and `async_backfill` to false ?

JoeHowarth: Yeah would have used it on Sunday. We wanted to upload traces only to a new bucket and have it follow the tip. Though I admit it's a more niche use case.

More common to do the inverse and have `async_backfill=true` and `traces_only=false`

PR #2608, opened 2025-11-25
txgen: Shutdown CommittedTxWatcher with shutdown flag

eclark0426:


monad-eth-testutil/examples/txgen/workers/committed_tx_watcher.rs
```diff
@@ -60,12 +62,18 @@ impl CommittedTxWatcher {
...
+            if shutdown.load(Ordering::Relaxed) {
+                break;
+            }
+
             let block = match block {
                 Ok(b) => b,
                 Err(e) => {
                     warn!("Blockstream returned error: {e}");
+                    let now = Instant::now();
+                    self.sent_txs.retain(|_, v| *v + self.delay > now);
```
dshulyak: does it make sense that it is cleared on error? maybe document why it makes sense.

maybe would be better to clear self.sent_txs where sent_txs are extended or atleast setup timer with select! above

monad-eth-testutil/examples/txgen/workers/committed_tx_watcher.rs
```diff
@@ -60,12 +62,18 @@ impl CommittedTxWatcher {
         }
     }
 
-    pub async fn run(mut self) {
+    pub async fn run(mut self, shutdown: Arc<AtomicBool>) {
         while let Some(block) = self.blockstream.next().await {
+            if shutdown.load(Ordering::Relaxed) {
+                break;
+            }
```
dshulyak: if blockstream is not producing anything then shutdown won't be reachable, there are some utilities to have awaitable shutdown

dshulyak (approved):

monad-eth-testutil/examples/txgen/workers/committed_tx_watcher.rs
eclark0426: This probably isn't actually needed any more. I think this is just a remnant of different things I tried to make sure things were getting cleaned up, before I realized it was the task as a whole that was not getting shut down. I'll revert and check that it doesn't cause issues

monad-eth-testutil/examples/txgen/workers/committed_tx_watcher.rs
eclark0426: Currently, we always use the same blockstream so it won't ever stop producing blocks. My [initial version](https://github.com/category-labs/monad-bft/commit/20ad1cd544724ecb0ed9fd1258a315a7ff51f05e) of the change did actually account for this, but it was a much smaller diff to do it how it is now. I can go back to something more like the original if you'd prefer

dshulyak: no strong opinion

JoeHowarth: Related issue - nothing shuts down blockstream! A new one is created for each workload group

See line 55

JoeHowarth: Nvrm, blockstream is not a worker but an owned stream

eclark0426: Seems like @JoeHowarth also preferred the original version, so I've reverted to that and propagated the clear on error change

dshulyak (approved):

PR #2610, opened 2025-11-25
Archive: Refactor monad-block-writer

JoeHowarth:
- Add streaming support to automatically follow tip
- Default to using 1M, 2M, etc. folders instead of dumping all files together
- Add FsStorage KVStore + KVReader - generally useful
- Updated monad-compress BoundedWriter with zero-cost Into<Vec<u8>> support
- Drive by fix for monad-indexer --fallback-data-source not having correct clap #[arg(..)]

### DIFF
--- a/monad-archive/src/bin/monad-block-writer/main.rs
+++ b/monad-archive/src/bin/monad-block-writer/main.rs
@@ -13,16 +13,15 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-use std::sync::Arc;
+use std::vec::IntoIter;
 
 use alloy_consensus::Block as AlloyBlock;
 use alloy_rlp::Encodable;
 use clap::Parser;
-use futures::future::join_all;
 use monad_archive::prelude::*;
 use monad_compress::{brotli::BrotliCompression, CompressionAlgo};
-use tokio::sync::Semaphore;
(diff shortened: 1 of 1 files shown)


monad-archive/src/bin/monad-block-writer/main.rs
```diff
@@ -51,50 +51,156 @@ async fn process_block(
...
-            .await?;
+                if start >= stop {
+                    info!(
+                        "No new blocks to process, sleeping for {} seconds",
+                        stream_args.sleep_secs
+                    );
+                    tokio::time::sleep(Duration::from_secs_f64(stream_args.sleep_secs)).await;
+                    continue;
+                }
+                stop = stop.min(start + 1_000_000);
```
eclark0426: Why do we need to run in increments of a million blocks at most? Is that just to make sure that the latest block is updated reasonably often?

eclark0426 (dismissed):
Seems fine to me. One comment that's mostly just for my understanding, but otherwise LGTM

monad-archive/src/bin/monad-block-writer/main.rs
JoeHowarth: Copying from slack:
> yeah it makes it save latest more often, but also means that blocks get written closer to in-order which means that execution can progress while it's still downloading. That also motivated the switch from semaphores to stream::buffered

PR #2613, opened 2025-11-26
RPC: allow rpc to run without monad-bft

JoeHowarth:
This is useful for genesis replay boxes. The current solution involves spoofing the ipc path and sending a bincode serialized empty HashSet through before closing - this is clearly not ideal and should not be part of the instructions given to external operators.

### DIFF
--- a/monad-rpc/src/cli.rs
+++ b/monad-rpc/src/cli.rs
@@ -21,8 +21,9 @@ use clap::Parser;
 #[command(name = "monad-rpc", about, long_about = None, version = monad_version::version!())]
 pub struct Cli {
     /// Set the mempool ipc path
+    /// If not set, the tx pool will be disabled.
     #[arg(long)]
-    pub ipc_path: PathBuf,
+    pub ipc_path: Option<PathBuf>,
 
     /// Set the monad triedb path
     #[arg(long)]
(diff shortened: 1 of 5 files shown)


monad-rpc/src/cli.rs
```diff
@@ -22,7 +22,12 @@ use clap::Parser;
 pub struct Cli {
     /// Set the mempool ipc path
     #[arg(long)]
-    pub ipc_path: PathBuf,
+    pub ipc_path: Option<PathBuf>,
+
+    /// Disable the tx pool
+    /// Cannot be used with --ipc-path
+    #[arg(long)]
+    pub disable_tx_pool: bool,
```
eerkaijun: do we need `disable_tx_pool`? if we already make `ipc_path` to be Option, can we just infer that tx pool is disabled if `ipc_path` is None?

JoeHowarth: That is logically true, but I'm worried it's too easy for an operator to forget ipc path by accident without intending to disable the txpool. Thoughts?

omegablitz: There's already precedent for this, eg execution can be started without statesync-socket set, and the stack won't start up correctly.

I don't think it'll be a problem in practice to make it optional

JoeHowarth: ok sounds good, new revision up

omegablitz (approved):

PR #2614, opened 2025-11-26
Adds additional indexer workflow to txgen

philipglazman:
Adds an indexer rpc pattern to rpc request gen that includes tracing blocks with calltracer and prestatetracer. Also updates the configuration option for txgen

### DIFF
--- a/monad-eth-testutil/examples/txgen/config.rs
+++ b/monad-eth-testutil/examples/txgen/config.rs
@@ -13,7 +13,7 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-use std::{collections::HashSet, str::FromStr};
+use std::str::FromStr;
 
 use eyre::bail;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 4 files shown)


monad-eth-testutil/examples/txgen/workers/rpc_request_gen.rs
```diff
@@ -108,23 +116,405 @@ where
...
+                                    .send(Err(BlockIndexError {
+                                        block_number,
+                                        error: err,
+                                    }))
+                                    .await;
+                                return Err(());
+                            }
+                        }
+                    }
+                    Ok(receipts.pop().unwrap())
```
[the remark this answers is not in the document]
philipglazman: requests_per_block == 0 is an invalid toml

monad-eth-testutil/examples/txgen/workers/rpc_request_gen.rs
```diff
@@ -108,23 +116,408 @@ where
...
+                            Self::debug_trace_block_by_number(
+                                &client,
+                                block_number,
+                                GethDebugBuiltInTracerType::CallTracer,
+                            )
+                        })
+                        .collect::<FuturesUnordered<_>>();
+
+                    let mut traces = Vec::new();
+                    while let Some(result) = futures.next().await {
```
JoeHowarth: iirc this sends all the requests as fast as possible without rate limiting. Is this the intention? Makes sense from a worst case stress testing perspective but hopefully not what real indexers would do

monad-eth-testutil/examples/txgen/workers/rpc_request_gen.rs
```diff
@@ -108,23 +116,408 @@ where
...
+    ) {
+        let block = match Self::get_block_by_number(&client, block_number).await {
+            Ok(header) => header,
+            Err(err) => {
+                warn!(?err, "Failed to get block by number");
+                return;
+            }
+        };
+
+        let uniq_addrs = if !block.transactions.is_empty() {
```
JoeHowarth, line 367: This assignment from a very large block typically means the block should be a fn instead. I won't block on this but consider breaking this apart or factoring it out

monad-eth-testutil/examples/txgen/workers/rpc_request_gen.rs
```diff
@@ -529,3 +970,66 @@ impl RpcWsCompare {
...
+    ) -> Self {
+        Self {
+            tip: 0,
+            tip_sender: sender,
+            client,
+        }
+    }
+
+    async fn run(mut self) {
+        loop {
```
JoeHowarth: This init code is confusing and at least deserves a comment. 

Overall though I would recommend collapsing these into 1 loop with a an additional check 

```
// init self.tip
if self.tip == 0 {
    self.tip = tip;

} else if tip > self.tip {
    // only start indexing once tip has been initialized
    ... existing logic
}
```

monad-eth-testutil/examples/txgen/workers/rpc_request_gen.rs
```diff
@@ -346,7 +739,54 @@ impl RpcWalletSpam {
...
+        let mut uniq_addrs = BTreeSet::new();
+
+        // select on tip refresher channel and completion channel
+        while !shutdown.load(Ordering::Relaxed) {
+            tokio::select! {
+                Some((index_from, index_to)) = tip_receiver.recv() => {
+                    debug!(from=index_from, to=index_to, "index range");
+
+                    for block_number in index_from..=index_to {
+                        RpcRequestGenerator::index_block(self.rpc_client.clone(), block_number, self.requests_per_block, index_done_sender.clone()).await;
```
JoeHowarth: If you are doing these calls synchronously, then why are you sending the result back on a channel instead of returning them from the fn? 
My guess is you meant for these to run async? Since there's a decent chance doing this sync could cause the indexer to fall behind.

monad-eth-testutil/examples/txgen/workers/rpc_request_gen.rs
```diff
@@ -346,7 +739,54 @@ impl RpcWalletSpam {
...
+                        Err(err) => {
+                            total_failed += 1;
+                            warn!(?err.block_number, ?err.error,"failed to index block");
+                            err.block_number
+                        }
+                    };
+                }
+
+                _ = status_interval.tick() => {
+                    debug!(indexed=total_indexed, failed=total_failed, addrs=uniq_addrs.len(), "indexer status");
```
JoeHowarth, line 791: Do we want any other mechanism for reporting or metrics here besides logs? Might have missed the plumbing if you have it already

monad-eth-testutil/examples/txgen/config.rs
JoeHowarth, line 826: Can you add the rpc req generator config to one of the sample configs so there is an example of how to use it? 

This is also useful to see the ux and iterate on it if it's not clear / easy.

monad-eth-testutil/examples/txgen/workers/rpc_request_gen.rs
```diff
@@ -108,23 +116,408 @@ where
...
+                            ..Default::default()
+                        },
+                    ),
+                )
+                .await;
+            match resp {
+                Ok(_) => return resp,
+                Err(err) => {
+                    Self::handle_rpc_error(err)?;
+                    tokio::time::sleep(Duration::from_millis(10)).await;
```
eclark0426: nit: all these sleep times are 10ms, can that be pulled into a constant so that it's easily configurable throughout?

eclark0426: Also is it intentional that you sleep on errors, but not on successful calls?

monad-eth-testutil/examples/txgen/config.rs
```diff
@@ -394,6 +386,49 @@ impl Default for TrafficGen {
...
+    Indexer,
+    /// Spam RPC and WebSocket with wallet workflow requests and compare responses
+    SpamRpcWs,
+    /// Compare block headers returned from RPC and WebSocket
+    CompareRpcWs,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[serde(default)]
+pub struct RpcRequestGeneratorConfig {
```
eclark0426: To maintain parity with how traffic generators are instantiated in a config file, I wonder if it might make sense to have request generator configs also be an array so that if/when they need generator-specific configs, it's easier to add them than it is currently with just a hashmap of the enabled generators. And in fact I think we could already use gen-specific configs since `Indexer` and `SpamRpcWs` use the `requests_per_block` configuration, while I don't ever see `CompareRpcWs` making use of it

monad-eth-testutil/examples/txgen/workers/rpc_request_gen.rs
philipglazman: Updated. The sleep is a primitive way to handle retry logic for spurious errors

monad-eth-testutil/examples/txgen/workers/rpc_request_gen.rs
philipglazman: Will include reporting in a followup to keep diff small

PR #2617, opened 2025-11-26
Update earliest block

eerkaijun:


omegablitz:
How is earliest used in practice?

eerkaijun:
it's rarely used in practice, by convention it usually just points to the genesis block. we never properly implement it (there has been previous discussion to use `earliest` as the tag to indicate the earliest block stored locally in triedb). the change is mostly to remove ambiguity for audits

omegablitz:
Ok, it seems a bit weird because generally RPCs won't work on block 0, but if noone uses it anyways then it's fine

PR #2619, opened 2025-11-27
Add timeout to txpool ipc stream

andr-dev:


monad-eth-txpool-ipc/src/lib.rs
```diff
@@ -100,7 +95,17 @@ impl EthTxPoolIpcStream {
 
                     let events = bincode::serialize(&events).expect("txpool events are serializable");
 
-                    stream.send(events.into()).await?;
+                    match tokio::time::timeout(std::time::Duration::from_secs(1), stream.send(events.into())).await {
```
dshulyak: so the stream is read in the same task (e.g right above this)? so if we await here, that endpoint should become ready?

i am trying to understand when it actually can timeout without failure, e.g if stream is closed i would expect `stream.send(events.into()).await?;` to exit with error

andr-dev: If the other end of the socket stops processing messages, the socket will fill up until the max size is hit.

ariqchowdhury (approved):

PR #2631, opened 2025-12-01
Minor performance improvements / cleanup for RPC eth_getLogs

andr-dev:
On top of reducing memcpys and unifying the receipt -> logs process, these changes facilitate a new `eth_getLogs` feature coming soon which will heuristically short-circuit requests based on whether the expected JSON response size exceeds the configured max response size to enable large log request ranges.

### DIFF
--- a/monad-rpc/src/chainstate/mod.rs
+++ b/monad-rpc/src/chainstate/mod.rs
@@ -941,17 +941,18 @@ async fn get_logs_with_index(
     to_block: u64,
     filter: &Filter,
 ) -> monad_archive::prelude::Result<Vec<Log>> {
-    let potential_matches = get_receipts_stream_using_index(reader, from_block, to_block, filter)
-        .await?
-        .try_collect::<Vec<_>>()
-        .await?;
-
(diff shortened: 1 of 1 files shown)


monad-rpc/src/chainstate/mod.rs
```diff
@@ -929,20 +931,28 @@ async fn get_logs_with_index(
...
+    reader: &ArchiveReader,
+    from_block: u64,
+    to_block: u64,
+    filter: &Filter,
+) -> monad_archive::prelude::Result<Vec<Log>> {
+    let filtered_params = FilteredParams::new(Some(filter.clone()));
+
+    get_receipts_stream_using_index(reader, from_block, to_block, filter)
+        .await?
+        .map_ok(|receipt| {
```
philipglazman, line 948: Might be worthwhile to add a map_err to for logging or otherwise if item is Error

andr-dev: I took a quick look and it seems like all callers of the function let the error bubble up to some point where it eventually gets logged.

PR #2633, opened 2025-12-01
Persist peers periodically

eerkaijun:
closes https://github.com/category-labs/category-internal/issues/1659

supercedes https://github.com/category-labs/monad-bft/pull/2529

### DIFF
--- a/monad-peer-discovery/src/discovery.rs
+++ b/monad-peer-discovery/src/discovery.rs
@@ -16,6 +16,7 @@
 use std::{
     collections::{BTreeMap, BTreeSet, HashMap, HashSet},
     net::SocketAddrV4,
+    path::PathBuf,
     time::Duration,
 };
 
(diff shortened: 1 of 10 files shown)


monad-peer-discovery/src/discovery.rs
```diff
@@ -608,6 +615,123 @@ impl<ST: CertificateSignatureRecoverable> PeerDiscovery<ST> {
...
+                auth_port: name_record.name_record.authenticated_udp_port(),
+            })
+            .collect();
+        for (peer_id, conn_info) in self.pending_queue.iter() {
+            peer_configs.push(NodeBootstrapPeerConfig {
+                address: conn_info.name_record.udp_address().to_string(),
+                record_seq_num: conn_info.name_record.seq(),
+                secp256k1_pubkey: peer_id.pubkey(),
+                name_record_sig: conn_info.name_record.signature,
+                auth_port: conn_info.name_record.name_record.authenticated_udp_port(),
```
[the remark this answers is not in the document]
eerkaijun: current integration is such that udp and tcp is the same port

monad-peer-discovery/src/discovery.rs
```diff
@@ -2462,4 +2592,133 @@ mod tests {
         let selected_peers = state.select_peers_to_lookup_from();
         assert_eq!(selected_peers.len(), 3);
     }
+
+    #[test]
+    fn test_persist_peers() {
```
dshulyak, line 2561: could you please add an insta test to easily spot that persisted format is modified, and that we can recover previously persisted data?

monad-peer-discovery/src/discovery.rs
```diff
@@ -608,6 +615,123 @@ impl<ST: CertificateSignatureRecoverable> PeerDiscovery<ST> {
...
+        let mut peer_configs: Vec<NodeBootstrapPeerConfig<ST>> = self
+            .routing_info
+            .iter()
+            .map(|(peer_id, name_record)| NodeBootstrapPeerConfig {
+                address: name_record.udp_address().to_string(),
+                record_seq_num: name_record.seq(),
+                secp256k1_pubkey: peer_id.pubkey(),
+                name_record_sig: name_record.signature,
+                auth_port: name_record.name_record.authenticated_udp_port(),
+            })
```
dshulyak, line 632: i would add try_from/try_into (or non-fallible variants) between name record and bootstrap peer config

monad-peer-discovery/src/discovery.rs
eerkaijun: great point, added

monad-peer-discovery/src/discovery.rs
eerkaijun: updated

monad-peer-discovery/src/lib.rs
```diff
@@ -518,6 +519,61 @@ impl<ST: CertificateSignatureRecoverable> TryFrom<&MonadNameRecord<ST>> for Peer
     }
 }
 
+// Helper function to convert (NodeId, MonadNameRecord) to NodeBootstrapPeerConfig
+pub fn peer_to_bootstrap_config<ST: CertificateSignatureRecoverable>(
```
dshulyak: i made this for similar case https://github.com/category-labs/monad-bft/blob/5fc291c7202c4f92328cc38f980214c400a70b65/monad-peer-discovery/src/lib.rs#L467-L476

i think adding another cast to that type will be clear

eerkaijun: ok

PR #2634, opened 2025-12-01
Cleanup iterations across RPC

andr-dev:


JoeHowarth (approved):

monad-rpc/src/handlers/debug.rs
```diff
@@ -738,27 +738,32 @@ async fn include_code_output<T: Triedb>(
...
         let depth = value.depth.to::<usize>();
         let new_node = Rc::new(RefCell::new(MonadCallFrame::from(value)));
-        while let Some(last) = stack.last() {
-            if last.borrow().depth < depth {
-                last.borrow_mut().calls.push(Rc::clone(&new_node));
+
+        loop {
+            let mut last = stack
+                .last()
+                .expect("Call tree root node is never removed from stack")
```
JoeHowarth: For wellformed data this is likely true, but we should make rpc defensive to malformed data and not crash the program. Which implies this should error! loudly and also return Err()

JoeHowarth (changes requested):

monad-rpc/src/handlers/debug.rs
andr-dev: Panics in RPC get caught by actix and drop the connection instead of crashing the program. That being said, this is still worse than just returning an error, updated!

philipglazman (approved):

PR #2641, opened 2025-12-03
EthTxPoolIpcServer: refactor as trait

robertkiel:
Refactor `EthTxPoolIpcServer` as a trait so that (e2e) testing does not rely on OS sockets.

```rust
pub trait TxPoolServer: Sized {
    fn new(config: EthTxPoolIpcConfig) -> Result<Self, io::Error>;

    fn broadcast_tx_events(self: Pin<&mut Self>, events: BTreeMap<TxHash, EthTxPoolEventType>);

    fn poll_txs(
        self: Pin<&mut Self>,
        cx: &mut Context<'_>,
        generate_snapshot: impl Fn() -> EthTxPoolSnapshot,
    ) -> Poll<Vec<TxEnvelope>>;
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5141,6 +5141,7 @@ name = "monad-dataplane"
 version = "0.1.0"
 dependencies = [
  "bytes",
+ "clap 4.5.47",
  "env_logger 0.10.2",
  "futures",
  "futures-util",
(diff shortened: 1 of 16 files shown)


andr-dev:
Resolved by https://github.com/category-labs/monad-bft/pull/2718.

PR #2644, opened 2025-12-03
Simplify RPC txpool status tracking

andr-dev:
Previously, RPC tracked pool state using a tuple of the last status received from txpool (tracked, committed, etc.) and an optional one-shot sender which was used to allow tasks to "subscribe" to the next change in status. This "subscription" capability was added so that RPC methods like `eth_sendRawTransaction` can submit a tx to the pool and be notified once a status is available for the tx (which occurs after the pool processes the tx).

This design had two key flaws:
1. Only one task could "subscribe" at a time (causes issues when the same tx is replayed)
2. Adds additional complexity since the status value is stored in two places (in tuple + possibly in channel)

To fix both of these issues, the tuple was replaced with a `tokio::sync::watch` channel which retains the last value sent. This allows multiple tasks to "subscribe" to tx status changes while storing the last status in a single place.

A test was added showing that multiple tasks can now "subscribe" to the same status channel at the same time. Additionally, another test shows how snapshots do not notify subscribers, ensuring that tasks continue waiting for a _new_ status if the txpool socket gets reconnected.

### DIFF
--- a/monad-rpc/src/handlers/eth/txn.rs
+++ b/monad-rpc/src/handlers/eth/txn.rs
@@ -219,37 +219,31 @@ pub async fn monad_eth_sendRawTransaction(
     Ok(tx_hash.to_string())
 }
 
-fn validate_and_decode_tx<F>(
+fn validate_and_decode_tx(
     hex_tx: &[u8],
     chain_id: u64,
     allow_unprotected_txs: bool,
-    decode_error_fn: F,
-) -> Result<TxEnvelope, JsonRpcError>
-where
-    F: Fn() -> JsonRpcError,
-{
-    match TxEnvelope::decode(&mut &hex_tx[..]) {
-        Ok(tx) => {
-            // drop pre EIP-155 transactions if disallowed by the rpc (for user protection purposes)
-            if !allow_unprotected_txs && tx.chain_id().is_none() {
-                return Err(JsonRpcError::custom(
-                    "Unprotected transactions (pre-EIP155) are not allowed over RPC".to_string(),
-                ));
-            }
+    decode_error_fn: impl FnOnce() -> JsonRpcError,
+) -> Result<TxEnvelope, JsonRpcError> {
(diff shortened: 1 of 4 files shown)


monad-rpc/src/handlers/eth/txn.rs
```diff
@@ -258,42 +252,59 @@ async fn submit_to_txpool(
...
+                TxStatus::Tracked
+                | TxStatus::Dropped { .. }
+                | TxStatus::Evicted { .. }
+                | TxStatus::Committed => Err(JsonRpcError::custom(
+                    "rpc no longer tracking tx".to_string(),
+                )),
+            };
+        }
+        Err(_) => {
+            warn!("txpool bridge not responding, tx status has not changed");
```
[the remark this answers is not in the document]
andr-dev: Added comment explaining the intended behavior.

monad-rpc/src/handlers/eth/txn.rs
```diff
@@ -258,42 +252,61 @@ async fn submit_to_txpool(
...
+            // currently has, even if it might be stale.
+            warn!("txpool bridge not responding, tx status has not changed");
         }
     }
 
-    Err(JsonRpcError::custom("txpool not responding".to_string()))
+    let latest_tx_status = tx_status_recv.borrow_and_update().to_owned();
+
+    match latest_tx_status {
+        TxStatus::Evicted { reason: _ } => Err(JsonRpcError::custom("rejected".to_string())),
```
JoeHowarth, line 302: Why do we include the reason for Dropped but not Evicted?

JoeHowarth (approved):

monad-rpc/src/handlers/eth/txn.rs
andr-dev: Txs should never get evicted when being submit, so this branch should never get hit. If it does, something weird happened so we can just blanket state that it was rejected by txpool.

JoeHowarth (approved):

PR #2645, opened 2025-12-03
Fix publisher group garbage cleaning logic

michael-yxchen:
The group spans are exclusive of the end round. Publisher should only retain groups with end rounds greater than the current round


michael-yxchen:
yeah I'll add a unit test

andr-dev (approved):

mikety (approved):

PR #2649, opened 2025-12-04
Add WritePolicy to KVStore for conditional write protection

JoeHowarth:
Add WritePolicy enum (AllowOverwrite, NoClobber) and PutResult enum (Written, Skipped) to control whether put operations can overwrite existing keys.

Implementation:
- S3: Uses if_none_match("*") for conditional writes, handles PreconditionFailed error
- FS: Uses OpenOptions::create_new(true), handles AlreadyExists error
- Memory: Checks contains_key() before insert
- MongoDB: uses insert_one 
- DynamoDB: Stub implementations that always overwrite since no batch api exists for NoClobber policy

Block data archive now uses NoClobber for blocks, receipts, and traces while using AllowOverwrite for the "latest" marker.

### DIFF
--- a/monad-archive/src/bin/monad-archiver/bft_archive_worker.rs
+++ b/monad-archive/src/bin/monad-archiver/bft_archive_worker.rs
@@ -176,7 +176,7 @@ async fn process_single_file(
         .wrap_err("Failed to read local BFT file")?;
 
     store
-        .put(&key, bytes, WritePolicy::AllowOverwrite)
+        .put(&key, bytes, WritePolicy::NoClobber)
         .await
         .wrap_err("Failed to upload BFT block")?;
     metrics.inc_counter(MetricNames::BFT_BLOCK_FILES_UPLOADED);
@@ -377,7 +377,7 @@ mod tests {
             .put(
                 &header_key,
                 b"already uploaded".to_vec(),
-                WritePolicy::AllowOverwrite,
+                WritePolicy::NoClobber,
             )
             .await
             .unwrap();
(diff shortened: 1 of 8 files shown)


monad-archive/src/kvstore/mod.rs
```diff
@@ -97,10 +116,19 @@ impl From<KVStoreErased> for KVReaderErased {
...
 pub trait KVStore: KVReader {
-    async fn put(&self, key: impl AsRef<str>, data: Vec<u8>) -> Result<()>;
-    async fn bulk_put(&self, kvs: impl IntoIterator<Item = (String, Vec<u8>)>) -> Result<()> {
+    async fn put(
+        &self,
+        key: impl AsRef<str>,
+        data: Vec<u8>,
+        policy: WritePolicy,
+    ) -> Result<PutResult>;
+    async fn bulk_put(
```
mikety, line 125: `bulk_put` is not using individual `PutResult` - I think it will be consistent to report if all puts were successful, i.e. `Skipped` if at least one put returned `Skipped` ?

monad-archive/src/kvstore/mongo.rs
```diff
@@ -236,7 +236,13 @@ impl KVStore for MongoDbStorage {
...
     }
 
-    async fn put(&self, key: impl AsRef<str>, data: Vec<u8>) -> Result<()> {
+    async fn put(
+        &self,
+        key: impl AsRef<str>,
+        data: Vec<u8>,
+        _policy: WritePolicy,
+    ) -> Result<PutResult> {
+        // Note: WritePolicy is ignored for MongoDB - always overwrites
```
mikety: Why not to implement the WritePolicy?

monad-archive/src/bin/monad-archiver/bft_archive_worker.rs
```diff
@@ -176,7 +176,7 @@ async fn process_single_file(
         .wrap_err("Failed to read local BFT file")?;
 
     store
-        .put(&key, bytes)
+        .put(&key, bytes, WritePolicy::AllowOverwrite)
```
mikety: bft blocks do not need overwrite protection? Arguably bft blocks are primary since the rest can be derived from them via re--execution

PR #2650, opened 2025-12-04
raptorcast: safe parsing of the raptorcast packet

dshulyak:


monad-raptorcast/src/udp.rs
```diff
@@ -851,6 +943,7 @@ mod tests {
     }
 
     #[test]
+    #[ignore] // ignoring because this test is very slow
```
dshulyak: maybe refactor it, it should not be taking 30s to run it

monad-raptorcast/src/udp.rs
```diff
@@ -367,194 +593,60 @@ pub enum MessageValidationError {
...
-    let signature = <ST as CertificateSignature>::deserialize(&cursor_signature)
-        .map_err(|_| MessageValidationError::InvalidSignature)?;
-
-    let cursor_version = split_off(2)?;
-    let version = u16::from_le_bytes(cursor_version.as_ref().try_into().expect("u16 is 2 bytes"));
-    if version != 0 {
+    // we do basic validation in parse to guarantee that all fields can be read safely
+    let packet = RaptorcastPacket::parse(&message)?;
+    // below is the protocol specific validation
+    if packet.header.version.get() != 0 {
```
xinyuan-dev: Nit: maybe move this check to `RaptorcastPacket::parse` since if the version doesn't match it probably isn't useful to parse the remaining of the fields.

dshulyak: i updated to have a enum with versioned payload, and reject if its not 0

monad-raptorcast/src/udp.rs
```diff
@@ -38,14 +39,307 @@ use crate::{
...
+        if rest.len() < RaptorcastHeaderV0::SIZE {
+            return Err(MessageValidationError::TooShort);
+        }
+        let (header_bytes, rest) = rest.split_at(RaptorcastHeaderV0::SIZE);
+        let header: Ref<&[u8], RaptorcastHeaderV0> =
+            Ref::from_bytes(header_bytes).map_err(|_| MessageValidationError::TooShort)?;
+
+        let tree_depth = header.tree_depth();
+        #[allow(clippy::manual_range_contains)]
+        if tree_depth < MIN_MERKLE_TREE_DEPTH || tree_depth > MAX_MERKLE_TREE_DEPTH {
```
eclark0426: Is there a reason for ignoring the clippy error instead of using the `!(MIN_MERKLE_TREE_DEPTH..=MAX_MERKLE_TREE_DEPTH).contains(&tree_depth)` syntax that clippy wants?

dshulyak: i wasn't sure if it compiles to the same code, but looks like it does compile to this in both cases. i will switch it

```
        .cfi_startproc
        decq    %rdi
        cmpq    $9, %rdi
        setb    %al
        retq
```

eclark0426 (approved):

PR #2652, opened 2025-12-04
ConfirmGroup message fix

michael-yxchen:
It was possible to create a ConfirmGroup message with peer list longer than max_group_size in the original invitation. This ConfirmGroup message will get rejected by full nodes. This PR introduces a NoConfirm message that's immediately sent to the client when group reaches max size. When client receives the NoConfirm message, it releases the round to other invites.

Regarding backwards compatibility, the NoConfirm message is a new addition. Older clients will simply ignore this message (fail to deserialize) and will not experience any negative effects.

### DIFF
--- a/monad-raptorcast/src/raptorcast_secondary/client.rs
+++ b/monad-raptorcast/src/raptorcast_secondary/client.rs
@@ -28,7 +28,10 @@ use super::{
     super::config::RaptorCastConfigSecondaryClient,
     group_message::{ConfirmGroup, PrepareGroup, PrepareGroupResponse},
 };
-use crate::util::{SecondaryGroup, SecondaryGroupAssignment};
+use crate::{
+    raptorcast_secondary::group_message::NoConfirm,
+    util::{SecondaryGroup, SecondaryGroupAssignment},
+};
 
 monad_executor::metric_consts! {
     pub CLIENT_NUM_CURRENT_GROUPS {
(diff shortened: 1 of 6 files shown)


monad-raptorcast/src/raptorcast_secondary/publisher.rs
```diff
@@ -479,16 +479,27 @@ where
...
             debug!(
-                ?self.full_nodes_accepted.list,
+                accepted = ?self.full_nodes_accepted.list,
                 "RaptorCastSecondary Publisher confirm group formed",
             );
             self.next_invite_tp = TimePoint::MAX; // lock the group
+
+            // the first max_group_size peers are confirmed into the group
+            let confirmed_peers = self
+                .full_nodes_accepted
```
dshulyak: so this effectively just drops other responses? does it make sense to drop them in the on_candidate_response? with a log or maybe metric

not a big difference, but might be easier to debug

dshulyak (commented):
would be nice to have a test too, to prevent regressions

monad-raptorcast/src/raptorcast_secondary/client.rs
```diff
@@ -318,43 +318,55 @@ where
...
 
-        let maybe_entry = invites.get(&confirm_msg.prepare.validator_id);
-        let Some(old_invite) = maybe_entry else {
+        // Validator sends confirmed group once per invite. Removing from
+        // pending_confirms. If ConfirmGroup is not validated/accepted, removing
+        // it from pending_confirms releases the rounds to other invites
+
+        // TODO: discount validator reputation if it has sent an invalid
+        // ConfirmGroup
+        let Some(accepted_invite) = invites.remove(&confirm_msg.prepare.validator_id) else {
```
eerkaijun: does a full node that doesn't get included in the confirm group message still get sent this message? (assuming the full node has previously accepted the prepare group)

monad-raptorcast/src/raptorcast_secondary/client.rs
```diff
@@ -318,43 +318,55 @@ where
...
         }
 
         if !confirm_msg.peers.contains(&self.client_node_id) {
+            if confirm_msg.peers.len() == confirm_msg.prepare.max_group_size {
+                debug!(
+                    ?confirm_msg,
+                    "Group invite has reached max_group_size. Node participation is not confirmed"
+                );
+                return;
+            }
```
eerkaijun, line 381: what's the reason for needing this additional check if we are already returning anyways?

monad-raptorcast/src/raptorcast_secondary/client.rs
michael-yxchen: on FN side, accepting an invite reserves the given round span and occupies a slot towards max_num_group. I'm sending a ConfirmGroup to all that have responded so the excluded ones can release the lock to other invites

thinking now, i want to introduce a NoConfirm message response when we have already reached max_num_peers. I'll put up that change in this PR

monad-raptorcast/src/raptorcast_secondary/client.rs
michael-yxchen: not including a FN when reaching proposed max_group_size is valid. the other branch represents a malformed message

monad-raptorcast/src/raptorcast_secondary/publisher.rs
michael-yxchen: yeah i think it'd be cleaner to drop response in `on_candidate_response`. i also want to introduce a NoConfirm message type that can be sent back to the full node immediately, to release reserved rounds for a given invite

michael-yxchen:
You're right that `handle_full_node_raptorcast_request` has a guard that caps connected public fullnodes + prioritized nodes under max_group_size. I believe we don't have such a check on the config update path for updating prioritized full nodes. This change should still be valuable in guarding against mismatches like this

PR #2653, opened 2025-12-05
monad-dataplane: return bound socket addresses synchronously

dshulyak:
it allows to run tests with dataplane configured to :0 port, which guarantees that kernel will pick a port not used at the time of binding

fixes test flakes due to port binding in concurrent tests


xinyuan-dev (approved):

dshulyak:
@xinyuan-dev @omegablitz resolved conflict. please reapprove

eerkaijun (approved):

PR #2655, opened 2025-12-05
Use port from peer discovery config for constructing self name record

michael-yxchen:


monad-node/src/main.rs
```diff
@@ -580,18 +580,24 @@ where
     }
     dp_builder = dp_builder.extend_udp_sockets(udp_sockets);
 
+    // auth port in peer discovery config and network config should be set and unset simultaneously
+    assert_eq!(
+        peer_discovery_config.self_auth_port.is_some(),
+        network_config.authenticated_bind_address_port.is_some()
+    );
```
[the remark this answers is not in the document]
michael-yxchen: they can be different when running behind NAT

monad-node/src/main.rs
```diff
@@ -580,18 +580,24 @@ where
...
+
     let self_id = NodeId::new(identity.pubkey());
-    let self_record = match network_config.authenticated_bind_address_port {
+    let self_record = match peer_discovery_config.self_auth_port {
```
dshulyak, line 590: i would try to separate inputs that are used to generate name record somehow to make it harder to misuse.
e.g 

fn generate_name_record_from_config(resolved_address, peer_discovery_config) 

and maybe document on that function that this is intentionally not using bind addresses to support port mapping

but i can do that in separate change

michael-yxchen: cool i'll merge this change first for the release

PR #2656, opened 2025-12-06
RPC enable large eth_getLogs ranges

andr-dev:


monad-archive/src/model/logs_index.rs
```diff
@@ -193,20 +193,17 @@ impl LogsIndexArchiver {
...
-                    let (_id, bytes) = doc
-                        .resolve(&collection)
-                        .await
-                        .wrap_err("Failed to resolve logs index")?;
-                    reader
-                        .resolve_from_bytes(&bytes)
-                        .await
-                        .wrap_err("Failed to decode TxIndexedData when querying logs")
-                }
+            .then(move |doc| async move {
```
JoeHowarth, line 196: nice

monad-rpc/src/chainstate/mod.rs
```diff
@@ -781,6 +813,83 @@ impl<T: Triedb> ChainState<T> {
...
+                        ?num_blocks_processed,
+                        ?block_number,
+                        ?num_blocks_total,
+                        "logs stream block number exceeded range"
+                    );
+                    return Err(JsonRpcError::internal_error(format!("Logs out of range")));
+                }
+
+                response_logs.extend(logs.into_iter().map(|log| {
+                    heuristic_response_size += compute_heuristic_log_len(&log) as u64;
```
JoeHowarth: flup idea: could we return a hint for the largest block range that would fall below the limit?

monad-rpc/src/chainstate/mod.rs
```diff
@@ -781,6 +813,83 @@ impl<T: Triedb> ChainState<T> {
...
+                    MonadLog(log)
+                }));
+
+                if heuristic_response_size > max_response_size as u64 {
+                    return Err(JsonRpcError::max_size_exceeded());
+                }
+
+                if heuristic_response_size >= EXTRAPOLATION_CHECK_MIN_RESPONSE_SIZE
```
JoeHowarth, line 874: good example of where a comment explaining the rule and a justification for why this rule is reasonable would be good documentation

monad-rpc/src/chainstate/mod.rs
JoeHowarth, line 958: Note: we can remove this fn from production code path, though don't have to do so in this pr. 
It could be useful for testing new vs. old impls in the future, but not strictly necessary right now

monad-rpc/src/chainstate/mod.rs
```diff
@@ -1120,19 +1285,94 @@ async fn get_receipt_from_triedb<T: Triedb>(
     }
 }
 
+const HEURISTIC_SMALLEST_LOG_RESPONSE: &str = r#"{"address":"0x0000000000000000000000000000000000000000","blockHash":,"blockNumber":,"data":"0x","logIndex":,"removed":,"topics":[],"transactionHash":,"transactionIndex":}"#;
```
JoeHowarth: nit: this should definitely be in it's own module with it's own unit tests

monad-rpc/src/chainstate/mod.rs
```diff
@@ -1268,9 +1508,54 @@ mod tests {
...
+            "invalid arbitrary",
+            |bytes| {
+                let mut u = Unstructured::new(&bytes);
+                u.arbitrary().ok()
+            },
+        )
+    }
+
+    proptest! {
+        #[test]
```
JoeHowarth: What's the wall clock on these tests? proptests can sometimes take a while

monad-rpc/src/chainstate/mod.rs
andr-dev: Yeah this is a great idea, I'm skeptical on how many clients would actually parse it out to use for their next query but at least the information is there.

monad-rpc/src/chainstate/mod.rs
andr-dev: Under one second. It's pretty simple to generate an arbitrary tx, the heuristic function itself is super fast, and JSON serialization is minimal cost.

monad-rpc/src/chainstate/mod.rs
andr-dev: I'll do this in a separate PR to keep things clean, I want to change a couple of the error methods which will add to the line count and this one is already getting a bit unwieldy.

monad-rpc/src/chainstate/mod.rs
andr-dev: Added comments to the constants explaining their purpose.

PR #2658, opened 2025-12-08
Include log index in traces

philipglazman:
This PR adds an `index` field to each traced log in the debug call tracer output. A log index is the global order of a log within a transaction.

We compute the index by walking the execution trace in call order.

### DIFF
--- a/monad-ethcall/src/lib.rs
+++ b/monad-ethcall/src/lib.rs
@@ -132,6 +132,7 @@ impl From<MonadTracer> for u32 {
 
 pub const ETH_CALL_SUCCESS: i32 = 0;
 pub const EVMC_OUT_OF_GAS: i32 = 3;
+pub const EVMC_MONAD_RESERVE_BALANCE_VIOLATION: i32 = 18;
 
 #[derive(Clone, Debug, Default)]
 pub enum EthCallResult {
@@ -385,6 +386,24 @@ pub async fn eth_call(
                         gas_refund,
                         output_data,
                     })
+                }
+            }
+            EVMC_MONAD_RESERVE_BALANCE_VIOLATION => {
+                if tracer_cval == bindings::monad_tracer_config_NOOP_TRACER {
+                    CallResult::Failure(FailureCallResult {
+                        error_code: EthCallResult::OtherError,
+                        message: "reserve balance violation".to_string(),
+                        data: None,
+                    })
+                } else {
+                    let output_data_len = (*result).encoded_trace_len;
+                    let output_data = if output_data_len != 0 {
+                        std::slice::from_raw_parts((*result).encoded_trace, output_data_len)
+                            .to_vec()
+                    } else {
+                        vec![]
+                    };
+                    CallResult::Revert(RevertCallResult { trace: output_data })
                 }
             }
             _ => {


dhil (dismissed):

monad-rpc/src/handlers/debug.rs
```diff
@@ -745,15 +751,22 @@ async fn build_call_tree(
         return Ok(None);
     };
 
+    let mut log_index_counter = 0usize;
+
     let root = Rc::new(RefCell::new(MonadCallFrame::from(root)));
-    let mut stack = vec![Rc::clone(&root)];
+
+    // Stack tracks frame and the next log index to assign for that frame
```
JoeHowarth: I think this is right, but it's a bit tricky to follow. 
Consider a more descriptive comment detailing the algorithm?

JoeHowarth:
Also needs to update flexnet call trace tests

monad-rpc/src/handlers/debug.rs
philipglazman: Added comments

monad-ethcall/src/lib.rs
```diff
@@ -387,6 +388,24 @@ pub async fn eth_call(
...
+                    })
+                } else {
+                    let output_data_len = (*result).encoded_trace_len;
+                    let output_data = if output_data_len != 0 {
+                        std::slice::from_raw_parts((*result).encoded_trace, output_data_len)
+                            .to_vec()
+                    } else {
+                        vec![]
+                    };
+                    CallResult::Revert(RevertCallResult { trace: output_data })
```
JoeHowarth, line 406: This should still be a Failure, not revert. Even if it's not noop tracer, reserve balance checking happens before execution iirc, so a non-empty trace will never be emitted. 
This means that the reserve balance error message will never make it into the response. 

If I'm understanding this right, this whole block should be collapsed down to
```
  EVMC_MONAD_RESERVE_BALANCE_VIOLATION => {
      CallResult::Failure(FailureCallResult {
          error_code: EthCallResult::OtherError,
          message: "reserve balance violation".to_string(),
          data: None,
      })
  }
```

monad-rpc/src/handlers/debug.rs
```diff
@@ -686,9 +691,17 @@ pub async fn decode_call_frame<T: Triedb>(
...
                     Ok(None)
                 } else {
-                    let mut call_frame = call_frames.remove(0);
-                    include_code_output(&mut call_frame, triedb_env, block_key).await?;
-                    Ok(Some(MonadCallFrame::from(call_frame)))
+                    if let Some(root_frame) = call_frames.first_mut() {
+                        include_code_output(root_frame, triedb_env, block_key).await?;
+                    }
+
+                    let mut root = build_call_tree(call_frames)
```
JoeHowarth, line 698: This used to take the fast path and not decode the entire call tree just to grab the root node. Doesn't hugely matter, but a perf regression

monad-ethcall/src/lib.rs
JoeHowarth: Phil correctly pointed out I made an incorrect assumption and `EVMC_MONAD_RESERVE_BALANCE_VIOLATION` is being set after execution in `execute_call_message` / `execute_create_message`, not before execution

PR #2662, opened 2025-12-10
feat: eth_fillTransaction RPC endpoint

haythemsellami:
This PR implement the `eth_fillTransaction` RPC endpoint which fills the defaults (nonce, gas, gasPrice or 1559 fields) on a given unsigned transaction, and returns it to the caller for further processing (signing + broadcast).

This will be quite useful when combined with `eth_sendRawTransactionSync`.

For reference: https://github.com/ethereum/execution-apis/pull/512/files.

### DIFF
--- a/monad-rpc/src/handlers/eth/gas.rs
+++ b/monad-rpc/src/handlers/eth/gas.rs
@@ -15,7 +15,9 @@
 
 use std::ops::{Div, Sub};
 
-use alloy_consensus::{Header, Transaction, TxEnvelope};
+use alloy_consensus::{
+    Header, SignableTransaction, Transaction, TxEip1559, TxEip7702, TxEnvelope, TxLegacy,
+};
 use alloy_primitives::{Address, TxKind, U256, U64};
 use alloy_rpc_types::{FeeHistory, TransactionReceipt};
 use futures::stream::StreamExt;
@@ -28,10 +30,13 @@ use serde::Deserialize;
 use tracing::trace;
 
 use crate::{
-    chainstate::{get_block_key_from_tag_or_hash, ChainState},
-    handlers::eth::call::{fill_gas_params, CallRequest},
+    chainstate::{get_block_key_from_tag, get_block_key_from_tag_or_hash, ChainState},
+    handlers::eth::call::{fill_gas_params, CallRequest, GasPriceDetails},
     types::{
-        eth_json::{BlockTagOrHash, BlockTags, MonadFeeHistory, Quantity},
+        eth_json::{
+            BlockTagOrHash, BlockTags, FillTransactionResult, MonadFeeHistory, Quantity,
+            UnformattedData,
+        },
         jsonrpc::{JsonRpcError, JsonRpcResult},
     },
 };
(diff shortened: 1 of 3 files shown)


monad-rpc/src/handlers/eth/call.rs
```diff
@@ -1002,6 +1006,226 @@ pub async fn monad_admin_ethCallStatistics(
...
+        )
+        .await?;
+
+        tx.gas = Some(U256::from(estimated_gas));
+    }
+
+    let (raw, filled_tx) = build_unsigned_transaction(&tx, chain_id)?;
+
+    Ok(FillTransactionResult { raw, tx: filled_tx })
+}
```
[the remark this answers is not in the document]
haythemsellami: All of this should be covered in https://github.com/category-labs/monad-integration/pull/394

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -1002,6 +1006,226 @@ pub async fn monad_admin_ethCallStatistics(
...
+        gas_price_details: tx.gas_price_details.clone(),
+        value: Some(value),
+        input: tx.input.clone(),
+        nonce: Some(tx.nonce.unwrap_or_default()),
+        chain_id: Some(tx.chain_id.unwrap_or(U64::from(chain_id))),
+        ..tx.clone()
+    };
+
+    Ok((UnformattedData(raw_bytes), filled))
+}
```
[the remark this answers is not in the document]
haythemsellami: Added.

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -1002,6 +1006,226 @@ pub async fn monad_admin_ethCallStatistics(
...
+            }
+            Some(input)
+        }
+        (None, data) | (data, None) => data,
+    };
+
+    let from = match tx.from {
+        Some(from) => from,
+        None => return Err(JsonRpcError::invalid_params()),
+    };
```
andr-dev:
```rust
let from = tx.from.unwrap_or(JsonRpcError::invalid_params)?;
```

haythemsellami:
Related tests https://github.com/category-labs/monad-integration/pull/394.

monad-rpc/src/handlers/eth/call.rs
haythemsellami: Compiler don't like it as `.from` is `Option<Address>`.

haythemsellami: Updated this to use `ok_or_else()`.

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -1002,6 +1006,220 @@ pub async fn monad_admin_ethCallStatistics(
...
+    pub tx: CallRequest,
+}
+
+#[rpc(
+    method = "eth_fillTransaction",
+    ignore = "chain_id,provider_gas_limit,eth_call_executor"
+)]
+#[allow(non_snake_case)]
+pub async fn monad_eth_fillTransaction<T: Triedb>(
+    triedb_env: &T,
```
philipglazman: why not use `chain_state: &ChainState<T>` instead of Triedb?

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -1002,6 +1006,220 @@ pub async fn monad_admin_ethCallStatistics(
...
+
+    // Estimate gas if not specified
+    let gas_specified = tx.gas.is_some();
+    if !gas_specified {
+        let protocol_gas_limit = header.header.gas_limit;
+        let eth_call_provider_gas_limit = provider_gas_limit.min(protocol_gas_limit);
+        let original_tx_gas = U256::from(protocol_gas_limit);
+
+        // Set initial gas for estimation
+        tx.gas = Some(U256::from(eth_call_provider_gas_limit));
```
philipglazman: Do we want to set the initial gas limit  to the provider limit or use `fill_gas_params`? That method will fill both prices and the gas limit to the sender's actual gas allowance based on their balance.

monad-rpc/src/handlers/eth/call.rs
haythemsellami: Yeah can access Triedb through `&ChainState<T>`, I just found that triedb is used only in `call.rs` and was like why adding another dependency. 

Are there any specific upsides from using `chain_state: &ChainState<T>` ?

monad-rpc/src/handlers/eth/call.rs
haythemsellami: I'm not sure we want to use `fill_gas_params` as it will revert if sender have 0 balance. I guess we want `eth_fillTransaction` to fill missing tx object field without a strict verification.

monad-rpc/src/handlers/eth/call.rs
andr-dev: Yeah, chain state can use data cached from execution events so it can reduce triedb pressure + endpoint latency. Eventually, we'll probably move to a model where triedb isn't even exposed so it always tries grabbing things from memory first.

haythemsellami:
Should we also get this merged @philipglazman @andr-dev ?

monad-rpc/src/handlers/eth/call.rs
philipglazman: I'm not sure what the desired behavior is wrt to how strict the method should be, so will leave that to you.

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -1002,6 +1006,220 @@ pub async fn monad_admin_ethCallStatistics(
...
+    {
+        let priority_fee = suggested_priority_fee().await.unwrap_or_default();
+        tx.gas_price_details = GasPriceDetails::Eip1559 {
+            max_fee_per_gas: *max_fee_per_gas,
+            max_priority_fee_per_gas: Some(U256::from(priority_fee)),
+        };
+    }
+
+    // Fill gas prices (max_fee_per_gas, max_priority_fee_per_gas, or gas_price)
+    let base_fee = U256::from(header.header.base_fee_per_gas.unwrap_or_default());
```
philipglazman: We assign the last voted block's base fee, but the next block's base fee might be different because of dynamic base fees. We might want to consider adding some buffer for better ux

philipglazman (commented):
Looks good to me! Left some comments and suggest we use `chain_state` instead of `triedb_env` directly

monad-rpc/src/handlers/eth/call.rs
haythemsellami: Added 1.5x buffer.

haythemsellami:
> Looks good to me! Left some comments and suggest we use `chain_state` instead of `triedb_env` directly

fixed.

monad-rpc/src/handlers/eth/call.rs
haythemsellami: I kept it as it is, I think reth also do not revert or check for user balance.

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -1003,6 +1007,226 @@ pub async fn monad_admin_ethCallStatistics(
...
+    let base_fee = U256::from(header.base_fee_per_gas.unwrap_or_default());
+    let fill_base_fee = match &tx.gas_price_details {
+        GasPriceDetails::Eip1559 { .. } => base_fee.saturating_mul(U256::from(3)) / U256::from(2),
+        _ => base_fee,
+    };
+    tx.fill_gas_prices(fill_base_fee)?;
+
+    // Estimate gas if not specified
+    let gas_specified = tx.gas.is_some();
+    if !gas_specified {
```
eerkaijun: nit: maybe just do tx.gas.is_none()?

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -1003,6 +1007,226 @@ pub async fn monad_admin_ethCallStatistics(
...
+        let gas_estimator = GasEstimator::new(
+            chain_id,
+            header.clone(),
+            from,
+            block_key,
+            StateOverrideSet::default(),
+            false, // gas not specified by user
+        );
+
+        let Quantity(estimated_gas) = estimate_gas(
```
eerkaijun: since we call into execution eth call, I think we should also limit concurrent requests to this endpoint. 

in eth_call and eth_estimate endpoint we have something like this to limit concurrent requests
```
let _permit = match app_state.rate_limiter.try_acquire() {
        Ok(permit) => permit,
        Err(_) => {
            if let Some(tracker) = &app_state.eth_call_stats_tracker {
                tracker.record_queue_rejection().await;
            }
            return Err(JsonRpcError::internal_error(
                "eth_estimateGas concurrent requests limit".into(),
            ));
        }
    };
```

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -1003,6 +1007,226 @@ pub async fn monad_admin_ethCallStatistics(
...
+    let gas_limit: u64 = tx
+        .gas
+        .unwrap_or_default()
+        .try_into()
+        .map_err(|_| JsonRpcError::invalid_params())?;
+    let value = tx.value.unwrap_or_default();
+    let input = tx.input.input.clone().unwrap_or_default();
+    let to = tx.to;
+
+    let raw_bytes = match &tx.gas_price_details {
```
eerkaijun: we have `impl TryFrom<CallRequest> for TxEnvelope` - can that be reused?

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -1003,6 +1007,226 @@ pub async fn monad_admin_ethCallStatistics(
...
+
+    let from = tx.from.ok_or_else(JsonRpcError::invalid_params)?;
+
+    if tx.value.is_none() {
+        tx.value = Some(U256::from(0));
+    }
+
+    tx.chain_id = Some(U64::from(chain_id));
+
+    // Get header from chain_state (uses buffer cache if available)
```
andr-dev: Let's remove these comments, they're just stating what the code is doing.

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -1003,6 +1007,226 @@ pub async fn monad_admin_ethCallStatistics(
...
+            .await
+            .map_err(JsonRpcError::internal_error)?;
+        tx.nonce = Some(U64::from(account.nonce));
+    }
+
+    // Fill max_priority_fee_per_gas with suggested fee if not specified (EIP-1559 only)
+    if let GasPriceDetails::Eip1559 {
+        max_fee_per_gas,
+        max_priority_fee_per_gas: None,
+    } = &tx.gas_price_details
```
andr-dev: Can we match here and show the other variants explicitly? My thinking here is that there _could_ be a 7702 variant that would use 1559 semantics, I checked the enum def and there isn't but good for future proofing.

monad-rpc/src/handlers/eth/gas.rs
```diff
@@ -43,15 +43,15 @@ use crate::{
...
+pub(super) trait EthCallProvider {
     async fn eth_call(
         &self,
         txn: TxEnvelope,
         eth_call_executor: Option<Arc<EthCallExecutor>>,
     ) -> CallResult;
 }
 
-struct GasEstimator {
+pub(super) struct GasEstimator {
```
andr-dev: Instead of making this pub, why don't we put the endpoint in this file?

monad-rpc/src/handlers/eth/call.rs
haythemsellami: fixed.

monad-rpc/src/handlers/eth/call.rs
haythemsellami: Seems like `TryFrom<CallRequest> for TxEnvelope` injects a dummy signature, while eth_fillTransaction needs unsigned signable bytes... I kept it as it is for now

monad-rpc/src/handlers/eth/gas.rs
haythemsellami: I implemented the handler in `call.rs` because it mostly does request normalization and unsigned-tx construction; if we want to reduce coupling later, the cleaner refactor I guess is to expose one narrow gas-estimation helper from gas.rs instead of moving the whole endpoint...

I'll keep it as it is for now unless u strongly disagree, lmk.

monad-rpc/src/handlers/eth/call.rs
haythemsellami: done

monad-rpc/src/handlers/eth/call.rs
haythemsellami: done

monad-rpc/src/handlers/eth/call.rs
haythemsellami: Applied the same concurrency gate.

monad-rpc/src/handlers/eth/gas.rs
haythemsellami: refactored it in gas.rs

haythemsellami:
> would be nice to add a test in monad-integration

Already have them https://github.com/category-labs/monad-integration/pull/394, should pass once we merge this.

JoeHowarth (approved):

PR #2664, opened 2025-12-10
Pre tfm base fee cleanup

eerkaijun:
as part of the reserve balance cleanup and refactor, removing pre tfm base fee which is not used anymore

### DIFF
--- a/monad-eth-block-policy/src/lib.rs
+++ b/monad-eth-block-policy/src/lib.rs
@@ -181,9 +181,9 @@ struct CommittedBlock {
     timestamp_ns: u128,
     fees: BlockTxnFeeStates,
 
-    base_fee: Option<u64>,
-    base_fee_trend: Option<u64>,
-    base_fee_moment: Option<u64>,
+    base_fee: u64,
+    base_fee_trend: u64,
+    base_fee_moment: u64,
     block_gas_usage: u64,
 }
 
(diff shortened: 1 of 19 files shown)


monad-executor-glue/src/lib.rs
```diff
@@ -1236,36 +1208,9 @@ where
...
-                    2 => Some(u64::decode(&mut base_fee_moment_payload)?),
-                    _ => {
-                        return Err(alloy_rlp::Error::Custom(
-                            "failed to decode unknown base_fee_moment in mempool event",
-                        ))
-                    }
-                };
+                let base_fee = u64::decode(&mut payload)?;
+                let base_fee_trend = u64::decode(&mut payload)?;
+                let base_fee_moment = u64::decode(&mut payload)?;
```
eerkaijun, line 1224: this breaks the base fee wal encoding & decoding for mempool event pre upgrade, but think it's a good change rather than keeping it

monad-eth-block-policy/src/lib.rs
```diff
@@ -1283,9 +1238,7 @@ where
         let validator = EthBlockPolicyBlockValidator::new(
             block.get_seq_num(),
             self.execution_delay,
-            block
-                .get_base_fee()
-                .unwrap_or(monad_tfm::base_fee::PRE_TFM_BASE_FEE),
```
michael-yxchen, line 1288: PRE_TFM_BASE_FEE constant can be removed

PR #2670, opened 2025-12-12
Use QC event for latest voted in RPC

andr-dev:
Previously, the execution event for a QC was sent before the triedb voted metadata was updated, causing a race condition where RPC websockets would push a new head and subsequently querying something for triedb would fail.

After https://github.com/category-labs/monad/pull/1964 (which has been updated in `monad-bft` in a separate PR), we can go back to using the buffer (populated from execution events) for the latest block number. This will reduce user latency and put less pressure on triedb on top of the execution fix writing the QC before executing the proposal which should also further reduce latency.

### DIFF
--- a/monad-rpc/src/chainstate/mod.rs
+++ b/monad-rpc/src/chainstate/mod.rs
@@ -101,11 +101,11 @@ impl<T: Triedb> ChainState<T> {
     }
 
     pub fn get_latest_block_number(&self) -> u64 {
-        // Return triedb's latest block number.
-        // There is a race condition between buffer and triedb for common wallet workflows.
-        // For example, a wallet might call `eth_getBalance` after calling `eth_getBlockByNumber`
-        // and the balance might not be updated in triedb yet.
-        self.triedb_env.get_latest_voted_block_key().seq_num().0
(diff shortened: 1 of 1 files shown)


omegablitz:
I don't think this is sound - any rpc that hits triedb_env before get_latest_voted_block_key is updated will fail.
It's possible for triedb_env to lag the buffer

omegablitz:
before:
```
getLatest            <- triedbenv <- triedb
<all other rpcs>     <- triedbenv
```
after:
```
getLatest             <- chainstate <- triedb
<all other rpcs>      <- triedbenv <- triedb
```

So after, getlatest and then calling another rpc can fail.
Before, it couldn't, because everything returned from getLatest is always accessible in triedbenv

andr-dev:
Triedb metadata is a memory mapped region so I assumed we were reading it directly from memory. Instead, triedb_env polls the region every `5ms` and caches the value thus introducing the issue mentioned above. There's good value in changing this and reading directly since it'll shave off `2.5ms` on average for every triedb related request which is a UX win. I will resolve that separately in another PR and link it here once it's ready.

andr-dev:
Being handled in separate PR chain.

PR #2675, opened 2025-12-16
Latest blocktag uses proposed blocks

philipglazman:


monad-rpc/src/chainstate/buffer.rs
```diff
@@ -77,10 +80,21 @@ impl ChainStateBuffer {
             _ => return,
         };
 
-        if block_event.commit_state != BlockCommitState::Voted {
-            if block_event.commit_state == BlockCommitState::Finalized {
-                let height = block_event.data.header.number;
-                self.finalized.fetch_max(height, Ordering::SeqCst);
+        if block_event.commit_state != BlockCommitState::Proposed {
+            match block_event.commit_state {
```
andr-dev: Let's add a separate branch in the commit state match for proposed that doesn't return

monad-triedb-utils/src/triedb_env.rs
```diff
@@ -679,9 +707,25 @@ impl TriedbEnv {
...
+            let triedb_handle: TriedbHandle = TriedbHandle::try_new(triedb_path, node_lru_max_mem)
+                .expect("triedb should exist in path");
+            (
+                SeqNum(triedb_handle.latest_proposed_block().unwrap_or_default()),
+                BlockId(Hash(
+                    triedb_handle.latest_proposed_block_id().unwrap_or_default(),
+                )),
+            )
+        };
+
```
andr-dev: Why do we create two different triedb_handles? This should be merged into a single block.

monad-triedb-utils/src/triedb_env.rs
```diff
@@ -271,9 +291,15 @@ fn polling_thread(
                     }
                 }
 
+                if proposed_is_updated {
+                    last_proposed = latest_proposed;
+                    populate_cache(&tokio_handle, &triedb_handle, meta.clone(), latest_proposed);
+                }
```
andr-dev, line 311: Should the `voted_is_updates` branch still be checking if it's a `BlockKey::Proposed` variant? Similarly, should we have something similar for the `latest_proposed`?

omegablitz:
This needs to be addressed as a prerequisite: https://github.com/category-labs/category-internal/issues/1968

Full list of deps: https://github.com/category-labs/category-internal/issues/2017

monad-rpc/src/chainstate/buffer.rs
```diff
@@ -99,11 +121,18 @@ impl ChainStateBuffer {
...
                 ?block_height,
-                "ChainStateBuffer received block event for existing block height"
+                old_hash = ?old_block.header.hash,
+                new_hash = ?block_hash,
+                "ChainStateBuffer received block event for existing block height, replacing old block"
             );
+
+            // Remove old block's hash from by_hash
+            self.block_height_by_hash
+                .remove(&FixedData(old_block.header.hash.0));
```
eerkaijun, line 135: do we also need to clear `self.tx_loc_by_hash` here? I guess there could be some transactions that were included in the old block but not in the new block

monad-triedb-utils/src/triedb_env.rs
```diff
@@ -245,11 +249,27 @@ fn polling_thread(
                 // retry in case of a race
             }
 
+            let mut latest_proposed = last_proposed;
+            for _ in 0..3 {
+                if let (Some(proposed_block_num), Some(proposed_block_id)) = (
+                    triedb_handle.latest_proposed_block(),
+                    triedb_handle.latest_proposed_block_id(),
```
eerkaijun: these are two non atomic calls, I think there can be a race condition. for e.g. in `get_latest_voted_block_key`, we check for consistency

monad-rpc/src/chainstate/buffer.rs
```diff
@@ -136,17 +165,6 @@ impl ChainStateBuffer {
...
-            warn!(
-                ?voted_block_height,
-                event_block_height = block_height,
-                "ChainStateBuffer received voted block event with lower height than existing voted block height"
-            );
-            return;
-        }
-
         let mut block_heights = self.block_heights.lock().await;
         block_heights.push_front(block_height);
```
eerkaijun, line 176: does it matter if there are two blocks with the same seqnum in `block_heights`?

monad-triedb-utils/src/triedb_env.rs
```diff
@@ -674,14 +709,25 @@ impl TriedbEnv {
...
-        let latest_finalized = {
+        let (latest_finalized, latest_proposed_seq_num, latest_proposed_block_id) = {
             let triedb_handle: TriedbHandle = TriedbHandle::try_new(triedb_path, node_lru_max_mem)
                 .expect("triedb should exist in path");
-            SeqNum(triedb_handle.latest_finalized_block().unwrap_or_default())
+            (
+                SeqNum(triedb_handle.latest_finalized_block().unwrap_or_default()),
+                SeqNum(triedb_handle.latest_proposed_block().unwrap_or_default()),
+                BlockId(Hash(
+                    triedb_handle.latest_proposed_block_id().unwrap_or_default(),
```
eerkaijun: also potential race condition i think

monad-rpc/src/chainstate/buffer.rs
andr-dev: I'm addressing this in a separate change, will post soon.

monad-rpc/src/chainstate/buffer.rs
philipglazman: Yea we should. Fixing and adding test.

monad-triedb-utils/src/triedb_env.rs
```diff
@@ -674,14 +716,22 @@ impl TriedbEnv {
...
+                .unwrap_or(BlockKey::Finalized(latest_finalized));
+            (latest_finalized, latest_proposed)
         };
+
         let meta = Arc::new(Mutex::new(TriedbEnvMeta {
-            latest_finalized: FinalizedBlockKey(latest_finalized),
-            latest_voted: BlockKey::Finalized(FinalizedBlockKey(latest_finalized)),
+            latest_finalized,
+            latest_voted: BlockKey::Finalized(latest_finalized),
+            latest_proposed,
```
omegablitz: For simplicity, I would just initialize latest_proposed to latest_finalized, similar to how it's done for latest_voted. You can then delete the diff at the start of this function for fetching latest_proposed.

monad-rpc/src/chainstate/buffer.rs
eerkaijun: i think right now it can cause some inconsistency with the eviction logic, i.e. during eviction we do `self.block_by_height.remove(&evicted_block_height)` which then clear the `tx_loc_by_hash` and `block_height_by_hash` based on the "old" block

eerkaijun: i'm not sure whether this matters in practice, but seems like a race condition where the transactions are cleared while the block height still exists.

PR #2688, opened 2025-12-29
Archive: add exists(key) to kvstore

JoeHowarth:
Replaces previous store.scan_prefix with exact match key. 
That impl was `O(log(N))` which works for quite a while, but eventually degrades after tens of millions of keys.

### DIFF
--- a/monad-archive/src/kvstore/mod.rs
+++ b/monad-archive/src/kvstore/mod.rs
@@ -147,6 +147,10 @@ pub trait KVStore: KVReader {
 #[enum_dispatch]
 pub trait KVReader: Clone {
     async fn get(&self, key: &str) -> Result<Option<Bytes>>;
+
+    /// Lightweight existence check; prefer this over `get` when you don't need payload bytes.
+    async fn exists(&self, key: &str) -> Result<bool>;
+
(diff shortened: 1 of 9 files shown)


monad-archive/src/kvstore/mod.rs
```diff
@@ -147,6 +147,9 @@ pub trait KVStore: KVReader {
 #[enum_dispatch]
 pub trait KVReader: Clone {
     async fn get(&self, key: &str) -> Result<Option<Bytes>>;
+
+    async fn exists(&self, key: &str) -> Result<bool>;
```
mikety, line 152: nit. a comment of `get` vs `exists` may be useful to clarify why not to use `get` all the times (because its expensive)

monad-archive/src/kvstore/mongo.rs
```diff
@@ -226,6 +226,20 @@ impl KVReader for MongoDbStorage {
         }
     }
 
+    async fn exists(&self, key: &str) -> Result<bool> {
+        let start = Instant::now();
+        let count = self
+            .collection
```
mikety, line 234: nit: `findOne` with `_id` projection is simpler and more efficient

mikety (approved):
ok to merge after "mongo" nit is addressed.

PR #2692, opened 2025-12-30
monad-wireauth: parametrize metrics

dshulyak:
wireauth will be used for both TCP, UDP (raptorcast and separate point to point) and we need separate
metrics , and since we don't support labels we just generate different instances of metrics


ariqchowdhury (approved):

dshulyak:
@omegablitz @michael-yxchen rebased, please reaprove

PR #2695, opened 2025-12-31
monad-wireauth: add message buffering during handshake with limits

dshulyak:
for tcp workload we need to buffer message instead of falling back to original tcp protocol.
so if session is not established, we initiate a session and buffer message, once session completes
buffered messages are sent to a peer immediately

for safety we enforce that there are atmost 1000 concurrent initiated sessions and each session can buffer atmost 128KB of data

this functionality will be also useful for some udp workloads when we disable unauthenticated udp,
for example discovery ping/pong will have to be buffered, instead of dropping the message

### DIFF
--- a/monad-wireauth/README.md
+++ b/monad-wireauth/README.md
@@ -140,6 +140,13 @@ session_decrypt         time:   [166.11 ns 168.75 ns 171.20 ns]
 | `monad.wireauth.enqueued.cookie_reply` | cookie challenges added to outbound queue |
 | `monad.wireauth.enqueued.keepalive` | keepalive packets added to outbound queue |
 
+### initiator buffering
+
+| metric | description |
+|--------|-------------|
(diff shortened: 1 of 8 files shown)


xinyuan-dev (approved):

dshulyak:
Run cargo update --workspace --locked
error: failed to load manifest for workspace member `/__w/monad-bft/monad-bft/monad-execution`
referenced via `monad-*` by workspace at `/__w/monad-bft/monad-bft/Cargo.toml`

Caused by:
  failed to read `/__w/monad-bft/monad-bft/monad-execution/Cargo.toml`

Caused by:
  No such file or directory (os error 2)
Error: Process completed with exit code 101.

michael-yxchen:
Should be addressed after #2782 merges. https://github.com/category-labs/monad-bft/pull/2782/changes#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R4

PR #2698, opened 2026-01-05
Seed blocksync RNG from entropy

michael-yxchen:
closes: https://github.com/category-labs/category-internal/issues/1711


omegablitz:
Is mock-swarm still deterministic with this?

PR #2700, opened 2026-01-06
Archive: rework bft archiving and indexing

JoeHowarth:
The previous way bft blocks were archived was unsatisfactory in 2 ways:
1. There was no `seq_num => block_id` index, so blocks were effectively unqueriable
2. The key-space did not follow the on-disk ledger/headers ledger/bodies directory structure, causing confusion and making laying down blocks for re-execution more difficult.

Unlike archiving execution outputs, the ledger folder is keyed by hash (block_id / body_id) so we must walk back from the committed_head to ensure we're creating an index and archive without gaps.

Additionally, we must create an index and migrate data from the existing bft archive.

This work is best understood commit by commit


monad-archive/src/bin/monad-indexer/main.rs
```diff
@@ -53,28 +57,39 @@ async fn main() -> Result<()> {
             archive_sink,
             async_backfill,
         }) => run_set_start_block(block, archive_sink, async_backfill).await,
+        Some(cli::Commands::MigrateBftIndex {
+            source,
+            sink,
+            concurrency,
+            batch_size,
+            no_copy_data,
+        }) => run_migrate_bft_index(source, sink, concurrency, batch_size, !no_copy_data).await,
```
mikety: Why double negative? `!no_copy_data` instead of `copy_data` ?

monad-archive/src/bin/monad-indexer/migrate_bft_archive.rs
```diff
@@ -0,0 +1,357 @@
...
+            .scan_prefix_with_max_keys(LEGACY_BFT_BLOCKS_PREFIX, concurrency * 20)
+            .await?;
+
+        info!(?candidates, "Found {} candidates", candidates.len());
+
+        let candidates = candidates.into_iter().into_iter().filter_map(|key| {
+            info!("Candidate key: {}", key);
+            let hex_str = key
+                .strip_prefix(LEGACY_BFT_BLOCKS_PREFIX)?
+                .strip_suffix(".header")?;
```
mikety, line 251: nit:  ".body" suffixes will be counted even if filtered out - suggest to increase "20" multiplier to account for that.

monad-archive/src/bin/monad-indexer/migrate_bft_archive.rs
```diff
@@ -0,0 +1,357 @@
...
+use monad_types::{BlockId, Hash};
+use serde::{Deserialize, Serialize};
+
+const LEGACY_BFT_BLOCKS_PREFIX: &str = "bft_block/";
+const BFT_INDEX_MARKERS_PREFIX: &str = "bft/index_markers/";
+const RETRY_DELAY: Duration = Duration::from_millis(500);
+const MAX_STARTUP_RETRIES: usize = 10;
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+struct IndexedRangeMarker {
```
mikety, line 41: a comment about range markers and their purpose will help, also `end_id` is less than `start_id` since its parent by design. Suggest change it to `head_id`, `tail_id`

monad-archive/src/bin/monad-indexer/migrate_bft_archive.rs
```diff
@@ -0,0 +1,357 @@
...
+                break;
+            }
+
+            let (header, _) = self.fetch_legacy_header_by_id(&candidate).await?;
+            match self.find_committable_head(header).await {
+                Ok(Some(committable_id)) => {
+                    info!("Found committable head {:?}", committable_id);
+                    let (committable_header, _) =
+                        self.fetch_legacy_header_by_id(&committable_id).await?;
+                    known_committable_heads.insert(IndexedRangeMarker {
```
mikety: is it used anywhere as a range? if it is a  `seq_num` -> `block_id` index entry it should be changed accordingly?

monad-archive/src/bin/monad-indexer/migrate_bft_archive.rs
```diff
@@ -0,0 +1,357 @@
...
+            info!(
+                "Indexing sub-chain from num: {}, to num: {}",
+                marker.end_num,
+                marker.end_num.saturating_sub(max_num_per_batch as u64)
+            );
+            for _ in 0..max_num_per_batch {
+                if other_sub_chain_tips.contains(&marker.end_id) {
+                    return Ok(());
+                }
+
```
mikety, line 367: should it have a condition that this is already reindexed by checking the new index as well? 
`self.model.get_id_by_num(marker.end_num).await.is_ok() `

mikety (commented):
solid overall - a few nits in the comments

monad-archive/src/bin/monad-indexer/migrate_bft_archive.rs
```diff
@@ -0,0 +1,357 @@
...
+                let current_id = marker.end_id;
+                let (next_id, next_num) = retry_forever(
```
mikety: add - check if parent exists: - return if `self.fetch_legacy_header_by_id(&current_id).await.is_err()`  - it will stuck at the genesis block otherwise

monad-archive/src/bin/monad-archiver/ledger_archiver_and_indexer.rs
```diff
@@ -0,0 +1,235 @@
...
+};
+
+const UPLOAD_CONCURRENCY: usize = 10;
+const BFT_BLOCK_BODY_FILE_PATH: &str = "bodies/";
+/// Threshold for keeping header bytes in memory during scan.
+/// Above this, we discard bytes to bound memory and re-read during upload.
+const HEADER_BYTES_MEMORY_THRESHOLD: usize = 10_000;
+
+struct BftUploadItem {
+    num: u64,
```
mikety: `seq_num` instead of `num` -> self documenting

monad-archive/src/bin/monad-archiver/ledger_archiver_and_indexer.rs
```diff
@@ -0,0 +1,235 @@
...
+
+            Ok((item, header_bytes, body_bytes))
+        })
+        .buffered(UPLOAD_CONCURRENCY)
+        .filter_map(|result: Result<(BftUploadItem, Vec<u8>, Vec<u8>)>| async {
+            match result {
+                Ok(item) => Some(item),
+                Err(e) => {
+                    error!("Failed to read bft block from disk: {e:?}");
+                    None
```
mikety: if read fails it will cause gaps later in the upload

PR #2701, opened 2026-01-06
Consensus: Improve liveness when advancing round using TC from timeout messages

xinyuan-dev:
This PR improves networking liveness when advancing round using a TC from a timeout message:

- when advancing to the next round, immediately timeout locally and broadcast the timeout
- when advancing to a future round, broadcast advance round message with the TC

This ensures honest validators can either receive a TC or locally form a TC to enter the new round.

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -746,10 +746,37 @@ where
             cmds.extend(handle_vote_cmds);
         }
 
+        let current_round = self.consensus.pacemaker.get_current_round();
         match &timeout.last_round_certificate {
-            Some(RoundCertificate::Tc(tc)) => {
+            Some(RoundCertificate::Tc(tc)) if tc.round == current_round => {
                 self.metrics.consensus_events.remote_timeout_msg_with_tc += 1;
+
+                // broadcast Timeout message immediately if received
+                // TC to advance round. this helps other validators to
+                // form their own TC
+                cmds.extend(self.handle_timeout_expiry(tc.round));
                 cmds.extend(self.process_tc(tc));
+            }
+            Some(RoundCertificate::Tc(tc)) if tc.round > current_round => {
+                self.metrics
+                    .consensus_events
+                    .remote_timeout_msg_with_future_tc += 1;
+
+                // broadcast AdvanceRound message with the TC for
+                // skipped round. this helps other validators to
+                // advance their rounds.
+                //
+                // Note: cannot broadcast a Timeout message for due
+                // to the lack of required round certificate for the
+                // skipped round.
+                let round_certificate = RoundCertificate::Tc(tc.clone());
+                let messages = self.broadcast_advance_round_message(round_certificate);
+
+                cmds.extend(messages);
+                cmds.extend(self.process_tc(tc));
+            }
+            Some(RoundCertificate::Tc(_)) => {
+                // ignore TC from past rounds
             }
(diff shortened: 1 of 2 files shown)


monad-consensus-state/src/lib.rs
```diff
@@ -779,13 +806,46 @@ where
...
+        };
+
+        let message = ConsensusMessage {
+            message: ProtocolMessage::AdvanceRound(AdvanceRoundMessage {
+                last_round_certificate,
+            }),
+            version: self.version,
+        };
+        let command = ConsensusCommand::Publish {
+            target: RouterTarget::Broadcast(epoch),
```
dshulyak, line 842: is it intentionally broadcast and not raptorcast? i think with raptorcast target it will write less data, but otherwise will be functionally equivalent.

do you know if there are any other assumptions? maybe document them here for future

xinyuan-dev: AdvanceRoundMessage here functions similarly to the TimeoutMessage, which already uses broadcast. I don't think we need distribute the chunks based on stake.

also, the TC is not as large as a proposal. I'd expect the message to fit in one or two chunk typically, or at most ~10s of chunks (assuming a ~20KiB TC in worst case). For small messages raptorcast has an overhead proportional to the number of validators due to rounding chunks, so I'm not sure if it will actually save much data.

dshulyak: so broadcast is sending same data (moreover with redundancy factor which is a mistake in this case) to every peer? that seems to have higher overhead than any overhead that raptorcast can introduce. 

i run a script to check how much data is generated based on different variables (msg size, validator set)

  | Message | Validators | Broadcast              | Raptorcast          |
  |---------|------------|------------------------|---------------------|
  | 2 KiB   | 100        | 495 pkts (712 KiB)     | 99 pkts (142 KiB)   |
  | 20 KiB  | 100        | 4,257 pkts (5.98 MiB)  | 99 pkts (142 KiB)   |
  | 2 KiB   | 1,000      | 4,995 pkts (7.01 MiB)  | 999 pkts (1.40 MiB) |
  | 20 KiB  | 1,000      | 42,957 pkts (60.3 MiB) | 999 pkts (1.40 MiB) |

dshulyak: it is probably fine to use broadcast, as the traffic to low to observe any issues, but i think it needs to be revisited based on plans to support more validators

dshulyak (approved):

monad-consensus-state/src/lib.rs
xinyuan-dev: I don't think in reality the traffic saving matters much as this TC is only broadcasted on a relatively rare scenario (consecutive timeouts) and only when the validator is lagging and needing to skip rounds, which is even rarer. In addition, raptorcast requires two hops, which we need further evaluation for its robustness, here I think lower latency broadcast is probably preferred especially because this message is meant for a liveness improvement. In addition, the implementation here is consistent with other broadcasted messages, namely RoundRecoveryMessage and TimeoutMessage.

dshulyak: to me it seemed that 60MB vs 1.4MB (for 1000 validators and 20KB tc message) makes it obvious that broadcast won't work, as it will require 0.5s to send all the data to peers. and in all my simulations raptorcast propagation is faster than that even with larger payloads, so i am quite confident that raptorcast will have lower latency as well

PR #2704, opened 2026-01-07
Remove consecutive seqnum assertion

omegablitz:


monad-eth-block-policy/src/lib.rs
```diff
@@ -1300,13 +1300,6 @@ where
     ) -> Result<(), BlockPolicyError> {
         let chain_id = chain_config.chain_id();
 
-        let first_block = extending_blocks
-            .iter()
-            .chain(std::iter::once(&block))
-            .next()
-            .unwrap();
-        assert_eq!(first_block.get_seq_num(), self.last_commit + SeqNum(1));
```
dshulyak, line 1308: so there is no need for checking this at all? invalidate with error for example?

what would be the expected behavior if one adversarial validator proposes block with u64::MAX seqnum?

dshulyak: looks like this is checked in another place

dshulyak (dismissed):

monad-eth-block-policy/src/lib.rs
omegablitz: Yes it's checked elsewhere

xinyuan-dev (dismissed):

monad-mock-swarm/tests/nonces.rs
```diff
@@ -649,4 +664,27 @@ mod test {
...
+                    ByzantineConfig::default()
+                }
+            });
+
+            while swarm
+                .step_until(&mut UntilTerminator::new().until_block(20))
+                .is_some()
+            {}
+
+            swarm_ledger_verification(&swarm, 2);
```
michael-yxchen: nit: we can bump the min ledger length to 18 or 19 with swarm terminating at block 20

Issue #1956, opened 2025-07-07
rpc socket watcher: potential locking when (re)connecting to an IPC socket


robertkiel:
The socket watcher that is used for the `TxPool` IPC connection in the RPC module spawns a blocking future on the current thread. This is an issue because the Tokio runtime explicitly asks the developer to spawn [blocking futures](https://docs.rs/tokio/latest/tokio/index.html#cpu-bound-tasks-and-blocking-code) using `spawn_blocking` to prevent the runtime from running out of worker threads.

For the RPC endpoint, this is slightly more severe because it is started with exactly [2 worker threads](https://github.com/category-labs/monad-bft/blob/4287361bfbe642f15847c92101d7a7347fee3853/monad-rpc/src/main.rs#L47) which means there is only one "backup" thread to pick up the task that receives the `SocketWatcher` event through the channel.

```rust
let mut watcher = notify::INotifyWatcher::new(
    move |res| {
        // spawns a blocking task
        block_on(async {
            if let Err(send) = tx.send(res).await {
                error!("cannot send on socket watcher channel: {:?}", send);
            };
        })
    },
    notify::Config::default(),
)
```

### Suggestion

Why don't clone the `tx` endpoint beforehand and then pass the cloned instance to `INotifyWatcher` and run all of it as a Tokio `task`?

```rust
let cloned_tx = tx.clone();

let mut watcher = notify::INotifyWatcher::new(
    move |res| {
        tokio::spawn(async {
            if let Err(send) = cloned_tx.send(res).await {
                error!("cannot send on socket watcher channel: {:?}", send);
            };
        }))
    },
    notify::Config::default(),
)
```

PR #2705, opened 2026-01-08
Cleanup txpool bridge reconnection logic

andr-dev:
The `EthTxPoolBridgeHandle` runtime code was grafted onto an existing socket watching approach that was convoluted and involved complex state across poll calls. This change simplifies the watching logic for new txpool sockets along with tests. Additionally, it removes the `notify` dependency which helps cut back on several other transient dependencies. Finally, this change also resolves https://github.com/category-labs/monad-bft/issues/1956.

### Testing

Added unit tests to the `SocketWatcher` utility and `EthTxPoolBridge` demonstrating various behaviors. The change was also tested on `devnet7` by manually restarting `monad-bft` several times and observing successful reconnections.

### DIFF
--- a/monad-rpc/src/txpool/mod.rs
+++ b/monad-rpc/src/txpool/mod.rs
@@ -16,26 +16,22 @@
 use std::{
     io,
     path::{Path, PathBuf},
-    pin::Pin,
-    task::Poll,
     time::Duration,
 };
 
 use alloy_consensus::TxEnvelope;
 use flume::Receiver;
-use futures::{ready, Future, Sink, SinkExt, Stream, StreamExt};
+use futures::{SinkExt, StreamExt};
 use monad_eth_txpool_ipc::{EthTxPoolIpcClient, EthTxPoolIpcTx};
-use monad_eth_txpool_types::{EthTxPoolEvent, EthTxPoolSnapshot};
-use pin_project::pin_project;
 use state::TxStatusReceiverSender;
-use tokio::pin;
-use tracing::warn;
+use tracing::{debug, error, info, warn};
 
 pub use self::{client::EthTxPoolBridgeClient, handle::EthTxPoolBridgeHandle, types::TxStatus};
 use self::{
     socket::SocketWatcher,
     state::{EthTxPoolBridgeEvictionQueue, EthTxPoolBridgeState},
 };
+use crate::txpool::socket::SocketWatcherEvent;
 
 mod client;
 mod handle;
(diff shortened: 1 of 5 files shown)


monad-rpc/src/txpool/mod.rs
```diff
@@ -102,35 +89,115 @@ impl EthTxPoolBridge {
...
         cleanup_timer.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
 
-        let err = loop {
+        let err = 'err: loop {
+            let ipc_client = match self.ipc_client.as_mut() {
+                Some(ipc_client) => ipc_client,
+                None => loop {
+                    let Some(result) = self.socket_watcher.next().await else {
+                        error!("TxPoolBridge SocketWatcher died while attempting to reconnect");
+                        break 'err None;
```
JoeHowarth: run() is doing too much for 1 function - the nesting is confusing and the overall flow is unclear for someone reading this without context.

The outer loop body should be it's own fn and convert the break 'err into return Err(..). I'd also consider putting this inner loop into another fn as well - you have 2 options, either add comments or make fn names that serve as comments

monad-rpc/src/txpool/mod.rs
```diff
@@ -102,127 +89,476 @@ impl EthTxPoolBridge {
...
+                                ?event,
+                                "TxPoolBridge detected socket delete event while trying to reconnect"
+                            );
+                        }
+                    }
+                },
+            };
+
             tokio::select! {
+                biased;
```
JoeHowarth, line 105: A 50 line select w/ 4 branches is right on the edge for me. Much more than this and I prefer to make each arm it's own handler and have the select serve as a router. This isn't strictly necessary here since each branch has pretty simple logic, but you may consider it fr the tx_receiver branch - up to you

monad-rpc/src/txpool/mod.rs
```diff
@@ -103,126 +90,479 @@ impl EthTxPoolBridge {
...
+            let ipc_client = match self.ipc_client.as_mut() {
+                Some(ipc_client) => ipc_client,
+                None => match self.try_create_ipc_client().await {
+                    Err(err) => break err,
+                    Ok(()) => continue,
+                },
+            };
+
             tokio::select! {
+                biased;
```
philipglazman, line 105: The `biased` might starve tx_receiver and the other branches since we expect high volume from ipc_client

andr-dev: This was intentional, while there are txpool events we should process them to
1. prioritize returning existing requests
2. clear up the txpool buffer so txpool can send more events which then allows it to process more txs

PR #2711, opened 2026-01-12
Raptorcast: Refactor ValidatorSet types and types with ST generics

xinyuan-dev:
This PR contains two separate refactories:

- unify the `ValidatorSet` type used in `util::EpochValidators` and `decoding::ValidatorSet`
- change unnecessary `ST` generics to `PT`

Rationale for unifying `ValidatorSet` type: The `monad_validator::validator_set::ValidatorSet` type semantically guarantee that the validator set is valid, which is an essential assumption in many parts of Raptorcast, but not obvious in types like `EpochValidators` or `decoding::ValidatorSet`. Unifying the validator set types allows us to clear up duplicated instances of the same concept and preclude unsound usages like `impl Default for EpochValidators`.

Rationale for converting `ST` to `PT`: A type with a generic `ST` implies that it contains a signature (or something related to a signature, like a keypair), which is not actually true for most types depending on `ST`. For example, types like `EpochValidators`/`Group` actually only contains a set of node ids, which only depends on a public key. Moreover, the usage of ST is contagious - e.g. `BuildTarget` must depend on `ST` because `ValidatorsView` uses `ST`. The refactory is expected to clean up the excessive usages of `ST`.

### DIFF
--- a/monad-peer-discovery/src/lib.rs
+++ b/monad-peer-discovery/src/lib.rs
@@ -468,7 +468,7 @@ impl<ST: CertificateSignatureRecoverable> MonadNameRecord<ST> {
     pub fn with_pubkey(
         &self,
         pubkey: CertificateSignaturePubKey<ST>,
-    ) -> MonadNameRecordWithPubkey<ST> {
+    ) -> MonadNameRecordWithPubkey<'_, ST> {
         MonadNameRecordWithPubkey {
             record: self,
             pubkey,
--- a/monad-raptorcast/benches/encode_bench.rs
+++ b/monad-raptorcast/benches/encode_bench.rs
@@ -28,6 +28,7 @@ use monad_raptorcast::{
 use monad_secp::SecpSignature;
 use monad_testutil::signing::get_key;
 use monad_types::{Epoch, NodeId, Stake};
+use monad_validator::validator_set::ValidatorSet;
 
 const NUM_NODES: usize = 100;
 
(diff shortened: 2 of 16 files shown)


michael-yxchen (dismissed):

xinyuan-dev:
> We seem to be making behavioral changes (ValidatorSet::new_unchecked, get_members() calls) in this refactor

I don't think there is any behavioral change - the original `EpochValidators` and `decoding::ValidatorSet` all have exactly the same type (`BTreeMap<NodeId, Stake>`) as `ValidatorSet` (except for the additional `total_stake` value which is not really used in this refactory). The `new_unchecked` method was added to signify that it's the caller's responsibility to ensure the invariance assumed of the type, but other than the semantic change, the usage is actually [identical](https://github.com/category-labs/monad-bft/pull/2711/changes#diff-a7a0c418338285b73aa56490b3f74e7fad73f870617ea93822f8cb78b2a6b7e9R798).

Initially I was only unifying the `ValidatorSet` types, but then it just pulled on more strings and brought a cascade of changes to the generics - that's why this PR is like this. I agree it's easier to look at the two set of changes separately, let me split it up.

UPDATE: done. See https://github.com/category-labs/monad-bft/pull/2711/commits/61fefac89db44a1b01b3eea396d1c4d0958245b7 and https://github.com/category-labs/monad-bft/pull/2711/commits/38c0baaedfa9ed2dbae4c17885b9cf1520c0b304.

PR #2715, opened 2026-01-14
Archive: fix mongo replica names in checker

JoeHowarth:
The archive checker would key replicas by the replica_name given by ArchiveArgs, which for mongo db was simply db_name. Since all the db names are typically the same, this meant that only 1 mongo replica could be checked at a time.

This change does 2 things:

1. make block data reader and archive args `replica_name()` impls the same for mongo storages, notably including the redacted mongo url along with the db name
2. Add an optional `--replica-name` arg for mongo url strings that overrides this

### DIFF
--- a/monad-archive/src/cli.rs
+++ b/monad-archive/src/cli.rs
@@ -264,13 +264,7 @@ impl BlockDataReaderArgs {
         match self {
             Aws(aws_cli_args) => aws_cli_args.bucket.clone(),
             Triedb(trie_db_cli_args) => trie_db_cli_args.triedb_path.clone(),
-            MongoDb(mongo_db_cli_args) => {
-                format!(
-                    "{}:{}",
-                    redact_mongo_url(&mongo_db_cli_args.url),
-                    mongo_db_cli_args.db
-                )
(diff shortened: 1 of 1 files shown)


monad-archive/src/cli.rs
```diff
@@ -711,8 +708,17 @@ impl MongoDbCliArgs {
...
+            replica_name: kv.remove("replica-name"),
         })
     }
+
+    /// Returns the replica name for this MongoDB connection.
+    /// Uses explicit replica_name if set, otherwise derives from redacted URL and db name.
+    pub fn replica_name(&self) -> String {
+        self.replica_name
+            .clone()
+            .unwrap_or_else(|| format!("{}:{}", redact_mongo_url(&self.url), self.db))
```
mikety, line 720: this approach mixes 2 separare concepts:
- mongdb replica set name
- replica in monad sense, i.e. the data source. 
I think it will be more clear to separate those so the monad_replica is either mongo rsname or the derivative of the connection string.

ariqchowdhury (approved):

PR #2718, opened 2026-01-15
Make txpool tx input stream generic

andr-dev:
This change enables creating txpool executors without starting up the txpool IPC server making testing/fuzzing simpler.


monad-eth-txpool-executor/src/lib.rs
```diff
@@ -223,7 +225,7 @@ where
     }
 }
 
-impl<ST, SCT, SBT, CCT, CRT> EthTxPoolExecutor<ST, SCT, SBT, CCT, CRT>
+impl<ST, SCT, SBT, CCT, CRT> EthTxPoolExecutor<ST, SCT, SBT, CCT, CRT, EthTxPoolIpcServer>
```
robertkiel: Any reason for using the concrete type and not a generic here?

andr-dev: Good catch, this one can use the generic.

PR #2721, opened 2026-01-15
Raptorcast: recognize unspecified broadcast mode

xinyuan-dev:
This is a small refactory to support the upcoming protocol dispatcher feature.

### DIFF
--- a/monad-raptorcast/src/packet/assembler.rs
+++ b/monad-raptorcast/src/packet/assembler.rs
@@ -32,7 +32,7 @@ use super::{
 };
 use crate::{
     udp::GroupId,
-    util::{compute_hash, Redundancy},
+    util::{compute_hash, BroadcastMode, Redundancy},
     SIGNATURE_SIZE,
 };
 
@@ -600,16 +600,10 @@ fn encode_symbols<PT: PubKey>(
     Ok(())
 }
 
-pub(crate) enum BroadcastType {
-    Secondary,
-    Primary,
-    Unspecified,
-}
-
 // return the shared header for all chunks sans the signature.
 pub(crate) fn build_header(
     version: u16,
-    broadcast_type: BroadcastType,
+    broadcast_mode: BroadcastMode,
     merkle_tree_depth: u8,
     group_id: GroupId,
     unix_ts_ms: u64,
@@ -633,10 +627,10 @@ pub(crate) fn build_header(
 
     let (cursor_broadcast_merkle_depth, cursor) =
         cursor.split_at_mut_checked(1).expect("header to short");
-    let mut broadcast_byte: u8 = match broadcast_type {
-        BroadcastType::Primary => 0b10 << 6,
-        BroadcastType::Secondary => 0b01 << 6,
-        BroadcastType::Unspecified => 0b00 << 6,
+    let mut broadcast_byte: u8 = match broadcast_mode {
+        BroadcastMode::Primary => 0b10 << 6,
+        BroadcastMode::Secondary => 0b01 << 6,
+        BroadcastMode::Unspecified => 0b00 << 6,
     };
     // tree_depth max 4 bits
     if (merkle_tree_depth & 0b1111_0000) != 0 {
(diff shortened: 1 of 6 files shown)


monad-raptorcast/src/packet/assembler.rs
```diff
@@ -633,10 +627,10 @@ pub(crate) fn build_header(
...
     let (cursor_broadcast_merkle_depth, cursor) =
         cursor.split_at_mut_checked(1).expect("header to short");
-    let mut broadcast_byte: u8 = match broadcast_type {
-        BroadcastType::Primary => 0b10 << 6,
-        BroadcastType::Secondary => 0b01 << 6,
-        BroadcastType::Unspecified => 0b00 << 6,
+    let mut broadcast_byte: u8 = match broadcast_mode {
+        BroadcastMode::Primary => 0b10 << 6,
+        BroadcastMode::Secondary => 0b01 << 6,
+        BroadcastMode::Unspecified => 0b00 << 6,
```
andr-dev, line 633: Let's make this an associated function of `BroadcastMode`.

monad-raptorcast/src/packet/builder.rs
```diff
@@ -454,13 +454,13 @@ where
...
 where
     PT: PubKey,
 {
     match build_target {
-        BuildTarget::Raptorcast { .. } => BroadcastType::Primary,
-        BuildTarget::FullNodeRaptorCast { .. } => BroadcastType::Secondary,
-        _ => BroadcastType::Unspecified,
+        BuildTarget::Raptorcast { .. } => BroadcastMode::Primary,
+        BuildTarget::FullNodeRaptorCast { .. } => BroadcastMode::Secondary,
+        _ => BroadcastMode::Unspecified,
```
andr-dev: Can we specify each of the remaining `BuildTarget` variants explicitly? This forces the compiler to immediately flag to people when adding a new `BuildTarget` so they can change this branch as well.

monad-raptorcast/src/udp.rs
```diff
@@ -469,12 +471,12 @@ where
...
-        (true, false) => Some(BroadcastMode::Primary),
-        (false, true) => Some(BroadcastMode::Secondary),
-        (false, false) => None, // broadcast or unicast
+    let broadcast_mode = match (broadcast, secondary_broadcast) {
+        (true, false) => BroadcastMode::Primary,
+        (false, true) => BroadcastMode::Secondary,
+        (false, false) => BroadcastMode::Unspecified, // unicast or broadcast
         (true, true) => {
-            return Err(MessageValidationError::InvalidBroadcastBits);
+            return Err(MessageValidationError::InvalidBroadcastBits(0b11));
```
andr-dev, line 477: What does `0b11` mean, should this be a constant?

monad-raptorcast/src/packet/assembler.rs
xinyuan-dev: I think it's best not have this an associated function because the way the broadcast mode is encoded can be tied to the protocol (e.g. raptorcast versions/specific header layout), which is why the assembler and parser should be the one holding this information, instead of the modes themselves, which can stay abstract.

monad-raptorcast/src/udp.rs
xinyuan-dev: It's the the two 1 bits (`(true, true)`), which is currently unused and considered invalid.

monad-raptorcast/src/packet/builder.rs
xinyuan-dev: good point, fixed in fe3762af146f69e461c662ce553e6865eaf3a2f5.

monad-raptorcast/src/udp.rs
```diff
@@ -342,9 +342,11 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
                     // TODO: cap rebroadcast symbols based on some multiple of esis.
                     try_rebroadcast_symbol();
 
-                    if let Some(mode) = parsed_message.maybe_broadcast_mode {
-                        self.metrics
-                            .record_broadcast_latency(mode, parsed_message.unix_ts_ms);
+                    if !matches!(parsed_message.broadcast_mode, BroadcastMode::Unspecified) {
```
andr-dev: This no longer needs to be wrapped in a `matches` because `record_broadcast_latency` internally matches on the `mode`.

andr-dev (approved):

PR #2722, opened 2026-01-15
monad-dataplane: add per-peer byte limit for TCP egress queue

dshulyak:
closes: https://github.com/category-labs/category-internal/issues/1557

in addition to 150 message limit per connection we enforce 4MB byte limit, thats 2 max block sizes and slightly more than 2 max statesync responses. we set upper bound for outoging connections, in addition to incoming, just as a safety measure. and reduce the size of the global response queue from 1024 to 256, to reduce potential memory pressure there, that queue is relevant only if we do a burst of responses larger than the queue


dshulyak:
@xinyuan-dev @ariqchowdhury please reapprove, fixed conflicts

PR #2729, opened 2026-01-16
Rate limit for handling peer lookup request

eerkaijun:
Closes https://github.com/category-labs/category-internal/issues/2086

This rate limit doesn't need to be configurable since it doesn't affect a node's ability to discover peers, easier to just set a hardcoded default

### DIFF
--- a/monad-peer-discovery/src/discovery.rs
+++ b/monad-peer-discovery/src/discovery.rs
@@ -40,15 +40,17 @@ use crate::{
     ipv4_validation::{IpCheckError, validate_socket_ipv4_address},
 };
 
-type RateLimiter = governor::RateLimiter<
+type RateLimiter<C> = governor::RateLimiter<
     governor::state::NotKeyed,
     governor::state::InMemoryState,
-    governor::clock::QuantaClock,
-    governor::middleware::NoOpMiddleware<governor::clock::QuantaInstant>,
+    C,
+    governor::middleware::NoOpMiddleware<<C as governor::clock::Clock>::Instant>,
 >;
 
 /// Maximum number of peers to be included in a PeerLookupResponse
 const MAX_PEER_IN_RESPONSE: usize = 16;
+/// Rate limit for incoming peer lookup requests (per second)
+const PEER_LOOKUP_RATE_LIMIT_PER_SECOND: u32 = 10;
 /// Number of peers to send lookup request to
 const NUM_LOOKUP_PEERS: usize = 3;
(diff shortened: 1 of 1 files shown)


monad-peer-discovery/src/discovery.rs
```diff
@@ -2004,51 +2026,89 @@ mod tests {
...
+            assert!(check_accepted(&cmds), "request {i} should be accepted");
+        }
+
+        // next request should be rate limited and dropped
+        let cmds = send_request(state, rate_limit + 1);
+        assert_eq!(cmds.len(), 0, "request should be rate limited and dropped");
+        assert_eq!(state.metrics[GAUGE_PEER_DISC_RATE_LIMITED], 1);
+
+        // advance time by 1 second
+        std::thread::sleep(Duration::from_secs(1));
```
michael-yxchen: https://docs.rs/governor/latest/governor/_guide/index.html#constructing-a-direct-rate-limiter

```
let clock = FakeRelativeClock::default();
RateLimiter::direct_with_clock(Quota::per_second(nonzero!(50u32)), clock);
```

this might be helpful for tests

eerkaijun: i think it serves slightly different purposes though, here i added a few integration tests that treat the rate limiter as a black box and made sure requests are dropped accordingly. if we use a fake timer we will be testing the accounting logic of the governor rate limiter works correctly, which I think we probably don't need to test that

michael-yxchen: I think that FakeRelativeClock allows you to manually advance clocks without sleeping in tests

eerkaijun: yeah right but I mean that is really testing a different thing, i.e. we set up a rate limiter with FakeRelativeClock, then manually advancing the clock is just testing the accounting logic of this mock governor impl

dshulyak: i am also not following, how is it different from sleep? you are using os as source of time, and adjusting it with sleep. with fake clock the source of time is moved into memory and can be adjusted more efficiently

frankly 1s is too much time for tests, it is generally needs to be avoided

eerkaijun: ok sure i updated it

michael-yxchen: sry I wasn't clear before. I was making more of a general comment that we should avoid sleep in unit tests when possible

PR #2735, opened 2026-01-21
Remove trivially-updatable unmaintained dependencies

eclark0426:
Removes direct and indirect dependencies that can be updated with minimal changes to the codebase. There will be more to come, but will require more significant changes


ariqchowdhury (dismissed):

eclark0426:
Not sure what was up with the build pipeline failing a WASM check, but recreating the lockfile on the first failing commit seems to have resolved it locally

ariqchowdhury (approved):

andr-dev:
Suggesting bumping `reqwest` to `0.12.28` to use the same `reqwest` version across all crates, as it is now we're using both `0.12` and `0.13`. Version `>=0.13` also pulls in a bunch of java and windows deps, likely a mistake by crate author.

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -906,17 +906,20 @@ pub async fn monad_createAccessList<T: Triedb + TriedbPath>(
...
-    let v: serde_cbor::Value = if raw_payload.is_empty() {
-        serde_cbor::Value::Array(vec![])
+    let access_list: AccessList = if raw_payload.is_empty() {
+        AccessList::default()
     } else {
-        serde_cbor::from_slice(&raw_payload)
-            .map_err(|e| JsonRpcError::internal_error(format!("cbor decode error: {}", e)))?
+        let v: ciborium::Value = ciborium::de::from_reader(raw_payload.as_slice())
+            .map_err(|e| JsonRpcError::internal_error(format!("cbor decode error: {}", e)))?;
+        serde_json::from_value(serde_json::to_value(v).map_err(|e| {
```
andr-dev: This looks pretty bad from a performance standpoint, first deserialize into cbor Value, then turn into JSON Value, then deserialize. I haven't used `ciborium` before, can we just deserialize straight from the raw_payload to the type instead of going through `ciborium::Value`?

eclark0426: To summarize our Slack DMs here: I agree this is subpar, but because of the types used in different places, this was the most straightforward solution. For now, we're omitting the final ciborium commit and will work out the best solution separately.

andr-dev: @eclark0426 and I agreed to move this to another PR.

andr-dev (approved):

PR #2744, opened 2026-01-26
Persist voted_head in ledger

philipglazman:


monad-block-persist/src/lib.rs
```diff
@@ -198,6 +203,14 @@ where
         Ok(())
     }
 
+    fn update_voted_head(&mut self, block_id: &BlockId) -> std::io::Result<()> {
+        let mut wip = PathBuf::from(&self.voted_head_path);
+        wip.set_extension(".wip");
+        std::os::unix::fs::symlink(self.header_path(block_id), &wip)?;
+        std::fs::rename(&wip, &self.voted_head_path)?;
+        Ok(())
+    }
```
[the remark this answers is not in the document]
philipglazman: Don't need Windows support

monad-consensus-state/src/lib.rs
```diff
@@ -1200,6 +1200,19 @@ where
         self.metrics.consensus_events.process_qc += 1;
 
         let mut cmds = Vec::new();
+
+        // Emit Voted commit for the highest coherent block on the path to the QC block.
+        let qc_block_id = qc.get_block_id();
+        if let Some(voted_block) = self
```
omegablitz, line 1206: I think it's possible that the same block gets emitted as voted multiple times here in a row. I don't think this breaks anything but just something to keep in mind.

omegablitz (approved):
ledger-tail could utilize this too

omegablitz:
Another thing to be wary of is that proposed_head is not guaranteed to be on the same chain as voted_head. This could cause issues in execution depending on how the proposed_head/voted_head following is implemented there.

In short, either consensus needs to guarantee that proposed_head is on the same chain, or execution needs to know not to execute proposed_head if it's not on the same chain.

monad-consensus-state/src/lib.rs
philipglazman: Yep, it can be emitted multiple times - but I don't think that causes issues

PR #2745, opened 2026-01-27
Raptorcast: move utilities to util module

xinyuan-dev:
In this PR:

- move `Collector`, `PeerAddrLookup`, `Recipient`, `UdpMessage` from the `packet` module into the `util` module.
- add helper types that implement the traits: `EmptyCollector`, `BlackholeCollector`, `KnownSocketAddr` that will be used later.
- add a few tests

These traits/types are expected to be useful beyond the packet builder module, so it makes better sense to put them in the shared `util.rs` file.

### DIFF
--- a/monad-raptorcast/src/util.rs
+++ b/monad-raptorcast/src/util.rs
@@ -14,16 +14,19 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 use std::{
+    cell::OnceCell,
     cmp::Ordering,
     collections::{BTreeMap, HashSet},
     fmt,
     net::SocketAddr,
+    rc::Rc,
 };
 
+use bytes::Bytes;
 use fixed::{types::extra::U11, FixedU16};
 use itertools::Itertools;
 use monad_crypto::{
-    certificate_signature::PubKey,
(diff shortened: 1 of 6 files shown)


monad-raptorcast/src/util.rs
```diff
@@ -679,6 +682,222 @@ impl<PT: PubKey> ReBroadcastGroupMap<PT> {
...
+#[derive(Clone, Eq)]
+struct RecipientInner<PT: PubKey> {
+    node_id: NodeId<PT>,
+    node_hash: OnceCell<[u8; 20]>,
+    addr: OnceCell<Option<SocketAddr>>,
+}
+
+impl<PT: PubKey> PartialEq for RecipientInner<PT> {
+    fn eq(&self, other: &Self) -> bool {
+        self.node_hash == other.node_hash
```
michael-yxchen: should this be `self.node_hash() == other.node_hash()` since `node_hash` is lazily computed now?

xinyuan-dev: You're right. In fact, `RecipientInner` shouldn't implement `PartialEq` on its own, as it was meant to be a property of `Recipient` only.

PR #2746, opened 2026-01-27
Raptorcast: refactor dual udp packet sender

xinyuan-dev:
This commit refactors `send`/`send_with_record` functions with a new versatile `Collector` type `DualUdpPacketSender` that can be used directly on a packet builder.

Note, the new `DualUdpPacketSender` removes batching support for improved latency and removed coupling with the hard-coded batch size, which should be considered an implementation detail of the packet builder.

### DIFF
--- a/monad-raptorcast/src/lib.rs
+++ b/monad-raptorcast/src/lib.rs
@@ -14,7 +14,7 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 use std::{
-    collections::{BTreeMap, HashMap, VecDeque},
+    collections::{BTreeMap, HashMap, HashSet, VecDeque},
     future::Future as _,
     marker::PhantomData,
     net::{IpAddr, SocketAddr, SocketAddrV4},
@@ -33,7 +33,7 @@ use message::{InboundRouterMessage, OutboundRouterMessage};
 use monad_crypto::{
     certificate_signature::{
         CertificateKeyPair, CertificateSignature, CertificateSignaturePubKey,
-        CertificateSignatureRecoverable,
+        CertificateSignatureRecoverable, PubKey,
     },
     signing_domain,
 };
@@ -62,7 +62,8 @@ use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender};
 use tracing::{debug, debug_span, error, trace, warn};
 use udp::GroupId;
 use util::{
-    BuildTarget, EpochValidators, FullNodes, Group, PeerAddrLookup, ReBroadcastGroupMap, Redundancy,
+    BuildTarget, Collector, EpochValidators, FullNodes, Group, KnownSocketAddr, PeerAddrLookup,
+    ReBroadcastGroupMap, Recipient, Redundancy, UdpMessage,
 };
 
 use crate::{
@@ -336,10 +337,12 @@ where
         &mut self,
         outbound_msg: SecondaryOutboundMessage<CertificateSignaturePubKey<ST>>,
     ) {
-        let Some(secondary_mb) = self.secondary_message_builder.as_mut() else {
+        let Some(builder) = self.secondary_message_builder.as_mut() else {
             error!("secondary_message_builder not configured");
             return;
         };
+
+        let mut sink = DualUdpPacketSender::new(&mut self.dual_socket, &self.peer_discovery_driver);
 
         match outbound_msg {
             SecondaryOutboundMessage::SendSingle {
@@ -353,15 +356,11 @@ where
                     "raptorcastprimary handling single message from secondary"
                 );
                 let build_target = BuildTarget::PointToPoint(&dest);
-                send(
-                    &mut self.dual_socket,
-                    &self.peer_discovery_driver,
-                    secondary_mb,
-                    &msg_bytes,
-                    &build_target,
-                    UdpPriority::Regular,
-                    group_id,
-                );
+                builder
+                    .prepare()
+                    .group_id(group_id)
+                    .build_into(&msg_bytes, &build_target, &mut sink)
+                    .unwrap_log_on_error(&msg_bytes, &build_target)
             }
             SecondaryOutboundMessage::SendToGroup {
                 msg_bytes,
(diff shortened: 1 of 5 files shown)


monad-raptorcast/src/lib.rs
```diff
@@ -1214,26 +1209,27 @@ where
...
                                 &mut this.dual_socket,
                                 &this.peer_discovery_driver,
-                                &mut this.message_builder,
-                                &router_message,
-                                UdpPriority::Regular,
-                                &target,
-                                &name_record,
-                            );
+                            )
+                            .with_target_addr(&target, SocketAddr::V4(name_record.udp_socket()));
```
dshulyak, line 1216: this doesn't change the behavior, right? as i understand it will use chained lookup in dual packet sender, if wireauth returns something it will use that, and otherwise fallback to target_addr 

also there was an inconsistency there and fixed in https://github.com/category-labs/monad-bft/pull/2751 , looks like i will just add another implementation for lookup that uses (dual_socket,auth_addr) tuple

monad-raptorcast/src/lib.rs
```diff
@@ -1357,79 +1353,108 @@ where
...
+impl<'a, ST, PD, AP> Drop for DualUdpPacketSender<'a, ST, PD, AP>
+where
     ST: CertificateSignatureRecoverable,
+    AP: auth::AuthenticationProtocol<PublicKey = CertificateSignaturePubKey<ST>>,
     PD: PeerDiscoveryAlgo<SignatureType = ST>,
+{
+    fn drop(&mut self) {
+        let targets = self.targets.iter().map(|r| r.node_id());
+        ensure_authenticated_sessions(self.dual_socket, self.peer_disc_driver, targets);
+    }
```
dshulyak, line 1431: i am not sure if this is the right use of drop. also once we are ready to switch to remove non-auth we will always want to establish sessions before sending any messages, so it is also not very future proof either

dshulyak: and the other thing it is also should not use discovery for establishing sessions when we use name record. if you want to use drop i guess i will change with_target_addr to with_target_name_record and then i will do switch based on that in the followup here https://github.com/category-labs/monad-bft/pull/2751

monad-raptorcast/src/lib.rs
xinyuan-dev: > this doesn't change the behavior, right?

Yes, `impl PeerAddrLookup for (PL1, PL2)` works in a fallback manner - first try PL1, then PL2.

> also there was an inconsistency there and fixed in https://github.com/category-labs/monad-bft/pull/2751

I suggest we postpone PR after this one, and we can design a combinator or wrapper for that specific lookup mechanism. It needs to be implemented in a different way as it's not a simple chained lookup.

monad-raptorcast/src/lib.rs
xinyuan-dev: > i am not sure if this is the right use of drop.

Drop may not run in safe rust only if the value is leaked via `mem::forget` or with `panic=abort`. I don't think we need to worry about those cases in normal operation.

> once we are ready to switch to remove non-auth we will always want to establish sessions before sending any messages

I agree, during the refactory of this part I do find the `establish_authenticated_sessions` a bit absurd since I expected the authenticated session to establish first before we start sending data. If I understand correctly, this is to avoid delaying the current transmission and only open sessions for future uses. For this PR I hope to not modify existing behavior.

> and the other thing it is also should not use discovery for establishing sessions when we use name record

Yes I think it requires a different design as the peer lookup trait in raptorcast currently don't fit very well. To be specific, the current peer lookup is only about returning an address, it has no knowledge of the type of address, that's why it's convoluted to make it fit with wireauth. Ultimately, for a packet to be sent via wireauth I'm expecting we can get move the socket address lookup to DualUdpSocket and only use the recipient's node id to specify the destinations. But again, I don't think we should introduce such behavior changes here.

dshulyak: > Drop may not run in safe rust only if the value is leaked via mem::forget or with panic=abort. I don't think we need to worry about those cases in normal operation.

i meant that it is just weird that we establish sessions in drop, sure it works, but why to use it at all like that? i would rather just have a function that does it

```
send()
establish_sessions()
```

drop on the other hand makes sense when it restores correct state or performs cleanup

monad-raptorcast/src/lib.rs
dshulyak: are you sure that this refactoring is required though? after referenced pr the code in send and send_with_record won't share that much code at all, and i will just have to adjust DualPacketSender to have additional switch to handle send_with_record properly

monad-raptorcast/src/lib.rs
dshulyak: >  If I understand correctly, this is to avoid delaying the current transmission and only open sessions for future uses. For this PR I hope to not modify existing behavior.

yes that was the reason. i am not asking to modify behavior. to me it seems that it makes future changes more complicated that they have to be, which is the opposite goal of the refactoring (i think)

monad-raptorcast/src/lib.rs
xinyuan-dev: I think send and send_with_record are conceptually doing the same thing, it's just the way we lookup the address is different. That's why I think the rerfactory is sound.

Also, IIUC most of the NameRecordLookup, RefCell complications are there to get around the borrowing issues with UdpMessageBatcher. I think DualPacketSender would be a cleaner substitute of that, especially if we deem batching unnecessary.

dshulyak: ok, lets re-examine when i will rebase https://github.com/category-labs/monad-bft/pull/2751

monad-raptorcast/src/lib.rs
xinyuan-dev: I see what you mean. This pattern is indeed not very typical, but it's not that unusual (e.g. DropTimer, RebroadcastGuard). I see the usage here as doing some extra works that are called when the sender is out of scope, sort of like doing flushing. Also, we may consider the `establish_sessions` step "best-effort". Also, do we consider not calling `establish_sessions` a flaw? Adding the explicit call would imply such possibility, but Drop can preclude that interpretation. 

Beyond those reasons, I think having an explicit `establish_sessions` method does make sense, although I personally still prefer the Drop style at least for this specific place. I'm happy to change it if you still prefer the more explicit way.

PR #2747, opened 2026-01-27
Raptorcast: Signature verifier

xinyuan-dev:
This PR combines the rate limiter and signature verifier into a unified type called `SignatureVerifier` and integrates it into the `udp::parse_message` function.

This type captures the concept of signature verification, with optional caching and optional rate limiting.

The rate limiter bypassing mechanism is now reduced into a simpler predicate. This callback can be further reduced to a single boolean value once the message parser is specialized to each protocol in follow up commits.

### DIFF
--- a/monad-raptorcast/src/parser/signature_verifier.rs
+++ b/monad-raptorcast/src/parser/signature_verifier.rs
@@ -1,0 +1,262 @@
+// Copyright (C) 2025 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+use std::{hash::Hash, marker::PhantomData, num::NonZeroUsize};
+
+use governor::{DefaultDirectRateLimiter, Quota, RateLimiter};
+use lru::LruCache;
+use monad_crypto::certificate_signature::{
+    CertificateSignaturePubKey, CertificateSignatureRecoverable,
+};
+use monad_types::NodeId;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SignatureVerifierError {
+    /// The verifier's rate limit was exceeded for this key or source.
+    RateLimited,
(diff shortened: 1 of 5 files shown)


monad-raptorcast/src/parser/signature_verifier.rs
```diff
@@ -0,0 +1,256 @@
...
+    where
+        SD: monad_crypto::signing_domain::SigningDomain,
+    {
+        let author_pubkey = signature
+            .recover_pubkey::<SD>(data)
+            .map_err(|_| MessageValidationError::InvalidSignature)?;
+        let author = NodeId::new(author_pubkey);
+
+        Ok(author)
+    }
```
michael-yxchen, line 129: Does it make sense to always try save to cache after successful verification? and load from cache before rate limit checks

xinyuan-dev: That was my initial design actually. However, if we merge the caching and verification together, then `udp::parse_message` wouldn't be able to lazily call the rate limit bypass callback depending on whether cache is used (or it must be passed as a callback to SignatureVerifier). The slight performance implication is probably fine, but I think it's more straightforward in the current style.

On the other hand, the exposed methods, namely `verify`, `load_cache`, `save_cache` do make sense as valid standalone concepts on their own, despite there are no actual usage at the moment (e.g. verifying without caching, checking the cache without verifying).

michael-yxchen: cool gotcha

michael-yxchen (dismissed):

dshulyak (dismissed):

monad-raptorcast/src/parser/signature_verifier.rs
```diff
@@ -0,0 +1,256 @@
...
+            rate_limiter: None,
+            _signing_domain: PhantomData,
+        }
+    }
+
+    pub fn with_cache(mut self, size: usize) -> Self
+    where
+        CacheKey: Eq + Hash + Clone,
+    {
+        let cache_size = NonZeroUsize::new(size).expect("cache size must be non-zero");
```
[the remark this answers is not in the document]
xinyuan-dev: that's a valid good point!

monad-raptorcast/src/parser/signature_verifier.rs
```diff
@@ -0,0 +1,256 @@
...
+use std::{hash::Hash, marker::PhantomData, num::NonZeroUsize};
+
+use governor::{DefaultDirectRateLimiter, Quota, RateLimiter};
+use lru::LruCache;
+use monad_crypto::certificate_signature::{
+    CertificateSignaturePubKey, CertificateSignatureRecoverable,
+};
+use monad_types::NodeId;
+
+use crate::udp::MessageValidationError;
```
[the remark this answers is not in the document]
xinyuan-dev: nice advice!

michael-yxchen (approved):

PR #2748, opened 2026-01-27
spec for tx ingestion/forwarding

dshulyak:


tklenze:
I just had a conversation with Dmitry. I generally agree with this proposal. Some open questions / comments:

### Lack of backpressure
Senders (RPCs) do not know how many transactions they are allowed to send to the receiver (leader). This is an issue if the leader's bandwidth becomes a limiting factor.
- **Congestion feedback loop**. If the receiving leader experiences congestion (either through natural demand, spam, or DoS) and drops packets, the incentive for rational RPCs is to send more traffic to the receiver. For example, if there is 20% packet loss at the networking layer, then an RPC has the incentive to send duplicates of its transactions (or at least its most valuable ones), to ensure that all of them are received by the validator. Thus, congestion is a positive feedback loop.
- **Solution**. A possible solution is go give backpressure to senders, and give them explicit limits on the amount of traffic that they are allowed to send. If the authenticated traffic originating from a source exceeds its limit, the receiver drops the excess traffic, and can furthermore punish the sender for misbehavior (for example, but dropping all of its transactions). This breaks the feedback loop.

I think it would be good to build confidence that a backpressure system can be incorporated into the design later on.

### Sybill incentive
The proposal's `max_gas_contribution` limits the amount of priority that any one RPC can get. This creates an incentive for large RPC operators to artificially split up their RPC into many smaller RPCs, which may have undesirable implications. Proposal: remove the `max_gas_contribution`, but use a Exponential Moving Average to count the more recent gas contribution more. Then, we can also remove the extra decay factor in the calculation.

```
# Constants
block_time = 0.4  # seconds
half_life = 86400 / block_time  # set to 24 hours. This is just a proposal...
decay_factor = 0.5^(1 / half_life)

# On each new block:
gas_contribution = decay_factor * gas_contribution + new_block_gas_contribution
time_weight = min((time_known / time_weight_unit)^2, max_time_weight) # to prevent rapid new IP address spawning
score = gas_contribution × time_weight
```
Notice that the decay here takes into consideration "recent contribution rate" rather than "time since last seen." The EMA replaces both the gas_contribution cap and the decay factor, since it handles both concerns.

docs/0004_tx_ingestion.md
```diff
@@ -0,0 +1,232 @@
...
+
+| Property | Priority Pool | Regular Pool |
+|----------|---------------|--------------|
+| Bandwidth share | 90% | 10% |
+| Scheduling | Weighted by score | Equal weights |
+| Max size | 100,000 | 100,000 |
+
+Incoming transactions are routed based on authentication status and score. Authenticated identities with scores at or above `promotion_threshold` enter the priority pool; all others enter the regular pool.
+
+The bandwidth split between pools defaults to 90/10. During dequeue, a counter cycles to probabilistically select which pool to serve, achieving the configured ratio over time.
```
0xSqualo: Nit: The cycling counter achieves the 90/10 ratio over time, but the spec should clarify behavior when one pool is empty. If the priority pool is empty (e.g., right after a node restart before any peers have scores), does a priority dequeue attempt fall back to the regular pool, or does it return nothing? If it's work-conserving (falls back to the other pool), a brief note here could make that explicit.

docs/0004_tx_ingestion.md
```diff
@@ -0,0 +1,232 @@
...
+finish_time(i) = virtual_time + (1 / score(i))
+```
+
+When a transaction is dequeued, the global virtual time advances to the finish time of the selected identity, and that identity's finish time is recalculated for its next queued transaction.
+
+The algorithm guarantees that over time, each identity receives bandwidth proportional to its score. An identity with score 10 receives 10x the bandwidth of an identity with score 1.
+
+### Priority Pool Fallback
+
+When the priority pool cannot accept a transaction, promoted identities fall back to the regular pool rather than being rejected. This provides a safety mechanism if scoring is abused to block new identities.
```
0xSqualo, line 147: Nit: What's the behavior when both pools are at capacity - drop incoming transactions? Might be worth a one-liner to make it explicit.

docs/0004_tx_ingestion.md
```diff
@@ -0,0 +1,232 @@
...
+### Two-Tier Identity Storage
+
+To prevent identity table eviction attacks, the scorer maintains two separate pools:
+
+| Pool | Capacity | Entry Condition | Eviction |
+|------|----------|-----------------|----------|
+| Promoted | 90,000 | score ≥ promotion_threshold | LRU within pool |
+| Newcomers | 10,000 | score < promotion_threshold | LRU within pool |
+
+New identities enter the newcomers pool. When score crosses the promotion threshold, the identity moves to the promoted pool. Flooding with new identities only evicts other newcomers; promoted peers remain protected.
```
dnkolegov-ar: Is it possible to be transferred from the Promoted pool to the Newcomers pool?

docs/0004_tx_ingestion.md
```diff
@@ -0,0 +1,232 @@
...
+    - [Reassembly Pool Eviction](#reassembly-pool-eviction)
+  - [Impact on Other Components](#impact-on-other-components)
+  - [Security Analysis](#security-analysis)
+    - [Naive Flooding](#naive-flooding)
+    - [Mempool Monopolization](#mempool-monopolization)
+    - [Graceful Degradation](#graceful-degradation)
+
+## Rationale
+
+The current implementation processes transactions in arrival order without identity-based prioritization. This creates two vulnerabilities:
```
dnkolegov-ar, line 29: Is the goal to protect the nodes themselves or the entire network? What are the assumptions in the security mechanisms?  Are there any presuppositions that the nodes are incentivised to follow and enforce these rules?

docs/0004_tx_ingestion.md
```diff
@@ -0,0 +1,232 @@
...
+The scoring model intentionally avoids negative reputation (penalties for invalid transactions or spam). This keeps the model simple. Penalties may be introduced in future iterations if the positive-only approach proves insufficient against observed attack patterns.
+
+### Two-Tier Identity Storage
+
+To prevent identity table eviction attacks, the scorer maintains two separate pools:
+
+| Pool | Capacity | Entry Condition | Eviction |
+|------|----------|-----------------|----------|
+| Promoted | 90,000 | score ≥ promotion_threshold | LRU within pool |
+| Newcomers | 10,000 | score < promotion_threshold | LRU within pool |
```
dnkolegov-ar, line 110: What if Newcomers or Promoted is overflowed?

0xSqualo:
Clarification: The spec introduces `AuthTxIngestion` as a new port separate from the existing `AuthenticatedUDP` used by raptorcast. Who is expected to establish wireauth sessions on this port - only validators/consensus peers, or any node with a keypair (e.g., RPCs, fullnodes)? 

This affects the threat model significantly: if only validators can connect, then scoring tracks validators as relayers; if anyone can connect, scoring tracks RPCs directly. A sentence on the intended session audience would help.

msetrin:
```
# Constants
block_time = 0.4  # seconds
half_life = 86400 / block_time  # set to 24 hours. This is just a proposal...
decay_factor = 0.5^(1 / half_life)

# On each new block:
gas_contribution = decay_factor * gas_contribution + new_block_gas_contribution
time_weight = min((time_known / time_weight_unit)^2, max_time_weight) # to prevent rapid new IP address spawning
score = gas_contribution × time_weight
```

Something seems off about this formulation.  It seems like you can get the full time weight benefit just from having connected X minutes before sending any traffic - why can't an attacker just spawn a bunch of IPs an hour before doing any sort of short-term attack?  It seems like the higher time_weight should come with some additional requirement on the total contribution to achieve the full benefit.

tklenze:
> Something seems off about this formulation. It seems like you can get the full time weight benefit just from having connected X minutes before sending any traffic - why can't an attacker just spawn a bunch of IPs an hour before doing any sort of short-term attack? It seems like the higher time_weight should come with some additional requirement on the total contribution to achieve the full benefit.

The score is `score = gas_contribution × time_weight`, and `gas_contribution` is initially 0 (actually, it makes sense to have a floor on the gas contribution, e.g., `score = max(gas_contribution, gas_contribution_floor) × time_weight`).

The `time_weight` is just designed to prevent DoS by quickly cycling through new IP addresses. I did not touch this, it is part of the original proposal. (I am not sure how important this is.)

msetrin:
Yeah sorry I didn't mean to suggest it was a point specific to one of the two formulations (I just picked the easier one to quote).  I guess I still have skepticism about both formulations to some extent.  Just thinking out loud:
- I agree with the rough intuition that there should be a boost based on the duration of the connection, but it seems like there needs to be some sort of penalty that needs to be overcome to realize it (like some target contribution that is a function of the time that needs to be overcome to get a boost).
- Something feels missing in that we aren't tracking the actual bandwidth utilization anywhere (maybe for simplicity).  I feel like intuitively tracking the contribution per unit bandwidth with some confidence component that scales as a function of total bandwidth and/or time could work.

docs/0004_tx_ingestion.md
dshulyak: documented both goals

docs/0004_tx_ingestion.md
dshulyak: we evict based on lru within each pool, for promoted also comparing score of the newly promoted to the one that will be evicted. maybe for newcomes i should add some time-based protection

docs/0004_tx_ingestion.md
dshulyak: documented that yes, identities can migrate if space is requested

docs/0004_tx_ingestion.md
dshulyak: updated that we fallback to other pool

docs/0004_tx_ingestion.md
dshulyak: i will go with dropping for now, but it looks like evicting the other tx from identity with lower score provides better outcomes. just need to understand if it is worth the complexity

dshulyak:
> I agree with the rough intuition that there should be a boost based on the duration of the connection, but it seems like there needs to be some sort of penalty that needs to be overcome to realize it (like some target contribution that is a function of the time that needs to be overcome to get a boost).

it doesn't give them relative advantage over other honest long running identities. only over other newcomers. in my mind this is mostly to offset any immediate spam attempts. i am not sure if making this part more sophisticated offers other advantages 

> Something feels missing in that we aren't tracking the actual bandwidth utilization anywhere (maybe for simplicity). I feel like intuitively tracking the contribution per unit bandwidth with some confidence component that scales as a function of total bandwidth and/or time could work.

EMA (proposed by @tklenze ) is more self-penalizing? if they submit but don't land any transactions in a block it will naturally decrease over time. would it work as a proxy for bw utilization?

docs/0004_tx_ingestion.md
```diff
@@ -0,0 +1,233 @@
...
+### Score Computation
+
+```
+gas_contribution = min(total_gas_spent, max_gas_contribution)
+time_weight = min((time_known / time_weight_unit)^2, max_time_weight)
+decay = 0.5^(idle_time / decay_half_life)
+score = gas_contribution × time_weight × decay
+```
+
+Gas contribution measures useful work by summing gas spent on transactions included in blocks. Time weight grows quadratically from 0 to 1 over 1 hour, giving established identities a 3,600x advantage over 1-minute-old ones. Quadratic growth offsets attacks using short-term IP rental. Decay halves score every 30 minutes of inactivity, requiring consistent uptime to maintain priority. The gas contribution cap ensures that gaining more bandwidth requires acquiring more identities and IPs.
```
robertkiel: What if nodes perform gas mining? E.g. they allocate storage on-chain to increase gas consumption and release it afterwards. This is still expensive for the adversary but it leads to lower attack costs than the ones presented below.

docs/0004_tx_ingestion.md
```diff
@@ -0,0 +1,237 @@
...
+
+An identity is a public key that owns a wireauth session. All reputation tracking uses this public key as the identifier. The number of identities per IP address is limited to prevent a single IP from creating unbounded sessions.
+
+### Score Computation
+
+```
+ema_gas(t) = α × gas_this_block + (1 - α) × ema_gas(t - 1)
+α = 1 - 0.5^(block_time / ema_half_life)
+time_weight = min((time_known / time_weight_unit)^2, max_time_weight)
+score = ema_gas × time_weight
```
robertkiel, line 76: `ema_gas` and `time_weight` linearly contribute to score. Due to this, an attacker could compensate low "time scores" by high gas consumption and vice versa. Why don't apply the quadratic voting mechanism? E.g.

`score = sqrt(ema_gas * normalization_factor) + sqrt(time_weight)`

docs/0004_tx_ingestion.md
```diff
@@ -0,0 +1,243 @@
...
+Incoming transactions are routed based on authentication status and score. Authenticated identities with scores at or above `promotion_threshold` enter the priority pool; all others enter the regular pool.
+
+The bandwidth split between pools defaults to 90/10. During dequeue, a counter cycles to probabilistically select which pool to serve, achieving the configured ratio over time. The design is work-conserving: when the selected pool is empty, the other pool is served instead.
+
+### Weighted Fair Queueing
+
+Within each pool, bandwidth is allocated using weighted fair queueing (WFQ). Each identity maintains a virtual finish time that determines scheduling priority:
+
+```
+finish_time(i) = virtual_time + (1 / score(i))
```
robertkiel, line 138: Assuming we first calculate the score for a transaction `tx_score`. Based on `tx_score`, we select a pool, `priority` or `regular` and put the transaction in one of the pools.

Then, we apply the virtual finish time algorithm and select transactions from both pools according to the score.

In case any of the pools gets empty, let's say `regular` gets empty. In this case, `regular` won't advance its virtual clock and new incoming transactions in `regular` will get a "low" virtual time and thus get prioritized over transactions in the `priority` queue

A [recommendation](https://intronetworks.cs.luc.edu/current/uhtml/fairqueuing.html#a-second-virtual-finish-example) is to run the the "virtual clock" at `1/N` where `N` is the number of active queues, in our case `0, 1, 2`

dshulyak: i will look more into suggestion, but pools were meant to be completely isolated. they are integrated into one pool using 90/10 split (configurable, maybe 80/20 split), meaning for every 10 pops we pop 9 from priority according to its own order, and 1 from regular according to separate order

docs/0004_tx_ingestion.md
```diff
@@ -0,0 +1,243 @@
...
+Messages are keyed by `(identity, message_id)` tuple. The 3-byte message ID only needs to be unique per identity, not globally. This allows each identity to have up to 16M concurrent message IDs while keeping the header compact.
+
+Default parameters:
+
+| Parameter | Default | Description |
+|-----------|---------|-------------|
+| max_fragment_payload | 1432 | Bytes per fragment |
+| max_message_size | 128 KB | Maximum assembled message size |
+| max_fragments_per_message | 92 | Maximum fragments per message (128 KB / 1432) |
+| max_priority_messages | 10,000 | Concurrent priority reassemblies |
```
robertkiel: An attacker may keep sending multi-fragment chunks and withhold the last fragment. This fills the reassembly buffer and ultimately degrades `priority` transactions into `regular`. 

This goes under the assumption that the attacker has some reputation (score) and keeps sending incomplete transactions.

dshulyak: yes, i realized that. there should be a comment somewhere that we fallback to regular pool if priority pool is full and "stuck", also 10,000 might be too much, kept it as default. we just need to tolerate udp reordering, maybe 10-100 should be enough.

and regular pool provides randomized eviction, same as current raptorcast, so it is harder to attack but qos is worse

docs/0004_tx_ingestion.md
dshulyak: thanks, will need to model this, it basically reduces cost per gas, but i assume not to a zero

docs/0004_tx_ingestion.md
dshulyak: the goal of time factor is to make any short range attacks non-feasible, e.g make it impossible to compete with rpc's that are expected to be online anyway. so i think linear is not a problem here, but i will check what is the difference with your function

docs/0004_tx_ingestion.md
robertkiel: The point with the eviction into the `regular` pool is that this bypasses the whole mechanism, which means that nodes with higher reputation (score) directly compete with nodes with normal or no reputation.

dshulyak: so this is a fallback to prevent dos, degrading qos instead of rejecting txs. i considered penalizing peers that timeout in the buffer, it can be done to drain score of identities that do such thing, but need to be certain that necessary

priority pool is gated by certain threshold, that can be crossed only if actual monetary value was spent. together with limit on concurrent txs per identity i think should be sufficient to offset such attacks.

e.g you spent 1-10K to block the pool, and as the result only qos is degraded, so nothing to gain

robertkiel: Penalizing nodes whose transactions tend to trigger timeouts seems to be very effective

JasonSome:
I like the idea of a positive feedback loop from txpool to inform the DoS protection of the networking layer. A couple comments:
1. I wonder whether we should focus on **total priority fees** spent during contentious times, rather than gas spent; focusing on total priority fees would be more aligned with the validator's incentives, and also better predict (in contentious times) the probability of inclusion. Note that, in the case of about constant priority fees, this would just be a proxy for gas spent, except fees would be more incentive-aligned.
2. We would benefit from some degree of **customization** on this by validators, e.g., known trusted peers of validators (like RPCs) to be assigned boosted values. We can do this just by allowing a validator to set the total priority fees to some **virtual** value (i.e., virtual fees) or even a function for future extensibility. Note that this is again more incentive-aligned.

tklenze:
> 1. I wonder whether we should focus on **total priority fees** spent during contentious times

I see two issues with this:
1. It lowers the cost of boosting your reputation. An attacker can submit transactions with close to zero gas but a very high priority fee in order to boost its reputation. For example, if the priority fee is normally 10% of the gas cost of a transaction, then the attacker can get a 10x advantage. (Not saying that it's worth for an attacker to do this, but it does lower the cost).
2. ~A more serious attack is that a malicious RPC can collude with a validator to recoup its investment in priority fees. Concretely, the attacker creates transactions with almost zero gas and a high priority fee, and submits it to the colluding validator, who includes it in its block. Since priority fees go to the leader instead of being burned, this comes at no cost to the attacker. In contrast, gas is burned.~ (not relevant as pointed out by Lioba)

Re customizability: the easiest would just to allow for a whitelist. Hosts on the whitelists would have some very high limit.

liobaheimbach:
I agree with @tklenze that focussing on priority fees only might an issue when base fees are a significant chunk of the total fee. One could focus on total fees(priority and base fees) though as opposed to total priority fees. I do agree with @JasonSome that fees might be good to look at, not only gas. 

> A more serious attack is that a malicious RPC can collude with a validator to recoup its investment in priority fees. Concretely, the attacker creates transactions with almost zero gas and a high priority fee, and submits it to the colluding validator, who includes it in its block. Since priority fees go to the leader instead of being burned, this comes at no cost to the attacker. In contrast, gas is burned.

I might be missing something, but I do not fully understand this attack. If I understand it correctly, this would only affect the rpc scores of the colluding validator. The colluding validator/leader can overwrite the rpc score altogether. Or am I missing something and these scores are not local?

tklenze:
> I might be missing something, but I do not fully understand this attack. If I understand it correctly, this would only affect the rpc scores of the colluding validator. The colluding validator/leader can overwrite the rpc score altogether. Or am I missing something and these scores are not local?

Yes, that makes sense! Sorry for the confusion.

docs/0004_tx_ingestion.md
dshulyak: i think monad doesn't do gas refund https://github.com/category-labs/monad/blob/48f6d827640bcd19824abfd23a05a37580c50a70/category/execution/monad/monad_transaction_gas.cpp#L23-L30 . so atleast we can ignore that

docs/0004_tx_ingestion.md
dshulyak: i tried to look into this, the biggest issue for me was normalization factor. i could't make sense of it. also i don't think that this is a problem at all

> Due to this, an attacker could compensate low "time scores" by high gas consumption and vice versa

time factor is capped, so "vice versa" is not possible. in order to compensate for low  time score you have to spend disproportionately more gas

dshulyak:
> We would benefit from some degree of customization on this by validators, e.g., known trusted peers of validators (like RPCs) to be assigned boosted values

we can definitely add support for trusted peers. i will think about specific, we can completely prioritize traffic from them for example vs what @tklenze proposed

> I wonder whether we should focus on total priority fees 

from block proposer perspective this is exactly what they want to maximize profit, so even if it is more gameable it is probably the correct tradeoff

docs/0004_tx_ingestion.md
```diff
@@ -0,0 +1,243 @@
...
+### Packet Format
+
+Large messages are fragmented for transmission over authenticated UDP sessions. Each fragment carries an 8-byte header:
+
+| Field | Size | Description |
+|-------|------|-------------|
+| Version | 1 byte | Protocol version (currently 1) |
+| Message ID | 3 bytes | Identifies fragments belonging to the same message within a single identity |
+| Sequence Number | 2 bytes | Fragment index within message (0-65535) |
+| Flags | 2 bytes | START (0x0001) and END (0x0002) flags indicating message boundaries |
```
xinyuan-dev, line 168: Why choose two bytes for the flags of 4 states that fit in 2 bits?

Furthermore, I think we may further reduce it to a 1 bit "end" flag:

| end | condition | description             |
|-----|-----------|-------------------------|
| 0   | seq = 0   | initial fragment        |
| 1   | seq = 0   | single-fragment message |
| 0   | seq > 0   | interior fragment       |
| 1   | seq > 0   | final fragment          |

This 1-bit flag would also eliminate invalid combinations of the flags and the seq number, e.g. seq=0 for non-initial fragment.

dshulyak: i wanted to avoid bit packing for simplicity, as savings are minimal. i think lowest i can go is 4 bytes. 
and also have a room to add new flags without bumping version if there will be new cases, even if it seems unlikely  

wrt flags think this is good suggestion. i can trim message id to 2 bytes and leave 1 byte reserved for flags, and keep only END flag in the implementation

PR #2749, opened 2026-01-29
Zeroize KeystoreSecret::to_secp

omegablitz:


monad-secp/src/secp.rs
```diff
@@ -128,11 +128,12 @@ impl KeyPair {
         keypair
     }
 
-    pub fn from_ikm(ikm: &[u8]) -> Result<Self, Error> {
+    pub fn from_ikm(ikm: &mut [u8]) -> Result<Self, Error> {
         let dst = b"monad-ecdsa-keygen";
         let scalar =
             <K256Secp256k1 as GroupDigest>::hash_to_scalar::<ExpandMsgXmd<Sha256>>(&[ikm], &[dst])
                 .map_err(|_| Error(secp256k1::Error::InvalidSecretKey))?;
+        ikm.zeroize();
```
dshulyak, line 136: i think you should consume then, otherwise this is very easy to misuse. 

but why is it even needed. ikm is a special type that is zeroized on drop

PR #2751, opened 2026-02-02
monad-raptorcast: use name record address for PingPongCommand

dshulyak:
PingPongCommand strictly uses only the auth address specified in the name record. existing wireauth session is reused only if its address matches the one in the name record.

verified by a test that routes a ping message to the address in the name record when there is an existing session with the target identity at a different address.


eerkaijun (approved):

ariqchowdhury (approved):

dshulyak:
>  Error: File was unable to be removed Error: EACCES: permission denied, unlink '/__w/monad-bft/monad-bft/monad-raptorcast/src/parser/mod.rs'

PR #2752, opened 2026-02-02
monad-peer-discovery: validate ping/pong source address against name record

dshulyak:
discovery relies on send_with_record to route messages to the owner of the name record, it got broken and fixed in https://github.com/category-labs/monad-bft/pull/2751.  this change is meant to meant to make whole flow more robust since checking that the claimed ip address matches the source address is harder to break accidentally.

for example if we introduce some sort of flaw in the ping/pong randomness then current assumptions won't be enough to stop poisoning. basically checking source is more fool proof

test suite extended to cover ping/pong source address validation.

### DIFF
--- a/monad-peer-discovery/src/discovery.rs
+++ b/monad-peer-discovery/src/discovery.rs
@@ -15,7 +15,7 @@
 
 use std::{
     collections::{BTreeMap, BTreeSet, HashMap, HashSet},
-    net::SocketAddrV4,
+    net::{IpAddr, SocketAddrV4},
     num::NonZeroU32,
     path::PathBuf,
     time::Duration,
@@ -36,7 +36,7 @@ use crate::{
     MonadNameRecord, MonadNameRecordWithPubkey, NameRecord, PeerDiscoveryAlgo,
     PeerDiscoveryAlgoBuilder, PeerDiscoveryCommand, PeerDiscoveryEvent, PeerDiscoveryMessage,
     PeerDiscoveryMetricsCommand, PeerDiscoveryTimerCommand, PeerLookupRequest, PeerLookupResponse,
-    Ping, Pong, TimerKind,
+    PeerSource, Ping, Pong, TimerKind,
     ipv4_validation::{IpCheckError, validate_socket_ipv4_address},
 };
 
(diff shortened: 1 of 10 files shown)


monad-peer-discovery/src/discovery.rs
```diff
@@ -804,11 +804,25 @@ where
...
+        ping: SourcedMessage<Ping<Self::SignatureType>>,
     ) -> Vec<PeerDiscoveryCommand<ST>> {
-        debug!(?from, ?ping_msg, "handling ping request");
+        let src_addr = ping.src_addr;
+        let ping_msg = ping.message;
+        debug!(?from, ?src_addr, ?ping_msg, "handling ping request");
         self.metrics[GAUGE_PEER_DISC_RECV_PING] += 1;
 
+        let expected_ip = IpAddr::V4(ping_msg.local_name_record.name_record.ip());
+        if src_addr.ip() != expected_ip {
```
eerkaijun: any reason to only check expected ip instead of expected socket?

monad-peer-discovery/src/lib.rs
```diff
@@ -641,9 +647,11 @@ pub enum PeerDiscoveryEvent<ST: CertificateSignatureRecoverable> {
     },
     FullNodeRaptorcastRequest {
         from: NodeId<CertificateSignaturePubKey<ST>>,
+        src_addr: SocketAddr,
```
eerkaijun: is `src_addr` needed for `FullNodeRaptorcastRequest` and `FullNodeRaptorcastResponse`?

monad-peer-discovery/src/lib.rs
```diff
@@ -625,11 +631,11 @@ pub enum PeerDiscoveryEvent<ST: CertificateSignatureRecoverable> {
     },
     PeerLookupRequest {
         from: NodeId<CertificateSignaturePubKey<ST>>,
-        request: PeerLookupRequest<ST>,
+        request: SourcedMessage<PeerLookupRequest<ST>>,
```
eerkaijun: does this need to be `SourcedMessage`?

monad-peer-discovery/src/lib.rs
dshulyak: it was for consistency, but i changed approach. now there is a source tuple instead of NodeID

monad-peer-discovery/src/lib.rs
dshulyak: this was also for consistency, but as above i changed approach to use source instead of node id, so message are not touched

monad-peer-discovery/src/discovery.rs
dshulyak: i just didn't see a reason to have strict socket check, also i recall there was a desire to use tcp for secondary raptorcast and discovery (or only raptorcast?) , in that case it will be one more place to fix

dshulyak: if you prefer i can switch to full address

eerkaijun: there was desire to switch secondary raptorcast to use tcp

i don't have a very strong opinion but since different nodes can be behind a NAT and share the same IP, it feels more accurate to check against the socket

eerkaijun (dismissed):

monad-peer-discovery/src/discovery.rs
dshulyak: realized that ping can be received either from auth or non auth ports, so we would need to check for both here. i will keep ip

monad-raptorcast/src/lib.rs
```diff
@@ -1048,11 +1048,10 @@ where
                                 "RaptorCastPrimary rx deserialized PeerDiscoveryMessage: {:?}",
                                 peer_disc_message
                             );
-                            // handle peer discovery message in driver
                             this.peer_discovery_driver
                                 .lock()
                                 .unwrap()
-                                .update(peer_disc_message.event(from));
+                                .update(peer_disc_message.event_with_source(from, src_addr));
```
xinyuan-dev, line 1055: I'm looking to reduce udp state's knowledge about the networking layer stuff like the socket addr. instead of modifying the output of `UdpState::handle_message`, how about we retrieve it here from `message.src_addr` directly?

xinyuan-dev (dismissed):

monad-raptorcast/src/lib.rs
dshulyak: updated

PR #2753, opened 2026-02-02
monad-wireauth: drop tai64 dependency, implement tai64n in-place

dshulyak:
trivial zerocopy-based implementation preserving full compatibility

### DIFF
--- a/monad-wireauth/src/protocol/tai64.rs
+++ b/monad-wireauth/src/protocol/tai64.rs
@@ -1,0 +1,124 @@
+// Copyright (C) 2025 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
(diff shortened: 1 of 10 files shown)


monad-wireauth/src/protocol/tai64.rs
```diff
@@ -0,0 +1,123 @@
...
+    Hash,
+    FromBytes,
+    IntoBytes,
+    Immutable,
+    KnownLayout,
+)]
+#[repr(C)]
+pub struct Tai64N {
+    seconds: U64<BigEndian>,
+    nanoseconds: U32<BigEndian>,
```
xinyuan-dev, line 40: Why do we use big endian here? Other wireauth fields seem to only use little endian.

dshulyak: there is canonical encoding for tai64 https://cr.yp.to/libtai/tai64.html

btw i can't change even if we want to deviate from canonical, as it is only re-implementation to reduce bug surface, and changing to little endian will break compatibility

xinyuan-dev: I see. Also, I notice that the doc says:

> A TAI64N label is normally stored or communicated in external TAI64N format, consisting of twelve 8-bit bytes. The **first eight bytes** are the TAI64 label in [external TAI64 format](https://cr.yp.to/libtai/tai64.html#tai64). The **last four bytes** are the nanosecond counter in big-endian format.

Should we add a u32 padding between u64 and u32? Or if we explicitly don't want the padding, we probably want to `repr(packed)` on the struct/or statically assert that its size is 12 bytes.

xinyuan-dev: Just realized you added the `packed` directive, nice.

PR #2754, opened 2026-02-02
Raptorcast: refactor raptorcast chunk parser

xinyuan-dev:
This commit adapts the safe parser implemented in
https://github.com/category-labs/monad-bft/pull/2650 on top of the refactored `SignatureVerifier` type.

### DIFF
--- a/monad-raptorcast/src/parser/legacy.rs
+++ b/monad-raptorcast/src/parser/legacy.rs
@@ -1,0 +1,364 @@
+// Copyright (C) 2025 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// Legacy packet parser implementation.
+
+use bytes::Bytes;
+use monad_crypto::{
+    certificate_signature::{
+        CertificateSignature, CertificateSignaturePubKey, CertificateSignatureRecoverable,
+    },
+    hasher::{Hasher, HasherType},
+    signing_domain,
+};
+use monad_merkle::{MerkleHash, MerkleProof};
+use monad_types::{Epoch, Round};
+
+use crate::{
+    message::MAX_MESSAGE_SIZE,
+    packet::assembler::HEADER_LEN,
+    parser::signature_verifier,
+    udp::{
+        GroupId, MessageValidationError, ValidatedMessage, MAX_MERKLE_TREE_DEPTH, MAX_REDUNDANCY,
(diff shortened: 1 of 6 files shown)


monad-raptorcast/src/parser/legacy.rs
```diff
@@ -0,0 +1,363 @@
...
+
+        let known_addresses = HashMap::new();
+        (keys.pop().unwrap(), validators, known_addresses)
+    }
+
+    const EPOCH: Epoch = Epoch(5);
+    const UNIX_TS_MS: u64 = 5;
+
+    #[test]
+    fn test_legacy_vs_new_parser_equivalence() {
```
dshulyak, line 329: it would be better to use proptest or differential fuzzer to cover this part.

for single example it would be better to introduce insta test with expected input/output

monad-raptorcast/src/parser/packet_parser.rs
```diff
@@ -0,0 +1,425 @@
...
+
+const _: () = assert!(
+    RaptorcastChunkHeaderV0::SIZE == crate::packet::assembler::CHUNK_HEADER_LEN,
+    "RaptorcastChunkHeader size must match CHUNK_HEADER_LEN"
+);
+
+/// header + merkle root, used as cache key for signatures
+#[repr(transparent)]
+#[derive(Clone, PartialEq, Eq, Hash)]
+pub struct SignedOverDataV0([u8; RAPTORCAST_HEADER_V0_SIZE + MERKLE_HASH_SIZE]);
```
michael-yxchen, line 153: This struct is unused. Do you plan to switch the cache key in a later change?

xinyuan-dev: <s>yes, because for the deterministic rc the signed over data will no longer include an additional `MERKLE_HASH_SIZE`. let me mark it as unused for now.</s>

actually, let me just use this signed over data in the new packet parser and do some retrofitting for the legacy module.

monad-raptorcast/src/parser/legacy.rs
xinyuan-dev: differential fuzzer implemented in https://github.com/category-labs/monad-bft/pull/2754/commits/8c5cfccd67ad89e8a0399c90fc75e4e233186f57. let's merge the fuzz pr first.

monad-raptorcast/src/parser/packet_parser.rs
```diff
@@ -0,0 +1,422 @@
...
+
+        if chunk_header_and_payload.len() < RaptorcastChunkHeaderV0::SIZE {
+            return Err(MessageValidationError::TooShort);
+        }
+        let (chunk_header_bytes, payload) =
+            chunk_header_and_payload.split_at(RaptorcastChunkHeaderV0::SIZE);
+        let chunk_header: Ref<&[u8], RaptorcastChunkHeaderV0> =
+            Ref::from_bytes(chunk_header_bytes).map_err(|_| MessageValidationError::TooShort)?;
+
+        if payload.is_empty() {
```
robertkiel, line 227: Shouldn't there be a minimum size as well? An adversary may overwhelm nodes with e.g. 1-byte chunks

dshulyak: we should have that after parser

> if parsed_message.chunk.len() < MIN_CHUNK_LENGTH

i agree that checking it before signature is more defensive

xinyuan-dev: agree, issue noted in https://github.com/category-labs/monad-bft/issues/2825. i'll revisit this later.

PR #2755, opened 2026-02-02
Raptorcast: refactor publishing targets

xinyuan-dev:
This PR includes a bunch of changes around the `BuildTarget` group types used in message publishing:

- replace node view types like `FullNodesView`, `ValidatorsView` with simple reference of the owned group types
- simplify the use of `Group`/`FullNodes`/`EpochValidators` into explicit types like `ValidatorSet`, `SecondaryGroup`, and `Vec<NodeId>`
- delay the "`excl_self`" iterators to message builder step
- refactor Secondary Publisher usage of degenerate/invariance-violating `Group` type into `CurrentGroup` type that handles the cases explicitly

Rationale:

- the original types allow expressing invalid `BuildTarget` combinations:
  + Broadcast(NodesView) to non-validator nodes (should use individual point-to-point or use secondary)
  + FullNodeRaptorCast(Group) allows publishing to a validator group
- publishing to secondary group should never worry about "exclude self" because the publisher is never in the group.
- move all the "exclude self" logic to packet builder, which is closer to the actual protocols, other part of raptorcast can operate on the the full set
- unwrap unnecessary wrapper types:
  + `FullNodes` type doesn't require any invariances, or represent any extra semantic beyond `Vec<NodeId>` according to their uses.
  + similarly, `EpochValidators` is functionally and semantically identical to `ValidatorSet`
- enforce invariances on types, for example:
  + `SecondaryGroup` must be a non-empty set
  + `RoundSpan` must be an non-empty interval (e.g. `impl Default for {RoundSpan,Group}` are unsound)

### DIFF
--- a/monad-raptorcast/src/packet/assigner.rs
+++ b/monad-raptorcast/src/packet/assigner.rs
@@ -18,6 +18,7 @@ use std::{collections::HashMap, ops::Range};
 use bytes::BytesMut;
 use monad_crypto::certificate_signature::PubKey;
 use monad_types::{NodeId, Stake};
+use monad_validator::validator_set::{ValidatorSet, ValidatorSetType};
 use rand::{rngs::StdRng, seq::SliceRandom as _, SeedableRng as _};
 
 use super::{BuildError, Chunk, PacketLayout, Result};
@@ -315,13 +316,7 @@ pub(crate) struct Replicated<PT: PubKey> {
 }
 
 impl<PT: PubKey> Replicated<PT> {
-    pub fn from_unicast(node_id: NodeId<PT>) -> Self {
-        Self {
-            recipients: vec![Recipient::new(node_id)],
-        }
-    }
-
-    pub fn from_broadcast(recipients: Vec<NodeId<PT>>) -> Self {
+    pub fn from_broadcast(recipients: impl IntoIterator<Item = NodeId<PT>>) -> Self {
         Self {
             recipients: recipients.into_iter().map(Recipient::new).collect(),
         }
@@ -472,14 +467,16 @@ impl<PT: PubKey> StakeBasedWithRC<PT> {
     // for easy implementation in other languages, e.g., using Mt19937
     // and Fisher Yates shuffle.
     pub fn shuffle_validators(
-        view: &crate::util::ValidatorsView<PT>,
+        validator_set: &ValidatorSet<PT>,
+        self_id: &NodeId<PT>,
         seed: [u8; 32],
     ) -> Vec<(NodeId<PT>, Stake)> {
-        let mut validator_set = view
+        let mut validator_set = validator_set
+            .get_members()
             .iter()
-            .map(|(node_id, stake)| (*node_id, stake))
-            .collect::<std::collections::BinaryHeap<_>>()
-            .into_sorted_vec();
+            .filter(|(node_id, _stake)| *node_id != self_id)
+            .map(|(node_id, stake)| (*node_id, *stake))
+            .collect::<Vec<_>>();
         let mut rng = StdRng::from_seed(seed);
         validator_set.shuffle(&mut rng);
         validator_set
--- a/monad-raptorcast/src/packet/builder.rs
+++ b/monad-raptorcast/src/packet/builder.rs
@@ -21,7 +21,8 @@ use monad_crypto::certificate_signature::{
 };
 use monad_dataplane::udp::DEFAULT_SEGMENT_SIZE;
 use monad_types::NodeId;
-use rand::Rng;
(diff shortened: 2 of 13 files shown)


monad-raptorcast/src/raptorcast_secondary/publisher.rs
```diff
@@ -578,25 +621,22 @@ where
...
+        let members = self.full_nodes_accepted.clone();
 
-        Group::new_fullnode_group(
-            self.full_nodes_accepted.list.clone(),
-            &validator_id,
-            validator_id,
-            round_span,
-        )
+        if members.is_empty() {
+            CurrentGroup::inactive(self.start_round, self.end_round - self.start_round)
```
xinyuan-dev, line 629: @eerkaijun / @fabiotdf: the test case `gap_into_middle_of_scheduled_group` can produce a finalized group without any member. Please verify if this is desired behavior.

eerkaijun: yes, i believe it's possible to have an empty group. @fabiotdf can you confirm?

monad-raptorcast/src/packet/builder.rs
```diff
@@ -445,3 +462,14 @@ where
         BuildTarget::Broadcast(_) | BuildTarget::PointToPoint(_) => BroadcastMode::Unspecified,
     }
 }
+
+// introduce variation in the group member ordering
+fn randomize_secondary_group_nodes<PT: PubKey>(
+    group: &crate::util::SecondaryGroup<PT>,
+    seed: usize,
+) -> Vec<NodeId<PT>> {
+    let n = seed % group.len();
```
robertkiel, line 471: Why don't use [Fisher-Yates shuffle](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle) - this is simple and effective? This should be even more efficient than rotating the array.

monad-raptorcast/src/packet/builder.rs
robertkiel:
```suggestion
        rng: &mut impl Rng + CryptoRng,
```
this prevents from using bad RNGs

monad-raptorcast/src/packet/builder.rs
xinyuan-dev: We already use that in validator-to-validator raptorcast and I don't see why we wouldn't apply it here as well. However, this is a refactory PR so I intend to match the old logic without introducing any behavioral changes. See https://github.com/category-labs/monad-bft/pull/2755/changes#diff-51f39c3513b760064bef16aee658a1a480fef63221f21b3b22d3e846dabffd5bL451.

monad-raptorcast/src/packet/builder.rs
xinyuan-dev: fixed!

omegablitz (dismissed):

monad-raptorcast/src/util.rs
```diff
@@ -35,185 +36,57 @@ use tracing::{debug, warn};
 
 use crate::udp::GroupId;
 
-#[derive(Clone, Debug)]
-pub struct EpochValidators<PT: PubKey> {
-    pub validators: ValidatorSet<PT>,
+// Invariance: The group must be non-empty.
+#[derive(Debug, Clone, Default)]
```
michael-yxchen: nit: The Default derive contradicts the non-empty invariant. Is that necessary?

michael-yxchen (dismissed):

monad-raptorcast/src/util.rs
xinyuan-dev: fixed!

monad-raptorcast/src/packet/assigner.rs
```diff
@@ -472,12 +467,15 @@ impl<PT: PubKey> StakeBasedWithRC<PT> {
...
     ) -> Vec<(NodeId<PT>, Stake)> {
-        let mut validator_set = view
+        let mut validator_set = validator_set
+            .get_members()
             .iter()
-            .map(|(node_id, stake)| (*node_id, stake))
+            .filter(|(node_id, _stake)| *node_id != self_id)
+            .map(|(node_id, stake)| (*node_id, *stake))
             .collect::<std::collections::BinaryHeap<_>>()
             .into_sorted_vec();
```
dshulyak: q: is binary heap then sorting again necessary here given that data comes from btree? e.g `.collect::<Vec<_>>()` won't be equivalent?

dshulyak (approved):

monad-raptorcast/src/packet/assigner.rs
xinyuan-dev: fixed.

PR #2756, opened 2026-02-02
Proposed head must on canonical chain

philipglazman:


monad-consensus-state/src/lib.rs
```diff
@@ -1243,6 +1254,28 @@ where
...
+        {
+            if let Some(block) = self.consensus.pending_block_tree.get_block(&canonical_tip_id) {
+                cmds.push(ConsensusCommand::CommitBlocks(
+                    OptimisticPolicyCommit::Proposed {
+                        block: block.to_owned(),
+                        is_canonical: true,
+                    },
+                ));
+            }
+        }
```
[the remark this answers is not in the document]
philipglazman: Fixed

monad-blocktree/src/blocktree.rs
```diff
@@ -539,6 +539,21 @@ where
...
+        }
+        self.get_blocks_on_path_from_root(high_qc_block_id)
+            .and_then(|chain| {
+                chain
+                    .iter()
+                    .rev() // Start from high_qc, walk toward root
+                    .find(|block| self.is_coherent(&block.get_id()))
+                    .map(|block| block.get_id())
+            })
+    }
```
michael-yxchen: Children of the highest voted block are considered to be on the canonical chain. We should go beyond the `high_qc_block_id` in the block tree when looking for the highest coherent tip

monad-blocktree/src/blocktree.rs
```diff
@@ -539,6 +539,81 @@ where
 
         chain
     }
+
+    /// Returns the highest-round coherent descendant in the subtree rooted at `block_id`
```
omegablitz: nit: add comment saying that input block_id should be coherent

monad-blocktree/src/blocktree.rs
```diff
@@ -539,6 +539,81 @@ where
...
+    fn highest_round_coherent_descendant(&self, block_id: BlockId) -> BlockId {
+        let mut best_id = block_id;
+        let Some(mut best_round) = (if block_id == self.root.info.block_id {
```
omegablitz: nit:
I find this and the `let Some(children) = ...` lower down harder to read. Maybe something like this?
```rust
        let mut best_round = if block_id == self.root.info.block_id {
            self.root.info.round
        } else if let Some(entry) = self.tree.get(&block_id){
            entry.validated_block.get_block_round()
        } else {
            return block_id;
        };
```

And something similar for the `let Some(children)` below

monad-consensus-state/src/lib.rs
```diff
@@ -1443,30 +1449,64 @@ where
...
+        let high_cert_qc = self.consensus.pacemaker.high_certificate().qc();
+        let canonical_tip_id = self
+            .consensus
+            .pending_block_tree
+            .get_canonical_coherent_tip(high_cert_qc);
+        let block = self
+            .consensus
+            .pending_block_tree
+            .get_block(&canonical_tip_id)?;
+        Some(ConsensusCommand::CommitBlocks(
```
omegablitz: Probably worth deduplicating these

monad-consensus-state/src/lib.rs
```diff
@@ -1443,30 +1449,64 @@ where
...
+            OptimisticPolicyCommit::Proposed {
+                block: block.to_owned(),
+                is_canonical: true,
```
omegablitz: I think it would be clearer to make a new `OptimisticPolicyCommit::ProposedHead` or something.
Maybe the full lineup should be:
```rust
OptimisticPolicyCommit::Proposed
OptimisticPolicyCommit::UpdateProposedHead
OptimisticPolicyCommit::UpdatedVotedHead
OptimisticPolicyCommit::UpdateFinalizedHead
```
or something like that

The current naming was based on a scheme that predated the heads.

monad-consensus-state/src/lib.rs
```diff
@@ -1306,6 +1309,9 @@ where
                     )
                 }),
         );
+
+        cmds.extend(self.propose_canonical_head());
```
omegablitz: I think this is unnecessary now

monad-consensus-state/src/lib.rs
```diff
@@ -1443,30 +1449,64 @@ where
...
             &self.config.chain_config,
-        ) {
+        );
+
+        // Compute canonical coherent tip (highest coherent block on path to high_qc)
+        let high_cert_qc = self.consensus.pacemaker.high_certificate().qc();
+        let canonical_tip = self
+            .consensus
+            .pending_block_tree
+            .get_canonical_coherent_tip(high_cert_qc);
```
omegablitz: All of this canonical stuff can be deleted I think. All newly coherent blocks can be emitted as proposed, and at the end of this function, `self.propose_canonical_head()` can be called

monad-consensus-state/src/lib.rs
```diff
@@ -1493,6 +1485,40 @@ where
         cmds
     }
 
+    fn propose_canonical_head(
```
omegablitz: nit: add `#[must_use]`

omegablitz (dismissed):

monad-consensus-state/src/lib.rs
```diff
@@ -1477,6 +1487,41 @@ where
...
+            .consensus
+            .pending_block_tree
+            .get_block(&canonical_tip_id)?;
+
+        self.consensus.canonical_proposed_tip = Some(canonical_tip_id);
+
+        debug!(
+            seq_num =? block.header().seq_num,
+            block_round =? block.header().block_round,
+            "proposing canonical head"
```
michael-yxchen: nit: the log msg might be confused with proposals. We can change it to "updating proposed head". Maybe the function name can be changed too

monad-consensus-state/src/lib.rs
```diff
@@ -3715,22 +3766,35 @@ mod test {
         assert_eq!(n0.metrics.consensus_events.rx_bad_state_root, 1);
 
         assert_eq!(n0.consensus_state.get_current_round(), Round(6));
-        assert_eq!(cmds.len(), 5);
+        // Filter out Proposed commits (re-emitted for canonical tip) for assertion check
```
michael-yxchen: I think we should keep the commit proposed commands even if there're duplicates. It should still be deterministic and safe to assert on

monad-consensus-state/src/lib.rs
```diff
@@ -1477,6 +1487,41 @@ where
...
+            return None;
+        }
+
+        let block = self
+            .consensus
+            .pending_block_tree
+            .get_block(&canonical_tip_id)?;
+
+        self.consensus.canonical_proposed_tip = Some(canonical_tip_id);
+
```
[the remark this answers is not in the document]
philipglazman: cheap comparison and the staleness doesn't cause any correctness issue

monad-consensus-state/src/lib.rs
philipglazman: changed and updated function name

monad-consensus-state/src/lib.rs
philipglazman: Yea, good point. Updated test

PR #2760, opened 2026-02-03
Fix RPC request wrapper deserialize

andr-dev:
Previously, we were using the auto-derived serde `Deserialize` for `ResponseWrapper` in tests. Unfortunately, the autogenerated derive does not play nice with serde untagged so deserializing a batch would always fail. I encountered this issue when migrating `RequestWrapper` to use `RawValue`. For `ResponseWrapper` however, we never caught the issue because responses are never deserialized in normal RPC operations and there weren't any tests showing successful deserialization of a batch response.

This PR fixes the issue and adds a valid batch response deserialization test to the RPC batch conformance test suite. The issue does not affect production and is only usefu for writing tests that use `ResponseWrapper` to deserialize RPC call results.

### DIFF
--- a/monad-rpc/src/websocket/handler.rs
+++ b/monad-rpc/src/websocket/handler.rs
@@ -191,7 +191,7 @@ async fn handler(
                     Some(Ok(AggregatedMessage::Text(body))) => {
                         last_heartbeat = Instant::now();
 
-                        let request = to_request(body.as_bytes());
+                        let request = parse_request(body.as_bytes());
 
                         match request {
                             Ok(req) => {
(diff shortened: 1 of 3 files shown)


monad-rpc/src/websocket/handler.rs
```diff
@@ -665,7 +665,7 @@ fn to_response<S: Serialize + std::fmt::Debug>(resp: &S) -> String {
     }
 }
 
-fn to_request<'p>(body: &'p bytes::Bytes) -> Result<Request<'p>, JsonRpcError> {
```
andr-dev, line 668: This was a fly-by rename for clarity.

monad-rpc/src/websocket/handler.rs
```diff
@@ -665,7 +665,7 @@ fn to_response<S: Serialize + std::fmt::Debug>(resp: &S) -> String {
...
+fn parse_request<'p>(body: &'p bytes::Bytes) -> Result<Request<'p>, JsonRpcError> {
```
dshulyak, line 668: may as well add try_from(bytes: impl AsRef<[u8]) for relevant types

PR #2764, opened 2026-02-05
Raptorcast: clarify group types for primary/secondary raptorcast

xinyuan-dev:
This PR restructures the raptorcast group related types in order to make the types simpler and more explicit.

Major changes:

- Remove `Group`, split its usage into:
  + `SecondaryGroup` (newtype on `BTreeSet<Nodes>`, guaranteed non-empty)
  + `ValidatorSet`
  + `SecondaryGroupAssignment`, which bundles `publisher_id`, `round_span`, and a `SecondaryGroup` for the members
  + note in all above cases, self_id is not excluded from the member set, allowing these values unique universally (i.e. across nodes). If self_id is not desired, e.g. during rebroadcasting, the exclusion is delayed to that step.

- Remove `ReBroadcastGroupMap` type and the state in Raptorcast.
  + we already have `EpochValidators`, so we only add `FullNodeGroupMap` to the raptorcast state to track secondary groups
  + during the `handle_message`, we construct a `BroadcastGroup`, which is a borrowed view from group id and the groups. It's conceptually a raptorcast group of members related to the incoming raptorcast chunk, and implements relevant validation logic and utilities (i.e. group validity check, sender validity check, rebroadcasting peers).

Rationale:

The `Group` type was overloaded. It was used for both validator groups (primary raptorcast) and full node groups (secondary raptorcast). There are several downsides with this approach. First, it contains unnecessary data, e.g. round_span/validator_id for primary groups. Second, the condition that `other_peers` does not contains self is not applicable for secondary group where the self is a validator. The mixed semantics prevent `Group` type from representing an unambiguous, standalone concept.

The issue with `ReBroadcastGroupMap` is also from its overloaded meanings which makes its internal logic complex. In addition, the validator set per epoch is already stored independently in raptorcast state, duplicating the state in `ReBroadcastGroupMap` would be susceptible to desync if not managed correctly.

Please also see https://github.com/category-labs/monad-bft/pull/2843 for further refactory on these group types being used as part of build target.

### DIFF
--- a/monad-raptorcast/src/raptorcast_secondary/client.rs
+++ b/monad-raptorcast/src/raptorcast_secondary/client.rs
@@ -25,9 +25,10 @@ use tokio::sync::mpsc::UnboundedSender;
 use tracing::{debug, error, warn};
 
 use super::{
-    super::{config::RaptorCastConfigSecondaryClient, util::Group},
+    super::config::RaptorCastConfigSecondaryClient,
     group_message::{ConfirmGroup, PrepareGroup, PrepareGroupResponse},
 };
+use crate::util::{SecondaryGroup, SecondaryGroupAssignment};
 
 monad_executor::metric_consts! {
     pub CLIENT_NUM_CURRENT_GROUPS {
@@ -44,8 +45,6 @@ monad_executor::metric_consts! {
     }
 }
 
-type GroupAsClient<PT> = Group<PT>;
-
 // This is for when the router is playing the role of a client
 // That is, we are a full-node receiving group invites from a validator
 pub struct Client<ST>
@@ -58,10 +57,10 @@ where
     // upload bandwidth to broadcast chunk to a large group.
     config: RaptorCastConfigSecondaryClient,
 
-    // [start_round, end_round) -> GroupAsClient
+    // [start_round, end_round) -> SecondaryGroupAssignment
     // Represents all raptorcast groups that we have accepted and haven't expired
     // yet. The groups may overlap.
-    confirmed_groups: IntervalMap<Round, GroupAsClient<CertificateSignaturePubKey<ST>>>,
+    confirmed_groups: IntervalMap<Round, SecondaryGroupAssignment<CertificateSignaturePubKey<ST>>>,
 
     // start_round -> validator_id -> group invite
     // Once we receive an invite, we remember how the invite looked like, so
@@ -75,7 +74,7 @@ where
     >,
 
     // Once a group is confirmed, it is sent to this channel
-    group_sink_channel: UnboundedSender<GroupAsClient<CertificateSignaturePubKey<ST>>>,
+    group_sink_channel: UnboundedSender<SecondaryGroupAssignment<CertificateSignaturePubKey<ST>>>,
 
     // For avoiding accepting invites/confirms for rounds we've already started
     curr_round: Round,
@@ -95,7 +94,9 @@ where
 {
     pub fn new(
         client_node_id: NodeId<CertificateSignaturePubKey<ST>>,
-        group_sink_channel: UnboundedSender<GroupAsClient<CertificateSignaturePubKey<ST>>>,
+        group_sink_channel: UnboundedSender<
+            SecondaryGroupAssignment<CertificateSignaturePubKey<ST>>,
+        >,
         config: RaptorCastConfigSecondaryClient,
     ) -> Self {
         assert!(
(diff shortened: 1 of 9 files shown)


monad-raptorcast/src/util.rs
```diff
@@ -138,409 +109,300 @@ pub enum BroadcastMode {
...
-                    }
-                    debug!(
-                        ?msg_round,
-                        ?author_node_id,
-                        ?self.fullnode_map,
-                        ?src_addr,
-                        "msg_round not found for validator's round span");
-                    // FIXME: returning true for backward compatibility,
-                    // to be removed after upgrade
-                    return true;
```
xinyuan-dev, line 436: Please help confirm if this piece of backward-compatibility logic (as well as https://github.com/category-labs/monad-bft/pull/2764/changes#diff-51f39c3513b760064bef16aee658a1a480fef63221f21b3b22d3e846dabffd5bL461) is still applicable.

michael-yxchen: both can be removed

monad-raptorcast/src/udp.rs
```diff
@@ -146,20 +149,157 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
...
+        //     if !group.is_sender_valid(&sender_id) {
+        //         tracing::debug!(
+        //             ?parsed_message.group_id,
+        //             author =? parsed_message.author,
+        //             sender =? sender_id,
+        //             "dropping message from invalid sender"
+        //         );
+        //         return None;
+        //     }
+        // }
```
michael-yxchen: the check is valid

michael-yxchen (approved):
lgtm. Thanks for noting the redundant backwards compatibiltiy branches

monad-raptorcast/src/raptorcast_secondary/client.rs
```diff
@@ -367,11 +368,13 @@ where
             return false;
         }
 
-        let group = GroupAsClient::new_fullnode_group(
-            confirm_msg.peers.into_inner(),
-            &self.client_node_id,
+        let members = confirm_msg.peers.into_inner().into_iter().collect();
```
dshulyak, line 379: was into_iter().collect() intentional? original code just moved internal collection as i understand

monad-raptorcast/src/udp.rs
```diff
@@ -146,20 +149,155 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
         self.decoder_cache.metrics()
     }
 
+    pub fn handle_unicast(
+        &mut self,
+        epoch_validators: &BTreeMap<Epoch, ValidatorSet<CertificateSignaturePubKey<ST>>>,
+        parsed_message: &ValidatedMessage<CertificateSignaturePubKey<ST>>,
+        _sender_pk: Option<&CertificateSignaturePubKey<ST>>,
```
dshulyak, line 158: can it be removed?

monad-raptorcast/src/raptorcast_secondary/client.rs
xinyuan-dev: `members` needs to be a `BTreeMap`, but `LimitedVec::into_inner` returns a `Vec` which needs to be converted to a `BTreeMap`.

monad-raptorcast/src/udp.rs
xinyuan-dev: The sender's identity can be useful for validation here. Although we don't use it here because of I don't want to change the old behavior, I think it's nice to keep it explicitly listed here.

monad-raptorcast/src/util.rs
```diff
@@ -138,409 +109,353 @@ pub enum BroadcastMode {
...
+    }
+
+    pub fn len(&self) -> usize {
+        self.map.len()
+    }
+
+    #[must_use]
+    pub fn try_insert(&mut self, assignment: SecondaryGroupAssignment<PT>) -> Option<()> {
+        let group_map = self.map.entry(assignment.publisher_id).or_default();
+        group_map.try_insert(assignment.round_span, assignment.group)
```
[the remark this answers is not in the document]
xinyuan-dev: during normal operation i don't expect any invalid assignments from secondary client, and even if it happens the slight side-effect is insignificant and will be cleared up in the next round. preferring the original code for clarity.

monad-raptorcast/src/util.rs
```diff
@@ -138,409 +109,353 @@ pub enum BroadcastMode {
...
-            start_ix: 0,
+    pub fn get_current_or_next(&self, round: Round) -> Option<&SecondaryGroup<PT>> {
+        if let Some(group) = self.get(round) {
+            return Some(group);
         }
+        let (span, group) = self.group_map.smallest()?;
+        if round < span.start {
+            return Some(group);
+        }
+        None
```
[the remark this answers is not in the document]
xinyuan-dev: The new code is more restrictive: it only falls back to the smallest group when round is before all groups. This invariance is maintained with `delete_expired(current_round)` so it's logically equivalent to the compatibility logic.

PR #2768, opened 2026-02-07
Fix flaky dataplane byte limit test

andr-dev:
This test would occasionally fail during rust CI tests because it is probabilistic in nature.

Example: https://github.com/category-labs/monad-bft/actions/runs/21785358893/job/62855845345?pr=2767#step:15:2141

By increasing the message count, we dramatically increase the probability of the test reliably passing. Tested by running the test sequentially `1000` times.


dshulyak:
opened alternative fix here github.com/category-labs/monad-bft/pull/2770

andr-dev:
https://github.com/category-labs/monad-bft/pull/2770

Issue #2777, opened 2026-02-10
monad-wireauth: add soft/hard handshake rate limits (cookie-gated at soft, drop at hard)


dshulyak:
right now the "cookie/ip validation" path only kicks in once total_sessions >= low_watermark_sessions. an attacker can try to keep total_sessions just below that threshold while still generating enough handshake traffic to consume the global handshake budget, causing honest peers' new connections to get dropped without ever triggering cookie-gated, per-ip controls.

PR #2778, opened 2026-02-10
monad-wireauth: split handshake rate limits by cookie validity

dshulyak:
right now the "cookie/ip validation" path only kicks in once
total_sessions >= low_watermark_sessions. an attacker can try to keep
total_sessions just below that threshold while still generating enough
handshake traffic to consume the global handshake budget, causing honest
peers new connections to get delayed or dropped without ever triggering
cookie-gated, per-ip controls.

split the handshake budget into two independent rate limits:
- handshake_cookie_unverified_rate_limit for cookie-invalid initiations
- handshake_cookie_verified_rate_limit for cookie-valid initiations

once the cookie-unverified budget is exhausted we send cookie replies
instead of dropping, so honest peers can obtain a cookie and retry into 
the protected verified lane. cookie replies are smaller than initiation/response, 
so even if we reply under attack we are not amplifying traffic

closes: https://github.com/category-labs/monad-bft/issues/2777

### DIFF
--- a/monad-wireauth/src/filter.rs
+++ b/monad-wireauth/src/filter.rs
@@ -34,9 +34,11 @@ pub enum FilterAction {
 // Filter ...
 // NOTE that rate limiting for ipv6 is not properly supported
 pub struct Filter {
-    counter: u64,
+    cookie_unverified_counter: u64,
+    cookie_verified_counter: u64,
     last_reset: Duration,
-    handshake_rate_limit: u64,
+    handshake_cookie_unverified_rate_limit: u64,
+    handshake_cookie_verified_rate_limit: u64,
     handshake_rate_reset_interval: Duration,
     ip_request_history: LruCache<IpAddr, Duration>,
     ip_rate_limit_window: Duration,
@@ -48,9 +50,11 @@ pub struct Filter {
 }
 
 impl Filter {
+    #[allow(clippy::too_many_arguments)]
     pub fn new(
         metric_names: &'static MetricNames,
-        handshake_rate_limit: u64,
+        handshake_cookie_unverified_rate_limit: u64,
+        handshake_cookie_verified_rate_limit: u64,
         handshake_rate_reset_interval: Duration,
         ip_rate_limit_window: Duration,
         ip_history_capacity: usize,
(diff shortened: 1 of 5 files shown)


monad-wireauth/src/filter.rs
```diff
@@ -141,16 +147,49 @@ impl Filter {
...
+                    remote_addr = %remote_addr,
+                    counter = self.cookie_verified_counter,
+                    verified_rate_limit = self.handshake_cookie_verified_rate_limit,
+                    "cookie-verified rate limit exceeded - dropping handshake"
+                );
+                return Some(FilterAction::Drop);
+            }
+            self.cookie_verified_counter += 1;
+        } else {
+            if self.cookie_unverified_counter >= self.handshake_cookie_unverified_rate_limit {
```
robertkiel: Doesn't this reveal some of the internal state?

There are three observable outcomes, namely
- a handshake reply
- a cookie
- nothing, packet gets dropped

As an adversary, we can
- sample a random cookie
- use a previously retrieved cookie

To see if the node is busy, we send a random value. If the node replies with a cookie, we know that usage is very low and take the cookie. Otherwise, we use a correct cookie. If the packet gets dropped and we don't receive a reply, then usage has execeeded `self.handshake_cookie_verified_rate_limit`. If we receive a reply with the handshake reply, we know that usage is below `self.handshake_cookie_verified_rate_limit`

dshulyak: do you see a specific problem? this should not be an important factor in preventing exploit that i am trying to mitigate

cookie can be used to rate limit by ip, so to consume whole verified budget someone will need to get ips to bypass that rate limit. with default params that would be 10_000 ips

robertkiel: True, it does not affect the mitigation of the vulnerability but it might create a new side channel that leaks useful data for DoS attacks.

dshulyak:
```

Run actions/checkout@v4
/usr/bin/docker exec  bc8335f374ace18483ba311602f79149a00e6dc944952c39534df3d1e071eeb9 sh -c "cat /etc/*release | grep ^ID"
Syncing repository: category-labs/monad-bft
Getting Git version info
Temporarily overriding HOME='/__w/_temp/e4b93360-bf33-432a-a698-ace48bb2f818' before making global git config changes
Adding repository directory to the temporary git global config as a safe directory
/usr/bin/git config --global --add safe.directory /__w/monad-bft/monad-bft
Deleting the contents of '/__w/monad-bft/monad-bft'
Error: File was unable to be removed Error: EACCES: permission denied, unlink '/__w/monad-bft/monad-bft/target/.rustc_info.json'

```

PR #2779, opened 2026-02-11
[rpc] `eth_simulateV1` implementation

Baltoli:


monad-rpc/src/handlers/eth/simulate.rs
```diff
@@ -0,0 +1,331 @@
...
+                .map_err(|e| JsonRpcError::internal_error(format!("CBOR decode error: {}", e)))?;
+            serde_json::value::to_raw_value(&v).map_err(|e| {
+                JsonRpcError::internal_error(format!("json serialization error: {}", e))
+            })
+        }
+
+        SimulateResult::Failure(error) => {
+            Err(JsonRpcError::eth_call_error(error.message, error.data))
+        }
+    }
```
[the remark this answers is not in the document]
dhil: Fixing this problem is on our radar, but out of scope for this patch. We need to rearchitect parts of the execution client tracing logic. Won't fix for now.

monad-rpc/src/handlers/eth/simulate.rs
```diff
@@ -0,0 +1,331 @@
...
+            BlockTagOrHash::BlockTags(BlockTags::Number(Quantity(block_number - 1))),
+        )
+        .await
+        .ok_or_else(JsonRpcError::block_not_found)?;
+        match block_key {
+            BlockKey::Finalized(FinalizedBlockKey(SeqNum(_))) => None,
+            BlockKey::Proposed(ProposedBlockKey(SeqNum(_), BlockId(Hash(id)))) => Some(id),
+        }
+    } else {
+        None
```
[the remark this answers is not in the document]
dhil: I believe this suggestion is incorrect. First, there is no parent of block 0. Second, if the block is finalized then we correctly return `None`, which gets translated into to an zero `bytes32_t` in the execution client. The execution client distinguishes between finalized and proposed blocks by whether the provided `block_id : bytes32_t` is zero.

monad-rpc/src/handlers/eth/simulate.rs
```diff
@@ -0,0 +1,331 @@
...
+        .simulation
+        .block_state_calls
+        .iter()
+        .map(|bsc| {
+            bsc.calls
+                .iter()
+                .map(|call| call.from.unwrap_or_default())
+                .collect()
+        })
+        .collect();
```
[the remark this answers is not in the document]
dhil: I believe the current behaviour is spec-complaint.

dhil:
@claude please critically review this PR.

monad-rpc/src/handlers/eth/simulate.rs
```diff
@@ -0,0 +1,349 @@
...
+
+    if calls.len() != overrides.len() {
+        return Err(JsonRpcError::invalid_params());
+    }
+
+    for (call_list, (_, state_override)) in calls.iter_mut().zip(overrides.iter()) {
+        for call in call_list.iter_mut() {
+            // Merge `data` and `input` fields (compatible with
+            // go-ethereum, see https://github.com/ethereum/go-ethereum/issues/15628).
+            call.input.input = match (call.input.input.take(), call.input.data.take()) {
```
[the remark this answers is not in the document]
dhil: Block overrides are applied by execution.

monad-rpc/src/handlers/eth/simulate.rs
```diff
@@ -0,0 +1,349 @@
...
+        .iter()
+        .map(|bsc| {
+            bsc.calls
+                .iter()
+                .map(|call| call.from.unwrap_or_default())
+                .collect()
+        })
+        .collect();
+
+    let mut header = match data_provider
```
andr-dev: style nit:
```rust
let Some(mut header) = ... else {
  return Err(JsonRpcError::block_not_found());
};
```

monad-rpc/src/handlers/eth/simulate.rs
```diff
@@ -0,0 +1,349 @@
...
+        .block_state_calls
+        .iter()
+        .map(|bsc| bsc.calls.to_vec())
+        .collect();
+
+    if calls.len() != overrides.len() {
+        return Err(JsonRpcError::invalid_params());
+    }
+
+    for (call_list, (_, state_override)) in calls.iter_mut().zip(overrides.iter()) {
```
andr-dev, line 170: The length check is dead code, the two collections will have the same size since we're iterating over the same thing. Also, this zipping is confusing, can we first `iter_mut()` over `block_state_calls` to do this call input update, then below do the calls and overrides iter? Also, I vaguely remember that `eth_simulate_v1` asserts the size of the calls and overrides param passed in are the same, why not make `eth_simulate_v1` take them in a single iter?

dhil: If I am understanding your suggestion correctly, then you are suggesting to hoist the call input fixup into its own loop. Then afterwards do the zip to fill in gas parameters?

And your last suggestion is to zip `calls`, `senders`, and `overrides` together and pass the resulting iterator to the execution binding?

dhil:
@claude please review this PR.

monad-rpc/src/handlers/eth/simulate.rs
```diff
@@ -0,0 +1,350 @@
...
+                    call.try_into().map_err(|_| JsonRpcError::invalid_params())
+                })
+                .collect::<Result<Vec<_>, _>>()
+        })
+        .collect::<Result<Vec<_>, _>>()?;
+
+    let (block_number, block_id) = match block_key {
+        BlockKey::Finalized(FinalizedBlockKey(SeqNum(n))) => (n, None),
+        BlockKey::Proposed(ProposedBlockKey(SeqNum(n), BlockId(Hash(id)))) => (n, Some(id)),
+    };
```
philipglazman: There's a private fn called `block_key_to_parts` in `monad-rpc/src/handlers/eth/gas.rs` that we can potentially reuse

dhil:
@claude please review this PR again. Take into account the previous comments.

monad-rpc/src/handlers/eth/simulate.rs
```diff
@@ -0,0 +1,350 @@
...
+                        Some(input)
+                    }
+                    (None, data) | (data, None) => data,
+                };
+            }
+            Ok(calls)
+        })
+        .collect::<Result<Vec<_>, _>>()?;
+
+    for (call_list, (_, state_override)) in calls.iter_mut().zip(overrides.iter()) {
```
philipglazman, line 170: Should we check that the sum of the gas calls are less than the provider gas limit?

monad-rpc/src/handlers/eth/simulate.rs
```diff
@@ -0,0 +1,350 @@
...
+            .await?;
+        }
+    }
+
+    let calls: Vec<Vec<TxEnvelope>> = calls
+        .into_iter()
+        .map(|call_list| {
+            call_list
+                .into_iter()
+                .map(|call: CallRequest| {
```
philipglazman: If provided, check that call.chain_id matches our chain_id. Otherwise, fill in the chain_id to the call. Similar to `monad-rpc/src/handlers/eth/call.rs:640`

monad-rpc/src/handlers/eth/simulate.rs
```diff
@@ -0,0 +1,350 @@
...
+                    (None, data) | (data, None) => data,
+                };
+            }
+            Ok(calls)
+        })
+        .collect::<Result<Vec<_>, _>>()?;
+
+    for (call_list, (_, state_override)) in calls.iter_mut().zip(overrides.iter()) {
+        for call in call_list.iter_mut() {
+            // Inherit the base header. The execution client applies overrides to the simulation header.
```
philipglazman: Will execution fill in the baseFeePerGas or gasLimit for each call too? We call fill_gas_params without any block overrides for those two fields

monad-rpc/src/cli.rs
```diff
@@ -169,6 +169,14 @@ pub struct Cli {
     #[arg(long, default_value_t = 2)]
     pub eth_call_executor_queuing_timeout: u32,
 
+    /// Set the gas limit for `eth_simulateV1`
+    #[arg(long, default_value_t = u64::MAX)]
```
eerkaijun: in terms of DoS prevention, is the expectation that rpc providers should manage that themselves? I think with this gas limit even a single eth_simulateV1 call can block a thread for some time

monad-rpc/src/handlers/eth/simulate.rs
```diff
@@ -0,0 +1,350 @@
...
+            let mut header = header.header.clone();
+            fill_gas_params(
```
eerkaijun, line 174: `fill_gas_params` can do triedb reads, so it is potentially an expensive operation. seems like we do not bind `calls` by `max_calls` yet at this point, so I think a malicious user can issue many triedb reads. I think we should do the `max_calls` check at the beginning.

monad-rpc/src/cli.rs
dhil: Yes; we set a lower default. Geth defaults to `U256::MAX`.

monad-rpc/src/handlers/eth/simulate.rs
dhil: I suppose we can do an early exit here if that's the case. Execution will stop if the gas limit is exceeded.

monad-rpc/src/handlers/eth/simulate.rs
dhil: Yes, execution constructs the headers during simulation as they become necessary. I think the only thing we need here to set up the transaction objects accordingly.

monad-rpc/src/handlers/eth/simulate.rs
dhil: I suppose we want two enforce two limits here: maximum number of block simulations (which is actually what `max_calls` is supposed to governing now; I will rename this variable),
 and then we also want to restrict the total number of transaction invocations (I will use the name `max_calls` to denote this restriction).

monad-rpc/src/handlers/eth/simulate.rs
andr-dev: Yeah, looking at the `monad-ethcall` code, the first thing `eth_simulate_v1` does is assert that `calls`, `overrides`, and `senders` are all the same size. On top of that, we then assert that the lengths of the items in `calls` and `senders` are the same size. That indicates to me we should be passing in some collection of more structured objects.

monad-rpc/src/handlers/eth/simulate.rs
dhil: I've added both configurables.

monad-rpc/src/handlers/eth/simulate.rs
dhil: I've implemented a static check for this.

dhil:
@claude review this patch again. Take into account the previous reviews.

monad-rpc/src/cli.rs
dhil: OK, after some conversations, we have settled on a default value of 8 times Monad's block gas limit. Again, this is a configurable value that each provider can tune to their own needs.

dhil:
@claude one more review please. Note that the "CBOR → JSON conversion remains fragile" issue wont be resolved by this PR, and "Shared eth_call_handler semaphore for simulate requests" is as intended, and "Test coverage is deserialization-only" is as intended as we have end-to-end tests externally (and several unit tests in execution).

monad-rpc/src/handlers/eth/simulate.rs
dhil: I will make this change in a subsequent PR as it touches execution too.

monad-rpc/src/handlers/eth/simulate.rs
```diff
@@ -0,0 +1,383 @@
...
+    true
+}
+
+#[derive(Debug, Deserialize)]
+pub struct MonadSimulateParams {
+    pub simulation: MonadSimulation,
+    pub block: BlockTagOrHash,
+}
+
+pub async fn monad_simulate_v1<T: Triedb + TriedbPath>(
```
philipglazman, line 78: We should add the rpc macro to make this method available in the rpc docs

```
#[rpc(
    method = "eth_simulateV1",
    ignore = "chain_id,call_gas_limit,simulation_gas_limit,max_simulated_calls,max_simulated_blocks"
)]
```

monad-rpc/src/handlers/eth/simulate.rs
```diff
@@ -0,0 +1,390 @@
...
+                // Merge `data` and `input` fields (compatible with
+                // go-ethereum, see https://github.com/ethereum/go-ethereum/issues/15628).
+                call.input.input = match (call.input.input.take(), call.input.data.take()) {
+                    (Some(input), Some(data)) => {
+                        if input != data {
+                            return Err(JsonRpcError::invalid_params());
+                        }
+                        Some(input)
+                    }
+                    (None, data) | (data, None) => data,
```
eerkaijun, line 162: just noting this chunk of code has been duplicated a few times with eth_call and eth_estimateGas so probably worth modularizing it. I can clean it up in a separate PR though so don't have to fix it now

eerkaijun (approved):

philipglazman (approved):

monad-rpc/src/handlers/eth/simulate.rs
andr-dev: Side note for this later change that we should have different types for the RPC request that allows passing either "input" or "data" and the "transformed" type that only has "input". Then, we can force a conversion through a `new` function or similar, ensuring that all usages are identical rather than relying on a helper fn being called.

andr-dev (approved):

omegablitz (approved):

PR #2783, opened 2026-02-11
Create proposal ahead

bamrith:
Txpool optimistically creates a proposal for `Round(N)` assuming a QC will be formed for the block proposed in `Round(N-1)`. On formation of QC, consensus fetches the created proposal and broadcasts it immediately.

In the event that there's a timeout or consensus decides to extend a different block in `Round(N)`, txpool defaults back to creating the proposal when `Round(N)` begins

### DIFF
--- a/monad-eth-txpool/src/pool/mod.rs
+++ b/monad-eth-txpool/src/pool/mod.rs
@@ -15,9 +15,7 @@
 
 use std::time::Duration;
 
-use alloy_consensus::{
-    constants::EMPTY_WITHDRAWALS, transaction::Recovered, TxEnvelope, EMPTY_OMMER_ROOT_HASH,
-};
+use alloy_consensus::{transaction::Recovered, TxEnvelope};
 use alloy_primitives::Address;
 use alloy_rlp::Encodable;
 use itertools::{Either, Itertools};
@@ -26,10 +24,7 @@ use monad_chain_config::{
     revision::{ChainRevision, MockChainRevision},
     ChainConfig, MockChainConfig,
 };
-use monad_consensus_types::{
-    block::{BlockPolicyError, ConsensusBlockHeader, ProposedExecutionInputs},
-    payload::RoundSignature,
-};
+use monad_consensus_types::block::{BlockPolicyError, ConsensusBlockHeader};
 use monad_crypto::certificate_signature::{
     CertificateSignaturePubKey, CertificateSignatureRecoverable,
 };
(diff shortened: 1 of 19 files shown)


dshulyak:
as i understand this change will add 400ms latency to tx inclusion in the best case. as we are currently sending tx batches to next 3 leaders. and the closest leader will definitely miss such batch. so basically we are trading off inclusion latency for reliability, e.g preventing long proposal build times causing timeouts.

in my opinion this is not the right tradeoff and it would make sense to figure out algo that allows to maintain "best" proposal given current set of txs in  the mempool, so that when consensus requests mempool we can give it instantly or near instantly "best" proposal.

if we consider that this is a low risk, we should adjust tx forwarder to not send txs to the closest leader together with this change to not waste bandwidth, not necessarily in the pr.

bamrith:
@dshulyak 
> so basically we are trading off inclusion latency for reliability, e.g preventing long proposal build times causing timeouts.

the goal was to start creating proposals at the earliest stage instead of waiting for a QC. but, this also allows validators to run early simulations for MEV purposes.

> in my opinion this is not the right tradeoff and it would make sense to figure out algo that allows to maintain "best" proposal given current set of txs in the mempool, so that when consensus requests mempool we can give it instantly or near instantly "best" proposal.
> if we consider that this is a low risk, we should adjust tx forwarder to not send txs to the closest leader together with this change to not waste bandwidth, not necessarily in the pr.

maintaining a "best" proposal with all the mempool transactions was also my first idea. however, I figured that if full nodes are forwarding transactions to the next 3 leaders, all of them will have (almost) same "best" proposals, and this would cause 2nd/3rd leader to re-build the proposal anyway when they see the extending blocks.

dshulyak:
>  I figured that if full nodes are forwarding transactions to the next 3 leaders, all of them will have (almost) same "best" proposals, and this would cause 2nd/3rd leader to re-build the proposal anyway when they see the extending blocks.

they will rebuild it but 1 slot later, that was main point of my previous message.
tx inclusion latency is definitely one of the metrics that we want to optimize for, and here we are making it worse by additional 400ms. it is not always will be 400ms, but definitely will skew the distribution, i don't know if thats fine

monad-eth-txpool/src/pool/mod.rs
```diff
@@ -98,6 +109,7 @@ where
 
         Self {
             tracked: TrackedTxMap::new(config_limits),
+            ready_proposals: BTreeMap::new(),
```
andr-dev: Ready proposals should live in the txpool executor, not the pool datastructure. There should not be any differences in the sequencing process for creating a proposal ahead of time vs just in time.

PR #2785, opened 2026-02-12
Rlp cleanup and stricter decoding enforcement

eerkaijun:
No trailing data during rlp decoding of network messages

### DIFF
--- a/monad-blocksync/src/messages/message.rs
+++ b/monad-blocksync/src/messages/message.rs
@@ -69,13 +69,19 @@ impl Decodable for BlockSyncRequestMessage {
                 "expected to decode type BlockSyncRequestMessage",
             ));
         }
-        match u8::decode(&mut payload)? {
-            1 => Ok(Self::Headers(BlockRange::decode(&mut payload)?)),
-            2 => Ok(Self::Payload(ConsensusBlockBodyId::decode(&mut payload)?)),
-            _ => Err(alloy_rlp::Error::Custom(
-                "failed to decode unknown BlockSyncRequestMessage",
-            )),
+        let result = match u8::decode(&mut payload)? {
+            1 => Self::Headers(BlockRange::decode(&mut payload)?),
+            2 => Self::Payload(ConsensusBlockBodyId::decode(&mut payload)?),
+            _ => {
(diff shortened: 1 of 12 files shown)


omegablitz (dismissed):

monad-blocksync/src/messages/message.rs
```diff
@@ -69,13 +69,19 @@ impl Decodable for BlockSyncRequestMessage {
...
+        let result = match u8::decode(&mut payload)? {
+            1 => Self::Headers(BlockRange::decode(&mut payload)?),
+            2 => Self::Payload(ConsensusBlockBodyId::decode(&mut payload)?),
+            _ => {
+                return Err(alloy_rlp::Error::Custom(
+                    "failed to decode unknown BlockSyncRequestMessage",
+                ))
+            }
+        };
+        if !payload.is_empty() {
```
xinyuan-dev, line 81: I'm not sure if this is the best practice. Enforcing no trailing data may hurt composability for embedding these messages larger structs. I recommend tackling the problem from the caller end instead, and use functions like [`decode_exact`](https://docs.rs/alloy/latest/alloy/rlp/fn.decode_exact.html) to ensure no trailing bytes.

omegablitz: This is enforcing there's no trailing bytes within the list, correct?
From my understanding, decode_exact would enforce that there are no trailing bytes after the list

eerkaijun: Let me add some tests to verify the behaviour and rework this PR. I also wanted to audit all network messages to make sure I don't miss out any

eerkaijun: So decode_exact checks that there is no trailing byte after a list e.g. [content1, content2] JUNK
while we want to enforce no trailing byte inside the list e.g. [content1, content2, JUNK]

Since we are implementing custom decoding for these structs, I think the only way is the to check
`payload.is_empty()` inside these custom decode implementation. This is also the same behavior as the default `derive(RlpDecodable)`

eerkaijun:
Did an audit of all network messages and this PR covers the custom decoding of all these messages including their inner structs

PR #2790, opened 2026-02-16
monad-blocksync: cap peer request table for non-validators

dshulyak:


michael-yxchen:
The change by itself looks fine, though I'm not sure about its effect in practice. In steady state, the ledger executor only looks up blocks from its in-memory cache. The ledger executor is polled with [very high priority](https://github.com/category-labs/monad-bft/blob/1eac252a426209d8b48e084238f796d08fac087a/monad-updaters/src/parent.rs#L186-L187), before router. I doubt the table will fill up when we're under blocksync dos traffic, because it gets serviced immediately. Perhaps a rate limit addresses this more directly?

dshulyak:
@michael-yxchen i think if blocksync injects requests from network faster than ledger can service them there always will be growth in this maps and in unbounded channel owned by ledger executor. for the ledger itself the limiting factor should be disk access. 

so for example if we set rate limiter to one value and it turns out that worst case is slower then we won't prevent memory growth. like we set rate limit to 100rps and worst case is 1rps. so to me it seems that rate limiting here is error prone, maybe it is beneficial on the ledger side but that would solve different problem.

in general i think it is solid practice to limit collections that store data from network, very hard to get it wrong.  worst case will stay bounded even if some assumption about db speed or polling are broken

michael-yxchen:
@dshulyak I agree that this a safe change. I just don't think the request table will grow to more than 1 entry after cache is hydrated. The block file ledger fetch is synchronous and [only accesses memory in steady state](https://github.com/category-labs/monad-bft/blob/e05a05c02e1779680398d061c29a607ec6284aa7/monad-ledger/src/lib.rs#L188-L197). The synchronous nature prevents monad-state from ingesting the next event, and the high priority of ledger executor makes sure ledger response is handled before the next blocksync message from the router.
 
I attempted to bloat the request table by flooding different payload requests. The PoC is pushed to [this branch](https://github.com/category-labs/monad-bft/tree/michael/blocksync-poc/monad-blocksync-flood). I ran it against a single node in flexnet and confirmed that it never had more than 1 entries. Here's an excerpt of the logs
```
{"timestamp":"2026-02-24T21:09:09.620148Z","level":"WARN","fields":{"message":"blocksync: peer request triggered ledger fetch for payload","table_len":1},"target":"monad_blocksync::blocksync"}
{"timestamp":"2026-02-24T21:09:09.620166Z","level":"WARN","fields":{"message":"blocksync: ledger response cleared peer request table entry for payload","table_len":0,"removed_requesters":1},"target":"monad_blocksync::blocksync"}
{"timestamp":"2026-02-24T21:09:09.620230Z","level":"WARN","fields":{"message":"blocksync: peer request triggered ledger fetch for payload","table_len":1},"target":"monad_blocksync::blocksync"}
{"timestamp":"2026-02-24T21:09:09.620244Z","level":"WARN","fields":{"message":"blocksync: ledger response cleared peer request table entry for payload","table_len":0,"removed_requesters":1},"target":"monad_blocksync::blocksync"}
{"timestamp":"2026-02-24T21:09:09.620318Z","level":"WARN","fields":{"message":"blocksync: peer request triggered ledger fetch for payload","table_len":1},"target":"monad_blocksync::blocksync"}
{"timestamp":"2026-02-24T21:09:09.620333Z","level":"WARN","fields":{"message":"blocksync: ledger response cleared peer request table entry for payload","table_len":0,"removed_requesters":1},"target":"monad_blocksync::blocksync"}
```
The attacker was able to push 200 requests through before the connection got reset

So I think it's only guarding against a broken poll loop (e.g. poll order accidentally flipped), not blocksync flood. We can merge it if you think that's still valuable

dshulyak:
thanks for explanation, i definitely missed that we don't serve blocks from disk. i expected that we dispatch requests to db and while waiting for response new requests can be processed from raptorcast and queued.

i still think that putting caps where requests are handled is more secure, but i agree that it is impossible to abuse without significant changes in ledger

PR #2811, opened 2026-02-23
Raptorcast: deterministic protocol

xinyuan-dev:
This PR implements the deterministic raptorcast protocol ([MIP-10](https://github.com/monad-crypto/MIPs/blob/main/MIPS/MIP-10.md)) as an alternative to regular raptorcast protocol specifically designed for proposal dissemination.

The implementation is divided into commits. It's recommended to review each individually:

- the packet building & parsing capability
  + implements building & parsing & validation for v1 packets
  + this commit should look almost identical to v0 packet building
  + verify canonical encoding after decoding the message
- rebroadcast using deterministic assignment (also read https://github.com/category-labs/monad-bft/pull/2978)
  + removes recipient_hash field from chunk header because the assignment is fully deterministic
  + the chunk_id validation is now exact, and rebroadcasting targets source directly from the assignment
- the encoding commitment logic
  + adds commitment logic to a single encoding per round, this precludes encoding-level equivocation
- the rollout logic
  + implements a 4-stage gradual rollout plan

(Some tests are generated using Claude Opus 4.6)

### DIFF
--- a/monad-node-config/src/lib.rs
+++ b/monad-node-config/src/lib.rs
@@ -37,6 +37,9 @@ mod peers;
 pub mod fullnode_raptorcast;
 pub use fullnode_raptorcast::FullNodeRaptorCastConfig;
 
+pub mod raptorcast;
+pub use raptorcast::DeterministicProtocolRolloutStage;
+
 mod sync_peers;
 
 #[derive(Debug, Deserialize, Clone)]
@@ -71,6 +74,9 @@ pub struct NodeConfig<ST: CertificateSignatureRecoverable> {
     pub txpool_peer_score: ema::ScoreConfig,
 
     pub fullnode_raptorcast: FullNodeRaptorCastConfig<CertificateSignaturePubKey<ST>>,
+
+    #[serde(default)]
+    pub deterministic_raptorcast_rollout: DeterministicProtocolRolloutStage,
 
     // TODO split network-wide configuration into separate file
     ////////////////////////////////
--- a/monad-raptorcast/src/lib.rs
+++ b/monad-raptorcast/src/lib.rs
@@ -91,6 +91,7 @@ pub mod metrics;
 pub mod packet;
 pub mod parser;
 pub mod raptorcast_secondary;
+mod round_info;
 pub mod udp;
 pub mod util;
 
@@ -915,6 +916,7 @@ where
 
                         self.epoch_validators.retain(|e, _| *e + Epoch(1) >= epoch);
                     }
+                    self.udp_state.update_current_round(round);
                     self.full_node_groups.delete_expired(round);
                     self.peer_discovery_driver
                         .lock()
--- a/monad-raptorcast/src/lib.rs
+++ b/monad-raptorcast/src/lib.rs
@@ -61,7 +61,7 @@ use monad_validator::{
 };
 use packet::regular;
 use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender};
-use tracing::{debug, debug_span, error, trace, warn};
+use tracing::{debug, debug_span, error, info, trace, warn};
 use util::{
     AutoRebroadcast, BroadcastGroup, BroadcastGroupError, BuildTarget, Collector, FullNodeGroupMap,
     PeerAddrLookup, PrimaryBroadcastGroup, Recipient, Redundancy, SecondaryBroadcastGroup,
@@ -94,6 +94,7 @@ pub mod raptorcast_secondary;
 mod round_info;
 pub mod udp;
 pub mod util;
+pub mod v1_rollout;
 
 const SIGNATURE_SIZE: usize = 65;
 const DEFAULT_RETRY_ATTEMPTS: u64 = 3;
@@ -129,6 +130,7 @@ where
     current_epoch: Epoch,
 
     udp_state: udp::UdpState<ST>,
+    v1_rollout: v1_rollout::DeterministicProtocolRolloutStage,
     message_builder: OwnedMessageBuilder<ST>,
     secondary_message_builder: Option<OwnedMessageBuilder<ST>>,
 
@@ -213,6 +215,10 @@ where
         debug!(
             ?is_dynamic_fullnode, ?self_id, ?config.mtu, "RaptorCast::new",
         );
+        info!(
+            stage = %config.deterministic_protocol_rollout,
+            "deterministic raptorcast rollout",
+        );
 
         let dual_socket = auth::DualSocketHandle::new(
             auth::AuthenticatedSocketHandle::new(authenticated.0, authenticated.1),
@@ -250,6 +256,13 @@ where
             .segment_size(segment_size)
             .redundancy(secondary_redundancy);
 
+        let mut udp_state = udp::UdpState::new(
+            self_id,
+            config.udp_message_max_age_ms,
+            config.sig_verification_rate_limit,
+        );
+        udp_state.set_v1_rollout(config.deterministic_protocol_rollout);
+
         Self {
             self_id,
             is_dynamic_fullnode,
@@ -263,13 +276,13 @@ where
             message_builder,
             secondary_message_builder: Some(secondary_message_builder),
 
+            // TODO: call UpdateCurrentRound instead of pass in
+            // current_{epoch,round} as argument to allow downstream
+            // components to initialize appropriately.
             current_epoch,
 
-            udp_state: udp::UdpState::new(
-                self_id,
-                config.udp_message_max_age_ms,
-                config.sig_verification_rate_limit,
-            ),
+            udp_state,
+            v1_rollout: config.deterministic_protocol_rollout,
 
             tcp_reader,
             tcp_writer,
@@ -455,7 +468,7 @@ where
         });
 
         match target {
-            RouterTarget::Broadcast(epoch) | RouterTarget::Raptorcast(epoch) => {
+            RouterTarget::Broadcast(epoch) | RouterTarget::Raptorcast { epoch, .. } => {
                 let group = match PrimaryBroadcastGroup::of_epoch(
                     epoch,
                     &self_id, // author
@@ -488,7 +501,9 @@ where
 
                 let build_target = match &target {
                     RouterTarget::Broadcast(_) => BuildTarget::Broadcast(group),
-                    RouterTarget::Raptorcast(_) => BuildTarget::raptorcast(group),
+                    RouterTarget::Raptorcast { round, .. } => {
+                        v1_rollout::build_target(self.v1_rollout, *round, group)
+                    }
                     _ => unreachable!(),
                 };
                 let outbound_message =
@@ -791,6 +806,7 @@ where
             invite_future_dist_max: Round(5),
             invite_accept_heartbeat_ms: 100,
         },
+        deterministic_protocol_rollout: v1_rollout::CURRENT_STAGE,
     };
     let pd = PeerDiscoveryDriver::new(peer_discovery_builder);
     let shared_pd = Arc::new(Mutex::new(pd));
@@ -854,6 +870,7 @@ where
             invite_future_dist_max: Round(5),
             invite_accept_heartbeat_ms: 100,
         },
+        deterministic_protocol_rollout: v1_rollout::CURRENT_STAGE,
     };
     let pd = PeerDiscoveryDriver::new(peer_discovery_builder);
     let shared_pd = Arc::new(Mutex::new(pd));
@@ -916,6 +933,7 @@ where
 
                         self.epoch_validators.retain(|e, _| *e + Epoch(1) >= epoch);
                     }
+
                     self.udp_state.update_current_round(round);
                     self.full_node_groups.delete_expired(round);
                     self.peer_discovery_driver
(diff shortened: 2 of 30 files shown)


monad-raptorcast/src/udp.rs
```diff
@@ -91,10 +104,20 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
             decoder_cache: DecoderCache::default(),
             signature_verifier,
 
+            current_round: Round(0),
+            round_commitments: RoundCommitments::default(),
+            v1_rollout: v1_rollout::CURRENT_STAGE,
+
```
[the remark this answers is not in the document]
xinyuan-dev: this behavior is expected.

monad-raptorcast/src/packet/assigner.rs
```diff
@@ -13,633 +13,481 @@
...
-    // for easy implementation in other languages, e.g., using Mt19937
-    // and Fisher Yates shuffle.
-    pub fn shuffle_validators(validators: &mut [(NodeId<PT>, Stake)], seed: [u8; 32]) {
+    // a deterministic seed. It is required that the publisher and all
+    // validators compute the shuffling using the same seed and
+    // algorithm for deterministic raptorcast.
+    fn shuffle(&mut self, seed: [u8; 32]) {
         let mut rng = StdRng::from_seed(seed);
-        validators.shuffle(&mut rng);
+        self.validators.shuffle(&mut rng);
```
michael-yxchen, line 322: StdRng is platform dependent and future library versions might replace the algorithm. We should use chacha20 crate for determinism

monad-raptorcast/src/packet/deterministic.rs
```diff
@@ -0,0 +1,675 @@
...
+
+// Derive the seed used to shuffling validator set for deterministic raptorcast.
+// Layout: round (8) || floor(unix_ts_ms / 2048) (8) || author_pk[..16] (16) = 32 bytes
+pub fn derive_seed<PT: PubKey>(author: &NodeId<PT>, round: Round, unix_ts_ms: u64) -> [u8; 32] {
+    let author_bytes = author.pubkey().bytes();
+    let coarse_ts = unix_ts_ms / 2048; // ~2s resolution
+    let mut seed = [0u8; 32];
+    seed[..8].copy_from_slice(&round.0.to_le_bytes());
+    seed[8..16].copy_from_slice(&coarse_ts.to_le_bytes());
+    seed[16..].copy_from_slice(&author_bytes[..16]);
```
michael-yxchen: super nit: the first byte of a SECP pubkey is a tag (0x02/0x03), with less entropy

monad-raptorcast/src/packet/deterministic.rs
```diff
@@ -0,0 +1,675 @@
...
+        let num_base_symbols = layout.num_base_symbols(app_message_len);
+        stake_partition_num_chunks_hint(num_base_symbols, redundancy, validator_set_size)
+    })?;
+
+    Some(depth)
+}
+
+#[cfg(test)]
+pub(crate) fn build_with_given_header<PT, C>(
+    header: &[u8],
```
michael-yxchen, line 442: nit: could this function be merged with `build` below?

monad-raptorcast/src/udp.rs
```diff
@@ -138,20 +162,158 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
...
+    fn handle_deterministic_primary(
+        &mut self,
+        group: &PrimaryBroadcastGroup<'_, CertificateSignaturePubKey<ST>>,
+        round: Round,
+        chunk: &ValidatedChunk<CertificateSignaturePubKey<ST>>,
+        decoding_context: &DecodingContext<CertificateSignaturePubKey<ST>>,
+        rebroadcast_to: &mut impl FnMut(Vec<NodeId<CertificateSignaturePubKey<ST>>>),
+    ) -> Option<(NodeId<CertificateSignaturePubKey<ST>>, Bytes)> {
+        let Some(round_info) = self.round_info_cache.get_or_insert_primary(round) else {
```
eerkaijun, line 221: can anyone insert into `round_info_cache`? do we need to validate chunk correctness e.g. author?

monad-node-config/src/lib.rs
```diff
@@ -72,6 +75,9 @@ pub struct NodeConfig<ST: CertificateSignatureRecoverable> {
 
     pub fullnode_raptorcast: FullNodeRaptorCastConfig<CertificateSignaturePubKey<ST>>,
 
+    #[serde(default)]
+    pub deterministic_raptorcast_rollout: DeterministicProtocolRolloutStage,
```
eerkaijun, line 79: do we want this to be configurable by node operators? thinking we will bump the phase in each release, e.g. current release is `AcceptBothPublishV0`, in the next release we will bump the const to `AcceptBothPublishV1`

xinyuan-dev: under normal circumstances the node operators wouldn't need to change that, but having this as a configurable option allows a node operator to troubleshoot, e.g. if any issue is caused by the new protocol, by trying different rollout stages.

monad-raptorcast/src/udp.rs
xinyuan-dev: The author's signature and the merkle proof are validated during packet parsing, then the author is checked to be within validator set from the `PrimaryBroadcastGroup` construction, then we validate the sender (e.g. rebroadcaster) is of the message is valid through `group`. So at this time we're sure the chunk is from a rightful validator for the epoch. To further ensure only the leader can lock on the round we can further add leader info in raptorcast.

xinyuan-dev: Since it's not a trivial change. Let me implement the feature in a separate PR give raptorcast access to leader information in a separate PR before the v1 rollout (stage 3).

Update: tracked in https://github.com/category-labs/monad-bft/issues/3063

monad-raptorcast/src/util.rs
```diff
@@ -845,6 +886,18 @@ impl Redundancy {
         Redundancy(FixedU16::from_bits((num as u16) << Self::FRAC_BITS))
     }
 
+    pub const fn from_fract(
+        int_part: u8,
+        decimal_part: u8, // out of 100
```
michael-yxchen: nit: renaming to hundredth can avoid confusion

monad-raptorcast/src/parser/packet_parser.rs
```diff
@@ -387,8 +517,177 @@ impl<'a> RaptorcastPacketV0<'a> {
...
+
+        let app_message_len = self.ensure_app_message_length()?;
+        let timestamp = self.header.unix_ts_ms.get();
+        ensure_valid_timestamp(timestamp, max_age_ms)?;
+
+        let symbol_len = self.payload().len();
+        ensure!(
+            symbol_len >= deterministic::MIN_SYMBOL_LEN || symbol_len == app_message_len,
+            InvalidChunk::InvalidChunkLen
+        );
```
michael-yxchen: is `|| symbol_len == app_message_len` ever hit?

monad-raptorcast/src/round_info.rs
```diff
@@ -0,0 +1,401 @@
...
+    }
+
+    pub fn update_current_round(&mut self, round: Round) {
+        if let Some(current) = self.current_round {
+            debug_assert!(
+                round >= current,
+                "Cannot enter a past round: current {}, new {}",
+                current,
+                round
+            );
```
michael-yxchen: nit: It's a critical invariant (maybe not to this module but definitely for the entire system). Safe to bump to `assert!`

monad-raptorcast/src/round_info.rs
```diff
@@ -0,0 +1,401 @@
...
+        }
+
+        self.current_round = Some(round);
+
+        // Evict rounds from the cache
+        if let Some(cutoff_future) = round.checked_add(CACHE_MAX_FUTURE_ROUNDS) {
+            drop(self.primary.split_off(&cutoff_future));
+        };
```
michael-yxchen, line 63: just to confirm my understanding, we should never enter this branch right?

monad-raptorcast/src/lib.rs
```diff
@@ -262,13 +276,13 @@ where
             message_builder,
             secondary_message_builder: Some(secondary_message_builder),
 
+            // TODO: call UpdateCurrentRound instead of pass in
+            // current_{epoch,round} as argument to allow downstream
+            // components to initialize appropriately.
```
michael-yxchen, line 281: we can probably do that in MonadState::new()

monad-raptorcast/src/udp.rs
```diff
@@ -331,22 +493,43 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
...
                 }
-                BroadcastMode::Primary | BroadcastMode::Secondary => self.handle_broadcast(
+                BroadcastMode::Primary => {
+                    if !v1_rollout::should_accept(self.v1_rollout, &chunk) {
+                        self.metrics.executor_metrics_mut()
+                            [COUNTER_RAPTORCAST_CHUNKS_DROPPED_INCOMPATIBLE_VERSION] += 1;
+                        continue;
+                    }
+                    let accepted_metric = match chunk.version {
+                        1 => COUNTER_RAPTORCAST_V1_PRIMARY_CHUNKS_ACCEPTED,
```
michael-yxchen: can we make constants for these version numbers? in packet_parser.rs too

monad-raptorcast/src/packet/assigner.rs
michael-yxchen, line 1: I wonder if f64 shares can lead to non-determinism in chunk assignments on different platforms. Integer arithmetic removes this footgun. `num_scaled_symbols * stake` is upper bounded by `2**(MAX_MERKLE_TREE_DEPTH - 1) * 100B * 10**18 ~= 2^111` . That fits in a U128 or U256 if we want more consistent with alloy_rlp.

monad-raptorcast/src/parser/packet_parser.rs
xinyuan-dev: it's a legacy from v0 parser. in fact, any packet surviving `V1::parse` will unconditionally satisfy `symbol_len >= MIN_SYMBOL_LEN && symbol_len <= MAX_SYMBOL_LEN` as well. but let me simplify but keep the check defensively because the `symbol_len` is subsequently used as divisor.

monad-raptorcast/src/util.rs
xinyuan-dev: good suggestion. fixed.

monad-raptorcast/src/packet/deterministic.rs
xinyuan-dev: yup, in fact [kai's pr](https://github.com/category-labs/monad-bft/pull/2998/changes#diff-94dae8ce85beb92d06d1388c91e4e4a99201e55c51411a19a4c4adfd91f0e38bL441) simplifies the duplications using `InnerDeterministicEncoding` construct, which i think nicer to keep that.

monad-raptorcast/src/round_info.rs
xinyuan-dev: done. also changed `>=` to `>` for a stricter check.

monad-raptorcast/src/packet/deterministic.rs
xinyuan-dev: yup it should be acceptable. changed to `author[1..17]` regardless.

monad-raptorcast/src/round_info.rs
xinyuan-dev: before the first `update_current_round` call, the `RoundInfoCache` accepts info from any rounds, including rounds from arbitrary future. The branch can hit upon the first `update_current_round` invocation to prune round info too far into the future.

monad-raptorcast/src/udp.rs
xinyuan-dev: yup. i created an enum type for the version numbers

monad-raptorcast/src/lib.rs
xinyuan-dev: noted in https://github.com/category-labs/monad-bft/issues/3068, let's address it in a separate pr.

monad-raptorcast/src/packet/assigner.rs
xinyuan-dev: practically i don't think using f64 can leads to non-determinism. IEEE 754 operations (+, -, *, /) are bit-exact, as well as the conversions. however, i do think there is a potential for llvm to optimize the ops differently on different platforms.

let me address this issue in a separate pr as the current f64 impl is a lot simpler to reason about. it would be easier to review this single change. tracked as https://github.com/category-labs/monad-bft/issues/3070.

monad-raptorcast/src/round_info.rs
michael-yxchen: I would recommend keeping `>=`. I don't think anything can break it as Pacemaker is the single place that emits EnterRound commands. It guarantees strict monotonicity. `>=` is safer in case we emit it in other places that silently break this invariant, e.g. on init.

xinyuan-dev: is there any other cases that can break the invariant other than initialization? i think initialization is fine with the strict check as it will be recorded as a transition from `None` to `Some`.

michael-yxchen: in steady state things should be fine. I was just worries that init might double emit EnterRound for the same round. Not a concern now and we can probably fix it easily if spotted

michael-yxchen (approved):

eerkaijun (approved):

PR #2820, opened 2026-02-25
Add support for hive rpc conformance suite

es3n1n:
The rust part of the https://github.com/category-labs/monad/pull/2103 PR.

I tried to keep the number of changes small and made a new docker compose config to run specifically in the needed configuration without bloating the main single-node run.sh or compose.yaml.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,10 +77,6 @@ nets/run.sh --toolchain gcc-avx512
 
 This will start a single node with chain ID of `20143` and RPC at `localhost:8080`. The known [Foundry/Anvil accounts](https://getfoundry.sh/anvil/overview/) have each been loaded with [large initial balances](https://github.com/category-labs/monad/blob/ce4101b11701bf4ef3a9cd996a6144883735187f/category/execution/monad/chain/monad_devnet_alloc.hpp#L22). The easiest way to fund transactions is to import the private key from one of those pre-allocated accounts.
 
-For replaying Hive-generated chains through the local RPC harness used by
-`category-labs/hive-runner`, see
-[docker/single-node/hive/README.md](docker/single-node/hive/README.md).
-
 > [!TIP]
 > To avoid a lengthy rebuild after shutting down the docker containers, you can call `nets/run.sh` with the `--cached-build <full path to build dir>` arg, e.g.
 >
--- a/README.md
+++ b/README.md
@@ -77,6 +77,10 @@ nets/run.sh --toolchain gcc-avx512
 
 This will start a single node with chain ID of `20143` and RPC at `localhost:8080`. The known [Foundry/Anvil accounts](https://getfoundry.sh/anvil/overview/) have each been loaded with [large initial balances](https://github.com/category-labs/monad/blob/ce4101b11701bf4ef3a9cd996a6144883735187f/category/execution/monad/chain/monad_devnet_alloc.hpp#L22). The easiest way to fund transactions is to import the private key from one of those pre-allocated accounts.
 
+For replaying Hive-generated chains through the local RPC harness used by
+`category-labs/hive-runner`, see
+[docker/single-node/hive/README.md](docker/single-node/hive/README.md).
+
 > [!TIP]
 > To avoid a lengthy rebuild after shutting down the docker containers, you can call `nets/run.sh` with the `--cached-build <full path to build dir>` arg, e.g.
 >
@@ -107,7 +111,7 @@ Please consult the [official RPC docs](https://docs.monad.xyz/reference/) as the
 ### Using Cargo
 
 To run a Monad consensus client, follow instructions [here](monad-node/README.md).
- 
+
 To run a JsonRpc server, follow instructions [here](monad-rpc/README.md).
 
 #### Cargo targets that link Monad execution
--- a/docker/single-node/hive/run.sh
+++ b/docker/single-node/hive/run.sh
@@ -1,0 +1,98 @@
+#!/bin/bash
+
+set -ex
+
+CHAIN_RLP_PATH=""
+NBLOCKS=""
+
+usage() {
+    echo "Usage: $0 --chain-rlp /path/to/chain.rlp --nblocks <number>"
+    exit 1
+}
+
+while [[ "$#" -gt 0 ]]; do
+    case $1 in
+        --chain-rlp)
+            if [[ -z "$2" || "$2" == --* ]]; then
+                echo "Error: --chain-rlp requires a path argument." >&2
+                usage
+            fi
+            CHAIN_RLP_PATH="$2"
+            shift
+            ;;
+        --nblocks)
+            if [[ -z "$2" || "$2" == --* ]]; then
+                echo "Error: --nblocks requires a number argument." >&2
+                usage
+            fi
+            NBLOCKS="$2"
+            shift
+            ;;
+        *)
+            echo "Unknown parameter passed: $1"
+            usage
+            ;;
+    esac
+    shift
+done
+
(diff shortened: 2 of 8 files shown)


eclark0426 (dismissed):

monad-chain-config/src/lib.rs
```diff
@@ -35,6 +35,7 @@ pub const ETHEREUM_MAINNET_CHAIN_ID: u64 = 1;
 pub const MONAD_MAINNET_CHAIN_ID: u64 = 143;
 pub const MONAD_TESTNET_CHAIN_ID: u64 = 10143;
 pub const MONAD_DEVNET_CHAIN_ID: u64 = 20143;
+pub const HIVE_CHAIN_ID: u64 = 3503995874084926;
```
michael-yxchen, line 39: I'm curious if there's a reason behind this chain id?

docker/single-node/hive/run.sh
```diff
@@ -0,0 +1,95 @@
...
+export MONAD_EXECUTION_ROOT="${MONAD_BFT_ROOT}/monad-cxx/monad-execution"
+export GIT_TAG_VERSION=$(git -C "$MONAD_BFT_ROOT" describe --always)
+
+rand_hex=$(od -vAn -N8 -tx1 /dev/urandom | tr -d " \n" | cut -c 1-16)
+vol_root="logs/$(date +%Y%m%d_%H%M%S)-$rand_hex"
+
+mkdir -p "$vol_root/node/ledger"
+touch "$vol_root/node/ledger/wal"
+cp -r "$hive_dir"/* "$vol_root"
+cp -r "$devnet_dir/monad/config" "$vol_root/node"
```
michael-yxchen, line 101: `node.toml` in this directory has a chain_id of 20143. RPC gets its chain id from this file, which doesn't match `hive_net` chain_id used by execution. Did eth_call go through with this setup?

docker/single-node/hive/run.sh
michael-yxchen, line 1: Could you make this file executable?

monad-chain-config/src/lib.rs
es3n1n: good question, but im not sure. it was first introduced in https://github.com/lightclient/rpctestgen/pull/27 (the repo that later got merged into the execution-apis) though

docker/single-node/hive/run.sh
es3n1n: ah, good point. i forgot that in my env i had a different folder that i didnt commit

docker/single-node/hive/run.sh
```diff
@@ -0,0 +1,96 @@
...
+mkdir -p logs
+
+monad_bft_root="../.."
+rpc_dir="$monad_bft_root/docker/rpc"
+devnet_dir="$monad_bft_root/docker/devnet"
+hive_dir="$monad_bft_root/docker/single-node/hive"
+
+export MONAD_BFT_ROOT=$(realpath "$monad_bft_root")
+export RPC_DIR=$(realpath "$rpc_dir")
+export MONAD_EXECUTION_ROOT="${MONAD_BFT_ROOT}/monad-cxx/monad-execution"
```
michael-yxchen: we moved the submodule
```suggestion
export MONAD_EXECUTION_ROOT="${MONAD_BFT_ROOT}/monad-execution"
```

monad-chain-config/src/lib.rs
michael-yxchen: nit: can you add a comment here referencing the chain id?

docker/single-node/hive/run.sh
```diff
@@ -0,0 +1,96 @@
...
+    usage
+fi
+if [ -z "$NBLOCKS" ]; then
+    echo "Error: --nblocks is required." >&2
+    usage
+fi
+if [ ! -f "$CHAIN_RLP_PATH" ]; then
+    echo "Error: Chain RLP file does not exist: $CHAIN_RLP_PATH" >&2
+    exit 1
+fi
```
michael-yxchen: place argument validation immediately after parsing, above the mkdir block, so we don't create dir and block files if arguments are invalid

es3n1n:
Converted to draft until https://github.com/category-labs/monad/pull/2230 is merged, will update the submodule and rebase again once that is done

monad-execution
eclark0426, line 1: Execution bump can be dropped, we bumped a couple of days ago and this would move the submodule backwards atp

docker/single-node/hive/run.sh
```diff
@@ -0,0 +1,98 @@
...
+export GIT_TAG_VERSION=$(git -C "$MONAD_BFT_ROOT" describe --always)
+
+rand_hex=$(od -vAn -N8 -tx1 /dev/urandom | tr -d " \n" | cut -c 1-16)
+vol_root="logs/$(date +%Y%m%d_%H%M%S)-$rand_hex"
+
+mkdir -p "$vol_root/node/ledger"
+touch "$vol_root/node/ledger/wal"
+cp -r "$hive_dir"/* "$vol_root"
+cp -r "$devnet_dir/monad/config" "$vol_root/node"
+sed -i'' -e 's/chain_id = 20143/chain_id = 3503995874084926/' "$vol_root/node/config/node.toml"
```
eclark0426: Post-processing the devnet chain ID feels a bit brittle. Was there a reason for doing it this way vs making a separate config for Hive?

0xSqualo: We reuse the devnet node config because the Hive harness needs the same local single-node RPC shape, just with the execution-side `hive_net` chain ID so RPC and execution agree. I tightened this up so it no longer depends on the old devnet chain ID value: the script now rewrites exactly one `chain_id` entry in the generated copy and fails if the config has zero or multiple matches, which keeps the checked-in config from drifting while avoiding a mostly duplicated Hive config.

monad-execution
0xSqualo: Agreed we can drop from this, some new changes required in execution due to latest version dropping previous forks.

eclark0426 (dismissed):

michael-yxchen:
let's wait for https://github.com/category-labs/monad/pull/2338 to get merged first so we can point to an execution commit on main

eclark0426 (dismissed):
A few of the hive tests are failing, but I think those have already been brought up on Slack.

README.md
```diff
@@ -77,6 +77,10 @@ nets/run.sh --toolchain gcc-avx512
 
 This will start a single node with chain ID of `20143` and RPC at `localhost:8080`. The known [Foundry/Anvil accounts](https://getfoundry.sh/anvil/overview/) have each been loaded with [large initial balances](https://github.com/category-labs/monad/blob/ce4101b11701bf4ef3a9cd996a6144883735187f/category/execution/monad/chain/monad_devnet_alloc.hpp#L22). The easiest way to fund transactions is to import the private key from one of those pre-allocated accounts.
 
+For replaying Hive-generated chains through the local RPC harness used by
+`category-labs/hive-runner`, see
+[docker/single-node/hive/README.md](docker/single-node/hive/README.md).
+
```
eerkaijun: can we remove these from the top level README?

0xSqualo: Removed this as requested in commit e470ee92.

eclark0426 (approved):

eerkaijun (approved):

Baltoli (approved):

omegablitz (approved):

eclark0426:
CI failure here can be disregarded, it will be resolved once https://github.com/category-labs/monad-integration/pull/440 merges

PR #2837, opened 2026-03-04
Feed custom blocks for testing

philipglazman:
Adds a txpool variant that lets bft propose custom blocks for e2e testing


monad-eth-txpool-executor/src/scripted.rs
```diff
@@ -0,0 +1,499 @@
...
+    async fn try_read_block(stream: &mut UnixStream) -> io::Result<Option<Vec<TxEnvelope>>> {
+        let mut len_buf = [0u8; 4];
+        stream.read_exact(&mut len_buf).await?;
+        let msg_len = u32::from_be_bytes(len_buf) as usize;
+
+        if msg_len == 0 {
+            return Ok(Some(Vec::new()));
+        }
+
+        const MAX_MSG_LEN: usize = 4 * 1024 * 1024; // 4 MB
```
eclark0426: Where does the max size come from? I assume it's related to chain revision's `proposal_byte_limit`, so maybe should be assigned based on that rather than a separate hard-coded value here

monad-eth-txpool-executor/src/scripted.rs
```diff
@@ -0,0 +1,499 @@
...
+        }
+
+        let mut msg_buf = vec![0u8; msg_len];
+        stream.read_exact(&mut msg_buf).await?;
+
+        let txs = Vec::<TxEnvelope>::decode(&mut msg_buf.as_slice()).map_err(|e| {
+            io::Error::new(io::ErrorKind::InvalidData, format!("RLP decode error: {e}"))
+        })?;
+
+        Ok(Some(txs))
```
eclark0426: What is the expected use case for this, or more explicitly would it ever be intentionally used for things like proposing invalid transactions? We're just blindly putting the transactions in right now. If someone unintentionally adds an invalid transaction that they thought was valid, they might get confusing downstream errors from other validators rejecting the block with invalid transactions

monad-eth-txpool-executor/src/scripted.rs
philipglazman: Good point. fixed

monad-eth-txpool-executor/src/scripted.rs
philipglazman: I think having an adversarial validator is a valid workflow for using custom blocks. The expected use case is to have custom blocks with valid transactions in order to test certain workflows, but it doesn't need to be restricted to valid proposals.

monad-eth-txpool-executor/src/scripted.rs
```diff
@@ -0,0 +1,499 @@
...
+                result = Self::read_from_connection(&mut self.connection) => {
+                    match result {
+                        Ok(Some(txs)) => {
+                            info!(
+                                num_txs = txs.len(),
+                                queue_len = self.block_queue.len() + 1,
+                                "scripted txpool: received block from IPC"
+                            );
+                            self.block_queue.push_back(txs);
+                        }
```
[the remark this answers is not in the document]
philipglazman: It's a test only feature, not worth the complexity

andr-dev (dismissed):

andr-dev (dismissed):

eclark0426 (dismissed):

michael-yxchen (dismissed):

monad-eth-txpool-executor/Cargo.toml
```diff
@@ -39,12 +41,16 @@ serde = { workspace = true }
 tokio = { workspace = true }
 tracing = { workspace = true }
 
+[features]
+test-scripted-blocks = ["monad-system-calls", "tokio/io-util", "tokio/net", "dep:alloy-rlp"]
```
andr-dev: Is there a reason `monad-system-calls` is specified as is while `alloy-rlp` has the "dep:" prefix? It's been a while since I played with optional features.

andr-dev (dismissed):

monad-eth-txpool-executor/src/scripted.rs
```diff
@@ -0,0 +1,501 @@
...
+    ) -> io::Result<Option<Vec<TxEnvelope>>> {
+        match connection.as_mut() {
+            Some(conn) => Self::try_read_block(conn).await,
+            None => std::future::pending().await,
+        }
+    }
+
+    async fn try_read_block(
+        stream: &mut FramedRead<UnixStream, LengthDelimitedCodec>,
+    ) -> io::Result<Option<Vec<TxEnvelope>>> {
```
michael-yxchen: nit: I think Option<> is not necessary here

michael-yxchen (dismissed):
Comment is not blocking

eclark0426 (dismissed):

monad-node/Cargo.toml
michael-yxchen, line 1: should monad-node include a `test-scripted-blocks` feature?

michael-yxchen (commented):
Actually I wonder if a new `ScriptedTxPoolExecutorClient` type is needed. It wraps `EthTxPoolExecutorClient` type and `Executor/Stream` implementations are pass thru. If we don't plan to change those, we can move `start` to a function that returns `EthTxPoolExecutorClient`. Cleans up the TxPoolExecutor enum too. Not blocking and can go in separately

PR #2879, opened 2026-03-20
monad-node: switch to prometheus backed metrics

dshulyak:
there was some interest from operators to switch to plain prometheus, but beside that main motivation is to enable reporting from different threads without additional layers (e.g txpool and dataplane), use actual histograms for latency measurement, and labels instead of creating slightly differently named sets of metrics 

main changes:

- existing counters/gauges are stored in prometheus gauge type instead of u64 with custom plumbing, basically Arc<AtomicU64> with additional metadata. importantly increment/decrement is a simple atomic operation with additional load from arc ptr
- executor metrics interface is used to register gauges in prometheus on program startup. it obtains cheap copies to internal Arc<AtomicU64>. this removes the need to manually poll them in timer, instead they will be exposed on /metrics endpoint in a separate server after being cloned on startup
- otel collector is still preserved for smooth migration. once everybody migrated related code can be dropped

there is a number of follow up changes that we can make to support histograms/labels/seamless reporting from different threads. but i didn't include them in this pr for simplicity


dshulyak:
@claude review

monad-peer-discovery/src/lib.rs
```diff
@@ -658,9 +658,6 @@ pub enum PeerDiscoveryTimerCommand<E, ST: CertificateSignatureRecoverable> {
     },
 }
 
-#[derive(Debug, Clone)]
-pub struct PeerDiscoveryMetricsCommand(ExecutorMetrics);
```
dshulyak, line 674: since metrics are Arc<atomic64> we simply clone them and don't have to support this command

monad-executor/src/metrics.rs
```diff
@@ -46,6 +54,42 @@ impl PartialEq for MetricDef {
...
 
+fn is_valid_prometheus_metric_start(ch: char) -> bool {
+    ch.is_ascii_alphabetic() || matches!(ch, '_' | ':')
+}
+
+fn is_valid_prometheus_metric_char(ch: char) -> bool {
+    is_valid_prometheus_metric_start(ch) || ch.is_ascii_digit()
+}
+
+pub fn prometheus_metric_name(metric: &str) -> prometheus::Result<String> {
```
dshulyak, line 65: previously otel was fixing metrics into correct format

omegablitz (approved):

ariqchowdhury (approved):

PR #2945, opened 2026-04-01
 Add monad-chain-data crate to support queryX rpc methods

JoeHowarth:
[Intro Design Doc](https://www.notion.so/Bitmap-Indexed-History-Queries-32775b0ba84080a4bbb2ed1089a64a9d?source=copy_link)

These are the first 2 commits for upstreaming monad-chain-data. This crate will provide the backend for 
[QueryX](https://www.notion.so/RPC-Query-Methods-Draft-MIP-30b75b0ba840806186b9f6b2df4ccc7a?source=copy_link&utm_content=30b75b0b-a840-8061-86b9-f6b2df4ccc7a&utm_campaign=T081S0AP65N&n=slack&n=slack_link_unfurl&pvs=6)

Commits: 

1. inits crate dir and creates stub file scaffolds
2. creates naive log ingest and block-range scan query paths

The intention is to slowly build up to the completed codebase through vertical slices. 

Future slices will include:

- Accelerated queries using roaring-bitmaps and paged indexes (the main point of this crate)
- Integration into RPC and an ingest binary (eventually superseding monad-archive)
- Txs, Traces, Native Transfers, Blocks familes and query support
- Caching
- Production backends (Scylla + S3)
- Multi-writer support
- Zero-copy types

### DIFF
--- a/monad-chain-data/src/kernel/tables.rs
+++ b/monad-chain-data/src/kernel/tables.rs
@@ -1,0 +1,139 @@
+// Copyright (C) 2025 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+use crate::{
+    error::Result,
+    logs::LogBlockHeader,
+    primitives::state::{BlockRecord, PublicationState},
+    store::{BlobStore, BlobTable, BlobTableId, KvTable, MetaStore, TableId},
+};
+
+pub struct Tables<M: MetaStore, B: BlobStore> {
+    publication: PublicationTables<M>,
+    blocks: BlockTables<M>,
+    logs: LogTables<M, B>,
+}
+
+impl<M: MetaStore, B: BlobStore> Tables<M, B> {
+    pub fn new(meta_store: M, blob_store: B) -> Self {
+        Self {
+            publication: PublicationTables::new(meta_store.clone()),
+            blocks: BlockTables::new(meta_store.clone()),
+            logs: LogTables::new(meta_store, blob_store),
+        }
+    }
+
+    pub fn publication(&self) -> &PublicationTables<M> {
+        &self.publication
+    }
+
+    pub fn blocks(&self) -> &BlockTables<M> {
+        &self.blocks
+    }
+
+    pub fn logs(&self) -> &LogTables<M, B> {
+        &self.logs
+    }
+}
+
+pub struct PublicationTables<M: MetaStore> {
+    publication_state: KvTable<M>,
+}
+
+impl<M: MetaStore> PublicationTables<M> {
+    pub const PUBLICATION_STATE_TABLE: TableId = TableId::new("publication_state");
+    pub const PUBLICATION_STATE_KEY: &[u8] = b"state";
+
+    fn new(meta_store: M) -> Self {
+        Self {
+            publication_state: meta_store.table(Self::PUBLICATION_STATE_TABLE),
+        }
+    }
+
+    pub async fn load_published_head(&self) -> Result<Option<u64>> {
+        let _ = &self.publication_state;
+        todo!("load publication-state metadata for the current head")
+    }
+
+    pub async fn store_state(&self, state: PublicationState) -> Result<()> {
+        let _ = (&self.publication_state, state);
+        todo!("store publication-state metadata")
+    }
+}
+
+pub struct BlockTables<M: MetaStore> {
+    block_records: KvTable<M>,
+}
+
+impl<M: MetaStore> BlockTables<M> {
+    pub const BLOCK_RECORD_TABLE: TableId = TableId::new("block_record");
+
+    fn new(meta_store: M) -> Self {
+        Self {
+            block_records: meta_store.table(Self::BLOCK_RECORD_TABLE),
+        }
+    }
+
+    pub async fn load_record(&self, block_number: u64) -> Result<Option<BlockRecord>> {
+        let _ = (&self.block_records, block_number);
+        todo!("load a shared block record by block number")
+    }
+
+    pub async fn store_record(&self, block_number: u64, block_record: &BlockRecord) -> Result<()> {
+        let _ = (&self.block_records, block_number, block_record);
+        todo!("store a shared block record by block number")
+    }
+}
+
+pub struct LogTables<M: MetaStore, B: BlobStore> {
+    block_headers: KvTable<M>,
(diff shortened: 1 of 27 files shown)


JoeHowarth:
Addressed AI code review feedback and resolved others

 1. Non-atomic ingest — Already addressed with TODO comment.
  2. Limit not enforced — This is the spec's intended behavior (block-aligned responses). Not a bug.
  3. pub mod core shadows — Real point, addressed
  4. Dead core/types.rs — Already removed.
  5. Dead ApiSurface — Already removed.
  6. Genesis hardcoded at block 1 — Known first-pass limitation, documented in code.
  7. head + 1 overflow — would take 233,976,966,942 years, prefer cleaner `+`
  8. todo!() panics — These are in unimplemented store methods (delete, list). Fair point for trait completeness, but these aren't called by any code path today.
  9. std::sync::RwLock in async — Test-only stores, acceptable.
  10. Asymmetric clamping — Addressed
  11. Empty offsets validation — Minor, log_count() handles it gracefully.
  12. LogEntry::encode infallible — fixed
  13. PutResult ignored — Fair observation but not relevant until conditional-write backends exist.
  14. MonadChainDataError not Clone — Premature.

JoeHowarth:
Updated to use static methods on structs instead of free fns as suggested by Andre

monad-chain-data/src/primitives/page.rs
```diff
@@ -13,6 +13,8 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+pub const DEFAULT_LIMIT: usize = 100;
```
andr-dev: Default limit of what?

monad-chain-data/src/logs/materialize.rs
```diff
@@ -13,30 +13,170 @@
...
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct QueryLogsRequest {
     pub from_block: Option<u64>,
     pub to_block: Option<u64>,
     pub order: QueryOrder,
+    /// Target number of primary objects to return. The server always completes the
+    /// current block before stopping, so the actual count may exceed this value.
+    /// The server may also return fewer if an internal constraint is reached.
+    /// Defaults to [`DEFAULT_LIMIT`] (100).
     pub limit: usize,
```
andr-dev, line 50: I think we talked about renaming this field to `target_logs` or something similar since it encapsulates the essence of the "overshooting" in the comment better than "limit" does.

monad-chain-data/src/primitives/range.rs
```diff
@@ -0,0 +1,30 @@
...
+    pub to_block: BlockRef,
+}
+
+pub async fn resolve_block_window(
+    request: &QueryLogsRequest,
+    published_head: u64,
+) -> Result<ResolvedBlockWindow> {
+    let _ = (request, published_head);
+    todo!("resolve the published block window for the current query request")
+}
```
andr-dev: In the next commit, this becomes:
```rust
impl ResolvedBlockWindow {
    pub async fn resolve<M: MetaStore>(
        request: &QueryLogsRequest,
        published_head: u64,
        blocks: &BlockTables<M>,
    ) -> Result<Self> {
        ...
    }
}
```

andr-dev (dismissed):

monad-chain-data/src/logs/materialize.rs
JoeHowarth: Keeping the field to be consistent with other queryX methods in the spec but improved doc comment

monad-chain-data/src/query/runner.rs
```diff
@@ -0,0 +1,61 @@
...
+    for block_number in block_range(window, request.order) {
+        let (block_ref, block_logs) = materializer
+            .load_filtered_block_logs_for_block(block_number, request)
+            .await?;
+        logs.extend(block_logs);
+        cursor_block = block_ref;
+        if logs.len() >= target_log_count {
+            break;
+        }
+    }
```
0xSqualo, line 44: @JoeHowarth Should this also filter for a max number of blocks searched? 

To avoid the case in which a very narrow / impossible log match filter is passed and a massive block range is searched where no logs are found.

monad-chain-data/src/query/runner.rs
```diff
@@ -0,0 +1,61 @@
...
+    let mut cursor_block = window.from_block;
+
+    for block_number in block_range(window, request.order) {
+        let (block_ref, block_logs) = materializer
+            .load_filtered_block_logs_for_block(block_number, request)
+            .await?;
+        logs.extend(block_logs);
+        cursor_block = block_ref;
+        if logs.len() >= target_log_count {
+            break;
```
philipglazman, line 42: Should we warn callers that we break early ?

monad-chain-data/src/kernel/tables.rs
```diff
@@ -0,0 +1,209 @@
...
+fn block_number_key(block_number: u64) -> [u8; 8] {
+    block_number.to_be_bytes()
+}
+
+pub struct PublicationTables<M: MetaStore> {
+    publication_state: KvTable<M>,
+}
+
+impl<M: MetaStore> PublicationTables<M> {
+    pub const PUBLICATION_STATE_TABLE: TableId = TableId::new("publication_state");
```
eclark0426, line 63: How generic/widely used are the backends that the tables would be stored in? Mostly wondering if there's any concern with IDs colliding if this just mapped directly to an S3 bucket or something like that

monad-chain-data/src/query/runner.rs
JoeHowarth: Yes you're right and this lands in a later slice: jhow/chain-data/query-limits - PR #2989
The thought for this first commit was to set up a scaffold that was as simple as possible and then layer in more over time

monad-chain-data/src/query/runner.rs
JoeHowarth: It's part of the queryX spec. Basically when cursor blocks != to block the client knows it's a paginated resp

monad-chain-data/src/kernel/tables.rs
JoeHowarth: That's left up to the backend that impls the interface, but I added a comment there : )

monad-chain-data/src/logs/materialize.rs
```diff
@@ -0,0 +1,184 @@
...
+    header: &LogBlockHeader,
+    blob: &Bytes,
+    block_record: &BlockRecord,
+    request: &QueryLogsRequest,
+) -> Result<Vec<LogEntry>> {
+    let count = header.log_count();
+    let indices: Box<dyn Iterator<Item = usize>> = match request.order {
+        QueryOrder::Ascending => Box::new(0..count),
+        QueryOrder::Descending => Box::new((0..count).rev()),
+    };
```
philipglazman, line 146: Looks good. Suggesting an alternative w/o Box dyn: 

```
let indices = (0..count).map(|i| match request.order {
      QueryOrder::Ascending  => i,
      QueryOrder::Descending => count - 1 - i,
  });
```

philipglazman (dismissed):

monad-chain-data/src/logs/materialize.rs
JoeHowarth: Yeah that's cleaner. 
This gets cleaned up in a later pr though and I don't want to do another round of rebases so going to punt

PR #2946, opened 2026-04-01
Chain-Data: first indexed query path

JoeHowarth:
This PR takes the initial block-scan logs slice from the first PR and adds the first indexed-query path on top of it.

  The stack is intentionally incremental:

  - first, make logs globally addressable with stable LogId windows
  - then add a persisted primary directory so LogId -> block can be resolved
  - then add persisted per-clause bitmap fragments so indexed clauses can cheaply produce candidate IDs
  - finally, add the indexed query executor that combines those pieces and preserves the existing block-scan fallback for unindexed requests

  The result is a vertical slice of the eventual indexed logs architecture: ingest now persists enough structure to support indexed reads, and query_logs can execute through
  that indexed path when the request includes indexed clauses.

 Commit-by-Commit

1. Add global log ID windows
  Introduces typed LogId and records per-block log windows as first_log_id + count in BlockRecord.
  This gives logs a stable global primary-ID space and proves ID continuity across empty blocks.

2.  Persist primary directory fragments
  Adds persisted primary-directory fragments that map ranges of log IDs back to block numbers, partitioned by 10k-ID sub-buckets.
  This is the first half of the indexed read path: resolving LogId -> block efficiently.

3.  Persist per-block bitmap fragments
  Adds per-clause roaring bitmap fragments for address/topic streams, including shard/page keying and bitmap blob encoding.
  This is the second half of the indexed read path: resolving indexed clauses into candidate LogIds.

4. Add indexed log query execution
  Adds log-window resolution, bitmap loading/intersection, directory-backed LogId resolution, and the indexed executor behind query_logs.
  Indexed queries now use the new artifacts, while unindexed queries still fall back to the original block-scan path.


monad-chain-data/src/kernel/bitmap.rs
```diff
@@ -0,0 +1,193 @@
...
+use bytes::Bytes;
+use roaring::RoaringBitmap;
+
+use crate::{
+    error::{MonadChainDataError, Result},
+    primitives::state::LogId,
+};
+
+pub const LOCAL_ID_BITS: u32 = 24;
+pub const STREAM_PAGE_LOCAL_ID_SPAN: u32 = 65_536;
```
andr-dev, line 27: Clearer as `64 * 1024`.

monad-chain-data/src/kernel/bitmap.rs
```diff
@@ -0,0 +1,193 @@
...
+
+    let topic_kinds = ["topic0", "topic1", "topic2", "topic3"];
+    for (topic, kind) in topics.iter().zip(topic_kinds) {
+        entries.push((sharded_stream_id(kind, topic.as_slice(), shard), local));
+    }
+
+    entries
+}
+
+fn compacted_bitmap_blob(
```
andr-dev, line 175: Method on `RoaringBitmap`?

monad-chain-data/src/query/indexed.rs
```diff
@@ -0,0 +1,199 @@
...
+    store::{BlobStore, MetaStore},
+};
+
+/// Resolves a global log ID to its block number and position within that block.
+///
+/// Caches directory sub-bucket fragments to avoid redundant lookups when
+/// resolving many IDs from the same region.
+struct LogIdResolver<'a, M: MetaStore, B: BlobStore> {
+    tables: &'a Tables<M, B>,
+    fragment_cache: HashMap<u64, Vec<PrimaryDirFragment>>,
```
andr-dev, line 38: I'm not seeing a strong reason to couple LogId resolution and the caching layer. I think the cleanest approach is to "wrap" the base object (the resolver) with some cache object and have both of them implement the same trait. This also provides users of the library the most flexibility when it comes to customizing caching layers / approaches.

monad-chain-data/src/query/indexed.rs
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    async fn resolve(&mut self, log_id: LogId) -> Result<Option<ResolvedLogLocation>> {
+        let bucket = sub_bucket_start(log_id.as_u64());
+        if let Entry::Vacant(entry) = self.fragment_cache.entry(bucket) {
+            entry.insert(self.tables.logs().load_sub_bucket_fragments(bucket).await?);
+        }
+
+        let Some(fragments) = self.fragment_cache.get(&bucket) else {
+            return Ok(None);
```
andr-dev, line 56: I believe this *should* be dead code and thus probably result in an error.

monad-chain-data/src/query/indexed.rs
```diff
@@ -0,0 +1,199 @@
...
+        };
+        let Some(fragment) = fragments.iter().find(|f| {
+            log_id.as_u64() >= f.first_primary_id && log_id.as_u64() < f.end_primary_id_exclusive
```
andr-dev, line 59: Helper method on `LogId` that produces `Option<u64>` containing `log_block_idx` when this condition is satisfied?

monad-chain-data/src/query/indexed.rs
```diff
@@ -0,0 +1,199 @@
...
+            debug_assert!(
+                request.filter.matches(&log),
+                "indexed candidate at block {} idx {} should match filter",
+                location.block_number,
+                location.log_block_idx,
+            );
+
+            if stop_after_block.is_none() && logs.len() >= request.limit {
+                stop_after_block = Some(location.block_number);
+            }
```
philipglazman: duplicate code. same as L148

monad-chain-data/src/query/indexed.rs
JoeHowarth: Remove whole cache, coming later, fixed scannable kv cache

monad-chain-data/src/query/indexed.rs
JoeHowarth: Yes, on type

monad-chain-data/src/primitives/state.rs
```diff
@@ -20,11 +20,68 @@ use crate::{
...
+    pub const fn shard(self) -> u64 {
+        self.0 >> Self::LOCAL_ID_BITS
+    }
+
+    pub const fn local(self) -> u32 {
+        (self.0 & Self::LOCAL_ID_MASK) as u32
+    }
+
+    pub const fn from_parts(shard: u64, local: u32) -> Self {
+        Self((shard << Self::LOCAL_ID_BITS) | (local as u64))
```
philipglazman, line 55: should we use `checked_shl` for shard?

philipglazman (approved):

PR #2970, opened 2026-04-13
Raptorcast: Round-robin mode for stake-proportional chunk assigner

xinyuan-dev:
This PR adds the ability to assign chunks in round-robin fashion in `packet::assigner::StakeBasedWithRC`, which is used in deterministic rc (https://github.com/category-labs/monad-bft/pull/2811).

### DIFF
--- a/monad-raptorcast/src/packet/assigner.rs
+++ b/monad-raptorcast/src/packet/assigner.rs
@@ -13,7 +13,10 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-use std::{collections::HashMap, ops::Range};
+use std::{
+    collections::{HashMap, VecDeque},
+    ops::Range,
+};
 
 use bytes::BytesMut;
 use monad_crypto::certificate_signature::PubKey;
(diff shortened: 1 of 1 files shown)


monad-raptorcast/src/packet/assigner.rs
```diff
@@ -743,6 +812,88 @@ mod tests {
...
+            );
+        }
+    }
+
+    #[test]
+    fn test_stake_with_rc_round_robin() {
+        // With equal-stake validators, round-robin should alternate recipients.
+        let validator_set = vec![
+            (node_id(1), Stake::from(U256::from(1))), // gets 1/3 of the chunks
+            (node_id(2), Stake::from(U256::from(2))), // gets 2/3 of the chunks
```
eerkaijun, line 866: just wondering, comment says equal stake but stake assignment is not equal?

eerkaijun (dismissed):

monad-raptorcast/src/packet/assigner.rs
xinyuan-dev: that was an outdated comment, i updated the test since. let me remove that.

eerkaijun (approved):

monad-raptorcast/src/packet/assigner.rs
```diff
@@ -479,50 +490,108 @@ impl<PT: PubKey> StakeBasedWithRC<PT> {
...
+            .iter()
+            .zip(obligation.rc.iter())
+            .enumerate()
+            .map(|(val_idx, (ic, rc))| (val_idx, *ic + *rc as usize))
+            .collect();
+
+        let mut chunk_id = 0;
+        while !remaining.is_empty() {
+            // optimization to avoid iterating over the whole list
+            // when only one validator has remaining obligation
```
dshulyak, line 550: what does it optimize? wouldn't retain_mut do one operation as well if len is 1 and order is round robin?

dshulyak (approved):

monad-raptorcast/src/packet/assigner.rs
xinyuan-dev: it's probably premature optimization - if there is a validator with larger stakes than others, this trick could save the final looping over the single remaining node.

michael-yxchen (approved):

monad-raptorcast/src/packet/assigner.rs
dshulyak: makes sense, i missed that it pushes a range chunk_id..(chunk_id + rem) instead of chunk_id + 1

PR #2974, opened 2026-04-13
Use execution dep install scripts in dockerfiles

andr-dev:


docker/devnet/Dockerfile
michael-yxchen, line 1: gcc15/g++15 are from install-tools but aren't included in devnet/rpc dockerfile. How does the build work without them?

michael-yxchen (commented):
Build dependencies are now leaked into the runner stage, because everything is installed to the base image. It's better to keep them separated

andr-dev:
> Build dependencies are now leaked into the runner stage, because everything is installed to the base image. It's better to keep them separated

Yeah I was thinking about splitting the install scripts based on what's needed at runtime vs what's needed to compile so we can have that clear separation.

PR #2976, opened 2026-04-14
Chain Data: add logId resolution index

JoeHowarth:
Adds a logId -> block number index to improve log materialization.

This index is created every 10k logIds currently.

### DIFF
--- a/monad-chain-data/src/ingest/directory_compaction.rs
+++ b/monad-chain-data/src/ingest/directory_compaction.rs
@@ -1,0 +1,190 @@
+// Copyright (C) 2025 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+use crate::{
+    error::{MonadChainDataError, Result},
+    kernel::{
+        primary_dir::{
+            bucket_start, PrimaryDirBucket, PrimaryDirFragment, PrimaryDirTables,
+            DIRECTORY_BUCKET_SIZE,
+        },
+        tables::LogTables,
+    },
+    store::{BlobStore, MetaStore},
+};
+
+/// Compacts every directory bucket sealed by the given ingest transition.
+pub(crate) async fn compact_newly_sealed_log_directory_buckets<M: MetaStore, B: BlobStore>(
+    logs: &LogTables<M, B>,
+    from_next_primary_id: u64,
+    next_primary_id: u64,
+) -> Result<()> {
+    compact_newly_sealed_buckets(logs.dir(), from_next_primary_id, next_primary_id).await
+}
+
+async fn compact_newly_sealed_buckets<M: MetaStore>(
+    dir: &PrimaryDirTables<M>,
+    from_next_primary_id: u64,
+    next_primary_id: u64,
+) -> Result<()> {
+    for sealed_bucket_start in sealed_ranges(from_next_primary_id, next_primary_id) {
+        let bucket =
+            compact_bucket_from_fragments(&dir.load_bucket_fragments(sealed_bucket_start).await?)?;
+        // Sealed fragments are intentionally retained after the compacted bucket
+        // is written. Reads prefer the bucket via `load_bucket`, so the fragments
+        // are unreferenced but kept available for replay/recovery. Eager deletion
+        // is deferred to a backend that exposes safe delete semantics.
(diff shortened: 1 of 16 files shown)


monad-chain-data/src/ingest/directory_compaction.rs
```diff
@@ -0,0 +1,214 @@
...
+            "missing sealed primary directory bucket fragments",
+        ));
+    };
+    validate_fragment(first_fragment)?;
+
+    let start_block = first_fragment.block_number;
+    let mut first_primary_ids = Vec::with_capacity(fragments.len() + 1);
+    first_primary_ids.push(first_fragment.first_primary_id);
+
+    for (index, fragment) in fragments.iter().enumerate().skip(1) {
```
andr-dev, line 63:
```rust
use itertools::Itertools;

for (previous, fragment) in fragments.tuple_windows() {
  ...
}
```

monad-chain-data/src/ingest/directory_compaction.rs
```diff
@@ -0,0 +1,214 @@
...
+            ))?,
+    );
+
+    let bucket = PrimaryDirBucket {
+        start_block,
+        first_primary_ids,
+    };
+    validate_bucket(&bucket)?;
+    Ok(bucket)
+}
```
andr-dev, line 96: In this method, we call `validate_fragment` twice and `validate_bucket` manually, this seems like an error prone approach. Is there a reason why we can't enforce through the PrimaryDirBucket type that only valid bucket objects can be created? i.e. inline the `validate_bucket` method into a `PrimaryDirBucket::new` method that produces `Result<Self, E>`?

monad-chain-data/src/kernel/primary_dir.rs
```diff
@@ -124,8 +160,27 @@ impl<M: MetaStore> PrimaryDirTables<M> {
...
+
+fn validate_bucket(bucket: &PrimaryDirBucket) -> Result<()> {
+    if bucket.first_primary_ids.len() < 2 {
+        return Err(MonadChainDataError::Decode(
+            "primary directory bucket missing sentinel",
+        ));
+    }
+    if bucket
+        .first_primary_ids
+        .windows(2)
```
andr-dev, line 175: ditto on `tuple_windows`

monad-chain-data/src/query/directory_resolver.rs
```diff
@@ -0,0 +1,193 @@
...
+
+    Ok(Some(ResolvedLogLocation {
+        block_number: bucket.start_block + entry_index as u64,
+        log_block_idx: usize::try_from(log_id.as_u64() - bucket.first_primary_ids[entry_index])
+            .map_err(|_| MonadChainDataError::Decode("log block index overflow"))?,
+    }))
+}
+
+fn containing_bucket_entry(bucket: &PrimaryDirBucket, log_id: u64) -> Option<usize> {
+    if bucket.first_primary_ids.len() < 2 {
```
andr-dev, line 111: Why are first primary ids special when there's less than 2? This is the type of stuff I'd love to see encoded into some type rather than adding a comment explaining its purpose.

monad-chain-data/src/query/directory_resolver.rs
```diff
@@ -0,0 +1,193 @@
...
+    let Some(fragment) = fragments.iter().find(|fragment| {
+        log_id.as_u64() >= fragment.first_primary_id
+            && log_id.as_u64() < fragment.end_primary_id_exclusive
+    }) else {
+        return Ok(None);
+    };
+
+    Ok(Some(ResolvedLogLocation {
+        block_number: fragment.block_number,
+        log_block_idx: usize::try_from(log_id.as_u64() - fragment.first_primary_id)
```
andr-dev, line 140: I believe this is the same logic as the helper function I mentioned adding in the previous PR.

monad-chain-data/src/ingest/directory_compaction.rs
```diff
@@ -0,0 +1,190 @@
...
+    Ok(())
+}
+
+fn compact_bucket_from_fragments(fragments: &[PrimaryDirFragment]) -> Result<PrimaryDirBucket> {
+    let Some(first_fragment) = fragments.first() else {
+        return Err(MonadChainDataError::MissingData(
+            "missing sealed primary directory bucket fragments",
+        ));
+    };
+    let last_fragment = fragments.last().expect("fragments.first() returned Some");
```
philipglazman, line 61: nit: alternatively write as: 

```
let (Some(first_fragment), Some(last_fragment)) = (fragments.first(), fragments.last()) else {
    return Err(MonadChainDataError::MissingData(
        "missing sealed primary directory bucket fragments",
    ));
};
```

monad-chain-data/src/ingest/directory_compaction.rs
```diff
@@ -0,0 +1,190 @@
...
+};
+
+/// Compacts every directory bucket sealed by the given ingest transition.
+pub(crate) async fn compact_newly_sealed_log_directory_buckets<M: MetaStore, B: BlobStore>(
+    logs: &LogTables<M, B>,
+    from_next_primary_id: u64,
+    next_primary_id: u64,
+) -> Result<()> {
+    compact_newly_sealed_buckets(logs.dir(), from_next_primary_id, next_primary_id).await
+}
```
philipglazman, line 35: why have this indirection when you can make `compact_newly_sealed_buckets` public ?

philipglazman (approved):

PR #2977, opened 2026-04-14
Chain Data: bitmap pages index and compaction

JoeHowarth:


monad-chain-data/src/ingest/bitmap_compaction.rs
```diff
@@ -0,0 +1,296 @@
...
+
+    Ok(out)
+}
+
+fn build_compaction_plan(
+    previous_open_streams: BTreeSet<String>,
+    touched_by_page: BTreeMap<u64, BTreeSet<String>>,
+    from_next_primary_id: u64,
+    next_primary_id: u64,
+) -> BitmapCompactionPlan {
```
andr-dev, line 121: This function can be called with `next_primary_id < from_next_primary_id` in which case it'll loop forever. Maybe create a range type for the two?

monad-chain-data/src/ingest/bitmap_compaction.rs
```diff
@@ -0,0 +1,296 @@
...
+            BitmapFragmentWrite, STREAM_PAGE_LOCAL_ID_SPAN,
+        },
+        tables::LogTables,
+    },
+    store::{BlobStore, MetaStore},
+};
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct BitmapCompactionPlan {
+    sealed_pages: Vec<(u64, BTreeSet<String>)>,
```
andr-dev, line 32: :)

monad-chain-data/src/kernel/bitmap.rs
```diff
@@ -194,3 +399,13 @@ fn compacted_bitmap_blob(
         bitmap,
     }))
 }
+
+fn block_number_key(block_number: u64) -> [u8; 8] {
+    block_number.to_be_bytes()
+}
```
philipglazman, line 407: Duplicate function exists in tables.rs

monad-chain-data/src/kernel/bitmap.rs
```diff
@@ -135,6 +300,33 @@ pub fn encode_grouped_bitmap_fragments(
     Ok(out)
 }
 
+/// Merges retained bitmap fragments for one page into the compacted page meta
+/// and blob artifacts written once that page seals.
+pub(crate) fn compact_bitmap_page<I, T>(
+    page_start_local: u32,
+    fragments: I,
+) -> Result<Option<(BitmapPageMeta, Bytes)>>
```
philipglazman: Should we just result Result<> instead of Result<Option<>> ? Caller returns error anyway for None

monad-chain-data/src/kernel/bitmap.rs
```diff
@@ -194,3 +399,13 @@ fn compacted_bitmap_blob(
         bitmap,
     }))
 }
+
+fn block_number_key(block_number: u64) -> [u8; 8] {
+    block_number.to_be_bytes()
+}
+
+pub(crate) fn parse_stream_shard(stream_id: &str) -> Result<u64> {
+    let shard_hex = stream_id.rsplit('/').next().unwrap_or_default();
```
philipglazman: `ok_or` instead of unwrap_or_default? Think next() will not return None on legitimate stream_ids, but we should have a clear error in case None is returned.

PR #2978, opened 2026-04-15
Raptorcast: Chunk assignment module cleanup & improvements

xinyuan-dev:
implementation-level changes:

- remove code path unused in production
- clarify the concepts of Partition/OrderedNodes/ChunkAssignment/Chunk
- rename the algorithms to EvenPartion/StakePartition
- remove Assigner trait and its usage in favor of simple concrete types

behavioral changes:

- secondary raptorcast now uses Fisher-Yates shuffling instead of rotation to shuffle group members
- secondary raptorcast now also switch to round-robin assignment, like primary raptorcast, to reduce latency
- restrict num_symbols to <= MAX_TRIPLES, stricter and more semantically correct than <= MAX_NUM_PACKETS.

the new  chunk assignment usage:

```rust
  // Primary raptorcast
  let mut partition = StakePartition::from_group(primary_group);
  partition.shuffle(seed);
  let assignment = partition.assign(num_symbols, redundancy)?;
  let chunks = assignment.materialize(segment_len, &partition);

  // Secondary raptorcast
  let mut partition = EvenPartition::from_group(secondary_group);
  partition.shuffle(seed);
  let assignment = partition.assign(num_symbols, redundancy)?;
  let chunks = assignment.materialize(segment_len, &partition);

  // Broadcast/Unicast
  let assignment = ChunkAssignment::unicast(num_symbols);
  let chunks: Vec<_> = recipients.iter()
      .flat_map(|r| assignment.materialize(segment_len, r))
      .collect();
```

recipient check & rebroadcast (for deterministic raptorcast):

```rust
  let Some(routing) = assignment.resolve_chunk_id(chunk_id, &partition) else {
      // drop invalid chunk that is not in expected assignment
      return;
  };
  if routing.recipient() == &self_node_id {
      for target in routing.rebroadcast_targets() {
          rebroadcast_to(target);
      }
  }

```

the `ChunkAssignment` and `EvenPartition/StakePartition` can be reused in the receiving path to validate the received chunks for deterministic raptorcast.

(The tests are generated using Claude Opus 4.6)


monad-raptorcast/src/packet/assigner.rs
```diff
@@ -13,554 +13,408 @@
...
+            total_stake += *stake;
         }
-    }
 
-    pub fn calc_obligation(&self, num_symbols: usize) -> Obligation {
-        let num_validators = self.validator_set.len();
-        let mut ic = vec![0usize; num_validators];
-        let mut rc = vec![false; num_validators];
+        // Guaranteed by the invariant of validator set.
+        assert!(total_stake != Stake::ZERO);
```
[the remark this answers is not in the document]
xinyuan-dev: fixed the panicking. but a group containing only the author is still valid so the method will construct a valid StakePartition instance.

eerkaijun (approved):
nice, I think this nicely sets up secondary raptorcast to switch to deterministic assignment and chunk contribution tracking

monad-raptorcast/src/packet/assigner.rs
```diff
@@ -13,633 +13,484 @@
...
 
-        let obligations = self
-            .validator_set
-            .iter()
-            .map(|(_, s)| num_symbols as f64 * (*s / self.total_stake));
+        for (node, stake) in group.iter() {
+            if node == group.author() {
+                continue;
+            }
+            let share = *stake / total_stake;
```
dshulyak, line 365: would it make sense to use scaled u128 (or even larger integer) for division and then scale back so that it is guaranteed to be deterministic on all platforms? thought we are limiting that

i think it is safer to avoid f64 in consensus code, raptorcast probably will be considered such

dshulyak (approved):

monad-raptorcast/src/packet/assigner.rs
xinyuan-dev: let's address the specific concerns individually:

- overflow: u256 has max value of ~2^256, f64's range exceeds 2^1024. the subsequent division bring the `share` to [0,1] and multiplication with `num_scaled_symbols` in modest range will not trigger overflow.
- determinism: u256 to f64 conversion is pure pure bit-manipulation; f64 division is IEEE 754 mandated to be deterministic across platforms.
- precision: f64 has 53 bits of mantissa with full precision in range [0,1] as in `share`. the multiplication with `num_scaled_symbols` in modest range should be safe.
- precision loss from u256 to f64 conversion: the max absolute error is `2^(e-53)` where `e = floor(log2(value))`. analysis for practical stake values:

| Stake | Bit-width (e) | ULP (2^(e-52)) | Max abs error | Note |
|-------|-----------|-----|---------------|------|
| 100K MON | 76 | 2^24 = 16,777,216 wei | 8.4e-12 MON | min self-delegation for active set |
| 1B MON | 89 | 2^37 = ~1.37e11 wei | 6.9e-8 MON | current largest single validator on mainnet |
| 100B MON | 96 | 2^44 = ~1.76e13 wei | 8.8e-6 MON | current total supply |
| 10,000B MON | 102 | 2^50 = ~1.13e15 wei | 5.6e-4 MON |  |

i think such the precision ranges are acceptable. we should have no problem to tolerate even orders of magnitudes higher of precision loss without issue, so long determinism is not violated and there is no risk of overflow.

eerkaijun (approved):

monad-raptorcast/src/packet/assigner.rs
dshulyak: number of chunks will be rounded at some point, but if you think there is no way to have a non-deterministic bug because of that then i don't mind

dshulyak (approved):

xinyuan-dev:
@claude review

PR #2985, opened 2026-04-17
monad-wal: bound wal size with rotating chunks

dshulyak:
3 commits
- rotate wal chunks in the driver
- move wal writer to a separate thread not to interfere with consensus https://github.com/category-labs/monad-bft/pull/3022
- macro to exclude not needed events from writing to wal (for example forwarded txs) https://github.com/category-labs/monad-bft/pull/3024

also moves writer to a different thread so that wal writing doesn't delay proposal verification. mainly about proposal, as it is the only larger message in the protocol. in this case proposal was 400KB, larger size will take longer than 5ms

image

separate thread will be mostly idle and shed load on dos

image

### DIFF
--- a/monad-wal/src/wal.rs
+++ b/monad-wal/src/wal.rs
@@ -14,15 +14,18 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 use std::{
+    collections::VecDeque,
     fmt::Debug,
-    fs::{File, OpenOptions},
-    io::Write,
+    fs::{self, File, OpenOptions},
+    io::{self, Write},
     marker::PhantomData,
-    path::PathBuf,
+    path::{Path, PathBuf},
+    time::{Duration, SystemTime, UNIX_EPOCH},
 };
 
 use bytes::Bytes;
 use monad_types::Serializable;
+use tracing::debug;
 
 use crate::WALError;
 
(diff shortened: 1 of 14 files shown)


dshulyak:
@claude review

monad-wal/src/wal.rs
```diff
@@ -102,52 +134,187 @@ where
...
+        self.trim_oldest_chunks()?;
+        Ok(())
+    }
+
+    fn trim_oldest_chunks(&mut self) -> Result<(), WALError> {
+        while self.rotated.len() as u64 + 1 > self.chunks {
+            let Some(oldest) = self.rotated.pop_front() else {
+                break;
+            };
+            fs::remove_file(oldest.path)?;
```
omegablitz, line 193: Is there any likely chance this fails?
Only possibility I can think of is a permission issue. We could check perms on each chunk in discover_chunk but might be overkill.

omegablitz (dismissed):

monad-wal/src/wal.rs
dshulyak: there is a call to create_all for {SOURCE}/wal/ directory. so it will fail on startup if there are permissions issues. 

could be that permissions will change between restarts, but seems exiting during runtime seems good enough in such case

ariqchowdhury (dismissed):

omegablitz (approved):

ariqchowdhury (approved):

PR #2986, opened 2026-04-17
Cover log query block-alignment and remove redundant indexed limit check

JoeHowarth:
Pin the queryX spec invariant that the server always completes the current block before stopping, with three black-box tests covering both indexed and block-scan paths at exactly-limit and mid-block-limit boundaries.

The indexed query loop had two consecutive checks that both attempted to set stop_after_block when logs.len() crossed the limit. The pre-push check could never fire first: between iterations, stop_after_block remains None only if the previous iteration's post-push check did not fire, which means logs.len() < limit before the current push. The redundant check is removed.


JoeHowarth:
Closing after stack cleanup: the block-alignment coverage has been folded into the earlier indexed query execution commit, so this standalone PR node is no longer needed.

PR #2992, opened 2026-04-18
Join blocks onto query_logs pages

JoeHowarth:
Adds the blocks relation to query_logs: when the caller sets relations.blocks on the request, the response carries a deduped, ascending-by-number vec of `Block { hash, header }` entries for the blocks that contributed logs in this page. Hash comes from the BlockRecord since `alloy_consensus::Header` does not carry the hash.

Relations are packaged in a dedicated `LogsRelations` struct so adding the next relation (transactions, traces) extends the shape without churning the request envelope. Block-side types and helpers live in a new `src/blocks.rs` module — the natural home for cross-family block-query machinery.

### DIFF
--- a/monad-chain-data/src/api.rs
+++ b/monad-chain-data/src/api.rs
@@ -14,6 +14,7 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 use crate::{
+    blocks::load_blocks_for_logs,
     error::{MonadChainDataError, Result},
     family::FinalizedBlock,
     ingest::{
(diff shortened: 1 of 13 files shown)


monad-chain-data/src/logs/materialize.rs
```diff
@@ -75,6 +89,9 @@ impl Default for QueryLogsRequest {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct QueryLogsResponse {
     pub logs: Vec<LogEntry>,
+    /// Deduped blocks for the `logs` in this page, sorted ascending by
+    /// block number. Empty unless `QueryLogsRequest::relations.blocks`.
+    pub blocks: Vec<Block>,
```
JoeHowarth: This should be an Option<Vec>

monad-chain-data/src/blocks.rs
```diff
@@ -0,0 +1,63 @@
...
+/// caller opts in.
+pub async fn load_blocks_for_logs<M: MetaStore>(
+    blocks: &BlockTables<M>,
+    logs: &[LogEntry],
+) -> Result<Vec<Block>> {
+    let distinct: BTreeSet<u64> = logs.iter().map(|l| l.block_number).collect();
+    let mut out = Vec::with_capacity(distinct.len());
+    for number in distinct {
+        out.push(load_block(blocks, number).await?);
+    }
```
philipglazman, line 46: Should we switch to concurrent reads for loading blocks?

monad-chain-data/src/api.rs
```diff
@@ -161,10 +162,17 @@ impl<M: MetaStore, B: BlobStore> MonadChainDataService<M, B> {
...
-            return execute_indexed_log_query(&self.tables, &request, window).await;
+        let mut response = if request.filter.has_indexed_clause() {
+            execute_indexed_log_query(&self.tables, &request, window).await?
+        } else {
+            execute_block_scan_query(&self.tables, &request, window).await?
+        };
+
+        if request.relations.blocks {
+            response.blocks =
+                Some(load_blocks_for_logs(self.tables.blocks(), &response.logs).await?);
```
philipglazman, line 173: nit: if `response.logs.is_empty()` we don't need to load_blocks_for_logs

philipglazman (approved):
approving, but see two comments

PR #2993, opened 2026-04-18
Hoist shared query envelope primitives

JoeHowarth:
Extracts the shared query envelope primitives used by `query_logs` and the upcoming `query_blocks` path.

- Adds `QueryEnvelope` as the common `(from_block, to_block, order, limit)` carrier for range-style query requests. `QueryLogsRequest` now embeds this envelope instead of owning those fields directly.
- Updates `ResolvedBlockWindow::resolve` to take a `QueryEnvelope` reference, removing its dependency on the log-specific request type while preserving queryX from/to semantics.
- Hoists `ResolvedBlockWindow::iter(order)` from the local query runner helper so range executors can share traversal behavior.
- Adds `BlockSpan` for response pagination metadata and moves `QueryLimits::check_limit` into the shared limits primitive.

No behavior change: the query_logs path is still the only caller in this PR, and its semantics are unchanged.

### DIFF
--- a/monad-chain-data/src/primitives/range.rs
+++ b/monad-chain-data/src/primitives/range.rs
@@ -16,9 +16,8 @@
 use crate::{
     error::{MonadChainDataError, Result},
     kernel::tables::BlockTables,
-    logs::QueryLogsRequest,
     primitives::{
-        limits::{LimitExceededKind, QueryLimits},
+        limits::{LimitExceededKind, QueryEnvelope, QueryLimits},
         page::QueryOrder,
         refs::BlockRef,
     },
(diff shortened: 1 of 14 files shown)


monad-chain-data/src/primitives/range.rs
```diff
@@ -42,12 +41,15 @@ impl ResolvedBlockWindow {
     /// short-circuits if the resolved span exceeds `limits.max_block_range`,
     /// then loads the per-bound block records.
     pub async fn resolve<M: MetaStore>(
-        request: &QueryLogsRequest,
+        from_block: Option<u64>,
```
JoeHowarth: Group into type, embed in Req

philipglazman (approved):

PR #2994, opened 2026-04-18
Add query_blocks substrate over the block-header artifact

JoeHowarth:
Adds a finalized range-scan query for block headers mirroring the envelope of query_logs: queryX spec from/to semantics, limit, and cursor_block pagination. The spec does not define filters or relations for `eth_queryBlocks`, so the request carries only the range envelope.

Types and executor live in `src/blocks.rs` alongside the block-relation helper from the previous commit.

### DIFF
--- a/monad-chain-data/src/blocks.rs
+++ b/monad-chain-data/src/blocks.rs
@@ -18,10 +18,16 @@ use std::collections::BTreeSet;
 use crate::{
     error::{MonadChainDataError, Result},
     family::Hash32,
-    kernel::tables::BlockTables,
+    kernel::tables::{BlockTables, Tables},
     logs::LogEntry,
-    primitives::EvmBlockHeader,
-    store::MetaStore,
+    primitives::{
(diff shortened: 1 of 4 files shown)


monad-chain-data/src/blocks.rs
```diff
@@ -32,6 +37,77 @@ pub struct Block {
     pub header: EvmBlockHeader,
 }
 
+/// Public block query in queryX spec semantics: `from_block`/`to_block` are
+/// the inclusive range start/end, with the lower/upper roles depending on
+/// `order`. Omitted bounds default to `"earliest"`/`"latest"` per order.
+/// The spec does not support filters or relations on `eth_queryBlocks`.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct QueryBlocksRequest {
```
JoeHowarth, line 46: Should be a common Req with Family specific req having it as a field. Same for Resp

philipglazman (approved):

PR #2996, opened 2026-04-20
Fix ConsensusState Debug/PartialEq impls

michael-yxchen:


eerkaijun (approved):

monad-consensus-state/src/lib.rs
```diff
@@ -122,9 +122,17 @@ where
...
+            && self.scheduled_vote.eq(&other.scheduled_vote)
             && self.safety.eq(&other.safety)
             && self.pacemaker.eq(&other.pacemaker)
+            && self.block_sync_requests.eq(&other.block_sync_requests)
+            && self
+                .canonical_proposed_tip
+                .eq(&other.canonical_proposed_tip)
+            && self
+                .vote_delay_timer_start
+                .eq(&other.vote_delay_timer_start)
```
dshulyak, line 135: we can't derive PartialEq ?

dshulyak (approved):

omegablitz (approved):

monad-consensus-state/src/lib.rs
michael-yxchen: some generics, e.g. StateBackend can't have a PartialEq bound b/c `InMemoryState` is an Arc<Mutex<_>>, which has no PartialEq

PR #2997, opened 2026-04-20
Use TxHash and BlockHash aliases in RPC

andr-dev:
`ExecEventsBuffer` was using `FixedData<32>` for block and tx hashes when alloy aliases both to `FixedBytes<32>`. Changed these aliases and updated `DataProvider` for consistency.

### DIFF
--- a/monad-rpc/src/data/buffer.rs
+++ b/monad-rpc/src/data/buffer.rs
@@ -22,7 +22,7 @@ use std::{
 };
 
 use alloy_consensus::TxEnvelope;
-use alloy_primitives::Bloom;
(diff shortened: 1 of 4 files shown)


andr-dev:
> `BlockTagOrHash::Hash` still wraps `EthHash` (`FixedData<32>`), requiring manual conversion to `BlockHash` (`FixedBytes<32>`). 

Not ideal but the types in `monad-rpc/src/types/eth_json.rs` haven't been cleaned up yet. Changing the `BlockTagOrHash::Hash` variant here would change deserialization so deferring.

philipglazman (approved):

JoeHowarth (approved):

PR #2998, opened 2026-04-21
Kai/secondary deterministic raptorcast

eerkaijun:
Integrate secondary raptorcast to also follow the rollout plan as primary raptorcast to use deterministic encoding. Tested on flexnet with AlwaysV1 enabled. Probably worth reviewing commit by commit. 

There are some code which can be further modularized in PrimaryEncoding and SecondaryEncoding. Right now the main difference is just using different chunk assignments (stake weighted vs even partitioned). Can address in a separate PR to keep this PR simple enough.

This PR also makes the node operator controlled full node redundancy obselete, to be removed from node.toml in a future PR.

### DIFF
--- a/monad-raptorcast/src/packet/assigner.rs
+++ b/monad-raptorcast/src/packet/assigner.rs
@@ -297,6 +297,17 @@ impl<PT: PubKey> EvenPartition<PT> {
 
         Ok(assignment)
     }
+
+    pub fn num_chunks(&self, num_base_symbols: usize, redundancy: Redundancy) -> Option<usize> {
+        even_partition_num_chunks(num_base_symbols, redundancy)
+    }
+}
+
+#[inline(always)]
+pub fn even_partition_num_chunks(num_base_symbols: usize, redundancy: Redundancy) -> Option<usize> {
+    // EvenPartition::assign emits exactly redundancy.scale(num_base_symbols)
+    // chunks regardless of group size.
+    redundancy.scale(num_base_symbols)
 }
 
 // Proportional to stake, plus each validator gets an optional
--- a/monad-raptorcast/src/packet/deterministic.rs
+++ b/monad-raptorcast/src/packet/deterministic.rs
@@ -290,25 +290,37 @@ where
     Ok(buffer.freeze())
 }
 
-pub(super) struct GlobalMerkleTree<'a, PT: PubKey> {
-    chunks: &'a mut [Chunk<PT>],
-    tree: MerkleTree,
+// Shared inner pipeline for deterministic raptorcast encoding. Given
+// an assigned set of chunks for a group, this struct encodes Raptor
+// symbols into them and builds the global merkle tree, yielding the
+// global merkle root. Callers can then compute a header, write
+// headers and merkle proofs into each chunk, and emit them.
+pub(super) struct InnerDeterministicEncoding<PT: PubKey> {
     layout: PacketLayout,
+    chunks: Vec<Chunk<PT>>,
+    tree: MerkleTree,
 }
 
-impl<'a, PT: PubKey> GlobalMerkleTree<'a, PT> {
-    fn build(chunks: &'a mut [Chunk<PT>], layout: PacketLayout) -> Result<Self, BuildError> {
+impl<PT: PubKey> InnerDeterministicEncoding<PT> {
+    pub fn new(
+        layout: PacketLayout,
+        mut chunks: Vec<Chunk<PT>>,
+        app_message: &[u8],
+    ) -> Result<Self, BuildError> {
+        // Encode Raptor symbols into each chunk's payload.
+        encode_unique_symbols(app_message, &mut chunks, layout)?;
+
+        // Build the global merkle tree over (chunk header + symbol).
         chunks.sort_by_key(|c| c.chunk_id());
 
-        let mut hashes = Vec::with_capacity(chunks.len());
         let depth = layout.merkle_tree_depth();
         debug_assert!(chunks.len() <= 2usize.pow((depth - 1) as u32));
 
+        let mut hashes = Vec::with_capacity(chunks.len());
         for (leaf_index, chunk) in chunks.iter_mut().enumerate() {
             if chunk.chunk_id() != leaf_index {
                 return Err(BuildError::NonContiguousChunkIds);
             }
-
             layout.write_chunk_header(chunk)?;
             let hash = layout.chunk_hash(chunk);
             hashes.push(hash);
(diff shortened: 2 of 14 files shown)


monad-raptorcast/src/packet/deterministic.rs
```diff
@@ -251,14 +255,25 @@ where
...
+    // For Secondary, the header's round field identifies the group;
+    // the epoch field is unused and set to 0.
+    let epoch_bytes: [u8; 8] = match (broadcast_mode, group_id) {
+        (BroadcastMode::Primary, GroupId::Primary(epoch)) => epoch.0.to_le_bytes(),
+        (BroadcastMode::Secondary, GroupId::Secondary(group_round)) => {
+            // The group_id's round must match the encoding scheme's round.
+            if group_round != round {
+                return Err(BuildError::InvalidGroupId(group_id));
+            }
+            0u64.to_le_bytes()
```
xinyuan-dev: i'm thinking maybe we can keep `epoch` and `round` fields in both primary and secondary packet formats? the `epoch` field can still be used to locate a validator set, which can be useful for e.g. signature verification rate limiting, or prioritizing the decoding of validator chunks. These usages are beyond the notion of "group id".

monad-raptorcast/src/round_info.rs
```diff
@@ -149,36 +182,90 @@ impl<PT: PubKey> PrimaryRoundInfo<PT> {
...
 
-        let Some(commitment) = &mut self.commitment else {
-            // no commitment for this round yet, so we will commit to
-            // the first claim we see.
-            self.commitment = Some(EncodingCommitment::from(claim));
-            return Some(());
-        };
-        if commitment.is_compatible_with(claim) {
-            return Some(());
+pub struct SecondaryRoundInfo<PT: PubKey> {
```
xinyuan-dev: nit: since for secondary the encoding assignment/commitment is not actually per-round, but per-(round, publisher) tuple, maybe we should use a different name to reflect this difference.

monad-raptorcast/src/packet/deterministic.rs
eerkaijun: my initial thought is that epoch can be derived from rounds, but I guess it's cleaner to just populate the `epoch` field as well. Updated the code. `epoch` is currently not used by secondary raptorcast, but as you said can be extended if we want to

eerkaijun:
@claude review

monad-raptorcast/src/packet/assigner.rs
```diff
@@ -296,6 +296,24 @@ impl<PT: PubKey> EvenPartition<PT> {
...
+        &self,
+        num_base_symbols: usize,
+        redundancy: Redundancy,
+    ) -> Option<usize> {
+        even_partition_num_chunks_hint(num_base_symbols, redundancy)
+    }
+}
+
+#[inline(always)]
+pub fn even_partition_num_chunks_hint(
```
xinyuan-dev: we can probably kill the `_hint` suffix here. it was used in stake partition algorithm because the actual number of rounding chunks depend on the specific ordering of validator set, which requires the seed and adds a lot of complexity if we want the exact number. So some of the code simply use the heuristic that `#rounding_chunks < #valset`. for even partition we don't use rounding chunks at all, thus the number of chunks can be calculated exactly.

monad-raptorcast/src/packet/deterministic.rs
```diff
@@ -208,6 +211,7 @@ fn build_header<ST>(
     encoding_scheme: EncodingScheme,
     merkle_tree_depth: u8,
     group_id: GroupId,
```
xinyuan-dev: what about we just pass in round & epoch as args (i.e. remove group id)? the group type can be inferred from `broadcast_mode` already.

monad-raptorcast/src/round_info.rs
```diff
@@ -149,36 +182,90 @@ impl<PT: PubKey> PrimaryRoundInfo<PT> {
...
+
+        if let Some((encoding, assignment)) = &self.encoding {
+            return assignment.resolve_chunk_id(chunk.chunk_id as usize, encoding.partition());
+        }
+
         None
     }
+
+    // Returns None if there is a conflicting commitment suggesting
+    // publisher equivocation.
```
xinyuan-dev, line 229: "publisher equivocation" in the sense that a node receives a proposal of different merkle root is something to be expected for secondary raptorcast because it can receive the same proposals re-encoded from multiple validators as publishers. since the round leader's identity is lost from the re-encoding, i think we probably can't commit to a canonical encoding like we did for primary. we can scope the commitment by validator though, but i'm not sure how useful it is.

monad-raptorcast/src/util.rs
```diff
@@ -80,9 +84,28 @@ impl<'a, PT: PubKey> BuildTarget<'a, PT> {
...
+        BuildTarget::FullNodeRaptorCast {
+            group,
+            mode: RaptorcastMode::Regular,
+        }
+    }
+
+    pub fn deterministic_fullnode_raptorcast(
+        group: SecondaryBroadcastGroup<'a, PT>,
+        epoch: Epoch,
+        round: Round,
```
xinyuan-dev: the `round` can be acquired from `SecondaryBroadcastGroup` which is guaranteed correspond to the group by its type invariant.

monad-raptorcast/src/packet/deterministic.rs
```diff
@@ -415,6 +458,76 @@ impl<PT: PubKey> PrimaryEncoding<PT> {
...
+        let depth = optimal_merkle_tree_depth(|d| {
+            let layout = PacketLayout::new(segment_len, d);
+            let num_base_symbols = layout.num_base_symbols(app_message_len);
+            partition.num_chunks(num_base_symbols, redundancy)
+        })
+        .ok_or(BuildError::MerkleTreeTooDeep)?;
+        let layout = PacketLayout::new(segment_len, depth);
+
+        // Shuffle the full-node group members.
+        let seed = derive_seed(group.publisher(), round, unix_ts_ms);
```
xinyuan-dev, line 482: the seed derivation input is carefully tuned to primary raptorcast based on its threat model. which is different for secondary, i think we can use a different seed derivation here. reusing the algorithm for secondary should be harmless, but IMO we may put a note here that the choice is coincidental, not on purpose.

eerkaijun: added a comment, let me know if you prefer other seed too

monad-raptorcast/src/round_info.rs
eerkaijun: the `self.secondary` is keyed by the publisher id and round

eerkaijun:
@claude review

xinyuan-dev (approved):

monad-raptorcast/src/parser/packet_parser.rs
```diff
@@ -621,11 +639,17 @@ impl<'a> RaptorcastPacketV1<'a> {
...
+                num_source_symbols,
+                deterministic::DEFAULT_REDUNDANCY,
+                MAX_VALIDATOR_SET_SIZE,
+            ),
+            BroadcastMode::Secondary => {
+                even_partition_num_chunks(num_source_symbols, deterministic::DEFAULT_REDUNDANCY)
+            }
+            BroadcastMode::Unspecified => None,
+        }
         .ok_or(InvalidChunk::InvalidChunkId)?;
```
michael-yxchen, line 653: nit: maybe we can differentiate InvalidChunkId and InvalidBroadcastMode

monad-raptorcast/src/round_info.rs
```diff
@@ -35,13 +35,18 @@ const CACHE_MAX_PAST_ROUNDS: Round = Round(100);
 pub struct RoundInfoCache<PT: PubKey> {
     current_round: Option<Round>,
     primary: BTreeMap<Round, PrimaryRoundInfo<PT>>,
+    // Per-publisher secondary round info. Multiple validators can
+    // publish secondary broadcasts in the same round to independent
+    // full-node groups, so we key by publisher as well.
+    secondary: HashMap<NodeId<PT>, BTreeMap<Round, SecondaryGroupRoundInfo<PT>>>,
```
michael-yxchen, line 41: ~~nit: generally we prefer BTreeMap over HashMap when key is NodeId, as another hashing is unnecessary. I'm writing some benchmark to verify that~~

nvm, ignore. it seems that HashMap is actually faster because Hash is a transmute but comparator is a slow ffi call

michael-yxchen (approved):

monad-raptorcast/src/parser/packet_parser.rs
eerkaijun: sounds good, I'll update this in a follow up PR which also reduces duplicated code between primary and secondary

PR #3006, opened 2026-04-23
Fix gas estimation for eth_fillTransaction

eerkaijun:
From my understanding, we don't need to have separate behavior for gas estimation in eth_estimateGas and eth_fillTransaction. This commit makes them the same behaviour. 

Previously the subtle difference is that when calling an external contract, only 63/64 gas is forwarded. With the 1/64 reservation the inner frame runs out of gas. The inner OOG causes success=false, which is converted to ExecutionError

Confirmed on a testnet node this fixes discrepancies between eth_estimateGas and eth_fillTransaction

### DIFF
--- a/monad-rpc/src/handlers/eth/gas.rs
+++ b/monad-rpc/src/handlers/eth/gas.rs
@@ -63,35 +63,12 @@ fn block_key_to_parts(block_key: BlockKey) -> (u64, Option<[u8; 32]>) {
     }
 }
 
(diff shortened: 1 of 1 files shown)


haythemsellami (approved):
This was initially done to allow `eth_fillTransaction` on reverting transactions.

I guess it is cleaner to just have the same behavior as in `eth_estimateGas`.

ariqchowdhury (approved):

PR #3011, opened 2026-04-23
monad-wireauth: cache cookies by public key

dshulyak:
Handshake sessions expire after 10–11s by default, but cookies can remain valid for up to 120s and were lost with the session.

When both peers require cookies, this can cause repeated challenges: B loses its cookie before A retries, so A challenges B again. The cycle can exhaust A’s retries before the handshake completes.

This change caches cookies by peer public key, preserving them across session expiry, reconnects, and both handshake directions.

The configurable LRU defaults to 256,000 entries. Each pending handshake accepts one valid cookie reply; duplicates cannot refresh recency.


dshulyak:
@claude

PR #3012, opened 2026-04-23
monad-wireauth: split pending and transport limits

dshulyak:
Before filter allowed all requests while below low watermark. It effectively disabled ip limit while node had bellow 10k sessions, so one ip could accumulate that many within large interval of time. 

In this change we split pending (accepted/sent) and established. For established the total ip limit is enforced since the start, with low watermark threshold being removed.

For accepted sessions we control admission based on rates, while load is low we don't enforce ip checks, with higher load we allow only 1 request from ip in 10s. Higher load is achievable only during intentional dos.

### DIFF
--- a/monad-wireauth/README.md
+++ b/monad-wireauth/README.md
@@ -22,17 +22,20 @@ session management layer:
 
 high-level api with dos protection:
 
-the filter operates using two global handshake rate limits plus a hard session cap:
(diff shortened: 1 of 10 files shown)


dshulyak:
@claude

dshulyak:
@claude

Issue #2999, opened 2026-04-21
fair queue/txpool: enforce minimal and max tx size when enqueing


dshulyak:
transport can theoretically accept 1 byte transactions, that will create a lot of pressure on the queue and peer scoring.
the same for maximal, that will allow to enqueue more in memory than we anticipate.

ideally this should be enforced at the codec level, so that decoding itself doesn't consume too much resources. especially risky with tcp 3MB limit.

also we can limit collection size at the transport level, so that amount of work per message is bounded

PR #3015, opened 2026-04-23
monad-state: limit forwarded txs per message

dshulyak:
collection without limits allows to setup 3M 1 byte transactions over tcp, which takes ~100ms to decode and significant memory pressure about ~200 MiB to decode, and then has some downstream effects as well where we need to loop over them, that potentially can be abused to slow down a node.

this wraps txs with limited vector, with 5_000 length

related https://github.com/category-labs/monad-bft/issues/2999

### DIFF
--- a/monad-types/src/lib.rs
+++ b/monad-types/src/lib.rs
@@ -17,6 +17,7 @@ use std::{
     error::Error,
     fmt::{Debug, Display},
     io,
+    marker::PhantomData,
     ops::{Add, AddAssign, Div, Rem, Sub, SubAssign},
     str::FromStr,
     time::{Duration, Instant},
@@ -28,8 +29,12 @@ use alloy_rlp::{
 };
 use monad_crypto::certificate_signature::PubKey;
 pub use monad_crypto::hasher::Hash;
-use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use serde::{
+    de::{IgnoredAny, SeqAccess, Visitor},
+    Deserialize, Deserializer, Serialize, Serializer,
+};
 use serde_json::Value;
+use serde_with::{de::DeserializeAsWrap, DeserializeAs, SerializeAs};
 use zerocopy::IntoBytes;
 
 pub const GENESIS_SEQ_NUM: SeqNum = SeqNum(0);
@@ -37,6 +42,7 @@ pub const GENESIS_ROUND: Round = Round(0);
 
 pub const ETHERNET_MTU: u16 = 1500;
 pub const DEFAULT_MTU: u16 = ETHERNET_MTU;
+pub const MAX_FORWARDED_TXS_PER_MESSAGE: usize = 5_000;
 
 const PROTOCOL_VERSION: u32 = 1;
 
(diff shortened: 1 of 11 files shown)


monad-types/src/lib.rs
```diff
@@ -924,6 +934,103 @@ impl<T, const N: usize> IntoIterator for LimitedVec<T, N> {
...
+    }
+}
+
+impl Decodable for ForwardedTxList {
+    fn decode(buf: &mut &[u8]) -> alloy_rlp::Result<Self> {
+        LimitedVec::decode(buf).map(Self)
+    }
+}
+
+impl Serialize for ForwardedTxList {
```
andr-dev: Is there a reason we can't use `#[serde_as(as = "Vec<serde_with::hex::Hex>")]` like the old code did? A new implementation concerns me as I don't believe we have stability tests for the serialization of forwarded tx messages.

dshulyak: @andr-dev is it good now? updated to use serde_as and added a test to catch this

monad-types/src/lib.rs
```diff
@@ -962,6 +964,33 @@ impl<'de, T: Deserialize<'de>, const N: usize> Deserialize<'de> for LimitedVec<T
...
+where
+    TAs: DeserializeAs<'de, T>,
+{
+    fn deserialize_as<D: Deserializer<'de>>(deserializer: D) -> Result<LimitedVec<T, N>, D::Error> {
+        let vec = <Vec<TAs> as DeserializeAs<'de, Vec<T>>>::deserialize_as(deserializer)?;
+        if vec.len() > N {
+            return Err(<D::Error as serde::de::Error>::custom(
+                "list exceeds maximum length",
+            ));
+        }
```
eerkaijun: do we want to add this check to `impl Deserialize` too? there's no risk since network messages don't go through that code path but might be worth it to make it consistent and future proof

eerkaijun (dismissed):

monad-types/src/lib.rs
dshulyak: updated, also changed it so that allocation is done after length validation

dshulyak:
```
FAIL: test_eth_createAccessList_mip8_page_dedup (test_rpc_standard.EthClient.test_eth_createAccessList_mip8_page_dedup) (case='same_page')

Test eth_createAccessList page deduplication under MIP-8 page-gas.

----------------------------------------------------------------------

  File "/home/jenkins/workspace/ft_ci_dmitry_limit-forwarded-txs/monad-integration/tests/test_rpc_standard/test_rpc_standard.py", line 1572, in test_eth_createAccessList_mip8_page_dedup

    assert actual == want, f'{name}: access list mismatch\nExpected: {want}\nActual: {actual}'

           ^^^^^^^^^^^^^^

AssertionError: same_page: access list mismatch

Expected: {'0x19860f538cf66f78bb02215d2e651b96437ba405': {'0x0000000000000000000000000000000000000000000000000000000000000000'}}

Actual: {'0x19860f538cf66f78bb02215d2e651b96437ba405': {'0x0000000000000000000000000000000000000000000000000000000000000001', '0x0000000000000000000000000000000000000000000000000000000000000000'}}
```

eerkaijun (approved):

dshulyak:
```
....E

======================================================================

ERROR: test_eth_simulate_v1_smoketest (test_rpc_eth_simulate_v1.EthSimulateV1.test_eth_simulate_v1_smoketest)
web3.exceptions.MethodUnavailable: {'code': -32601, 'message': 'Method not supported'}
```

michael-yxchen (approved):

PR #3022, opened 2026-04-27
monad-node: persist wal events on waltrace thread

dshulyak:
Create a dedicated monad_bft_waltrace thread that owns the WAL logger. The node forwards log-friendly events over a bounded crossbeam channel in arrival order and never blocks the main event loop on WAL writes.

If the waltrace queue fills, warn and drop the WAL event. If WAL initialization or persistence fails.

### DIFF
--- a/monad-node/src/main.rs
+++ b/monad-node/src/main.rs
@@ -20,7 +20,7 @@ use std::{
     num::NonZeroU16,
     path::PathBuf,
     process,
-    sync::Arc,
+    sync::{mpsc::TrySendError, Arc},
     time::{Duration, Instant},
 };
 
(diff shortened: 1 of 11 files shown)


dshulyak:
@claude review

monad-node/src/main.rs
```diff
@@ -466,21 +476,24 @@ async fn run(node_state: NodeState) -> Result<(), ()> {
...
-                            return Err(());
+                    if let Some(waltrace_tx) = waltrace_tx.as_ref() {
+                        let wal_event = LogFriendlyMonadEvent {
+                            timestamp: Utc::now(),
+                            event: event.lossy_clone(),
+                        };
+                        match waltrace_tx.try_send(wal_event) {
+                            Ok(()) => {}
+                            Err(TrySendError::Full(_)) => {
+                                warn!("waltrace is lagging; dropping wal event");
```
omegablitz: Ideally this would be marked in the wal somehow, but if we expect it to be sufficiently rare maybe it doesn't matter

We can always cross-check against the logs if the wal looks messed up to look for the warn worst case

omegablitz (approved):

monad-node/src/main.rs
dshulyak: i expect it to happen only during intentional spam or a bug in the wal writer

marcograss:
approved

PR #3024, opened 2026-04-27
monad-wal: trace only wal-logged events

dshulyak:
Change-Id: I179d4f964292407d19480fffad7445f079c2ca53

### DIFF
--- a/monad-executor-glue/src/lib.rs
+++ b/monad-executor-glue/src/lib.rs
@@ -795,24 +795,28 @@ where
     }
 }
 
-#[derive(Clone, PartialEq, Eq, Serialize)]
+#[derive(Clone, PartialEq, Eq, Serialize, monad_wal::WALLog)]
 pub enum ConsensusEvent<ST, SCT, EPT>
 where
     ST: CertificateSignatureRecoverable,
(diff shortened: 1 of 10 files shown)


dshulyak:
@claude review

omegablitz:
What was the motivation for this? Was it for human readability or because the thread was getting backlogged or something?

monad-executor-glue/src/lib.rs
```diff
@@ -2020,26 +2035,31 @@ where
...
     /// Events to update validator set
+    #[wal(enable(nested))]
     ValidatorEvent(ValidatorEvent<SCT>),
     /// Events to mempool
+    #[wal(enable(nested))]
     MempoolEvent(MempoolEvent<ST, SCT, EPT>),
     /// Events for the debug control panel
     ControlPanelEvent(ControlPanelEvent<ST>),
     /// Events to update the block timestamper
     TimestampUpdateEvent(u128),
```
omegablitz, line 2061: I think this should be enabled

omegablitz (approved):
Maybe consider switching field inclusion to opt out instead of opt in for monad_wal::WALLog

PR #3032, opened 2026-04-28
Remove decompression support from RPC

andr-dev:


eerkaijun:
I think copilot review is correct, Actix still allows compressed rpc requests by default, need to remove that too

andr-dev:
> I think copilot review is correct, Actix still allows compressed rpc requests by default, need to remove that too

Removing the `compress` feature disables compression/decompression entirely. This is evident by the removal of the dependencies on brotli, flate, or zstd. Furthermore, https://docs.rs/actix-web/latest/src/actix_web/types/payload.rs.html#381-389.

That being said, copilot is right that I don't think the server will provide a specific error message about compression, but it will probably reply with a "bad http body" message or similar

eerkaijun:
can we add a test case for compressed rpc request? either unit or flexnet I guess

omegablitz (dismissed):

marcograss:
a couple of nits that are related:
- Compressed clients now get a misleading error. A request with Content-Encoding: gzip returns HTTP 200 + JSON‑RPC parse_error, because the bytes look
  like garbage to the JSON parser. A 415 (Unsupported Media Type) or 400 with an explicit "Content-Encoding not supported" message would be friendlier — a
   small middleware that rejects any non‑identity Content-Encoding would restore that contract without bringing decompression back.
   
- Unknown Content-Encoding values are silently ignored. Previously DecompressionGuard rejected unparseable encodings with 400; now they're not

ogechno:
Just a small nit: the `ws` feature on `actix-web` only exists from 4.12.0 onward, but the version constraint is "4.5.1". Listing `ws` is what forces cargo to resolve to ≥4.12.1 (implemented [here](https://github.com/actix/actix-web/releases/tag/web-v4.12.0)). Worth bumping the constraint to "4.12" to make the real floor explicit

andr-dev:
Kept middleware that rejects all messages with content encoding, added test showing this failure, and updated `actix-web` dep to `4.12.1` in `Cargo.toml` (no change in `Cargo.lock`).

philipglazman (approved):

eerkaijun (approved):

ariqchowdhury (approved):

PR #3037, opened 2026-04-30
Generalize the indexed query engine substrate

JoeHowarth:
Reorganizes the log query internals into a dedicated engine module, then introduces typed primary IDs, family-prefixed table construction, generic indexed clauses, and an IndexedFilter trait. The engine stops owning log-specific field vocabulary and codec details, leaving the logs module to provide domain-specific filtering and materialization.

This is mostly structural groundwork for additional indexed families: file moves, module boundary cleanup, and call-site rewrites preserve behavior while intentionally changing unshipped table naming and LogId encoding.


monad-chain-data/src/engine/family.rs
```diff
@@ -0,0 +1,57 @@
...
+/// canonical list of domains the engine supports.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum Family {
+    Log,
+}
+
+impl Family {
+    pub const fn table_ids(self) -> FamilyTableIds {
+        match self {
+            Family::Log => family_table_ids!("log"),
```
0xSqualo, line 54: @JoeHowarth  should existing `block_log_header` / `block_log_blob` data be considered disposable here, or do we need a migration/backcompat read? 

This changes the log family materialization tables to `log_block_*` while the block records and log directory/bitmap indices keep their old table names, so an upgraded store can still resolve old log locations but then read empty new header/blob tables and return `MissingData`.

PR #3039, opened 2026-04-30
Add transaction family substrate

JoeHowarth:
Introduces Family::Tx, TxId, tx windows on BlockRecord, and a txs module containing transaction ingest, storage, materialization, filtering, and response types. FinalizedBlock now carries optional transaction inputs, while existing log-only fixtures continue to ingest zero-tx windows until transaction ingest is wired in.

The tx storage shape indexes from, to, and selector streams over signed transaction envelopes and keeps RPC conversion feature-gated. TxEntry remains bytes-first, with alloy decoding exposed through accessors and an optional RPC conversion helper.

### DIFF
--- a/monad-chain-data/src/primitives/state.rs
+++ b/monad-chain-data/src/primitives/state.rs
@@ -113,6 +113,51 @@ impl From<LogId> for PrimaryId {
     }
 }
 
+/// `PrimaryId` scoped to the tx family. Kept as a distinct type so tx-domain
+/// signatures don't accept primary ids from other families by mistake.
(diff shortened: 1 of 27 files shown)


monad-chain-data/src/primitives/state.rs
```diff
@@ -131,6 +176,7 @@ pub struct BlockRecord {
     pub block_hash: Hash32,
     pub parent_hash: Hash32,
     pub logs: FamilyWindowRecord,
+    pub txs: FamilyWindowRecord,
```
0xSqualo, line 179: @JoeHowarth This changes the persisted block-record shape by adding a required tx window, while decoding still expects the new shape exactly; should this use a backward-compatible decode/default tx window or a storage migration so existing stores do not stall after upgrade?

PR #3042, opened 2026-04-30
Fence publication head updates with CAS

JoeHowarth:
Removes the unused versioned PutResult and Record wrappers from ordinary MetaStore reads and writes, leaving content-addressed artifact writes as plain idempotent put/get operations. A separate MetaStoreCas trait is introduced specifically for the publication head boundary.

Publication state moves onto MonadChainDataService and uses CAS rows that are distinct from normal kv rows. A stale publication advance now returns FencedOut, with tests covering stale-none, wrong-version, and a two-writer race.

### DIFF
--- a/monad-chain-data/src/api.rs
+++ b/monad-chain-data/src/api.rs
@@ -19,7 +19,11 @@ use crate::{
     blocks::{
         execute_query_blocks, load_blocks_by_numbers, QueryBlocksRequest, QueryBlocksResponse,
     },
-    engine::{family::Family, query::family_runner::IndexedFamilyQuery, tables::Tables},
+    engine::{
+        family::Family,
+        query::family_runner::IndexedFamilyQuery,
+        tables::{PublicationTables, Tables},
+    },
     error::{MonadChainDataError, Result},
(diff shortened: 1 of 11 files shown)


monad-chain-data/src/api.rs
```diff
@@ -147,11 +165,13 @@ impl<M: MetaStore, B: BlobStore> MonadChainDataService<M, B> {
...
-            .store_state(crate::primitives::state::PublicationState {
-                indexed_finalized_head: block.block_number(),
-            })
+        self.publication
+            .cas_advance(
+                expected_version,
+                crate::primitives::state::PublicationState {
+                    indexed_finalized_head: block.block_number(),
+                },
+            )
```
0xSqualo, line 174: @JoeHowarth is it intentional that `publication_state`/`state` moves to a physically separate CAS namespace and that block-number keyed artifacts are written before the CAS publish, even though this can make upgraded nodes ignore an existing KV-published head and lets a fenced stale writer overwrite already-published N+1 rows before `cas_advance` fails?

PR #3045, opened 2026-04-30
fix(rpc): return access list for createAccessList reverts

### LINKED ISSUE #2684: eth_createAccessList discrepancy
Reported by Asymmetric 

(eth_createAccessList\create-al-abi-revert)

The server should return the accessed slots regardless of failure, and should report the failure in the "error" field
https://github.com/category-labs/monad-bft/blob/35860968b2561283f9e345ef7319d2aeb6421cad/monad-rpc/src/handlers/eth/call.rs#L950
ref: https://github.com/ethereum/go-ethereum/blob/5d512083347d7bdd1c9d92416dfd1d20dfc01544/internal/ethapi/api.go#L1170

mmv08:
# Fix `eth_createAccessList` Revert Responses

## Summary

Fixes `eth_createAccessList` so reverted executions return a successful JSON-RPC result containing the generated access list, gas used, and an `error` field.

Also updates the OpenRPC return schema/docs for `eth_createAccessList` so the generated documentation exposes `accessList`, `gasUsed`, and the optional `error` field directly.

This matches the expected Ethereum JSON-RPC behaviour and resolves category-labs/monad-bft#2684.

## Problem

`eth_createAccessList` was returning a top-level JSON-RPC error when the simulated transaction reverted.

That behaviour was not compatible with the geth or the Ethereum execution API test expectation. For access-list creation, an EVM revert is still a useful result: the node should return the addresses and storage slots touched before the revert, and report the execution failure inside the result object's `error` field.

Before this change, callers could not receive the partial access list for reverting calls. This caused the `create-al-abi-revert` compatibility case to fail.

## Solution

This change separates access-list generation failures from transaction execution failures.

The first execution runs with the access-list tracer and decodes the tracer output even when execution reverts. The second execution runs with the generated access list to compute `gasUsed` and the optional result `error`.

Execution failures that are part of the simulated transaction, such as revert, out-of-gas, or reserve-balance violation, are now returned inside the `eth_createAccessList` result object. Internal/setup failures still remain top-level JSON-RPC errors.

The OpenRPC signature now returns a schema-backed `MonadCreateAccessListResult`, so generated RPC docs describe the same result object shape returned at runtime.

The follow-up execution is also pinned to the same resolved block as the tracer execution, so `latest` and `pending` requests cannot accidentally use different blocks between the two passes.

## Notes

This PR depends on category-labs/monad#2241. The RPC-side fix requires the execution layer to return trace output for traced executions that revert.

This keeps Monad's existing single traced pass plus one follow-up execution. Geth can iterate access-list tracing until the access list stabilizes, which can discover additional slots in gas-sensitive flows. That broader behavior is intentionally left out of this compatibility fix.

## Testing

- `cargo check -p monad-rpc`
- `cargo fmt --check`
- `TRIEDB_TARGET=triedb_driver CC=gcc-15 CXX=g++-15 CFLAGS="-march=haswell" CXXFLAGS="-march=haswell" ASMFLAGS="-march=haswell" cargo test -p monad-rpc handlers::eth::call::tests`

### DIFF
--- a/monad-rpc/src/handlers/eth/call.rs
+++ b/monad-rpc/src/handlers/eth/call.rs
@@ -21,10 +21,11 @@ use std::{
 use alloy_consensus::{Header, SignableTransaction, TxEip1559, TxEip7702, TxEnvelope, TxLegacy};
 use alloy_eips::eip7702::SignedAuthorization;
 use alloy_primitives::{Address, Bytes, Signature, TxKind, Uint, B256, U256, U64, U8};
-use alloy_rpc_types::{AccessList, AccessListItem, AccessListResult};
+use alloy_rpc_types::{AccessList, AccessListItem};
 use monad_chain_config::execution_revision::MonadExecutionRevision;
 use monad_ethcall::{
-    eth_call, CallResult, EthCallExecutor, EthCallRequest, MonadTracer, StateOverrideSet,
+    eth_call, CallResult, EthCallExecutor, EthCallRequest, EthCallResult, FailureCallResult,
+    MonadTracer, StateOverrideSet,
 };
 use monad_rpc_docs::rpc;
 use monad_triedb_utils::triedb_env::{
@@ -45,7 +46,7 @@ use crate::{
         parse_ethcall_chain_id,
     },
     types::{
-        eth_json::BlockTagOrHash,
+        eth_json::{BlockTagOrHash, MonadCreateAccessListResult},
         ethhex,
         jsonrpc::{JsonRpcError, JsonRpcResult},
     },
@@ -498,39 +499,90 @@ pub enum CallParams {
 
 impl CallParams {
     /// Destructure into the component parts needed by `prepare_eth_call`.
-    fn into_parts(self) -> (CallRequest, BlockTagOrHash, StateOverrideSet, MonadTracer) {
+    fn into_execution_params(self) -> (EthCallExecutionParams, BlockTagOrHash) {
         match self {
-            CallParams::Call(p) => (
-                p.transaction,
-                p.block,
-                p.state_overrides,
-                MonadTracer::NoopTracer,
-            ),
-            CallParams::Trace(p) => (
-                p.transaction,
-                p.block,
-                p.tracer.state_overrides,
-                p.tracer.tracer_params.into(),
-            ),
-            CallParams::AccessList(p) => (
-                p.transaction,
-                p.block,
-                StateOverrideSet::default(),
-                MonadTracer::AccessListTracer,
-            ),
+            CallParams::Call(p) => {
+                let execution_params = EthCallExecutionParams {
+                    transaction: p.transaction,
+                    state_overrides: p.state_overrides,
+                    tracer: MonadTracer::NoopTracer,
+                };
+                (execution_params, p.block)
+            }
+            CallParams::Trace(p) => {
+                let execution_params = EthCallExecutionParams {
+                    transaction: p.transaction,
+                    state_overrides: p.tracer.state_overrides,
+                    tracer: p.tracer.tracer_params.into(),
+                };
+                (execution_params, p.block)
+            }
+            CallParams::AccessList(p) => {
+                let execution_params = EthCallExecutionParams {
+                    transaction: p.transaction,
+                    state_overrides: StateOverrideSet::default(),
+                    tracer: MonadTracer::AccessListTracer,
+                };
+                (execution_params, p.block)
+            }
         }
     }
 }
 
+struct EthCallExecutionParams {
+    transaction: CallRequest,
+    state_overrides: StateOverrideSet,
+    tracer: MonadTracer,
+}
+
+/// Controls response shaping for provider-cap and execution out-of-gas cases.
+/// Other call failures pass through and are handled by the RPC method.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+enum OutOfGasHandling {
+    RpcError,
+    ReturnAsCallFailure,
+}
+
 #[tracing::instrument(level = "debug")]
-pub async fn prepare_eth_call<T: Triedb + TriedbPath>(
+async fn prepare_eth_call<T: Triedb + TriedbPath>(
     triedb_env: &T,
     eth_call_handler_config: &EthCallHandlerConfig,
     eth_call_executor: &EthCallExecutor,
     chain_id: u64,
     params: CallParams,
+    out_of_gas_handling: OutOfGasHandling,
 ) -> Result<(BlockKey, CallResult), JsonRpcError> {
-    let (mut tx, block_tag, state_overrides, tracer) = params.into_parts();
+    let (execution_params, block_tag) = params.into_execution_params();
+    let block_key = get_block_key_from_tag_or_hash(triedb_env, block_tag)
+        .await
+        .ok_or_else(JsonRpcError::block_not_found)?;
+
+    prepare_eth_call_at_block(
+        triedb_env,
+        eth_call_handler_config,
+        eth_call_executor,
+        chain_id,
+        execution_params,
+        out_of_gas_handling,
+        block_key,
+    )
+    .await
+}
+
+async fn prepare_eth_call_at_block<T: Triedb + TriedbPath>(
+    triedb_env: &T,
+    eth_call_handler_config: &EthCallHandlerConfig,
+    eth_call_executor: &EthCallExecutor,
+    chain_id: u64,
+    params: EthCallExecutionParams,
+    out_of_gas_handling: OutOfGasHandling,
+    block_key: BlockKey,
+) -> Result<(BlockKey, CallResult), JsonRpcError> {
+    let EthCallExecutionParams {
+        transaction: mut tx,
+        state_overrides,
+        tracer,
+    } = params;
 
     tx.input.input = match (tx.input.input.take(), tx.input.data.take()) {
         (Some(input), Some(data)) => {
@@ -542,22 +594,24 @@ pub async fn prepare_eth_call<T: Triedb + TriedbPath>(
         (None, data) | (data, None) => data,
     };
 
-    if tx.gas
-        > Some(U256::from(
-            eth_call_handler_config.provider_gas_limit_eth_call,
-        ))
-    {
-        return Err(JsonRpcError::eth_call_error(
-            "user-specified gas exceeds provider limit".to_string(),
-            None,
-        ));
+    let provider_gas_limit = U256::from(eth_call_handler_config.provider_gas_limit_eth_call);
+    if tx.gas > Some(provider_gas_limit) {
+        match out_of_gas_handling {
+            OutOfGasHandling::RpcError => {
+                return Err(JsonRpcError::eth_call_error(
+                    "user-specified gas exceeds provider limit".to_string(),
+                    None,
+                ));
+            }
+            OutOfGasHandling::ReturnAsCallFailure => {
+                // Geth caps eth_createAccessList gas above RPCGasCap instead
+                // of rejecting it before execution.
+                tx.gas = Some(provider_gas_limit);
+            }
+        }
     }
 
     // TODO: check duplicate address, duplicate storage key, etc.
-
-    let block_key = get_block_key_from_tag_or_hash(triedb_env, block_tag)
-        .await
-        .ok_or_else(JsonRpcError::block_not_found)?;
 
     let mut header = match triedb_env
         .get_block_header(block_key)
@@ -621,18 +675,29 @@ pub async fn prepare_eth_call<T: Triedb + TriedbPath>(
     )
     .await
     {
-        monad_ethcall::CallResult::Failure(error)
-            if matches!(error.error_code, monad_ethcall::EthCallResult::OutOfGas) =>
-        {
-            if eth_call_provider_gas_limit < header_gas_limit
-                && U256::from(eth_call_provider_gas_limit) < original_tx_gas
-            {
-                return Err(JsonRpcError::eth_call_error(
-                    "provider-specified max eth_call gas limit exceeded".to_string(),
+        CallResult::Failure(error) if matches!(error.error_code, EthCallResult::OutOfGas) => {
+            match out_of_gas_handling {
+                OutOfGasHandling::RpcError => Err(JsonRpcError::eth_call_error(
+                    if eth_call_provider_gas_limit < header_gas_limit
+                        && U256::from(eth_call_provider_gas_limit) < original_tx_gas
+                    {
+                        "provider-specified max eth_call gas limit exceeded".to_string()
+                    } else {
+                        "out of gas".to_string()
+                    },
                     None,
-                ));
+                )),
+                OutOfGasHandling::ReturnAsCallFailure => Ok((
+                    block_key,
+                    CallResult::Failure(FailureCallResult {
+                        error_code: error.error_code,
+                        gas_used: error.gas_used,
+                        gas_refund: error.gas_refund,
+                        message: "out of gas".to_string(),
+                        data: None,
+                    }),
+                )),
             }
-            return Err(JsonRpcError::eth_call_error("out of gas".to_string(), None));
         }
         result => Ok((block_key, result)),
     }
@@ -661,6 +726,7 @@ pub async fn monad_eth_call<T: Triedb + TriedbPath>(
         eth_call_executor,
         chain_id,
         CallParams::Call(params),
+        OutOfGasHandling::RpcError,
     )
     .await?;
     match result {
@@ -700,6 +766,7 @@ pub async fn monad_debug_traceCall<T: Triedb + TriedbPath>(
         eth_call_executor,
         chain_id,
         CallParams::Trace(params),
+        OutOfGasHandling::RpcError,
     )
     .await?;
     let raw_payload: Vec<u8> = match call_result {
@@ -739,7 +806,10 @@ pub async fn monad_debug_traceCall<T: Triedb + TriedbPath>(
     }
 }
 
-/// Returns an access list containing all addresses and storage slots accessed during a simulated transaction.
+/// Returns the derived access list and gas used for a simulated transaction.
+///
+/// Execution failures and reverts are reported in the result object's optional
+/// `error` field so callers can still use the generated access list.
 #[rpc(
     method = "eth_createAccessList",
     ignore = "eth_call_handler_config",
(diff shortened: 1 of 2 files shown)


monad-rpc/src/handlers/eth/call.rs
```diff
@@ -522,14 +528,65 @@ impl CallParams {
...
+}
+
+async fn prepare_eth_call_with_resolved_block<T: Triedb + TriedbPath>(
+    triedb_env: &T,
+    eth_call_handler_config: &EthCallHandlerConfig,
+    eth_call_executor: &EthCallExecutor,
+    chain_id: u64,
+    params: CallParams,
+    out_of_gas_handling: OutOfGasHandling,
+    resolved_block_key: Option<BlockKey>,
```
eerkaijun: nit: personally I find this a bit unintuitive that we use Option and do conditional matching later, I think getting the block key at the call sites and just passing `BlockKey` instead of `Option<BlockKey>` is cleaner, but up to you.

eerkaijun (dismissed):
Looks good to me. It would be good to add an integration test case at https://github.com/category-labs/monad-integration/blob/92b37945e8d36eab9ee7d6767d39ec22e2e9f8eb/tests/test_rpc_standard/test_rpc_standard.py#L1250 before merging this

mmv08:
@eerkaijun - good catch, i was going to do this but then i forgot 🙈 thanks!

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -897,11 +1070,327 @@ mod tests {
...
+        serde_cbor::to_vec(access_list).expect("access list encodes as CBOR")
+    }
+
+    fn out_of_gas_failure() -> FailureCallResult {
+        FailureCallResult {
+            error_code: EthCallResult::OutOfGas,
+            gas_used: 12_345,
+            gas_refund: 0,
+            message: "execution reverted".into(),
+            data: Some("0x".into()),
```
philipglazman: data: `None` ?

philipglazman (dismissed):

mmv08:
@eerkaijun, the PR for the tests is live in https://github.com/category-labs/monad-integration/pull/425

eerkaijun (dismissed):

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -345,12 +346,17 @@ pub fn check_contract_creation_size(input: Option<&Bytes>) -> Result<(), JsonRpc
     Ok(())
 }
 
+/// Merge the generated access list with caller-supplied entries for the
+/// follow-up execution used to compute `gasUsed` and the optional result
+/// `error`. The public response still uses the generated access list; ordered
+/// collections keep this internal follow-up transaction deterministic while
+/// preserving the existing address/storage-key deduplication.
```
andr-dev: Let's remove this comment.

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -345,12 +346,17 @@ pub fn check_contract_creation_size(input: Option<&Bytes>) -> Result<(), JsonRpc
...
 pub fn merge_access_lists(generated: AccessList, original: Option<AccessList>) -> AccessList {
     let Some(original) = original else {
         return generated;
     };
 
-    let mut access_map: HashMap<Address, HashSet<B256>> = HashMap::new();
+    let mut access_map: BTreeMap<Address, BTreeSet<B256>> = BTreeMap::new();
```
andr-dev: Why did we change this to BTreeMap/BTreeSet?

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -498,39 +504,108 @@ pub enum CallParams {
...
+        eth_call_executor,
+        chain_id,
+        params,
+        OutOfGasHandling::RpcError,
+    )
+    .await
+}
+
+#[tracing::instrument(level = "debug", name = "prepare_eth_call")]
+async fn prepare_eth_call_with_out_of_gas_handling<T: Triedb + TriedbPath>(
```
andr-dev: Instead of adding a new function, let's change the arguments to `prepare_eth_call` to accept the additional `out_of_gas_handling` param. Then, we can update the callsites which makes it clear where we're using the `RpcError` and `ReturnAsCallFailure` variants.

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -621,23 +698,61 @@ pub async fn prepare_eth_call<T: Triedb + TriedbPath>(
...
+/// but `eth_createAccessList` must return execution failures in the result
+/// object's `error` field after the access list has been generated.
+fn handle_out_of_gas_failure(
+    mut error: monad_ethcall::FailureCallResult,
+    eth_call_provider_gas_limit: u64,
+    header_gas_limit: u64,
+    original_tx_gas: U256,
+    out_of_gas_handling: OutOfGasHandling,
+) -> Result<CallResult, JsonRpcError> {
+    if out_of_gas_handling == OutOfGasHandling::RpcError
```
andr-dev: We can put this under the `match` branch  below right? Also, let's remove the comments both on the function and down below, they're fairly confusing as the function isn't inherently related to `eth_createAccessList`.

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -753,85 +871,136 @@ pub async fn monad_createAccessList<T: Triedb + TriedbPath>(
...
+            );
+            Err(JsonRpcError::internal_error(
+                "unexpected traced revert from NoopTracer eth_createAccessList follow-up call"
+                    .into(),
+            ))
         }
-    };
-
-    serde_json::value::to_raw_value(&result)
-        .map_err(|e| JsonRpcError::internal_error(format!("json serialization error: {}", e)))
```
andr-dev, line 834: I think it might be clearer / easier to review if we split out the `MonadCreateAccessListItem/MonadCreateAccessListResult` change from the revert fix as they are two separate changes.

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -897,11 +1070,327 @@ mod tests {
...
+        let result = access_list_result_from_call_result(access_list.clone(), call_result)
+            .expect("successful follow-up call produces access-list result");
+
+        assert_eq!(result.access_list, access_list);
+        assert_eq!(result.gas_used, U256::from(21_000));
+        assert_eq!(result.error, None);
+    }
+
+    #[test]
+    fn merge_access_lists_keeps_user_supplied_entries_for_follow_up_call() {
```
andr-dev: Why are we adding tests for the `merge_access_lists` method in this change? I'm not against adding more tests, but we seem to be conflating changes here.

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -753,85 +871,136 @@ pub async fn monad_createAccessList<T: Triedb + TriedbPath>(
...
+    match call_result {
+        CallResult::Success(monad_ethcall::SuccessCallResult { output_data, .. }) => {
+            decode_access_list_trace(&output_data)
+        }
+        CallResult::Failure(error) => Err(JsonRpcError::eth_call_error(error.message, error.data)),
+        CallResult::Revert(result) => decode_access_list_trace(&result.trace),
+    }
+}
+
+fn create_access_list_error_message(error: &monad_ethcall::FailureCallResult) -> String {
```
andr-dev: This method doesn't seem to be used anywhere, let's revert separating it into a separate fn and keep it inlined.

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -753,85 +871,136 @@ pub async fn monad_createAccessList<T: Triedb + TriedbPath>(
...
+        .map_err(|e| JsonRpcError::internal_error(format!("failed to decode access list: {}", e)))
+}
+
+/// Extracts access-list tracer output from a simulated execution.
+///
+/// A tracer `Revert` still contains the state trace produced before execution
+/// failed, so it is decoded the same way as a successful trace. A `Failure`
+/// has no trace payload to decode, including out-of-gas failures shaped as call
+/// failures, so it remains a top-level RPC error.
+fn access_list_from_trace_call_result(call_result: CallResult) -> Result<AccessList, JsonRpcError> {
```
andr-dev, line 879: This feels like it should be a method on `MonadCreateAccessListResult` that takes the call result and the original access list and internally performs the merging, etc.

monad-rpc/src/handlers/eth/call.rs
mmv08: I considered splitting this originally, but I think keeping it here is cleaner. The OpenRPC/type change is small and directly documents the response shape introduced by this fix. Splitting it would create a dependent PR where the docs/schema PR is not  very meaningful without the behaviour change, and coordinating review/merge order seems like more overhead than the ~20 lines of schema diff

monad-rpc/src/handlers/eth/call.rs
mmv08: I believe I was testing something out. But in the end, this change is not needed. Thanks for the catch

monad-rpc/src/handlers/eth/call.rs
mmv08: I moved the result construction to MonadCreateAccessListResult::from_follow_up_call_result, but kept the trace decoding/merging outside it.

The reason is that the merge is only used to prepare the internal follow-up call. The actual response should still use the access list produced by the tracer. So putting the merge inside MonadCreateAccessListResult felt a bit misleading to me, since that type is really about shaping the public result rather than preparing the next execution input

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -753,85 +826,135 @@ pub async fn monad_createAccessList<T: Triedb + TriedbPath>(
...
+    serde_cbor::from_slice(raw_payload)
+        .map_err(|e| JsonRpcError::internal_error(format!("failed to decode access list: {}", e)))
+}
+
+/// Extracts access-list tracer output from a simulated execution.
+///
+/// A tracer `Revert` still contains the state trace produced before execution
+/// failed, so it is decoded the same way as a successful trace. A `Failure`
+/// has no trace payload to decode, including out-of-gas failures shaped as call
+/// failures, so it remains a top-level RPC error.
```
andr-dev: This comment just explains what the code does, let's remove it.

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -621,18 +675,32 @@ pub async fn prepare_eth_call<T: Triedb + TriedbPath>(
...
-        {
-            if eth_call_provider_gas_limit < header_gas_limit
-                && U256::from(eth_call_provider_gas_limit) < original_tx_gas
-            {
-                return Err(JsonRpcError::eth_call_error(
-                    "provider-specified max eth_call gas limit exceeded".to_string(),
-                    None,
-                ));
+        CallResult::Failure(error) => {
+            if matches!(error.error_code, EthCallResult::OutOfGas) {
```
andr-dev: Let's keep the `if matches!()` as a match guard.

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -621,18 +675,32 @@ pub async fn prepare_eth_call<T: Triedb + TriedbPath>(
...
+                if out_of_gas_handling == OutOfGasHandling::RpcError
+                    && eth_call_provider_gas_limit < header_gas_limit
+                    && U256::from(eth_call_provider_gas_limit) < original_tx_gas
```
andr-dev: This can be moved into the match branch below.

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -621,18 +675,32 @@ pub async fn prepare_eth_call<T: Triedb + TriedbPath>(
...
+                }
+
+                match out_of_gas_handling {
+                    OutOfGasHandling::RpcError => {
+                        Err(JsonRpcError::eth_call_error("out of gas".to_string(), None))
+                    }
+                    OutOfGasHandling::ReturnAsCallFailure => {
+                        let mut error = error;
+                        error.message = "out of gas".to_string();
+                        error.data = None;
```
andr-dev: Mutating error like this might pose problems if we add other fields that *should* be changed / updated, why not just build a new error object here?

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -753,85 +826,135 @@ pub async fn monad_createAccessList<T: Triedb + TriedbPath>(
...
+    // This intentionally uses one traced pass plus one follow-up execution.
+    // Geth repeats tracing until the access list stabilizes, which can discover
+    // extra slots when warming the first pass lets gas-sensitive code run
+    // further. Keeping this single-pass behavior avoids expanding the scope of
+    // this compatibility fix.
+    //
+    // The public result uses the traced access list. Monad execution warms
+    // tx-initial access-list entries before tracing, so the trace already
+    // includes caller-supplied entries while preserving execution-side
+    // filtering. The merge here is only for the follow-up execution call.
```
andr-dev: What is "this compatibility fix?" This seems like an LLM generated comment, let's remove it.

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -753,85 +826,135 @@ pub async fn monad_createAccessList<T: Triedb + TriedbPath>(
...
+        CallResult::Failure(error) => Err(JsonRpcError::eth_call_error(error.message, error.data)),
+        CallResult::Revert(result) => decode_access_list_trace(&result.trace),
+    }
+}
+
+impl MonadCreateAccessListResult {
+    /// Builds the public `eth_createAccessList` result from the follow-up call.
+    ///
+    /// Unlike `eth_call`, execution failures are reported in the result object's
+    /// `error` field so callers can still use the generated access list.
```
andr-dev: How `eth_call` propogates errors isn't very relevant to how the create access list result does, let's remove this comment.

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -753,85 +826,135 @@ pub async fn monad_createAccessList<T: Triedb + TriedbPath>(
...
                 access_list,
                 gas_used: U256::from(gas_used),
                 error: None,
+            }),
+            CallResult::Failure(error) => match error.error_code {
+                EthCallResult::OutOfGas
+                | EthCallResult::ExecutionError
+                | EthCallResult::ReserveBalanceViolation => {
+                    // `eth_call` reports ABI-decoded revert reasons as a top-level error, but
+                    // geth's `eth_createAccessList` returns the raw VM error string instead.
```
andr-dev: This comment is also irrelevant, let's remove it.

mmv08: Why do you think it is irrelevant? I agreed the wording may not be great, but I still think a small comment helps here because otherwise the "execution reverted: ..." -> "execution reverted" change looks random. I will rephrase it to say that the primary reason for the code to do this is to match Geth’s eth_createAccessList behaviour

monad-rpc/src/handlers/eth/call.rs
mmv08: I added a comment because to me it seemed like an unusual path in this file. Most CallResult::Failure cases become top-level RPC errors, but here, some execution failures intentionally result in a successful eth_createAccessList response with an error set, so we can still return the access list. I think having the reason in the code helps future readers. I will rephrase it to avoid the eth_call comparison.

monad-rpc/src/handlers/eth/call.rs
mmv08: I will rephrase it. The intention was to document a possible edge case/implementation divergence that is handled in Geth but not by us, for future reference. The edge case is roughly the following:

```
contract Store {
      mapping(uint256 => uint256) public values;

      function read(uint256 i) external view returns (uint256) {
          return values[i];
      }
  }

contract Driver {
    Store immutable store;

    constructor(Store s) {
        store = s;
    }

    function run(uint256 t0, uint256 t1) external view returns (uint256 x) {
        x += store.read(0);

         if (gasleft() > t0) {
            x += store.read(1);

            if (gasleft() > t1) {
                x += store.read(2);
            }
        }
    }
}
```

With a tight gas limit and carefully chosen `t0/t1`:

1. First trace has no generated access list yet. It cold-accesses `Store/values [0]`, leaves too little gas for `gasleft() > t0`, and returns an access list containing `Store + values[0]`.
2. Geth runs again with that access list. Now `Store + values[0]` are warm, so execution has a bit more gas at the first branch and reaches `values[1]`. New access list now includes `values[1]`.
3. Geth runs again with `values[1]` warm too. That can push it past `t1`, so it reaches `values[2]`.

I've never seen a similar code in production. However, an LLM tells me such code is written, for example, in liquidation loops that just run until they run out of gas.

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -621,19 +675,33 @@ pub async fn prepare_eth_call<T: Triedb + TriedbPath>(
...
+                        gas_used: error.gas_used,
+                        gas_refund: error.gas_refund,
+                        message: "out of gas".to_string(),
+                        data: None,
+                    }),
+                )),
             }
-            return Err(JsonRpcError::eth_call_error("out of gas".to_string(), None));
         }
+        CallResult::Failure(error) => Ok((block_key, CallResult::Failure(error))),
```
andr-dev: I think we can remove this line, it's identical to the one below right?

eerkaijun (dismissed):

monad-rpc/src/handlers/eth/call.rs
```diff
@@ -621,19 +675,33 @@ pub async fn prepare_eth_call<T: Triedb + TriedbPath>(
...
+                        && U256::from(eth_call_provider_gas_limit) < original_tx_gas
+                    {
+                        Err(JsonRpcError::eth_call_error(
+                            "provider-specified max eth_call gas limit exceeded".to_string(),
+                            None,
+                        ))
+                    } else {
+                        Err(JsonRpcError::eth_call_error("out of gas".to_string(), None))
+                    }
+                }
```
andr-dev: Might be clearer as
```rust
Err(JsonRpcError::eth_call_error(
  if eth_call_provider_gas_limit < header_gas_limit && U256::from(eth_call_provider_gas_limit) < original_tx_gas {
    "provider-specified max eth_call gas limit exceeded".to_string()
  } else {
    "out of gas".to_string()
  },
  None,
))
```

eerkaijun (approved):

andr-dev (approved):

ariqchowdhury (approved):

PR #3046, opened 2026-04-30
MIP-9: Bump validator set max to 300

michael-yxchen:
[MIP-9](https://github.com/monad-crypto/MIPs/blob/main/MIPS/MIP-9.md) raises the validator set cap from 200 to 300, gated by a hard fork on the execution side. Consensus stays passive (reads whatever the staking contract returns); only the static caps need to move.

Consolidate the two redundant constants (MAX_VALIDATORS in monad-consensus-types and MAX_VALIDATOR_SET_SIZE in monad-raptorcast) into a single MAX_VALIDATOR_SET_SIZE in monad-validator. Both crates already depend on monad-validator.

Drop the "additional buffer" framing on the constant

### DIFF
--- a/monad-consensus-types/src/timeout.rs
+++ b/monad-consensus-types/src/timeout.rs
@@ -28,12 +28,12 @@ use monad_validator::{
         SignatureCollection, SignatureCollectionError, SignatureCollectionKeyPairType,
     },
     validator_mapping::ValidatorMapping,
+    validator_set::MAX_VALIDATOR_SET_SIZE,
 };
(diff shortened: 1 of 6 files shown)


eerkaijun (dismissed):

eerkaijun (approved):

omegablitz (approved):
If didn't already, need to sanity check TC path (receiving ~300 different TCs in a round) doesn't choke the network before enabling

xinyuan-dev (approved):

PR #3050, opened 2026-05-04
monad-executor-glue: split peer entry ports

dshulyak:
- 2nd refactoring for bootstrap entry records https://github.com/category-labs/monad-bft/pull/3051
- 3rd relaxes validation for udp port in name record itself https://github.com/category-labs/monad-bft/pull/3052

### DIFF
--- a/monad-executor-glue/src/lib.rs
+++ b/monad-executor-glue/src/lib.rs
@@ -15,7 +15,7 @@
 
 use std::{
     fmt::Debug,
-    net::{SocketAddr, SocketAddrV4},
+    net::{Ipv4Addr, SocketAddr, SocketAddrV4},
     num::NonZeroU16,
 };
 
@@ -276,13 +276,14 @@ pub enum GetMetrics {
     Response(Metrics),
 }
 
-#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
 #[serde(bound = "ST: CertificateSignatureRecoverable")]
 pub struct PeerEntry<ST: CertificateSignatureRecoverable> {
     #[serde(serialize_with = "serialize_pubkey::<_, CertificateSignaturePubKey<ST>>")]
     #[serde(deserialize_with = "deserialize_pubkey::<_, CertificateSignaturePubKey<ST>>")]
     pub pubkey: CertificateSignaturePubKey<ST>,
-    pub addr: SocketAddrV4,
+    #[serde(flatten)]
+    pub address: PeerEntryAddress,
 
     pub signature: ST,
     pub record_seq_num: u64,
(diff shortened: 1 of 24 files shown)


dshulyak:
@claude review

monad-peer-discovery/src/lib.rs
```diff
@@ -424,14 +424,22 @@ impl<ST: CertificateSignatureRecoverable> MonadNameRecord<ST> {
...
     fn try_from(peer: &PeerEntry<ST>) -> Result<Self, Self::Error> {
         let name_record = NameRecord::new_with_ports(
-            *peer.addr.ip(),
-            peer.addr.port(),
-            peer.addr.port(),
+            peer.address,
+            peer.tcp_port.get(),
+            peer.udp_port
+                .ok_or(PeerEntryConversionError::MissingUdpPort)?
+                .get(),
```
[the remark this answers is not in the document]
dshulyak: udp will be optional for name record as well in the latter commit in this stack

monad-executor-glue/src/lib.rs
```diff
@@ -296,33 +297,121 @@ pub struct PeerEntry<ST: CertificateSignatureRecoverable> {
...
+            Self::SocketAddr(endpoint) => NonZeroU16::new(endpoint.address.port())
+                .ok_or("socket address port must be non-zero"),
+        }
+        .expect("peer entry address is validated during config deserialization")
+    }
+
+    pub fn udp_port(&self) -> Option<NonZeroU16> {
+        match self {
+            Self::Split(endpoint) => endpoint.udp_port,
+            Self::SocketAddr(_) => Some(self.tcp_port()),
```
eerkaijun, line 350: just to understand, this is for backwards compatibility?

eerkaijun: I am wondering whether it will be cleaner to first remove unauth udp port in wire messages (e.g. removing PortTag::UDP in NameRecord) and we wouldn't need functions like `udp_port()` anymore

dshulyak: > just to understand, this is for backwards compatibility?

yes

if udp port is removed implicitly together with release then operator can't go back to original behavior without rolling back the release or requiring a fix

eerkaijun: is that not the case right now? because we mandate auth port to be a required field, it will always default to use auth udp and operators can't choose to not use it just by changing config. I might be missing something

dshulyak: there is a fallback. while session is being established (or if it can't be established for unknown reason) traffic will go through non auth.

in future patch when non auth is disabled traffic will be buffered and sent only when session is finally established, and a change to pre-warm validator-to-validator session

basically this change enables less risky rollout, where operator can fallback to original behavior just by updating config

---

i also think that most of this refactoring is useful even without udp port, e.g separating ip and tcp_port

monad-executor-glue/src/lib.rs
```diff
@@ -366,6 +466,21 @@ impl<ST: CertificateSignatureRecoverable> Decodable for PeerEntry<ST> {
...
+
+fn decode_optional_non_zero_u16(payload: &mut &[u8]) -> alloy_rlp::Result<Option<NonZeroU16>> {
+    let port = u16::decode(payload)?;
+    if port == 0 {
+        Ok(None)
+    } else {
+        Ok(Some(
+            NonZeroU16::new(port).expect("non-zero port must construct"),
+        ))
+    }
```
eerkaijun: can simplify with `Ok(NonZeroU16::new(u16::decode(payload)?))`

eerkaijun (dismissed):

eerkaijun (dismissed):

monad-executor-glue/src/lib.rs
```diff
@@ -296,33 +297,121 @@ pub struct PeerEntry<ST: CertificateSignatureRecoverable> {
     pub direct_udp_port: Option<NonZeroU16>,
 }
 
+#[derive(Clone, Copy, Debug, Serialize, PartialEq, Eq)]
+#[serde(untagged)]
+pub enum PeerEntryAddress {
+    Split(PeerEntrySplitEndpoint),
+    SocketAddr(PeerEntrySocketAddrEndpoint),
+}
```
michael-yxchen, line 305: Do we need this enum or can we use `PeerEntrySplitEndpoint` everywhere? It seems that all backward compatibility part can be handled at Deserialize

monad-node/src/main.rs
michael-yxchen, line 1: update error log to include new `MissingUdpPort` variant
```rust
match MonadNameRecord::try_from(&peer_entry) {
    Ok(monad_name_record) => Some((node_id, monad_name_record)),
    Err(_) => {
        warn!(?node_id, "invalid name record signature in config file");
        None
    }
}
```

monad-peer-discovery/src/mock.rs
```diff
@@ -282,7 +288,6 @@ where
     fn get_addr_by_id(&self, id: &NodeId<CertificateSignaturePubKey<ST>>) -> Option<SocketAddrV4> {
         self.known_addresses.get(id).copied()
     }
-
```
michael-yxchen, line 285: nit: deleted by accident?

monad-node/src/main.rs
```diff
@@ -807,7 +811,6 @@ fn resolve_domain_v4<P: PubKey>(node_id: &NodeId<P>, domain: &String) -> Option<
     warn!(?node_id, ?domain, "No IPv4 DNS record");
     None
 }
-
```
michael-yxchen, line 811: nit: deleted by accident?

monad-executor-glue/src/lib.rs
dshulyak: i think PeerEntryAddressSerdeRepr needs to be removed and deserialize derive just added here

dshulyak: updated to use 1 enum, that duplication was not necessary

michael-yxchen (dismissed):

eerkaijun (dismissed):

michael-yxchen (approved):

eerkaijun (approved):

PR #3051, opened 2026-05-04
monad-node-config: separate parts of the address for bootstrap entrypoints

dshulyak:


dshulyak:
@claude review

monad-node-config/src/bootstrap.rs
```diff
@@ -51,3 +51,294 @@ pub struct NodeBootstrapPeerConfig<ST: CertificateSignatureRecoverable> {
     )]
     pub direct_udp_port: Option<NonZeroU16>,
 }
+
+#[derive(Debug, Clone)]
+pub struct NodeBootstrapPeerAddress {
+    // Supported bootstrap address formats:
+    // - Legacy: address = "host:port"
+    // - Split: address = "host", tcp_port = 8000, udp_port = 8001 (optional)
```
eerkaijun: do we want to keep the legacy option? maybe forcing node operators to switch to the new format when they upgrade is better? otherwise I think people will just keep using the default (legacy)

dshulyak: maybe a warn log that legacy will be deprecated in next release? it is somewhat annoying to work with breaking config changes, also during testing

dshulyak: actually they will have to migrate when unauth udp will be deprecated, as the node will reject legacy at that point

eerkaijun: so might as well only support `Split` right now? since they will need to migrate soon and we will need to remove support for legacy soon anyways

but if you feel strongly otherwise we can defer removing legacy support too

dshulyak: with that change the annoying part is that you always have to modify config together with release, and if you want to rollback it you need to modify config again. also i think that MF or CL maintains a list of publicly available bootnodes, so that list have to be modified as well (with rollback considerations)

if both are supported config can be updated only once you know that rollback is not likely, or before the next release.  

beside that there will be a commit in some future release that drops unauthenticated path completely, so it may as well cleanup this part as well

eerkaijun (approved):

monad-node-config/src/bootstrap.rs
```diff
@@ -29,10 +29,10 @@ pub struct NodeBootstrapConfig<ST: CertificateSignatureRecoverable> {
 }
 
 #[derive(Debug, Deserialize, Serialize, Clone)]
-#[serde(deny_unknown_fields)]
```
michael-yxchen, line 32: does dropping `deny_unknown_fields` have any unwanted side effects? I understand that it doesn't work with flatten but want to understand if we should hand roll a flattened helper struct to maintain this strict check

monad-node-config/src/bootstrap.rs
```diff
@@ -51,3 +51,294 @@ pub struct NodeBootstrapPeerConfig<ST: CertificateSignatureRecoverable> {
...
+
+#[derive(Debug, Clone)]
+pub struct NodeBootstrapPeerAddress {
+    // Supported bootstrap address formats:
+    // - Legacy: address = "host:port"
+    // - Split: address = "host", tcp_port = 8000, udp_port = 8001 (optional)
+    // Host is an IPv4 literal or DNS name; IPv6 is not supported here.
+    // Split-format address must not include a port.
+    ip: Option<Ipv4Addr>,
+    domain: Option<String>,
```
michael-yxchen: nit: is an enum more appropriate here? it seems that it's always ip or domain

michael-yxchen (approved):

monad-node-config/src/bootstrap.rs
dshulyak: i agree it is better to keep deny_unknown_fields

PR #3052, opened 2026-05-04
monad-peer-discovery: allow auth-only name records

dshulyak:


dshulyak:
@claude review

monad-peer-discovery/src/discovery.rs
```diff
@@ -1819,19 +1819,28 @@ where
...
             .get(id)
-            .map(|info| info.name_record.udp_address())
+            .and_then(|info| info.name_record.udp_address())
     }
 
     fn get_addr_by_id(&self, id: &NodeId<CertificateSignaturePubKey<ST>>) -> Option<SocketAddrV4> {
         self.routing_info
             .get(id)
-            .map(|name_record| name_record.udp_address())
+            .and_then(|name_record| name_record.udp_address())
```
eerkaijun, line 1831: so for peers that no longer define a non-auth udp port, this function returns None. is that expected? I think raptorcast uses this function to retrieve a peer's address right?

dshulyak: i updated once place where it should be get_tcp_addr instead of get_addr . other than that if get_addr returns None there will be no fallback to unauthenticated path, but thats expected

dshulyak: also pushed update with more explicit names for those functions

michael-yxchen (approved):

eerkaijun (approved):

PR #3064, opened 2026-05-11
Mock-swarm: Transport using raptorcast chunks

xinyuan-dev:
This PR implements a mock-swarm router scheduler that uses raptorcast chunks as underlying transport using real deterministic raptorcast chunk assignment algorithm. It can be helpful to simulate scenarios involving two-hop delay, and chunk-based networking conditions.

Note: the scheduler does not (yet) expose the underlying chunks to the consensus.

Note 2: the scheduler doesn't perform actual raptor-code encoding or use real raptorcast packet format. The underlying transport is a simple `(Arc<Bytes>, chunk_id)` of the full serialized message.

### DIFF
--- a/monad-mock-swarm/src/raptorcast.rs
+++ b/monad-mock-swarm/src/raptorcast.rs
@@ -1,0 +1,575 @@
+// Copyright (C) 2025 Category Labs, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// Raptorcast-aware in-process router scheduler for mock-swarm.
+
+use std::{
+    collections::{BTreeMap, BTreeSet, HashMap, VecDeque},
+    marker::PhantomData,
+    sync::Arc,
+    time::Duration,
+};
+
+use bytes::Bytes;
+use monad_chain_config::{revision::MockChainRevision, MockChainConfig};
+use monad_consensus_types::{
+    block::{MockExecutionProtocol, PassthruBlockPolicy},
+    block_validator::MockValidator,
+};
+use monad_crypto::{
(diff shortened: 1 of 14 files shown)


xinyuan-dev:
> ### Bug: Wrong sender attribution on rebroadcast path

fixed.

> ### Dead config fields: `raptorcast_redundancy` and `segment_len`

removed `raptorcast_redundancy`. renamed `segment_len` to `symbol_len`, which is used in unicast/broadcast (raptorcast v0).

> ### Unbounded `decoding_states` growth

there is no trivial way to know when to prune the decoding state. let's optimize when we encounter any problems.

> ### Visibility widening

marked all touched types `pub`. this issue should be ultimately addressed with a minor architectural change in chunk assignment. tracked with https://github.com/category-labs/monad-bft/issues/3065.

@claude review

xinyuan-dev:
> num_source_symbols mismatch in raptorcast path

switched to use `encoding.layout()` as the source of truth for `num_source_symbols`.

> Rebroadcast clones stale recipient field

this behavior is expected. on a second look, the `ChunkMsg.recipient` field is not even needed: since the chunk assignment is deterministic, the recipient of every chunk is public knowledge. i removed the field.

michael-yxchen (dismissed):
lgtm. need a rebase

xinyuan-dev:
RaptorcastRouterScheduler updated to include an internal proposer schedule state as implemented in https://github.com/category-labs/monad-bft/pull/3114, and each chunk's `round` and `author` fields now get validated. The smoke test has been updated to cover epoch boundaries. Rebased on top of https://github.com/category-labs/monad-bft/pull/3114.

michael-yxchen (dismissed):

monad-mock-swarm/src/raptorcast.rs
```diff
@@ -0,0 +1,591 @@
...
+        msg_id: usize,
+        payload: &Arc<Bytes>,
+        to: NodeId<PT>,
+    ) {
+        let num_source_symbols = payload.len().div_ceil(self.config.symbol_len as usize);
+        let num_scaled_chunks =
+            ((num_source_symbols as f32) * self.config.unicast_redundancy).ceil() as usize;
+        let round = self.current_round.unwrap_or(Round(0));
+
+        let deadline = time + self.config.delay;
```
michael-yxchen: is the delay applied twice, once here and another by `LatencyTransformer` if that's populated?

xinyuan-dev: yes. i just removed the `delay` field here.

michael-yxchen (dismissed):

michael-yxchen (dismissed):

michael-yxchen (approved):

omegablitz (approved):

PR #3069, opened 2026-05-12
Upgrade alloy version to 2.0

eerkaijun:
Block timestamp is an optional field in rpc responses. The version bump introduces this field, and since we have access to block timestamp, we can populate this field

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -419,9 +419,9 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "alloy-consensus"
-version = "1.7.0"
+version = "2.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


eerkaijun:
Dependent on https://github.com/category-labs/monad/pull/2263

Cargo.lock
```diff
@@ -17,7 +17,7 @@ dependencies = [
  "actix-macros",
  "actix-rt",
  "actix_derive",
- "bitflags 2.10.0",
+ "bitflags 2.11.1",
```
andr-dev: I think we might have bumped every dep, let's just do alloy in this change, especially since cargo doesn't have a minimum age feature yet.

dshulyak (dismissed):

xinyuan-dev (dismissed):

andr-dev (dismissed):

philipglazman (dismissed):

michael-yxchen (approved):

xinyuan-dev (approved):

PR #3072, opened 2026-05-12
Split statesync executor into separate crate

andr-dev:
This change completes the statesync split by creating a new `monad-statesync-executor` crate and moving all non-ffi modules. This will enable moving the `monad-statesync` crate to the `category-labs/monad` repo.


philipglazman (dismissed):

philipglazman (dismissed):

omegablitz:
monad-statesync is unfortunately more than a thin wrapper around monad_statesync_client_context, and embeds some higher-level functionality, eg:

```rust
let num_prefixes = self::bindings::monad_statesync_client_prefixes();
for prefix in 0..num_prefixes {
    self::bindings::monad_statesync_client_handle_new_peer(
        ctx,
        prefix as u64,
        self.client_version,
    );
}
```

Would we want to move all of that into the execution repo, or just the smaller minimal rust type wrappers around the existing execution functions?

andr-dev:
> Would we want to move all of that into the execution repo, or just the smaller minimal rust type wrappers around the existing execution functions?

Agreed, we want to keep the execution repo crates as generic as possible, split out all the "higher-level" functionality.

monad-statesync/src/ffi.rs
```diff
@@ -87,40 +101,52 @@ impl SyncCtx {
...
         })
     }
 
-    /// Returns true if reached target and successfully finalized
-    pub fn try_finalize(&mut self) -> bool {
+    pub fn get_or_create_ctx_with_client_prefixes(
+        &mut self,
+    ) -> *mut monad_statesync_client_context {
+        let ctx = self.get_or_create_ctx();
+        add_client_prefixes_as_new_peers(ctx, self.client_version);
```
omegablitz: This looks like it has different behavior than before. Won't this re-add the client prefixes each time?

Worth checking if this passes in flexnet. If it passes, we're missing some test coverage.

omegablitz (commented):
I was thinking to make the wrapper in execution just a very thin wrapper around these:
https://github.com/category-labs/monad/blob/4e08950ff1dc928f9acb77d11bde23a355f5b83b/category/statesync/statesync_client.h#L30-L59

The add_client_prefixes_as_new_peers function for instance is more than a thin wrapper around the execution functions.

But I'm not really sure what other valid invocations of monad_statesync_client_handle_new_peer are, maybe worth checking with execution. Perhaps we could get rid of the monad_statesync_client_handle_new_peer function entirely and embed it into monad_statesync_client_context_create if we always expect to call it this way.

monad-statesync/src/ffi.rs
andr-dev: Yeah you're right, need to copy the initializer from `get_or_create_ctx`.

omegablitz (approved):

omegablitz:
@claude review

philipglazman (approved):

PR #3073, opened 2026-05-13
Raptorcast: Stake partition using integer arithmetic

### LINKED ISSUE #3070: Raptorcast: use integer arithemtic for StakePartition assignment algorithm

xinyuan-dev:
This PR changes the `StakePartition` chunk assignment algorithm to use integer arithmetic (U256) to avoid potential platform-dependent discrepancies. The original f64 version is moved to test as reference implementation for documentation & differential verification.

fixes https://github.com/category-labs/monad-bft/issues/3070.

some tests are written with the assist of opus 4.7.

### DIFF
--- a/monad-raptorcast/src/packet/assigner.rs
+++ b/monad-raptorcast/src/packet/assigner.rs
@@ -15,6 +15,7 @@
 
 use std::{collections::VecDeque, ops::Range};
 
+use alloy_primitives::U256;
 use bytes::BytesMut;
 use monad_crypto::certificate_signature::PubKey;
 use monad_raptor::r10::lt::MAX_TRIPLES;
@@ -302,10 +303,14 @@ impl<PT: PubKey> EvenPartition<PT> {
 // Proportional to stake, plus each validator gets an optional
 // rounding chunk
 pub(crate) struct StakePartition<PT: PubKey> {
-    // Publisher node excluded
-    //
-    // Invariant: validators.map(.1).sum() == 1.0
-    validators: Vec<(NodeId<PT>, f64)>,
+    // Validator set with the publisher node excluded.
+    // Invariant: all stake must be non-zero
+    validators: Vec<(NodeId<PT>, Stake)>,
+
+    // Invariant: total_stake == sum of validators' stakes
+    // Invariant: validators.is_empty() iff total_stake == Stake::ZERO
+    // (i.e. singleton validator set)
+    total_stake: Stake,
 }
 
 impl<PT: PubKey> OrderedNodes<PT> for StakePartition<PT> {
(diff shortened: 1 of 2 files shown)


monad-raptorcast/src/packet/assigner.rs
```diff
@@ -302,10 +303,14 @@ impl<PT: PubKey> EvenPartition<PT> {
 // Proportional to stake, plus each validator gets an optional
 // rounding chunk
 pub(crate) struct StakePartition<PT: PubKey> {
-    // Publisher node excluded
-    //
-    // Invariant: validators.map(.1).sum() == 1.0
-    validators: Vec<(NodeId<PT>, f64)>,
+    // Validator set with the publisher node excluded.
+    // Invariant: all stake must be non-zero
```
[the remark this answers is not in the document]
xinyuan-dev: not a bug. The constructor takes a `PrimaryBroadcastGroup` as input, whose invariant guarantees the validator set is valid: i.e. non-empty and all stakes non-zero.

monad-raptorcast/src/packet/assigner.rs
```diff
@@ -348,32 +353,33 @@ impl<PT: PubKey> StakePartition<PT> {
...
+            total_stake,
         }
-
-        let total_share = validators.iter().map(|(_, s)| *s).sum::<f64>();
-        debug_assert!((total_share - 1.0).abs() < 1e-6);
-
-        Self { validators }
     }
 
     #[cfg(test)]
```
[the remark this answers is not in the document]
xinyuan-dev: from the invariant, `total` is never zero if this function gets called. added a a note and guarded the function using `debug_assert`.

xinyuan-dev:
@claude review

monad-raptorcast/src/packet/assigner.rs
```diff
@@ -297,26 +298,19 @@ impl<PT: PubKey> EvenPartition<PT> {
...
 // Proportional to stake, plus each validator gets an optional
 // rounding chunk
 pub(crate) struct StakePartition<PT: PubKey> {
-    // Publisher node excluded
-    //
-    // Invariant: validators.map(.1).sum() == 1.0
-    validators: Vec<(NodeId<PT>, f64)>,
+    // Validator set with the publisher node excluded.
+    // Invariant: all stake must be non-zero
+    validators: Vec<(NodeId<PT>, Stake)>,
```
dshulyak, line 308: maybe Stake should be a normal fractional, e.g a (U256, U256) tuple. i don't think that additional 32 bytes per item is a concern. then `obligation` (and any other math) can be a method on that type

monad-raptorcast/src/packet/assigner.rs
```diff
@@ -396,67 +397,88 @@ impl<PT: PubKey> StakePartition<PT> {
...
+        let total = self.total_stake.0;
+        debug_assert!(!total.is_zero());
+
+        // SAFETY: obligation getting called implies the presence of
+        // at least one validator in `validators`, thus we must have
+        // total_stake > 0 from the invariant.
+        let (quo, rem) = prod.div_rem(total);
+        let quo = quo.try_into().ok()?;
+        let rem = Stake(rem);
+        Some((quo, rem))
```
dshulyak, line 431: for example here denominator is "lost" . i know that code just checks if remainder is non-zero. but returning tuple would be cleaner

dshulyak (approved):

monad-raptorcast/src/packet/assigner.rs
xinyuan-dev: indeed, but in that case all the denominators would be the same, so i think it would be simpler to have a common `total_stakes` stored outside. the current obligation method is equivalent.

monad-raptorcast/src/packet/assigner.rs
xinyuan-dev: the remainder will be useful for implementing stake-proportional multicast, where the rounding chunks will be rebroadcasted to limited targets.

dshulyak: my comment was related to the comment above about fractional. rem is returned without denominator (e.g total_stake), so if there will be more operations on the Stake it will be easier to misuse it

xinyuan-dev: I see. I do enjoy the idea of having an independent type for stake fraction, but IMO here as `total_stake` is global and unchanging in this context so it's more straightforward to be explicit. I'll explore this idea further later when we need more complex logic (e.g. limited targets).

michael-yxchen (approved):

PR #3077, opened 2026-05-13
Enforce N in all LimitedVec constructors

michael-yxchen:
A node could boot with a TOML config holding more than MAX_VALIDATOR_SET_SIZE validators with no error raised going through ValidatorsConfig::read_from_str

This PR enforce N in LimitedVec's Deserialize (Err), From<Vec<T>>, and FromIterator (assert). Decodable already enforced. The redundant assert in ValidatorSetData::new is dropped.

This code was generated using Claude Opus 4.7.

Closes: #3075

PR #3079, opened 2026-05-13
Remove DerefMut from LimitedVec, add try_push

michael-yxchen:
Prevents callers from bypassing the capacity bound N via `&mut Vec<T>` methods (push/extend/insert/...). Adds `LimitedVec::try_push` as the fallible bounded alternative and `Index/IndexMut` forwarding impls for length-safe element access. Migrates the one production push site in raptorcast_secondary to `try_push`

This code was generated using Claude Opus 4.7.

### DIFF
--- a/monad-types/src/lib.rs
+++ b/monad-types/src/lib.rs
@@ -862,7 +862,35 @@ impl<T, const N: usize> LimitedVec<T, N> {
     pub fn into_inner(self) -> Vec<T> {
         self.0
     }
+
+    /// Append `item` if the vector has not reached the capacity bound `N`.
+    /// Returns the rejected item on overflow so the caller can decide how to recover.
+    pub fn try_push(&mut self, item: T) -> Result<(), LimitedVecCapacityError<T>> {
+        if self.0.len() >= N {
+            return Err(LimitedVecCapacityError {
+                rejected: item,
+                capacity: N,
+            });
+        }
+        self.0.push(item);
+        Ok(())
+    }
 }
+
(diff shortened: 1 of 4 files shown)


monad-types/src/lib.rs
xinyuan-dev, line 853: maybe we can avoid exposing the underlying Vec publicly, instead impl `AsRef<Vec<T>>` (or the current Deref) so it can only be read immutably? <del>the `From<Vec>` can also be turned to a `TryFrom` to prevent leakage. similarly `FromIterator` can be replaced by a fallible version.</del> saw these addressed in https://github.com/category-labs/monad-bft/pull/3077.

michael-yxchen: yeah a fallible version is probably safer than a panic

monad-types/src/lib.rs
```diff
@@ -853,8 +853,36 @@ impl<T, const N: usize> LimitedVec<T, N> {
...
+    /// Returns the rejected item on overflow so the caller can decide how to recover.
+    pub fn try_push(&mut self, item: T) -> Result<(), LimitedVecCapacityError<T>> {
+        if self.0.len() >= N {
+            return Err(LimitedVecCapacityError {
+                rejected: item,
+                capacity: N,
+            });
+        }
+        self.0.push(item);
+        Ok(())
```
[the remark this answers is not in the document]
michael-yxchen: addressed in https://github.com/category-labs/monad-bft/pull/3077/

michael-yxchen:
@claude

omegablitz (approved):

eerkaijun (approved):

PR #3084, opened 2026-05-15
Drop docker workaround for gcc-15

mhumeSF:
## Summary
Remove docker container from Rust CI; gcc-15 is now available in ubuntu-toolchain-r/test for Ubuntu 24.04.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -9,6 +9,9 @@ on:
     branches: "*"
     types: [ opened, synchronize, reopened, edited ]
 
+permissions:
+  contents: read
+
 concurrency:
   group: >
     ${{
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -95,8 +124,18 @@ jobs:
...
           -A clippy::doc-overindented-list-items
 
+      - name: Tune network buffers
+        run: |
+          sudo sysctl -w net.core.rmem_default=62500000
+          sudo sysctl -w net.core.rmem_max=62500000
+          sudo sysctl -w net.core.wmem_default=62500000
+          sudo sysctl -w net.core.wmem_max=62500000
+          sudo sysctl -w net.ipv4.tcp_rmem="4096 62500000 62500000"
+          sudo sysctl -w net.ipv4.tcp_wmem="4096 62500000 62500000"
```
dshulyak: is it possible to apply this config file instead somehow https://github.com/category-labs/monad-bft/blob/master/debian/etc/sysctl.d/90-monad-network-buffer.conf for consistency?

dshulyak (dismissed):

michael-yxchen (dismissed):

.github/workflows/rust.yml
mhumeSF: Yea, that does sound better. I have updated the workflow to apply the file and reload sysctl

https://github.com/category-labs/monad-bft/pull/3084/changes/70e25ab636468798d3c9a45ba725dcfb43e0f038

dshulyak (approved):

michael-yxchen (approved):

PR #3085, opened 2026-05-16
Implement HistoricalDataSource for TriedbDataSource

andr-dev:


andr-dev:
@claude review.

andr-dev:
@claude review

monad-rpc/src/data/source/triedb.rs
```diff
@@ -0,0 +1,153 @@
...
+        BlockPointer::Finalized(block_num) => {
+            BlockKey::Finalized(FinalizedBlockKey(SeqNum(*block_num)))
+        }
+        BlockPointer::NonFinalized(block_num, block_id) => {
+            BlockKey::Proposed(ProposedBlockKey(SeqNum(*block_num), *block_id))
+        }
+    }
+}
+
+fn block_key_to_block_pointer(block_key: BlockKey) -> BlockPointer {
```
philipglazman, line 59: Impl `From`/`Into` ?

philipglazman: Same with `resolve_block_key`

philipglazman (approved):

JoeHowarth (approved):

monad-rpc/src/data/source/triedb.rs
andr-dev: I generally don't like From/Into unless anyone should be able to do the conversion, which I think in this case should only be explicitly done by the Triedb data source.

monad-rpc/src/data/source/triedb.rs
```diff
@@ -0,0 +1,153 @@
...
+            .get_block_header(block_key)
+            .await
+            .map_err(DataSourceError::Internal)?
+        else {
+            return Ok(None);
+        };
+
+        let Ok(transactions) = self.triedb.get_transactions(block_key).await else {
+            // During statesync, block headers might be available in triedb while transactions are
+            // not. Ignoring this error allows the DataSourceStack to continue to the next source.
```
dshulyak, line 135: it looks like this should return typed error and that error is ignored in DataSourceStack

dshulyak (approved):

ariqchowdhury (approved):

monad-rpc/src/data/source/triedb.rs
andr-dev: Yeah that's another way of writing it, I just opted for this since it's simpler for now. Will consider adding an explicit error type in the future.

PR #3086, opened 2026-05-18
raptorcast, node: optional non-auth UDP socket

dshulyak:


dshulyak:
@claude review

eerkaijun (dismissed):

monad-node/src/main.rs
```diff
@@ -573,18 +573,35 @@ where
     OM: Encodable + Clone + Send + Sync + 'static,
     DS: IdentityScore<Identity = NodeId<CertificateSignaturePubKey<ST>>>,
 {
-    let bind_address = SocketAddr::new(
-        IpAddr::V4(node_config.network.bind_address_host),
-        node_config.network.bind_address_port,
-    );
+    let network_config = node_config.network;
+    let self_udp_port = peer_discovery_config.udp_port();
+    let (tcp_bind_address, non_authenticated_bind_address) = if self_udp_port.is_some() {
```
xinyuan-dev: nit: i find this part of logic a little difficult to parse. i think it's clearer by intermediating through an enum like this:

```rust
  enum NonAuthRaptorcastBind {
      Shared { port: u16 }, // TCP + non-auth UDP on same port
      TcpOnly { tcp_port: u16 }, // no non-auth UDP socket
  }
```

then it would be cleaner on how the two cases are determined and their outputs (bind addr, etc).

xinyuan-dev (dismissed):

ariqchowdhury (dismissed):

michael-yxchen (dismissed):

dshulyak:
failure is unrelated. test_create_dip_reverted_legacy seems to be broken on master 

```
FAIL: test_create_dip_reverted_legacy (tests.test_account_balance.AccountBalance.test_create_dip_reverted_legacy)

Test that top level CREATE legacy txn that dips into reserve is being reverted.

----------------------------------------------------------------------

  File "/home/jenkins/workspace/ad-bft_ci_dmitry_unauth-part-4-4/monad-integration/net-lib/monad_flexnet/test_utils/safe_test.py", line 53, in wrapped

    return method(*args, **kwargs)

           ^^^^^^^^^^^^^^^^^^^^^^^

  File "/home/jenkins/workspace/ad-bft_ci_dmitry_unauth-part-4-4/monad-integration/net-lib/monad_flexnet/test_utils/topology_sharing/topology_sharing.py", line 138, in wrapper

    return func(*args, **kwargs)

           ^^^^^^^^^^^^^^^^^^^^^

  File "/home/jenkins/workspace/ad-bft_ci_dmitry_unauth-part-4-4/monad-integration/tests/test_account_balance.py", line 646, in test_create_dip_reverted_legacy

    assert client.eth.get_balance(receipt.contractAddress) == 0

           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

eerkaijun (dismissed):

monad-raptorcast/src/auth/socket.rs
```diff
@@ -104,16 +109,123 @@ where
...
+            .has_initiator_session_by_socket_and_public_key(&auth_addr, public_key);
+        let non_authenticated_addr = self
+            .non_authenticated
+            .is_some()
+            .then(|| name_record.udp_socket().map(SocketAddr::V4))
+            .flatten();
+
+        // A successful connect creates an initiator session synchronously. Later sends will either
+        // buffer while the handshake is pending or keep using non-auth fallback without reconnecting.
+        match (
```
xinyuan-dev: on styling the logic here, i think this full match can be simplified into two or three smaller units with cleaner semantics:

1. needs connect: `!has_authenticated_socket && !has_initiator_session`
2. delivery method: `non_authenticated_addr.is_some_and(!has_auth_socket) => fallback; otherwise => authed`. for fallback path call `write_unicast_with_priority`, for authed path call `write_with_buffering`.

monad-raptorcast/src/auth/socket.rs
```diff
@@ -104,16 +109,123 @@ where
...
+        &mut self,
+        public_key: &AP::PublicKey,
+        name_record: &NameRecord,
+        msg: Bytes,
+        stride: u16,
+        priority: UdpPriority,
+    ) {
+        let auth_addr = SocketAddr::V4(name_record.authenticated_udp_socket());
+        let has_authenticated_socket =
+            self.is_connected_socket_and_public_key(&auth_addr, public_key);
```
xinyuan-dev, line 147: just a small concern: `NameRecord` is borrowed at this point from locking the peer disc mutex. should be fine if all this function does is to send the packet through channel, but if there is in-place i/o or crypto computation then it may be safer to pass by copy to avoid contention. though in practice this likely doesn't matter at all because peer disc and raptorcast runs on single thread so there is no such risk.

xinyuan-dev (dismissed):

monad-raptorcast/src/auth/socket.rs
dshulyak: it is probably ok to copy, but i will also get rid of it after next release. api will deal with addresses rather than name records, as there will be only auth path

monad-raptorcast/src/auth/socket.rs
dshulyak: @xinyuan-dev cleaned up this part

xinyuan-dev (approved):

eerkaijun (approved):

PR #3087, opened 2026-05-18
raptorcast: buffer auth-only UDP sends

dshulyak:


dshulyak:
@claude review

marcograss:
it looks like  we have a regression of the invariant `test_send_with_record_uses_name_record_address` was added to defend (49e87e83b).

The old `lookup_addr` only reused an existing session if its addr matched the supplied NameRecord (`is_connected_socket_and_public_key(name_record.auth_addr, pubkey)`). The new `write_to_name_record` only checks `get_socket_by_public_key(pubkey).is_some()`, so any existing session for that pubkey wins regardless of address — `SendWithRecord` ends up sending to the old session's addr and the supplied NameRecord is ignored.

Repro:
```
cargo test --release --test wireauth_raptorcast \
    test_send_with_record_uses_name_record_address -p monad-raptorcast
```
Passes on the parent (`af76fc235`), fails on this PR's head with `bob1 should NOT have received the ping — it should go to bob2's address`.

Mostly a peer-mobility / Ping verification issue rather than a confidentiality break (session keys are still pubkey-bound), but it does break the Ping-at-new-addr flow until the old session times out.

monad-raptorcast/src/lib.rs
```diff
@@ -1765,80 +1751,25 @@ fn rebroadcast_packet<ST, PD, AP>(
...
-            .lock()
-            .ok()
-            .and_then(|pd| pd.get_udp_addr(target))
+    let Ok(peer_discovery) = peer_discovery_driver.lock() else {
+        return;
     };
 
-    let Some(target_addr) = target_addr else {
-        warn!(target=?target, "failed to find address for rebroadcast target");
+    let Some(name_record) = peer_discovery
```
xinyuan-dev, line 1809: originally the only purpose of the using peer discovery for addr lookup here is to get the socket address. if we're migrating off to wireauth to do the task, i wonder if we can remove the lookup here and delegate to dual_socket for whatever lookup is necessary - and pass in only the `UdpMessage` without the recipient addr resolved. I see this as a prospective improvement of decoupling raptorcast from the data-layer, not related to this pr specifically.

xinyuan-dev (approved):

monad-raptorcast/src/lib.rs
dshulyak: i could store one more reference to peer discovery in that dual socket, but then i need to have 2 methods. write_to_name_record (because one api call provides explicit name record for ping), and also write_to_id (which will select name record internally)

i think doing name record selection before write and keep single write method is cleaner

monad-raptorcast/src/auth/socket.rs
```diff
@@ -121,6 +126,102 @@ where
...
+            has_authenticated_socket,
+            non_authenticated_addr,
+            has_initiator_session,
+        ) {
+            // There is either a fully authenticated socket or an initiated connection.
+            (true, _, _) | (false, None, true) => {
+                let msg_len = msg.len() as u64;
+                if self
+                    .authenticated
+                    .write_with_buffering(public_key, msg, priority)
```
eerkaijun, line 165: just wanted to double check here: if a peer changes its name record and fall into this arm, `write_with_buffering` seems to still be using the old name record because it reads the address from the established session? is that expected?

dshulyak: it should fall into this arm only if one the check below is true, and both are using socket addr.

```
        let has_authenticated_socket =
            self.is_connected_socket_and_public_key(&auth_addr, public_key);
        let has_initiator_session = self
            .authenticated
            .auth_protocol
            .has_initiator_session_by_socket_and_public_key(&auth_addr, public_key);
```

i think the case you described can't happen, even if we still have 2 sessions from several versions of name records, the latter one will be prioritized. i will cleanup though to use both addr and public key consistently

eerkaijun (approved):

monad-raptorcast/src/auth/socket.rs
dshulyak: after another look i think it doesn't need to be refactored, this code will be dropped once i remove unauth path completely. practically it is hard to have two sessions from distinct sockets as they timeout if unreachable. and the latter will used for writing in the unlikely case that 2 or more are live

PR #3091, opened 2026-05-19
Emit UpdateCurrentRound on boot

### LINKED ISSUE #3068: Raptorcast: call UpdateCurrentRound on initialization
Currently, the initial epoch of raptorcast is specified as constructor argument. If we switch to calling `UpdateCurrentRound` instead of pass in `current_{epoch,round}` as argument, this can allow the downstream components (e.g. round info cache) to initialize properly without duplicate codes.

xinyuan-dev:
- [x] verify that the rounds emitted by any `UpdateCurrentRound` command remain strictly monotonic
- [x] verify that all handlers of `UpdateCurrentRound` can support the change:
  + [x] `RaptorCast::exec`
  + [x] `RaptorCastSecondary::exec`
  + [x] `MultiRouter::exec`
  + [x] `PeerDiscoveryDriver::update`

implements https://github.com/category-labs/monad-bft/issues/3068.

### DIFF
--- a/monad-router-multi/src/lib.rs
+++ b/monad-router-multi/src/lib.rs
@@ -163,7 +163,6 @@ where
             recv_group_messages,
             send_group_infos,
             send_outbound_to_primary,
-            current_epoch,
         );
 
         let mut rc_primary = RaptorCast::new(
@@ -175,7 +174,6 @@ where
             non_authenticated_socket,
             control,
             shared_pdd.clone(),
-            current_epoch,
         );
         rc_primary.bind_channel_to_secondary_raptorcast(
             secondary_mode,
@@ -222,7 +220,6 @@ where
             recv_group_messages,
             send_group_infos,
             send_outbound_to_primary,
-            current_epoch,
         );
         self.rc_secondary = rc_secondary;
     }
@@ -236,7 +233,6 @@ where
         channel_to_primary_outbound: UnboundedSender<
             SecondaryOutboundMessage<CertificateSignaturePubKey<ST>>,
         >,
-        current_epoch: Epoch,
     ) -> Option<RaptorCastSecondary<ST, M, OM, SE, PD>> {
         let secondary_instance: RaptorCastConfigSecondary<CertificateSignaturePubKey<ST>> =
             match mode {
@@ -303,7 +299,6 @@ where
                 recv_group_messages,
                 send_group_infos,
                 channel_to_primary_outbound,
-                current_epoch,
             )),
         }
     }
--- a/monad-state/src/lib.rs
+++ b/monad-state/src/lib.rs
@@ -1574,12 +1574,20 @@ where
             high_certificate.clone(),
         );
         let current_round = consensus.get_current_round();
+        let current_epoch = consensus.get_current_epoch();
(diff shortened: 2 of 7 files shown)


dshulyak (dismissed):

monad-router-multi/src/lib.rs
```diff
@@ -175,7 +175,6 @@ where
             non_authenticated_socket,
             control,
             shared_pdd.clone(),
-            current_epoch,
```
michael-yxchen, line 178: does secondary raptorcast still need epoch in new()?

monad-state/src/lib.rs
```diff
@@ -1574,12 +1574,20 @@ where
...
             "done syncing, initializing consensus"
         );
         self.consensus = ConsensusMode::Live(consensus);
+        // Pacemaker emits EnterRound only on a strictly higher
+        // certificate, seed RaptorCast/PeerDiscovery/etc with the
+        // bootstrap round here.
+        commands.push(Command::RouterCommand(RouterCommand::UpdateCurrentRound(
+            current_epoch,
+            current_round,
+        )));
```
michael-yxchen, line 1590: is it too late to update round after consensus finish syncing? the other option is to do that in `MonadStateBuilder::build()`, but this is better if Raptorcast runs fine with epoch == 0

also just noting that `AddEpochValidatorSet` should precede `UpdateCurrentRound`. Otherwise [dataplane.update_trusted](https://github.com/category-labs/monad-bft/blob/939aef624cf8330419b3b3be55722f36868ca483/monad-raptorcast/src/lib.rs#L945) is a noop. Maybe that's already the case on startup

monad-router-multi/src/lib.rs
xinyuan-dev: it appears that secondary rc only uses the epoch to publish point to point protocol messages, the resulting 0 in epoch fiel only influences which decoding cache bucket the message is sent to. so it should be safe to drop them too.

monad-state/src/lib.rs
xinyuan-dev: raptorcast would still run fine with epoch == 0 mostly because neither blocksync/statesync depends on the paths that use epoch inside raptorcast. although this will result in publishing point-to-point message with a 0 in epoch field, which will be received and decoded fine.

if we move the command to `MonadStateBuilder::build`, the downstream component's current_round will be stuck at the forkpoint round, as `UpdateCurrentRound` command will not fire again at going live. So we can opt for emitting the command again on going live, but that could result in duplicate command containing the same round, which breaks the assumed invariant.

ideally we may want to run raptorcast in a degraded state until consensus is live, like skipping publishing/receiving (most) consensus messages until the catches up. but at we do need to continue sending timeout and listening for proposals/tcs in order to learn about high certificates.

monad-router-multi/src/lib.rs
xinyuan-dev: removed `curr_epoch` field from secondary raptorcast.

michael-yxchen (approved):

dshulyak (approved):

PR #3097, opened 2026-05-20
run ci on all pull requests

dshulyak:
turns out `branches: "*"` doesn't match branch names with `/` , so if base branch was as `dmitry/somethig-something` ci won't trigger. so it should either use ** or remove all matchers (which i did)

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -2,7 +2,6 @@ name: Benchmark Regression Check
 
 on:
   pull_request:
-    branches: [ master ]
 
 jobs:
   benchmark:
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -6,7 +6,6 @@ on:
   push:
     branches: ["master"]
   pull_request:
-    branches: "*"
     types: [ opened, synchronize, reopened, edited ]
 
 concurrency:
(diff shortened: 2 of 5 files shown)


dshulyak:
> benchmarks.yml — hardcoded master checkout will break on stacked PRs.
The benchmark workflow checks out ref: master as the baseline (line 24). When this runs on a stacked PR targeting a feature branch, comparing against master is incorrect — the base should be the PR's base branch. This could produce misleading regression/improvement reports on stacked PRs. Consider using ${{ github.event.pull_request.base.ref }} instead of the hardcoded master.

i think this is not an issue, we can still comparer against master

andr-dev (approved):

andr-dev:
I don't think this will retrigger CI after a rebase, but that's fine for now. Ideally it reruns whenever base branch changes.

dshulyak:
what kind of rebase do you have in mind? if there is a chain: master <- c1 <- c2 <- c3 . and c1 was updated then developer needs to rebase/fix whole tree and repush each pr, so ci will be triggered on new push. i think it is the same story if c1 was merged, c2 needs to be rebased and ci will run on push

andr-dev:
> what kind of rebase do you have in mind? if there is a chain: master <- c1 <- c2 <- c3 . and c1 was updated then developer needs to rebase/fix whole tree and repush each pr, so ci will be triggered on new push. i think it is the same story if c1 was merged, c2 needs to be rebased and ci will run on push

Right I forgot github doesn't dedupe c1, yeah after rebase and push it'll rerun.

michael-yxchen (approved):

PR #3102, opened 2026-05-21
Separate RPC events buffer into write and view

andr-dev:
Separated block buffer in RPC into a cloneable read object and single producer write object. This more accurately reflects how events are ingested / read. This also allows for some synchronization primatives to be removed regarding the block number buffer (used for eviction).

This change will enable cleanly implementing `HistoricalDataSource` for the view object only.

### DIFF
--- a/monad-rpc/src/data/buffer/mod.rs
+++ b/monad-rpc/src/data/buffer/mod.rs
@@ -21,28 +21,34 @@ use std::{
     },
 };
 
-use alloy_consensus::TxEnvelope;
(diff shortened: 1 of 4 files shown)


andr-dev:
@claude thorough review

andr-dev:
Duplicate ring buffer entry only occurs during reorg / forking, not worth adding additional complexity to optimize the early eviction of a cache entry during a rare edge case.

philipglazman (approved):

JoeHowarth (approved):

PR #3103, opened 2026-05-22
monad-peer-discovery: split sign name record address

dshulyak:


monad-raptorcast/src/lib.rs
```diff
@@ -153,6 +162,130 @@ where
...
+    self_id: &NodeId<CertificateSignaturePubKey<ST>>,
+    validators: &ValidatorSet<CertificateSignaturePubKey<ST>>,
+    peer_discovery_driver: &Arc<Mutex<PeerDiscoveryDriver<PD>>>,
+) -> Vec<(NodeId<CertificateSignaturePubKey<ST>>, SocketAddr)>
+where
+    ST: CertificateSignatureRecoverable,
+    PD: PeerDiscoveryAlgo<SignatureType = ST>,
+{
+    let num_validators = validators.len();
+    if num_validators == 0 {
```
xinyuan-dev: this should be redundant by validator set invariant. an assert/debug_assert should be sufficient.

xinyuan-dev (approved):

monad-raptorcast/src/lib.rs
dshulyak: oh that commit wasn't supposed to be in this branch, i changed base and didn't rebase properly

eerkaijun (approved):

PR #3109, opened 2026-05-26
Implement HistoricalDataSource for BlockBufferView

andr-dev:
This change implements `HistoricalDataSource` for `BlockBufferView`, completing the vertical integration of unified RPC with `get_block` and `get_block_headers` methods completely migrated.

### DIFF
--- a/monad-rpc/src/data/mod.rs
+++ b/monad-rpc/src/data/mod.rs
@@ -155,37 +155,6 @@ fn resolve_block_height_from_buffer(view: &BlockBufferView, block: &BlockTagOrHa
     }
 }
 
-async fn resolve_block_tag_or_hash(
-    data_source: &impl HistoricalDataSource,
-    block_tag_or_hash: BlockTagOrHash,
-) -> DataSourceResult<Option<BlockPointer>> {
-    match block_tag_or_hash {
-        BlockTagOrHash::BlockTags(BlockTags::Number(Quantity(block_num))) => {
-            data_source.try_resolve_block_number(block_num).await
-        }
-        BlockTagOrHash::BlockTags(BlockTags::Latest) => {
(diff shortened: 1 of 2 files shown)


monad-rpc/src/data/buffer/view.rs
```diff
@@ -209,4 +209,121 @@ impl BlockBufferView {
...
+        let BlockTransactions::Full(txs) = &block.value().transactions else {
+            warn!(
+                block_number = block.value().header.number,
+                "BlockBufferView block does not contain full transaction list"
+            );
+            return Ok(None);
+        };
+
+        Ok(Some((
+            block.value().header.inner.clone(),
```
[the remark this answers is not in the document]
andr-dev: Small regression, this will be changed after we complete the unified RPC integration across all methods and we introduce the pre-serialized struct variants to make methods even faster.

monad-rpc/src/data/buffer/view.rs
```diff
@@ -209,4 +210,142 @@ impl BlockBufferView {
             None
         }
     }
+
+    fn get_block_ref(&self, pointer: BlockPointer) -> Option<DashMapRef<'_, u64, Block>> {
```
marcograss: there seems to be a narrow window where the triedb might have block X but the buffer may hold a fork proposal Y, not Finalized(X), 

Is it possible to carry the hash also on Finalized() ? cc @andr-dev

andr-dev: Yeah the original version had an `Option<Hash>` for finalized and generally speaking it's probably better to always do lookup by hash if it is cheaper. Still thinking through how to cleanly resolve this and consistency when doing sequential operations.

PR #3114, opened 2026-05-29
Raptorcast: Proposer validation per round (sync version)

xinyuan-dev:
This PR implements the same thing in https://github.com/category-labs/monad-bft/pull/3083. Instead of requesting an internal service hosted within MonadState, raptorcast now owns a opaque type (`impl ProposerSchedule`) that exposes the needed information. `ProposerSchedule` trait and its concrete implementation stays inside `monad-validator` alongside EpochManager and LeaderElection, so raptorcast can stay ignorant of the logic for locating the proposer of a round.

```rust
pub trait ProposerSchedule<PT: PubKey> {
    // Returns None when round falls outside the known epochs
    fn check_proposer(&self, node: &NodeId<PT>, round: Round) -> Option<bool>;
    fn check_epoch(&self, epoch: Epoch, round: Round) -> Option<bool>;

    fn insert_epoch(&mut self, epoch: Epoch, epoch_start: Round, val_set: ValidatorSet<PT>);
    fn prune_below(&mut self, cutoff: Round);
}
```

### DIFF
--- a/monad-raptorcast/src/udp.rs
+++ b/monad-raptorcast/src/udp.rs
@@ -22,7 +22,10 @@ use monad_crypto::{
 };
 use monad_executor::ExecutorMetricsChain;
 use monad_types::{Epoch, NodeId, Round};
-use monad_validator::validator_set::{ValidatorSet, ValidatorSetType as _};
+use monad_validator::{
+    proposer_schedule::ProposerSchedule,
+    validator_set::{ValidatorSet, ValidatorSetType as _},
+};
 
 pub use crate::packet::build_messages;
 use crate::{
@@ -171,6 +174,7 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
         &mut self,
         epoch_validators: &BTreeMap<Epoch, ValidatorSet<CertificateSignaturePubKey<ST>>>,
         full_node_group_map: &FullNodeGroupMap<CertificateSignaturePubKey<ST>>,
+        proposer_schedule: &dyn ProposerSchedule<CertificateSignaturePubKey<ST>>,
         chunk: &ValidatedChunk<CertificateSignaturePubKey<ST>>,
         rebroadcast_to: &mut impl FnMut(Vec<NodeId<CertificateSignaturePubKey<ST>>>),
         sender: Option<&NodeId<CertificateSignaturePubKey<ST>>>,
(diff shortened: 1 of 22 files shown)


monad-raptorcast/src/udp.rs
```diff
@@ -217,11 +228,44 @@ impl<ST: CertificateSignatureRecoverable> UdpState<ST> {
     fn handle_deterministic_primary(
         &mut self,
         group: &PrimaryBroadcastGroup<'_, CertificateSignaturePubKey<ST>>,
+        proposer_schedule: &dyn ProposerSchedule<CertificateSignaturePubKey<ST>>,
         round: Round,
         chunk: &ValidatedChunk<CertificateSignaturePubKey<ST>>,
         decoding_context: &DecodingContext<CertificateSignaturePubKey<ST>>,
         rebroadcast_to: &mut impl FnMut(Vec<NodeId<CertificateSignaturePubKey<ST>>>),
     ) -> Option<(NodeId<CertificateSignaturePubKey<ST>>, Bytes)> {
+        if proposer_schedule.check_epoch(group.epoch(), round) != Some(true) {
```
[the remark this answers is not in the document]
xinyuan-dev: when the current round is at stable horizon we'd expect the validator set for the next epoch to be ready, extending the stable horizon.

eerkaijun:
nice, I certainly think this is a simpler approach than https://github.com/category-labs/monad-bft/pull/3083. maybe @michael-yxchen will have some thoughts too

xinyuan-dev:
@claude review

eerkaijun (dismissed):

monad-raptorcast/src/udp.rs
```diff
@@ -1986,6 +2038,93 @@ mod tests_deterministic {
...
+                &proposer_schedule,
+                |_, _, _| {},
+                recv_msg,
+            ));
+        }
+
+        assert!(
+            decoded.is_empty(),
+            "chunks authored by a non-proposer must not decode"
+        );
```
michael-yxchen: this fails at
```rust
if proposer_schedule.check_epoch(group.epoch(), round) != Some(true)
```
because stub always return `Some(false)`, not the intended non proposer branch

xinyuan-dev: fixed! also added a case for asserting decoding failure when `check_epoch` returns `false`.

michael-yxchen (dismissed):

eerkaijun (dismissed):

michael-yxchen (approved):

eerkaijun (approved):

marcograss:
approved

PR #3116, opened 2026-05-31
Jhow/chain data/feature 3

JoeHowarth:


monad-chain-data/src/api.rs
```diff
@@ -1063,30 +1665,126 @@ impl<M: MetaStoreCas, B: BlobStore> MonadChainDataService<M, B> {
...
+            )
+        };
+
+        let epoch_blocks = self.tables.dicts().config().epoch_blocks;
+        let mut last_verified = first_number.saturating_sub(1);
+        for block in blocks {
+            let epoch = block.block_number() / epoch_blocks;
+            let version = u32::try_from(epoch)
+                .map_err(|_| MonadChainDataError::Decode("dict epoch/version overflow"))?;
+            self.ensure_epoch_dicts(version).await?;
```
0xSqualo, line 1732: @JoeHowarth Should standby checksum verification avoid the write-capable `ensure_epoch_dicts` path, or require write authority, so a reader-only verifier cannot persist missing dictionaries while checking artifacts?

PR #3120, opened 2026-06-03
Update vote pace and block reward activation

bamrith:
Changes include:
- New chain revision v0.12.0 with 300ms vote pace
- Functionality to fork block reward by round


omegablitz (approved):
Double checked that we were using block_header.round (rather than proposal.round) everywhere construction/validation of the block reward happens, and all lgtm.

michael-yxchen (approved):

anthdoesntcode (approved):

PR #3121, opened 2026-06-04
Jhow/chain data/feature

JoeHowarth:


monad-archive/src/bin/monad-archiver/chain_data_ingest_worker.rs
```diff
@@ -0,0 +1,189 @@
...
+    async fn fetch_finalized_block(&self, block_number: u64) -> Result<FinalizedBlock> {
+        let (block, receipts, traces) = tokio::try_join!(
+            self.reader.get_block_by_number(block_number),
+            self.reader.get_block_receipts(block_number),
+            async {
+                self.reader
+                    .try_get_block_traces(block_number)
+                    .await
+                    .map(|traces| traces.unwrap_or_default())
+            },
```
0xSqualo: @JoeHowarth Should chain-data ingest fail closed when `try_get_block_traces` returns `None` for a block with transactions? Otherwise archives populated with `--skip-traces` can publish finalized blocks with empty trace indexes, causing `eth_queryTraces` false negatives

PR #3123, opened 2026-06-05
Use StateBackend through &mut

andr-dev:
Previously, the `StateBackend` trait only had methods on `&self` leading us to pass it around by immutable reference. In a future change of the triedb API from the `monad-triedb` crate, we will change methods that internally mutate the triedb reference to use `&mut self`. As such, this change has been split out into a separate PR for easier reviewing.

This will also allow us to remove the unnecessary synchronization primitives on `StateBackendCache` and completely remove `InMemoryStateInner` (https://github.com/category-labs/monad-bft/pull/3124).

### DIFF
--- a/monad-blocktree/src/blocktree.rs
+++ b/monad-blocktree/src/blocktree.rs
@@ -239,7 +239,7 @@ where
         metrics: &mut Metrics,
         block_id: BlockId,
         block_policy: &mut BPT,
-        state_backend: &SBT,
+        state_backend: &mut SBT,
         chain_config: &CCT,
     ) -> Vec<BPT::ValidatedBlock> {
         let Some(path_from_root) = self.get_blocks_on_path_from_root(&block_id) else {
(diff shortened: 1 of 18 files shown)


xinyuan-dev (approved):

omegablitz:
What's the motivation of this? I understand that triedb mutates internally, but from the caller pov it behaves like a cache right?

omegablitz:
After this, everything that accesses state from triedb would need a &mut StateBackend? Readability-wise that seems worse, so just want to understand the motivation

andr-dev:
> After this, everything that accesses state from triedb would need a `&mut StateBackend`? Readability-wise that seems worse, so just want to understand the motivation

1. From caller POV, caches mutate themselves
2. Callers were already only using a single immutable reference to `StateBackend`, no logical code changes going from `&self` to `&mut self`, code readability is identical
3. The current triedb interface is incorrect in that many methods mutate internally but expose the functionality behind `&self` functions. This allows for UB in specific circumstances which fortunately we didn't do (specifically because we only hold a single immutable ref) but this will soon be updated to require `&mut self` on those specific methods.

omegablitz:
In what scenarios can there be UB?
> This allows for UB in specific circumstances which fortunately we didn't do (specifically because we only hold a single immutable ref) but this will soon be updated to require &mut self on those specific methods.

Is the claim that if we hold two immutable references to the same StateBackend, that there's UB? What sequence of function calls can cause that?

I don't think the readability is identical. We set mutability intentionally to better track what needs to be seeded when/where for statesync in [monad-state](https://github.com/category-labs/monad-bft/blob/0743c2c986ef2c8153a61b681bcef8353c682b24/monad-state/src/consensus.rs#L83-L116).

omegablitz (approved):

PR #3124, opened 2026-06-05
Remove synchronization primitives from ExecutionStateReadCache

andr-dev:
Now that `StateBackend` is passed around as `&mut self`, we can remove the silly synchronization primitives on `ExecutionStateReadCache`.

### DIFF
--- a/monad-execution-state-read-cache/src/lib.rs
+++ b/monad-execution-state-read-cache/src/lib.rs
@@ -16,7 +16,6 @@
 use std::{
     collections::{BTreeMap, HashMap},
     marker::PhantomData,
-    sync::{Arc, Mutex},
     time::Duration,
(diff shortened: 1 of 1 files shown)


xinyuan-dev (dismissed):

monad-execution-state-read-cache/src/lib.rs
```diff
@@ -46,7 +45,7 @@ where
     ESRT: ExecutionStateRead<ST, SCT>,
 {
     // used so that ExecutionStateReadCache can maintain a logically immutable interface
```
xinyuan-dev: comment outdated.

xinyuan-dev (approved):

philipglazman (approved):

PR #3141, opened 2026-06-15
node: move prometheus options to config

dshulyak:


monad-node-config/src/lib.rs
```diff
@@ -52,6 +54,9 @@ pub struct NodeConfig<ST: CertificateSignatureRecoverable> {
     pub node_name: String,
     pub network_name: String,
 
+    #[serde(default)]
+    pub prometheus: PrometheusConfig,
```
xinyuan-dev: nit: maybe make this field into an `Option<PrometheusConfig>` and remove `impl Default on PrometheusConfig`? that'll better reflect that this field is optional and Prometheus endpoint disabled by default.

ariqchowdhury (dismissed):

monad-node-config/src/lib.rs
dshulyak: thats indeed cleaner

xinyuan-dev (approved):

ariqchowdhury (approved):

PR #3144, opened 2026-06-16
Enable atomic resolve+query on HistoricalDataSource

andr-dev:
The commit fixes a consistency gap in `HistoricalDataSource` usage. Previously, block tag resolution and data querying methods ran sequentially one after another. This means a pointer resolved by source A could be queried against source B, which may hold a different/inconsistent view. The new `try_resolve_and_then` implemented on the extension trait keeps resolve+query coupled within the same source. This is required for correct usage of the `DataSourceStack` module once we implement `HistoricalDataSource` for `BlockBufferView` which can resolve non-finalized blocks which could be inconsistent with what `TriedbReader` produces.

### DIFF
--- a/monad-rpc/src/data/mod.rs
+++ b/monad-rpc/src/data/mod.rs
@@ -49,7 +49,7 @@ use self::{
     },
 };
 use crate::{
-    data::source::{DataSourceError, TriedbDataSource},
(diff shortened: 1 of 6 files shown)


andr-dev:
@claude review thoroughly

PR #3145, opened 2026-06-17
Raptorcast: Reject group invites with unreasonable round span

xinyuan-dev:
Impose a defensive limit on the length of accepted group span to avoid abuse.

Note: we can alternatively implement this limit as a config entry on `FullNodeRaptorCastConfig` (e.g. `full_node_raptorcast.invite_max_round_span`). Doing so would require all node ops to update their node config for a small defensive change.

**Note to node operators**: Please review the `full_node_raptorcast.round_span` config if your node participates in secondary raptorcast as a publisher. Please cap this config to be at most `960`, or your secondary group invites may be refused by clients. For reference, the default value defined in devnet node.toml is `240`.

### DIFF
--- a/monad-raptorcast/src/raptorcast_secondary/client.rs
+++ b/monad-raptorcast/src/raptorcast_secondary/client.rs
@@ -33,6 +33,10 @@ use crate::{
     util::{SecondaryGroup, SecondaryGroupAssignment},
 };
 
+// Default full_node_raptorcast.round_span is 240. Refuse to accept
+// invites that exceed 4x that.
+pub(crate) const MAX_ACCEPTED_ROUND_SPAN: Round = Round(960);
+
 monad_executor::metric_consts! {
     pub CLIENT_NUM_CURRENT_GROUPS {
         name: "monad.bft.raptorcast.secondary.client.num_current_groups",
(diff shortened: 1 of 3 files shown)


monad-raptorcast/src/raptorcast_secondary/client.rs
```diff
@@ -572,13 +585,21 @@ mod tests {
...
                 start_round: Round(7),
                 end_round: Round(5),
             },
+            // unreasonably long round span
+            PrepareGroup {
+                max_group_size: 1,
+                validator_id: nid(2),
+                start_round: Round(5),
+                // 5 + MAX_ACCEPTED_ROUND_SPAN + 1
+                end_round: Round(966),
```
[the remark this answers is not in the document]
xinyuan-dev: i purposely hardcoded the value so a future change of the constant can be noticed.

eerkaijun (dismissed):

michael-yxchen:
we can validate round span config on startup here: https://github.com/category-labs/monad-bft/blob/ba3f20bdb49af7cef5555cfd854419226257300c/monad-raptorcast/src/raptorcast_secondary/publisher.rs#L227-L240

monad-node-config/src/fullnode_raptorcast.rs
```diff
@@ -41,6 +41,8 @@ pub struct FullNodeRaptorCastConfig<P: PubKey> {
 
     // config for Client
     pub max_num_group: usize,
+    #[serde(default = 960)]
+    pub invite_max_round_span: Round,
```
michael-yxchen: what's the use case for this?

xinyuan-dev: please ignore. it was accidentally slipped from my stashed changes when working with the option to make this a configuration entry. removed.

eerkaijun (approved):

michael-yxchen (approved):

marcograss (approved):

PR #3147, opened 2026-06-23
Move monad-test-util/src/swarm to monad-mock-swarm

larskuhtz:
Remove a cyclic dependency between test modules of monad-testutil and monad-mock-swarm that caused redundant rebuilds when building tests.

The cycle is broken by moving the module `monad_testutil::swarm` to `monad_mock_swarm::swarm`.


michael-yxchen (approved):
Please squash all commits into one before merging

eerkaijun (approved):

xinyuan-dev (approved):

PR #3165, opened 2026-07-10
Discrete event simulation engine for MockSwarm

larskuhtz:
A generic discrete-event scheduler for MockSwarm with support for Monte Carlo simulations. It allows simulation of heterogeneous components by abstracting over the internal state and logic of the simulated system. It uses a single event queue which
* simplifies the implementation,
* improves performance,
* allows for richer simulation metrics,
* provides robust control of non-determinism and randomness, and
* is more faithful with respect to production behavior than the current implementation.

New crates:
* `monad-sim`: a discrete event simulation engine. *It is completely agnostic about the application domain*. The design supports simulation of concurrent behavior through the notion of "Processes" that act on isolated state. Each discrete simulation step has access to the state of exactly one process. Additionally, a step is provided with a restricted view of the simulation context that gives access to the global clock and grants the capability to schedule simulation steps.
* `monad-sim-swarm`: provides generic components and features that are useful across different application domains. For now, this includes timers, a generic network component, and tools for declaratively assembling processes that are connected via a network into a "swarm".

*This PR introduces only the generic engine and some generic tools.* The code does not depend on any other crate in monad-bft. *The integration with the existing mock-swarm is provided in a separate PR.*

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6138,6 +6138,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "monad-sim-swarm"
+version = "0.1.0"
+dependencies = [
+ "monad-sim",
+ "rand 0.8.5",
+ "rand_chacha 0.3.1",
+]
+
+[[package]]
 name = "monad-state"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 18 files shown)


larskuhtz:
The changes in this PR got merged as part of #3203

PR #3170, opened 2026-07-16
Enable atomic resolve+query on HistoricalDataSource

andr-dev:


monad-rpc/src/data/source/stack.rs
```diff
@@ -91,19 +101,72 @@ where
...
+                Err(err) => {
+                    let err = split_terminating_error(err)?;
+                    last_err.get_or_insert(err);
+                    continue;
+                }
+            }
+        }
+
+        if let Some(ptr) = pointer {
+            for source in &self.sources {
```
eerkaijun: this still iterates through all sources instead of using source where `pointer` is resolved?

monad-rpc/src/data/source/historical.rs
```diff
@@ -83,3 +85,24 @@ pub trait HistoricalDataSource: DynClone + Send + Sync {
 clone_trait_object!(HistoricalDataSource);
 
 pub type HistoricalDataSourceStack = DataSourceStack<Box<dyn HistoricalDataSource>>;
+
+pub trait HistoricalDataSourceExt: HistoricalDataSource + Sized {
```
eerkaijun, line 89: just thinking out loud, is it necessary to introduce this trait extension? 

instead of just doing the atomic source query in HistoricalDataSource, something like
```
async fn function_name(
    &self,
    block: BlockTagOrHash,
) -> DataSourceResult< .... > {
    match self.try_resolve(block).await? {
        Some(pointer) => self.get_block(pointer).await,
        None => Ok(None),
    }
}   
```
ie a non generic default in HistoricalDataSource and still dyn compatible i think

andr-dev: The idea is that we should separate block pointer resolution from block data querying. Some sources, like the recent block cache in RPC, can resolve block pointer but can't query certain block data, say traces. If we have this separation, we need a way of expressing in some unified way that we want to do a resolution followed by some series data query methods. This trait is that expression.

monad-rpc/src/data/source/stack.rs
```diff
@@ -91,19 +101,79 @@ where
...
+        )
+            -> Pin<Box<dyn Future<Output = DataSourceResult<Option<U>>> + Send + 's>>,
+    {
+        let mut last_err = None;
+
+        let mut sources = self.sources.iter().peekable();
+
+        let pointer = loop {
+            let Some(source) = sources.peek() else {
+                return Ok(None);
```
eerkaijun: nit: should this be `return last_err.map_or(Ok(None), Err);`?

eerkaijun (dismissed):

eerkaijun (approved):

omegablitz (approved):

philipglazman (approved):

PR #3180, opened 2026-07-21
Only publish first accepted proposal to full nodes

eerkaijun:
Right now a validator will still publish an equivocated deterministic raptorcast proposal (if sent by a byzantine propose) to its full node, which would be flagged by the full nodes (although nothing happens right now since we just log them). With this commit, the validator only publishes the first accepted one.

### DIFF
--- a/monad-consensus-state/src/lib.rs
+++ b/monad-consensus-state/src/lib.rs
@@ -457,6 +457,10 @@ where
         cmds
     }
 
+    pub fn has_handled_proposal(&self, round: Round) -> bool {
+        !self.consensus.safety.is_safe_to_handle_proposal(round)
+    }
+
     /// handles proposal messages from other nodes
     /// validators and election are required as part of verifying the proposal certificates
     /// as well as determining the next leader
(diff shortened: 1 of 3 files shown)


xinyuan-dev:
i think we could also need similar guard for primary raptorcast as well because an honest publisher should ensure not publishing twice to avoid equivocation. Although I don't think it's currently possible for an honest validator to propose twice, I think this invariant should still be retained from the raptorcast side (because it's where commitment is enforced) Therefore, I wonder if this should be implemented from the publishing side within raptorcast. For each grouping key, (round, nodeid) for secondary and (round) for primary, we can guard only one such message should be built.

eerkaijun:
agree, I think it makes sense to add the invariant on the raptorcast layer for both primary and secondary raptorcast. I think this commit is still needed though, because we want the validator to only publish the first _accepted_ proposal to the full nodes, i.e. any proposals that get dropped during `handle_proposal_message` will not take up the invariant slot

xinyuan-dev:
> because we want the validator to only publish the first accepted proposal to the full nodes, i.e. any proposals that get dropped during handle_proposal_message will not take up the invariant slot

yup that makes sense indeed. especially we don't just wish not to equivocate through raptorcast, we also don't want to equivocate to the dedicated full nodes that don't care about raptorcast.

xinyuan-dev (dismissed):

michael-yxchen (dismissed):

xinyuan-dev (approved):

omegablitz (approved):

PR #3181, opened 2026-07-23
Add defensive invariant guard on raptorcast messages

eerkaijun:
Add a defensive guard at raptorcast layer such that a node never sends conflicting commitments for raptorcast messages. 

More context here: https://github.com/category-labs/monad-bft/pull/3180#issuecomment-5032725734

### DIFF
--- a/monad-raptorcast/src/round_info.rs
+++ b/monad-raptorcast/src/round_info.rs
@@ -13,7 +13,7 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-use std::collections::{BTreeMap, HashMap};
+use std::collections::{BTreeMap, BTreeSet, HashMap};
 
 use monad_crypto::certificate_signature::PubKey;
 use monad_types::{NodeId, Round};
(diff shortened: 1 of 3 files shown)


monad-raptorcast/src/round_info.rs
```diff
@@ -306,6 +306,37 @@ fn try_commit_into<PT: PubKey>(
     None
 }
 
+// Max number of published rounds to keep track of
+pub(crate) const PUBLISHED_ROUNDS_CACHE_SIZE: usize = 100;
```
eerkaijun, line 310: we need this to keep track of sent rounds and make sure it's not unbounded, but feels like it also slightly defeats the purpose of the invariants

xinyuan-dev: i doubt if there is a way to perfectly adhere to the invariant for all rounds while keeping bounded memory.

it may be possible to reuse the round info cache to record the flag, as such would semantically bind the two invariants (publishing commitment & receiving commitment) so the limitation on # of rounds aligns up. but i'm not sure if it's a good idea to mix receiving & publishing side state because it feels awkward especially for secondary rc where the publisher and full nodes are asymmetric (i.e. publisher never receives their own message).

anyway, i think given that we're now only adding this check defensively, it's probably fine to just bound it conveniently as the current implementation. given that the publishing end is generally considred under the node's own control so unless there is a bug the invariant shouldn't be broken typically.

xinyuan-dev (dismissed):

xinyuan-dev (approved):

xinyuan-dev (approved):

michael-yxchen (approved):

PR #3185, opened 2026-07-30
Add slot number and BAL-hash header fields gated on V_NEXT

ryankeleti:
Adds `MonadExecutionRevision::V_NEXT` with `amsterdam_enabled: true` gating two new `ProposedEthHeader` fields:

- `block_access_list_hash` (EIP-7928), always zeroed, as monad doesn't implement block access lists. Like `requests_hash`, it's required to be populated for RLP encoding/decoding of `slot_number`
- `slot_number` (EIP-7843), set to the consensus block round

`V_NEXT` is activated at 0 on devnet and left as `u64::MAX` on testnet and mainnet.

Will bump `monad-execution` submodule on submit, changes are in PR here: https://github.com/category-labs/monad/pull/2439

### DIFF
--- a/monad-chain-config/src/execution_revision.rs
+++ b/monad-chain-config/src/execution_revision.rs
@@ -20,10 +20,11 @@ pub enum MonadExecutionRevision {
     V_ONE,
     V_TWO,
     V_FOUR,
+    V_NEXT,
 }
 
 impl MonadExecutionRevision {
-    pub const LATEST: Self = Self::V_FOUR;
+    pub const LATEST: Self = Self::V_NEXT;
 }
 
 impl MonadExecutionRevision {
@@ -33,6 +34,7 @@ impl MonadExecutionRevision {
             Self::V_ONE => &EXECUTION_CHAIN_PARAMS_V_ONE,
             Self::V_TWO => &EXECUTION_CHAIN_PARAMS_V_TWO,
             Self::V_FOUR => &EXECUTION_CHAIN_PARAMS_V_FOUR,
+            Self::V_NEXT => &EXECUTION_CHAIN_PARAMS_V_NEXT,
         }
     }
 }
@@ -41,29 +43,41 @@ impl MonadExecutionRevision {
 pub struct ExecutionChainParams {
     pub max_code_size: usize,
     pub prague_enabled: bool,
+    pub amsterdam_enabled: bool,
     pub validate_system_txs: bool,
 }
 
 const EXECUTION_CHAIN_PARAMS_V_ZERO: ExecutionChainParams = ExecutionChainParams {
     max_code_size: 24 * 1024,
     prague_enabled: false,
+    amsterdam_enabled: false,
     validate_system_txs: false,
 };
 
 const EXECUTION_CHAIN_PARAMS_V_ONE: ExecutionChainParams = ExecutionChainParams {
     max_code_size: 24 * 1024,
     prague_enabled: false,
+    amsterdam_enabled: false,
     validate_system_txs: false,
 };
 
 const EXECUTION_CHAIN_PARAMS_V_TWO: ExecutionChainParams = ExecutionChainParams {
     max_code_size: 128 * 1024,
     prague_enabled: false,
+    amsterdam_enabled: false,
     validate_system_txs: false,
 };
 
 const EXECUTION_CHAIN_PARAMS_V_FOUR: ExecutionChainParams = ExecutionChainParams {
     max_code_size: 128 * 1024,
     prague_enabled: true,
+    amsterdam_enabled: false,
+    validate_system_txs: true,
+};
+
+const EXECUTION_CHAIN_PARAMS_V_NEXT: ExecutionChainParams = ExecutionChainParams {
+    max_code_size: 128 * 1024,
+    prague_enabled: true,
+    amsterdam_enabled: true,
     validate_system_txs: true,
 };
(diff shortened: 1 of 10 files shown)


monad-chain-config/src/execution_revision.rs
```diff
@@ -20,10 +20,11 @@ pub enum MonadExecutionRevision {
     V_ONE,
     V_TWO,
     V_FOUR,
+    V_NEXT,
 }
 
 impl MonadExecutionRevision {
-    pub const LATEST: Self = Self::V_FOUR;
+    pub const LATEST: Self = Self::V_NEXT;
```
Baltoli, line 27: Not _immediately_ clear to me whether this is correct or not - need to review, but it might be the case that it's best kept at the latest pre-NEXT revision

monad-eth-types/src/lib.rs
```diff
@@ -441,6 +475,8 @@ mod test {
             excess_blob_gas: 0,
             parent_beacon_block_root: [0_u8; 32],
             requests_hash: None,
+            block_access_list_hash: None,
+            slot_number: None,
         };
```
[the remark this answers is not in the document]
ryankeleti: Will add

monad-chain-config/src/lib.rs
```diff
@@ -66,6 +66,7 @@ pub struct MonadChainConfig {
     pub execution_v_one_activation: u64,
     pub execution_v_two_activation: u64,
     pub execution_v_four_activation: u64,
+    pub execution_v_next_activation: u64,
 }
```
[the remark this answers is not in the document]
Baltoli: Might need some thought for what a sensible default is, or maybe we shouldn't provide one (and just let configuring users handle it?). See [here](https://github.com/category-labs/monad-integration/blob/9cb58f1154af00cc91d09dd569808e5e187e7544/net-lib/monad_flexnet/topology/config/chain_config.py#L24) for how this is used in Flexnet

monad-eth-types/src/lib.rs
```diff
@@ -158,6 +172,14 @@ impl Encodable for ProposedEthHeader {
         if let Some(requests_hash) = &self.requests_hash {
             requests_hash.encode(out);
         }
+
+        if let Some(block_access_list_hash) = &self.block_access_list_hash {
```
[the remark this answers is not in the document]
Baltoli: Seems reasonable to add this assertion; I think your execution code does this already

monad-chain-config/src/lib.rs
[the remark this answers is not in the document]
ryankeleti: Maybe mirroring the rust code with something like `execution_v_next_activation: int = sys.maxsize` in flexnet's `ExecutionVersions`?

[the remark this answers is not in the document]
Baltoli: Let's just set no default here and let flexnet / other users override - I suspect we will want NEXT to get activated unconditionally in those environments anyway

monad-chain-config/src/lib.rs
```diff
@@ -210,6 +214,8 @@ const MONAD_TESTNET_CHAIN_CONFIG: MonadChainConfig = MonadChainConfig {
     execution_v_one_activation: 0,
     execution_v_two_activation: 0,
     execution_v_four_activation: 0,
+    // amsterdam not yet on testnet
+    execution_v_next_activation: u64::MAX,
```
Baltoli, line 218: Add a TODO for Ellis here - Hive isn't testnet, but rather a mysterious third thing that will want different handling. The value should probably be 0 as well.

monad-eth-txpool/src/pool/mod.rs
```diff
@@ -389,6 +389,28 @@ where
...
+
+        let maybe_slot_number = if self
+            .execution_revision
+            .execution_chain_params()
+            .amsterdam_enabled
+        {
+            Some(round.as_u64())
+        } else {
+            None
+        };
```
Baltoli: Nit - we probably won't implement BALs at all so "yet" is maybe redundant.

I'm also not a Rust guru but maybe fold these together into one `let`:
```rust
let (maybe_block_access_list_hash, maybe_slot_number) = if self ...
```

Baltoli (commented):
All looks basically good to me - let's get the execution stack of PRs to a point where we're happy to merge, then we can get feedback on these changes from consensus folks

monad-chain-config/src/lib.rs
ryankeleti: Added a TODO and set to 0 - @eclark0426 does this seem right?

eclark0426: I think Github's way of collapsing the diff may have caused some confusion, the edit there _was_ on Testnet, not hive, although the collapsed view does make it look like it was on the hive section. Hive's config is unchanged, and just takes all of the configs from devnet except for the chain ID. So hive is already getting 0, not `u64::MAX`

Anyway, all this to say, the original edit this comment was on looks fine as-is, and I don't think needs a todo. I may need to regenerate some fixtures if there's any differences in gas pricing, which we'll find out the first time hive runs

eclark0426: Also, if the comment is accurate and Amsterdam is not on testnet, this line should have stayed at `u64::MAX`, not been changed to 0

ryankeleti: That makes sense, thank you!

ryankeleti:
Failing build is due to bumping monad-execution while waiting on https://github.com/category-labs/monad-bft/pull/3214 to be merged

eerkaijun:
i think this looks good to me, we can get CI to be green before merging

PR #3186, opened 2026-07-30
Update codeowners

andr-dev:


andr-dev:
@claude review.

.github/CODEOWNERS
```diff
@@ -1,9 +1,9 @@
 * @omegablitz @ariqchowdhury @michael-yxchen @xinyuan-dev @dshulyak
 
-/monad-archive/ @JoeHowarth @omegablitz @ariqchowdhury @michael-yxchen
-/monad-rpc/ @JoeHowarth @omegablitz @ariqchowdhury @michael-yxchen
-/monad-eth-testutil/ @JoeHowarth @omegablitz @ariqchowdhury @michael-yxchen
-/monad-chain-data/ @JoeHowarth @andr-dev @omegablitz @ariqchowdhury @michael-yxchen
+/Cargo.{toml,lock} @andr-dev @eclark0426 @omegablitz @ariqchowdhury @michael-yxchen
```
0xSqualo: @andr-dev I didn't think `CODEOWNERS` supports brace expansion, so `/Cargo.{toml,lock}` matches a literal path named `Cargo.{toml,lock}` (which never exists) rather than the root `Cargo.toml` and `Cargo.lock`. 

Both files fall through to the `*` owners on line 1 instead of the owners listed here - is that intended? To gate the root manifest + lockfile it needs two separate patterns:
```
/Cargo.toml @andr-dev @eclark0426 @omegablitz @ariqchowdhury @michael-yxchen
/Cargo.lock @andr-dev @eclark0426 @omegablitz @ariqchowdhury @michael-yxchen
```

.github/CODEOWNERS
```diff
@@ -1,9 +1,10 @@
 * @omegablitz @ariqchowdhury @michael-yxchen @xinyuan-dev @dshulyak
 
-/monad-archive/ @JoeHowarth @omegablitz @ariqchowdhury @michael-yxchen
-/monad-rpc/ @JoeHowarth @omegablitz @ariqchowdhury @michael-yxchen
-/monad-eth-testutil/ @JoeHowarth @omegablitz @ariqchowdhury @michael-yxchen
-/monad-chain-data/ @JoeHowarth @andr-dev @omegablitz @ariqchowdhury @michael-yxchen
+/Cargo.toml @andr-dev @eclark0426 @omegablitz @ariqchowdhury @michael-yxchen
+/Cargo.lock @andr-dev @eclark0426 @omegablitz @ariqchowdhury @michael-yxchen
```
dshulyak, line 4: btw it is weird that this is more restrictive than *

andr-dev: I applied the CO that makes sense in this context given Ellis and I have historically handled dependency management. Archive and RPC are similarly more restrictive.

ariqchowdhury (approved):

omegablitz (approved):

PR #3187, opened 2026-08-03
Update ruint

eclark0426:
Resolves https://rustsec.org/advisories/RUSTSEC-2026-0220

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1095,6 +1095,23 @@ dependencies = [
 ]
 
 [[package]]
+name = "ark-ff"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


eclark0426:
> These packages are the same versions in both base and PR, yet their resolved transitive deps changed. This suggests the lockfile was regenerated with cargo update (full) rather than cargo update -p ruint (targeted), or with a different cargo/resolver version.

It was run with `-p`, I found the additional diffs strange too. Ultimately the win packages aren't compiled for linux though so it shouldn't be functionally any different, I don't think

dshulyak (dismissed):

omegablitz (dismissed):

dshulyak (approved):

omegablitz (approved):

PR #3188, opened 2026-08-03
monad-node: export triedb disk capacity/usage as Prometheus gauges

maxkozlovsky:
Add monad.triedb.disk_capacity_bytes and monad.triedb.disk_used_bytes, sourced from the storage pool via the new TriedbReader::storage_stats(). Unlike the read-once migration-phase metric, disk usage is dynamic, so a long-lived read-only TriedbReader is kept on the main task (it is !Send) and both gauges are refreshed on a 30s ticker.

Bumps the monad-execution submodule to pick up the storage_stats FFI.

### DIFF
--- a/monad-node/src/main.rs
+++ b/monad-node/src/main.rs
@@ -78,15 +78,19 @@ use monad_wal::wal::{WALLog, WALoggerConfig};
 use opentelemetry::metrics::{Gauge, Meter, MeterProvider};
 use opentelemetry_otlp::{MetricExporter, WithExportConfig};
 use rand_chacha::{rand_core::SeedableRng, ChaCha8Rng};
-use tokio::signal::unix::{signal, SignalKind};
+use tokio::{
+    signal::unix::{signal, SignalKind},
+    time::{interval, MissedTickBehavior},
+};
 use tracing::{error, event, info, warn, Instrument, Level};
 
 use self::{
     cli::Cli,
     error::NodeSetupError,
     metrics::{
-        default_prometheus_labels, init_triedb_phase_metrics, record_triedb_phase_metrics,
-        start_metrics_server, MetricsServerState, NodePrometheusMetrics,
+        default_prometheus_labels, init_triedb_phase_metrics, init_triedb_storage_metrics,
+        record_triedb_phase_metrics, record_triedb_storage_metrics, start_metrics_server,
+        MetricsServerState, NodePrometheusMetrics,
     },
     state::NodeState,
 };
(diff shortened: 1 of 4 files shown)


monad-node/src/main.rs
```diff
@@ -528,6 +540,13 @@ async fn run(node_state: NodeState) -> Result<(), ()> {
     let mut sigterm = signal(SignalKind::terminate()).expect("in tokio rt");
     let mut sigint = signal(SignalKind::interrupt()).expect("in tokio rt");
 
+    // The !Send reader lives on this task, so the refresh has to run here. It
+    // reaches the scrape only because init_triedb_storage_metrics() created
+    // these gauges before NodePrometheusMetrics::new registered clones of them:
+    // a gauge first touched after registration is silently never scraped.
+    let mut triedb_storage_ticker = tokio::time::interval(TRIEDB_STORAGE_METRICS_INTERVAL);
+    triedb_storage_ticker.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
```
dshulyak: use imports?

monad-node/src/main.rs
```diff
@@ -110,6 +111,7 @@ const STATESYNC_REQUEST_TIMEOUT: Duration = Duration::from_secs(10);
 
 const EXECUTION_DELAY: u64 = 3;
 const WALTRACE_CHANNEL_CAPACITY: usize = 1024;
+const TRIEDB_STORAGE_METRICS_INTERVAL: Duration = Duration::from_secs(30);
```
dshulyak, line 117: this can be done without additional scraping delay and in general this is unnecessary workaround. in prometheus library  there is a PullingGauge type that can be reused in monad  metrics shim. or triedb c bindings can expose its own data source and implement collector trait for that. in both cases metrics will be read when scraped

monad-node/src/main.rs
```diff
@@ -528,6 +540,13 @@ async fn run(node_state: NodeState) -> Result<(), ()> {
     let mut sigterm = signal(SignalKind::terminate()).expect("in tokio rt");
     let mut sigint = signal(SignalKind::interrupt()).expect("in tokio rt");
 
+    // The !Send reader lives on this task, so the refresh has to run here. It
+    // reaches the scrape only because init_triedb_storage_metrics() created
+    // these gauges before NodePrometheusMetrics::new registered clones of them:
+    // a gauge first touched after registration is silently never scraped.
```
dshulyak: i think this comment can be removed

monad-node/src/main.rs
dshulyak: i guess i can just let agent do this refactoring later, so its ok to ignore

dshulyak (dismissed):

monad-node/src/main.rs
maxkozlovsky: the cost is not exactly trivial, and the value does not meaningfully change often, so I thought it'll be better to not have it pull the value. It can be revisited later.

dshulyak (approved):

michael-yxchen (approved):

PR #3192, opened 2026-08-04
hive: support running from a caller-supplied prebuilt image

eclark0426:


eclark0426:
May or may not actually be needed, the failures might have just been due to Ubuntu outages. Regardless, I think this would bring us in the right direction of a unified hive runner/monad-integration.

PR #3198, opened 2026-08-07
ci(builder): publish builder image to ghcr.io

mhumeSF:
Push the builder image to ghcr.io using the built-in `GITHUB_TOKEN`. Merging this triggers the first build. Consumer workflows switch over in a follow-up.

### DIFF
--- a/.github/workflows/build-builder.yml
+++ b/.github/workflows/build-builder.yml
@@ -5,6 +5,7 @@ name: Build and publish builder Docker image
   push:
     paths:
       - 'docker/builder/**'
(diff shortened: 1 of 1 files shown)


Yb2411 (approved):

michael-yxchen (commented):
`.github/workflows/test-command.yml` and `.github/workflows/fuzz.yml` both refer to the builder image at peach10 repo. should those be updated?

mhumeSF:
@michael-yxchen That'll happen in a follow-up PR; the other workflows will fail with no initial image published

michael-yxchen (approved):

PR #3207, opened 2026-08-14
feat(dataplane): parallelize UDP transmit

dshulyak:
- kernel: linux 6.17.0-35-generic
- cpu: amd epyc 4584px, 16 cores; 16 mib l2 and 128 mib l3 cache

first test runs rebroadcast with 45.6% concentration on a single validator, with 31 targets.
throughput peaks at 7000 mbps with 5 tx workers, bottlenecked by raptorcast thread, after that adding more workers will decrease throughput due to coordination overhead. can be further improved by moving wireauth to dataplane thread, which is already planned at some point.

for 120MiB obligation, e.g 13% stake concentration on mainnet 2 workers should be enough.

image
image

the second test is throughput.rs that doesn't involve raptorcast thread, meters only dataplane scaling. targets are sharded, but with some variance.

| UDP TX workers | Targets | Sender throughput |
 |---:|---:|---:|
| 4 | 64 | 7.212 Gbps |
| 6 | 96 | 10.179 Gbps | 
| 8 | 128 | 11.864 Gbps |

#### ops

adds new config option
```
[network]
udp_tx_workers = 4
```
Valid range: 1–64. If omitted, it defaults to 1.


dshulyak:
@claude

xinyuan-dev (dismissed):

dshulyak:
@claude

xinyuan-dev (approved):

PR #3211, opened 2026-08-17
statesync: retire v0/v1, derive the version constants from execution

maxkozlovsky:
Needs category-labs/monad#2487 merged first (submodule bump).

The protocol version numbers live in execution's `statesync_version.h`.
`StateSyncVersion` stays here because the wire encoding is ours, but its
`{major, minor}` now come from that header through `monad-statesync-version`, so
there is one definition rather than two that happen to agree.

That crate carries only the constants, with no `monad-cxx` dependency:
`monad-debugger` reaches `monad-executor-glue`, and `cargo check --target
wasm32-unknown-unknown -p monad-debugger` is a CI job that the C++ FFI crate
would break.

Raising `STATESYNC_VERSION_MIN` to v2 retires the v0 upsert encoding with it:
`StateSyncUpsertV0` and the version fork in `StateSyncResponse`'s `encode`,
`length` and `decode` are gone, as is the pre-v2 branch in the ipc server's send
gating. A peer still speaking v0 or v1 now gets `BadVersion`.

### Testing

- `statesync_version_aliases_match_execution` — deriving the rungs does not pin
  *which* rung each side calls current and which it calls the floor; those stay
  two independent edits, so this asserts `SELF_STATESYNC_VERSION` and
  `STATESYNC_VERSION_MIN` against execution's. Verified it fails
  (`65538 != 65539`) when only execution's is bumped.
- `statesync_version_is_compatible` now pins the negative case: `{1,0}` and
  `{1,1}` are refused.
- `statesync_response_roundtrip` replaces the three v0/v1 format tests and adds
  an `encode()`/`length()` agreement check, which the deleted format-selection
  test was the only thing covering.

### DIFF
--- a/monad-executor-glue/src/lib.rs
+++ b/monad-executor-glue/src/lib.rs
@@ -49,6 +49,7 @@ use monad_crypto::certificate_signature::{
     CertificateSignaturePubKey, CertificateSignatureRecoverable, PubKey,
 };
 use monad_execution_state_read::ExecutionStateRead;
+use monad_statesync_version::{MONAD_STATESYNC_MAJOR, MONAD_STATESYNC_MINOR_2};
 use monad_types::{
     deserialize_pubkey, serialize_pubkey, Epoch, ExecutionProtocol, FullnodeBroadcastMode,
     LimitedVec, NodeId, Round, RouterTarget, SeqNum, Stake, UdpPriority,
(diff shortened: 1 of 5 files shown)


monad-executor-glue/src/lib.rs
```diff
@@ -1460,12 +1460,11 @@ where
     }
 }
 
-const STATESYNC_VERSION_V0: StateSyncVersion = StateSyncVersion { major: 1, minor: 0 };
-const STATESYNC_VERSION_V1: StateSyncVersion = StateSyncVersion { major: 1, minor: 1 };
-// Client is required to send completions since this version
+// Client is required to send completions since this version. Minors 0 and 1
+// predate that and are no longer spoken; nothing deployed still runs them.
 pub const STATESYNC_VERSION_V2: StateSyncVersion = StateSyncVersion { major: 1, minor: 2 };
```
Chen-Yifan: should also mirror the `MONAD_STATESYNC_MAJOR` and `MONAD_STATESYNC_MINOR_V2` enums from execution for the major minor here.

Chen-Yifan (approved):

ariqchowdhury (approved):

PR #3214, opened 2026-08-18
[rpc] Adopt new `monad_ethcall` Result-annotated interface

dhil:
This patch adopts the new result-oriented interface of the upstream `monad_ethcall` crate along with the new error types.

Companion patch: https://github.com/category-labs/monad/pull/2381

### DIFF
--- a/monad-rpc/src/handlers/eth/call.rs
+++ b/monad-rpc/src/handlers/eth/call.rs
@@ -24,8 +24,7 @@ use alloy_primitives::{Address, Bytes, Signature, TxKind, Uint, B256, U256, U64,
 use alloy_rpc_types::{AccessList, AccessListItem};
 use monad_chain_config::execution_revision::MonadExecutionRevision;
 use monad_ethcall::{
-    overrides::StateOverrideSet, CallResult, EthCallRequest, EthCallResult, FailureCallResult,
-    MonadExecutor, MonadTracer,
+    overrides::StateOverrideSet, EthCallRequest, EthCallResult, MonadExecutor, MonadTracer,
 };
 use monad_rpc_docs::rpc;
 use monad_triedb_utils::triedb_env::{
@@ -43,6 +42,7 @@ use crate::{
     },
     handlers::{
         debug::{decode_call_frame, TracerObject},
+        eth::{CallResult, FailureCallResult, SuccessCallResult},
         parse_ethcall_chain_id,
     },
     types::{
@@ -664,7 +664,7 @@ async fn prepare_eth_call_at_block<T: Triedb + TriedbPath>(
 
     let header_gas_limit = header.header.gas_limit;
 
-    match eth_call_executor
+    let call_result = match eth_call_executor
         .eth_call(EthCallRequest {
             chain_id: ethcall_chain_id,
             transaction: &txn,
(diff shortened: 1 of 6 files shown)


monad-rpc/src/handlers/eth/call.rs
```diff
@@ -678,9 +678,15 @@ async fn prepare_eth_call_at_block<T: Triedb + TriedbPath>(
...
     {
-        CallResult::Failure(error) if matches!(error.error_code, EthCallResult::OutOfGas) => {
-            match out_of_gas_handling {
-                OutOfGasHandling::RpcError => Err(JsonRpcError::eth_call_error(
+        Ok(result) => result.into(),
+        Err(monad_ethcall::EthCallError::Failure {
+            error_code,
+            gas_used,
+            gas_refund,
+            ..
```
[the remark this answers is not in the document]
dhil: This is consistent with the current behaviour of the code. I want to preserve semantic parity for now.

monad-rpc/src/types/jsonrpc.rs
```diff
@@ -550,17 +550,60 @@ impl From<monad_archive::prelude::Report> for JsonRpcError {
...
 }
 
+impl From<monad_ethcall::EthTraceError> for JsonRpcError {
+    fn from(error: monad_ethcall::EthTraceError) -> Self {
+        use monad_ethcall::EthTraceError::*;
+        let message = match error {
+            InternalError => String::from("internal error"),
+            Other(message) => message,
+        };
+        Self::eth_call_error(message, None)
```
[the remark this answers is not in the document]
dhil: Fixed.

monad-rpc/src/handlers/eth/mod.rs
```diff
@@ -13,9 +13,128 @@
...
+            }),
+            EthCallError::ReserveBalanceViolation {
+                gas_used,
+                gas_refund,
+            } => Self::Failure(FailureCallResult {
+                error_code: EthCallResult::ReserveBalanceViolation,
+                gas_used,
+                gas_refund,
+                message: String::from("reserve balance violation"),
+                data: None,
```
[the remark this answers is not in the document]
dhil: Fixed.

dhil:
@claude please re-review this PR, taking into account the past comments and their resolutions.

dhil:
@claude please take another look. I've addressed your previous comments and tighten the data payload types such that we do not need to copy `Box<[u8]>` into an intermediary `Vec<u8>`.

eerkaijun (approved):

monad-rpc/src/handlers/eth/mod.rs
```diff
@@ -13,9 +13,125 @@
...
+    fn from(result: SuccessCallResult) -> Self {
+        Self::Success(result)
+    }
+}
+
+impl From<EthCallSuccess> for CallResult {
+    fn from(result: EthCallSuccess) -> Self {
+        Self::Success(result.into())
+    }
+}
```
andr-dev, line 60: These methods just create an enum variant, why not inline the enum creation?

philipglazman (approved):

monad-rpc/src/handlers/eth/gas.rs
```diff
@@ -56,6 +59,15 @@ use crate::{
...
 const CALL_STIPEND: u64 = 2_300;
 
+fn into_call_result(
+    result: Result<monad_ethcall::EthCallSuccess, monad_ethcall::EthCallError>,
+) -> CallResult {
+    match result {
+        Ok(success) => success.into(),
+        Err(error) => error.into(),
+    }
+}
```
andr-dev, line 69: This method is only used twice, I think we can just turn `expr` into `Ok(expr?.into())` at the callsites to keep things clean.

monad-rpc/src/handlers/eth/mod.rs
dhil: It was just be able to succinctly use .into() in the code of interest. Although, it is probably only really useful for the error case.

monad-rpc/src/handlers/eth/gas.rs
dhil: It is used inside two lambda contexts which don't allow the use of `?`. I think we'd have to change their types, which would have further cascading changes (which I would rather do in another PR).

monad-rpc/src/handlers/eth/mod.rs
dhil: Do you consider this a blocker for merging?

monad-rpc/src/handlers/eth/gas.rs
dhil: Likewise, is this a blocker for merging?

michael-yxchen (approved):

PR #3223, opened 2026-08-24
monad-raptorcast: avoid redundant wireauth lookups

dshulyak:
perf profiling of the raptorcast service thread under a 64-target load showed repeated public-key hashing and wireauth session lookup in the authenticated send path.

reuse the authenticated socket address after validating that its active session belongs to the target public key. also avoid rechecking connected and initiating sessions for each piece of a buffered message.

in the throughput stress test with 2 udp tx workers, a 25.625 gbps offer, a 25 gbps global limit, and a 1 gbps per-peer limit, raptorcast throughput increased from 15,796 to 18,427 chunks/s (+16.7%). authenticated dataplane handoff increased from 11.646 to 13.586 gbps (+16.7%), while udp tx remained saturated at about 3.5 gbps.

### DIFF
--- a/monad-raptorcast/src/auth/socket.rs
+++ b/monad-raptorcast/src/auth/socket.rs
@@ -201,11 +201,14 @@ where
                 }
 
                 let msg_len = msg.len() as u64;
-                if self
-                    .authenticated
-                    .write_with_buffering(public_key, msg, stride, priority)
-                    .is_err()
-                {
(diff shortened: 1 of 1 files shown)


dshulyak:
@claude

omegablitz (approved):

michael-yxchen (approved):

PR #3224, opened 2026-08-24
monad-node: run consensus state on dedicated thread

dshulyak:
full proposals contain 3,750 transactions. across 86 full proposals matched by round between original and
modified clients, proposal handling on the original averaged 72.4 ms, with 71.3 ms p50, 84.2 ms p95, and
90.0 ms max. at the target 3.3 full blocks per second, the mean corresponds to about 239 ms of blocked
main-thread time per second.

while state.update runs inline, main cannot poll raptorcast or the txpool client. bft cannot rebroadcast, service
wireauth requests, receive network transactions into mempool processing, or consume mempool events until the
update returns.

move consensus state, consensus-local executors, and ledger commits to monad-consensus. keep raptorcast, txpool
access, discovery, state sync, control-panel work, and router publication on main. bound both cross-thread channels
at 1,024 items.

instrument state updates, signer recovery, ledger persistence, configuration writes, and execution-state-read
waits so blocking and io remain observable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4996,6 +4996,7 @@ dependencies = [
  "monad-types",
  "monad-validator",
  "tempfile",
+ "tracing",
 ]
 
 [[package]]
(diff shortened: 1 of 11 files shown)


dshulyak:
@claude

dshulyak:
@claude

PR #3226, opened 2026-08-24
Fix error code for insufficient balance

philipglazman:
Previously we would return an internal error code for errors that were related to eth_call insufficient balance. 

This is an immediate fix, but the ideal fix is to build off of https://github.com/category-labs/monad-bft/pull/3214 and pass insufficient balance error reason as an enum variant through the FFI boundary.  

flexnet: https://github.com/category-labs/monad-integration/pull/460

### DIFF
--- a/monad-rpc/src/types/jsonrpc.rs
+++ b/monad-rpc/src/types/jsonrpc.rs
@@ -489,8 +489,8 @@ impl JsonRpcError {
 
     pub fn insufficient_funds() -> Self {
         Self {
-            code: -32003,
-            message: "Insufficient funds for gas * price + value".into(),
+            code: -32000,
+            message: "insufficient funds for gas * price + value".into(),
             data: None,
         }
     }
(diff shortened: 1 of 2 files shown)


monad-rpc/src/types/jsonrpc.rs
```diff
@@ -556,14 +556,21 @@ impl From<monad_ethcall::FailureCallResult> for JsonRpcError {
             monad_ethcall::EthCallResult::ExecutionError => {
                 Self::eth_call_execution_revert(error.message, error.data)
             }
+            monad_ethcall::EthCallResult::OtherError
+                if matches!(
+                    error.message.as_str(),
+                    "insufficient balance" | "insufficient balance for fee"
```
eerkaijun, line 562: rather than string matching, I think ideally in `EthCallResult` we can add a variant of InsufficientBalance, which is propagated by `TransactionError::InsufficientBalance` or `MonadTransactionError::InsufficientBalanceForFee` in execution

eerkaijun: oops sorry saw your PR description, yeah agree it's better to extend #3214

eerkaijun:
do we also want to change this for the eth_simulateV1 path?

philipglazman:
> do we also want to change this for the eth_simulateV1 path?

we call `sender_gas_allowance` during that path which would return the right error

philipglazman:
Users experiencing this issue have a workaround so i will close this infavor of working towards the ideal solution

PR #3248, opened 2026-09-08
Raptorcast: fix commitment

xinyuan-dev:
This PR implements the validation of secp256k1 signature and restrict deterministic raptorcast commitment to a subset of headers that's relevant to encoding scheme, excluding the signature itself.

### DIFF
--- a/monad-raptorcast/src/round_info.rs
+++ b/monad-raptorcast/src/round_info.rs
@@ -25,7 +25,6 @@ use crate::{
     },
     udp::ValidatedChunk,
     util::{EncodingScheme, GlobalMerkleRoot, PrimaryBroadcastGroup, SecondaryBroadcastGroup},
-    SIGNATURE_SIZE,
 };
 
 pub(crate) const CACHE_MAX_FUTURE_ROUNDS: Round = Round(100);
@@ -294,11 +293,8 @@ fn try_commit_into<PT: PubKey>(
     if !commitment.conflict_logged {
         tracing::error!(
             author = ?chunk.author,
-            round = ?claim.round,
-            chunk_merkle_root = ?claim.global_merkle_root,
-            commit_merkle_root = ?commitment.global_merkle_root,
-            chunk_signature = ?claim.signature,
-            commit_signature = ?commitment.signature,
+            chunk_claim = ?claim,
+            committed_claim = ?commitment.claim,
             "Conflicting commitment"
         );
         commitment.conflict_logged = true;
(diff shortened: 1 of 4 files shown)


monad-raptorcast/src/round_info.rs
```diff
@@ -337,68 +333,61 @@ impl PublishedRounds {
...
-    fn try_from(chunk: &'a ValidatedChunk<PT>) -> Result<Self, ()> {
+    fn try_from(chunk: &ValidatedChunk<PT>) -> Result<Self, ()> {
         let round = match chunk.encoding_scheme {
             EncodingScheme::Deterministic25(round) => round,
             EncodingScheme::Unspecified => return Err(()), // not applicable
         };
+        let encoding_scheme_variant = chunk
+            .encoding_scheme
+            .variant()
+            .expect("deterministic rc must have encoding scheme variant");
```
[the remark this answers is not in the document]
xinyuan-dev: D25 is already matched immediately above so the `expect` is safe here.

monad-secp/src/secp.rs
```diff
@@ -239,7 +239,8 @@ impl SecpSignature {
         sig_vec.try_into().unwrap()
     }
 
-    /// Deserialize the signature
+    /// Deserialize the signature, accepting only the low-S form so each
+    /// signature has exactly one byte representation.
```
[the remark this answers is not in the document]
xinyuan-dev: comment updated.

marcograss (dismissed):

monad-raptorcast/src/round_info.rs
```diff
@@ -337,68 +333,61 @@ impl PublishedRounds {
...
-            .expect("signature of validated chunk must have correct length");
 
         Ok(Self {
-            signature,
-            global_merkle_root,
             round,
+            global_merkle_root: *global_merkle_root,
+            encoding_scheme_variant,
+            app_message_len: chunk.app_message_len,
+            coarse_ts: deterministic::coarse_ts(chunk.unix_ts_ms),
```
eerkaijun: if i understand correctly, the decoder cache still uses milliseconds precision, while the commitment uses coarse precision. so technically a proposer can form chunks that bypasses the decoder cache?

xinyuan-dev: indeed. i think the commitment as a whole should identify the signed header - so two different signed headers must map to the same commitment. it's the merkle root's role to identify the symbol assignment/encoding, so coarse_ts doesn't help here.

xinyuan-dev: fixed. now the commitment acts on the equivalence of the full header (with https://github.com/category-labs/monad-bft/pull/3247).

eerkaijun (approved):

PR #3257, opened 2026-09-14
Simplified chain certificate

michael-yxchen:
A node broadcasts an unsigned CapAdvance{cap} whenever its finalization cap moves. A peer cap at least `lag_threshold` ahead is trusted as a stand-in for the chain-state certificate: the tracker jumps to it, and the conductor closes the skipped slots and reopens windows from the new cap at their natural deadlines, abandoning the undecided deadline-agreement round.

ConductorMessage now multiplexes deadline agreement and cap advances over a tag-prefixed RLP encoding, ConductorConfig takes the lag_threshold, and handle_message gains the arrival timestamp.

No signatures or quorum yet: a single peer's message is a certificate. Block and DA sync for skipped slots are out of scope.

This code was generated using Claude Opus 5.


monad-mcp-chorus/src/conductor/mod.rs
```diff
@@ -388,6 +515,25 @@ where
         Ok(())
     }
 
+    // Slots below the cap stay closed and skipped windows keep their natural
+    // deadline schedule, matching the TODO at deadline_agreement.rs:70.
+    fn open_window_after_jump(&mut self, cap: Slot) -> Result<(), ConductorError> {
```
xinyuan-dev, line 520: nit: this seems a bit inefficient to try open each in-between windows until the cap is reached. given that cap is already given, would it be possible to jump to the desired window immediately? maybe this method can offer a fast-track path for cap exceeding next_window_cap * slots_per_window?

xinyuan-dev (approved):

PR #3267, opened 2026-09-16
txpool: forward smaller batches in sender nonce order

dshulyak:
large forwarding messages must be fully reassembled before the leader can process any of their transactions. smaller messages let the leader start processing sooner, without waiting for the rest of the batch.

if a burst of datagram loss prevents a message from being reassembled, smaller messages leave fewer transactions waiting for a retry.

partition each input list by sender and nonce with a 16 kib batch target. send larger valid transactions alone and rotate groups between batches.


dshulyak:
@claude

